PR #258, opened 2014-08-07
There's currently no advantage to using dual slow interleaved mode for a...

jablko:
...nalogRead()

See http://community.spark.io/t/why-dual-slow-interleaved/5550

### DIFF
--- a/inc/spark_wiring.h
+++ b/inc/spark_wiring.h
@@ -123,8 +123,7 @@
 #define SCL  1
 
 #define ADC1_DR_ADDRESS		((uint32_t)0x4001244C)
-#define ADC_DMA_BUFFERSIZE	10
-#define ADC_SAMPLING_TIME	ADC_SampleTime_7Cycles5	//Allowed values: 1.5, 7.5 and 13.5 for "Dual slow interleaved mode"
+#define ADC_SAMPLING_TIME	ADC_SampleTime_239Cycles5
 
 #define TIM_PWM_FREQ 500 //500Hz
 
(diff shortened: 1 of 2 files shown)


m-mcgowan:
ping @technobly

m-mcgowan:
@technobly  is this still relevant for the core?

technobly:
Absolutely swamped at the moment... but this is in our internal project tracking system currently, search ADC.  I'd suggest we add this to next Sprint.

This comment in the code that was deleted is particularly interesting: 
`#define ADC_SAMPLING_TIME ADC_SampleTime_7Cycles5 //Allowed values: 1.5, 7.5 and 13.5 for "Dual slow interleaved mode"`

technobly:
Added to Clubhouse 1664.  Let's work on this one soon!

jablko:
I think this issue is still relevant. Would you like me to rebase the PR on the latest firmware code?

technobly:
Agreed, it's definitely relevant :)  I've added it to Clubhouse which is our internal issue tracker.  This PR may be reopened and used as-is, with rebase, or we may take it in a different direction.  The goal will be to give the ADC attention and make sure it is performing in the best way (full resolution sampling, high impedance requirements, while still allowing the users to reconfigure it for faster sample rates with the trade off of lower impedance requirements).  Nothing to do now, but if you'd like to rebase the PR, that would be welcome.  Thanks @jablko !

technobly:
It should also be rebased against `develop`.

PR #264, opened 2014-08-12
Fix for https://github.com/spark/core-firmware/issues/260 "delay(1) not ...

mumblepins:
...as accurate as expected" 

I'll be honest, this is my first ever attempt at a pull request, so if I screwed this up, I'm sorry about that!

### DIFF
--- a/inc/spark_utilities.h
+++ b/inc/spark_utilities.h
@@ -43,8 +43,6 @@
 
 #define SPARK_LOOP_DELAY_MILLIS			1000	//1sec
 
-#define MIN_MILLIS_DELAY                        5       //switch to delayMicrosends if delay(n) is below this
-
 #define USER_VAR_MAX_COUNT				10
 #define USER_VAR_KEY_LENGTH				12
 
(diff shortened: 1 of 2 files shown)


m-mcgowan:
You're completely correct - using `micros()` wouldn't work well here. I reviewed the code again, and it's all good. The only point remaining is with waiting for the end to be reached

```
while (DWT->CYCCNT < DWT_END)
```

This will will immediately exit if DWT_END happens to wrap around. We can reformulate this using subtraction rather than comparison to avoid issues with wrap around:

```
while ((DWT->CYCNT-DWT_START)<(ms*SYSTEM_US_TICK))
```

m-mcgowan:
Just an update here - we have a fix for `micros()` so that it uses the full 32-bit value space and wraps around like `millis()`. https://github.com/spark/firmware/pull/493

m-mcgowan:
Thanks for this. I'm sorry I'm unable to merge this - the repo structure has changed in many ways since this PR was issued, and we're now working against develop. I'm going to write some unit tests to demonstrate the problem and then a fix based on the new `micros()` implementation, combined with your code above. I'll be sure to give you a shout out!

PR #300, opened 2014-09-10
Refactor/usb hal

satishgn:
All usb_xxx files and STM32 usb device specific code moved from core-firmware to core-common-lib


m-mcgowan:
@satishgn  - I believe you've already merged this into feature/hal, so closing this.

Issue #302, opened 2014-09-10
ability to duplicate the onboard LED color


m-mcgowan:
A fair few requests for having access to the current LED value for creating an external LED that mirrors the onboard led.

- https://community.spark.io/t/can-we-connect-an-rgb-led-directly-to-cores-status-led/6872
- https://community.spark.io/t/how-can-i-get-the-rgb-led-current-color/6772
- https://community.spark.io/t/providing-access-to-the-mode-button-and-rgb-led-from-gpio/3055

PR #303, opened 2014-09-10
led change notify

m-mcgowan:
This change addressses  #302.

User code can implement a function

``` cpp
void onChangeRGBLED(uint8_t r, uint8_t g, uint8_t b) {
// do stuff with the LED values
}
```

Which is called any time the LED changes color (either due to user code calling `RGB.color()` or the system updating the LED, e.g. breathing cyan.)

For example, this could be used to use analogWrite() to set PWM outputs to an external RGB led:

``` cpp
void onChangeRGBLED(uint8_t r, uint8_t g, uint8_t b) {
   analogWrite(D0, r);
   analogWrite(D1, g);
   analogWrite(D2, b);
}
```


m-mcgowan:
Refactored this and included for 0.4.0.

PR #314, opened 2014-09-17
Adding possibility for callback value registration with Spark::function

### LINKED ISSUE #313: parameterized function callbacks
At present, the function callbacks registratars (`attachInterrupt()` and `Spark.function()` et al.) take a function pointer. It's not presently possible to further parameterize the function called, for example, to customize the function behavior or which object is invoked at runtime.

There are other API designs where callback registrations allow you to specify an additional data parameter. This data parameter isn't used by the system but is simply passed back to the registered function when invokved, allowing the behaviour of the function to be parameterized at the point of registration.

A common use case is to be able to call a method on a class, which I sketched out in this thread. 

http://community.spark.io/t/how-to-point-spark-function-to-a-class-function/6986

rastapasta:
...allowing implementations of function calls to class members, should resolve #313

Allows following pattern as mentioned by @m-mcgowan on this thread: http://community.spark.io/t/how-to-point-spark-function-to-a-class-function/6986/4

```cpp
class MyClass {
   static int callback(String arg, void* pv) {
       return ((MyClass*)pv)->handle(arg);
   }

   virtual int handle(String arg) {
       return 42;
   }
}

MyClass my_class;
Spark.function("myfunc", MyClass::callback, &my_class);
```

### DIFF
--- a/inc/spark_utilities.h
+++ b/inc/spark_utilities.h
@@ -107,6 +107,7 @@ class SparkClass {
 public:
 	static void variable(const char *varKey, void *userVar, Spark_Data_TypeDef userVarType);
 	static void function(const char *funcKey, int (*pFunc)(String paramString));
+	static void function(const char *funcKey, int (*pFunc)(String paramString, void* pv), void* pv);
 	static void publish(const char *eventName);
 	static void publish(const char *eventName, const char *eventData);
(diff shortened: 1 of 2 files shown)


m-mcgowan:
I've used this code as inspiration for refactoring the way functions and variables are handled. Thanks for the contribution!

rastapasta:
<3

PR #318, opened 2014-09-22
C++11 and attach interrupts using lambdas

lbt:
I wanted to support multiple instances of a class in a library which uses attachInterrupt

I can now do:
    auto mylambda = [&]() {this->touchSense();} ;
    attachInterrupt(m_sensorPin, mylambda , RISING);

Standard function pointers still work.

I needed some cleanups to allow the entire library to build using gnu++11 and some minimal tidy up to allow <functional> to be #included.

### DIFF
--- a/build/makefile
+++ b/build/makefile
@@ -48,10 +48,12 @@ SRC_PATH = $(PROJECT_DIR)
 # ensure defined
 USRSRC += ""
 
+# The entire firmware is gnu+11 safe
+CPPFLAGS += -std=gnu++11
+
 # determine where user sources are, relative to project root
 ifdef APP
 USRSRC = applications/$(APP)/
-CPPFLAGS += -std=gnu++11
 
 ifndef TARGET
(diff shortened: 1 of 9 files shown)


m-mcgowan:
Just a note to say we've not forgotten about this, and I'd love to merge this to coincide with the photon release. 

I think using std::function also makes it easier to wrap user code for system management of user functions (as in issue #257)

m-mcgowan:
Still keeping an eye on this! The photon repo has moved far ahead of the master branch, so I won't merge this commit, but will borrow from it and factor it into the new repo structure. (We are promised the necessary licenses to distribute the photon repo at the end of this month.)

I will add lambda support in Wiring, which is all C++, which then calls the C HAL api that uses a function pointer and void* data object.

m-mcgowan:
I've pulled the parts that were needed from here, thanks for the inspiration to incorporate lambdas!

PR #319, opened 2014-09-24
Fix PINMAP for BTN. Was pointing to LED4

geekbozu:
Updated STM32_Pin_Info PIN_MAP[TOTAL_PINS] to have the proper info for the 20th Pin which has been defined as BTN. 

Previously was LED4 according to the schematic and local debugging.


m-mcgowan:
Hey @geekbozu! Good catch! Thanks for spotting this. I've implemented this in `develop` against the new HAL repo structure.

PR #334, opened 2014-10-24
Fix for Time.zone issue.

kennethlimcp:
This will fix https://github.com/spark/firmware/issues/280 for:

1.) Changed `Get_UnixTime` to factory in the **time_zone_cache** parameter

2.) Change the `Time.zone()` formula to multiple GMT_offset by 1800 instead 0f 3600 as tests showed that the original increase/decrease 2 hours per UTC

Sample code for test verification is here: https://gist.github.com/kennethlimcp/fe25df7bc1619654ff76

### DIFF
--- a/wiring/src/spark_wiring_time.cpp
+++ b/wiring/src/spark_wiring_time.cpp
@@ -158,11 +158,8 @@ static void Set_CalendarTime(struct tm calendar_time)
 /* Refresh Unix/RTC time cache */
 static void Refresh_UnixTime_Cache(time_t unix_time)
 {
-	if(unix_time != unix_time_cache)
-	{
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Just looked over this now - the changes are that the cache value is not checked and the time refreshed on each call. Can someone say what problem this fixes? I would like to see a unit test to illustrate the problem and prove the fix.

I don't see why we need to avoid the cache. Instead we can invalidate the cache when the timezone is changed, in addition to when the time changes, since the time and timezone make up the data used to generate the cached calendar value.

m-mcgowan:
This has been recently raised again, and I fixed it in #508. Thanks for the contribution!

PR #337, opened 2014-10-27
Added support for enabling just SWD without JTAG

elcojacobs:
This pull request adds the option USE_SWD, which enables SWD debugging, but keeps JTAG disabled.
This is useful when the JTAG pins are already in use.

To compile with SWD enabled, JTAG disabled, run:

    make clean all USE_SWD=y

The ST-Link V2 works in combination with ST-util with a very simple circuit, directly connecting SWDIO, CLK, VDD, GND and RESET. I'll post the details on the forum:
https://community.spark.io/t/working-jtag-programming-and-debugging-from-netbeans/7679


zsup:
Don't forget to document!



—
Zach Supalla
312.953.3413

m-mcgowan:
Sure! :+1: where to put it? AFAIK there's no docs for the existing `USE_SWD_JTAG` flag, just posts on discourse. 

I wonder if we should have an Advanced Firmware section in the docs site? It could cover topics like debugging, best practices using interrupts, threading (photon), etc...  

At present, the build system docs are stored as markdown in the git repo. I can put the new flag there if that's suitable.

zsup:
Good question. It does feel like there is room for an 'advanced firmware' section on the docs, or on the guide/tutorial that we will be starting to work on this sprint. In the meantime if there is an existing place for build flags, that should do the trick for now



—
Zach Supalla
312.953.3413

PR #354, opened 2014-11-30
Add in description for device id printout

kennethlimcp:
Seems like this line was took out after the hal branch implementation:

https://github.com/spark/firmware/blob/master/src/wifi_credentials_reader.cpp#L80

Changed to "device" for future hardware and not only core.


kennethlimcp:
fixed in this commit: https://github.com/spark/firmware/commit/d7079dfc608de89d134316db01fbbd9aa1b81e47#diff-6217688b0e210a067a90d2cc775ae4a3R80

m-mcgowan:
:+1 Coincidence we both fixed this at the same time! :-)

PR #355, opened 2014-11-30
Printout mac address in listening mode

kennethlimcp:
This will help new users get the mac address simply via Listening mode "i". 

Community issue: https://community.spark.io/t/getting-mac-address-of-unconnected-core/8473


kennethlimcp:
implemented in this commit: https://github.com/spark/firmware/commit/d7079dfc608de89d134316db01fbbd9aa1b81e47#diff-6217688b0e210a067a90d2cc775ae4a3R80

Issue #357, opened 2014-12-02
Connect should have retry limits


harrisonhjones:
As per this comment: http://community.spark.io/t/semi-automatic-mode-question/7855/29 it would be nice if we could call connect(retryCount) to have the core attempt to connect to the cloud retryCount times and then return. A user could then check the status of the connection to determine if it had indeed succeeded to failed

m-mcgowan:
I'm going to close this since the discussion is happening in 594.

PR #367, opened 2014-12-20
Fix for #366

### LINKED ISSUE #366: MANUAL mode should return control to the user loop if Cloud is not connected
As per issue #357, wherein connect should have a retry limit, and issue #347 (to a certain degree) the current mode handling is a bit odd, both in terms of actuality as well as documentation.    The current way the main() loop works is that it if `Spark.connect()` is called, which sets the `SPARK_CLOUD_CONNECT = 1` flag, the `SPARK_WLAN_Loop()` call in the beginning of the `main()` loop will force a blockage of the user `loop()`.  This appears to be solely due to this line:
```c++
if(!SPARK_WLAN_SETUP || SPARK_WLAN_SLEEP || !SPARK_CLOUD_CONNECT || SPARK_CLOUD_CONNECTED || SPARK_WIRING_APPLICATION)
```
which must be true to proceed to the `setup()` or `loop()` user code sections.  This also maybe should pass if mode is SEMI_AUTOMATIC.

mumblepins:
I believe this should be allow for a Spark.connect() call to only block for one connection attempt.  Should fix #366, and #357.  Might also need to add a `|| System.mode() == SEMI_AUTOMATIC`, but that maybe should be a debate about how "automatic" semi automatic should be.


m-mcgowan:
After a break, I'm looking at this again in a new light, and am a bit puzzled. Could you explain how the fix avoids user code from being blocked? 

As I see it, once `setup()` has been called, `SPARK_WIRING_APPLICATION` is set to true, and consequently that condition to execute user code is always true - without the need for a system mode check. The blocking is happening in `SPARK_WLAN_Loop()` which is attempting to re-connect wifi and cloud, and that's independent from these flags.

Can you help me understand how this ensures the connection is only made once - is there more code that's expected to be used in conjunction with this?

mumblepins:
Ah! You're right.  I didn't see that before.  However, I also can't find anything that should block in the SPARK_WLAN_Loop() for cloud connection.  I realize that the Wifi connect is blocking for a short while, but my understanding of tracing the code is that the Cloud connection only gets tried once in SPARK_WLAN_Loop.  

The really confusing bit is that this fix seems to work, without any user code used with it.  I admit to being very confused now...

m-mcgowan:
This sprint the firmware team will be working on reducing the times users code is blocked, so we'll have chance to investigate this fix.

m-mcgowan:
Thanks! I added a similar check in the develop branch. This will catch the case where the cloud tries to reconnect but fails, so it will try once and then run the user loop, then try again, and continue alternating like that until either the loop calls `Spark.disconnect()` or the cloud connects.

Issue #347, opened 2014-11-08
Spark.process() should call SPARK_WLAN_Loop()


towynlin:
Per the discussion here:
https://github.com/spark/docs/pull/242

We should make `Spark.process()` call `SPARK_WLAN_Loop()` rather than the other way around.

m-mcgowan:
Closing - implemented and working in feature/hal branch.

PR #372, opened 2015-01-02
Implementation as discussed for issue #347.

m-mcgowan:
- `Spark.publish()` now runs the full background processing - this is wifi/clout connection management, plus pumping cloud events
- `SPARK_WLAN_Loop()` renamed to `Spark_Idle()`. The old function name is still available, but deprecated.
- `delay()` does not run the background loop when in manual mode, nor when in semi automatic mode and not connected. The rationale being that in manual mode, users are given full control over cloud processing, while in semi-automatic mode, the promise is to maintain the cloud connection once it is established.

### DIFF
--- a/main/src/main.cpp
+++ b/main/src/main.cpp
@@ -212,7 +212,7 @@ void app_setup_and_loop(void)
     while (1)
     {
         DECLARE_SYS_HEALTH(ENTERED_WLAN_Loop);
-        SPARK_WLAN_Loop();
+        Spark_Idle();
 
         static uint8_t SPARK_WIRING_APPLICATION = 0;
         if(SPARK_WLAN_SLEEP || !SPARK_CLOUD_CONNECT || SPARK_CLOUD_CONNECTED || SPARK_WIRING_APPLICATION)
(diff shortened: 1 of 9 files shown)


zsup:
:+1:

If we are deprecating WLAN_Loop, is there anything we should do to provide understandable compiler errors for people who are using it in their code?



—
Zach Supalla
312.953.3413

m-mcgowan:
Thanks for the prod! I did some hunting and found that gcc does support deprecation messages, so I added a pointer to use `Spark.process()`. 

The message looks like this - not uber friendly, but does point the user in the right direction:

`
tests/core-v1/testapp1/cloud.cpp:7:21: error: 'void SPARK_WLAN_Loop()' is deprecated (declared at ../wiring/inc/spark_wlan.h:48): Please use Spark.process() instead. [-Werror=deprecated-declarations]
     SPARK_WLAN_Loop();  
`

zsup:
Not the prettiest message I've ever seen, but I'll take it over nothing :-)

m-mcgowan:
Closing this - it's been merged in the photon repo.

Issue #362, opened 2014-12-06
LED user control error


robertdolca:
When the download of a new firmware starts the LED is not being reinitialized. The reinitialization is being done only when the flashing starts.

This happens only when the user takes control of the LED.

m-mcgowan:
The same problem also occurs when entering listening mode. 

The fix is to not use `RGB.color()` calls in system code, but instead to set a system level value that unconditionally returns control to the system for the LED color. (E.g. a system active flag - off by default for regular background LED notifications, so the user is free to take control of the led, set to on when the system wishes to take control, which overrides any user settings.)

m-mcgowan:
I just encountered this again. The fix is now clear to me: at present, the user LED color always overrides the system assigned color, so that when entering OTA or setup mode, the flash sequence is correct, but not the color.  

We will split the system LED notifications into two kinds - background and important - with important notifications always overriding the user-assigned color, while background notifications show only when the user hasn't overridden the LED color. 

Important notifications  - these appear always:
- connecting to WiFi (flashing green)
- connecting to the cloud (flashing cyan)
- setup mode (slow flashing blue)
- OTA in progress (flashing magenta)

background notifications - these appear when the user hasn't taken control of the LED:
- resting, WiFI off (breathing White)
- resting, WiFi on (breathing blue)
- connected WiFI (breathing green)
- connected cloud (breathing cyan)

m-mcgowan:
There is a race condition between code in the main thread setting the LED color and the SysTick interrupt. Occasionally I've seen that the LED color is incorrect, e.g. flashing green during setup mode, or slow flashing blue in other cases, due to interplay between the main thread and the SysTick interrupt both setting the LED color.

Avoiding this, and also opening the door to multithreaded LED control can be done by updating the LED hardware only from the SysTick interrupt, rather than from the main thread. This can be conveniently done by having the main thread call a signalling service to set the desired LED state. The signalling service manages the LED state, including the priority scheme presented above above. 

This will encapsulate in one place all the LED logic, avoiding race conditions, also open up to more flexibility for the user in assigning the LED color, such as use of custom fades/flashes by the user.

m-mcgowan:
Fixed!

PR #373, opened 2015-01-03
fix for #362. 

m-mcgowan:
`RGB.control()` was being called (to return control of the RGB led to the system) but called with the wrong value! :-)


m-mcgowan:
D'oh - looking at a different function. closing.

PR #383, opened 2015-01-19
use default arguments to reduce number of method overloads required.

### LINKED ISSUE #376: Spark.publish("event",PRIVATE) version
Needs a constructor that takes two arguments, not the 4-argument complexity. Private should be used frequently, but this deters novices (who are most likely to spam) from making them so. TTL's a bit out of reach for the n00bs, and sometimes you just want to flip your events to private without adding other bits.

m-mcgowan:
Fixes #376


m-mcgowan:
closing - replicated in feature/hal (working repo)

PR #386, opened 2015-01-24
Add #define for ADC resolution

zfields:
Now I can programmatically determine voltage when you decide to change to a 10-bit ADC.
`3.3V * (analogRead(A0) / ADC_MAX_VALUE)`
-- or --
If you want to implement [analogReadResolution()](http://arduino.cc/en/Reference/AnalogReadResolution)
`#define ADC_RESOLUTION_BITS 12`
`3.3V * (analogRead(A0) / ((1 << ADC_RESOLUTION_BITS) - 1))`

### DIFF
--- a/inc/spark_wiring.h
+++ b/inc/spark_wiring.h
@@ -126,6 +126,7 @@
 
 #define ADC1_DR_ADDRESS		((uint32_t)0x4001244C)
 #define ADC_DMA_BUFFERSIZE	10
+#define ADC_MAX_VALUE	4095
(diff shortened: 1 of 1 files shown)


zfields:
Ping.

m-mcgowan:
Hi there! The firmware development is mainly happening in the `feature/hal` branch.  To accommodate future hardware, it would be best if the ADC resolution was specified as a function in `hal/inc/adc_hal.h`, A function like this would be good.

```
int HAL_ADC_Resolution();
```

PR #413, opened 2015-04-16
Adding support for custom serial bit parity

jmekler:
Updates to the HAL branch to add support for other serial modes enabled on the STM32F103 based on discussions in the online forum: http://community.spark.io/t/custom-serial-bit-parity-with-spark-core-e-g-serial-8n2/10675.

Syntax follows the Arduino standard. For example, if you wanted communicate with a device requiring 8 data bits, 2 stop bits, and even parity @ 9600baud on Serial1, you would initialize as follows:

     Serial1.begin(9600, SERIAL_8E2);

Define statements are included for the most common 8 data bit serial modes (none/even/odd parity and 1/2 stop bits). The code will throw a compile error if the user tries to compile with a predefined configuration that isn't supported/defined for the device (e.g. SERIAL_5N1 on the Spark Core). However, the code will fail silently if a user manually enters an unsupported serial configuration (e.g. `0b00000000` = `SERIAL_5N1`). 

The STM32 also supports some more advanced features, such as flow control (currently disabled in `hal/src/core-v1/usart_hal.c`, though potentially supported if the pinouts are exposed). The full set of features can be found here: `platform/MCU/STM32F1xx/STM32_StdPeriph_Driver/inc`.

### DIFF
--- a/hal/inc/usart_hal.h
+++ b/hal/inc/usart_hal.h
@@ -50,8 +50,6 @@ typedef enum HAL_USART_Serial {
 /* Exported constants --------------------------------------------------------*/
 
 /* Exported macros -----------------------------------------------------------*/
-// IS_USART_CONFIG_VALID (return true for 8 data bit, no flow control, any parity, any stop byte configurations
-#define IS_USART_CONFIG_VALID(CONFIG) ( ((CONFIG>>2) != 0b11) && ((CONFIG>>4)==0b11) && ((CONFIG >> 6)==0b00) )
 
 /* Exported functions --------------------------------------------------------*/
 
--- a/hal/inc/usart_hal.h
+++ b/hal/inc/usart_hal.h
@@ -50,6 +50,8 @@ typedef enum HAL_USART_Serial {
 /* Exported constants --------------------------------------------------------*/
 
 /* Exported macros -----------------------------------------------------------*/
+// IS_USART_CONFIG_VALID (return true for 8 data bit, no flow control, any parity, any stop byte configurations
+#define IS_USART_CONFIG_VALID(CONFIG) ( ((CONFIG>>2) != 0b11) && ((CONFIG>>4)==0b11) && ((CONFIG >> 6)==0b00) )
 
 /* Exported functions --------------------------------------------------------*/
 
@@ -58,7 +60,7 @@ extern "C" {
 #endif
 
 void HAL_USART_Init(HAL_USART_Serial serial, Ring_Buffer *rx_buffer, Ring_Buffer *tx_buffer);
-void HAL_USART_Begin(HAL_USART_Serial serial, uint32_t baud);
+void HAL_USART_Begin(HAL_USART_Serial serial, uint32_t baud, uint8_t config);
 void HAL_USART_End(HAL_USART_Serial serial);
 uint32_t HAL_USART_Write_Data(HAL_USART_Serial serial, uint8_t data);
 int32_t HAL_USART_Available_Data(HAL_USART_Serial serial);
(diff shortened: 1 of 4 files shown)


m-mcgowan:
Thanks for this! Could you please rebase this on the develop branch since that's where we are working for the next release. 

If you have any thoughts on how to implement this for the photon, please share!

romerod:
I need the same modification, why do you ( @m-mcgowan ) ask how to implement it for the photon, in hal>src>photon i don't find the usart_hal.c file, isn't the hal>src>core>usart_hal.c file used for the photon too?

technobly:
@romerod yes that's the common hal for all stm32f2xx devices, photon/p1/electron

I think Mat just would like to offer you the chance to implement the same code for the photon since this is fresh in your head and you probably have a good test setup already. If this interests you and you have time, it would be super appreciated by us. If not that's totally cool too, thanks so much for your contributions!

technobly:
I meant to reply to both @jmekler and @romerod ;-)

clay-to-n:
Very interested in this :+1:

dseemakurty:
Can we please get custom serial parity functioning on the photon? My product development is stalled due to the lack of even parity. Any help is appreciated!

jmekler:
@romerod @technobly @m-mcgowan I'm happy to move this implementation over to the develop branch and incorporate photon support. This wouldn't address some of @technobly's feature requests in #697, but would at least give folks an Arduino-style API to access other serial configurations.

I don't see any functional differences b/t the photon and core since they both use hal/src/stm32f2xx/usart_hal.c. So I'll just implement the same SERIAL_XXX options for the two platform IDs.

Anything I'm missing? It looks like the electron has a lot more Serial options. I'll leave those for another PR.

m-mcgowan:
That's great to hear!  Just a small correction - the Core doesn't use the stm32f2xx directory, but does use "core" and "stm32" directories. If the code is the same for the core and photon you could put it in the "stm32" folder and have the code shared between both platforms rather than duplicating it. I hope that helps!

jmekler:
Thanks @m-mcgowan! 

High-level USART functionality is the same for the core (STM32F1) and photon (STM32F2) – see pg. 10 on http://www.st.com/web/en/resource/technical/document/application_note/DM00033267.pdf. However, they do have different include statements and the STM32F2 version of usart_hal.c is also used by the Electron. I think it might be best to leave them as separate implementations for now rather than getting too sneaky with conditional macros.

Barring any disagreement, I'll make those changes for the core and stm32f2xx, test it on a photon/core, and then close out this PR and initiate a new PR on the develop branch.

clay-to-n:
Any progress on this, @jmekler?  A project I'm working on really needs this, so if it's no longer a priority for you I might be able to see if I can get this implemented in the Photon in the next few weeks.

I appreciate the work you've all done so far on this issue and the platform as a whole!

mirekhk:
I am interested in this project as well. Great work done so far. It seems that all builds were fine and there was only problem with final check. Do I understand it correct? Would it be possible to find what was wrong with checks?

clay-to-n:
@mirekhk, it sounds like the code in this PR is working on the Spark Core but not the Photon yet, but the process of making it work on Photon sounds straightforward.

mirekhk:
@clay-to-n, thx a lot for good news. I will test it over weekend. I have 0.4.7 version on my core and I have tried to find a confirmation in the source code, but i was not successful. Can you confirm that it is included in 0.4.7?

m-mcgowan:
USART parity was implemented in #757, released in 0.5.0. Thanks for the contribution!

PR #423, opened 2015-05-21
Added EXTRA_CFLAGS to the build make file

prices:
This allows extra CFLAGS to be added when the make command is called.  Much of what I want integrate into your firmware for my project requires certain constants to be set in the CFLAGS when they are compiled.  This will make that easy for me to accomplish.

### DIFF
--- a/build/makefile
+++ b/build/makefile
@@ -138,6 +138,8 @@ else
 CFLAGS += -DRELEASE_BUILD
 endif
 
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Thanks for this! Could you rebase this on to develop please, since that's where we are preparing the next release.

m-mcgowan:
merged #424, which replaces this one.

PR #424, opened 2015-05-22
Added the a hook to EXTRA_CFLAGS in user/build.mk. 

prices:
This allows the user code to be built with extra CFLAGS.  The user just has to set the shell variable EXTRA_CFLAGS and they will get added.

Something like:

export EXTRA_CFLAGS="flags"; make APP=userapp


m-mcgowan:
Thanks for rebasing. I've not tested this, but I hope it scratches the itch you had.

PR #425, opened 2015-05-24
Added BSSID scan result

kennethlimcp:
Sample output:
```
GOOD DAY, WIFI TESTER AT YOUR SERVICE!!!
WIFI_SCAN:;checking command: WIFI_SCAN:;
SCAN_START
xxx,-73,0x08:0x0F:0x0E:0x0D:0x03:0x04
kenken,-45,0x0C:0x05:0x02:0x01:0x06:0x02
xxx,-76,0x06:0x06:0x09:0x03:0x0C:0x05
xxxxxx,-76,0x02:0x00:0x0F:0x02:0x06:0x09
xxxx,-58,0x00:0x0C:0x06:0x0C:0x09:0x02
SCAN_STOP
```

@mdma, seems like the Atom editor i'm using stripped the tabs/spaces. Let me know if that's not ok and i will try to re-do it again!


m-mcgowan:
The changes look good to me (would have been easier to see the actual changes without all the whitespace changes! :-P)

Rather than printing out 0x in front of each octet, it's more customary to just print the octet, separated by a colon. E.g. `08:0F:0E:0D:03:04` for the first example.

In `wlan_scan_callback` in wifitester.cpp, check that ssid is not NULL before copying with memcpy. (This is my code, so I should have done this, just an oversight.)

kennethlimcp:
Just to confirm the syntax should be: (https://github.com/spark/firmware/blob/develop/system/src/wifitester.cpp#L163)

```
if(ssid != NULL){
   ------
    }
```

Will checking `ssid_len` for `0` be doing the same thing?

kennethlimcp:
@m-mcgowan it's done :)

```
SCAN_START
00:00:00:00:00:00,kenken,-58
00:00:00:00:00:00,Yiiiii,-84
00:00:00:00:00:00,1008,-63
SCAN_STOP
```
Spoofing the BSSID and not an error

m-mcgowan:
Cool,are the BSSIDs correct? Or is that just an example?

kennethlimcp:
@m-mcgowan , they are correct on my side. (i.e not 00:00:00:00) :)

system/src/wifitester.cpp
```diff
@@ -156,23 +156,36 @@ void WiFiTester::printInfo() {
...
 {
     WiFiTester& tester = *(WiFiTester*)data;
-    char str_ssid[33];
-    memcpy(str_ssid, ssid, ssid_len);
-    str_ssid[ssid_len] = 0;
-    tester.serialPrint(str_ssid);
-    tester.serialPrint(",");
-    itoa(rssi, str_ssid, 10);
-    tester.serialPrintln(str_ssid);
+    if(ssid != NULL){
```
m-mcgowan: When ssid is NULL you can still use bssid, just not ssid. So the code that deails with bssid can come out of the loop. When ssid is null, just set str_ssid to an empty string: `str_ssid[0] = 0;`

kennethlimcp: ok got it! Might need to wait for more tests. reading BSSID with no SSID has some hang issue in our testst

m-mcgowan:
Hi there @kennethlimcp!  Any status updates on this? Is it still something you'd like to merge? Something to keep in mind is that we will have the WiFi.scan() API soon if that's a good alternative.

kennethlimcp:
@m-mcgowan i'm happy to close this and rebase once the scan() API is available.

Thanks! I'm sure it will be useful for people since identifying various AP will be useful. ;)

m-mcgowan:
We now have WiFi.scan() which provides the BSSID as part of the result, so I'm going to close out this PR without merging. I hope the new API gives you what you ned.

PR #429, opened 2015-05-28
Feature/eeprom update

satishgn:
New EEPROM Interface Methods (FIRM-27)


m-mcgowan:
Nice! Could you please also add API tests under `user/tests/wiring/api`.

PR #433, opened 2015-05-31
set IWDG_SYSTEM_RESET to 0

kennethlimcp:
- currently not used even in the new core bootloader
- prevents breathing white state when using SEMI_AUTO/MANUAL with latest firmware
- fixes https://github.com/spark/firmware-private/issues/99

### DIFF
--- a/hal/src/core/core_hal.c
+++ b/hal/src/core/core_hal.c
@@ -103,7 +103,7 @@ void HAL_Core_Config(void)
 	if (RCC_GetFlagStatus(RCC_FLAG_IWDGRST) != RESET)
 	{
 		/* IWDGRST flag set */
-		IWDG_SYSTEM_RESET = 1;
+		IWDG_SYSTEM_RESET = 0;
 
 		/* Clear reset flags */
 		RCC_ClearFlag();


m-mcgowan:
@satishgn  - could you review this please? The combinations for when we need and don't need IWDG is too much for me to remember, combined with new and old bootloaders. It would be great if you could document IWDG after looking at this so the current situation is clear for all.

technobly:
@kennethlimcp I'm not understanding what a watchdog timeout reset has to do with root cause of the LED breathing white when it should breathe blue, when the WiFi connection is cut off at the router in your referenced issue: spark/firmware-private#99 and why resetting a flag that is supposed to be set to indicate we have booted from a watchdog timeout is an appropriate fix for this.

Would you please elaborate in both areas?

kennethlimcp:
Look at this line: https://github.com/spark/firmware/blob/develop/system/src/main.cpp#L218.

I'm not sure who decided that the var `IWDG_SYSTEM_RESET` should be set to 1 and this causes that line to fail once wifi is cut off in SEMI_AUTO/MANUAL mode.

Dont't see that var used in anywhere. Anyways, hook up a debugger and try out the behavior. You will notice that the core gets skips the entire user code due to that line above.

technobly:
Right, this all actually makes sense.  The function HAL_watchdog_reset_flagged() let's you know the system reset from a watchdog timeout.. I.e., some hard loop preventing code execution caused a system reset.  That condition should not exist in user code, and if it does it and the system resets, user code is not executed.  That's the way it's implemented.

Honestly I think user code should continue to be executed, and the system LED should be a different color to indicated we are running after a watchdog timeout.  Or provide system events that tell us all of the details, like how many times we've reset from a watchdog timeout.

That said, I was pretty sure the watchdog timer is still being cleared in the system tick ISR handler so there really shouldn't ever be a watchdog timeout happening.

All that said, what is causing a watchdog timeout to occur when you cut off the WiFi?

By changing IWDG_SYSTEM_RESET to equal 0, you are saying... "no, the system didn't boot up after a watchdog reset" but it did, and that's why we set IWDG_SYSTEM_RESET to 1.  So we need to figure out why a watchdog timeout is occurring because that's not something that should be happening with normal operation of system code.

kennethlimcp:
The key point is... That var is not used/set anywhere except the HAL function for core.

technobly:
Well that's a good point but I believe not the key point, and sheds light on the fact that we should probably initialize IWDG_SYSTEM_RESET to 0 instead of assuming this crucial variable is initialized to 0 by the system.  We could initialize it here to 0: https://github.com/spark/firmware/blob/develop/hal/src/core/core_hal.c#L45

The key point I believe is that it does have a use: https://github.com/spark/firmware/blob/develop/hal/src/core/core_hal.c#L356

Maybe not the use I would like (blocking user code when a watchdog timeout occurs)

But stubbing it out with a 0 where you suggested causes it to not be useful for any purpose.

This PR is trying to fix a problem you are having with WiFi cutting out, and ultimately changing the main loop to just say `if(!SPARK_FLASH_UPDATE) { // execute setup() and loop() }` would be what you really want this PR to be.  However, it's still ignoring the fact that something in the system code is causing a watchdog timeout when that doesn't even seem possible (if the watchdog timer is being reset via ISR still).  YEP: https://github.com/spark/firmware/blob/develop/system/src/main.cpp#L151

---

It's funny really, because somehow the Photon code does effectively stub this out: https://github.com/spark/firmware/blob/develop/hal/src/photon/watchdog_hal.c#L48  Perhaps there is no watchdog timer on the Photon?  I find that hard to believe because the WICED platform requires a platform_watchdog_kick(); function to be defined.  I've been meaning to figure this out and add more detail to those comments.

m-mcgowan:
I did reproduce the issue, and seems that shutting down WiFi causes the CC3000 to stall in a DMA transaction. I feel the best solution is to remove the watchdog check in the main loop, since we don't really use the watchdog anymore (the intent is that the timer is always reset in the SysTick so it never fires.)

technobly:
There is a lot going on in this PR :-) let's continue the discussion over in the issue 99 thread. https://github.com/spark/firmware-private/issues/99

PR #436, opened 2015-06-02
newehal --> newhal

kennethlimcp:


technobly:
Thanks! I was trying not to add more typos than I was fixing ;-)

PR #438, opened 2015-06-02
minor doc tweaks

kennethlimcp:


m-mcgowan:
ca means "circa" or "about". sure, I should have added the abbreviation mark after, so "ca. 3000 byte reduction". The figures are approximate since I was doing it from memory (and it will change as the firmware moves forward.)

PR #442, opened 2015-06-05
remove obsolete command

kennethlimcp:


m-mcgowan:
I'm a bit confused - brew has worked for me, that's how I got 4.9 on my machine.

kennethlimcp:
@m-mcgowan, give it a try again with the same command.

technobly:
I just upgraded recently to 4.9.3 using the `upgrade` version of this command. I did have to remove my old sim links and add new ones using prompted commands, but it worked fine.

technobly:
I think perhaps you didn't have Homebrew installed?  I added more info to the dependencies page:
https://github.com/spark/firmware/commit/db018f69127f527a334abf5763e0b290dbdbd268

kennethlimcp:
I did this:

- `brew update`
-  `brew install gcc-arm-none-eabi-49`

```
Error: No available formula for gcc-arm-none-eabi-49 
Searching formulae...
Searching taps...
```

I have `brew` installed for sure and this shouldn't be closed since i'm not the only one with the installation issue.

technobly:
I only closed the PR because I know the answer is not to remove the brew commands completely... we can definitely keep discussing your issue here though.  Did you get this resolved by adding the BREW TAP?  See latest docs now:
https://github.com/spark/firmware/blob/develop/docs/dependencies.md#1-download-and-install-dependencies

PR #447, opened 2015-06-09
Windows Driver manufacturer fix

harrisonhjones:
Fixed manufacturer string. Updated Spark Core Manufacturer to Particle.io

This was a joint effort by myself, @kennethlimcp, and @ScruffR

### DIFF
--- a/misc/driver/windows/photon.inf
+++ b/misc/driver/windows/photon.inf
@@ -12,7 +12,7 @@ DriverVer=01/10/2014,5.2.0.0
 CatalogFile=photon.cat
 
 [Manufacturer] 
-%DEVICE_DESCRIPTION%=DeviceList, NTx86, NTia64, NTamd64, NTARM
+%PROVIDER%=DeviceList, NTx86, NTia64, NTamd64, NTARM
 
 [DeviceList.NTx86] 
 %DEVICE_DESCRIPTION%=DriverInstall, USB\VID_2B04&PID_C006
--- a/misc/driver/windows/spark_core.inf
+++ b/misc/driver/windows/spark_core.inf
@@ -12,7 +12,7 @@ DriverVer=01/10/2014,5.2.0.0
 CatalogFile=spark_core.cat
 
 [Manufacturer] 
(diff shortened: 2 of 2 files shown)


m-mcgowan:
Not only did the manufacturer string change (Which is fine in itself) but also DEVICE_DESCRIPTION has changed to PROVIDER in the config. Can someone explain what that does/why it's necessary?

kennethlimcp:
@m-mcgowan,

It all depends on what Particle IO wants the manufacturer name to be..

If we use `DEVICE_DESCRIPTION` for the "MANUFACTURER" field in Windows:
- Spark Core --> Spark Core
- Photon --> Photon

It should be either `Spark Devices` `Particle IO` or whatever the company wants to be known as.

This https://github.com/spark/firmware/pull/447/files#diff-84eb7381a951faef223f48e35068bf2aL88 may not be required if the Manufacturer for Spark Core is listed as `Spark Devices`

TLDR: The `[Manufacturer] ` is not taking the correct string

ScruffR:
If you're refering to the change of `%DEVICE_DESCRIPTION%` into `%PROVIDER%` that was the whole solution, since the `%...%` notation is sort of a forward reference to the actual string that should be used as value descriptor under the [Manufacturer] section.
It's only a placeholder for the value which is given to `PROVIDER` further down the file, rather than the one of `DEVICE_DESCRIPTIONA`.

m-mcgowan:
Don't we then need similar changes to the photon.inf file?

kennethlimcp:
This fix changes both `.inf` file for `[Manufacturer]` to take `%PROVIDER%`

For `photon.inf`, --> https://github.com/harrisonhjones/firmware/blob/develop/misc/driver/windows/photon.inf#L88
already set to `Particle IO`

For `spark_core.inf`, `PROVIDER` is updated to `Particle IO`

technobly:
@m-mcgowan @harrisonhjones @kennethlimcp @ScruffR Great collab on this guys!  FYI: Just confirmed we are doing business as just "Particle", no " IO" at the end.

harrisonhjones:
Thanks @technobly, see #450

PR #449, opened 2015-06-09
fast pin access

m-mcgowan:
This covers fast pin set/reset. Does anyone have a suggestion for multiple pin set/reset, port access etc..?

### DIFF
--- a/user/inc/application.h
+++ b/user/inc/application.h
@@ -47,7 +47,7 @@
 #include "spark_wiring_servo.h"
 #include "spark_wiring_wifi.h"
 #include "spark_wiring_network.h"
-#include "spark_wiring_client.h"  
+#include "spark_wiring_client.h"
 #include "spark_wiring_tcpclient.h"
 #include "spark_wiring_tcpserver.h"
 #include "spark_wiring_udp.h"
@@ -56,6 +56,8 @@
 #include "spark_wiring_eeprom.h"
 #include "spark_wiring_version.h"
 #include "spark_wiring_thread.h"
+#include "fast_pin.h"
+
 
 // this was being implicitly pulled in by some of the other headers
 // adding here for backwards compatibility.
--- a/wiring/inc/fast_pin.h
+++ b/wiring/inc/fast_pin.h
@@ -35,12 +35,12 @@ extern "C" {
 
 #ifdef STM32F10X
 
-inline void fastPinSet(pin_t _pin)
+inline void pinSetFast(pin_t _pin)
 {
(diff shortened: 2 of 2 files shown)


satishgn:
Yes. Bit-banding is available on ARM Cortex-M3 which is the fastest/atomic way to set/reset IO pins.
Please feel free to assign it to me. I can take this up in my spare time after hardware design.
For more info on this: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0337h/Behcjiic.html

m-mcgowan:
Good call satish - right now focus is primarily on the API (as you see the implementation is simple), when that API settles down feel free to add a bit-banding implementation if you wish. I'm curious exactly how much faster bit banding will be and how fast do we need it to be?

ScruffR:
May be a bit late, but for consistency with `digitalWriteFast()` would not `pinSetFast()`/`pinResetFast()` be "nicer" ;-)

technobly:
Yep, Mat made them all consistent with Fast at the end.

I'm pretty excited about this bit-banding thing... have not experimented with that but it sounds super powerful for fast access.

pkourany:
Oh, I like the bit banding.  Nice catch @satishgn.  @ScruffR, the old way requires a lot of bit operations whereas with banding, it a simple RAM offset.  To the user, it would be transparent via @m-mcgowan 's API but under the hood it would scream!  @satishgn, I would love to see your propose implementation.

m-mcgowan:
We can expand with port writes as another PR. Is everyone happy with this as it is so it can be included for the imminent 0.4.2 release?

pkourany:
I agree, do as another PR

technobly:
Has anyone really tested this? Personally I have not had the time but really want to before we release it.

ScruffR:
@technobly, tested what? Bit-banding or my `portSetMasked()`?
The latter is tested on the Core and the books tell that this should work on the Photon just the same, but I haven't tested it - due to lack if HW ;-)

@m-mcgowan, I'm fine with that :+1:

technobly:
Tested the speed of inline functions setting and clearing GPIO, vs macros we are currently using. Anyone have a scope besides me?

pkourany:
@technobly, I have a scope and logic analyzer but I can only test late tonight or tomorrow.

technobly:
Easiest thing to do is look at v006 of the NeoPixel library with a scope. First verify my documented timings with a scope for WS2812B with Photon and Core, not logic analyzer. Then swap out the macros for these new inline functions. Appreciate the help Paul!

m-mcgowan:
Hi Everone, I'm preparing a 0.4.2 release, so going to merge this in. If any bugfixes are needed, we can discuss those as separate issues.

PR #450, opened 2015-06-10
Particle IO -> Particle

harrisonhjones:
Fixed manufacturer from Particle IO to Particle in the windows drivers.


ScruffR:
When trying to install this driver from this branch, I get an error message claiming that the hash value for the INF is not to be found in the connected CAT file.
Windows suspects unauthorized temering with the INF file and hence refueses to install.

(tested on Win 8.1 Pro 64bit)

harrisonhjones:
Well, we definitely did "tamper" with it. Interesting. Can I get a screenshot of the error?

harrisonhjones:
Looks like we need to generate new .cat files (see https://msdn.microsoft.com/en-us/library/windows/hardware/ff547089(v=vs.85).aspx). I'll work on it later today

ScruffR:
I'll have to switch my Win over to English, otherwise you'd only see my German version - may take a bit since I'm in the middle of downloading and installing some huge app ;-)

harrisonhjones:
Don't worry about a screenshot, just need new cat files.

PR #456, opened 2015-06-16
Fix return type for NO_CLOUD version of SyncTime()

Rukkh:
When the boolean return type was added to SyncTime, the SPARK_NO_CLOUD variant of the 
 CLOUD_FN macro in the syncTime function wasn't updated from void.

Compiling with SPARK_NO_CLOUD enabled throws an error.


m-mcgowan:
Thanks for this. I'll make the change by hand, rather than go through the formal process of signing CLAs etc..!

PR #461, opened 2015-06-22
Patch to enable half duplex mode in Serial1

prices:
This patch will allow user code to enable half-duplex mode in Serial1.  I have not tried it with Serial2.  The patch works great on my Core.  I have yet to try it on a Photon, because I have not received them yet.  I will be trying it when I receive one.

Setting the serial port to half duplex enables the one-wire on the hardware.  Both TX and RX use the TX pin.

Thanks!

### DIFF
--- a/hal/inc/usart_hal.h
+++ b/hal/inc/usart_hal.h
@@ -68,6 +68,7 @@ int32_t HAL_USART_Read_Data(HAL_USART_Serial serial);
 int32_t HAL_USART_Peek_Data(HAL_USART_Serial serial);
 void HAL_USART_Flush_Data(HAL_USART_Serial serial);
 bool HAL_USART_Is_Enabled(HAL_USART_Serial serial);
+void HAL_USART_Half_Duplex(HAL_USART_Serial serial, bool Enable);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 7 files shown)


m-mcgowan:
NIce work! If you haven't already done so, could you please sign our CLA.  https://docs.google.com/a/spark.io/forms/d/1_2P-vRKGUFg5bmpcKLHO_qNZWGi5HKYnfrrkd-sbZoA/viewform

And, if you'd like to, could you also add some API tests under `user/tests/wiring/api`. You build this app by adding TEST=wiring/api to the build command line (instead of any APP/APPDIR defines.) Thanks!

prices:
I have already signed the agreement with the last pull request I did to add EXTRA_CFLAGS into the user code compile.

Do you have more information on the testing?  I am looking at the test code an am not understanding it.  I got it to compile, but is it actually meant to get loaded into a Photon or Core?  It compiles with the arm compiler.

m-mcgowan:
It's compile only to assert that the API hasn't regressed and that it can be linked. (The Photon uses dynamic linking - and if a dynamic link function is missing, it shows up as a static link error.)

prices:
Ah.  So if it compiles, that is a success?  That should be easy enough.  Thanks.

prices:
I think that I added it properly.  It compiles properly.

m-mcgowan:
Thanks! :+1: I'll merge this.

m-mcgowan:
We forgot about CHANGELOG.md :-) - I added an entry under Features for the upcomming 0.4.3 release. If you have the time, could you have a go at submitting a PR to the spark/docs project for the new halfdupliex API?

The firmware docs are https://github.com/spark/docs/blob/master/src/content/en/shared/firmware.md

Thanks!

PR #465, opened 2015-06-25
User API for controlling an external RGB led

technobly:
This is currently working if anyone needs to control an external RGB LED as a system RGB LED, but be _warned_ it is likely to change.

Here's a sample user app:
```
#include "application.h"

SYSTEM_MODE(MANUAL);

void onChangeRGBLED(void* data, uint8_t r, uint8_t g, uint8_t b, void* reserved) {
   analogWrite(D0, r);
   analogWrite(D1, g);
   analogWrite(D2, b);
}

void setup() {
    pinMode(D0, OUTPUT);
    pinMode(D1, OUTPUT);
    pinMode(D2, OUTPUT);
    RGB.attachHandler(onChangeRGBLED, 0);
}

void loop() {
    RGB.control(true);
    RGB.color(255,0,0);
    delay(1000);
    RGB.color(0,255,0);
    delay(1000);
    RGB.color(0,0,255);
    delay(1000);
    RGB.control(false);
}
```

### DIFF
--- a/services/inc/services_dynalib.h
+++ b/services/inc/services_dynalib.h
@@ -39,6 +39,7 @@ DYNALIB_FN(services,LED_Toggle)
 DYNALIB_FN(services,LED_Fade)
 DYNALIB_FN(services,Get_LED_Brightness)
         
+DYNALIB_FN(services,set_rgb_led_change_handler)
 DYNALIB_FN(services,set_logger_output)
 DYNALIB_FN(services,panic_)
 
--- a/services/inc/services_dynalib.h
+++ b/services/inc/services_dynalib.h
@@ -39,12 +39,13 @@ DYNALIB_FN(services,LED_Toggle)
 DYNALIB_FN(services,LED_Fade)
 DYNALIB_FN(services,Get_LED_Brightness)
         
-DYNALIB_FN(services,set_rgb_led_change_handler)
 DYNALIB_FN(services,set_logger_output)
 DYNALIB_FN(services,panic_)
 
 DYNALIB_FN(services, jsmn_init)
 DYNALIB_FN(services, jsmn_parse)
+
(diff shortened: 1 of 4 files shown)


services/inc/services_dynalib.h
```diff
@@ -39,6 +39,7 @@ DYNALIB_FN(services,LED_Toggle)
 DYNALIB_FN(services,LED_Fade)
 DYNALIB_FN(services,Get_LED_Brightness)
         
+DYNALIB_FN(services,set_rgb_led_change_handler)
```
m-mcgowan: These must go at the end, or it will change the offsets of the other dynamically linked functions. This means existing firmware that tries to call panic_ would now call set_logger_output.

wiring/src/spark_wiring_rgb.cpp
```diff
@@ -61,4 +60,10 @@ void RGBClass::brightness(uint8_t brightness, bool update)
         LED_On(LED_RGB);
 }
 
+void RGBClass::attachHandler(led_update_handler_fn fn)
```
m-mcgowan: I suggest adding an optional void* data parameter so that users can have an optional data item passed back to their handler.

services/inc/services_dynalib.h
technobly: Good to know!  Is there a document that describes how the DYNALIB works?

technobly:
@m-mcgowan I incorporated your feedback, and edited the user app above.  Tested and working.  Feel free to edit my original post with an example of how you'd most likely see the `data` parameter being used.  I just stubbed it out for now.  I guess it can really be just about anything, but a pointer to the current value of something that would change the proxied system RGB LED output might be nice.

m-mcgowan:
We merged in #518 as agreed so closing this one.

PR #466, opened 2015-06-25
valid JSON output for pretty printing tools like JSON.stringify(DATA,null,4)

technobly:
When connecting to the device in Listening mode via serial, and receiving 's', this super small change makes the Describe Message output as valid JSON ready for copy/pasting into pretty print tools online or code like `JSON.stringify(DATA,null,4)`

@emilyrose just wanted to double check with you if this is needed for the CLI at all, and if so it would require some refactoring on your end most likely.

### DIFF
--- a/system/src/wifi_credentials_reader.cpp
+++ b/system/src/wifi_credentials_reader.cpp
@@ -209,8 +209,9 @@ void WiFiCredentialsReader::handle(char c)
     else if ('s' == c)
     {
         StreamAppender appender(serial);
+        print("{");
         system_module_info(append_instance, &appender);
-        print("\r\n");
+        print("}\r\n");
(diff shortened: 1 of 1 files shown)


technobly:
@emilyrose just pinging you on this again.  If you do not use this output for any reason in the CLI, this can be merged.  If you do, then we just need to figure out how to make the CLI work for both cases, implement that and then we can merge this.  Essentially, this just add `{ }` around the current output to make it valid JSON.

m-mcgowan:
I don't believe the cli presently uses the output, since I've not made this feature widely known - the primary intent was for debugging.  Having the output as valid JSON is of course a nice step in the right direction! So :+1: for merging this change.

PR #467, opened 2015-06-26
Fixed the halfduplex code so that it works on the Photon.

prices:
I got my Photons in the mail today.  ;)

I fixed the half duplex code so that it runs properly on the Photon.   I missed a couple of the hal entries that pertained to the photon.


technobly:
Thanks for the update @prices

PR #479, opened 2015-07-08
Add string_convert.h to includes in application.h

bkobkobko:
This will add itoa and friends so they can be used by users.


m-mcgowan:
thanks! :-1:

PR #493, opened 2015-07-18
micros() uses the full 32-bit value space

m-mcgowan:
https://community.particle.io/t/micros-rolling-over-after-59-652323-seconds-not-71-minutes-solved/3468/32

https://github.com/spark/firmware/issues/133


m-mcgowan:
The main code change is in platform/MCU/shared/STM32/hw_ticks.c

Rather than duplicate the same code between the STM32F2 and STM32F1 platforms, I factored it into a shared area for both STM32 platforms.  (Each time you copy and paste code, a unicorn dies... :-))

platform/MCU/shared/STM32/src/hw_ticks.c
```diff
@@ -0,0 +1,85 @@
...
+ */
+static volatile system_tick_t system_millis_clock = 0;
+
+/**
+ * Increment the millisecond tick counter.
+ */
+void System1MsTick(void)
+{
+    system_millis++;
+    system_millis_clock = DWT->CYCCNT;
```
rvnash: You may want to comment here that the order of these two statements are critical. If someone changes them in the future, the clever trick below that ensures that you weren't interrupted between these two statements would fail.

PR #495, opened 2015-07-19
fix for I2C timeout using micros(). Fixes #494

### LINKED ISSUE #494: [Photon] I2C doesn't allow 1.7Mhz
When i set the I2C speed to 1.7Mhz, checking on `endTransmission` returns **4** and boils down to      the code `if(EVENT_TIMEOUT < (isr_safe_micros() - _micros)) return 4;` 

https://github.com/spark/firmware/blob/latest/hal/src/stm32f2xx/i2c_hal.c#L257

m-mcgowan:
Changes the timeout to use the onboard cycle counter, since micros() doesn't wrap around at 32-bits, the subtraction trick to compute delays produces incorrect results.

### DIFF
--- a/hal/src/stm32f2xx/i2c_hal.c
+++ b/hal/src/stm32f2xx/i2c_hal.c
@@ -29,6 +29,7 @@
 #include "timer_hal.h"
 #include "pinmap_impl.h"
 #include <stddef.h>
+#include "hw_config.h"
 
 /* Private typedef -----------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


satishgn:
@mat, this fix fails the testcase : "TEST=wiring/i2c_shield" on my pressure sensor board.
However the feature/latest works fine on the test board. 

Not sure what exactly is the issue without a proper I2C shield where the feature/latest fails.

m-mcgowan:
Closing this since it doesn't fix the problem. We'll either stick with millis() or move to the 32-bit micros() implementations.

PR #497, opened 2015-07-19
Early implementation of timers in hal, only ms for now

maghis:
@m-mcgowan super simple implementation of how timers could work.

Let me know what you think.

### DIFF
--- a/hal/inc/hal_dynalib.h
+++ b/hal/inc/hal_dynalib.h
@@ -56,6 +56,8 @@ DYNALIB_FN(hal,HAL_Delay_Milliseconds)
 DYNALIB_FN(hal,HAL_Delay_Microseconds)
 DYNALIB_FN(hal,HAL_Timer_Get_Micro_Seconds)
 DYNALIB_FN(hal,HAL_Timer_Get_Milli_Seconds)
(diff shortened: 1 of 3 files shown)


pkourany:
@maghis, great start!  I am working in the background with @m-mcgowan to develop a more extensive timer library possibly based on my SparkIntervalTimer library.  The goal is to support several timer functions including input and output capture.  No decision has been made as to where the library will reside - system or user.  Would you like to collaborate on this effort?

maghis:
@pkourany of course man, super happy to help.

This was a very early effort to get the basics going because I need it for the product I'm developing.

Do you want to connect and have a chat via hangouts or skype about it?

pkourany:
@maghis we could do hangouts or we could also start a topic on the forum.  Have you tried the SparkIntervalTimer library?

maghis:
Yup, I was using that on the core. I believe now it makes sense to move the timer related functions to the hal and maybe have a higher level functionality library in the user part.

What are your thoughts?

pkourany:
@maghis, I've started a topic for this discussion here:

https://community.particle.io/t/hal-hardware-timer-library-development/13510

PR #500, opened 2015-07-19
Feature/wakeup wifi in all modes

m-mcgowan:
This wakes up wifi regardless of the mode, so that wifi is automatically enabled on wakeup from sleep in SEMI_AUTOMATIC/MANUAL modes.

One problem I see with this:
- if the wifi module was off beforehand, it will now be switched on. The previous wifi state (OFF,ON,CONNECTED) should be saved so that it can be restored to this state on wakeup.


m-mcgowan:
I've fixed this by storing the pre-sleep state and then restoring that state to wifi/cloud on wakeup. This works for all modes. So this PR is obsolete.

PR #518, opened 2015-08-04
Add a change handler to the RGB class

monkbroc:
A user can easily implement an external RGB LED by registering a change handler with `RGB.onChange` and doing `analogWrite` to another set of pins inside the change andler.

Multiple community topics show this is wanted.
http://community.particle.io/t/mirroring-on-board-rgb-led-to-external-led/13724/2
http://community.particle.io/t/best-way-to-implement-an-outboard-bright-rgb-led/12063/2

How to use:

```
// Needed to catch the initial cloud connection blinks
SYSTEM_MODE(SEMI_AUTOMATIC);

void externalLEDHandler(uint8_t r, uint8_t g, uint8_t b) {
  // reproduce green to an external LED to get a pretty good idea if the cloud is connected
  analogWrite(D0, g);
}

void setup()
{
  pinMode(D0, OUTPUT);
  RGB.onChange(&externalLEDHandler);
  Spark.connect();
}
```

Note: `Set_RGB_LED_Values(0,0,0);` was changed to `Set_RGB_LED_Color(0);` otherwise the handler is not called when the LED is toggled during the initial cloud connected.

### DIFF
--- a/services/inc/rgbled.h
+++ b/services/inc/rgbled.h
@@ -57,7 +57,7 @@ bool LED_RGB_IsOverRidden(void);
  */
 typedef void (*led_update_handler_fn)(void* data, uint8_t r, uint8_t g, uint8_t b, void* reserved);
 
-void set_rgb_led_change_handler(led_update_handler_fn fn, void* data);
+void LED_RGB_SetChangeHandler(led_update_handler_fn fn, void* data);
 
 
 #ifdef __cplusplus
--- a/services/inc/services_dynalib.h
+++ b/services/inc/services_dynalib.h
@@ -45,6 +45,7 @@ DYNALIB_FN(services,panic_)
 DYNALIB_FN(services, jsmn_init)
 DYNALIB_FN(services, jsmn_parse)
 DYNALIB_FN(services, log_print_)
+DYNALIB_FN(services,LED_RGB_SetChangeHandler)
 DYNALIB_END(services)
 
 
--- a/services/src/rgbled.c
+++ b/services/src/rgbled.c
@@ -61,6 +61,11 @@ uint8_t Get_LED_Brightness()
 static void* data;
 static led_update_handler_fn handler;
 
+void LED_RGB_SetChangeHandler(led_update_handler_fn fn, void* fn_data) {
+  handler = fn;
+  data = fn_data;
+}
+
 uint8_t asRGBComponent(uint16_t ccr) {
     return (uint8_t)((ccr<<8)/Get_RGB_LED_Max_Value());
 }
@@ -140,7 +145,7 @@ void LED_Off(Led_TypeDef Led)
         break;
 
     case LED_RGB:
-        Set_RGB_LED_Values(0,0,0);
+        Set_RGB_LED_Color(0);
         led_fade_step = 0;
         led_fade_direction = 1; /* next fade is rising. */
         break;
@@ -175,7 +180,7 @@ void LED_Toggle(Led_TypeDef Led)
         color = LED_RGB_OVERRIDE ? lastSignalColor : lastRGBColor;
         Get_RGB_LED_Values(rgb);
         if (rgb[0] | rgb[1] | rgb[2])
-            Set_RGB_LED_Values(0,0,0);
+            Set_RGB_LED_Color(0);
         else
             Set_RGB_LED_Color(color);        
         break;
--- a/user/tests/wiring/api/wiring.cpp
+++ b/user/tests/wiring/api/wiring.cpp
@@ -54,6 +54,15 @@ test(api_wiring_system_interrupt) {
 }
 #endif
 
+void externalLEDHandler(uint8_t r, uint8_t g, uint8_t b) {
+}
+
+class ExternalLed {
+  public:
+    void handler(uint8_t r, uint8_t g, uint8_t b) {
+    }
+} externalLed;
+
 test(api_rgb) {
     bool flag; uint8_t value;
     API_COMPILE(RGB.brightness(50));
@@ -63,5 +72,6 @@ test(api_rgb) {
     API_COMPILE(RGB.color(255,255,255));
     API_COMPILE(RGB.color(RGB_COLOR_WHITE));
     API_COMPILE(flag=RGB.brightness());
-
+    API_COMPILE(RGB.onChange(externalLEDHandler));
+    API_COMPILE(RGB.onChange(&ExternalLed::handler, &externalLed));
 }
--- a/user/tests/wiring/no_fixture/led.cpp
+++ b/user/tests/wiring/no_fixture/led.cpp
@@ -171,3 +171,29 @@ test(LED_BrightnessIsPersisted) {
     
     assertLEDColorIs(ledAdjust(255,128),ledAdjust(255,128),0);
 }
+
+uint8_t rgbUser[3];
+void userLEDChangeHandler(uint8_t r, uint8_t g, uint8_t b) {
+    rgbUser[0] = r;
+    rgbUser[1] = g;
+    rgbUser[2] = b;
+}
+
+void assertChangeHandlerCalledWith(uint8_t r, uint8_t g, uint8_t b) {
+    assertEqual(rgbUser[0], r);
+    assertEqual(rgbUser[1], g);
+    assertEqual(rgbUser[2], b);
+}
+
+test(LED_ChangeHandlerCalled) {
+    // given
+    RGB.onChange(userLEDChangeHandler);
+
+    // when
+    RGB.control(true);
+    RGB.brightness(255);
+    RGB.color(10, 20, 30);
+
+    // then
+    assertChangeHandlerCalledWith(ledAdjust(10),ledAdjust(20),ledAdjust(30));
+}
--- a/wiring/inc/spark_wiring_rgb.h
+++ b/wiring/inc/spark_wiring_rgb.h
@@ -21,7 +21,11 @@
  */
 
 #include <stdint.h>
+#include <functional>
 #include "rgbled.h"
+
+typedef void (raw_rgb_change_handler_t)(uint8_t, uint8_t, uint8_t);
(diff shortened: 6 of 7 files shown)


monkbroc:
C++ example using function wrapper:

```C++
class ExternalLed {
  public:
    ExternalLed(int pin);
    void begin();
    void handler(uint8_t r, uint8_t g, uint8_t b);

  private:
    int pin;
};

ExternalLed::ExternalLed(int pin) : pin(pin) {
}

void ExternalLed::begin() {
  pinMode(pin, OUTPUT);
  using namespace std::placeholders;
  RGB.onChange(std::bind(&ExternalLed::handler, this, _1, _2, _3));
}

void ExternalLed::handler(uint8_t r, uint8_t g, uint8_t b) {
  // Mirror green LED to external LED
  analogWrite(pin, g);
}
```

technobly:
Hey @monkbroc this is superb!  I actually was hacking on this recently and submitted a PR as well so that we could discuss internally: https://github.com/spark/firmware/pull/465

But your implementation works much better :grin:  The LED actually blinks (turns off when it should), whereas with mine it never called the handler when the LED color was 0 for some reason.  Love all of the extra C++ support as well.

I actually need this for a customer, and was trying to get the external RGB to be initialized and functioning as close to boot time as possible.  The following modification of your C++ example implements a full RGB early enough in runtime that you can even see the White breathing LED for a bit.  Notice because of the constructor setup, there is no need to use SEMI_AUTOMATIC mode.  Ultimately I think this will need to extend all of the way to the bootloader LED states, but this is a great start.

```c++
/*
 ******************************************************************************
 *  Copyright (c) 2015 Particle Industries, Inc.  All rights reserved.
 *
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation, either
 * version 3 of the License, or (at your option) any later version.
 *
 * This library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 * Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
 * License along with this library; if not, see <http://www.gnu.org/licenses/>.
 ******************************************************************************
 */

#include "application.h"

class ExternalRGB {
	public:
		ExternalRGB(pin_t r, pin_t g, pin_t b);
		void handler(uint8_t r, uint8_t g, uint8_t b);

	private:
		pin_t pin_r;
		pin_t pin_g;
		pin_t pin_b;
};

ExternalRGB::ExternalRGB(pin_t r, pin_t g, pin_t b) : pin_r(r), pin_g(g), pin_b(b) {
	pinMode(pin_r, OUTPUT);
	pinMode(pin_g, OUTPUT);
	pinMode(pin_b, OUTPUT);
	using namespace std::placeholders;
	RGB.onChange(std::bind(&ExternalRGB::handler, this, _1, _2, _3));
}

void ExternalRGB::handler(uint8_t r, uint8_t g, uint8_t b) {
	// Mirror system RGB LED to external common cathode RGB LED
	// Called every 1ms, keep this as short as possible!
	analogWrite(pin_r, r);
	analogWrite(pin_g, g);
	analogWrite(pin_b, b);
}

// Create a new ExternalRGB instance on D0, D1 and D2 (R, G, and B)
ExternalRGB myRGB(D0, D1, D2);

void setup() {
	// Nothing to setup for ExternalRGB, already done in constructor.
}

void loop() {
}
```

monkbroc:
Thanks @technology.

It does work better to set up the external led handler in a C++ constructor.

m-mcgowan:
These fixes look good, but one critical change is needed:

The dynalib order is critically important - the new dynalib function
 `DYNALIB_FN(services,LED_RGB_SetChangeHandler)` must go at the end of the list. 

It would also be much appreciated it you could add tests:
- `user/tests/wiring/api` - to add compilation tests for the new api (so you can test both regular and lambda functions compile). The code will not be executed, just compiled. 
- `user/tests/wiring/no_fixture/led.cpp` - as an actual test to validate that the RGB callback is invoked. (E.g. a test that sets the RGB led color, and verifies that the registered handler is called.)

monkbroc:
OK. I didn't even notice that this repository had tests because it wasn't mentioned in the main README.md and the tests are not in a top-level folder.

A link to the test readme from the main readme would help.

monkbroc:
Also, since the DYNALIB_FN order is important, could you add a comment in every dynalib header file like:

```C++
DYNALIB_BEGIN(services)
...
DYNALIB_FN(services, jsmn_parse)
DYNALIB_FN(services, log_print_)
// IMPORTANT: add new functions at the end (above this line) only!
DYNALIB_END(services)
```

Are there restrictions about removing dynalib functions (don't ever ever do it, or do it in a certain way)?

monkbroc:
@m-mcgowan I can't figure out how to actually run the tests. The test readme says to go to the unit folder and run `make` but that doesn't seem to pick up the tests in the `wiring` folder.

Do you have an updated document somewhere on running the tests?

monkbroc:
The CI tests pass, but I wasn't able to run them locally and see them fail, so I wouldn't trust that the right stuff is being tested.

m-mcgowan:
see https://github.com/spark/firmware/blob/develop/user/tests/readme.md#building-platform-tests for building platform tests.  so you'd add `TEST=wiring/api` to your build command to build the API tests. (These are only built and not executed.)

The actual integration tests can go in `tests/wiring/no_fixture/led.cpp`, which you build by adding `TEST=wiring/no_fixture` to the make command. I hope that helps.

monkbroc:
Thanks. That works.

I'll update the test readme file with instructions to run the wiring/api and wiring/no_fixture tests, ok?

One other question for you @m-mcgowan and @technobly. I was looking for an easy way to have interrupt/subscription/change handlers into C++ classes. While `std::function` makes this technically possible, it's still very cryptic how you write those. I added a template method on `RGB` so that you can simply pass the method and object without knowing the weird C++ `std::bind` syntax.

If you like that syntax, I can submit a PR with the same for attachInterrupt, Spark.subscribe, and a couple more APIs with callbacks.

How familiar are you with C++ templates, lambdas and functors? Do you think the implementation below is the simplest for users and most straightforward for maintainers?

```C++
class RGBClass {
    template <typename T>
    static void onChange(void (T::*handler)(uint8_t, uint8_t, uint8_t), T *instance) {
      using namespace std::placeholders;
      onChange(std::bind(handler, instance, _1, _2, _3));
    }
};

class ExternalLed {
  public:
    ExternalLed() {
      RGB.onChange(&ExternalLed::handler, this);
    }
    void handler(uint8_t r, uint8_t g, uint8_t b) {
      // ...
    }
} externalLed;
```

m-mcgowan:
Not sure I follow. We have std::function support for functions and interrupts and don't need to use std::bind. 

A large majority of our less advanced users will be using plain functions, or possibly lambdas, but using class methods is a stretch for some, so we can't make this the only option.

monkbroc:
I agree with straight functions pointers for simple cases. That should
definitely stay the main API.

What I was looking for was a simple way to have a non-static C++ method be
an interrupt handler. How would you do that?

m-mcgowan:
Just use a lambda?

```
class MyClass
{

   void isr_handler() { ... }

   void attach() {
      attachInterrupt(D5, RISING, []->{ this->isr_handler(); });
   }
}
```

monkbroc:
The syntax is quite verbose for other cases:

```
RGB.onChange([this](uint8_t r, uint8_t g, uint8_t b) -> void { this->handler(r, g, b); });
Spark.function("move", [this](String arg) -> int { return this->moveHandler(arg); });
```

Which is why I was proposing to put the complicated syntax inside the firmware, and keep the user-facing syntax simple.

```
RGB.onChange(&MyClass::handler, this);
Spark.function("move", &MyClass::moveHandler, this);
```

m-mcgowan:
Sure. I like with what you did in the RGB class - covers all the bases - class methods, lambdas and raw functions.

monkbroc:
Cool. This PR is ready then. Let me know if you have other comments.

m-mcgowan:
Almost done! The `wiring/api` tests are compile-only tests - to get integration tests run on the device, please also add the test code to `wiring/no_fixture/led.cpp`. When I checked yesterday there were some tests failing (!), but you can of course ignore these, so long as the new tests you add are ok. Thanks!

monkbroc:
I added both compile and integration tests yesterday. I ran the integration test on a photon by building with `modules$ make clean all program-dfu TEST=wiring/no_fixture PLATFORM=photon` and starting it with `cmd exclude=*`, `cmd include=LED_ChangeHandlerCalled`, `cmd start`. The test passed.

monkbroc:
BTW, thanks for the frequent updates on this PR. I understand it's not a priority with the upcoming 0.4.4 release, so I appreciate you taking some time to iron this out.

m-mcgowan:
no probs, glad to help.

hey @technobly  - you mentioned you needed the LED notifications for a customer - should I merge this PR now or wait until 0.4.5?

technobly:
@m-mcgowan I ended up merging this with the feature/ota-stall-fix branch to create feature/serial2-ext-rgb.  Looks like I should probably swap around the dynalib inclusion.  But if you are ready to merge this, and it's ready.. please do.  I can just delete the branch and point the customer to the next release.

monkbroc:
The commits in this PR have the new dynalib at the end by the way.

m-mcgowan:
I'm good with this. Gonna pull it in.

m-mcgowan:
I almost forgot - there's two other pieces to implementing a feature, adding an entry in the change log and the user documentation. I can take care of the change log. Would you take the user docs? You can directly edit https://github.com/spark/docs/blob/master/src/content/reference/firmware.md to submit a PR for the docs - please set the milestone to 0.4.4

monkbroc:
Sure, but I don't think I can set the milestone since I don't have commit rights.

m-mcgowan:
ok, no probs, I'll set that.

Issue #516, opened 2015-08-02
Photon: PANIC when TCPClient fails to connect 


maghis:
When trying to connect to a closed port or just failing to connect, the Photon restarts with an SOS instead of simply returning false.

Update:
- crashes and restarts w a PANIC if you try to connect right after the unit has connected to the internet (if you wait 2 seconds it works, less and it fails)
- crashes and restart w a PANIC if: address doesn't exist, nobody is listening on the port or if the tcp connection fails after a successful connect

maghis:
I often happens when it's connected too, when calling write().

monkbroc:
I have seen similar PANIC when the cloud connection fails. The `wiced_tcp_delete_socket` call in `Spark_Disconnect` causes the crash.

monkbroc:
This is probably the same underlying issue as #490

m-mcgowan:
Are you building locally? If so please check you have flashed your system firmware with this commit, which addresses a SOS issue with sockets - https://github.com/spark/firmware/commit/ae42b7114d2a0e7814fb22d7e5dd7bad6c450072

andyw-lala:
Is that already in develop ? I am still seeing problems with latest develop

maghis:
@m-mcgowan I imported that change in my local 0.4.3 (0.4.4 doesn't compile https://github.com/spark/firmware/issues/503).

I'll see if it works.

maghis:
@m-mcgowan the fix works but I believe the sockets get in a strange state after multiple connections, probably because I'm missing all the changes to the socket_hal you did before.

monkbroc:
Comment by @andyw-lala originally posted on #519:

> As of 4b09814, client.connect() causes SOS under 3 of the 3 common scenarios.
> This includes Mat's rework of your PR #519.
> 
> From the comment in the test case:
> 
> Modify the following to create the 3 common scenarios: *
> 1. Server and listener exist (e.g. run "netcat -l -p 8888" on the server)
> 2. Server exists, no listener on port (don't do the netcat)
> 3. Server doesn't exist (point the IPAddress somewhere in hyperspace *
> As of 4b09814 on develop branch:
> #1 works fine (duh)
> #2 fails instantly
> #3 fails after ~5 seconds (the timeout in socket_connect())
> Full test case can be found here:
> https://www.dropbox.com/s/ql7nq1a58xjydht/490-unit-test.ino?dl=0
> 
[...]

PR #519, opened 2015-08-04
Don't swallow timeouts in socket_connect

monkbroc:
There seems to be a mix-up between `sock_result_t` and `wiced_result_t`
types in `socket_connect`. The result of `wiced_tcp_connect` should be
assigned to a variable of type `wiced_result_t` not `sock_result_t`.

When `wiced_tcp_connect` times out and returns 2, the old version would
set `result = as_sock_result(wiced_result)` but since `wiced_result`
was only set by the call to `wiced_tcp_bind`, the timeout errror would
be swalled and `socket_connect` incorrectly return 0 (success) on
timeout.

### DIFF
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -622,12 +622,11 @@ sock_result_t socket_connect(sock_handle_t sd, const sockaddr_t *addr, long addr
             //wiced_tcp_register_callbacks(tcp(socket), socket_t::notify_connected, socket_t::notify_received, socket_t::notify_disconnected, (void*)socket);
             SOCKADDR_TO_PORT_AND_IPADDR(addr, addr_data, port, ip_addr);
             unsigned timeout = 5*1000;
-            result = wiced_tcp_connect(tcp_socket, &ip_addr, port, timeout);
-            if (result==WICED_SUCCESS)
+            wiced_result = wiced_tcp_connect(tcp_socket, &ip_addr, port, timeout);
+            if (wiced_result==WICED_SUCCESS)
                 tcp_socket->connected();
         }
-        if (result!=WICED_SUCCESS)
-            result = as_sock_result(wiced_result);
+        result = as_sock_result(wiced_result);
     }
     return result;
 }


towynlin:
Thanks @monkbroc — looks good. Next week some of the firmware team will be back from vacation, and we can look at merging this then.

monkbroc:
Sure @towynlin. I figured m-mcgowan was on vacation because he's super responsive usually.

andyw-lala:
I see the logic error in the code that you fixed, but applying your patch to my development branch makes my test cases that do not result in a connection (non existent host, good host but port without a listener) fail with SOS.

No idea why at this time.

monkbroc:
The call to `wiced_delete_tcp_socket` causes the SOS but without being able to trace into the WICED library I couldn't figure out why.

monkbroc:
Memory management issue or problem in the way the WICED library was compiled maybe?

m-mcgowan:
Thanks for raising this issue @monkbroc. This prompted me to rework the return value handling in that function to have just one result variable. (See the commit above.) I hope everyone agrees this is how it should have been to start with. :-)

andyw-lala:
As of 4b09814, client.connect() causes SOS under 3 of the 3 common scenarios.
This includes Mat's rework of your PR.

From the comment in the test case:
/*
 * Modify the following to create the 3 common scenarios:
 *
 * 1. Server and listener exist (e.g. run "netcat -l -p 8888" on the server)
 * 2. Server exists, no listener on port (don't do the netcat)
 * 3. Server doesn't exist (point the IPAddress somewhere in hyperspace
 *
 * As of 4b09814 on develop branch:
 * #1 works fine (duh)
 * #2 fails instantly
 * #3 fails after ~5 seconds (the timeout in socket_connect())
 */

Full test case can be found here:
https://www.dropbox.com/s/ql7nq1a58xjydht/490-unit-test.ino?dl=0

Going to spend some (but sadly, not a lot of) quality time digging into this right now.

monkbroc:
Thanks for the update @m-mcgowan. There's a lot of activity around SOS related to failed connects. I'll see over the next couple of days if I run into any further issues.

@andyw-lala I'll copy your comment to #516 since it talks specifically about panic for failed client.connect and close this PR.

andyw-lala:
Sure, I lose track of which of these 3 closely related issues is which. I
suspect the root cause will turn out to be similar.

PR #524, opened 2015-08-08
tidy up network_on/network_connect calls and remove duplication.

technobly:
...and removes seemingly unnecessary wiced_network_resume() call.  Tested with deep sleep, goes to sleep and resumes fine.  Presented as PR for easy testing and discussion before merge.

### DIFF
--- a/system/src/system_network.cpp
+++ b/system/src/system_network.cpp
@@ -236,7 +236,7 @@ void network_connect(network_handle_t network, uint32_t flags, uint32_t param, v
     {
         bool was_sleeping = SPARK_WLAN_SLEEP;
 
-        network_on(network, flags, param, NULL);
+        // network_on(network, flags, param, NULL); // This was causing OTA to stall.
 
         WLAN_DISCONNECT = 0;
         wlan_connect_init();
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Thanks so much for digging into this. I'm hoping the culprit is the call to `wiced_network_resume()`. I'll investigate just removing that so none of the platform neutral code needs to change.

technobly:
I tested just removing the wlan_activate() call and it does not stall.  That said, minus this one call to wlan_activate() which we are removing, the rest of network_on() is all pretty much setting the same flags that are in network_connect() plus changing the LED to blue.

We should think about what's happening in this block of code and if it's all in the correct order, and not redundant:
```c++
        network_on(network, flags, param, NULL);

        WLAN_DISCONNECT = 0;
        wlan_connect_init();
        SPARK_WLAN_STARTED = 1;
        SPARK_WLAN_SLEEP = 0;
```

m-mcgowan:
The OTA stalling was fixed based on the researched done here. ITO keep the platform-neutral parts unchanged, I only changed the photon HAL. https://github.com/spark/firmware/commit/b184ef5de33cf3e408122744436d9919fbd1ce19

The other changes stand as a reminder to refactor network_on / network_connect and wlan_connect_init / wlan_activate, since there is some overlap (resulting in wiced_wifi_connectivity_init being called multiple times.) This does no harm, but is redundant. 

Thanks again for the help @technobly.

PR #527, opened 2015-08-09
Fix inet_gethostbyname() behaviour and usage.

andyw-lala:
Use return value directly in system_cloud.cpp, instead of inferring error
from an IP address of 0.0.0.0.
Stop writing garbage into ip address in photon/inet_hal.cpp when the underlying
wiced call failed.
Use and pass on inet_gethostbyname() return value in spark_wiring_tcpclient.cpp


m-mcgowan:
Thanks very much for this, and for the in-depth analysis on the forums.

andyw-lala:
Were you able to validate on the core ? I did not build/test for the core
(I wondered if that was where the ipaddr == 0.0.0.0 test came from...

m-mcgowan:
I'll do a cursory check on the Core too to be sure, but the code looks good. The inet_gethostbyname should function in the same way on the Core.

PR #530, opened 2015-08-12
Add missing SysTick_Disable prototype

monkbroc:
Running `make all PARTICLE_DEVELOP=1 PLATFORM=photon` from the firmware root folder fails with `implicit declaration of function 'SysTick_Disable'` in `bootloader/src/main.c`

### DIFF
--- a/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/hw_config.h
+++ b/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/hw_config.h
@@ -64,6 +64,7 @@ typedef enum
 void Set_System(void);
 void NVIC_Configuration(void);
 void SysTick_Configuration(void);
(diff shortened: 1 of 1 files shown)


technobly:
I'm not sure if compiling from the firmware/ folder is or should be supported.  I only use main/, modules/ and bootloader/.  cc: @m-mcgowan

monkbroc:
A valid point, but the error still occurs when compiling from `bootloader/

technobly:
That's strange.  Thanks for raising this issue... will have to see if this is a recent regression or something with the makefile.

m-mcgowan:
This was a regression with changes I made when implementing `micros()` in #493 where I removed SysTick_Disable() from the header, getting ready to refactor. I think I may have been interrupted part way through this. The commit above addresses the issue. Thanks for reporting!

PR #534, opened 2015-08-13
Add C++ handlers and handler_data to subscribe

monkbroc:
Spark.function, Spark.subscribe and attachInterrupt can now take a pointer to a member and an instance to easily attach a callback to a C++ object as shown in the tests.

subscribe was missing lambda support and `spark_subscribe` did not save the data pointer to be passed to the `EventHandler` callback so I implemented the following solution to add lambda support:
* Don't change the signature of any dynalib method
* Save the handler data pointer in `FilteringEventHandler`
* If the data pointer is non-null, call the handler with the extra data pointer (3 arguments)
* If the data pointer is null, call the handler with 2 arguments

The idea is that this should be compatible with previous user code and HAL versions.

Compile and device tests pass (I didn't know how to add tests for attachInterrupt and Spark.function since those were not tested previously).

Note: This needs #533 for the tests to compile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
  - `Spark.connected()` et al. is now `Particle.connected()`. The former `Spark` library is still available but is deprecated.
  - `System.freeMemory()` API to determine the amount of available RAM.
  - `STARTUP()` macro to define blocks of code that execute at startup.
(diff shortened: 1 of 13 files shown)


m-mcgowan:
I've merged #533. That may require you to rebase this PR.

monkbroc:
Rebased

m-mcgowan:
This looks good to me - could you please rebase again and I'll merge. All the changes are additive so it's low risk. Also please add an issue to the spark/docs repo to fill out the corresponding documentation (or submit a PR to the docs) and link to here. Thanks!

monkbroc:
Done. The docs change is linked above.

PR #537, opened 2015-08-14
Link fix in gettingstarted.md

Darma:
"docs/" in the [Download and Install Dependencies] should be removed otherwise when it's clicked there will be additional "docs/" in the URL leads to 404.


m-mcgowan:
I'm having trouble with github...it just says "couldn't update branch. something went wrong". retrying doesn't help. I'll make the change separately. Thanks for the tip @Darma, much appreciated!

Issue #490, opened 2015-07-16
client.connect(hostname, port) causes SOS if DNS not resolved


pkourany:
On latest develop and 0.4.3 branches, running client.connect(hostname,port) where the DNS is not resolving (in my case perhaps because of a VPN), the function will cause an SOS and the Photon resets.

HardWater:
comments copied from the community site
First Posting 
-----------------------------------------------------------------------------------------------------------------
@peekay123 I am continuing to drill down to find out where this problem is. What I have found out so far is the problem occurs at this call in spark_wiring_tcpclient.cpp

       uint32_t ot = HAL_WLAN_SetNetWatchDog(S2M(MAX_SEC_WAIT_CONNECT));
       DEBUG("_sock %d connect",_sock);
       //gets here first and second request
       connected = (socket_connect(_sock, &tSocketAddr, sizeof(tSocketAddr)) == 0 ? 1 : 0);
       DEBUG("_sock %d connected=%d",_sock, connected);
       //gets here only on first request on the second request WDT seems to trip before here

@mdma is this something that you would be looking into or is there someone else that should be pinged?
-----------------------------------------------------------------------------------------------------------------
Second Posting

@mdma Making further progress on finding where the problem happens here is part of socket_hal.cpp

sock_result_t socket_connect(sock_handle_t sd, const sockaddr_t *addr, long addrlen)
{
    sock_result_t result = SOCKET_INVALID;
    socket_t* socket = from_handle(sd);
    if (is_tcp(socket)) {
        wiced_result_t wiced_result = wiced_tcp_bind(tcp(socket), WICED_ANY_PORT);
        if (wiced_result==WICED_SUCCESS) {
            wiced_tcp_register_callbacks(tcp(socket), socket_t::notify_connected, socket_t::notify_received, socket_t::notify_disconnected, (void*)socket);
            SOCKADDR_TO_PORT_AND_IPADDR(addr, addr_data, port, ip_addr);
            unsigned timeout = 5*1000;
           //-----------> gets here on first and second request
            result = wiced_tcp_connect(tcp(socket), &ip_addr, port, timeout);
           //-----------> gets here only on the first request WDT reset before here on the second
        }

        if (result!=WICED_SUCCESS)
            {
            result = as_sock_result(wiced_result);
            }
    }
    return result;
}

I do not seem to be able to get closer to the problem as I cannot find the source for wiced_tcp_connect() I am assuming that is part of the WICED binary image. Please Help?

Third posting
[...]

Issue #538, opened 2015-08-14
TCPClient::connect() does not close an existing socket.


andyw-lala:
TCPClient::connect() doesn't close an existing connection. It will orphan a socket if you call it without stopping the previous one. Sure, you can make the case of "Doctor, it hurts when I do this", but we should make the tcpclient (and server) classes as robust as possible, IMHO.

You will hit _this_ limit after something like 8 iterations with code similar to this:

```cpp
TCPClient client;

/*
 * Pick a host that exists, with an open http server.
 */
IPAddress server(192,168,121,51); 
int port = 80;

unsigned loop_count = 0;

void loop() {
  int x;
  
  Serial1.println("connecting to server (loop: " + String(loop_count) + ")");
  x = client.connect(server, port);
  Serial1.println("   connect returned: " + String(x));
  x = client.println("Hello World!");
  Serial1.println("   println returned: " + String(x));
  x = client.connected();
  Serial1.println("   connected returned: " + String(x));

/* With the following code, you can probably run for hundreds of connections, without it, maybe 8 or so. */
  if (x) {
    client.stop();
  }

  delay(2000);
  loop_count++;
}
```
However, it looks like there is some kind of async cleanup needs to happen in the wiced stack, because simply adding a call to TCPClient::stop() at the top of TCPClient::connect() doesn't work, whereas adding a call to client.stop(), followed by delay(2000) in application code, as shown above, does work.

If I was a betting man, I'd wager the same medicine will need to be applied to the TCPServer class as well.
~

m-mcgowan:
Totally agree, the API should make things as simple as possible and hard to shoot yourself in the foot. 

There have been many stability improvements to sockets lately (I've spent the day testing and refining), so I'm hoping simply calling stop() before connect in TCPClient/TCPServer is sufficient to avoid a resource leakage.

m-mcgowan:
For anyone watching, I believe this bug is around 2 years old from the original TCPClient implementation.

m-mcgowan:
Resolved in commit above.

Issue #536, opened 2015-08-14
TCPClient stop() does not work on first connection


nylen:
I wrote a simple HTTP server for the Photon.  When a client connects, I write a basic HTTP response in a few lines, then call `client.flush()` and `client.stop()`.

This works well, except for one thing:  my code does not disconnect the very first client to connect to the Photon after it boots up, so the connection hangs.  Every connection after the first one works fine.

Here's my full code:

```
int led1 = D6; // connected to solid-state relay
int led2 = D7; // the little blue LED built into the board
int buttonLow = D2;
int buttonInput = D1;
int buttonHigh = D0;

int ledState = HIGH;

TCPServer server = TCPServer(80);
TCPClient client;
char addr[16];

const int MAX_HTTP_LENGTH = 255;
char httpBuffer[MAX_HTTP_LENGTH + 1];
int httpLength;

void setup() {
    pinMode(buttonLow, OUTPUT);
    digitalWrite(buttonLow, LOW);
    
    pinMode(buttonInput, INPUT);
    
    // Writing an INPUT mode pin HIGH does not seem to work here,
    // like I hear it does on Arduino
    // so we're using an extra pin
    pinMode(buttonHigh, OUTPUT);
    digitalWrite(buttonHigh, HIGH);

    pinMode(led1, OUTPUT);
    pinMode(led2, OUTPUT);
    setState(HIGH);
    
    IPAddress localIP = WiFi.localIP();
    sprintf(addr, "%u.%u.%u.%u", localIP[0], localIP[1], localIP[2], localIP[3]);
    Spark.variable("Address", addr, STRING);
    Spark.variable("Request", httpBuffer, STRING);
    server.begin();
}

void setState(int state) {
    digitalWrite(led1, state);
    digitalWrite(led2, state);
    ledState = state;
}

void loop() {
    delay(100);
    
    if (digitalRead(buttonInput) == LOW) {
        setState(ledState == HIGH ? LOW : HIGH);
        while (digitalRead(buttonInput) == LOW) {
            delay(100);
        }
    }
    
    client = server.available();
    if (client.connected()) {
        httpLength = 0;
        bool foundNewLine = false;
        while (client.available()) {
            char c = client.read();
            if (c == '\r' || c == '\n') {
                foundNewLine = true;
            } else if (!foundNewLine && httpLength < MAX_HTTP_LENGTH) {
                httpBuffer[httpLength++] = c;
            }
        }
        httpBuffer[httpLength] = '\0';
[...]

PR #539, opened 2015-08-14
Work around a WICED bug crashing when socket connection fails

### LINKED ISSUE #516: Photon: PANIC when TCPClient fails to connect

monkbroc:
The reason for the bug is WICED deallocating the connection handler twice.

It occurs because in `WICED/network/LwIP/WICED/tcpip.c:920` the `netconn_delete( socket->conn_handler );` is not followed by `socket->conn_handler = NULL;` as it is in other places in the code for example in `WICED/network/LwIP/WICED/tcpip.c:931`. This leads `wiced_tcp_delete_socket` to delete the connection handler again, leading to a crash.

This should fix #516 and #490. @andyw-lala can you run your tests against this branch please?

This should also be reported to Broadcom so they can fix this upstream.

### DIFF
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -107,20 +107,14 @@ struct tcp_packet_t
  */
 struct tcp_socket_t : public wiced_tcp_socket_t {
     tcp_packet_t packet;
-    bool open; bool opened;
+    bool open;
     volatile bool closed_externally;
 
-    tcp_socket_t() : open(false), opened(false), closed_externally(false) {}
+    tcp_socket_t() : open(false), closed_externally(false) {}
 
-    ~tcp_socket_t()
-    {
-        // calling wiced_tcp_delete_socket when the socket didn't connect causes a hard fault.
-        // connected is only called after successful connection, setting opened to true.
-        if (opened)
-            wiced_tcp_delete_socket(this);
-    }
+    ~tcp_socket_t() { wiced_tcp_delete_socket(this); }
 
-    void connected() { open = true; opened = true; }
+    void connected() { open = true; }
 
     bool isClosed() {
         return !open || closed_externally;
@@ -630,8 +624,14 @@ sock_result_t socket_connect(sock_handle_t sd, const sockaddr_t *addr, long addr
             SOCKADDR_TO_PORT_AND_IPADDR(addr, addr_data, port, ip_addr);
             unsigned timeout = 5*1000;
             result = wiced_tcp_connect(tcp_socket, &ip_addr, port, timeout);
-            if (result==WICED_SUCCESS)
+            if (result==WICED_SUCCESS) {
                 tcp_socket->connected();
+            } else {
+              // Work around WICED bug that doesn't set connection handler to NULL after deleting
+              // it, leading to deleting the same memory twice and a crash
+              // WICED/network/LwIP/WICED/tcpip.c:920
+              tcp_socket->conn_handler = NULL;
+            }
         }
     }
     return as_sock_result(result);


andyw-lala:
So this would allow us to continue calling the upper level shutdown/disconnect functions, which should take care of housekeeping better ?

monkbroc:
> So this would allow us to continue calling the upper level shutdown/disconnect functions, which should take care of housekeeping better ?

Correct.

I reverted afd45d2e before implementing the fix.

andyw-lala:
I can apply and test tonight (US time).

andyw-lala:
Test started. Might not be able to check in on the results for a few hours (see also: life)

monkbroc:
Life? Don't talk to me about life.
-- Marvin

andyw-lala:
Still fails after the same number of passes (approx 350), with the same syndrome.

My guess is we're still running the wiced stack out of resources, or memory, or perhaps the whole photon out of memory - I lack the diagnostic to dump the current size of the memory pool, or we could see if there was a leak in effect.

monkbroc:
That's a really disappointing lack of quality in WICED :-(

stevie67:
I would like to debug this in a bit more depth. Which version of the WICED SDK is needed to debug this further and is there a guide how to install it?

monkbroc:
@stevie67 @m-mcgowan commented in another issue that Particle is in discussion with Broadcom for a permissive license to WICED. Right now they use WICE 3.3.0 with some modifications that they are not allowed to share with the community.

However, if you agree to Broadcom's terms, you can join the [WICED community](http://community.broadcom.com/community/wiced-wifi), [download the SDK](http://community.broadcom.com/community/wiced-wifi/wiced-wifi-documentation) and [download the Photon platform files](http://community.broadcom.com/community/wiced-wifi/wiced-wifi-forums/blog/2014/11/04/bcm9wcdusi09-platform-files-for-sdk-3x).

You can compile WICED with `./make snip.scan-BCM9WCDUSI09-FreeRTOS-LwIP-SDIO`. Althrough that library won't be compatible with the Photon build right now, it's still good enough to follow along what's going on and debug. (I adapted the WICED compile command from `hal/src/photon/wiced.sh`)

monkbroc:
What's your test case again @andyw-lala? I don't remember in which thread you posted it...

stevie67:
Thanks @monkbroc, I had joined the WICED community, just needed to know, which one to download. I will try that.

andyw-lala:
Here's the test I'm using:

```cpp
TCPClient client;

IPAddress server1(192,168,121,34); /* This server does not exist */
IPAddress server2(192,168,121,51); /* This server exists and has an open http port */
int port = 80;

unsigned loop_count = 0;

void handler(const char *topic, const char *data)
{
    Serial1.println("received " + String(topic) + ": " + String(data));
}

void setup() {
  Serial1.begin(115200);
  Serial1.println("********************** RESTART ***********************");
  
  Spark.subscribe("spark/", handler);
}

void loop() {
  int x;
  
  Serial1.println("connecting to server1 (loop: " + String(loop_count) + ")");
  x = client.connect(server1, port);
  Serial1.println("   connect returned: " + String(x));
  x = client.println("Hello World!");
  Serial1.println("   println returned: " + String(x));
  x = client.connected();
  Serial1.println("   connected returned: " + String(x));

  if (x) {
    client.stop();
  }

  delay(2000);
  
  Serial1.println("connecting to server2 (loop: " + String(loop_count) + ")");
  x = client.connect(server2, port);
  Serial1.println("   connect returned: " + String(x));
  x = client.println("Hello World!");
  Serial1.println("   println returned: " + String(x));
  x = client.connected();
  Serial1.println("   connected returned: " + String(x));

  if (x) {
    client.stop();
  }

  Spark.publish("spark/device/name");
  delay(2000);
  loop_count++;
}
```

See the commnents for server1 vs server2.

The output is as follows:

********************** RESTART ***********************
connecting to server1 (loop: 0)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 0)
   connect returned: 1
   println returned: 14
   connected returned: 1
connecting to server1 (loop: 1)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 1)
   connect returned: 1
   println returned: 14
   connected returned: 1
received spark/device/name: andyw-201
connecting to server1 (loop: 2)
   connect returned: 0
   println returned: -3
   connected returned: 0
received spark/device/name: andyw-201
connecting to server2 (loop: 2)
   connect returned: 1
   println returned: 14
   connected returned: 1
...
connecting to server1 (loop: 349)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 349)
   connect returned: 1
   println returned: 14
   connected returned: 1
received spark/device/name: andyw-201
connecting to server1 (loop: 350)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 350)
   connect returned: 1
   println returned: 14
   connected returned: 1
connecting to server1 (loop: 351)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 351)
   connect returned: 1
   println returned: 14
   connected returned: 1
connecting to server1 (loop: 352)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 352)
   connect returned: 1
   println returned: 14
   connected returned: 1
connecting to server1 (loop: 353)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 353)
   connect returned: 1
   println returned: -20
   connected returned: 1
connecting to server1 (loop: 354)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 354)
   connect returned: 1
   println returned: -48
   connected returned: 1
connecting to server1 (loop: 355)
...
connecting to server2 (loop: 360)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server1 (loop: 361)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 361)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server1 (loop: 362)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 362)
...

You can see how it starts off working well, and staying in contact with the cloud (proven by the pub/sub resulting in printing of the photon's device name. That seems to be the first casualty (around  pass 350), then the connection to server2 (the one that exists) starts acting up, with return values of -20 and -48 from client.println(), finally it settles down to a steady-state failure mode around loop 360.

andyw-lala:
Here's the pcap file that corresponds to the behaviour reported above. It might be missing some of the start of the test, but it certainly captures the end of the photon's connectivity.

https://www.dropbox.com/s/8y6kee36rrlkykv/516-soak-test-20150814-1.pcap?dl=0

m-mcgowan:
Andy, just to be sure I'm undrstanding correctly - are you saying that there's still a resource leak even with this fix?  I'll add a memory available function to the HAL.

andyw-lala:
Yes- with all the latest fixes, after 350 iterations of my test, the photon is dead to the network. At this point I do not know if it is a vanilla memory exhaustion problem, or if it is some other limitation buried in the wiced stack; but we're running out of *something*.

m-mcgowan:
Thanks. So this fix neither helps nor hinders the firmware? I'm surprised since this is what I would have expected to fix things and avoid the double delete. I think I'll merge this fix all the same and make a rc.3 release, since we are already many steps ahead of where we were with the 0.4.4-rc.2 release.

andyw-lala:
It's better, that's for sure.
Let me know when you've got a memfree call I can invoke every time around the loop of my testing.

andyw-lala:
Confirmed with 2fa9708 the long term failure remains.

monkbroc:
I think Particle needs to a serious memory management review of `hal/src/photon/socket_hal.cpp`.

There are several sketchy things going on in that file.

In `socket_create()`, memory is allocated for the Particle socket type `socket_t`: `socket_t* socket = new socket_t();`

`socket_create()` also contains `sock_handle_t result = SOCKET_INVALID;` (line 867) and the only other place it's used is in `socket_dispose(result);` (line 880). That's a relic from previous versions.

If the socket creation fails `wiced_result!=WICED_SUCCESS`, the `socket` memory allocated is not deallocated, or tracked anywhere.

If socket creation succeeds, the socket is added to a linked list (either `servers` or `clients`).

However, that linked list is never cleaned. The `clients` list had 775 Particle `socket_t` in it when @andyw-lala's test code stopped working. The only place where `socket_t` are deleted is in `remove_disposed_list` but that's never called in the current firmware.

So yeah, any code that uses sockets for a long time will eventually run out of RAM in the micro.

m-mcgowan:
> If the socket creation fails wiced_result!=WICED_SUCCESS, the socket memory allocated is not deallocated, or tracked anywhere.

The code path in `socket_create()` either adds it to the list or the socket memory is reclaimed by a call to  `socket_dispose()`.  I'd removed the `delete` call in socket_dispose while we were investigating SOS's to eliminate memory bugs in our code. Now we have this fix, I can re-instatate that an also verify that the lists are cleaned up appropriately.

m-mcgowan:
In the develop branch I've added resource cleanup, but there's definitely still a leak coming from WICED. 

Test case is to just connect to a non-existent port. With the normal code, I see a memory leak. If I comment out wiced_tcp_connect() in socket_hal and hard-wire to failure result, the lead disappears. 

So there's definitely a leak in WICED code. Once memory is exhausted, the cloud connection cannot be re-established. 

Unlike the Core, exhausted memory doesn't presently cause a PANIC, the intent being that we will later be adding better memory management with soft caches and the like. For now, I will introduce a panic on exhausted heap so we can see when the system has exhausted resources.

m-mcgowan:
After re-instating the delete and list management, my test case is showing no leaks at all after thousands of both successful and failed socket connection attempts.

andyw-lala:
Sorry, I forgot to post the commit I tested against, and now I'm out for
dinner. I'll post the commit as soon as we get home, but I did a git pull
about 20 mins before I sent my last update.

monkbroc:
That's good news Matt! No crashes and no leaks. That's a huge improvement from last week :rocket:

andyw-lala:
Using the new System.freeMemory() call, as of 3ab6ef7:

********************** RESTART ***********************
connecting to server1 (loop: 0)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 0)
   connect returned: 1
   println returned: 14
   connected returned: 1
Heap: 60984
...
received spark/device/name: andyw-201
connecting to server1 (loop: 25)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 25)
   connect returned: 1
   println returned: 14
   connected returned: 1
Heap: 60984


Nice work everyone. Early indications are that it's fixed - I'll leave the test running overnight & report back.

Now onto: #538, #536 & TCPClient::connected() [I don't think there's an issue opened for the connected(0 problem, or is it already fixed ?

m-mcgowan:
Great, thanks for confirming. I'm not sure exactly what the TCPClient::connected() issue is - if there isn't already an issue for that, I'd be grateful if someone could write it up.

monkbroc:
> TCPClient::connected() 

That's still an issue. Can `wiced_tcp_register_callbacks` come back to fix this?

stevie67:
Same for me. I am using the System.freeMemory() call (many thanks for that @m-mcgowan). With the newest version it does not leak memory anymore, did for the earlier version - which I did not realize because in my test case the memory leak was so small and it occurred so infrequently. And it still handles the failing connection gracefully. Thanks to @monkbroc and @m-mcgowan and all others involved! This makes the Photon usable in real projects...

andyw-lala:
So, 12 hours later - the free memory has changed, but there appears to be no persistent leak:

...
received spark/device/name: andyw-201
connecting to server1 (loop: 4142)
   connect returned: 0
   println returned: -3
   connected returned: 0
connecting to server2 (loop: 4142)
   connect returned: 1
   println returned: 14
   connected returned: 1
Heap: 57192
...

nlambuca:
I am facing the same here...
If you do 500mS loops gets even worst.
I am using a local webserver, but it is possible that a connection request not always returns OK.

The smaller the loop time, the easier to make it fail.

This is still open with 0.4.6 photon fw, right?

Issue #542, opened 2015-08-17
TCPClient::connected() doesn't detect when the socket is closed


monkbroc:
[Reported by @javier-rodas inside another issue](https://github.com/spark/firmware/issues/490#issuecomment-128799318)

When a server closes the socket, `connected()` keeps returning `true`.

Test case:

```c++
SYSTEM_MODE(MANUAL);

const enum {
  TEST_READ,
  TEST_PRINT,
  TEST_CONNECTED_ONLY
} test_mode = TEST_READ;

TCPClient client;

IPAddress server(192,168,0,4); // Start a server with nc -l -p 8080
int port = 8080;

void setup() {
  Serial.begin(9600);
  WiFi.connect();
}

void loop() {
  Serial.println("Loop started");

  client.connect(server, port);
  Serial.println("Connect finished");

  while(client.connected()) {
    switch(test_mode) {
      case TEST_READ:
        if(client.available()) {
          char c = client.read();
          Serial.print(c);
        } else {
          delay(100);
        }
        break;
      case TEST_PRINT:
        client.print(".");
        delay(100);
        break;
    }
    // Closing the server doesn't break out of this loop
    // Make sure it works with all values of test_mode
  }

  Serial.println("Not connected");
}
```

[edit] test case updated

monkbroc:
0ee169c didn't fix the test case above when only reading, or when neither reading or writing. The disconnect callback was not called.

That callback looks like it causes a lot of trouble because of the asynchronicity.

What about checking the LwIP error directly in some of the `socket_hal` functions? The key is `netconn_err`

I made a PR with this idea #543

JumpMaster:
I'm seeing the same issue with a simple web request using the develop branch.  Although the server disconnects the photon does not detect it.

    #include "application.h"
    
    TCPClient client;
    byte server[] = { 45, 56, 96, 24 }; // whatismyip.org
    unsigned long startTime;
    void setup()
    {
    	Serial.begin(9600);
        while (!Serial.available()) {
    		Serial.println("Press any key to start.");
    		delay (1000);
    	}
    
    	Serial.println("connecting...");
    	startTime = millis();
    	if (client.connect(server, 80))
    	{
    		Serial.println("connected");
    		client.println("GET / HTTP/1.1");
    		client.println("host: whatismyip.org");
    		client.println("connection: close");
    		client.println();
    	}
    	else
    	{
    		Serial.println("connection failed");
    	}
    }
    
    void loop()
    {
[...]

PR #543, opened 2015-08-17
Add a method to retrieve the last error on a socket

monkbroc:
This is an alternative approach to #542: check the LwIP connection error in the `isClosed`. Instead of reaching through to LwIP directly, use a WICED-style wrapper `wiced_last_error` that check LwIP `netconn_err`.

This avoid the asynchronicity of the callback and works in more cases. It makes the test program in #542 work for the TEST_READ and TEST_PRINT case. The TEST_CONNECTED_ONLY still fails (connected always returns true), but that test case fails with the callback too.

### DIFF
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -33,6 +33,8 @@
 #include "lwip/api.h"
 #include "network_interface.h"
 
+wiced_result_t wiced_last_error( wiced_tcp_socket_t* socket);
+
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Could you rebase please and I'll take it for a spin.

monkbroc:
Have at it Matt.

m-mcgowan:
Thanks muchly for this. Going to merge now and test in develop.

PR #548, opened 2015-08-19
Feature/otastress

m-mcgowan:
Simple firmware and script to repeatedly send OTA updates to a device and collect metrics.

Opens up system events for OTA progress.

### DIFF
--- a/communication/src/spark_protocol.cpp
+++ b/communication/src/spark_protocol.cpp
@@ -162,7 +162,6 @@ bool SparkProtocol::event_loop(void)
     if (updating)
     {
       system_tick_t millis_since_last_chunk = callbacks.millis() - last_chunk_millis;
-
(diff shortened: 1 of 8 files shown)


joegoggins:
@m-mcgowan , this is a great diagnostic! Two comments:
- [`System.on`](https://github.com/spark/firmware/blob/develop/user/applications/otastress/otastress.cpp#L102) looks incredibly powerful, would love to see that get [into the documentation](https://docs.particle.io/reference/firmware/photon/#system-calls).
- Let me know if you want to cook this into an end to end test (i.e. "fail if the OTA reliability threshold for 20 flashes is less than 90%") or something we should graph (i.e. OTA Reliability)

m-mcgowan:
Totally. Maybe I should officially release system events this sprint. They were just something added for a customer.  Yeah, having it in a end to end test would be great.

PR #550, opened 2015-08-19
Feature/antenna switch

m-mcgowan:
Don't use WICED AUTO mode and instead manually alternate between internal and external antennae)

### DIFF
--- a/hal/src/photon/wlan_hal.cpp
+++ b/hal/src/photon/wlan_hal.cpp
@@ -183,7 +183,8 @@ wlan_result_t wlan_connect_finalize()
         result = wiced_network_up(WICED_STA_INTERFACE, WICED_USE_EXTERNAL_DHCP_SERVER, NULL);
     }
     if (result) {
-        if (antennaSelection==ANT_AUTO && connection_failures & 2) {   // in auto mode, switch to auto after 2 connects
+        connection_failures++;
+        if (antennaSelection==ANT_AUTO && (connection_failures & 2)) {   // in auto mode, switch to auto after 2 connects
                     // time to flip antenna? (start flipping after 2 failures)) and on each 2nd failure
             wwd_selected_antenna = (wwd_selected_antenna==ANT_INTERNAL) ? ANT_EXTERNAL : ANT_INTERNAL;
             wlan_select_antenna_impl(wwd_selected_antenna);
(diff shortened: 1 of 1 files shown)


andyw-lala:
I want to go on record not liking this work around.

What's the path either fix this ourselves int he wiced code base, or get bcom to address it. ?

It beggars belief to think the photon is either the first or only device to hit this.

m-mcgowan:
The AUTO mode in wiced is handled in the wifi firmware, so we have no hope to fix it, and not the time to wait for BCM to fix it.

What are your concerns with this approach?

andyw-lala:
Because the choice of optimal antenna is not a persistent one in the general case.

AUTO needs to be "continually re-assess the best antenna", not "I picked the best one at startup".

As I said, I highly doubt that Particle is the first or only user of this silicon to find this - that just does not pass the sniff test to me. Particle may be the only people doing so under semi-public scrutiny.

I understand the pressure for a *workaround*; I want to make sure it is not seriously considered a *solution*.

technobly:
@andyw-lala I tested this for hours last night, and I agree... picking even the best one at startup is not "AUTO" in the sense that it keeps reassessing, but even picking the best at startup would avoid an issue where your external only antenna connected device can get a system update that results in internal antenna by default at boot and prevents it from getting further updates.  This branch currently doesn't even pick the best at startup.. it assumes that internal might be completely useless and will then switch until it finds one that connects.  But in practice I was getting it to connect to internal with -80dB when the external antenna was -42dB.  So for now we're going to re-work this feature and not include in the v0.4.4 release.

andyw-lala:
Thank you.

Has testing also included independently monitoring the behaviour of the antenna select control line ? I think that will be crucial to validation of a solution. I suspect finding the right person to talk to @ BCOM will be the hardest part of this, I'm pretty sure this is either already fixed, a regression, or a known problem with a known workaround.

technobly:
Testing did in fact include monitoring the behavior of the RF switch control lines, which contributed a lot to understanding why things didn't appear to work like we might have thought they would.  We have some really great contacts at Broadcom/USI and I'll be reaching out to them to better understand the things we are seeing with ANT_AUTO mode.

m-mcgowan:
@technobly  If you could reach out to contacts at Broadcom that would be great. For now, I'm closing this PR - for now we have the user specify internal or external antenna explicitly.

PR #554, opened 2015-08-20
Add support for joining / leaving multicast groups

stevie67:


m-mcgowan:
Thanks for this, nicely done! I'll pair review the code and comment inline.

hal/inc/socket_hal.h
```diff
@@ -92,7 +92,21 @@ sock_result_t socket_accept(sock_handle_t sd);
...
 sock_handle_t socket_handle_invalid();
 
+/*
+ * Join a multicast address for all UDP sockets. This will allow reception of multicast packets
+ * sent to the given address on all UDP sockets which have bound the port to which the multicast
+ * packet was sent.
+ * @param addr IP multicast address to join
+ * @return Return the result of the join operation
+ */
+sock_result_t socket_join_multicast(const sockaddr_t *addr);
```
m-mcgowan: Just wondering if this function should take a HAL_IPAddress? Does multicast make sense on IPv6 networks?

m-mcgowan: Please parameterize with a `network_interface_t` value since the Photon will soon support multiple networks.  I wonder if this is a good reason to make it an instance method on UDP rather than static, so we know the network interface to use?

Finally we need one or more void* reserved param at the end for future expansion. (Once set, the function signatures cannot change.)

hal/src/core/socket_hal.c
```diff
@@ -195,3 +195,15 @@ sock_handle_t socket_handle_invalid()
 {
     return SOCKET_INVALID;
 }
+
+sock_result_t socket_join_multicast(const sockaddr_t *addr)
+{
+    /* Not supported on Spark */
```
m-mcgowan: Core :-)

wiring/inc/spark_wiring_udp.h
```diff
@@ -73,6 +73,22 @@ class UDP : public Stream, public Printable {
...
         virtual size_t printTo(Print& p) const;
 
+	/*
+	 * Join a multicast address for all UDP sockets. This will allow reception of multicast packets
+	 * sent to the given address for on UDP sockets which have bound the port to which the multicast
+	 * packet was sent.
+	 * @param addr IP multicast address to join
+	 * @return Return the result of the join operation
+	 */
+	static int joinMulticast(const IPAddress& ip);
```
m-mcgowan: If begin() is called first before joining a multicast group, then we could make this an instance method so we have the associated network interface.

hal/inc/socket_hal.h
stevie67: Multicast makes sense on IPv6. To be honest, I could not find the red line telling me where to use the HAL_IPAddress and where to use sockaddr_t, but it should be easy to change it in that direction.

The network_interface_t makes sense, I will add it. Not connected for the moment, right? And nice to hear about the multiple interfaces :-).

Sorry, I don't really get the reason for the reserved parameter. Why would this be necessary? Is this done in other functions as well?

hal/src/core/socket_hal.c
stevie67: Aeeh, yes :-).

wiring/inc/spark_wiring_udp.h
stevie67: Sounds good!

hal/inc/socket_hal.h
m-mcgowan: Dynamic linking - function changes have to be backwardly compatible. Adding parameters would break that.  I plan to write up a doc explaining this, and steps to contribute in general.

stevie67: Okay, just saw that all the inet_hal.cpp have the reserved parameters and use the HAL_IPAddress. Maybe I should move the functions there, then it would fit better in style. On the other hand it would fit less with the rest of the UDP socket stuff.

stevie67:
Okay, the changes are implemented as discussed. Please have a look.

hal/src/photon/socket_hal.cpp
```diff
@@ -978,3 +978,17 @@ sock_handle_t socket_handle_invalid()
 {
     return SOCKET_INVALID;
 }
+
+sock_result_t socket_join_multicast(const HAL_IPAddress *address, network_interface_t /*nif*/, void * /*reserved*/)
+{
+    wiced_ip_address_t multicast_address;
+    SET_IPV4_ADDRESS(multicast_address, address->ipv4);
+    return as_sock_result(wiced_multicast_join(WICED_STA_INTERFACE, &multicast_address));
```
m-mcgowan: let's pass `nif` instead of `WICED_STA_INTERFACE`

hal/src/photon/socket_hal.cpp
```diff
@@ -978,3 +978,17 @@ sock_handle_t socket_handle_invalid()
...
+}
+
+sock_result_t socket_leave_multicast(const HAL_IPAddress *address, network_interface_t /*nif*/, void * /*reserved*/)
+{
+    wiced_ip_address_t multicast_address;
+    SET_IPV4_ADDRESS(multicast_address, address->ipv4);
+    return as_sock_result(wiced_multicast_leave(WICED_STA_INTERFACE, &multicast_address));
```
m-mcgowan: again, `nif` please!

wiring/inc/spark_wiring_udp.h
```diff
@@ -73,6 +73,23 @@ class UDP : public Stream, public Printable {
          */
         virtual size_t printTo(Print& p) const;
 
+	/*
+	 * Join a multicast address for all UDP sockets which are on the same interface as this one.
+	 * This will allow reception of multicast packets sent to the given address for UDP sockets
+	 * which have bound the port to which the multicast packet was sent.
+	 * NOTE: Can be called only after begin() was called.
+	 * @param addr IP multicast address to join
+	 * @return Return the result of the join operation
```
m-mcgowan: nice description.

m-mcgowan:
This is shaping up nicely! With the points above addressed, we're looking code complete. 

One other thing to add is an API test under `user/tests/wiring/api` so that we capture the essence of this API call. Let me know if you need any help with that. 

Finally, would you consider making a PR to the docs repo to add user documentation for these functions?  Do we need these for the Core also, or does that just work out the box?

hal/src/photon/socket_hal.cpp
stevie67: Okay, I just didn't know, if the nif converts 1-to-1 from hal to wiced. I did not find a place elsewhere where it is used. And it's slightly underspecified, I think...

wiring/inc/spark_wiring_udp.h
stevie67: Done.

hal/src/photon/socket_hal.cpp
m-mcgowan: ah..sorry my mistake. We need to convert nif to either WICED_STA_INTERFACE or WICED_AP_INTEFACE.

m-mcgowan: you can leave as is for now. I'll fix it when AP mode is supported.

stevie67:
I addressed the above points. 
I think I figured the unit test stuff out, but I was not able to compile it. So please have a look.
I will have to look into the documentation stuff later.

m-mcgowan:
Hi, I merged this then realized you hadn't signed our CLA, so I reset the develop branch. Could you send your email to me - mat at particle dot io and I'll forward you the CLA. Thanks!  You'll also need to resubmit the latest code as a PR. Sorry for the extra work this causes you.

PR #556, opened 2015-08-20
fast-pin: Cortex M3 Bit-Band access to perform fast GPIO atomic read-modify-write

satishgn:
The bit-banding code is common across all STM32Fx series

### DIFF
--- a/hal/src/core/pinmap_impl.h
+++ b/hal/src/core/pinmap_impl.h
@@ -36,6 +36,7 @@ extern "C" {
 typedef struct STM32_Pin_Info {
   GPIO_TypeDef* gpio_peripheral;
   pin_t gpio_pin;
+  uint8_t gpio_pin_source;
   uint8_t adc_channel;
   TIM_TypeDef* timer_peripheral;
   uint16_t timer_ch;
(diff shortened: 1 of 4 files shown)


hal/src/core/pinmap_impl.h
```diff
@@ -36,6 +36,7 @@ extern "C" {
 typedef struct STM32_Pin_Info {
   GPIO_TypeDef* gpio_peripheral;
   pin_t gpio_pin;
+  uint8_t gpio_pin_source;
```
m-mcgowan: This must come at the end of the pinmap or older applications using the old version of pinmap will crash if newer system firmware is updated.

pkourany:
Bad news - doing a straight sequence of pinSetFast() and pinResetFast() with bitbanding gives a pulse time of 126ns.  Without bitbanding (straight to register), the same sequences gives a pulse time of 42ns.  So bitbanding is much slower!

hal/src/core/pinmap_impl.h
satishgn: The above addition was for the Core...

m-mcgowan: Totally, I only noticed afterwards. I should have written a follow up note -  my appologies for the noise!

PR #571, opened 2015-08-22
TCPClient.remoteIP

m-mcgowan:
Fetches the remote IP for a TCPClient.


m-mcgowan:
```
TCPServer server = TCPServer();
server.begin(80);

TCPClient client = server.available();

Serial.println(client.remoteIP());
```

Issue #120, opened 2014-02-23
Servo doesn't adjust speeds correctly on continuous rotation servos


zsup:
At RobotsConf, many people were working with the Servo library using robots driven by continuous rotation servos. Theoretically the way they should work is that servo.write(0) should turn one direction, servo.write(90) should stop the servo, and servo.write(180) should turn the other direction. This worked correctly if the servo was calibrated appropriately.

However, in the case above, servo.write(91) should move very slowly in one direction, and servo.write(89) should move very slowly in the other direction. This behavior was not observed; instead, it seemed like servo.write functions always moved the continuous rotation servo at full speed, but just in different directions.

m-mcgowan:
Anyone know if this has been resolved? cc: @satishgn, @technobly

technobly:
@m-mcgowan I don't have any continuous rotation servos on hand to test this on, but I'm guessing this is most likely an issue with the servo if it doesn't move slowly at pulse widths near "center".  Also, for that to properly work, the servo would need to be precisely calibrated to the pulse width that servo.write(90) outputs as "center".  Based on the tolerance of the servo PWM output, this may not be completely in line with what the servo expects.

It would be best to analyze this problem by sending me a few of the servos that were used at RobotsConf that had issues... hopefully they are servos with a decent datasheet.  I'll use a scope and figure it out :)

technobly:
Ok so I nabbed up the two servos continuous rotation that Adafruit sells and tested them.  They work fine!  The issue is well stated as @zsup said where `servo.write(90);` does not make the servo stop.  However, I was able to make these servos stop, creep in either direction and go all of the way to full speed; but the center position value varied between them.  The large standard size servo needed to be set to exactly 95 to be stopped, where 94 and 96 would creep in either direction.  The micro servo was stopped between 90 and 98, and would creep in either direction at 89 and 99.

Each servo responds a little differently, and I think this could just be a tolerance issue.  Every RC Transmitter I've ever known has analog or digital trimmers on the different output channels, which essentially sets the neutral output pulse value.  Similar concepts could be implemented in the software to make something like `servo.write(90+trim);` actually stop servo.

What the user needs is a straight forward way to control a servo and determine what values stop the servo, and make it creep in either direction or move and full speed in either direction.  What to do with the range of values in between is really up to the user to implement, eg., proportional control or a low - med - high setting... etc..

What follows is a modified version of Tinker that allows you to control a servo using the Particle iOS or Android app.  

[...]

PR #575, opened 2015-08-22
servo trim

m-mcgowan:
fix for #120 - provide a setTrim() method that users can minutely adjust the servo timing. Untested.


m-mcgowan:
Our firmware API allows users to monitor button pushes, and be notified of changes to the LED. That's everything needed to resolve this issue and allow externalized buttons and LEDs.

PR #579, opened 2015-08-23
spark -> particle

m-mcgowan:
This PR tracks all the changes needed for the lower level code to migrate spark -> particle.

As changes are pushed to develop, they will be pulled also into this branch. 

Key changes:
- spark_wiring -> particle_wiring
- SPARK_ globals 
- spark_protocol -> particle_protocol in comms lib

Non changes:
- kept SPARK_WLAN_Loop
- did not rename `spark/` events
- did not rename makefile variables since these are part of the make interface.


m-mcgowan:
Note that this should not be pulled until after the 0.4.5 release.

technobly:
Because this was un-merged for so long, there are about 70 files with conflicts and it's quite a mess now.  It might be better to reapply the same changes to the latest develop.  Closing this for now, and adding a task to Clubhouse 1654.

PR #582, opened 2015-08-24
Initial softap-page commit

mebrunet:
# SoftAP connection page

## Description
These changes add a SoftAP setup page to the Photon's firmware (i.e. a self-contained, "app-less" way to connect the Photon to the WiFi).  When put into Listening Mode, the Photon serves a simple html/javascript page that walks the user through connecting the device to their WiFi. It assumes the device has already been claimed to a Particle account (as would often be the case if re-distributing the Photon as the heart of an IoT product).

## Screenshot of SoftAP Page:
image

## Approach
This was done by:

- Assembling a set of browser compatible RSA encryption tools,
- Writing a very simple html/javascript setup page,
- Compressing all assets into sufficiently short strings (const char arrays) declared in a C++ header file,
- Including that header file in `hal/source/photon/softap_ip.c' and serving the strings in the same way as the content at '/hello'.
- Tweaking 'hal/src/photon/softap.cpp` to handle the extra pages.

This was all accomplished by using the build tools posted at [this repo](https://github.com/mebrunet/softap-setup-page).


m-mcgowan:
Many thanks for this. We think it's a great contribution!

We've discussed this internally in the team and have made a plan on how to move forward with this.  What we'll do is create the necessary APIs in the system firmware to allow the softAP embedded webpages to be included in the application code. This will make it easy to update the pages when the application updates without needing to push a new system firmware release.

m-mcgowan:
Hi @mebrunet! It's been a while, but we've not forgotten about this.  The 0.5.0 release will include the necessary APIs to allow applications to provide their own SoftAP pages. I've linked to the PR where this is implemented. The test app was your softap-page app that you committed here. 

Would you be interested in making this available as a library so that app devs only need include your library in their project and have this available to them? 

I can provide assistance in creating the library. If you don't have time for this right now, I'm happy to create the library myself.

mebrunet:
Hi Matt -

Oh awesome! Sure, happy to make it into a library. Currently in China on a
tight schedule, but I'll have a look in when I get back in April and be in
touch.

Also, we've built a cool feature into our firmware / js that favors a local
network connection over the cloud for function calls whenever possible. Cut
our response time by an order of magnitude - lots of happy users! I know
this isn't the place to chat about it, but I think it would be a very cool
library and would be glad to discuss as well.

blackadmin:
Thank you very much for such a nice feature, this will be very helpful.

m-mcgowan:
You can compile the app yourself, @blackadmin - see PR #906

m-mcgowan:
Now that we have this available in application firmware, I'll close this PR. See #906 for the application code to add to your app to include this feature.

joliva:
@mebrunet, you mentioned having built the capability to favor local network to cloud function calls.  Do you have that code available somewhere where we can get access? Thanks.

mebrunet:
@joliva - yes, we wrote an ultra-light HTTP server that accepts function calls directly over the local network, then we append a hash to every command so the device only acts on the first command it receives (whether local or cloud). I've been meaning to tear it out into something open source, but haven't had the time yet. I'll see if I can get around to it next week.

joliva:
@mebrunet, that would be great.  I didn't quite follow how the hash is being used.

Djsharma07:
@mebrunet , Everything works perfectly for softAP setup page. I am having just a single question out of curiosity. I've seen your GitHub link here [softap-setup-page](https://github.com/mebrunet/softap-setup-page/blob/master/README.md)
And I've opened the softAP setup page by using "localhost:8000". However, when I've included the code given here [softAP HTTP Pages](https://docs.particle.io/reference/firmware/photon/#complete-example) in my `.ino` file, I could not do the same. I had to use IP Address rather(i.e. 192.168.0.1). Is it possible to open the page in my `.ino` same as I've opened by running `server.py`  command? 
Thank You.

PR #586, opened 2015-08-27
Hotfix for product macros

jme783:


dmiddlecamp:
just using branch 'photon_041_v2' directly for 041 targeting on the cloud instead of merging this fix in, lemme know if people would rather just merge it

m-mcgowan:
Hey guys, do we still need this?

m-mcgowan:
the compile server is using photon_041_v2 so this fix is already available.

PR #587, opened 2015-08-27
Feature/i2c multiple interface support

m-mcgowan:
adds an additional I2C selector to the HAL interface.


m-mcgowan:
Superceded by #588

PR #588, opened 2015-08-27
Develop feature/electron merge

m-mcgowan:
Merge electron back into develop branch. Tested I2C firmware upgrade specifically as this was an area that recently changed.


hal/inc/interrupts_irq.h
```diff
@@ -95,6 +95,7 @@ typedef enum hal_irq_t {
     SysInterrupt_TIM8_Compare1,
     SysInterrupt_TIM8_Compare2,
     SysInterrupt_TIM8_Compare3,
+    SysInterrupt_TIM8_Compare4,
```
m-mcgowan: This will break all clients that use interrupts further down in the list. Fortunately this API is very new an undocumented so that seems very unlikely in practice. (And the resolution is to recompile the application.) In future, add new values to the end of the list, or manage the values explicitly. (E.g. SysInterrupt_TIM8_Compare4=46)

PR #592, opened 2015-08-31
Fix RTC issue on Electron - override RTC alarm interrupt handler

satishgn:
@technobly @mohitbhoite @towynlin @brandoaire The RTC on electron works now after overriding the RTC alarm interrupt handler. So it's not an issue with the crystals or the loading caps as we concluded earlier


m-mcgowan:
Does this need merging? @technobly

m-mcgowan:
I've rebased so it's clear for merge.

technobly:
:+1:

PR #596, opened 2015-09-05
Updated readme.md

sdelavega:
Updated readme.md for better clarity and set example build instructions to default to photon (with optional dfu-flash procedures)


m-mcgowan:
I'm a bit puzzled by this PR. The bjam changes look good for the boost libraries, but the change of PLATFORM_ID to 6 doesn't make sense. This readme is specifically for our cross-platform compile, which is PLATFORM_ID 3. Mention of the photon in the PR makes me wonder if you're mistaking this for part of the embedded firmware build?

PR #605, opened 2015-09-11
fixed unsigned signed compare issue

kennethlimcp:
- system_tick_t is defined as `uint32_t`

Not sure if i am doing this correctly but here's the PR! :smile_cat:


m-mcgowan:
Thanks! I'd already fixed this as we talked before noticing the PR :-)

PR #606, opened 2015-09-13
Feature/backupram

m-mcgowan:
This PR implements retained variables in backup RAM.

## Syntax overview
To enable the backup SRAM (since it does require more power and there are apps that don't require this)
add

```
STARTUP(System.enableFeature(FEATURE_RETAINED_MEMORY));
```

to the stop of your sketch (alongside  `SYSTEM_MODE`)

Then you can declare retained variables. These are variables whose value is remembered between warm starts, such as resets while powered on, and wakeup from deep sleep.

```
retained int counter = 10;
```

This will create a variable whose initial value is 10 when powered up first time. 

Subsequent changes to that variable are stored in backup ram and retained during sleep so long as VBAT has power.

Here's an example app to test:

```
#include "application.h"

SYSTEM_MODE(MANUAL);    // not using the cloud here for the demo app
STARTUP(System.enableFeature(FEATURE_RETAINED_MEMORY));


// some retained variables most with initial values
retained int count = 5;
retained int uninitialized;        // uninitialized will be 0!
retained int i1 = 1;
retained int i2 = 2;

void setup()
{
    Serial.begin(9600);
    while (!Serial.available()) {}  // wait for keypress

    if (System.featureEnabled(FEATURE_WARM_START))
    {
        Serial.println("warm start");
    }
    else
    {
        Serial.println("cold start");
    }

    Serial.print("uninitialize value is ");
    Serial.println(uninitialized);

    count++;

    Serial.print("Count ");
    Serial.println(count);

    Serial.print("i1 ");
    Serial.println(i1);

    Serial.println("sleeping");

    delay(1000);

    System.sleep(SLEEP_MODE_DEEP, 1);
}
```

To test:

- flash system modules and this app to the device
- remove power
- restore power and connect to serial
- hit a key
- you'll see the message "cold start", and the current value of the variables. Counter starts a 5.
- the photon will sleep, then wake up, so reconnect serial and press a key
- you'll see the message "warm start" and the current value of the variables. Counter is now 6.
- similarly for powered resets - hitting the reset button increments the counter. 
- To reset the values, disconnect power.


m-mcgowan:
cc @jenesaisdiq new proposed API for persistent variables during deep sleep

m-mcgowan:
Note from richard: be sure to emphasise that that data is only retained so long as there is power. For permanently persistent storage that is retained even after power failure, use the EEPROM library.

PR #607, opened 2015-09-13
Print.printf/printfln

m-mcgowan:
This little gem allows you to write printf-like statements to any Print instance (Serial, TCPClient etc...)

```
Serial.printfln("There are %s types of people in the world, those that understand binary and those that don't", "10");

Serial.printfln("There are %d sweets in the jar", rand()%560);
```

Without the API we'd have to write:

```
Serial.print("There are");
Serial.print(10);
Serial.println("types of people in the world, those that understand binary and those that don't");

Serial.print("There are ");
Serial.print(rand()%560);
Serial.println("sweets in the jar");

```

I've wanted this for ages - formatting messages otherwise is quite tedious, requiring them to be split over many lines of code.

`printf` formats the message as given, `printfln` adds a newline at the end.

### DIFF
--- a/wiring/src/spark_wiring_print.cpp
+++ b/wiring/src/spark_wiring_print.cpp
@@ -25,6 +25,8 @@
  */
 
 #include <math.h>
+#include <stdarg.h>
+#include <stdio.h>
 #include "spark_wiring_print.h"
 #include "spark_wiring_string.h"
 #include "spark_wiring_stream.h"
@@ -240,3 +242,29 @@ size_t Print::printFloat(double number, uint8_t digits)
 
   return n;
 }
+
+size_t Print::printf_impl(bool newline, const char* format, ...)
+{
+    char test[5];
+    va_list marker;
+    va_start(marker, format);
+    size_t n = vsnprintf(test, 1, format, marker);
+    va_end(marker);
+
+    if (n<5)
+    {
+        n = print(test);
+    }
+    else
+    {
+        char bigger[n+1];
+        va_start(marker, format);
+        n = vsnprintf(bigger, n+1, format, marker);
+        va_end(marker);
+        n = print(bigger);
+    }
+    if (newline)
+        n += println();
+    return n;
+}
+
--- a/wiring/src/spark_wiring_print.cpp
+++ b/wiring/src/spark_wiring_print.cpp
@@ -245,13 +245,14 @@ size_t Print::printFloat(double number, uint8_t digits)
 
 size_t Print::printf_impl(bool newline, const char* format, ...)
 {
-    char test[5];
+    const int bufsize = 20;
+    char test[bufsize];
     va_list marker;
     va_start(marker, format);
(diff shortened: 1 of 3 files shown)


m-mcgowan:
cc @jenesaisdiq - new proposed API

wiring/src/spark_wiring_print.cpp
```diff
@@ -240,3 +242,29 @@ size_t Print::printFloat(double number, uint8_t digits)
...
+
+size_t Print::printf_impl(bool newline, const char* format, ...)
+{
+    char test[5];
+    va_list marker;
+    va_start(marker, format);
+    size_t n = vsnprintf(test, 1, format, marker);
+    va_end(marker);
+
+    if (n<5)
```
brycekahle: Is this an optimization?

m-mcgowan: Not sure I follow. It allows you to use printf-like syntax - placeholders like %s and %d in the string. This wasn't previously available on Serial, TCPClient etc.

wiring/src/spark_wiring_print.cpp
```diff
@@ -240,3 +242,29 @@ size_t Print::printFloat(double number, uint8_t digits)
 
   return n;
 }
+
+size_t Print::printf_impl(bool newline, const char* format, ...)
+{
+    char test[5];
+    va_list marker;
+    va_start(marker, format);
+    size_t n = vsnprintf(test, 1, format, marker);
```
brycekahle: Doesn't passing `1` here mean that test will never contain the correct string for `n<5`?

brycekahle:
:+1:

technobly:
Looks great to me, goodbye sprintf() and String hacks.  One question, what's driving the 20 character test, because it seems like it prints the same way <20 or >20 with vsnprintf() not having some 20 char limit.

m-mcgowan:
The buffer size is not known at compile time. We need a fixed buffer to be able to call vsnprintf() to find out the actual size needed.

technobly:
Ah, that makes sense... was thinking 20 was magical.  Add some multiple argument tests and %x %2.4f formatting?

```cpp
SCENARIO("Print.printfln() with a multiple args", "[print]")
{
    BufferPrint print;
    print.printfln("float hex %2.4f 0x%x", 12.345678, 3735928559);
    REQUIRE(String("float hex 12.3457 0xdeadbeef\r\n") == print.result());
}
```

m-mcgowan:
We have tests for sprintf("%f") elsewhere (since it needs to be explicitly included in nanolib). Here I don't think we need to unit test sprintf itself, only that we are calling it correctly.

technobly:
It's vsnprintf() that you are testing here (maybe that boils down to sprintf()), and multiple arguments will exercise it more.  I only stress it because I've seen weirdness with expansion of arguments with this type of code recently.  I would think more tests the merrier, but if you feel this is covered somewhere else no worries.  It's rare that we debate about adding too many tests! haha.

Issue #406, opened 2015-03-19
Provide a user API for the 32-bit DWT hardware counter on Photon and Core


technobly:
Please reference this community post: https://community.spark.io/t/micros-rolling-over-after-59-652323-seconds-not-71-minutes-solved/3468/18

Working with micros() values can be tricky, since they top out at 59652323 and not UINT_MAX(32bit unsigned long).  It puts more burden on the user to protect against rollover which happens every minute (i.e., very frequently)

This super lightweight function would give users access to a more precise counter with full 32-bit resolution, which will allow them to use unsigned integer subtraction to compute time deltas (which automatically protect against rollover).  The user must still be aware this is only valid for time deltas less than ~1 minute (59,652,323us).  For very short and precise time measurements, this works perfectly.  For anything larger than 1 minute, the millis() counter is a better choice.  Also a combination of the two could yield very precise measurements with intervals greater than 1 minute.

m-mcgowan:
Yes. I think some inline functions would be perfect here so that the value of the DWT counter can be inspected, and inline so we have no function call overhead.

I don''t think we specifically need to do anything other than provide guidance as to which stm32 platform headers to include and some examples.

technobly:
Great!  Is there a foreseeable way to inline it all of the way out to the Wiring library, so no headers are necessary?

m-mcgowan:
we could simply add the header to the default application.h includes. Do we need to make a special API? 

If the answer is yes, we could have an inline function like

```
uint32_t SystemClass::ticks() {
    return DWT->CYCCNT;
}

uint32_t SystemClass::ticksPerMicro() {
   return SYSTEM_US_TICKS;
}

```

technobly:
Yeah I like that a lot.  ticksPerMicros() would be a clearer name I think.  

Part of me thinks that `System.ticksPerMicros()` is making `SYSTEM_US_TICKS` more complicated... but maybe it's easier to remember this way?

m-mcgowan:
cc: @jenesaisdiq.  Richard, we're discussing a potential API to allow users to measure sub-microsecond delays using the system clock counter.

technobly:
Not sure if this needs to be a separate github feature request, or if we should combine it here... 

Many like to use the millis() and micros() counters to make timed measurements.  There are some details with using both of these that can be tricky.

A proposal for a simple method for capturing time deltas that would look something like this:

    millisDelta(RESET); // to reset the timed measurement
[...]

PR #608, opened 2015-09-14
Feature/system ticks

m-mcgowan:
#406 

Allows implementation of accurate microsecond delays.  (Although not that when we introduce the RTOS loops like this will need to be made atomic so there is no chance of a context switch.)

```
// get the current system ticks value
uint32_t now = System.ticks();

// for converting an the unknown system tick frequency into microseconds
uint32_t scale = System.ticksPerMicrosecond();

// delay a given number of ticks. 
System.ticksDelay(20*System.ticksPerMicrosecond());

```

The main motivation for this is that it allows very precise timing measurements and delays without function call overhead (all functions are ideally inlined by the compiler.)


m-mcgowan:
cc: @technobly @jenesaisdiq

PR #609, opened 2015-09-14
Feature/safemode api

m-mcgowan:
API and events to trigger safe mode.


m-mcgowan:
Device should also publish a system event when it is in safemode, the reason for being in safe mode. (user initiated, invalid app, missing dependencies etc..)

PR #610, opened 2015-09-14
Serial.flush()

m-mcgowan:
Attempting to fix this, but this is not a solution since there is no way to know if there is a USB device connected that will consume the data.  While there is no device connected, the application will wait indefinitely for the data to be sent, blocking the call to flush().

It's not clear to me if that should be the correct behaviour. I can see it breaking existing programs.

Here's a test app:

```
void setup()
{
    Serial.begin(9600);
    pinMode(D7, OUTPUT);
}

void loop()
{
    Serial.println("Hello world here is a very long string to full a buffer");
    Serial.flush();
    delay(1000);
    digitalWrite(D7, !digitalRead(D7));
}
```

When the serial monitor is open (e.g. screen on OSX) then the application prints data and the D7 LED flashes on and off. Once the serial monitor is closed, the LED stops blinking for 10s (the default timeout.)

### DIFF
--- a/hal/inc/hal_dynalib_usart.h
+++ b/hal/inc/hal_dynalib_usart.h
@@ -53,7 +53,9 @@ DYNALIB_FN(hal_usart,HAL_USART_Peek_Data)
 DYNALIB_FN(hal_usart,HAL_USART_Flush_Data)
 DYNALIB_FN(hal_usart,HAL_USART_Is_Enabled)
 DYNALIB_FN(hal_usart,HAL_USART_Half_Duplex)
-
+#ifdef USB_CDC_ENABLE
+DYNALIB_FN(hal_usart,USB_USART_Flush_Data)
+#endif
 DYNALIB_END(hal_usart)
 
 #endif	/* HAL_DYNALIB_USART_H */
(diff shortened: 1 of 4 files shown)


dmiddlecamp:
don't most two-wire serial connections just send in the blind without acknowledgement?

m-mcgowan:
Not over USB

dmiddlecamp:
Sure, but if we're using that as an example, it wouldn't be unreasonable for us to assume users would prefer their code not block when waiting on transferring usb serial data.

m-mcgowan:
When you're sending data you don't block but sometimes you need to know the data has been fully sent, hence Serial.flush(). The issue I have is that when there is no serial reader on the other end of USB then the data is never sent and the application blocks until the timeout. That seems like it follows from the definition of flush - wait until the data is received, but it still might be unexpected that the application hangs when no serial monitor is connected.

dmiddlecamp:
Hmm... Flush to me generally indicates that we're pushing anything on a pending buffer across the wire, it doesn't necessarily require guaranteed delivery.  Maybe there could be an extra param, or different function, like "blocking flush"?

Personally I might expect a tcp send flush to block, but I suspect I'd be surprised if serial.flush blocked?

m-mcgowan:
If it doesn't block how do you want for the data to be sent?

m-mcgowan:
This PR is here for testing. It doesn't work as desired, causing code to block indefinitely when the host isn't connected to serial.

PR #611, opened 2015-09-14
Feature/button events

m-mcgowan:
reworked button events to be more flexible, simpler to code and easier to hook by applications.

e.g. example app that turns the LED magenta while the button is held down

```
void button_handler(system_event_t event, uint32_t duration, void* )
{
    if (!duration) { // just pressed
        RGB.control(true);
        RGB.color(255,0,128);
    }
    else {    // just released
        RGB.control(false);
    }
}

void setup()
{
    System.on(button_status, button_handler);
}

void loop()
{
// it would be nice to fire routine events while the button is being pushed, rather than rely upon loop
    if (System.buttonPushed()>1000) {
        RGB.color(255,128,0);
    }
}
```

### DIFF
--- a/system/inc/system_dynalib.h
+++ b/system/inc/system_dynalib.h
@@ -47,6 +47,7 @@ DYNALIB_FN(system, system_sleep)
 DYNALIB_FN(system, system_sleep_pin)
 DYNALIB_FN(system, system_subscribe_event)
 DYNALIB_FN(system, system_unsubscribe_event)
+DYNALIB_FN(system, system_button_pushed_duration)
 DYNALIB_END(system)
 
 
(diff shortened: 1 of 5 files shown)


ScruffR:
This still wants documenting ;-)

m-mcgowan:
@jenesaisdiq  - any thoughts on who could document this? example usage is given above. my hands are full for the moment.

jenesaisdiq:
I'll find someone @m-mcgowan, thanks!

chobberoni:
i gotcha @m-mcgowan and thanks @jenesaisdiq for sending this to me

m-mcgowan:
Thanks to you both!

chobberoni:
@m-mcgowan if you had to sum up this new functionality in 1-3 sentences, what would you want it to say?

technobly:
System.buttonPushed() documented 3c7ed09

PR #614, opened 2015-09-14
Photon Multithreading 

m-mcgowan:
This is an alpha of the Photon multithreading code.

Known issues:
- sockets are not thread safe, so using a socket from the application loop may have unexpected behavior.
- not investigated for memory leaks


monkbroc:
Wow, this is some headache inducing code right there. I don't know how you managed to make all of this work!

What's the general operating principle? It seems like you have 2 threads (SystemThread and ApplicationThread), each with a queue where function calls (lambdas) get pushed and executed. When there's nothing to do the background process runs.

I realize this is a work in progress, but why do you mix gthreads, cpp::channels (Go-style channels in C++), std::futures/promises and FreeRTOS queues? Shouldn't using only FreeRTOS components be sufficient, and less confusing than mixing different libraries and styles? Plus there's a lot of unused concurrent library code in there right now.

A key thing for multithread to be effective will be to reduce blocking user APIs, or at least provide non-blocking equivalent. Is that the idea with _THREAD_CONTEXT_ASYNC_RESULT vs SYSTEM_THREAD_CONTEXT_SYNC?

Regarding comments like [unguarded updates](https://github.com/spark/firmware/pull/614/files#diff-facd3e0b5c78e9abc8e54b7188873187R520), my background is in automotive embedded controls and we got away with few synchronization primitives because memory writes 64 bit and smaller were guaranteed to be atomic (don't know if it's the case on the STM). So as long as the programmer understands that some of the variables used in a statement can change during that statement, you should be able to set up your integer reads and even writes so that state remains consistent.

m-mcgowan:
Thanks for taking time to give this feedback, much appreciated!

I will write up docs on the general operating principle - it's currently a work in progress here https://gist.github.com/m-mcgowan/8b11f012b694089b7820

freeFtos primitives are used as the backbone for implementation. gthreads build on that to provide the STL mutex, condition variable, std::thread etc. FreeRTOS is hidden behind the HAL - everything above that doesn't touch FreeRTOS directly and instead uses the standard C++ concurrency primitives. 

I started wanting to use channels for the Active Object queues but moved away from that since I needed more control (e.g. running the background task.) But I do want channels to be exposed to Applications as a convenient way for them to create CSP-style threaded apps.  The lower-level concurrency primitives are also there for folks that want to work directly with mutexes and the like.

Single accesses to one memory location are ok, but that expression involved several reads so there is potential for an inconsistent read.  I didn't look deeply into the order of accesses and writes to figure out if it would truly be a problem or not. 

Cheers,
mat.

monkbroc:
Sure thing Matt.

Reading your docs gist I expect that the biggest issue with existing user code will be in peripheral drivers that bang bits to pins because the pre-emptive round-robin scheduler will introduce 1ms delays once in a while. It will be necessary to have some very big bold warnings about adding critical sections to the docs, forums, examples.

Another note is that posting a message to the application queue in Particle.subscribe will delay running that user function until the current loop() completes instead of interrupting the user's code and running it now. It's a semantic change that could break code that expects user functions to interrupt loop().

On a slightly different topic, any time I see delay() in embedded code I cringe. It makes it so much harder to add functionality when most of the loop time is already spent busy-waiting. I think that introducing system threading would present a good opportunity to start showing better patterns as you mentioned. The challenge is coming up with those good patterns in the first place (I certainly wouldn't call what I did with automotive RTOS good patterns!)

m-mcgowan:
The scheduler is already there and has been since 0.4.0 release in May, so any code doing critical time tasks is already subject to being interrupted unless it disables irqs. This is also true on the core, with SysTick interrupt so I'm not seeing a big change there.

User functions did not interrupt loop() previously and would only run at the end of loop(), so the threaded behavior is in fact in line with existing behavior. This was a conscious decision - I could have made the user functions execute on yet another thread, but that would depart from the existing model and mean everyone has to become familiar with multithreaded code issues. 

I agree about delay, and in fact busy waiting in general. Over time we will try to provide best-practice patterns to help users get the most from their device. It has implications for power consumption so will be important for battery powered devices.

monkbroc:
OK. Good to know. I hadn't noticed this behavior.

monkbroc:
I had to comment this line in system/src/system_threading.cpp to compile.
```
    __future_base::_State_base::~_State_base() = default;
```

The error was:
```
src/system_threading.cpp:52:5: error: definition of explicitly-defaulted 'std::__future_base::_State_baseV2::~_State_baseV2()'
     __future_base::_State_base::~_State_base() = default;
     ^
In file included from ./inc/active_object.h:27:0,
                 from ./inc/system_threading.h:25,
                 from src/system_threading.cpp:3:
/home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/future:316:15: error: 'virtual std::__future_base::_State_baseV2::~_State_baseV2()' explicitly defaulted here
       virtual ~_State_baseV2() = default;
               ^
```

andyw-lala:
Julien,
         I agree that busy waits are bad, but you and I both come from the
world where every cycle you free up can be used by someone else; and that
is not the case for the arduino demographic. So, we can expect to see code
with some variation of "delay(magic_number)" all over the place.

This is one of the reasons that I'm not as gung-ho about multi-threading
being the answer to all our problems, I think it will fix a lot of things,
and also create some new issues - in addition to being just plain difficult
to implement & test. Don't get me wrong, we need to do this, and we'll all
be thankful for the work, eventually.

One thing we can do is help by making the standard delay() calls do the
right thing wrt yielding the CPU and using underlying OS facilities to
implement the delay. It might even help to make the user thread
non-preemptable - while this means that the user can still screw things up
if they spin inside loop() - that behaviour is already present today - just
making network-related calls non-blocking would be enough of a win in my
book.

I have a sense we will have to choose between doing the most compatible
thing, and a more pure thing - if we choose the pure route, we'll claim the
moral high ground, but it will be a massive education campaign - simply
because of the range of programming expertise/experience that the average
arduino user has. In part that's the absolute beauty of that platform - you
can be productive, even if the solution is <1% efficient and would get you
laughed out of CompSci 101. I'll take all the creativity and
cross-fertilization that this brings any day, despite all the abysmal code
that I see.

Extra credit internally for finding a clean way to have an expert mode for
people who do grok this stuff and can then instantiate their own
multi-threaded apps on the particle platform; while still allowing a
non-expert universe where a middle-schooler can write code with lots of
delays.

On Tue, Sep 15, 2015 at 1:22 PM, Julien Vanier <notifications@github.com>
wrote:

> Sure thing Matt.
>
> Reading your docs gist I expect that the biggest issue with existing user
> code will be in peripheral drivers that bang bits to pins because the
> pre-emptive round-robin scheduler will introduce 1ms delays once in a
> while. It will be necessary to have some very big bold warnings about
> adding critical sections to the docs, forums, examples.
>
> Another note is that posting a message to the application queue in
> Particle.subscribe will delay running that user function until the current
> loop() completes instead of interrupting the user's code and running it
> now. It's a semantic change that could break code that expects user
> functions to interrupt loop().
>
> On a slightly different topic, any time I see delay() in embedded code I
> cringe. It makes it so much harder to add functionality when most of the
> loop time is already spent busy-waiting. I think that introducing system
> threading would present a good opportunity to start showing better patterns
> as you mentioned. The challenge is coming up with those good patterns in
> the first place (I certainly wouldn't call what I did with automotive RTOS
> good patterns!)
>
> —
> <https://github.com/spark/firmware/pull/614#issuecomment-140489823>.
>



-- 
Andy

monkbroc:
I've been trying to make sense of the code and it's very hard to follow because many pieces are aliased or reimplemented in the namespace std.

I was trying to figure out how C++ STL threads, mutexes and queues related to FreeRTOS counterparts.

In gthr_default.h, gthread mutex methods delegate to HAL mutex methods (concurrent_hal.h) which delegate to FreeRTOS semaphore methods (concurrent_hal.cpp). Under the covers std::mutex is implemented using gthread (right?)

In system_threading.cpp, std::thread start is implemented using HAL methods, which delegate to FreeRTOS task methods. So std::thread is essentially a FreeRTOS task.

On the other hand system_threading.cpp uses HAL queues directly, not an STL implementation, so it was confusing while reading this file to figure out why queues were handled differently than mutexes and threads.

This is hard stuff...

Do you have a reference for this active object pattern you are using? I'm not familiar with it.

m-mcgowan:
Andy,
Delays are implemented using task queuing in freeRTOS so they are cycle-efficient. Disabling interrupts aside, it's not possible for the user thread to permanently block the application thread since they are both at the same priority, so in the worst case you'll have each executing for 1ms before switching out to the other. The system thread aims to be well behaved and is I/O bound rather than CPU bound so should not steal much time from the application thread.

Making some calls non-blocking might provide a solution to some degree, but callback-based programming is a significant departure from the synchronous programming model that we and the Arduino community have been pushing for years. By using multhreading we allow users to continue writing code they are familiar with while also ensuring individual tasks run freely and allow the RTOS to manage scheduling, especially when tasks are I/O bound. (Incidentally, the WICED library API is synchronous so implementing asynchronous I/O would be a significant undertaking both for us to build and for our users having to learn a new API.)

m-mcgowan:
@monkbroc  - there is no gthread implementation on ARM, that's why we implement the necessary gthread platform functions so that we provide the missing platform stubs as required.

I try to use the STL concurrency primitives where possible, but didn't feel comfortable with the queue implementation (I don't recall why, but it wasn't an exact fit for what we needed, but the HAL queue was, so I used that being an ideal fit.)

As you noted the implementation is a series of layers -  from stl threads to gthreads to HAL to FreeRTOS. Given the system we are working with, the separation of concerns that we require, and use of STL on gcc (which implies gthreads), is there an alternative? Not that I really see a need for an alternative - I'm sure you're also aware that layered systems are commonplace and a way to augment existing functionality or adhere to specific interfaces.

I can appreciate it's a lot of code to take in on first read, so thanks for bearing with it and that headache! But I hope you also see that we have impacted relatively little code looking at the codebase as a whole, and now have both compile time and runtime enable/disable of threading behavior, which I'm personally very happy with!

monkbroc:
It was mostly the mix of layers that got me.

active_object.h uses HAL methods directly, it uses reimplemented STL objects from system_threading.cpp and it uses standard STL objects from <mutex>, <future>

You definitely deserve congratulations for achieving all this behavior with very little effect on the firmware and user codebase! I'm commenting with the hope that early efforts to better organize the threading code will prevent errors from being introduced when this code is modified in the future.

system/src/system_threading.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <string.h>
+
+
+#if PLATFORM_THREADING
+void system_thread_idle()
+{
+    Spark_Idle_Events(true);
+}
+
+ActiveObjectThreadQueue SystemThread(ActiveObjectConfiguration(system_thread_idle, 1024*3, 100));
```
monkbroc: Did you swap the stack size and take_wait timeout parameters?

system/src/system_threading.cpp
```diff
@@ -0,0 +1,84 @@
...
+    void invoke_thread(void* ptr)
+    {
+        thread::_Impl_base* call = (thread::_Impl_base*)ptr;
+        call->_M_run();
+    }
+
+    void thread::_M_start_thread(thread::__shared_base_type base)
+    {
+        if (os_thread_create(&_M_id._M_thread, "", 0, invoke_thread, base.get(), 1024*20)) {
+            PANIC(AssertionFailure, "%s %", __FILE__, __LINE__);
```
monkbroc: The second %s is missing in the format string

hal/src/photon/concurrent_hal.cpp
```diff
@@ -38,52 +52,23 @@ static_assert(sizeof(uint32_t)==sizeof(void*), "Requires uint32_t to be same siz
...
-}
+
 
 /**
  * Determines if the given thread is the one executing.
  * @param   The thread to test.
  * @return {@code true} if the thread given is the one currently executing. {@code false} otherwise.
  */
-bool os_thread_current(os_thread_t* thread)
+bool os_thread_current(os_thread_t thread)
```
monkbroc: This function is called os_thread_is_current in concurrent_hal.h
It should be renamed in this file.

Adding os_thread_t os_thread_current() to return the current thread would be useful.

system/inc/active_object.h
```diff
@@ -0,0 +1,335 @@
...
+    ActiveObjectQueue(const ActiveObjectConfiguration& config) : ActiveObjectBase(config), queue(NULL) {}
+
+    void start()
+    {
+        createQueue();
+    }
+};
+
+
+class ActiveObjectCurrentThreadQueue : public ActiveObjectQueue
```
monkbroc: How about ActiveObjectQueueInCurrentThread and ActiveObjectQueueInSeparateThread?

system/inc/active_object.h
```diff
@@ -0,0 +1,335 @@
...
+{
+public:
+    ActiveObjectCurrentThreadQueue(const ActiveObjectConfiguration& config) : ActiveObjectQueue(config) {}
+
+    void start()
```
monkbroc: I'd add the warning that this method does not return.

hal/src/photon/concurrent_hal.cpp
```diff
@@ -114,11 +103,239 @@ os_result_t os_thread_join(os_thread_t* thread)
...
+
+    void wait(lock_t* lock)
+    {
+        taskENTER_CRITICAL();
+        lock->unlock();
+        queue.enqueue();
+        taskEXIT_CRITICAL();
+        vTaskSuspend(NULL);
+        lock->lock();
+    }
```
monkbroc: I think a deadlock can happen if the task is pre-empted between the call to taskEXIT_CRITICAL and vTaskSuspend, and then the other task signals the CV. vTaskResume will be called on the original task before vTaskSuspend.

Instead of using vTaskResume and vTaskSuspend, I think you'll need one [binary semaphore](http://www.freertos.org/xSemaphoreCreateBinary.html) per thread to handle the suspend/resume logic.

m-mcgowan: Thanks but I'm not seeing the deadlock. The critical section is over so the scheduler is free to preempt the task - all that will happen is that the task ends up being suspended earlier (when it's preempted it's not running.) And when it's not preempted it hit hits vTaskSuspend() so it also ends up not running.  Which resources do you think are in contention to cause a deadlock?

monkbroc: The scheduler preempting a task is not the same at suspending it.

The vTaskSuspend docs say "When suspended a task will never get any microcontroller processing time, no matter what its priority"

The vTaskResume docs say "A task that has been suspended by one or more calls to vTaskSuspend () will be made available for running"

So in case vTaskResume is called first on the task trying to wait it will be a no-op since it has not been suspended yet (not currently running, but allowed to be run by the scheduler). Then when the task trying to wait is scheduled again it will call vTaskSuspend and be blocked from executing forever.

system/src/system_threading.cpp
m-mcgowan: Yep! Well caught! Fortunately the stack_size isn't used, but it did mean the system thread was waiting longer than I'd like to run the background loop.

system/src/system_threading.cpp
m-mcgowan: :+1:

hal/src/photon/concurrent_hal.cpp
m-mcgowan: The mutex is provided by the lock used with std::condition_variable. Even so, without the lock it's a typical race condition between two threads - if vTaskResume is called before vTaskSuspend then that task simply misses the notification - just as if it called wait a few microseconds later. I'm not seeing a deadlock here, just a missed notification, and the guard against that is the lock owned by threads using wait and notify on the condition variable.

andyw-lala: Is it possible/realistic/straightforward to restructure the code so there
is not even the whiff of a deadlock, rather than debating the
presence/absence of one ?

On Fri, Sep 18, 2015 at 1:53 PM, Matthew McGowan <notifications@github.com>
wrote:

> In hal/src/photon/concurrent_hal.cpp
> <https://github.com/spark/firmware/pull/614#discussion_r39888667>:
>
> > +{
> > +    ThreadQueue queue;
> > +
> > +    typedef std::unique_lock<std::mutex> lock_t;
> > +public:
> > +    ConditionVariable(UBaseType_t max_size) : queue(max_size) {}
> > +
> > +    void wait(lock_t* lock)
> > +    {
> > +        taskENTER_CRITICAL();
> > +        lock->unlock();
> > +        queue.enqueue();
> > +        taskEXIT_CRITICAL();
> > +        vTaskSuspend(NULL);
> > +        lock->lock();
> > +    }
>
> The mutex is provided by the lock used with std::condition_variable. Even
> so, without the lock it's a typical race condition between two threads - if
> vTaskResume is called before vTaskSuspend then that task simply misses the
> notification - just as if it called wait a few microseconds later. I'm not
> seeing a deadlock here, just a missed notification, and the guard against
> that is the lock owned by threads using wait and notify on the condition
> variable.
>
> —
> <https://github.com/spark/firmware/pull/614/files#r39888667>.
>



-- 
Andy

m-mcgowan: To my min the code is already simple and deadlock free - std::condition_variable semantics mean that the caller must own the lock on calling wait() or notify() so mutual exclusion is already guaranteed.  So there's no whiff of a deadlock by design. The only deadlock cause would be if either wait or notify would hang indefinitely while holding the lock. You can see from the code that isn't the case.  There is no blocking code while owning the lock.

monkbroc: Deadlock may be the wrong term. Task death is more what I meant: the waiting task is in danger of never being scheduled again if it suspends itself and no other task comes later to resume it.

I realize you disagree with my reasoning and you've got other things going on so I'll table this for now.

m-mcgowan: Thanks for clarifying. I'm happy to continue discussing - it's important to be sure it's right! The guard against that is simply that only one process can execute wait or notify at a time since these functions are guarded by a mutex, just as you mentioned in your proposed solution.  See the PromiseBase class which uses the condition variable - all uses are guarded by acquiring the same shared lock for that variable.

monkbroc: I agree that only only one task may enter wait or notify because of the mutex / shared lock. My point was that since the lock is unlocked in wait there is a possibility (however unlikely) that the other task will enter notify before the first task suspends itself.

Doing something like vTaskSuspend outside a critical section and without a lock while another thread is trying do vTaskResume around the same time has to have a failure mode.

I wrote that scenario in code. I used os_thread_scheduling to force the scheduler not to preempt the application task until right after the end of the critical section in wait, then forced the scheduler to preempt the application task (this could happen "in the wild" without the extra code). In this case vTaskResume is called before vTaskSuspend and the application task freezes forever (loop is never called in application.cpp).

https://github.com/munirent/particle-firmware/pull/1/files

m-mcgowan: Thanks. I fully understand your point now.  We can either implement using a semaphore to provide the task sleep/resume functionality, or we can have "spurious wakeups" as allowed by the spec. I'll take a look implementing sleep/wakeup with semaphores first.  I've seen that linked paper before, I don't recall why I didn't implement it like that first time. I don't really like the code as it is - those enforced critical sections are unnecessary once the correct locking primitives are used. I'll rework this. Thanks for prompting!

monkbroc: Glad to help. It take a lot of eyeballs to spot issues with concurrent programming.

One other thought I had: would it be possible to replace the promise in invoke_work by a queue with one item? Create a queue in the caller thread, pass it to the worker thread, when the work completes, the worker puts the result onto the queue. The caller takes from the queue and (potentially) blocks until the result is ready, then deletes the queue (could be handled by shared_ptr). It's less general, but it would allow you to remove promises and condition variables completely from the code (that's a pretty large bit of code).

Right now the only case where you need that promise is when the app blocks on a system call. A promise isn't returned from SYSTEM_THREAD_CONTEXT_ASYNC so you really don't have to worry about the case where the app does a system call and never gets the result of the promise (although it looks like it would work fine because you're using shared_ptr for that reason).

Just a suggestion to try to simplify the task completion synchronization code.

hal/src/photon/concurrent_hal.cpp
```diff
@@ -127,3 +344,5 @@ void os_thread_scheduling(bool enabled, void* reserved)
     else
         taskEXIT_CRITICAL();
 }
+
```
monkbroc: The logic is inverted in os_thread_scheduling if the meaning of os_thread_scheduling(false) is to disable the scheduler.

hal/inc/concurrent_hal.h
```diff
@@ -132,4 +181,3 @@ void os_thread_scheduling(bool enabled, void* reserved);
 #endif
```
monkbroc: The reserved argument of os_thread_scheduling should have a default value of NULL

m-mcgowan: It's a C API - there are no default values in C. As soon as we raise the bar to having everything C++ then we can do this (and fix some other long standing issues, like having to use #defines for pin constants.)

monkbroc: You learn something new everyday! I never realized that was the case. I
must never have tried to compile a function with default arguments in ANSI
C 😁

monkbroc:
At this point I haven't actually been able to successfully use this on any of my projects. I get hard faults that I can't unravel deep in the lambda to functor conversion code.

This is a simple program that crashes for me.

```c
#include "application.h"

SYSTEM_THREAD(ENABLED);

double a = 1.0;
double b = 2.0;

void setup()
{
    Particle.variable("a", &a, DOUBLE);
    Particle.variable("b", &b, DOUBLE);
}

/* Crashes at invocation of spark_variable("b", ...) in the application thread

Stack before the crash

#0  0x080808e4 in _malloc_r ()
#1  0x08062a96 in _M_init_functor (__f=..., __functor=...) at /home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/functional:1955
#2  _M_init_functor (__f=..., __functor=...) at /home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/functional:1926
#3  function<spark_variable(char const*, void const*, Spark_Data_TypeDef, void*)::<lambda()>, void> (__f=..., this=0x20002798)
    at /home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/functional:2426
#4  FFL<spark_variable(char const*, void const*, Spark_Data_TypeDef, void*)::<lambda()> > (func=...) at ./inc/system_threading.h:95
#5  spark_variable (varKey=0x80a0802 "b", userVar=0x20000308, userVarType=<optimized out>, reserved=0x0) at src/system_cloud.cpp:86
#6  0x080643e2 in app_loop (threaded=<optimized out>) at src/main.cpp:186
#7  0x0806302e in operator() (this=0x200161c0 <ApplicationThread+4>) at /home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/functional:2439
#8  ActiveObjectBase::run (this=this@entry=0x200161bc <ApplicationThread>) at src/active_object.cpp:51
#9  0x08064488 in start (this=0x200161bc <ApplicationThread>) at ./inc/active_object.h:313
#10 app_setup_and_loop () at src/main.cpp:256
#11 0x08067c32 in application_start () at src/stm32f2xx/core_hal_stm32f2xx.c:473
#12 0x0807aa2a in application_thread_main (arg=<optimized out>) at WICED/RTOS/FreeRTOS/WICED/wiced_rtos.c:177
#13 0x00000000 in ?? ()
   
Registers at start of HardFault_Handler ()

xPSR            0xa1000000
ReturnAddress   0x8080900
LR (R14)        0x8062a97
R12             0x0
R3              0xffffffff
R2              0xfffc17e8
R1              0x2e2d20
R0              0x2001ec5c
Return instruction:
   0x8080900 <_malloc_r+44>:	ldr	r2, [r1, #0]

*/
```

bkobkobko:
I would try adding the `volatile` keyword to your Particle variables.  Or you could increment them in loop() or something.  It looks like gcc decided it could optimize them out.

monkbroc:
The same crash happens with my entire application (with variables computed
in loop) at the same spot.

hal/src/photon/concurrent_hal.cpp
m-mcgowan: :+1:

m-mcgowan:
I tried the app with the latest state (just pushed) and don't see a crash. (Although in principle there are no changes regarding the thread invocation code, so I would have expected it to also work with the previous code.)

monkbroc:
> don't see a crash

:disappointed: I hate to pull the "it works on my machine" argument (or in this case "it fails on my machine").

My test program still crashes with the latest update. It crashes if I add `volatile` to the variables.

I build like this with the test code in `user/src/application.cpp`:
```
modules$ make clean all program-dfu PLATFORM=photon
```

Note that I have to comment out this line in `system_threading.cpp`:
```
    __future_base::_State_base::~_State_base() = default;
```

Otherwise I get this error:
```
src/system_threading.cpp:52:5: error: definition of explicitly-defaulted 'std::__future_base::_State_baseV2::~_State_baseV2()'
     __future_base::_State_base::~_State_base() = default;
     ^
In file included from ./inc/active_object.h:27:0,
                 from ./inc/system_threading.h:25,
                 from src/system_threading.cpp:3:
/home/monkbroc/bin/gcc-arm-embedded/gcc-arm-none-eabi-4_9-2015q2/arm-none-eabi/include/c++/4.9.3/future:316:15: error: 'virtual std::__future_base::_State_baseV2::~_State_baseV2()' explicitly defaulted here
       virtual ~_State_baseV2() = default;
               ^
```

My compiler version is
```
arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 4.9.3 20150529 (release) [ARM/embedded-4_9-branch revision 224288]
```

m-mcgowan:
Ok, I'm using 4.8.4 20140725. I have 4.9.3 - I'll try that next time I pass by the multithreading stuff again.

system/system-threading.md
```diff
@@ -0,0 +1,389 @@
...
+
+2. Have a generic way to run code asynchronously on the system thread
+
+```
+ON_SYSTEM_THREAD(
+   Particle.subscribe("event", theHandler);
+   Particle.subscribe("metoo", anotherHandler);
+);
+```
+
```
monkbroc: Or 3. Change the semantics when threading is enabled to make blocking functions become non-blocking. That's the point of system threading, right? Particle.subscribe only blocks on sending the subscription, which doesn't even factor into the return value.

m-mcgowan: I'm not sure I'm following you. The function returns a bool for success/failure. If someone were to register 10 subscriptions, most of them would fail since there's a maximum of 4 subscription. We can't simply throw away that status information, and we have to honour the original contract that the result indicates success/failure.

monkbroc: Particle.subscribe will block until the event is registered in the cloud, but will return true/false only based on adding the subscription to the local list. So why not make it return true/false right away and perform the registration asynchronously without looking at the return value. It changes the semantics, but it's OK since threading is opt-in.

m-mcgowan: Ah, understood. Yes for this specific method that is a possibility, but it won't save much. The issue is that the system thread could be busy servicing the network, so any function calls that execute on the system thread will block until the system thread is free to service them. So even just adding the event to the local store will block. The alternative is to make the local event store thread-safe so that it can be updated from non-system threads, but that's something I'd rather not do since it introduces a lot more complexity.

In this case, I feel the caller can simply request the entire Particle.subscribe() to run on the system thread as I outlined int he last example.

monkbroc:
Regardless of when you pull the changes in #647, take a look at my change for thread priorities here:
https://github.com/spark/firmware/commit/1d4816835277fc96376661b543289f264d2e4f1b

Having OS_THREAD_PRIORITY_DEFAULT = 0 and the std::thread creation routine use priority 0 meant that the system thread and other new threads ran at max priority.

Let's just use the FreeRTOS convention instead of the WICED convention for thread priorities: low numbers mean low priority.

m-mcgowan:
Definitely, those priority changes are good, thanks for that. You should be able to build and run the latest state now without issue, fixed the SOS that was happening.

m-mcgowan:
darn. I wanted to merge some of this, but github has close the PR. I will merge the remainder manually.

PR #619, opened 2015-09-16
strict typechecks for Particle.variable()

m-mcgowan:
Catches this error:

```
Particle.variable("myvar", &string, STRING);
```

which comes up often in community discussions.

### DIFF
--- a/system/inc/system_cloud.h
+++ b/system/inc/system_cloud.h
@@ -59,31 +59,34 @@ typedef enum
 	CLOUD_VAR_BOOLEAN = 1, CLOUD_VAR_INT = 2, CLOUD_VAR_STRING = 4, CLOUD_VAR_DOUBLE = 9
 } Spark_Data_TypeDef;
 
+#ifdef PLATFORM_ID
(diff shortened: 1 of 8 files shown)


system/inc/system_cloud.h
```diff
@@ -56,9 +56,43 @@ void system_set_time(time_t time, unsigned param, void* reserved);
 
 typedef enum
 {
-	BOOLEAN = 1, INT = 2, STRING = 4, DOUBLE = 9
```
bkobkobko: Is this backward compatible?  These enums are what users write in their code, aren't they?

m-mcgowan:
It's compatible - we have a suite of API tests that still compile.

system/inc/system_cloud.h
m-mcgowan: It's compatible - we have a suite of API tests that still compile.  The names are now assigned to class constants instead, `INT`, `STRING` etc. so we get more complete type info about what the corresponding variable type should be.

PR #621, opened 2015-09-17
Another attempt at merging develop into feature/electron-socket

brycekahle:
This is a new attempt at merging develop into feature/electron-socket before Brett's initial merge commit (a77496a032756c237982a21abcf19cb2f1abf791).

Compiles for all platforms. Wifi connect behavior tested on a Photon.

I used P4Merge to handle merge conflicts. When conflicted code had moved files, I had to search for the new location and apply the changes from develop manually. A small number of changes were required to get it compiling on the Core (just a cast for the `void*`). 

There was some confusion around `network.listen()`. The parameter `stop` is confusing because `network.listen(true)` looks like it should start listening, but actually stops. We should consider refactoring this. 

Brett reviewed all the diffs and changes with me. Twice.

cc @m-mcgowan @technobly


technobly:
Looks good!  P4merge for the win!  Great job Bryce!  I'm going to go through this with P4merge myself just to get acquainted with the new tool and a very difficult merge.  If you compare this to the latest commit in feature/electron-socket you'll see the various bits of code that were completely lost in the initial merge.

m-mcgowan:
:+1 agreed on the stop parameter. Let's change that.

Issue #616, opened 2015-09-15
The cloud connection status is set to disconnected when cloud events are not handled within 15 seconds


m-mcgowan:
Reports in the community of the device connection status not matching the cloud connection status.

What can happen is that the application thread can block on something, which then causes the cloud connection to terminate after 15 seconds even though the device is still breathing cyan.

When the app is blocked, no cloud commands can be serviced. After 15 seconds or more of being blocked then the cloud connection is terminated by the cloud.

I'd like to add a status LED notification when this happens so the device indicates that it's not online (e.g. flashing cyan or breathing green would be good cc: @jenesaisdiq) so at least the local cloud connection status indicator (the LED) and the actual cloud connection status are in agreement.

When the application becomes unblocked again, the cloud connection will be re-established.

PR #622, opened 2015-09-17
cloud disconnect detect

m-mcgowan:
addresses #616 - revert to breathing green when the cloud isn't serviced for 15 seconds.  System modes are unchanged - in automatic mode, the system will try to reconnect once the background loop is executed again. Semi-automatic and manual mode requires the application code to call `Particle.connect()`.

### DIFF
--- a/system/src/main.cpp
+++ b/system/src/main.cpp
@@ -33,6 +33,7 @@
 #include "system_task.h"
 #include "system_network.h"
 #include "system_network_internal.h"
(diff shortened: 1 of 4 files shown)


m-mcgowan:
To test add a delay to your loop():

```
void loop()
{
    static int count = 0;
    if ((count++ % 5)==0)
        HAL_Delay_Milliseconds(20*1000);
}
```

PR #625, opened 2015-09-18
removed mbed license

kennethlimcp:
Not sure if this is used but it should be replaced with a Particle license if needed?

Signed-off-by: Kenneth Lim <kennethlimcp@gmail.com>


m-mcgowan:
We cannot change the license of someone else's work without permission from them.

kennethlimcp:
I thought i would just mention it since i see that files in the `modem` directory has Particle license in it and i don't see where this license is applied to.

Issue #598, opened 2015-09-09
Addition of Wire.reset()


pomplesiegel:
We have experienced an issue wherein a MCP23017 port expander gets into a bad state and holds SDA down to ground indefinitely. When this occurs, it seems the only way to purge the bus is to send additional clock pulses down SCL to force the slave to release SDA. See [this AD app note] (http://www.analog.com/media/en/technical-documentation/application-notes/54305147357414AN686_0.pdf) and [this forum post] (http://www.microchip.com/forums/m175368.aspx).
After days of testing, it appears the solution below is quite effective, successfully purging the bus on 7 photons hundreds of times over this period. Would it make sense to include something similar to this as `Wire.reset()` in the Wire library?

````cpp
void RestartI2cBus()
{
  HAL_Pin_Mode(D0, INPUT_PULLUP); //Turn SCA into high impedance input
  HAL_Pin_Mode(D1, OUTPUT); //Turn SCL into a normal GPO 
  HAL_GPIO_Write(D1, HIGH); // Start idle HIGH
  
  //Generate 9 pulses on SCL to tell slave to release the bus
  for(int i=0; i <9; i++)
  {
    HAL_GPIO_Write(D1, LOW);
    delayMicroseconds(100);
    HAL_GPIO_Write(D1, HIGH);
    delayMicroseconds(100);
  }

  //Change SCL to be an input
  HAL_Pin_Mode(D1, INPUT_PULLUP);

  //Start i2c over again
  Wire.begin(); 
  delay(50);
}
````

m-mcgowan:
This sounds like a great idea!

pomplesiegel:
@m-mcgowan, Awesome! Glad to see this made it to 0.4.6.

PR #626, opened 2015-09-18
I2C fix for MCP23017 / Adafruit RGBLCDShield

m-mcgowan:
contribution from @pra in the community. Thanks! https://community.particle.io/t/i2c-lcd-freezes-photon/15666

If you have a MCP23017 or Adafruit RGBLCDShield please test and report if this improves I2C connectivity!

### DIFF
--- a/hal/src/stm32f2xx/i2c_hal.c
+++ b/hal/src/stm32f2xx/i2c_hal.c
@@ -315,26 +315,10 @@ uint32_t HAL_I2C_Request_Data(HAL_I2C_Interface i2c, uint8_t address, uint8_t qu
     /* While there is data to be read */
     while(numByteToRead)
     {
-		switch	(numByteToRead) {
-			case 2:
-				/* Disable Acknowledgement on last byte which is being assembled in shift register*/
-				I2C_AcknowledgeConfig(i2cMap[i2c]->I2C_Peripheral, DISABLE);
-				break;
-
-			case 1:
-				/* Send STOP Condition */
-				if (stop == true)
-					I2C_GenerateSTOP(i2cMap[i2c]->I2C_Peripheral, ENABLE);
-				break;
-
-			default:
-				break;
-		}
-
-
-        /* Wait for the byte to be received */
+        /* Wait for data register to be full */
+		/* This means the byte in the RX has already been acked */
+		/* Ack is sent on the xfer of teh byte to RX from the shift register */
         _millis = HAL_Timer_Get_Milli_Seconds();
-        //while(!I2C_CheckEvent(i2cMap[i2c]->I2C_Peripheral, I2C_EVENT_MASTER_BYTE_RECEIVED))
         while(I2C_GetFlagStatus(i2cMap[i2c]->I2C_Peripheral, I2C_FLAG_RXNE) == RESET)
         {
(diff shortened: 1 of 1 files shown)


dresslerc:
I've been using the fix on 2 MCP23017 for 24 hours and they are still running strong.

My test code would usually freeze the i2c bus within 10-30 mins and zero issues so far.  I'll keep testing and report back after it runs for a couple days.

pomplesiegel:
The MCP23017 locking up SDA was the original reason we were testing the fix `Wire.reset()` #598. Hopefully this takes things to the next level, preventing the problem before it occurs.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -1,726 +1,740 @@
...
+            /* SW Reset the I2C Peripheral */
+            HAL_I2C_SoftwareReset(i2c);
+
+            return 0;
+        }
+    }
+
+    /* perform blocking read into buffer */
+    uint8_t *pBuffer = i2cMap[i2c]->rxBuffer;
+    uint8_t numByteToRead = quantity;
```
technobly: These three lines and comment are disconnected from the code they are used with on line 315.  Doesn't look like they needed to move up here.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -1,726 +1,740 @@
...
+
+	if (quantity == 1)
+		/* Disable Acknowledgement */
+		I2C_AcknowledgeConfig(i2cMap[i2c]->I2C_Peripheral, DISABLE);
```
technobly: Disabling ACK if quantity == 1 here (no comment as to why that is needed before Slave Address send), but disabling ACK if numByteToRead (aka quantity) == 2 below later on receive.  Can we explain this difference better in comments?

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -1,726 +1,740 @@
...
+        }
+    }
+
+
+    /* While there is data to be read */
+    while(numByteToRead)
+    {
+        /* Wait for data register to be full */
+		/* This means the byte in the RX has already been acked */
+		/* Ack is sent on the xfer of teh byte to RX from the shift register */
```
technobly: These comments are very confusing.  I want to fix them, but do not fully understand yet what was changed here.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -1,726 +1,740 @@
...
+                        I2C_GenerateSTOP(i2cMap[i2c]->I2C_Peripheral, ENABLE);
+                break;
+
+            default:
+                break;
+        }
+
+        /* Read the byte from the Slave */
+		/* This will enhable the transfer of the next byte (if any) from the shift register to the RX */
+		/* and also sending ACK/NACK for that byte */
```
technobly: These comments are easier to understand, but some reference to numByteToRead >2, == 2 and == 1 in context would be helpful.

technobly:
Not trying to nitpick this change which is obviously working for users, just trying to make sure this fairly  complicated peripheral gets documented adequately as we make improvements ;-)  Thus ensuring the changes are widely understood and stay fixed as the code evolves.

PR #631, opened 2015-09-20
Don't modify registers in HardFault_Handler

monkbroc:
To make it easier to figure out where a hard fault came from I set a breakpoint in HardFault_Handler but the variables r1, r2, r3, r12, lr, pc, psr are incremented because of the missing braces in if(false). It's cleaner to just use (void)variable to silence the compiler error.


m-mcgowan:
Thanks. makes perfect sense. This is code I copied from an example online so I should have sanity checked it.

PR #636, opened 2015-09-22
Fix LED states

brycekahle:
* Fix LAC and Cell ID parsing
* Fix value used for finding a free socket
* If a SIM is not present after powerOn completes, the Electron will
flash blue.
* Cellular connection is moved to connect_finalize so it flashes green
during cellular registration and attachment.

Note: I was unable to get the uBlox module to allow persistent SIM polling. Thus if a SIM is not detected at startup, once it is inserted a RESET is required.

LED now behaves as below:

Power on: breathing white for a second or two
SIM present? No - flashing blue
Cellular connecting - Flashing green
Cloud connecting - Flashing cyan
Connected - Breathing cyan

### DIFF
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -36,6 +36,8 @@ typedef int cellular_result_t;
  */
 cellular_result_t  cellular_on(void* reserved);
 
+cellular_result_t  cellular_init(void* reserved);
+
 /**
  * Power off the cellular module.
  */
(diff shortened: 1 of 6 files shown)


hal/inc/cellular_hal.h
```diff
@@ -122,6 +124,7 @@ cellular_result_t cellular_credentials_set(const char* apn, const char* username
  */
 CellularCredentials* cellular_credentials_get(void* reserved);
 
+bool cellular_sim_ready(void);
```
technobly: `bool cellular_sim_ready(void* reserved);` and we need to make sure `cellular_sim_ready` and `cellular_init` get added to the DYNALIB once the modular PR is merged to develop.  Looks good Bryce!

m-mcgowan: modular is merged to develop so go ahead and add those to the cellular dynalib in `hal/inc/hal_dynalib_cellular.h`

brycekahle:
Updated with dynalib changes

PR #644, opened 2015-09-27
Include uxTopUsedPriority in non-modular build

monkbroc:
Include uxTopUsedPriority used by OpenOCD for FreeRTOS support in main.elf when building non-modular firmware.


m-mcgowan:
Thanks. This should go in build.mk. I'll close this and make the corresponding change.

PR #647, opened 2015-09-28
Update Wiring Thread implementation

monkbroc:
### Changes

* Include HAL thread functions in dynalib
* Allow lambdas as thread functions
* Support periodic tasks with the helper class PeriodicWork, which delegates to vTaskDelayUntil to ensure constant execution period
* Add simple CriticalSection concurrency primitive
* Removed ThreadFactory because `thread t = Thread.create(...)` is different than other Wiring APIs

### Motivation

The Particle platform is missing a straightforward way to service hardware at a regular interval in a predictable fashion with absolutely no possibility of blocking. Adding lightweight Thread support to Wiring allows this. PeriodicWork ensures that periods will be accurate, especially for short periods (< 5ms).

We've been having lots of discussions in #614 about blocking/non-blocking calls. Even with the multi-threading approach many calls will block for unpredictably long periods of time. I've been thinking about this a lot and rather than fix the blocking, I think it will be easier to provide a way to deal with the physical world in its own reserved thread.

### Approach

 Wiring threads are complementary to multithreading. It's possible to use either or both.

Personally with Wiring threads I don't think my applications will need SYSTEM_THREAD(ENABLED). I'll just leave all the blocking stuff together in setup() and loop() and service the hardware in its own thread.

It will be up to developers to only use appropriate calls in their threads (GPIO, I2C, SPI, etc) and make sure multiple threads don't use the same peripherals. Cloud calls should be made on the application thread.

For concurrency control, I think a simple CriticalSection will go a long way, but this is the weakest area of this PR. More concurrency primitives could be added later if they are needed.

### (Contrived) Example

```c++
#include "application.h"
void setup()
{
    RGB.control(true);

    Thread hardwareThread = Thread("hardware", [&]() {
        // 100ms periodic task
        PeriodicWork work(100);

        bool state = false;
        for(int i = 0; i < 10; i++) {
            work.start();
            RGB.color(0, 0, state ? 180 : 0);
            state = !state;
            // Will sleep for 100ms from the call to work.start()
            work.end();
        }
        // a task should never return, but delete itself when it is done
        // but most tasks will live forever anyway
        hardwareThread.dispose();
    }, OS_THREAD_PRIORITY_DEFAULT + 1);

    // join should be as rare as dispose
    hardwareThread.join();
    RGB.color(0, 255, 0);
}
```

### Next step

I'll start using this in my work and personal applications this week and report back here.

Unleash the comments :stuck_out_tongue_winking_eye:


m-mcgowan:
Thanks. 

- :+1: for the HAL concurrency functions. I'd also like to expose what is needed so that std::thread, std::mutex et. al. can be used in application code.
- :+1: For lambdas as thread functions.
- :+1: For removing the thread factory, if we can do away with it. This was a working concept design and far from being finalized. 
- :+1: for Critical section RAII.  I also wonder if that should disable interrupts as well. 

For periodic work, I'm wondering why we don't expose [FreeRTOS timers](http://www.freertos.org/RTOS-software-timer.html).   This could be wrapped up in a high-level `Timer` class:

```
   Timer t(2*MINUTE, handleTimer)
   t.start();
```

We'd also provide options for one-shot timers, and also the ability to use subclassing for the timer callback.

monkbroc:
> Critical section RAII. I also wonder if that should disable interrupts as well.

Critical section uses portENTER_CRITICAL/portEXIT_CRITICAL which disables interrupt and has a nesting counter.

> For periodic work, I'm wondering why we don't expose FreeRTOS timers.

Because FreeRTOS timers essentially run in interrupt context, not as a normal task that can be pre-empted. In a timer you can't use any system calls that may use delay(). Here be :dragon: 

From the bottom of the page you linked:
> Timer callback functions execute in the context of the timer service task. It is therefore essential that timer callback functions never attempt to block. For example, a timer callback function must not call vTaskDelay(), vTaskDelayUntil(), or specify a non zero block time when accessing a queue or a semaphore.

Regarding C++ thread primitives, I'm not really excited about using those in an embedded environment. The API surface is large (many types with subtly different semantics) so they are hard to reason about. They are quite complex to implement since they don't match 1-to-1 with the RTOS. There's still stuff missing like a thread-safe queue. Wouldn't you rather just wrap up what FreeRTOS gives you and call it a day?

m-mcgowan:
Re: the large surface area of the C++ threading primitives, that's true from a client perspective (they are fully featured) but not from an implementation perspective. I made std::thread, std::mutex work by only providing a handful of functions in the HAL to implement missing pieces in gth.h that isn't supported by ARM GCC. Sure - if there are other currency primitives that aren't so easily ported, then we can provide an API in hal and implement our own high level equivalents, but I would look to start with the standard first and only move away when there's hinderances preventing that. 

PeriodicWork.start()/end() - that seems to be a nice case for a RAII class. e.g.

```
    PeriodicWork work(100);
    for (;;) {
        PeriodicIteration iteration(work);
         // ... 
    }
```

I still think for the majority of cases timers are fine if the operations performed can be done quickly, e.g. toggle a pin, or start a DMA SPI transaction. Using a dedicated thread is ok too, but will require more resources, particularly the stack in RAM. 

Your discussion of not needing SYSTEM_THREAD(ENABLED) - that's fine if you're an advanced coder that want's to get involved with threading directly, but for beginners, having an application loop that runs freely always (like Arduino) is a good thing, and is what many people expect to happen when they come to the Particle platform.

monkbroc:
That's why I'm presenting this as complementary to SYSTEM_THREAD, not instead of. Beginners can use one `loop()` until that doesn't work anymore for their particular project.

One of my personal projects is an espresso machine controller. I want absolutely no chance of the heater being stuck on in case there's a WiFi hiccup (it has happened several times before. Luckily there's a failsafe thermostat). A dedicated periodic hardware thread is perfect to read the analog temperature, compute the control loop, update the display through I2C and command the relay with simple IO and `delay()`.

For the concurrency primitives, I was thinking of promises and conditional variables. CVs need a thread queue, a mutex and a binary semaphore per thread. Promises need a mutex, a CV and a std::function. Those all need to be deleted safely. That's a lot of stuff to get right.

Good suggestion for the periodic iteration. Can you think of a better name for PeriodicWork? It isn't great...

andyw-lala:
I have to raise a questioning eyebrow at this:

I still think for the majority of cases timers are fine if the operations
performed can be done quickly, e.g. toggle a pin, or start a DMA SPI
transaction.


Unless you're willing to document the cases, and spend even more cycles
answering questions in the forums, people will put publish(), printf() and
all kinds of rubbish  in these timers. I see a conflict between this
statement and the assertion that we need to plan for the average arduino
user in your next paragraph.

I realise that we are up to our necks in trade-offs at this point, some of
them quite tricky - but I have to state that exposing more interrupt
contexts where some reasonable number of calls can cause (possibly hard to
diagnose) problems is something that should be considered very carefully.

I'm still trying to work through in my own mind a good set of use cases for
all levels of users, and what those map onto in terms of services and
programming environment on the devices themselves.

On Mon, Sep 28, 2015 at 2:19 PM, Julien Vanier <notifications@github.com>
wrote:

> That's why I'm presenting this as complementary to SYSTEM_THREAD, not
> instead of. Beginners can use one loop() until that doesn't work anymore
> for their particular project.
>
> One of my personal projects is an espresso machine controller. I want
> absolutely no chance of the heater being stuck on in case there's a WiFi
> hiccup (it has happened several times before. Luckily there's a failsafe
> thermostat). A dedicated periodic hardware thread is perfect to read the
> analog temperature, compute the control loop, update the display through
> I2C and command the relay with simple IO and delay().
>
> For the concurrency primitives, I was thinking of promises and conditional
> variables. CVs need a thread queue, a mutex and a binary semaphore per
> thread. Promises need a mutex, a CV and a std::function. Those all need to
> be deleted safely. That's a lot of stuff to get right.
>
> Good suggestion for the periodic iteration. Can you think of a better name
> for PeriodicWork? It isn't great...
>
> —
> <https://github.com/spark/firmware/pull/647#issuecomment-143848745>.
>



-- 
Andy

m-mcgowan:
To be precise, the timers don't execute on an interrupt context, but rather they execute on a thread - the timer service thread - so calling Serial and other longer duration operations will not cause problems in the same way a hardware timer interrupt would. The sole reason not to block in a timer handler is so that other timers can be serviced. Timers are meant to be lightweight periodic actions - and will be documented as such. So even the expresso service loop could be run as a timer if the duration is small compared to the timer granularity required. If the user needs something more heavyweight, then a new thread is the way to go. Ultimately, timers and threads are two tools in a toolbox that we have to provide guidance about which to use for different situations.

m-mcgowan:
> CVs need a thread queue, a mutex and a binary semaphore per thread. Promises need a mutex, a CV and a std::function. Those all need to be deleted safely. That's a lot of stuff to get right.

C++ takes care of this for you if you structure the objects correctly (RAII). If dynamic allocation is required there are also pointer wrappers that can help manage that using RAII. I know you know this, so I'm surprised you're concerned by this since given there are tools and techniques to manage the complexity.

monkbroc:
I wouldn't say I'm concerned. Just that a lot of stuff needs to be checked for correct behavior. At least one other person agrees with me :smile: 

memory leaks

system/src/main.cpp
```diff
@@ -263,19 +263,24 @@ void app_setup_and_loop(void)
...
+        SystemThread.start();
         ApplicationThread.start();
     }
     else
-#endif
     {
+        SystemThread.setCurrentThread();
+        ApplicationThread.setCurrentThread();
+    }
+#endif
```
monkbroc: @m-mcgowan Do you agree with this: if PLATFORM_THREADING is compiled in, but threading is disabled, the thread that runs `loop()` should return true for `isCurrentThread()` for both the application and system thread. I achieved this by marking `SystemThread` and `ApplicationThread` as attached to the current thread.

monkbroc:
Rebased to latest feature/non-blocking

system/inc/active_object.h
```diff
@@ -280,6 +280,10 @@ class ActiveObjectBase
         return _thread_id == std::this_thread::get_id();
     }
 
+    void setCurrentThread() {
```
m-mcgowan: could you make this a protected method - seems to make sense from the subclass rather than general public access.

system/src/system_task.cpp
```diff
@@ -368,12 +368,13 @@ void system_delay_ms_non_threaded(unsigned long ms, bool force_no_background_loo
 
 void system_delay_ms(unsigned long ms, bool force_no_background_loop=false)
 {
-    if (system_thread_get_state(NULL)!=0) {
-        HAL_Delay_Milliseconds(ms);
+    if (system_thread_get_state(NULL) == spark::feature::DISABLED &&
```
m-mcgowan: I'm not sure I understand this. If threading is disabled, then application/system thread has no real meaning. Could you please clarify.

m-mcgowan:
I've reviewed this in detail.  A great contribution overall, thanks!  I'd prefer to drop the PeriodicWork while we work on that separately.  I'm also confused about needing to reference  APPLICATION_THREAD_CURRENT() when threading is disabled since that will always be true..

system/src/system_task.cpp
monkbroc: Yes, the ApplicationThread / SystemThread active objects don't have a meaning when threading is disabled (yet the objects still exist), but SYSTEM_THREAD_CURRENT / APPLICATION_THREAD_CURRENT have a clear meaning.

To explain, I'll work back from the issue I had.

`delay()` should only call `Spark_Idle_Events` when called from the user's `setup()/loop()` and threading is disabled. It should not be called from the system thread or from a wiring thread.

How to make the difference for that last case? Check if the current thread is the application thread.

When threading is disabled, the application thread and system thread are the original one created by WICED. They just happen not to be running a deferred execution loop.

So in main.cpp, the ApplicationThread and SystemThread are associated with the current thread if threading is disabled so that SYSTEM_THREAD_CURRENT / APPLICATION_THREAD_CURRENT return the correct result with Wiring threads.

Similarly if PLATFORM_THREADING is not compiled in, SYSTEM_THREAD_CURRENT / APPLICATION_THREAD_CURRENT always return true.

With this change, you can drop the check of `isStarted` in the `SYSTEM_THREAD_CONTEXT_...` macros because `isCurrentThread` will return the correct result whether threading is enabled or not. Do you agree?

system/inc/active_object.h
monkbroc: This has to be public since it's called from main.cpp.

I thought of calling this `attach` or `attachToCurrentThread`. Would that make more sense?

system/src/system_task.cpp
monkbroc: I dunno if a blurb where 15% of the words contain "thread" counts as clarification :laughing:

system/inc/active_object.h
m-mcgowan: I think I'd rather have an optional parameter to start()  that indicates if threading is enabled or not, so that the thread internally takes care of setting up the thread ID. 

But really I'm not seeing why we need to reference SystemThread/ApplicationThread when threading is disabled. (I haven't read the next comment yet.)

monkbroc:
I'll take out the PeriodicWork stuff since I can easily reimplement this in my application as long as `os_thread_delay_until` stays in the HAL

system/src/system_task.cpp
m-mcgowan: Ok, I see. I wasn't considering the case of creating additional threads since this release is focused on just the system/application thread.

monkbroc:
Thanks Matt

PR #654, opened 2015-10-01
Make the errors clearer for Particle.variable()

monkbroc:
Errors with Particle.variable("name", ..., STRING); are common. The changes in 0.4.7 introduce errors for incorrect cases instead of silently compiling. However the errors are weird template errors that are absolutely incomprehensible.

This at least shows clear error to users.

```C++
char buf[10];
String str;
Particle.variable("name", &buf, STRING);
Particle.variable("name", str, STRING);
Particle.variable("name", &str, STRING);
```
```
../wiring/inc/spark_wiring_cloud.h:67:9: error: static assertion failed:

Use Particle.varible("name", myVar, STRING); without & in front of myVar

../wiring/inc/spark_wiring_cloud.h:73:9: error: static assertion failed:

In Particle.varible("name", myVar, STRING); must be declared as char myVar[] not String myVar
```

See this forum discussion for why this is necessary...
http://community.particle.io/t/error-no-matching-function-for-call-to-cloudclass-variable/16125

### DIFF
--- a/wiring/inc/spark_wiring_cloud.h
+++ b/wiring/inc/spark_wiring_cloud.h
@@ -60,6 +60,26 @@ public:
         return CLOUD_FN(spark_variable(varKey, (const void*)userVar, CloudVariableTypeInt::value(), NULL), false);
     }
 
+    // Return clear errors for common misuses of Particle.variable()
+    template<typename T, std::size_t N>
+    static inline bool variable(const char *varKey, const T (*userVar)[N], const CloudVariableTypeString& userVarType)
+    {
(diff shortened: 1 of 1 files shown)


monkbroc:
Or if you feel generous, you could actually accept `Particle.variable("name", &myVar, STRING)` and just dereference the pointer.

```C++
    template<typename T, std::size_t N> static inline bool variable(const char *varKey, const T (*userVar)[N], const CloudVariableTypeString& userVarType)
    {
        return variable(varKey, *userVar, userVarType);
    }

```

m-mcgowan:
Thanks - this makes much sense. I was thinking of having the IDE compiler simplify the compiler output - this does the job nicely.

PR #656, opened 2015-10-02
pulseIn() for Photon/P1/Core/Electron, with tests

technobly:
Quick implementation of a [**pulseIn()**](https://www.arduino.cc/en/Reference/PulseIn) from Arduino/Wiring.
* Times out automatically after 3 seconds to keep from blocking background tasks.
* Tests ran and passed on Photon and Core, but Electron TEST=wiring/no_fixture has compiler errors from different tests, so could not run for E.  Ran tests for P1 and pulseIn() passed but about 8 other tests failed :(
* Thought it might be hard to test pulseIn() without a fixture, but surprise... you can read the state of an output when the PWM peripheral is banging out bits :joy:

### DIFF
--- a/hal/src/stm32f2xx/gpio_hal.c
+++ b/hal/src/stm32f2xx/gpio_hal.c
@@ -28,6 +28,7 @@
 #include "pinmap_impl.h"
 #include "stm32f2xx.h"
 #include <stddef.h>
+#include "hw_ticks.h"
 
 /* Private typedef ----------------------------------------------------------*/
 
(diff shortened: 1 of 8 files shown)


hal/src/stm32f2xx/gpio_hal.c
```diff
@@ -230,3 +231,37 @@ int32_t HAL_GPIO_Read(uint16_t pin)
...
+    STM32_Pin_Info* SOLO_PIN_MAP = HAL_Pin_Map();
+    #define pinReadFast(_pin) ((SOLO_PIN_MAP[_pin].gpio_peripheral->IDR & SOLO_PIN_MAP[_pin].gpio_pin) == 0 ? 0 : 1)
+
+    volatile uint32_t timeoutStart = SYSTEM_TICK_COUNTER; // total 3 seconds for entire function!
+
+    /* If already on the value we want to measure, wait for the next one.
+     * Time out after 3 seconds so we don't block the background tasks
+     */
+    while (pinReadFast(pin) != value) {
+        if (SYSTEM_TICK_COUNTER - timeoutStart > 360000000UL) {
```
m-mcgowan: I noticed these timeout values are the same on the stm32f2 and on the core - would think they would need to be different due to the different clock speed?

technobly: Good catch!  Added a test first, saw it fail... fixed the code, ran Core and Photon tests again and they passed :)

m-mcgowan: Woohoo! TDD ftw!

PR #664, opened 2015-10-04
initialize backup sram only on first use

technobly:
* The code currently (v0.4.6) attempts to initialize SRAM if it's not waking up from Standby mode, but should instead be initializing based on if it powered up when the Backup RAM is enabled. Since it's a bit tricky to store a system variable in SRAM for this purpose, checking the Backup Regulator Ready status bit will do nicely; which is a hardware controlled bit that won't be cleared if we are waking up from Standby, software or hardware reset. This backup regulator is only enabled if we are enabling Backup RAM, so this is perfect.

* I've been testing this in all modes of operation it's rock solid now. Docs really need a matrix table to show all of the different states for SRAM. This will make it super clear when SRAM is initialized and when it's retained (at least that's the hope).  

* Fixes issue: https://github.com/spark/firmware/issues/661

* What do you think about this table? Is the expectation for SRAM clear under the following conditions?

| Power_Down_Method | Power_Up_Method | When VIN Powered | When VBAT Powered | SRAM Initialized | SRAM Retained |
| :-: | :-: | :-: | :-: | :-: | :-: |
| Power removed on VIN and VBAT | Power applied on VIN | - | - | Yes | No |
| Power removed on VIN | Power applied on VIN | - | Yes | No | Yes |
| System.sleep(SLEEP_MODE_DEEP) | Rising edge on WKP pin, or Hard Reset | Yes | Yes/No | No | Yes |
| System.sleep(SLEEP_MODE_DEEP,10) | RTC alarm after 10 seconds | Yes | Yes/No | No | Yes |
| System.reset() | Boot after software reset | Yes | Yes/No | No | Yes |
| Hard reset | Boot after hard reset | Yes | Yes/No | No | Yes |

### DIFF
--- a/wiring/inc/spark_wiring_system.h
+++ b/wiring/inc/spark_wiring_system.h
@@ -156,8 +156,18 @@ public:
 
     inline int enableFeature(HAL_Feature feature)
     {
+        /* PRE-ENABLE TASKS */
+        bool init_sram = false;
+        // Initialize SRAM on first use of retained memory
+        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_RETAINED_MEMORY)) {
+            init_sram = true;
+        }
+
+        /* ENABLE FEATURE */
         int result = HAL_Feature_Set(feature, true);
-        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_WARM_START)) {
+
+        /* POST-ENABLE TASKS */
+        if (feature==FEATURE_RETAINED_MEMORY && init_sram) {
             system_initialize_user_backup_ram();
         }
         return result;
--- a/wiring/inc/spark_wiring_system.h
+++ b/wiring/inc/spark_wiring_system.h
@@ -156,21 +156,7 @@ public:
 
     inline int enableFeature(HAL_Feature feature)
     {
-        /* PRE-ENABLE TASKS */
-        bool init_sram = false;
-        // Initialize SRAM on first use of retained memory
-        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_RETAINED_MEMORY)) {
-            init_sram = true;
-        }
-
-        /* ENABLE FEATURE */
-        int result = HAL_Feature_Set(feature, true);
-
-        /* POST-ENABLE TASKS */
-        if (feature==FEATURE_RETAINED_MEMORY && init_sram) {
-            system_initialize_user_backup_ram();
-        }
-        return result;
+        return HAL_Feature_Set(feature, true);
     }
 
     inline int disableFeature(HAL_Feature feature)
--- a/build/arm/linker/stm32f2xx/backup_ram.ld
+++ b/build/arm/linker/stm32f2xx/backup_ram.ld
@@ -1,0 +1,8 @@
+    .backup :
+    {
+        link_global_retained_initial_values = LOADADDR( .backup ); /* This is the location in flash of the initial values of retained global variables */
+        link_global_retained_start = .;
+        *(.retained_user*)
+        link_global_retained_end = .;
+        . = ALIGN(., 4);
+    }>BACKUPSRAM AT> APP_FLASH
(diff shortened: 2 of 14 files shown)


wiring/inc/spark_wiring_system.h
```diff
@@ -148,8 +148,18 @@ class SystemClass {
...
+        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_RETAINED_MEMORY)) {
+            init_sram = true;
+        }
+
+        /* ENABLE FEATURE */
         int result = HAL_Feature_Set(feature, true);
-        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_WARM_START)) {
+
+        /* POST-ENABLE TASKS */
+        if (feature==FEATURE_RETAINED_MEMORY && init_sram) {
```
m-mcgowan: I'm wondering if we should always initialize the backup RAM in HAL_Core_Config() or some other early startup function, and only if the backup regular is off.

technobly: I could see the only reason to do that would be to ensure SRAM is initialized when FEATURE_RETAINED_MEMORY is not enabled (for when you might want an extra 4KB of RAM that is not retained after power loss or reset).  I guess the question is, how long does 4KB of memcpy() take? Or does it even initialize more than is actually used? I'd prefer not to tack on more boot time by default, regardless of feature use.  If we assumed 42ns per byte that's about 176us... not bad I guess but is that a realistic estimate?

andyw-lala: toggle a gpio and measure the time.

On Fri, Oct 9, 2015 at 2:17 PM, Technobly <notifications@github.com> wrote:

> In wiring/inc/spark_wiring_system.h
> <https://github.com/spark/firmware/pull/664#discussion_r41667856>:
>
> >          int result = HAL_Feature_Set(feature, true);
> > -        if (feature==FEATURE_RETAINED_MEMORY && !HAL_Feature_Get(FEATURE_WARM_START)) {
> > +
> > +        /* POST-ENABLE TASKS */
> > +        if (feature==FEATURE_RETAINED_MEMORY && init_sram) {
>
> I could see the only reason to do that would be to ensure SRAM is
> initialized when FEATURE_RETAINED_MEMORY is not enabled (for when you might
> want an extra 4KB of RAM that is not retained after power loss or reset). I
> guess the question is, how long does 4KB of memcpy() take? Or does it even
> initialize more than is actually used? I'd prefer not to tack on more boot
> time by default, regardless of feature use. If we assumed 42ns per byte
> that's about 176us... not bad I guess but is that a realistic estimate?
>
> —
> <https://github.com/spark/firmware/pull/664/files#r41667856>.
>



-- 
Andy

technobly: Yes Andy, as I like to say... GET DUMB.  aka, don't over think it, just do it.

m-mcgowan: We are already initializing about 50 KBytes of regular RAM, so another 4K is a very small increase. I imagine we are talking sub millisecond times here, and nothing compared to the default wait of 1 second for the USB bus to initialize.   Also only the data that is used is initialized and not the entire 4KB.

technobly: Ok, good thing I went and tested this... because I found a new problem.  Before that though, initializing one `retained int` takes less than 1us (the GPIO toggling is taking some time... not sure how much, let's say very little).  Conversely initializing a `retained uint8_t array[4090] = {0};` took about 220us.  So not bad, and it's dependent on how much you are using which is great.

The problem I found, was that if I leave VBAT floating when powering up for the first time, SRAM remains uninitialized.  This was not expected as the first time we are powering up, the Backup Regulator Ready bit should be cleared.  If I attached VBAT to GND with a wire, it powers up that first time initialized.  If I attach VBAT to 3V3 with a wire, it typically also powers up initialized.  Only if floating do I typically see it not initialized.  This means when it's floating, sometimes the BRR bit is being read as SET on power up... which is bizarre.

If you are using this feature for Backup RAM, it is recommended then to have VBAT connected to a known good power source on system boot.

If you are using this feature for Extra RAM and want to ensure it always initializes, it is recommended to jumper VBAT to GND.

m-mcgowan:
Testing: 

// something like this:

```
int retained = 10;

void setup()
{
     if (!System.featureEnabled(FEATURE_RETAINED_MEMORY)) {
         // cold start
         assertEqual(retained, 10);
         System.enableFeature(FEATURE_RETAINED_MEMORY);
         retained++;
         System.reset();
    }
    else {
         assertEqual(retained, 11);
    }
}
```

Not tested so far.

PR #668, opened 2015-10-05
Revert error for Particle.variable with String

monkbroc:
Since Particle.variable String kinda works now and will be better supported soon, don't annoy users now by making code that previously compiled return an error.


m-mcgowan:
Howdy! We added full support for `String` variables in the latest release.

monkbroc:
Nice work!

PR #674, opened 2015-10-06
Electron multi-threading and OpenOCD support

monkbroc:
I needed Electron threading to work for my application so I tweaked the FreeRTOS config for it to work. With these changes I'm able to use `SYSTEM_THREAD(ENABLED)` and wiring threads.

* Call the RTOS SysTick after the OS has been set up through `SysTickChain`
* Use RTOS `vTaskDelay` in `HAL_Delay_Milliseconds`
* Add `uxTopUsedPriority` to Electron

I don't know if you're ready to add multi-threading to the Electron, but I think this will be welcomed by all the beta testers.

### DIFF
--- a/hal/src/stm32f2xx/concurrent_hal.cpp
+++ b/hal/src/stm32f2xx/concurrent_hal.cpp
@@ -30,6 +30,8 @@
 #include "semphr.h"
 #include <mutex>
 
+// For OpenOCD FreeRTOS support
+extern const int  __attribute__((used)) uxTopUsedPriority = configMAX_PRIORITIES;
 
 // use the newer name
 typedef xTaskHandle TaskHandle_t;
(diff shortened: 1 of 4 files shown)


hal/src/stm32f2xx/concurrent_hal.cpp
```diff
@@ -30,6 +30,8 @@
 #include "semphr.h"
 #include <mutex>
 
+// For OpenOCD FreeRTOS support
```
m-mcgowan: cool, but I thought it had to be extern C?

monkbroc: extern "C" is only for functions. I doesn't do anything special for variables. A simple extern is sufficient here.

PR #678, opened 2015-10-08
Setting _remoteIP by index doesn't set it as IPv4

r2jitu:
I ran into an issue where I couldn't compare the remoteIP() to another IPAddress since the version was set to 0. Using index assignment doesn't set it to 4, but assigning uint8_t* to an IPAddress does set it as IPv4.


m-mcgowan:
Have you checked this? A quick look and I'm wondering if the little-endian order is going to bite us here.

Thanks for raising this issue. I'll add a method for setting all parts of the address, so that the version is also set.

r2jitu:
This should be endian-agnostic because the assignment operator is overridden:
```c++
IPAddress& IPAddress::operator=(const uint8_t* address)
{
    set_ipv4(address[0], address[1], address[2], address[3]);
    return *this;
}
```

PR #679, opened 2015-10-08
Support RedBear Duo platform

XuGuohui:
This pull request includes integrate source files and makefiles to support Duo, a new platform from RedBear. It can be built successfully and works well on Duo. It also fixes the DFU interface issue. 
Build bootloader: "make PLATFORM=duo PARTICLE_DEVELOP=1".
Build modular firmware: Navigate to the `modules` and "make PLATFORM=duo PARTICLE_DEVELOP=1".

Thanks,
Guohui

### DIFF
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -451,6 +451,11 @@ public:
         WLAN_CAN_SHUTDOWN = 1;
     }
 
+    void notify_cannot_shutdown()
+    {
+        WLAN_CAN_SHUTDOWN = 0;
+    }
+
 
     void listen_loop() override
     {
--- a/system/src/system_network_wifi.h
+++ b/system/src/system_network_wifi.h
@@ -75,7 +75,7 @@ protected:
 
     virtual void on_start_listening() override
     {
-        WLAN_CAN_SHUTDOWN = 0;
+        notify_cannot_shutdown();
         /* If WiFi module is connected, disconnect it */
         network_disconnect(0, 0, NULL);
 


technobly:
Nice and clean commits, good job! :)

Regarding the PID, please add an action item to discuss creation of signed windows drivers for serial comms.

XuGuohui:
Do you mean that open an issue to discuss creation of signed windows drivers for serial comms? To be honest, I 'm not familiar with signing a windows driver.

technobly:
@XuGuohui Hi :)  On the Photon, P1, Electron and Core, we have Windows drivers for the CDC serial communication that needs to be installed when doing things like Serial.print() over USB on Windows.  This is not required for Mac.  By changing the PID, you can set up a unique product name that is associated with a fairly generic CDC serial driver.  We have these files here: https://github.com/spark/firmware/tree/develop/misc/driver/windows

They need to be signed though to install without windows complaining: https://github.com/spark/firmware/wiki/Creating-a-signed-USB-driver

We've got the certificates for this.. and by action item I just mean to bring it up with someone on the Particle Team.  I think you might be in communication with David?

You might also need to add/update the `usbd_desc_device.h` in your hal with the Product String:
https://github.com/XuGuohui/firmware/blob/develop/hal/src/electron/usbd_desc_device.h

I might be getting ahead of myself though... do you have a USB connection on the Duo and will you be supporting serial communication over it?

XuGuohui:
Hi @technobly , Thank you for your kindly help. I am in communication with Matthew on particle community. Will he be able to help us sign the windows driver? 
The Duo also uses the high speed USB interface as Photon and it also supports serial communication over it, since I can read the device id and add WiFi credential using serial terminal.

technobly:
Absolutely!  Yes, Mat can definitely help with that.  If you want to create your .INF and test it (you can do this by disabling signed driver enforcement on Windows), make sure everything looks and works ok... then ask Mat to sign the driver :)

XuGuohui:
Perfect! Thanks. I'll ask Mat for help about signing the driver on the particle forum.

m-mcgowan:
I've reviewed all the changes to common files in detail and browsed the new files for the duo. Overall it looks great and I have no problem merging this.  There are some places where we have code duplication, which would be great to avoid so that fixes to the photon automatically apply to the duo.

Here's some suggested changes:

- ideally `bootloader/duo/stdperiphdriver` would be factored to a shared location to be used by the photon and duo.
- `bootloader/src/stm32f2xx/usbd_desc.c` - could the platform specific parts be factored out into the `duo` directory so that each platform has it's own copy of the parts that are platform specific.
- the HAL implementation has duplicated files from the photon - it would be good to factor these out to a WICED subfolder so they can be shared rather than duplicated.
-

XuGuohui:
That sounds great! I was always thinking about merging the duplicate files for Photon and Duo, but not dare to change the file structure. It's good for us to maintain  the source code for Photon and Duo, or the similar platforms arriving in the future. I totally agree with your suggestions! :+1:

PR #681, opened 2015-10-10
serialEventX() 

m-mcgowan:
Emulation of `serialEvent()` from arduino wiring.  Implemented as weak functions in user app to allow the application to know if there is serial data available.

This is a requested feature. I'm a bit puzzled why we might need this since the user loop can easily add code like

```
void loop()
{
   // ... loopy stuff here
   if (Serial1.available())
          serialEvent1();
}

void serialEvent1()
{
    char c = Serial1.read();  
}
```

Which is more efficient than what we are able to provide from the system code.

### DIFF
--- a/modules/shared/stm32f2xx/inc/user_part_export.c
+++ b/modules/shared/stm32f2xx/inc/user_part_export.c
@@ -71,6 +71,7 @@ void module_user_setup() {
 
 void module_user_loop() {
     loop();
+    serialEventRun();
 }
 
 #include "user_dynalib.h"
--- a/system/inc/system_user.h
+++ b/system/inc/system_user.h
@@ -26,6 +26,7 @@ extern "C" {
 void setup();
 void loop();
 
+void serialEventRun();
 
 void system_initialize_user_backup_ram();
 
--- a/system/src/main.cpp
+++ b/system/src/main.cpp
@@ -205,7 +205,6 @@ void manage_safe_mode()
     }
 }
 
-
 void app_loop(bool threaded)
 {
     DECLARE_SYS_HEALTH(ENTERED_WLAN_Loop);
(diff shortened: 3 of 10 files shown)


technobly:
Since Serial data is received with an interrupt handler, loop() may be blocking the if(Serial.available()) call which will delay the realtime response.  If the serialEvent1() handler is called when data is available, that would be preferred, but it seems like that's not the way it's implemented here and it only runs after loop() ends.  This actually makes it less responsive than more responsive, or just a more complicated way to know *when* your serial data is going to get processed.

Arduino's examples seem to indicate this works based on calling it directly from loop().  At least their comments seems to suggest a different operation than their example shows: 
https://www.arduino.cc/en/Tutorial/SerialEvent

But looking at the SerialEvent example bundled with Arduino IDE, it seems more correct and inline with the changes you've made Mat.  Although, I agree... this doesn't seem very necessary or beneficial to use as-is.

This seems to come from Processing which does more asynchronous work for you, also by Tom Igoe.
https://processing.org/reference/libraries/serial/serialEvent_.html

m-mcgowan:
I looked at the arduino sources - the serialEvent() call is invoked after calling loop() `main.cpp` - http://arduino.stackexchange.com/questions/4110/how-does-serialevent-gets-into-main. I'd prefer this over trying to have code run directly from the ISR, which would lead to problems when user code blocks or tries to output to serial based on the data received. 

The main point to including this is that it means we have arduino compatibility for `serialEvent()` so porting code is straightforward.

technobly:
That's a good main point then :)

m-mcgowan:
Testing:

- implement a serial echo server:
```
void serialEvent()
{
      char c = Serial.read();
      Serial.print(c);
} 
```

Tested on the Photon and the Core.

PR #684, opened 2015-10-12
standardized print format

kennethlimcp:
v,s,i are all doing a one line print follow by a linefeed but not sure why `m` needs to be special

### DIFF
--- a/system/src/system_setup.cpp
+++ b/system/src/system_setup.cpp
@@ -89,7 +89,7 @@ template<typename Config> void SystemSetupConsole<Config>::handle(char c)
     }
     else if ('m' == c)
     {
-        print("Your device MAC address is\r\n");
+        print("Your device MAC address is ");
         WLanConfig ip_config;
         ip_config.size = sizeof(ip_config);
         network.fetch_ipconfig(&ip_config);


m-mcgowan:
While I agree consistency is a good thing, this change makes me worry about breaking compatibility with any existing tools (particle or otherwise) that are parsing the output already. @emilyrose - does the CLI attempt to parse the MAC address output from Serial?

m-mcgowan:
Hey @brycekahle - does the CLI parse the MAC address output? Currently it's on one line, but this PR breaks it into two. Do you see that causing any issues?

brycekahle:
It does parse it. I would need to make it more flexible to handle two lines.

brycekahle:
Looking at the code it should handle it OK. Doesn't this PR change it to one line from two?

m-mcgowan:
Yes, it's from two lines to one (I wrote the converse.)

technobly:
Closing due to this needing to be be a synchronized change with the CLI, and also needs to remain backwards compatible with older versions of the CLI.  This is a low enough priority that I have not added a Clubhouse issue for this.  If a PR is submitting for the CLI addressing the above synchronous changes, this PR can be reopened. Thanks!

PR #693, opened 2015-10-14
reduce socketSetBlocking timout and let URCs recv

technobly:


technobly:
This PR represents the soft release of Electron v0.0.2-rc.2, which reduces background blocking delays and improves loop() iteration speed.

Issue #680, opened 2015-10-08
WICED HTTP Server: Incorrect TCP Packet Handling


msolters:
It has been shown that the problem with SoftAP failure (most notably on Safari, some other browsers in iOS) is due to the fact that some some HTTP requests from the browser will naturally be fragmented into multiple packets.

**The current implementation of the WICED HTTP server only considers the first packet**.  This may work most of the time, but in the case of fragmentation, it will return `{r: 0}` to the first packet (spuriously representing "success" to the client), and then ignore the second or subsequent packet(s) *which actually contains the SSID & PW payload.*

This is a problem with the WICED HTTP server.  HTTP is a stream implementation of TCP.  HTTP does not care about packets.  It cares about the bytes in the stream.  The fact the WICED server does not actually accumulate packets and then defer processing the finished stream data means, quite simply, it is a very broken HTTP server.

Community user indraastra reports in more detail:
http://community.particle.io/t/softap-configure-ap-requests-thwarted-by-wiced-http-server/16331

And in addition it is discussed in the end of the great community saga, "Photon SoftAP Usage":
http://community.particle.io/t/photon-softap-usage/11035/178

msolters:
Here is a reference for determining an answer to the question:  How do I know I've reached the end of the HTTP request?
http://greenbytes.de/tech/webdav/rfc7230.html#message.body.length

indraastra:
(copying over from the community thread)

I just checked the HTTP server daemon in WICED SDK 3.3.1 and it seems to suffer from the same problem of not really handling HTTP as a stream. Despite it being a blocker for us, I'm not sure it'd be a worthy investment on your part to fix their broken HTTP server. In the interim, supporting a GET version of the configure-ap and connect-ap HTTP calls is a really great idea.

emilyrose:
@msolters, great sleuthing work! I believe I've encountered this bug myself. Thank you!

indraastra:
Python 3 script to trigger this issue:

https://gist.github.com/indraastra/6ba088ef4f7d5b1fb5af

m-mcgowan:
Thanks so much for the script, that was very useful. 

Rather than rewriting the HTTP server, I've added the minimum necessary changes to the HTTP callback interface so that we can retrieve the additional packets as needed.  It's in the PR above - those that can please try and let me know how it functions both for Chrome and Safari. Thanks!

m-mcgowan:
[...]

PR #694, opened 2015-10-14
fix for WICED HTTP server only reading one packet

m-mcgowan:
fix for issue #680.  When content length is specified, ensure subsequent packets are read as necessary.

Known issues:
- this probably doesn't work with chunked transfer across multiple packets. I'm hoping that this doesn't arise in practice. 

To apply this change:

- update to 0.4.6 if not already done
- cd modules/photon/system-part2
- make PLATFORM_ID=6 all program-dfu


m-mcgowan:
Tested by using a python HTTP softAP client. Also tested using a python script that sents 2 packets. Old code didn't return an error, while the new code does (decryption error since the pwd isn't encrypted.) Community testing underway.

indraastra:
Thanks, this seems to be working! On the client side, setting the [content-type appropriately](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Simple_requests) is required to prevent an OPTIONS request (that the server never replies to) preceding the POST, but once this is done, the server handles the POST like a champ!

m-mcgowan:
Great to hear. For anyone else following a long, I posted the system-part2.bin image in the github issue https://github.com/spark/firmware/issues/680

hal/src/photon/softap.cpp
```diff
@@ -901,16 +902,121 @@ static void tcp_stream_reader(Reader& r, wiced_tcp_stream_t* stream) {
...
+        if (result)
+            return -1;
+
+        body->user = packet;         // ensure we clean up the packet
+        uint16_t packet_length;
+        uint16_t available_data_length;
+        uint32_t offset =0;
+        uint8_t* data;
+
+        if (length>available_data_length)
```
pingdynasty: @m-mcgowan comparing uninitialised variable `available_data_length`

m-mcgowan: Thanks!

pingdynasty:
any chance of posting the `http_server.c` patch, rather than just the binary?

PR #695, opened 2015-10-15
Feature/update zero 046

m-mcgowan:
- b:1 flag for minimal part2
- workaround dct corruption - erase non current sector
- refactored programming targets (program-dfu et al.) so that they can be used with update-zero target
- removed non-critical parts so update-zero image continues to fit in available space


m-mcgowan:
Testing

- build and flash to the device from mono firmware
 - cd modules && make update-zero PLATFORM=photon program-dfu
- verify tinker functions display in CLI
- send 's' to serial, verify that system part 2 has 'b':'1' set indicating minimal version and that system part1 is descriptor is complete.
- OTA flash a full system-part2 to the device and verify that this is accepted and that the modular info no longer includes 'b':'1'

These tests should be performed also on the P1.

technobly:
@m-mcgowan can you comment on if any changes in this PR are necessary to be merged into develop still?

technobly:
Added to Clubhouse 1666, to determine what parts of this PR still might need to be merged into develop.

PR #696, opened 2015-10-15
ota update control

m-mcgowan:
Provides system event notification and system flags to control when an OTA update is started and when the system can reset to apply the update.


m-mcgowan:
Tested 
- the happy path (OTA update)
- OTA update disabled in the app - OTA updates not received by the device. CLI reports flash failed.
- reset disabled in the app and then subsequently enabled (causing the device to reset)

PR #699, opened 2015-10-17
multithreading feedback

m-mcgowan:
Fixes based on the feedback so far received regarding multithreading.


indraastra:
Thanks for all your efforts to make this work as seamlessly with app code as possible!

It looks like listening mode no longer blocks user code, which is awesome, but it still seems to take some issue with hasCredentials(). Here's some minimal code that surfaces the issue:

https://gist.github.com/indraastra/955f40ea35b0348e9032

Steps to reproduce:
1. Run code
2. Open console and see 1's (or 0's, I suppose) being printed 
3. Enter listening mode by holding mode button
4. Output halts
Step to work around:
5. Comment line 50 and uncomment line 51 and rerun
6. Enter listening mode
7. Output continues

Sorry to keep throwing issues your way!

indraastra:
Oops, looks like this just got merged. I was building directly off of this branch, but I'll test again on develop just to be sure.

m-mcgowan:
WiFi.hasCredentials() is a blocking function call and waits for the system to be available so that all interactions with the credentials happens on one thread. 

When in listening mode, as it is now, all system functions will be delayed until listening mode is exited, since the present implementation of listening mode keeps the system thread in a tight loop, so it's not able to service the queue. This isn't ideal. 

I'll add it to our backlog to recode how listening mode works so that the system thread can still process messages.

Thanks for trying this and for the good feedback!

indraastra:
That makes sense, thanks for the explanation. Luckily, I think hasCredentials() is probably the least important function of the lot for providing user feedback, so I should be able to get by without it.

PR #700, opened 2015-10-17
Speed up I2C 100ms timeout on Slave_Address NACK - fixes #675

### LINKED ISSUE #675: Speed up I2C 100ms timeout on Slave_Address NACK with a flag
See discussion and proposed solution here: https://community.particle.io/t/issue-with-i2c-on-photon/12161/45

technobly:
* 100ms timeout is now immediately aborted if ACK failure is detected, i.e., no response to Slave Address being sent.  Master sees NACK on 9th SCL clock.
* NOTE: If Slave is disconnected, Master must still have pull up resistors on SDA and SCL to properly detect an ACK failure.  Without these, the I2C interface will see many Bus Errors (BERR) and potentially ARLO failure.
* Improved comments for I2C Receive routine (HAL_I2C_Request_Data())
* Added missing re-enable commands for interrupts in SW RESET function
* mode and ackFailure elements added to end of STM32_I2C_Info structure to preserve organization in memory for DYNALIB, allows older I2C apps to work with new I2C library.

### DIFF
--- a/hal/src/stm32f2xx/i2c_hal.c
+++ b/hal/src/stm32f2xx/i2c_hal.c
@@ -335,6 +335,16 @@ uint32_t HAL_I2C_Request_Data(HAL_I2C_Interface i2c, uint8_t address, uint8_t qu
             /* Send STOP Condition */
             I2C_GenerateSTOP(i2cMap[i2c]->I2C_Peripheral, ENABLE);
 
+            /* Wait to make sure that STOP control bit has been cleared */
+            _millis = HAL_Timer_Get_Milli_Seconds();
+            while(i2cMap[i2c]->I2C_Peripheral->CR1 & I2C_CR1_STOP)
+            {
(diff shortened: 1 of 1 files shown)


technobly:
Fixes: https://github.com/spark/firmware/issues/675

m-mcgowan:
Regarding dynalib and the STM32_I2C_Info struct - that struct isn't passed out to other modules (the definition is in a C file so no sharing of the struct event). Thus it's only used internally within the I2C HAL meaning versioning isn't a concern here - feel free to place the new values where they make sense.

technobly:
Ah ok cool, probably a good habit to get into anyway as structs rarely need to be organized :)

m-mcgowan:
:+1:

technobly:
Once I know this is working well for everyone, I can make the same changes to the Core.

PR #708, opened 2015-10-22
Support platform Duo

XuGuohui:
Hi,

This pull request includes changes on common used files to support Duo, a new platform from RedBear. Please review and comment it!

Thanks,
Guohui


hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -240,6 +240,11 @@ void HAL_Core_Config(void)
 #ifdef HAS_SERIAL_FLASH
     //Initialize Serial Flash
     sFLASH_Init();
+#if PLATFORM_ID == PLATFORM_DUO_PRODUCTION
```
m-mcgowan: I suggest factoring out the target flash device to it's own #define in one place then referencing that each time, rather than repeating the compile time conditional each time it's used.

platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/spi_flash.c
```diff
@@ -355,6 +358,37 @@ static void sFLASH_WriteEnable(void)
 
     /* Deselect the FLASH: Chip Select high */
     sFLASH_CS_HIGH();
+
```
m-mcgowan: Do these changes need to be compile-time conditional for the Duo?

m-mcgowan:
Apart from the missing conditional compilation around the changes to the sFLASH_WriteEnable function, these changes look very good to me.

hal/src/stm32f2xx/core_hal_stm32f2xx.c
XuGuohui: Please pardon me for having no idea what exactly to do as you told. Could you tell me more details?

platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/spi_flash.c
XuGuohui: Done. Is there any other changes to be compile-time conditional for Duo?

XuGuohui:
@m-mcgowan , Regarding to the commit 0a260cbc6639b2b841eb85ad2419e480a2f0be79, I'm wondering why you add 1 second delay before setting the linecoding handler? To avoid unnecessary interrupt when device being plugged onto PC?

Thanks.

m-mcgowan:
Sorry for the delay in response - I don't know why we delay there. I'm guessing it's to allow the usb connection time to stabilize, but it does feel pretty arbitrary!

technobly:
This PR currently serves to highlight refactoring that needs to occur in system firmware so that Compounds only need to add the required files for integration support.

This system refactoring for compounds has been added to Clubhouse 1669.

I will close this PR for now, changes should still be observable and here is an additional comparison link:
https://github.com/spark/firmware/compare/develop...redbear:develop

PR #716, opened 2015-10-26
Add timerChangePeriod()

pkourany:
Add timerChangePeriod() function to change timer period


pkourany:
PR gone weird. #716, 717 and 718 should be a single PR.

PR #717, opened 2015-10-26
Add timerChangePeriod()

pkourany:
Add 	OS_TIMER_CHANGE_PERIOD to os_timer_change_t


pkourany:
PR gone weird. #716, 717 and 718 should be a single PR.

PR #720, opened 2015-10-28
Add timerChangePeriod()

pkourany:
Add software timer timerChangePeriod() function.  Also, order of fromISR parameter in os_timer_change() was incorrect.


m-mcgowan:
Thanks for this Paul! Do you want to have a go at writing some unit tests for these changes - the time period change feature in particular.

pkourany:
Matt, I have (oddly) never written one and not sure how to go about it!

m-mcgowan:
Changes look good. I'll add tests later since we're in a pinch. Paul, will you add docs for this please?

PR #721, opened 2015-10-29
Fixes Soft Power Down, Listening Mode operation and CFOD issue on Electron

technobly:
### Soft Power Down
  - double-tap SETUP button to activate
  - Dim white RGB LED while shutting down
  - Deactivates PDP context
  - Detaches GPRS connection
  - Puts modem to sleep
  - Puts STM32 to sleep
  - Press RESET button to wake up (or rising HIGH signal on WKP pin)

### Listening Mode
  - Hold SETUP button for >3 seconds to enter
  - Hold SETUP button for >3 && <8 seconds && let go of button to exit
  - alternatively press `x` in serial terminal to exit
  - Previously setup PDP context will resume if available on exit

### Also fixes some issues with CFOD
  - Sockets were being used up when system would go through soft power down or when STM32 went to sleep.  This was because the modem might still be powered and active, yet system code was resetting so all knowledge of which sockets were in use was reset as well.  After a couple power downs, all sockets were taken and it appeared no more were available.  Yet, all 6 of them were technically available.  Now when the first socket is created on system boot, a new routine is run that seeks out open sockets on the modem and frees them.

### Tinker App for Electron
  - Removed unnecessary 3 second delay in setup()
  - Changed USB debugging baud from 115200 to 9600 for ease of use with most terminal apps (they default to 9600).

### DIFF
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -23,6 +23,7 @@
 #include <stdint.h>
 #include <string.h>
 #include "net_hal.h"
+#include "wlan_hal.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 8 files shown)


technobly:
#### Two known issues that don't need to be solved with this PR:

  :cyclone: Soft power down does not shut off the modem when Multithreading is enabled.
  :lollipop:  The SETUP button doesn't seem to respond to an initial press and hold until some time after boot, >1 or 2 seconds.

Issue #671, opened 2015-10-06
analogWrite to DAC pins on Photon no longer works after upgrading to 0.4.6


reinerge:
After upgrading to 0.4.6 analogWrite on photon no longer outputs the given DA value but instead outputs 3.3V
Repro steps: 
1.   run code below on a photon device with 0.4.5 firmware and watch DAC output pin using a scope. It should show a sawtooth waveform signal.
2.  cloud compile code for 0.4.6 and flash device. Device update to 0.4.6 starts. 
3.  Wait for update to 0.4.6 firmware to finish. Scope shows 3.3V constant signal. If there's still a sawtooth visible, re-power device.

    void setup() {

    }

    void loop() {
        for(int i = 0; i < 4096; i++) {
            pinMode(DAC, OUTPUT);
            analogWrite(DAC, i);
        }
        delay(100);
    }

reinerge:
**Update:**
I tried to find out what could be the reason for that issue and in order to do so I started to add some Serial.print's inside the top level function analogWrite in the 0.4.6 firmware code just to see what is called and what is not (I switched to local build for that, syncing to the 0.4.6 repository on my machine).

**So here's the surprise:**
adding Serial.println() between the safety check and the call to HAL_Validate_Pin_Function made the analogWrite function work again.
adding a simple delayMicroseconds(0) instead does the same. Adding the delay function at the beginning of the analogWrite function does however not work.
This smells like voodoo...

In addition, with this change it is no longer necessary to call pinMode(DAC, OUTPUT) everytime you call analogWrite.

Here's my modified analogWrite function:

    void analogWrite(pin_t pin, uint16_t value)
    {
        // Safety check
        if (!pinAvailable(pin))
        {
            return;
        }
        delayMicroseconds(0);
        if (HAL_Validate_Pin_Function(pin, PF_DAC) == PF_DAC)
        {
            HAL_DAC_Write(pin, value);
        }
        else if (HAL_Validate_Pin_Function(pin, PF_TIMER) == PF_TIMER)
        {
            PinMode mode = HAL_Get_Pin_Mode(pin);

            if (mode != OUTPUT && mode != AF_OUTPUT_PUSHPULL)
            {
                return;
            }

            HAL_PWM_Write(pin, value);
        }
    }

I would guess that there has been an issue with analogWrite also in the previous releases that was hidden by the call to pinMode before. Now some timing or whatever has changed and that "trick" does not work anymore -- means the issue is somewhere in the HAL code.
[...]

Issue #662, opened 2015-10-03
Photon: analogWrite() for DAC requires pinMode() each time


ScruffR:
As in this docs issue spark/docs#170

I think this is an undesirable behaviour, that each call to analogWrite() for a DAC pin compulsory requires a preceding call to pinMode(DAC, OUTPUT).
It should rather be like with PWM pins, that one single pinMode(DAC, OUTPUT) is enough.

See test
https://community.particle.io/t/analogwrite-dac-only-once-for-each-pinmode-output/15393/2

---

*Sorry, I originally opened this issue in private firmware*

ScruffR:
#671 makes this one look a minor glitch :-(

m-mcgowan:
Verified working.

PR #727, opened 2015-11-03
Zero out unused fields in DAC_InitStructure - seems to fix the DAC

unreason:
Zeroing out the only unused field in DAC_InitStructure - DAC_InitTypeDef::DAC_LFSRUnmask_TriangleAmplitude - appears to completely fix DAC behavior for me (issues #671 and #662). Setting the unused field to 255 brings back erratic behavior.

Can someone else please try this fix and report if it works for them?


m-mcgowan:
Many thanks for hunting down this bug! We incorporated the fix in https://github.com/spark/firmware/pull/728

PR #730, opened 2015-11-05
add test for github issue 671

peterqu:
Add the failure case for issue 671.  The original failure is somewhat
random depends on the state of the memory stack.  The issue is no
longer reproducible when compiled against the latest development branch.

### DIFF
--- a/user/tests/wiring/adc_dac/adc_dac.cpp
+++ b/user/tests/wiring/adc_dac/adc_dac.cpp
@@ -60,6 +60,7 @@ test(ADC_AnalogReadOnPinWithVoltageDividerResultsInCorrectValue) {
 }
 #endif
 
+
 #if (PLATFORM_ID == 6)
 test(ADC_AnalogReadOnPinWithDACOutputResultsInCorrectValue) {
     pin_t pin = A5;//pin under test (Voltage divider with equal resistor values)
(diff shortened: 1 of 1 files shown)


user/tests/wiring/adc_dac/adc_dac.cpp
```diff
@@ -71,3 +72,20 @@ test(ADC_AnalogReadOnPinWithDACOutputResultsInCorrectValue) {
     //To Do : Add test for remaining pins if required
 }
 #endif
+
+#if (PLATFORM_ID == 6)
```
technobly: This should be `#if PLATFORM_ID >= 3` to make sure this runs properly on P1 and Electron as well.  Both PLATFORM_ID==6 in this file could be changed to this.  Also what does the for() loop test since the pin is set back to `analogWrite(DAC, 2048);` afterwards?  Also looking at the setup, it seems A5 is biased to Vdd/2, which is what the DAC is set to as well.  I would think we would want to bias the pin to maybe 1/3Vdd and then set the DAC to 2/3Vdd for a proper test.  As-is the DAC could be not outputting anything at HI-Z and the test would pass.  100k ohm resistors might be too high as well for a proper ADC reading.  Sorry for all the grief I know you did not write the initial test, just documenting some comments about it now that I'm actually looking at it for the first time ;)

peterqu: The for loop is to reproduce the failing sequence of operations in issue #671.   In that failure, the DAC would be stuck at a random value between 0 and Vdd, so would fail the test.

The resistor divider setup only applies to Core, which does not have a DAC.  In that case, we do a resistor divider to bias the ADC to 1/2VDD.  100K might be too much in comparison to the input impedance of the ADC, changing it to 5 or 10k should be better.  However, since the test check for a ADC in the range 2000 to 2100, the test should still pass even with the loading of input impedance on A5. 

For the Photon, there is no resistor divider.  The output buffer of DAC drives the ADC directly.

PR #734, opened 2015-11-11
Added Arduino.h - compatability shim for Particle applications

plocher:
For discussion and debate - by adding this header to the Particle env, simple/portable apps written for the Wiring/Arduino IDE will tend to compile and run the first time - and, more importantly, work correctly.

This probably isn't enough - there should be a (maybe empty) "wiring.h" and a few others, such that compiles don't break unnecessarily.

There is still a need for architecture specific defines (for ADC sizes, pins/LEDs...), but this shim provides a way to ease the transition for many applications/sketches and libraries.


m-mcgowan:
Please take a look at the spark_wiring_arduino.h file - I think many of the defines are already there.

plocher:
While there are some defines there, they are superficial and don't address
the delta between the env seen in the Arduino IDE and in the Particle IDE...
Not to mention that the file is NOT included by application.h

IMO, even if the file *had* the right contents, the name difference
effectively hides the file and its contents from 100% of the app writers
out there.
The "key" to this pull request is the name, so that sketches that already
"#include <Arduino.h>" just work.

  -John

technobly:
Closing due to work being done here https://github.com/spark/firmware/commits/feature/arduino_compatibility superseding this PR.

PR #735, opened 2015-11-11
Cellular.RSSI() API for Electron

technobly:
### SYNTAX

```c++
CellularSignal sig = Cellular.RSSI();
Serial.println(sig.rssi);
Serial.println(sig.ber);
Serial.println(sig); // Complete structure also printable
```

### Output
```
-95    // RSSI
19     // BER
-95,19 // RSSI,BER
```

### EXAMPLE

```c++
#include "application.h"

// ALL_LEVEL, TRACE_LEVEL, DEBUG_LEVEL, WARN_LEVEL, ERROR_LEVEL, PANIC_LEVEL, NO_LOG_LEVEL
SerialDebugOutput debugOutput(9600, ALL_LEVEL);

uint32_t lastUpdate = 0;
#define now millis()

SYSTEM_MODE(SEMI_AUTOMATIC);

void setup()
{
  Particle.connect(); // blocking call to connect
  while (!Particle.connected()) Particle.process();
  lastUpdate = now;
}

void loop()
{
  // Publish signal strength and bit rate error every minute
  if (now - lastUpdate > 60000UL) {
    lastUpdate = now;
    CellularSignal sig = Cellular.RSSI();
    String s = String(sig.rssi) + String(",") + String(sig.ber);
    Particle.publish("electron-signal", s, 60, PRIVATE);
  }
}
```


user/tests/wiring/api/system.cpp
```diff
@@ -81,18 +81,21 @@ test(system_freememory) {
 }
 
 test(system_waitfor) {
+#if PLATFORM_ID!=10
```
m-mcgowan: Rather than hard-code on platform IDs, you can include hal_platform.h and test against HAL_PLATFORM_WIFI, or, since this is at the Wiring level, we also have Wiring_WiFi conditional define from the wiring API. This will future-proof the code for any other non-wifi devices.

technobly: Good call, I'll use Wiring_WiFi and Wiring_Cellular where appropriate ;-)

user/tests/wiring/api/system.cpp
```diff
@@ -81,18 +81,21 @@ test(system_freememory) {
...
     API_COMPILE(waitFor(WiFi.ready, 10000));
     API_COMPILE(waitUntil(WiFi.ready));
+#endif
 }
 
 
 test(system_config_set) {
 
     API_COMPILE(System.set(SYSTEM_CONFIG_DEVICE_KEY, NULL, 123));
+#if PLATFORM_ID!=10
```
technobly: This one really should be: `#if PLATFORM_ID >= 6 && PLATFORM_ID <= 8` or better yet `#if PLATFORM == photon`

PR #740, opened 2015-11-17
Ensure socket exists before removing it when closing in socket_hal

kierangraham:
We've have had an ongoing issue where once our Photon reconnects to WiFi it immediately hangs whilst breathing cyan.

Inside our application we make use of a UDP socket which once the WiFi connection comes up seems to be a stale resource no longer within `this->items` and once execution enters the following `while` block, hangs forever.

Our Lead Firmware Engineer (@mhazley) has [documented this on the forum](https://community.particle.io/t/photon-v0-4-7-hanging-in-udp-stop/17553) and managed to catch this in the debugger and it's looking very likely that the root of our issue is from recycling the socket every 30s (internal detail of our application) _after_ a WiFi disconnect / reconnect has happened.

**Work Done**

* [x] Check that a `socket_t` exists before attempting to remove it, otherwise `return false`

### DIFF
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -511,13 +511,14 @@ public:
     bool remove(socket_t* item)
     {
         bool removed = false;
-            if (items==item) {
-                items = item->next;
+        if (items==item) {
+            items = item->next;
             removed = true;
         }
-        else
+        else if (exists(item))
         {
-                socket_t* current = items;
+            socket_t* current = items;
+
             while (current) {
                 if (current->next==item) {
                     current->next = item->next;
@@ -526,6 +527,7 @@ public:
                 }
             }
         }
+
         return removed;
     }
 
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -508,23 +508,23 @@ public:
      * @param item
      * @param list
      */
-    bool remove(socket_t* item)
-    {
+     bool remove(socket_t* item)
+     {
         bool removed = false;
         if (items==item) {
             items = item->next;
             removed = true;
         }
-        else if (exists(item))
+        else
         {
             socket_t* current = items;
-
             while (current) {
                 if (current->next==item) {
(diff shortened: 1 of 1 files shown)


mhazley:
Perhaps re-writing the remove function to do the following would have same effect? 

```
bool remove(socket_t* item)
{
   bool removed = false;
   if (items==item) {
       items = item->next;
       removed = true;
   }
   else
   {
       socket_t* current = items;
       while (current) {
           if (current->next==item) {
               current->next = item->next;
               removed = true;
               break;
           }
           current = current->next;
       }
   }
   return removed;
}
```
This way we'll fall out with `removed = false` if the socket does not exist.

kierangraham:
@mhazley thanks for that, updated the PR to use that change instead.

mhazley:
Raised an issue to go along with this at #742

m-mcgowan:
Thanks for this. Just from visual inspection, I can see the current pointer isn't updated - I was sure I'd fixed this before, so a bit puzzled about this regression. I will add a test case that creates 2 sockets then removes them. This should highlight the problem and demonstrate the fix.  I will do this as a separate PR - no point having you sign our CLA for a one line change!

kierangraham:
@m-mcgowan I already signed the CLA before submitting this PR. If you could give some pointers on what way you'd like the test case structured, we could look at adding that to this PR also.

m-mcgowan:
oh, that's great, thanks for stepping up! The the tests would follow the same lines as our existing unit tests in `user/tests/wiring/no_fixture`.  Please see the docs folder in the repo for how to run the tests.

m-mcgowan:
We are making a new release today so I've applied the fix manually. Let's keep this PR open so you still have a place to add the unit tests when you have time for those.

mhazley:
Thanks @m-mcgowan - I'm going to try and get a look at them today actually so will add to this PR either myself or via @kierangraham

m-mcgowan:
I'll close this due to inactivity - feel free to submit another PR if you get around to writing unit tests for the change.

PR #744, opened 2015-11-23
Bugfix issue650

peterqu:
Refactor PWM library to init timer only once.  Subsequent analogWrite will only update the output compare register (CCR).  This should eliminate glitches when call to analogWrite is made more frequently than the PWM frequency.


m-mcgowan:
I'm closing this since it looks to be superseded by #752.

PR #750, opened 2015-12-01
will now show LED color long enough during cloud connection failure

technobly:
Before this fix, the LED would flash and be very hard to distinguish between RED, ORANGE or MAGENTA
https://www.dropbox.com/s/9mqtyuhh0wu7xcj/led-before.mp4?dl=0

After this fix, the LED color is much easier to see
https://www.dropbox.com/s/1recgqz9fr90wbx/led-after.mp4?dl=0

### DIFF
--- a/system/src/system_task.cpp
+++ b/system/src/system_task.cpp
@@ -246,29 +246,31 @@ void handle_cloud_connection(bool force_events)
             if (err)
             {
                 cloud_connection_failed();
+                LED_Signaling_Start();
                 if (0 > err)
                 {
                     // Wrong key error, red
-                    LED_SetRGBColor(RGB_COLOR_RED);
+                    LED_SetSignalingColor(RGB_COLOR_RED);
                 }
                 else if (1 == err)
                 {
                     // RSA decryption error, orange
-                    LED_SetRGBColor(RGB_COLOR_ORANGE);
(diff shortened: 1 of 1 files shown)


m-mcgowan:
The change from 250ms to 500ms makes sense.  Why do we need to use LED signalling? Does that have an effect?

technobly:
Without overriding the system signaling, the LED blinks super fast and is hard to see.  The other way to solve this is with a new LED state and stop the blinking in the HAL_SysTick_Handler.  Perhaps that is the proper way to do it and this PR can easily progress in that direction.  Using the override is most likely going to collide with user implementation of RGB.control().  This is something that's bothered me for a while and I thought I'd give it a tiny bit of attention ;-)

m-mcgowan:
Ah,yes, I'd forgotten about the blinking in the SysTick handler! good catch!

The LED code will be reworked with a series of importance levels and sources, so that we can be sure that important LED notifications from the system are shown while also allowing the application to override the background system LED states.

technobly:
If you want to close this out as a proof of concept received, totally fine by me.  I knew about the LED refactor and wanted to share thoughts on this part clearly ;-)

technobly:
Closing.  Issue added to Clubhouse 1655.

PR #759, opened 2015-12-08
Implement WiFi.getCredentials

### LINKED ISSUE #453: Function to list the SSIDs
Can we create a function on the Photon to return the list of SSIDs which the Photon has available ?

An example would be :

Wifi.getCredentials() or Wifi.getSsids()

returns an array of strings with the SSIDs in it.

This was discussed on the forum :
https://community.particle.io/t/manage-wifi-credentials/9982

Thank-you

monkbroc:
* User provides an array of WiFiAccessPoint that get filled with the APs saved on the device
* Returns number of APs on device
* Photon only. Returns 0 on the Core since it cannot read APs back from the CC3000 :disappointed:

Changes
* Implement wlan_get_credentials in HAL
* Implement WiFi.getCredentials with array arguments
* Add API test

Fixes #453

Usage:
```
    WiFiAccessPoint ap[5];
    int found = WiFi.getCredentials(ap, 5);
    for (int i=0; i<found; i++) {
        Serial.print("ssid: ");
        Serial.println(ap[i].ssid);
        Serial.println(ap[i].security);
        Serial.println(ap[i].channel);
        Serial.println(ap[i].rssi);
    }
```

For discussion: my first cut of this was to return a `vector<WiFiAccessPoint>` (typedefed to `WiFiAccessPointList`) from `WiFi.getCredentials`. This made for very nice application code:
```
WiFiAccessPointList list = WiFi.getCredentials();
for(WiFiAccessPoint ap : list) {
  Serial.println(ap.ssid);
}
```
But I noticed that `WiFi.scan` is currently implemented to accept an array allocated on the stack by the caller, so I went with that solution in the end.

Any strong reasons why not create a `vector` that allocates a few `WiFiAccessPoint` on the heap whenever `WiFi.scan` or `WiFi.getCredentials` is called? It is simple to make a version of `WiFi.scan` that returns a `WiFiAccessPointList` too.

Another point of discussion:
Printing the list of available networks on one of my Photon gives me this:
```
Listing stored access points
Count=5
SSID=TechBreweryNew, security=3
SSID=TechBreweryNew, security=3
SSID=TechBreweryNew, security=3
SSID=TechBreweryNew, security=3
SSID=TechBreweryNew, security=3
End listing
```

Is it OK that there are 5 copies of the same network? Now that people will be able to read this list out it might cause confusion.

Would it be better to overwrite existing credentials if the SSID is the same rather than make multiple copies?

### DIFF
--- a/hal/inc/hal_dynalib_wlan.h
+++ b/hal/inc/hal_dynalib_wlan.h
@@ -67,6 +67,7 @@ DYNALIB_FN(hal_wlan,wlan_select_antenna)
 DYNALIB_FN(hal_wlan,wlan_set_ipaddress)
 DYNALIB_FN(hal_wlan,wlan_set_ipaddress_source)
 DYNALIB_FN(hal_wlan,wlan_scan)
+DYNALIB_FN(hal_wlan,wlan_get_credentials)
 DYNALIB_END(hal_wlan)
 
 #endif	/* HAL_DYNALIB_WLAN_H */
--- a/hal/inc/wlan_hal.h
+++ b/hal/inc/wlan_hal.h
@@ -261,6 +261,15 @@ typedef void (*wlan_scan_result_t)(WiFiAccessPoint* ap, void* cookie);
  */
 int wlan_scan(wlan_scan_result_t callback, void* cookie);
 
+/**
+ * Lists all WLAN credentials currently stored on the device
+ * @param callback  The callback that receives each stored AP
+ * @param callback_data An opaque handle that is passed to the callback.
+ * @return count of stored credentials, negative on error.
+ */
+
+int wlan_get_credentials(wlan_scan_result_t callback, void* callback_data);
+
 #ifdef	__cplusplus
 }
 #endif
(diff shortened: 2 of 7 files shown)


m-mcgowan:
How about we call it `WiFi.getCredentials()` to mirror `WiFi.setCredentials()`?

monkbroc:
Sure. You have write access to my fork if you want to make small changes like that before merging.

monkbroc:
I renamed this API to `WiFi.getCredentials()`

jmaxxz:
I don't see how this gets the configured credentials. It appears to get the everything about the configured wifi networks besides the credentials. Am I misreading this? Or is does this api exposes the preshared keys of previously configured networks.

monkbroc:
I think what you're saying is it would be best described as "get connections" rather than "get credentials" since it doesn't actually return the passwords, but the connection information for which passwords are available, but cannot be retrieved. I can update the docs to that effect.

jmaxxz:
@monkbroc yes that is exactly what I am saying. Thanks!

Issue #704, opened 2015-10-20
Photon: Possible issue with "unhandled" "brownout"?


ScruffR:
Recently there seem to pop up topics about Photons becoming unusable or loosing data due to power issues (e.g. battery drainage, cheap DC chargers, ...)

e.g.
 https://community.particle.io/t/photon-ran-low-on-battery-now-stuck-in-sos/16710
 https://community.particle.io/t/eeprom-persistence-issue/16514

Could this be helped via software or does it need hardware?
If the latter a proper warning in the docs for existing devices might be required.

nicodegunst:
I have experienced this also. It is a new and I believe related to 0.4.6 fw. I have worked with both the core and photon with all iterations of the firmware and have tested with the same battery-powered power source all the time, draining the batteries. So far it didn't happen that eeprom lost its data, and now suddenly it does.
I hope this will be resolved soon..

kikolobo:
I have similar issues where a photon running on battery power becomes unusable after the batteries are drained and then re charge.. The photon turns on either the D7 LED and has to be re flashed using JTAG... Or it looses wifi credentials after the reboot.

Don't know what causes but it doesn't happen with direct power ... Only when running on batteries and the batteries run low.

InMyOrbit:
I am also experiencing this issue. I used the board with the SparkFun battery shield and since then I have not been able to use the board continuously. I manage to configure the board using the cli-interface and perform a remote upload. But as soon as the device reboots it looses its wifi credentials and gets stuck. The RGB-led flashes in an infinite pattern like this: 25 seconds of cyan flash, 3 orange flashes.

Tried updating keys without success.

monkbroc:
[I proposed in the community forum to set the brown-out reset configuration registers.](http://community.particle.io/t/eeprom-persistence-issue/16514/39?u=jvanier) If you are having this issue please report back here if setting a higher brown-out reset level helps.

```
void setBrowoutResetLevel() {
  const uint8_t desiredBOR = OB_BOR_LEVEL3; // Supply voltage ranges from 2.70 to 3.60 V
  if(FLASH_OB_GetBOR() != desiredBOR) {
    /* Steps from http://www.st.com/web/en/resource/technical/document/programming_manual/CD00233952.pdf */
[...]

PR #762, opened 2015-12-10
Improve EEPROM stability and persistence

monkbroc:
### Changes
* [Don't erase a page of flash ON EVERY BOOT](https://github.com/monkbroc/firmware/blob/18eda04106ea7f54d1a4a794840aa2403dad28bd/hal/src/stm32f2xx/eeprom_hal.c#L539) to increase Flash life and avoid erasing Flash in every low power situations
* Don't format the EEPROM in case both pages are valid, just erase one
* Lock the Flash after initialization or variable write

### Background

Bad chargers and depleted batteries cause the EEPROM contents to be lost on the Photon and Electron.
http://community.particle.io/t/eeprom-persistence-issue/16514
https://community.particle.io/t/photon-ran-low-on-battery-now-stuck-in-sos/16710
#704

The root cause is still not clear but one factor making it worse is the fact that the STM EEPROM emulation code (AN3390) erases one page of Flash on every boot. This is unnecessary and aweful for Flash life. Also what happens exactly if the microcontroller looses power during this erase operation? Experience shows that the EEPROM contents will be lost.

This is not a complete fix since an EEPROM erase could still happen during a low battery situation, but it will decrease the occurence of these issues.

### Steps to reproduce EEPROM persistence issue

1. Flash this application code

```
#include "application.h"

// Write something to the EEPROM on demand
int eepromWrite(String value) {
  EEPROM.write(0, 0xA5);
  return 0;
}

void setup()
{
  RGB.control(true);
  if(EEPROM.read(0) == 0xFF) {
    // Red means the EEPROM was wiped
    RGB.color(255, 0, 0);
  } else {
    // Yellow means the EEPROM was preserved
    RGB.color(255, 255, 0);
  }
  Particle.function("eepromWrite", eepromWrite);
}
```

2. Call `particle call device eepromWrite`
3. Reboot the Photon and notice the yellow LED meaning EEPROM contents are saved
4. Unplug the USB
5. Replug it partially, fiddle with it, making sure the white LED turn on and off multiple times and even turns partially on
6. Plug USB firmly
7. Notice the red LED meaning the EEPROM contents were lost. Try again if it doesn't fail on the first try

[Video of failure before the fix](https://youtu.be/YSzFfkvTlE8)
[Video of correct behavior after the fix](https://youtu.be/iKj_9vaid5A)

### DIFF
--- a/hal/src/stm32f2xx/eeprom_hal.c
+++ b/hal/src/stm32f2xx/eeprom_hal.c
@@ -86,7 +86,11 @@ static uint16_t EEPROM_WriteVariable(uint16_t EepromAddress, uint16_t EepromData
 
 void HAL_EEPROM_Init(void)
 {
+    /* Unlock the Flash Program Erase controller so EEPROM_Init can
+     * format EEPROM in case it is invalid */
+    FLASH_Unlock();
     EEPROM_Init();
+    FLASH_Lock();
 }
 
 size_t HAL_EEPROM_Length()
@@ -110,7 +114,9 @@ void HAL_EEPROM_Write(uint32_t address, uint8_t data)
 {
     if (address < EEPROM_SIZE)
     {
+        FLASH_Unlock();
         EEPROM_WriteVariable(address, data);
+        FLASH_Lock();
     }
 }
 
@@ -397,9 +403,6 @@ static uint16_t EEPROM_Init(void)
     uint16_t  FlashStatus;
     uint16_t EepromDataVar = 0;
 
-    /* Unlock the Flash Program Erase controller */
-    FLASH_Unlock();
-
     /* Calling this is here is critical on STM32F2 Devices Else Flash Operation Fails */
     FLASH_ClearFlags();
 
@@ -414,13 +417,7 @@ static uint16_t EEPROM_Init(void)
         case ERASED:
             if (PageStatus1 == VALID_PAGE) /* Page0 erased, Page1 valid */
             {
-                /* Erase Page0 */
-                FlashStatus = FLASH_EraseSector(PAGE0_ID,VOLTAGE_RANGE);
-                /* If erase operation was failed, a Flash error code is returned */
-                if (FlashStatus != FLASH_COMPLETE)
-                {
-                    return FlashStatus;
-                }
+                /* Normal valid state -> nothing to do */
             }
             else if (PageStatus1 == RECEIVE_DATA) /* Page0 erased, Page1 receive */
             {
@@ -523,10 +520,10 @@ static uint16_t EEPROM_Init(void)
             break;
 
         case VALID_PAGE:
-            if (PageStatus1 == VALID_PAGE) /* Invalid state -> format eeprom */
+            if (PageStatus1 == VALID_PAGE) /* Harmless invalid state -> pick one page and format the other */
             {
-                /* Erase both Page0 and Page1 and set Page0 as valid page */
-                FlashStatus = EEPROM_Format();
+                /* Erase Page0 */
+                FlashStatus = FLASH_EraseSector(PAGE0_ID, VOLTAGE_RANGE);
                 /* If erase/program operation was failed, a Flash error code is returned */
                 if (FlashStatus != FLASH_COMPLETE)
                 {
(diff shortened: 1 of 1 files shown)


nicodegunst:
Just tested this pull request, it absolutely works! Without it, I can very easily erase/corrupt the flash memory by simply switching usb on and off quickly (once or twice is enough). With this fix in place, I was unable to erase or corrupt the flash memory!

m-mcgowan:
Perfect! Thank you @nicodegunst  for testing and verifying the fix. I'll review the code and apply this or something similar to the system firmware for the 0.4.8 release.

towynlin:
:+1: Outstanding! Thanks so much Julien!

monkbroc:
@m-mcgowan @towynlin You're welcome!

STM's EEPROM emulation code is pretty rancid. This PR is a quick fix for the worst issues, but I wouldn't be surprised if there are other failure modes in there like @andyw-lala believes.

m-mcgowan:
Just to echo Zac's comment above, a big thanks for this! Catching the unnecessary erase on each boot is a big win.  It's been on my mind for a long time to refactor this code to be testable and verify with tests that erases don't happen more often than required, but it's continually been pushed down the list by other priorities.  Great to have an expert pair of eyes review the code. :+1:

andyw-lala:
And it's official - no one is happier than me to see this step taken.

As Julien says, I believe there are likely to be other issues (rancid is an
understatement wrt AN3390), but never let perfection get in the way of
progress.

On Mon, Dec 14, 2015 at 12:05 PM, Matthew McGowan <notifications@github.com>
wrote:

> Just to echo Zac's comment above, a big thanks for this! Catching the
> unnecessary erase on each boot is a big win. It's been on my mind for a
> long time to refactor this code to be testable and verify with tests that
> erases don't happen more often than required, but it's continually been
> pushed down the list by other priorities. Great to have an expert pair of
> eyes review the code. [image: :+1:]
>
> —
> <https://github.com/spark/firmware/pull/762#issuecomment-164512511>.
>



-- 
Andy

m-mcgowan:
"but never let perfection get in the way of progress." What do you mean there @andyw-lala?

andyw-lala:
Meaning that a fix that fixes the biggest problems but may not address
every single less common failure case, is to be preferred vastly over
holding out for the perfect fix to all problems (not that I believe anyone
was pushing for that in this case.)

It wasn't a loaded or sarcastic remark, if that's the way it came across.


On Mon, Dec 14, 2015 at 12:43 PM, Matthew McGowan <notifications@github.com>
wrote:

> "but never let perfection get in the way of progress." What do you mean
> there @andyw-lala <https://github.com/andyw-lala>?
>
> —
> <https://github.com/spark/firmware/pull/762#issuecomment-164522055>.
>



-- 
Andy

monkbroc:
👍

Enjoy your holidays Mat.

m-mcgowan:
Thanks very much for the good wishes and this PR. I'd like to get the brownout reset logic in place, so I might pick your brains about that in the new year! 

I hope you find time to relax over the holidays.  Festive greetings! :-)

peterqu:
Hi @monkbroc 

Thank you for this great fix.  There is one minor thing I want to check with you.

For the case where page0=valid, page1=valid, the new change is to only erase page 0.  I think you have forgot to set page0 status to ERASED.  I think the current code will result in both page0 and page1 have the VALID_PAGE status, but page0 will contains all FF data (erased).

What do you think?

~~~c
// Line 522, page0=VALID, page1=VALID
 if (PageStatus1 == VALID_PAGE) /* Harmless invalid state -> pick one page and format the other */
 {
        /* Erase Page0 */
        FlashStatus = FLASH_EraseSector(PAGE0_ID, VOLTAGE_RANGE);
        /* If erase/program operation was failed, a Flash error code is returned */
        if (FlashStatus != FLASH_COMPLETE)
         {
                    return FlashStatus;
         }
        // I think you need to add this line 
        FlashStatus = FLASH_ProgramHalfWord(PAGE0_BASE_ADDRESS, ERASED);
}
~~~

m-mcgowan:
ERASED is 0xFFFF so writing this has no affect when written to a freshly erased page.

peterqu:
Great, then it's all good.

andyw-lala:
In addition to avoiding gratuitous sector erase operations, Julien's fix
also only unlocks the flash controller during the required operations, and
locks the controller again afterwards. The original code appears to have
unlocked the controller regardless and left it unlocked, which is an
accident waiting to happen.

Since there is discussion about brownout protection, we should also take
steps to ensure the the flash configuration (see section 2.4 and 2.5 of
PM0059 (
http://www.st.com/web/en/resource/technical/document/programming_manual/CD00233952.pdf)
is supported by the voltage settings of the brownout controller.

It certainly looks like bad things can happen just reading flash at low
voltages, and it would be far better to just hold the part in reset rather
than execute unreliably. With Julien's fixes in place, I think the
opportunity for flaky operation due to low voltage will no longer corrupt
data required for reliable operation, but we should at least audit the
settings to ensure devices either operate reliably or not at all.

On Wed, Dec 23, 2015 at 2:35 AM, Peter Qu <notifications@github.com> wrote:

> Great, then it's all good.
>
> —
> <https://github.com/spark/firmware/pull/762#issuecomment-166833649>.
>



-- 
Andy

PR #767, opened 2015-12-14
Electron RSSI UX via SETUP button

technobly:
* One click of the SETUP button invokes a blocking call for signal strength, when received it is converted to 0-5 "bars" of signal strength via a green LED.  First the LED dims to make the flashes more apparent and easy to count, then the LED is flashed bright green once for each bar of signal strength.  If the signal is really weak, dim green with no flashes equates to zero bars.
* Two clicks of the SETUP button invokes the Soft Power Down feature that was previously implemented, which powers off the modem and puts the STM32 in deep sleep mode.
* Disables the single/double tap feature for non-electron devices.

A little preview of how this works with 1 bar of signal strength.  Sorry my signal is not the best right now :)
https://www.dropbox.com/s/hph5mldtxysr185/rssi-ux-and-soft-power-down.mov?dl=0

### DIFF
--- a/system/src/main.cpp
+++ b/system/src/main.cpp
@@ -56,6 +56,7 @@ using namespace spark;
 /* Private typedef -----------------------------------------------------------*/
 
 /* Private define ------------------------------------------------------------*/
+#if Wiring_SetupButtonUX
 #if defined(DEBUG_BUTTON_WD)
 #define BUTTON_WD_DEBUG(x,...) DEBUG(x,__VA_ARGS__)
 #else
@@ -78,6 +79,7 @@ inline void CLR_BUTTON_TIMEOUT() {
     button_timeout_duration = 0;
     BUTTON_WD_DEBUG("Button WD Cleared, was %d",button_timeout_duration);
 }
+#endif // #if Wiring_SetupButtonUX
 
 /* Private macro -------------------------------------------------------------*/
 
@@ -95,15 +97,16 @@ static volatile bool wasListeningOnButtonPress;
  */
 static volatile uint16_t pressed_time;
 
-/* flag used to initiate system_handle_single_click() from main thread */
-static volatile bool SYSTEM_HANDLE_SINGLE_CLICK = false;
-
 uint16_t system_button_pushed_duration(uint8_t button, void*)
 {
     if (button || network.listening())
         return 0;
     return pressed_time ? HAL_Timer_Get_Milli_Seconds()-pressed_time : 0;
 }
+
+#if Wiring_SetupButtonUX
+/* flag used to initiate system_handle_single_click() from main thread */
+static volatile bool SYSTEM_HANDLE_SINGLE_CLICK = false;
 
 static uint32_t prev_release_time = 0;
 static uint8_t clicks = 0;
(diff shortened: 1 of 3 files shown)


system/src/main.cpp
```diff
@@ -65,52 +93,118 @@ static volatile uint32_t TimingIWDGReload;
...
+}
+
+void handle_button_click(uint16_t depressed_duration, uint32_t release_time)
+{
+    uint32_t start_time = release_time - depressed_duration;
+    uint32_t since_prev = start_time - prev_release_time;
     bool reset = true;
-    if (duration<500) {                                 // a short button press
+    if (depressed_duration<500) {               // a short button press
         if (!clicks || (since_prev<1000)) {		// first click or within 1 second of the previous click
```
jme783: So we give the user 1s in between clicks to trigger the RSSI mode? Seems long enough, but was this an arbitrary choice?

technobly: I wouldn't say arbitrary, but it allows for a fast or slow click and double click.

* A 10ms to 999ms single click should kick off the RSSI UX
* An additional 10 to 999ms click within 1500ms of the original click will kick off the Soft Power Down feature.

system/src/main.cpp
```diff
@@ -65,52 +93,118 @@ static volatile uint32_t TimingIWDGReload;
...
+     *   5: >= -57 dBm
+     */
+    if (SYSTEM_HANDLE_SINGLE_CLICK) {
+        SYSTEM_HANDLE_SINGLE_CLICK = false;
+        volatile system_tick_t startTime = SYSTEM_TICK_COUNTER;
+        RGB.control(true);
+        RGB.color(0,10,0);
+        int rssi = 0;
+        int bars = 0;
+#if Wiring_WiFi == 1
```
jme783: Ah! So is this reusable for the Photon as well? Are there different mappings between `dBm` and # of bars for cellular vs. WiFi, or is it cool to use the same ranges?

technobly: I added this for the future.  Currently `Wiring_SetupButtonUX is set to 0` for all devices except the Electron.  Before I did that though I did test it on a photon and it worked.  We want to add a user API that allow users to OPT IN to enable this feature though, because many users are using the SETUP button for various things already (i.e. Short button presses up to 3 seconds, remember anything over 3 seconds activates Listening Mode).

I looked for a while for WiFi based RSSI values... every system is a bit different so we'll have to come up with values that seem appropriate for the Photon.

m-mcgowan:
How about using different color for no signal? e.g. orange? that way the user can clearly distinguish having a signal from not having one. cc: @jenesaisdiq

technobly:
@m-mcgowan We have experimented with white for zero bars, but it's ends up being a bunch of different colors happening that gets to be confusing.  Going into this mode with dim green for a dedicated period of time with no green flashes is more clear as zero bars than showing white and then green flashes if there are BAR levels to display.

m-mcgowan:
I wasn't advocating for a color transition, just a different color continually for 0 bars (not white then green.)   To me green means the connection is ok, but when there's no connection that's a misnomer.

jenesaisdiq:
Hey guys, the really dim green makes sense to me for 0 bars in this context. It won't be the only indication- remember that it's just when the user queries it for RSSI. If there isn't a connection, you'll also be seeing other status patterns to reflect that.

PR #773, opened 2015-12-17
Fixed unset wiced_result for tcp clients on socket_send

tlangmo:
I noticed that on the Photon platform, TCPClient::write does ALWAYS return -7009 if it is a client connection of a TCPServer. Looking at the code, I see that the socket return code is written to the wrong variable, result instead of wiced_result. This is leads to a constant return value of WICED_TCPIP_INVALID_SOCKET (7009) which subsequently gets negated.

Casting the int return code to wiced_result_t and assigning it to wiced_result is the fix.


m-mcgowan:
Thanks for the heads up about this oversight. :+1:  I changed the return type of tcp_server_client_t::write so that we avoid the cast.

PR #778, opened 2015-12-21
Fix system sleep #655

### LINKED ISSUE #655: [Photon]: System.sleep(_pin, _edge) never wakes
There is a similar issue for Core ( #640 ).
There is some confusion, since the docs state this should work, but it doesn't.
I've tested all pins with all FALLING/RISING/CHANGE and apart from some CHANGE combinations none of the other combinations did wake the Photon - not even WKP.

I tested with this code
```
void setup() 
{
  pinMode(D7, OUTPUT);
  Particle.function("GoToSleep", go2sleep);
}

void loop() 
{
    digitalWrite(D7, (millis() >> 2) & 0x88);
}

int go2sleep(String cmd)
{
    int i;
    if (i = cmd.toInt())
    {
        int pin = abs(i) % 20;  // 20 denotes pin D0
        PinMode input = INPUT;
        InterruptMode edge = CHANGE;

        if (abs(i) < 100) // >= 100 CHANGE
        {
            input = (i > 0) ? INPUT_PULLDOWN : INPUT_PULLUP;
            edge = (i > 0) ? RISING : FALLING;
        }
        pinMode(pin, input);
        System.sleep(pin, edge);
    }
    return i;
}

```
I also realized, that after the few CHANGE wakes that worked, the code continued from within the `Particle.function()` which does not fit the docs
```
When the specific interrupt arrives, the device awakens from stop mode, it will behave 
as if the device is reset and run all user code from the beginning with no values being 
maintained in memory from before the stop mode.
```
Although the note states
```
(Note: The new Particle Photon firmware will not reset before going into stop mode so 
all the application variables are preserved after waking up from this mode. The voltage 
regulator is put in low-power mode. This mode achieves the lowest power consumption 
while retaining the contents of SRAM and registers.)
```
What is meant with "new firmware"? A version number should be stated.
And also that the code carries on from where it left off (preserved registers doesn't mean too much to average users - as instruction and stack pointer would not do either)

peterqu:
## Change

1. Initialized GPIO_initStructure in pinMode()
2. Initialized EXTI_InitStructure and NVIC_InitStructure in HAL_Interupts_Attach() and HAL_Interrupts_Detach().

## Testing

Tested on Photon and Core pin A0~A7, D0~D7, WK(Photon only).  
Tested RISING, FALLING and CHANGE mode.  
Tested both System.sleep(pin, interruptMode, delay) and System.sleep(pin, interruptMode) usages.

## Test Application

~~~c
#include "application.h"

// Cloud Function example usage.
// Invoke go2sleep cloud function from Particle Dev IDE>Particle>Show Code Functions
// go2sleep(A2,C) // Setup wake up pin to A2 and change interrupt mode
// go2sleep(D0,F) // Setup wake up pin to D0 and falling interrupt mode
// go2sleep(D1,R) // Setup wake up pin to D1 and rising interrupt mode
int go2sleep(String cmd)
{

	// Select Pin
	int pinNumber = cmd.charAt(1) - '0';
	if (pinNumber< 0 || pinNumber >7) return -1;
	if (cmd.startsWith("A")){
		pinNumber += 10;
	}
	if (pinNumber>=10){
		Serial.printlnf("Pin A%d selected", pinNumber-10);
	} else {
		Serial.printlnf("Pin D%d selected", pinNumber);
	}

	// Select Mode
	InterruptMode interruptMOde = CHANGE;
	switch (cmd.charAt(3)){
		case 'C':
			interruptMOde = CHANGE;
			Serial.println("Mode is CHANGE");
			break;
		case 'F':
			interruptMOde = FALLING;
			Serial.println("Mode is FALLING");
			break;
		case 'R':
			interruptMOde = RISING;
			Serial.println("Mode is RISING");
			break;
		default:
			Serial.println("Illegal mode, use upper case letter!");
			return -1;
			break;
	}
	Serial.println("Going to sleep");
	delay(1000);
	System.sleep(pinNumber, interruptMOde);
	//System.sleep(pinNumber, interruptMOde, 0);
	return 0;
}

void setup()
{
	Particle.function("GoToSleep", go2sleep);
	Serial.begin(9600);
}

void loop()
{
	Serial.println("I am alive");
	delay(1000);
}
~~~

Fixes #655


m-mcgowan:
Awesome to hear we have a resolution. As we know, this is an intermittent bug due to stack conditions and unitialized stack variables. Can you confirm that this change works when applied to non-working firmware?

peterqu:
Hi @m-mcgowan 

I just found one more failure case with wake up issue.  This PR would failed at some timing-dependent scenarios.

If I add a delay before going into sleep.
   
    delay(x);
    System.sleep(D0, RISING);

For x=1000, 500, 300, 200, wake up works flawlessly.  For x=100, 50, the wake up didn't work.  I was wondering if there is anything timing-related could be the culprit here, maybe sysclk, iwdg?

technobly:
@peterqu IWDG is disabled so doubt it's that.  Might this be more uninitialized structures because adding any code definitely can rearrange memory.  Maybe try adding other non-delay type pieces of code that could have a similar effect in keeping sleep from waking?

m-mcgowan:
I will merge this even though there are unresolved issues, since these changes are useful.

schwiet:
Oddly, on the latest branch this worked for me flawlessly during my tests (using RISING on a black Core). Today could not figure out why it wasn't working. Glad I found this bug, will it make it to the latest branch soon? if not, is there a workaround?

PR #782, opened 2015-12-23
Reset WiFi credentials right after initializing WLAN hardware

### LINKED ISSUE #736: Factory Reset doesn't clear WiFi credentials before invoking network.connect()
Hi there,

Generically, clearing the WiFi credentials only happens when performing a factory reset and it actually executes when invoking the network.connect(). This would sound fine at most of time.

But there is a situation we must consider in that if we perform a factory reset and following set `SYSTEM_MODE(MANUAL);` in the sketch, it would not invoke the network.connect() at startup. Thus it will not  clear the WiFi credentials even if the flag `NVMEM_SPARK_Reset_SysFlag` was set. Basing on this situation, It will cause these problems:

1. If we invoked the function `WiFi.hasCredentials()` before `WiFi.connect()`, it would return `true`, instead of expected `false`.

2. If we set WiFi credentials using `WiFi.setCredentials()` and then invoke `WiFi.connect()` to trying connecting to local AP, it will enter the listen mode, since in the `connect()` function it checks the `NVMEM_SPARK_Reset_SysFlag` and executs clearing the WiFi credentials we just set before. Unless we have to perform a reset so that the `NVMEM_SPARK_Reset_SysFlag` was cleared and it can connect to local AP.

I think if we actually want to perform a factory reset to clear the credentials, the device should execute this action upon finishing factory firmware reset or at startup of system part 2.

Cheers!
Guohui

sergeuz:
Fixes #736

### DIFF
--- a/system/src/system_network_wifi.h
+++ b/system/src/system_network_wifi.h
@@ -94,11 +94,6 @@ protected:
     {
         wlan_connect_init();
 
-        if (wlan_reset_credentials_store_required())
-        {
-            wlan_reset_credentials_store();
-        }
-
         Set_NetApp_Timeout();
     }
 
@@ -174,6 +169,11 @@ public:
     void setup() override
     {
         wlan_setup();
+
+        if (wlan_reset_credentials_store_required())
+        {
+            wlan_reset_credentials_store();
+        }
     }
 
     void fetch_ipconfig(WLanConfig* target) override


m-mcgowan:
I'm guessing this is tested on a Photon? In theory this looks fine - the code was placed during connection since that's how it was on the Core originally. I will try this PR on a Core before merging, since the wifi device might behave differently.

sergeuz:
Yes, seems to work fine on Photon. I also had some doubts regarding the Core, as its wlan_* calls are implemented differently. Let me know if it indeed behaves differently there.

m-mcgowan:
I just tested on the Core - unfortunately it hardfaults. I'll investigate and see if there's a way tthat the Core HAL can initialize WiFi differently to work around this.

PR #783, opened 2015-12-24
Patch Time.now with zone

ScruffR:
Since `Time.now()` always returns UTC/GMT irrespective of the set `Time.zone()` I thought it might be useful to add an easy way to have `Time.now()` return epoch time either with or without (=default) the offset.
`static time_t now(bool withZone = false);`
`Time.now()` will behave as always
`Time.now(false)` returns the same 
`Time.now(true)` returns the epoch time + timeoffset in seconds

### DIFF
--- a/wiring/inc/spark_wiring_time.h
+++ b/wiring/inc/spark_wiring_time.h
@@ -12,7 +12,7 @@
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation, either
   version 3 of the License, or (at your option) any later version.
-
+9
   This library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
@@ -56,7 +56,7 @@ public:
 	static int     month(time_t t);   			// the month for the given time
 	static int     year();            			// current four digit year
 	static int     year(time_t t);    			// the year for the given time
-	static time_t  now();              			// return the current time as seconds since Jan 1 1970
+	static time_t  now(bool withZone = false);		// return the current time as seconds since Jan 1 1970
 	static void    zone(float GMT_Offset);		// set the time zone (+/-) offset from GMT
 	static void    setTime(time_t t);			// set the given time as unix/rtc time
 
--- a/wiring/src/spark_wiring_time.cpp
+++ b/wiring/src/spark_wiring_time.cpp
@@ -243,9 +243,9 @@ int TimeClass::year(time_t t)
 }
 
 /* return the current time as seconds since Jan 1 1970 */
-time_t TimeClass::now()
+time_t TimeClass::now(bool withZone)
 {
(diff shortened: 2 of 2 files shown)


m-mcgowan:
Looks good. I wonder if we should have a separate method to retrieve the local time vs UTC time. e.g. 

```
Time.local();
```

ScruffR:
I was thinking of that too, but couldn't decide on a name, but `local()` seems fine to me.
I thought of `Time.nowLocal()` vs. `Time.nowWithZone()` vs. `Time.nowTimezoned()` - thinking too complicated :stuck_out_tongue_winking_eye:

m-mcgowan:
Would be much appreciated if you could also add API tests in `user/tests/wiring/api` for the new overloaded function, and a test case in `user/test/wiring/no_fixture`. You'll see we already  have tests for the time API there.  

@jenesaisdiq Any thoughts on this proposed API:

```
uint32_t utc_time = Time.now();  
// time is current UTC time - this is what we presently have

// proposed Time.local() API
Time.zone(-5);
uint32_t  local_time = Time.local();
// local_time will be 5 hours earlier than utc_time
```

jenesaisdiq:
Sounds fine to me!

bkobkobko:
I like this too.  This will help with the ported TImeAlarms since I had to peek under the hood and get the current timezone in the current version.

m-mcgowan:
To fetch the current timezone offset, we could also add `Time.zone()` - with 0 arguments it returns the current timezone:

```
Time.zone(5);   // +5 hours
float tz = Time.zone();
// tz == 5.0
```

avtolstoy:
@m-mcgowan Wouldn't TZ offset in seconds work better? Much like tm_gmtoff in struct time.

m-mcgowan:
Possibly, but the `Time.zone(hours float)` API already exists, setting the precedent. I'm guessing the rationale is that expressing zone offsets in hours is conceptually simpler since timezones are commonly described in hours difference.

avtolstoy:
@m-mcgowan The world is big and there are apparently some half hour or even 45-minute timezones out there.

m-mcgowan:
That's right,  consequently `Time.zone()`  takes a float, so fractional hours can be specified.

ScruffR:
@avtolstoy, there are even some with 45 min offsets, but I second Mat here with the use of `float` since the standardized way to denote timezones is as hours (and in the odd cases hh:mm).
So if we wanted to ríd the user of the need to think ;-) even more than with `float` I'd only add an overload `Time.zone(int hh, int mm)`, but having `Time.zone(int ss)` seems rather prone to error - and it looks odd and more complicated (e.g `Time.zone(44820)` vs. `Time.zone(+12.75)` for Catham Island (+12:45))

---

@m-mcgowan: Was the comment about adding a `test` meant for me? 
If so, I'm happy to PR it.

m-mcgowan:
Hi @ScruffR   I didn't see your latest comment before had already coded up the tests, and the new APIs discussed.  Thanks for this PR and the prompt to add local time support.

ScruffR:
:+1:

PR #785, opened 2015-12-24
Break listening loop before disabling listening mode in multithreaded setup

### LINKED ISSUE #743: Photon remains in listening mode after WiFi.listen(false)
My code (which I haven't tried boiling down to a minimal example yet) is running on a Photon with SYSTEM_THREAD(ENABLED); and firmware version 0.4.7 installed.

I call WiFi.listen(true) to put the Photon into listening mode, and it does so correctly (on-board LED switches to blue flash, and Photon WiFi AP shows up on my nearby iPhone). 

I later call WiFi.listen(false) to take the Photon out of listening mode, but nothing changes.

Is there a known problem with WiFi.listen(false) not working?

sergeuz:
Fixes #743

### DIFF
--- a/system/src/system_network.cpp
+++ b/system/src/system_network.cpp
@@ -142,7 +142,12 @@ void network_off(network_handle_t network, uint32_t flags, uint32_t param, void*
  */
 void network_listen(network_handle_t network, uint32_t flags, void*)
 {
-    SYSTEM_THREAD_CONTEXT_ASYNC_CALL(nif(network).listen(flags & NETWORK_LISTEN_EXIT));
+    const bool stop = flags & NETWORK_LISTEN_EXIT;
+    if (!SYSTEM_THREAD_CURRENT() && stop) {
+        // Break current listening loop to ensure that system thread can process asynchronous calls
+        nif(network).break_listen_loop();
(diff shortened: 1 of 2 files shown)


sergeuz:
@m-mcgowan I'm not sure that reusing WLAN_SMART_CONFIG_START as a loop flag is a good idea here, so any thoughts are welcomed

m-mcgowan:
Those changes look good. In the long run, I'd to break out the listening loop so that it doesn't lock up the system thread, but for now this is a good approach. 

You're right to raise which flag to use. I think here `WLAN_SERIAL_CONFIG_DONE` would be a better fit here since that flag exists to terminate the listen loop.  It's not a strong preference, but seems to better reflect the semantics of what we are doing here. 

An aside - the flag could be renamed to something more generic like, NETWORK_LISTEN_COMPLETE to make the intent clear, although feel free to leave that to a later revisiting of the code where we fix all the flags.

sergeuz:
I've updated PR accordingly. Thanks)

m-mcgowan:
I cherry picked the key change from your PR #793, which does the same job as this PR, so closing out.

PR #787, opened 2015-12-26
Check for listening loop running when Particle.publish() is invoked in multithreaded setup

### LINKED ISSUE #761: Particle.publish() hangs on Photon when called in listening mode with SYSTEM_THREAD(ENABLED)
I'm using the following to test:

    #include "application.h"

    SYSTEM_MODE(MANUAL);
    SYSTEM_THREAD(ENABLED);

    #define PUBLISH_PERIOD_MS 2000

    int last_publish_attempt_ms;

    void setup() {
      Serial.begin(9600);
      Particle.connect(); // Should be run without credentials present.
      last_publish_attempt_ms = 0;
      delay(2000);
    }

     void loop() {
       if (millis() - last_publish_attempt_ms >= PUBLISH_PERIOD_MS) {
         last_publish_attempt_ms = millis();
         Serial.println("pre-publish");
         Particle.publish("test");
         Serial.println("post-publish");
       }
     }

When monitoring the serial output over USB, a single "pre-publish" message is printed, but "post-publish" is never printed. Furthermore, the blinking blue light of listening mode becomes very slow after ~10 seconds. This issue is present in MANUAL, SEMI_AUTOMATIC, and AUTOMATIC modes.

sergeuz:
Fixes #761

### DIFF
--- a/system/inc/system_network_internal.h
+++ b/system/inc/system_network_internal.h
@@ -19,13 +19,12 @@
 #ifndef SYSTEM_NETWORK_INTERNAL_H
 #define	SYSTEM_NETWORK_INTERNAL_H
 
-
 enum eWanTimings
 {
     CONNECT_TO_ADDRESS_MAX = S2M(30),
     DISCONNECT_TO_RECONNECT = S2M(2),
 };
-    
+
 extern volatile uint8_t WLAN_CONNECTED;
 extern volatile uint8_t WLAN_DISCONNECT;
 extern volatile uint8_t WLAN_DHCP;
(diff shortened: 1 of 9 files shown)


sergeuz:
@m-mcgowan I moved some flag declarations to separate header file, as previously the same flags were declared with both C and C++ linkage in different files, which led to compilation errors

m-mcgowan:
Nice - this is a good start!  We could use this as is with no problems. I have just a few ancillary comments to help refine a little:

- the system flags are an internal implementation detail of the system module, in contrast to part of the system API, so I suggest moving the new header into the `system/src` folder so there's no chance of them being used outside the system module. 

- in future, listening mode may be fixed so that the system thread continues to be responsive, so listening mode itself isn't the problem, but rather that the system thread is busy. We can reflect the intent more directly with a method `is_system_thread_busy()` that returns whether the system thread is busy (at present this means in listening mode.)  When we later fix listening mode we have only to change this one method, rather than hunt for uses of `network_listening()` and change only those that are used to determine if the system thread is blocked.

sergeuz:
Mat, I've updated this PR accordingly. I decided not to introduce separate busy flag, as this fix is still more a temporary workaround until we start processing system loop normally when listening mode is enabled

m-mcgowan:
I've cherry picked the key change in #793 into develop, which does the same job as this PR , so closing this one out.

PR #792, opened 2016-01-04
Feature/electron/rssi ux threaded

m-mcgowan:
Moves the RSSI UX to the system tick thread, and the handling of the button press to the system thread so that it continues to work when the application thread is blocked. 

I upped the delay between pips from 250 to 350ms - 250ms seemed too quick and I had trouble counting. 

Manages system override and restoring the LED color, and fixes https://github.com/spark/firmware/issues/784

### DIFF
--- a/system/src/main.cpp
+++ b/system/src/main.cpp
@@ -88,6 +88,21 @@ static volatile uint32_t TimingLED;
 static volatile uint32_t TimingIWDGReload;
 
 /**
+ * When non-zero, causes the system to play out a count on the LED.
+ * For even values, the system displays a bright color for a short period,
+ * while for odd values the system displays a dim color for a longer period.
+ * The value is then decremented and the process repeated until 0.
+ */
+static volatile uint8_t SYSTEM_LED_COUNT;
+/**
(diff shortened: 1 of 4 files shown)


m-mcgowan:
I also noticed a delay when pressing the button and the LED changing - this leads so some user confusion about whether the button press has been registered.  I believe (but haven't proven) the delay was due to the time it takes to get the RSSI value from the modem - the LED was only changed after the RSSI was available. The new behavior is to change the LED as soon as the button press is registered - before getting the RSSI value, and then the LED bars are signalled once the RSSI value is known.

m-mcgowan:
I tested:
- RGB.control(false) then pushed the button - reverts back to breathing cyan
- RGB.control(true) then pushed the button - reverts back to the color previously set by the application
- powering off during the RSSI sequence. 

- 0 bars and 5 bars. For 0 bars we now just get dim green with 0 pips rather than no UX at all. 

Not tested:
- possible race conditions - since the RSSI is being displayed in parallel with other things, e.g. the app switching rapidly back between RGB.control(true), RGB.control(false).

Issue #788, opened 2015-12-26
Entering listening mode does not block the system thread


m-mcgowan:
Currently, when the device enters listening mode, the thread running system code enters a tight loop servicing the listening mode implementation:

- on entry to listening mode, setup the radio hardware for listening (Core: enter smart setup mode, Photon: start soft AP.)
- blinking the LED blue at 250ms intervals
- detecting when the WiFI profiles should be deleted and deleting them (flashing the LED quickly in the process)
- on exit from listening mode, the radio os reset to the normal state. (Core: exit smart config mode, Photon: tear down SoftAP and reinstate STA mode.)

Some consequences of this implementation:

- the cloud is not serviced during listening mode, nor are any of the other background thread processes.
- application code does not execute during listening mode in single threaded mode. In multithreaded mode, application code is intended to be executed (but may fail presently due to issues with the system thread being blocked.)
- Serial is used to receive commands during listening mode - this is interleaved with whatever the application was presently doing with Serial.

On the Photon, entering listening mode specifically tears down all sockets and the current wifi connection. This is to mirror the behavior on the Core.  In future, we can run listening mode in parallel with regular STA WiFi without having to tear down the existing wifi connection.  To this end, we have plans to allow the SoftAP mode to be used by application code, where the photon supports 2 network interfaces at the same time.

The solution will need to consider how Serial is handled in the case when the application loop continues to run, since Serial will be used by both the listening mode loop and application code. One possible approach is to allow the application to specify if listening mode should activate the Serial interface or not. The setting also takes affect when the end user manually enters listening mode by pressing Setup for 3 seconds.

karlhenricksen:
@m-mcgowan - I see slow blinking blue LED when calling Wifi.hasCredentials() in listening mode and the user app is blocked due to that. Is this behavior the same issue?
https://www.youtube.com/watch?v=SNmeHpOhAB0

m-mcgowan:
Yes. If you are building locally, you can try the latest develop state which allows the application to run while in listening mode.

karlhenricksen:
[...]

PR #793, opened 2016-01-04
Process system queue events while in listening mode

sergeuz:
This is to keep up the discussion summarized here: #788, and also a generic fix for #743 and #761.

The patch ensures that system queue is processed while listening mode is enabled in multithreaded configuration, so that synchronous calls that need to be executed in the context of system thread don't block application code.

Solution is far from ideal, as listening loop is still separated from the main system loop, and thus additional checks in unrelated parts of the code are required in order to break this listening loop when necessary (e.g. https://github.com/sergeuz/firmware/blob/loop/system/src/system_network_internal.h#L378). In general, it seems that our whole flags and loops stuff needs to be refactored in favor of more or less strict state machine with centralized control over initialization and uninitialization of different device modes, related LED indication, etc -- but I'm not sure whether it goes well with our current priorities.

As for potentially interleaved user and system serial communications, proposed solution is to allow user to explicitly disable serial console in listening mode:

```c++
SYSTEM_THREAD(ENABLED)
SYSTEM_OPTIONS(SERIAL_CONSOLE_DISABLED)

void setup() {
    // ...
}

void loop() {
    // ...
}
```

`SYSTEM_OPTIONS(...)` macro expands to a series of calls to `system_set_option(int option, void *value)` function exposed by the system module, and can be extended to support other options if necessary:

```c++
SYSTEM_OPTIONS(
    SOME_SYSTEM_FLAG, // or SOME_SYSTEM_FLAG = true
    SOME_SYSTEM_TIMEOUT = 10000,
    SOME_STRING_OPTION = "hello"
)
```

### DIFF
--- a/user/tests/wiring/threading/application.cpp
+++ b/user/tests/wiring/threading/application.cpp
@@ -26,61 +26,41 @@
 #include "system_threading.h"
 #include <functional>
 
-void* abc = NULL;
-
 UNIT_TEST_APP();
 SYSTEM_THREAD(ENABLED);
 
-int test_val;
+volatile int test_val;
 void increment(void)
 {
 	test_val++;
+}
+
+test(system_thread_can_pump_events)
+{
+    test_val = 0;
+
+    ActiveObjectBase* system = (ActiveObjectBase*)system_internal(1, nullptr); // Returns system thread instance
+    system->invoke_async(std::function<void()>(increment));
+
+    uint32_t start = millis();
+    while (test_val != 1 && millis() - start < 1000); // Busy wait
+
+    assertEqual((int)test_val, 1);
 }
 
 test(application_thread_can_pump_events)
 {
 	test_val = 0;
 
-	ActiveObjectBase* app = (ActiveObjectBase*)system_internal(0, nullptr);
+	ActiveObjectBase* app = (ActiveObjectBase*)system_internal(0, nullptr); // Returns application thread instance
 	std::function<void(void)> fn = increment;
 	app->invoke_async(fn);
 
 	// test value not incremented
-	assertEqual(test_val, 0);
+	assertEqual((int)test_val, 0);
 
     Particle.process();
 
     // validate the function was called.
-    assertEqual(test_val, 1);
-
+    assertEqual((int)test_val, 1);
 }
-
-uint32_t last_listen_event = 0;
-void app_listening(system_event_t event, uint32_t time, void*)
-{
-	last_listen_event = time;
-	if (time>2000)
-		WiFi.listen(false); // exit listening mode
-}
-
-// This test ensures the RTOS time slices between threads when the system is in listening mode.
-// It's not a complete test, since we never exit or re-enter loop so the loop-calling logic isn't exercised.
-// This test indirectly ensures the system thread also runs, since if it didn't then the listen update events wouldn't be
-// published.
-test(application_thread_runs_during_listening_mode)
-{
-	System.on(wifi_listen_update, app_listening);
-
-	uint32_t start = millis();
-	WiFi.listen();
-	delay(10);		// time for the system thread to enter listening mode
-
-	while (millis()-start<1000);		// busy wait 1000 ms
-
-	uint32_t end = millis();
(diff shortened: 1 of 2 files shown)


m-mcgowan:
For the folks following along, the key change is calling `SystemThread.process()` from within the listen loop to allow the system thread to process requests while in listening mode. :+1:  to that. 

Can you explain the rationale behind using a new system flags macro - we already have `System.set()` to set various flags. For disabling serial during listening, we have `WiFi.listen()` could be extended to take additional flags so that serial can be disabled at the point listen mode is invoked.

m-mcgowan:
The PR contains a number of changes above the core functional change. Here's some tips for making code review simpler:

- please try to keep commits and PRs focused on a single change. 
- whitespace changes, moving content around are best put in their own commit
- when making API changes (such as the system flags), please write up a brief proposal first, so many get the chance to offer their viewpoint.  

Thanks!

sergeuz:
Mat, thanks for the tips. Indeed, overall diff looks a bit messy. Key changes are isolated in this commit: 16b253f14fa29de52e0a28f03d69286ec6f0a07a.

Rationale behind `SYSTEM_OPTIONS` macro was to have a way to adjust some system defaults (not necessarily flags) globally before any code is executed -- in similar way `SYSTEM_MODE` and `SYSTEM_THREAD` macros work. It seems I overlooked that existent `system_flag_t` flags can be changed by user code for configuration purposes.

As for adding parameter to `WiFi.listen()` call, what about the case when listening mode is enabled via setup button?

m-mcgowan:
I agree that disabling serial when the setup button is pressed is something to think about, and agree with the other points too. It was discussion that I wanted to encourage!

I like the system configuration macro - it was a nice implementation -  but my worry is that we have many different ways to configure system features. Perhaps the different ways can be normalized under this one scheme (in time)? For now, I think we should use an existing configuration API (e.g. System.set()) and focus later on normalizing a good configuration API. 

Thinking further afield, all the default functionality in our system should be eventually overridable, e.g the onboard LED, the setup button, listening interfaces (e.g. serial interface / HTTP interface in listening mode.)  It's not clear to me the best way to achieve this in a consistent and simple way, so I'd prefer to shelve this until we have more time to work it through.

sergeuz:
Summing all above let's do the following: I'll update this PR, leaving only changes related to listening loop here, and then register another PR allowing to disable serial in listening mode (via additional `system_flag_t` flag). Latter is a quick fix, but anyway let me know if it's not something necessary at the moment.

m-mcgowan:
Sounds great :+1:

m-mcgowan:
I'd like to also automate testing for these changes. Could you please add a new test suite for mutithreading in `user/tests/wiring/threading` so that we can write integration tests for multithreading behaviors. You can use the existing suite `user/tests/wiring/no_fixture` as a template if that helps.

I can appreciate this might be tricky so feel free to bounce ideas off me.

sergeuz:
I've updated this PR, so now it contains only tests related to listening mode with enabled threading (primary changes are already in `develop` branch: 0945540412009b1d9d7b1569c32980daea26a111).

sergeuz:
@m-mcgowan There's a problem with stability of updated wiring/threading suite: while all tests succeed individually, some of them may fail when you run them as part of the suite. Likely it happens because of the listening mode that is toggled between the cases, but I haven't tracked down the issue yet :(

m-mcgowan:
I fixed up the test cases - they run reliably for me. :+1:

sergeuz:
Wow, thanks!

Just curious, why do we need such long timeouts while waiting for relatively simple operations to complete: https://github.com/spark/firmware/blob/develop/user/tests/wiring/threading/application.cpp#L47?

m-mcgowan:
I didn't give it much thought, but thanks for asking since now I have. ;-) 

When the system comes out of listening mode, it will then attempt to reconnect to wifi and then connect to the cloud - this can take some time.  Ideally this should be part of the listening_stop() call so we can factor it out of the subsequent actions.

sergeuz:
ah, got it, thanks)

PR #796, opened 2016-01-05
Fixes USART_ISR infinite loop due to ORE flag not handled

jakeypoo:
Fixes the issue discussed here: https://github.com/spark/firmware/issues/776

### DIFF
--- a/hal/src/stm32f2xx/usart_hal.c
+++ b/hal/src/stm32f2xx/usart_hal.c
@@ -384,12 +384,12 @@ static void HAL_USART_Handler(HAL_USART_Serial serial)
 		}
 	}
 
-	// // If Overrun occurs, clear the OVR condition
-	// if (USART_GetFlagStatus(usartMap[serial]->usart_peripheral, USART_FLAG_ORE) != RESET)
-	// {
-	// 	(void)USART_ReceiveData(usartMap[serial]->usart_peripheral);
-	// 	USART_ClearITPendingBit (usartMap[serial]->usart_peripheral, USART_IT_ORE);
-	// }
+    	if (USART_GetFlagStatus(usartMap[serial]->usart_peripheral, USART_FLAG_ORE) != RESET)
+    	{
+    		// If Overrun flag is still set, clear it
+        	(void)USART_ReceiveData(usartMap[serial]->usart_peripheral);
+    	}
+
 }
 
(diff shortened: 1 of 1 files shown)


technobly:
@jakeypoo thanks for the PR :)

Is there any reason you didn't also leave in the code to clear the OVR IT pending bit?
```c++
USART_ClearITPendingBit (usartMap[serial]->usart_peripheral, USART_IT_ORE);
```

jakeypoo:
@technobly The ClearITPendingBit function doesn't handle the ORE flag. 
That bit is only cleared by reading the USART_SR followed by reading the UART Data register. So all we need to do is just read the data buffer, no need to manually clear that bit.

technobly:
I just dug into the ref manual and I see that now.  Also there is this:

>The ORE bit, when set, indicates that at least 1 data has been lost. There are two possibilities:
>• if RXNE=1, then the last valid data is stored in the receive register RDR and can be read,
>• if RXNE=0, then it means that the last valid data has already been read and thus there is nothing to be read in the RDR. This case can occur when the last valid data is read in the RDR at the same time as the new (and lost) data is received. It may also occur when the new data is received during the reading sequence (between the USART_SR register read access and the USART_DR read access).

Because this test for the ORE flag is after RXNE and TXNE IT status flags are serviced, we can be reasonably sure that we are not throwing away valid data by just reading the RDR.  Do you agree?

jakeypoo:
Yup, the byte in the RDR has already been read. The RDR just needs to see a
read operation to clear the ORE.

A side note, depending on your application, the ORE flag being raised could
mean that the data coming in is not valid if it relies on the preceding
byte(s) (which were lost).
It would be nice to implement a larger [than one byte] hardware UART buffer
to avoid the ORE condition. I'm surprised the stm32 doesn't have this
option enabled by default. Instead you have to enable DMA and allocate a
buffer.

technobly:
Tested with and without the fix, looks great.  Thanks again @jakeypoo your test app was crucial in validating this fast ;-)

technobly:
@jakeypoo would you please fill out this contributor license agreement for me?
https://docs.google.com/a/spark.io/forms/d/1_2P-vRKGUFg5bmpcKLHO_qNZWGi5HKYnfrrkd-sbZoA/viewform
Thanks!

PR #797, opened 2016-01-05
Fixes sleep (STOP mode) on Photon platform

### LINKED ISSUE #655: [Photon]: System.sleep(_pin, _edge) never wakes

avtolstoy:
HAL_Core_Enter_Stop_Mode should be wrapped with a critical section (cpsid/cpsie) to avoid
having an interrupt triggered while the device is entering sleep mode. Interrupts can be safely
disabled while in STOP mode.
All EXTI interrupts but the one associated with selected wakeup pin should be disabled as well
and restored after exiting STOP mode.

Fixes #655

### DIFF
--- a/hal/src/stm32f2xx/core_hal_stm32f2xx.c
+++ b/hal/src/stm32f2xx/core_hal_stm32f2xx.c
@@ -48,6 +48,9 @@
 #include "dct.h"
 #include "hal_platform.h"
 
+#define STOP_MODE_EXIT_CONDITION_PIN 0x01
+#define STOP_MODE_EXIT_CONDITION_RTC 0x02
+
 void HardFault_Handler( void ) __attribute__( ( naked ) );
 
 __attribute__((externally_visible)) void prvGetRegistersFromStack( uint32_t *pulFaultStackAddress )
@@ -366,7 +369,7 @@ void HAL_Core_Enter_Safe_Mode(void* reserved)
 
 void HAL_Core_Enter_Stop_Mode(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long seconds)
 {
-    if (!((wakeUpPin < TOTAL_PINS) && (edgeTriggerMode <= FALLING)) && seconds <= 0)
+    if (!((wakeUpPin < TOTAL_PINS) && (wakeUpPin >= 0) && (edgeTriggerMode <= FALLING)) && seconds <= 0)
         return;
 
     HAL_disable_irq();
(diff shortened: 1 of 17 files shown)


avtolstoy:
I have updated the PR. Waking up on RTC should now work correctly on Photon.
b92448f0c21055fb5219f9d1437dfd2cb6f1d2bc also correctly suspends networking before entering STOP mode and restores it after waking up.

m-mcgowan:
Tested on Core, Photon, Electron with a test app:

 - registers a `sleep` function with the cloud, arg is the number of seconds to sleep.  The sleep call is `System.sleep(A1, RISING, seconds)`
 - loop() writes the current millis() count to serial every 100ms (no delays, polls millis())
 - A1 connected to GND

Results:

- calling `sleep 10`, the device shutdown the network, then slept for 10 seconds. After waking up, the network/cloud reconnected and serial output resumed. (On the Core, the serial connection terminates on sleep, since the device resets as part of executing stop mode.)
- calling `sleep 1000` the device sleeps for a long time. Wakes up when connecting A1 to 3v3.  No other gpio pins wake the device.
- after waking up, verified that the setup button continues to work
- tried this 10 times.

Outlier on the Electron, U270 rev15:

- when the A1 pin is floating, the electron would not sleep but wake up immediately. When this happened the setup button no longer continued to function. This was quite hard to reproduce reliably, and didn't happen when the pin was connected to gnd/3v3.

m-mcgowan:
As you can see from the test above, this works well on all our platforms. Nice job @avtolstoy :+1:

hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -364,37 +367,84 @@ void HAL_Core_Enter_Safe_Mode(void* reserved)
     HAL_Core_System_Reset();
 }
 
-void HAL_Core_Enter_Stop_Mode(uint16_t wakeUpPin, uint16_t edgeTriggerMode)
+void HAL_Core_Enter_Stop_Mode(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long seconds)
 {
+    if (!((wakeUpPin < TOTAL_PINS) && (wakeUpPin >= 0) && (edgeTriggerMode <= FALLING)) && seconds <= 0)
```
m-mcgowan: I think this should be `|| seconds < 0` - I will add that edit in develop.

PR #800, opened 2016-01-08
notify system of PSD disconn. during GPRS detach & PDP deactivate URCs

technobly:
UDP currently pings the server every 23 minutes... 

If a PDP context is remotely deactivated or GPRS detached remotely, it would normally take up to 23 minutes to find out about it.  This PR catches the URCs associated with these events and notifies the system of network disconnect.

For GPRS detach, we set a 15 second watchdog timer because GPRS detach events have been seen to re-attach right after detaching.  No need to churn the whole system for that.  If it re-attaches, the WDT is cleared.

PDP deactivation will issue an immediate system network disconnect and retry.

To test:
* Try removing your antenna, waiting for logs to indicate a WDT is set... then replace the antenna and see if it clears.  Try publishing before and after.
* Try the same thing, but this time wait longer until the network is disconnected.
* Try to force a PDP context deactivation by deactivating the SIM, and reactivate.. make sure publish works before and after reconnect.

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -34,6 +34,7 @@
 #include "service_debug.h"
 #include "concurrent_hal.h"
 #include <mutex>
+#include "net_hal.h"
 
 std::recursive_mutex mdm_mutex;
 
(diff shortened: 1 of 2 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -257,10 +296,15 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
 
                 // GSM/UMTS Specific -------------------------------------------
                 // +UUPSDD: <profile_id>
-                if (sscanf(cmd, "UUPSDD: %d",&a) == 1) {
-                    if (*PROFILE == a) {
+                if (sscanf(cmd, "UUPSDD: %c", s) == 1) {
+                    DEBUG_D("UUPSDD: %s matched\r\n", PROFILE);
```
m-mcgowan: shouldn't the `PROFILE` param be `c`?

technobly: I started with %c but then switched to %s because I already had a buffer defined.  Need to change %c to %s

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -257,10 +296,15 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
...
+                    if ( strcmp(s, PROFILE) != -1 ) {
```
technobly: This should be `!strcmp(s, PROFILE)` as well...

PR #801, opened 2016-01-09
SPI_DMA_UserCallback should be called in SPI RX DMA IRQ Handler

### LINKED ISSUE #791: SPI DMA transfers are calling the callback too early
https://community.particle.io/t/bug-in-spi-block-transfer-complete-callback/18568

avtolstoy:
Fixes #791

SPI hardware might still be busy pushing out bytes when SPI TX DMA transfer complete IRQ handler is fired. To ensure that SPI user callback is called when the actual transfer has been completed, it should be called from SPI RX DMA transfer complete IRQ handler.

1 (Yellow) - MISO/MOSI (loopback mode)
2 (Blue) - SPI User callback indicator pin (pulled low in callback)
3 (Purple) - SCK

This is the behavior observed in develop branch at 100KHz.
Left cursor indicates a point the user callback was called, right cursor indicates a start of a non-DMA single byte transfer sequence (0x00, 0xff, 0x00). Despite the fact that all the bytes were pushed out, this is still incorrect behavior, since neither actual TX transfer has been completed, nor RX DMA has been completed when user callback was called.
ds1z_quickprint3

The behavior observed with this PR again at 100KHz.
The cursor indicates a point the user callback was called, immediately followed by non-DMA single byte transfer sequence (0x00, 0xff, 0x00).
ds1z_quickprint4

A test application used:

```C++
#include "application.h"

Serial1DebugOutput dbg(57600);

static volatile bool SPI_DMA_TransferCompleted = false;

void SPI_User_Callback(void)
{
    digitalWrite(D0, LOW);
    SPI_DMA_TransferCompleted = true;
}

static uint8_t tx[512];
static uint8_t rx[512];

void test() {
    memset(rx, 0, sizeof(rx));
    SPI_DMA_TransferCompleted = false;
    digitalWrite(D0, HIGH);
    SPI.transfer(tx, NULL, sizeof(tx), SPI_User_Callback);
    while(!SPI_DMA_TransferCompleted);
    SPI.transfer(0x00);
    SPI.transfer(0xff);
    SPI.transfer(0x00);    
    DEBUG("Complete");
}

void setup()
{
    for (int i = 0; i < sizeof(tx); i++)
        tx[i] = (uint8_t)i;
    Serial.begin(9600);
    pinMode(D0, OUTPUT);
    pinMode(SCK, AF_OUTPUT_PUSHPULL);
    pinMode(MOSI, AF_OUTPUT_PUSHPULL);
    pinMode(MISO, INPUT);
    SPI.setDataMode(SPI_MODE0);
    SPI.setBitOrder(MSBFIRST);
    SPI.setClockSpeed(100000);
    SPI.begin();
}

void loop()
{
    test();
    delay(2000);
}
```

### DIFF
--- a/hal/src/electron/core_hal.c
+++ b/hal/src/electron/core_hal.c
@@ -133,7 +133,7 @@ IDX[x] = added IRQ handler
 71 [x] TIM7_IRQHandler                   // TIM7
 72 [ ] DMA2_Stream0_IRQHandler           // DMA2 Stream 0
 73 [ ] DMA2_Stream1_IRQHandler           // DMA2 Stream 1
-74 [ ] DMA2_Stream2_IRQHandler           // DMA2 Stream 2
+74 [x] DMA2_Stream2_IRQHandler           // DMA2 Stream 2
 75 [ ] DMA2_Stream3_IRQHandler           // DMA2 Stream 3
 76 [ ] DMA2_Stream4_IRQHandler           // DMA2 Stream 4
 77 [ ] ETH_IRQHandler                    // Ethernet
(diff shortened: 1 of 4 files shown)


technobly:
Very nice @avtolstoy :+1:

m-mcgowan:
Excellent! :+1:  Is there a test we can do to verify this also on the Electron? Suggestions for a hardware fixture for an automated test are very welcome.

PR #802, opened 2016-01-10
short circuit debugging

m-mcgowan:
short-circuits debug functions if no debug output defined (making the debug function mostly a no-op) - parameters to the function are still evaluated by the caller, but vsnprintf() is not invoked.

Adds a log level and void* extension point to `log_direct_`


towynlin:
:+1:

PR #805, opened 2016-01-13
Rust on Particle

m-mcgowan:
Blinky in Rust - tested on Photon, but compiles also for Core and Electron.


m-mcgowan:
These are the first steps to incorporating Rust as a supported language for coding application on Particle devices.

## Features

- It's Rust! :-)
- Works statically linked or dynamically linked (Core/Photon/Electron.)
- Can call any C function using the support in Rust for calling C 

## Limitations

- each Rust source file is compiled separately, which in practice limits the app to a single source file until the Rust module format is integrated into the make system
- libraries not supported 

## Getting started

- Install the latest Nightly build of Rust - https://www.rust-lang.org/downloads.html
- find the commit hash of the rust compiler:
```
$ rustc -v --version
rustc 1.7.0-nightly (4405944b9 2016-01-11)
binary: rustc
commit-hash: 4405944b9414d9d39d98c988c420b77e93acba96
commit-date: 2016-01-11
host: x86_64-apple-darwin
release: 1.7.0-nightly
``` 
- git clone https://github.com/rust-lang/rust
- checkout the commit listed above:
 - `git checkout <commit>`
- in this firmware repo, rebuild the core library for your specific version of the compiler
```
cd build/arm/rust
./build-core.sh <path-to-rust-repo>
```
- Build the default app  (rust blinky) just as you normally would
```
cd main
make PLATFORM=photon all program-dfu
```

## Next Development Steps
- Since Rust doesn't support calling C++, but only C, we have to code all our C++ API (mainly Wiring) into Rust
- add support for Rust libraries in the build
- add support for multi-file projects

dbrgn:
Great to see! :)

dbrgn:
Are you sure you want to commit `libcore.rlib` into this branch?

dbrgn:
Another aspect is what the goals of the Rust support are. Will there be official support of Rust at some point in time? If yes, there are a few questions about the API:

- Do we want to keep naming as similar to the Particle/Arduino API as possible, or should we instead use the Rust coding style? (`pinMode(PinMode.HIGH)` vs `pin_mode(PinMode.High)`)
- Do we want a simple C-like API, or an idiomatic rust-y API that takes full advantages of the traits system?

dbrgn:
For me, after doing those steps, the LED breathes magenta. If I understand it correctly, that means that the user application could not be loaded properly, right?

francisdb:
:+1: for working on this!

m-mcgowan:
Breathing magenta means the app dependencies aren't satisfied  -be sure you have built and flashed the system modules, since the freshly-compiled application will be expecting the latest versions of these. 

Regarding the API style, it's mostly unconsidered at this point - this is just a proof of concept - so we welcome feedback and direction on this. Here's my initial thoughts:

- on the one hand, keeping the API as close as possible to to our existing API would make transitioning to Rust very simple, and all the existing examples and documentation would still be valid. I understand this might break a few Rust conventions, such as use of camelCase, and importing enums into global space.  So the `pinMode` call would be as I have it in the code here - `pinMode(D7, HIGH)` just as with our current C++ API, compared to the more conventional Rust `pin_mode(Pin.D7, PinState.HIGH);`

- reworking the API into something that is mainly object-oriented and idiomatic would allow a more expressive API. I have only spent an hour of two with Rust so I'm not yet fully versed on what this would bring, but imagine there is huge potential here. 

These two options aren't mutually exclusive - we could support both styles of API so that transitioning is easy for those familiar with our API but new to Rust, while opening up for more advanced coding for seasoned coders.

dbrgn:
> I understand this might break a few Rust conventions, such as use of camelCase

What if you'd go for the Rust conventions and add a note to the docs that `camelCase` function calls should just be translated to `under_scores`? It works for the people that use AngularJS (there controllers use camelCase in JavaScript, but are called through html-style-names in the HTML).

> reworking the API into something that is mainly object-oriented and idiomatic would allow a more expressive API. I have only spent an hour of two with Rust so I'm not yet fully versed on what this would bring, but imagine there is huge potential here.

The rust type system is mostly based around [Traits](http://blog.rust-lang.org/2015/05/11/traits.html). If you master them, you can create really composable APIs.

For example, if you specify the digitalWrite function as follows:

```rust
pub fn digital_write<S>(pin: Pin, value: S) where S: Into<PinState> {
    unsafe {
        HAL_GPIO_Write(pin as i16, value.into() as u8);
    }
}
```

...then any type that implements the `Into<PinState>` trait can be used as a specification of pin state. I'm not sure how useful this example would be, but you could for instance implement the trait for a `MotorState` type. The `Into` implementation would then convert a `MotorState.TurnedOn` to a `PinState.High`. This means that a user of the motor library could pass a `MotorState` object directly to the `digitalWrite` function, while ensuring full type correctness.

zsup:
I'm not familiar with Rust, but I do think we should use proper Rust conventions rather than making it as Arduino-like as possible.

The way I see it, Arduino already exists as a simultaneously wonderful and awful mix of C/C++ syntax and Java-like syntax (e.g. `camelCase`). It serves its purpose well.

A Rust implementation, in my mind, should be more geared for professional software developers, and therefore should be written using Rust conventions, but pulling important _concepts_ over from Arduino-land (`digitalWrite` or `digital_write` as an important GPIO API, for instance). However I could also see arguments for a more Arduino-like implementation as well, so my opinion isn't super strong.

towynlin:
Awesome! :smiley: I'd vote for following Rust conventions as closely as possible rather than breaking the conventions to look more like Arduino.

dbrgn:
> Breathing magenta means the app dependencies aren't satisfied -be sure you have built and flashed the system modules, since the freshly-compiled application will be expecting the latest versions of these.

```
git checkout latest
cd modules
make clean all program-dfu PLATFORM=photon
cd ../main
make clean all program-dfu PLATFORM=photon
```

Flashing like this through DFU works, but even when using the `latest` branch instead of the rust one, the LED simply breathes magenta.

m-mcgowan:
You have to use this branch - latest has an older version of system firmware.

dbrgn:
Yeah, but if I flash both `modules` and `main` from latest, I should get either blinky or tinker, right?

Even if I flash `modules` and then `main` from your `feature/rust` branch, I get the magenta breathing.

m-mcgowan:
Please put the device in listening mode, then send 's' over serial, it will print out the versions of the system modules and the application code. Look for the `v` attribute - the 2nd and 3rd ones should read `"v"=10`.  

{"p":6,"m":[{"s":16384,"l":"m","vc":30,"vv":30,"f":"b","n":"0","v":4,"d":[]},{"s":262144,"l":"m","vc":30,"vv":30,"f":"s","n":"1",**"v":10**,"d":[]},{"s":262144,"l":"m","vc":30,"vv":30,"f":"s","n":"2",**"v":10**,"d":[{"f":"s","n":"1","v":10,"_":""}]},{"s":131072,"l":"m","vc":30,"vv":30,"u":"0A44251BDE164CBE919B015296C5C491DE7581349A480AD7FBD50A26D7E91C45","f":"u","n":"1","v":3,"d":[{"f":"s","n":"2",**"v":10**,"_":""}]},{"s":131072,"l":"f","vc":30,"vv":0,"d":[]}]}

Hitting 'v' should print `system firmware version: 0.4.8-rc.6`

dbrgn:
```
{"p":6,"m":[{"s":16384,"l":"m","vc":30,"vv":30,"f":"b","n":"0","v":4,"d":[]},{"s":262144,"l":"m","vc":30,"vv":28,"f":"s","n":"1","v":10,"d":[]},{"s":262144,"l":"m","vc":30,"vv":28,"f":"s","n":"2","v":10,"d":[{"f":"s","n":"1","v":10,"_":""}]},{"s":131072,"l":"m","vc":30,"vv":28,"u":"38B9447B99708D1D31FE871F22EA6139990B452C9888D3E2057FB348981B7EC8","f":"u","n":"1","v":3,"d":[{"f":"s","n":"2","v":10,"_":""}]},{"s":131072,"l":"f","vc":30,"vv":0,"d":[]}]}
system firmware version: 0.4.8-rc.6
```

Firmware seems to be at the right version.

I re-did the entire process again to make sure:

```
multirust default nightly
git clone https://github.com/spark/firmware particle-firmware
cd particle-firmware
git checkout feature/rust
cd build/arm/rust
./build-core.sh ~/Projects/rust
cd ../../../modules/
# put photon into dfu mode
make PLATFORM=photon clean all program-dfu
# photon restarts and breathes magenta
cd ../main/
# put photon into dfu mode
make PLATFORM=photon clean all program-dfu
# photon restarts and breathes magenta
```

The rustc version used is "rustc 1.7.0-nightly (4405944b9 2016-01-11)".

m-mcgowan:
It looks like the crc32 appended to the application is incorrect - `"vc":30,"vv":28` this means the CRC validation failed, so the system won't load the app. If you're building on Windows, be sure that the crc32 from mingw is used, and not the one from msys, since it uses a different algorithm.

dbrgn:
Aah, you found the culprit! I actually didn't have a `crc32` binary at all. I'm on Arch Linux. Installing `perl-archive-zip` helps, the provided crc32 binary seems to be compatible. It works now, wohoo!

Is this something that could be fixed in the build process, in order to show a proper warning? Maybe a small test with expected crc32 output?

m-mcgowan:
The lack of crc32 should have caused the build to fail.  That's a good suggestion to validate the crc32 output. There are lots of things I would like to improve in the build (including checking prerequisites), but not using GNUMake, which limits us greatly. 

For updates to this PR,  I can pull from a branch on your repo when needed or accept a PR depending upon the nature of the change.  I'll open a new issue for Rust support so our development convo can go there, so we don't clutter up the list of commits here. 

Do you have any thoughts on how we can provide multiple source file projects? I briefly skimmed the docs and it looks like we have to get into modules and crates, but I had to punt on that for the PoC. I'd like to move the "scaffolding" from the app so that it's just setup/loop and the necessary imports.

creationix:
This is great!  I also vote for keeping Rust conventions.  Anyone going through the trouble to learn rust won't mind translating APIs names to be rust style.

dbrgn:
Here are some improvements, mostly style:

https://github.com/spark/firmware/compare/feature/rust...dbrgn:rust

> Do you have any thoughts on how we can provide multiple source file projects? I briefly skimmed the docs and it looks like we have to get into modules and crates, but I had to punt on that for the PoC. I'd like to move the "scaffolding" from the app so that it's just setup/loop and the necessary imports.

I'm currently playing around with simple multi-file crates, see https://github.com/dbrgn/particle-firmware/commit/b3d2daef986d1ee77bf86bc164e161f764911836. Compiling this works if calling `rustc` directly:

```
cd user/src
ln -s ../../build/arm/rust/thumbv7m-none-eabi.json
rustc --target thumbv7m-none-eabi -O -g -Z no-landing-pads -L ../../build/arm/rust --emit obj -o out application.rs
```

...but when running it through Make, the build script tries to compile `gpio.rs` separately, which fails because it's not the crate root.

In the end, we could get this to work, but would still be kind of limited to single crate applications. The better solution would be to somehow integrate Cargo into this building process, and let Cargo handle all the building. But I'm not sure how well it already supports adding compiler flags.

Another option would be to build libraries into `.rlib` files and then link to these from the main application `rustc` invocation. [This](http://www.hashmismatch.net/2015/05/18/pragmatic-bare-metal-rust.html) might be useful.

LeDominik:
This is very very cool. Something for the weekend :+1:

khyperia:
This is very cool! I'm very excited about using Rust!

@m-mcgowan @dbrgn I ran into the crc32 issue as well (Arch Linux). I believe the failure was not detected because bash (and likely other shells) do not have "-o pipefail" set by default (this means that if a subcommand in a pipe fails, for example due to a missing binary, it's undetectable). The crc32 invocation is indeed in a pipe, and I think it can be "fixed" by a few different ways (that I know of, there may be others):

- Prepending `set -o pipefail;` before every pipe invocation in the build system (on the same Makefile line as the pipe)
- Adding `export SHELLOPTS=pipefail` to some global makefile (e.g. the makefile entry point)
- Attempting to execute the potentially missing binary in some validation step
- Executing something like `which crc32` and checking the exit code (nonzero = not found)
- Telling the user to carefully scroll through output to see the message `/bin/sh: line 6: crc32: command not found` :)

I'm not sure how portable each of those are, though. (and a new issue about this may be helpful, as this isn't about Rust)

eddyb:
> compared to the more conventional Rust `pin_mode(Pin.D7, PinState.HIGH);`

Did you mean `pin_mode(Pin::D7, PinMode::Output);`? Assuming `Pin` and `PinMode` are enums, and `D7` and `High` are variants.

But the *actual* idiomatic Rust convention would be `D7.mode(Output)`.
`D7` would be either `Pin::D7` (enum variant) or `pin::D7` (constant).

I think I would prefer constants for pins, especially if never really makes sense to `match` over pins.

It's also really easy to alias pins then, e.g. `use spark::pin::{D6 as LED1, D7 as LED2};`

dbrgn:
> But the actual idiomatic Rust convention would be D7.mode(Output). D7 would be either Pin::D7 (enum variant) or pin::D7 (constant).

It depends on whether we're talking about a pin representation or a pin specification. If it's a *representation*, then yes, it might make sense that a pin might set its own mode and output.

If we want to have pin *specifications*, then it would probably make more sense that the `Board` (or something like that) can set the output for the pin specified by the pin specification.

eddyb:
@dbrgn Fair enough. I would actually be glad to have some kind of zero-sized argument to `setup` and `loopy` (or even move the code calling those two functions to Rust entirely), as the only way to control the hardware (instead of global functions).

user/src/application.rs
```diff
@@ -0,0 +1,84 @@
...
+ 
+ #[no_mangle] 
+ pub fn setup()
+ {
+ 	pinMode(D7, OUTPUT);
+ }
+
+ #[no_mangle] 
+ #[export_name = "loop"]
+ pub fn loopy()
```
eddyb: These two functions should be `extern fn` btw (which is short for `extern "C" fn`).

zsup:
Should we adopt any concepts or even the framework developed with Zinc?

https://zinc.rs/
https://github.com/hackndev/zinc

dbrgn:
Note https://github.com/hackndev/zinc/issues/284

tjpeden:
@zsup I really love your enthusiasm for this!

I'm going to get this branch running on my Ubuntu partition today and start taking a look at it.

zsup:
@dbrgn good call, looks like maybe that project isn't in a great place.

zsup:
If you're interested in contributing, go to our forums and mention it in the post I wrote on "Rust on Particle" so we can get you some free test hardware!

http://community.particle.io/t/rust-on-particle-call-for-contributors/19090/4

zsup:
FYI, new place for development and discussion of Rust on Particle:

https://github.com/zsup/firmware-rust/

https://github.com/zsup/firmware-rust/issues

technobly:
Closing due to new development location mentioned in previous post.

PR #808, opened 2016-01-14
A test for SPI DMA issue #791 fixed in PR #801

avtolstoy:
This test checks the issue #791 of SPI User callback firing before the actual SPI transfer has finished.

### DIFF
--- a/user/tests/wiring/spi_shield/spi_fram.cpp
+++ b/user/tests/wiring/spi_shield/spi_fram.cpp
@@ -328,3 +328,126 @@ test(SPI_Test3_MODE3_MSBFIRST_ReadWriteSucceedsWithoutUserIntervention) {
     // Then - SPI bus is disabled
     assertFalse(SPI.isEnabled());
 }
+
+static void SPI_Test4_Prepare_SPI() {
+    // When - Setting up the SPI bus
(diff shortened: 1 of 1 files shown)


m-mcgowan:
@technobly  would you mind testing this - I don't have the required hardware.

technobly:
@avtolstoy I've tested this on a Photon and Electron with the FRAM Shield and all 4 tests pass
```
make clean all PLATFORM_ID=6 TEST=wiring/spi_shield COMPILE_LTO=n -s program-dfu
```
However, I don't see the SPI user callback worked into test 4.  Was this meant to be included or is the fix from #791 tested by some other method?

avtolstoy:
@technobly It should pass with develop branch, however it should fail on test 4 with v0.4.8 for example. Basically the test disables SPI right after the user callback is called and forcefully pulls MISO line low. The result is that if the callback is called before the transfer is completed, the last bytes will not be received at all or received with errors (since we pull MISO low). I've confirmed that this works in loopback mode and in theory should work with actual hardware as well.

technobly:
Tested with v0.4.7 on Photon, this test fails #4 with FRAM shield.  Also this PR with Photon and Electron both pass all tests in FRAM shield.

PR #810, opened 2016-01-14
initialize wifi on the core when determining if credentials should be cleared

m-mcgowan:
fix for #782 on the Core


m-mcgowan:
Tested on the Core. Without this, booting the core causes a hardfault. Tested by flashing as current and factory backup firmware then performing a factory reset. The WiFi credentials were successfully cleared.

PR #824, opened 2016-01-18
Feature/criticalsection

m-mcgowan:
SingleThreadeSection and AtomicSection RAII classes.


user/tests/wiring/no_fixture/interrupts.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+	uint32_t start_micros = micros();
+	uint32_t start_millis, end_millis;
+	{
+		ATOMIC_SECTION();
+		start_millis = millis();
+		while (micros()-start_micros<2000)
+		{
+			// validates that atomic sections can be nested and interrupts restored when the outermost one exits
+			ATOMIC_SECTION();
```
technobly: Won't interrupts be restored while the `micros()-start_micros<2000` condition is being evaluated? I would think millis() would advance a bit each time, and if your test was longer than 2ms you might start to accumulate enough of a millis() count to fail the test.

m-mcgowan: The atomic section will disable interrupts until the end of the section (line 22), so millis() will not increment at all.  The block of code after tries it without atomic and millis() continues incrementing.

technobly: I'm thinking `HAL_enable_irq(prev);` will be executed in the while() expression test, but now that I look a bit closer since `prev` was created inside the while() loop it will already have the interrupts masked, therefor it will not re-enable them by accident.

PR #835, opened 2016-01-24
Feature/mutexes

m-mcgowan:
Addresses issue #669.  Multiple threads accessing serial produce jumbled output that is an interleaving of the output of both threads. 

This PR adds a system mutex for USB serial so that access to this resource can be serialized when used concurrently. 

To lock USB serial for a given code block, such as a function:

```
void print_status()
{
    WITH_LOCK(Serial);
    Serial.println("current status is:");
    Serial.println("happy");
}
```
Even when other threads attempt to access serial, their output will be blocked while this function is printing, so the status information will always appear intact.  Node that this method will block if another thread has acquired the lock.

To try to lock the resource, and conditionally execute code on successful locking:

```
void print_status()
{
    TRY_LOCK(Serial);
    if (IS_LOCKED(Serial)) {
       Serial.println("current status is:");
       Serial.println("happy");
   |
}
```

The try lock attempts to lock the resource, but doesn't block if the resource is already locked. 

cc: @jenesaisdiq

### DIFF
--- a/hal/inc/hal_dynalib_concurrent.h
+++ b/hal/inc/hal_dynalib_concurrent.h
@@ -40,6 +40,19 @@ DYNALIB_FN(hal_concurrent,os_timer_create)
 DYNALIB_FN(hal_concurrent,os_timer_destroy)
 DYNALIB_FN(hal_concurrent,os_timer_get_id)
 DYNALIB_FN(hal_concurrent,os_timer_change)
+
+DYNALIB_FN(hal_concurrent,os_mutex_create)
+DYNALIB_FN(hal_concurrent,os_mutex_destroy)
+DYNALIB_FN(hal_concurrent,os_mutex_lock)
+DYNALIB_FN(hal_concurrent,os_mutex_trylock)
+DYNALIB_FN(hal_concurrent,os_mutex_unlock)
+
+DYNALIB_FN(hal_concurrent,os_mutex_recursive_create)
+DYNALIB_FN(hal_concurrent,os_mutex_recursive_destroy)
+DYNALIB_FN(hal_concurrent,os_mutex_recursive_lock)
+DYNALIB_FN(hal_concurrent,os_mutex_recursive_trylock)
+DYNALIB_FN(hal_concurrent,os_mutex_recursive_unlock)
+
 #endif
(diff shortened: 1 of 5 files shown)


jenesaisdiq:
Super cool, @m-mcgowan! How does the resource become unlocked?

m-mcgowan:
Thanks! It's automatically unlocked at the end of the code block.

towynlin:
Looks great to me! When writing the documentation, we need to call out

- that the mutex will be scoped to a code block, giving examples, both functions and internal simple blocks if that's valid, and explaining that you don't want to lock your entire `loop`, and emphasizing that in fact you should lock the absolutely minimum amount of code — don't assume people know this :smile_cat: 
- a list of the resources you can pass to `WITH_LOCK`, `TRY_LOCK`, AND `IS_LOCKED`

technobly:
Agree with Doc's points.. looks good to me..  UPPERCASE is good :)  

It also looks like with TRY_LOCK(), the onus is on the user to test the LOCKED status with IS_LOCKED() and decide what to do thereafter, correct?  Would it make sense to have TRY_LOCK() also return the same status as IS_LOCKED() for simpler code?
```
void print_status()
{
    if (TRY_LOCK(Serial)) { // returns true if lock acquired, false if not
       Serial.print("current status is: ");
       Serial.println("happy");
    }
}
```

m-mcgowan:
I really wanted `TRY_LOCK()` to be a conditional expression itself, but couldn't find a way to do that, given that the macro expands to a declaration. 

We could do it if we drop the `if (...)`

```
TRY_LOCK(Serial) {
    ...
}
``

I kind of thought that was not so good, since it's missing the usual if(), but seeing it written down it's not so bad. 

To make it intuitive, the scope of the lock should be the block that comes after the declaration. (Which we can do by making the macro itself expand a new block.)   Contrast that with having the TRY_LOCK() scope in the containing block.

We could then unify the syntax also with the `WITH_LOCK()` too and have that apply to the following block.

```
WITH_LOCK(Serial) {
   Serial.println();
}
```

I like that because it makes it explicit what the scope of the lock is, rather than implicitly binding to the containing scope.

m-mcgowan:
I'll create a new PR with the block scope changes.

PR #836, opened 2016-01-25
adds random_seed_from_cloud() for the Core/Photon/Electron with test.

m-mcgowan:


m-mcgowan:
Tested on the Photon and the Core. (Electron runs the same code as the Photon.)

PR #838, opened 2016-01-25
locking API redux

m-mcgowan:
Implements

```
WITH_LOCK(Serial) {

}
```

and

```
TRY_LOCK(Serial) {

}
```

and

```
SINGLE_THREADED_BLOCK() {
   // single threaded code
}
```

and

```
ATOMIC_BLOCK() {

}
````

Tested on the Photon with an app that writes to serial during listening mode.

### DIFF
--- a/system/src/system_setup.cpp
+++ b/system/src/system_setup.cpp
@@ -70,12 +70,13 @@ template <typename Config> SystemSetupConsole<Config>::SystemSetupConsole(Config
 
 template<typename Config> void SystemSetupConsole<Config>::loop(void)
 {
-	WITH_LOCK(serial);
-	if (serial.available()) {
-        int c = serial.read();
-        if (c>=0)
-            handle((char)c);
-    }
+	TRY_LOCK(serial) {
+		if (serial.available()) {
+			int c = serial.read();
+			if (c>=0)
+				handle((char)c);
+		}
+	}
 }
 
 template<typename Config> void SystemSetupConsole<Config>::handle(char c)
--- a/user/tests/wiring/no_fixture/interrupts.cpp
+++ b/user/tests/wiring/no_fixture/interrupts.cpp
@@ -11,14 +11,15 @@ test(interrupts_atomic_section)
 	uint32_t start_micros = micros();
 	uint32_t start_millis, end_millis;
 	{
-		ATOMIC_SECTION();
-		start_millis = millis();
-		while (micros()-start_micros<2000)
-		{
-			// validates that atomic sections can be nested and interrupts restored when the outermost one exits
-			ATOMIC_SECTION();
+		ATOMIC_BLOCK() {
+			start_millis = millis();
+			while (micros()-start_micros<2000)
+			{
+				// validates that atomic sections can be nested and interrupts restored when the outermost one exits
+				ATOMIC_BLOCK();
+			}
+			end_millis = millis();
 		}
-		end_millis = millis();
 	}
 	assertEqual(start_millis, end_millis);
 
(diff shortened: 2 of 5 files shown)


andyw-lala:
Has the behaviour with return or goto inside the block been verified ?

Arguments about goto being evil > /dev/null, someone will do it.

On Mon, Jan 25, 2016 at 5:22 PM, Matthew McGowan <notifications@github.com>
wrote:

> Implements
>
> WITH_LOCK(Serial) {
>
> }
>
> and
>
> TRY_LOCK(Serial) {
>
> }
>
> and
>
> SINGLE_THREADED_BLOCK() {
>    // single threaded code
> }
>
> and
>
> ATOMIC_BLOCK() {
>
> }
>
> Tested on the Photon with an app that writes to serial during listening
> mode.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/spark/firmware/pull/838
> Commit Summary
>
>    - implements block constructs where the scope of the block is the one
>    that immediately comes after the block declaration.
>
> File Changes
>
>    - *M* system/src/system_setup.cpp
>    <https://github.com/spark/firmware/pull/838/files#diff-0> (13)
>    - *M* user/tests/wiring/no_fixture/interrupts.cpp
>    <https://github.com/spark/firmware/pull/838/files#diff-1> (15)
>    - *M* user/tests/wiring/no_fixture/thread.cpp
>    <https://github.com/spark/firmware/pull/838/files#diff-2> (28)
>    - *M* wiring/inc/spark_wiring_interrupts.h
>    <https://github.com/spark/firmware/pull/838/files#diff-3> (2)
>    - *M* wiring/inc/spark_wiring_thread.h
>    <https://github.com/spark/firmware/pull/838/files#diff-4> (12)
>
> Patch Links:
>
>    - https://github.com/spark/firmware/pull/838.patch
>    - https://github.com/spark/firmware/pull/838.diff
>
> —
> <https://github.com/spark/firmware/pull/838>.
>



-- 
Andy

m-mcgowan:
Any exiting of the block will cause the lock to be released. That's a guarantee provided by C++.

andyw-lala:
I understand what's supposed to happen, I was wondering of those two cases
had been explicitly tested, that's all.

I've been burned by something similar in the past.

I was going to ask about longjmp() as well, but I have no idea if that even
exists in the libc being used.

On Mon, Jan 25, 2016 at 5:30 PM, Matthew McGowan <notifications@github.com>
wrote:

> Any exiting of the block will cause the lock to be released. That's a
> guarantee provided by C++.
>
> —
> <https://github.com/spark/firmware/pull/838#issuecomment-174729200>.
>



-- 
Andy

technobly:
OMG goto's are EVIL!  ahaha..   j/k ;) goto's are pretty useful in the right context.  I doubt Mat tested goto's, but since they can be used only within a single function, if the jump is within the locked block of code (not likely, but maybe) it would remain locked.  If outside that block of code, the mutex will be unlocked.  longjmp's will clearly be out of scope if jumping to another function. Returns are clearly out of scope as well, and the mutex will be unlocked.

m-mcgowan:
The lock is released with `goto`, return, and, if they were enabled, exceptions. Tested with 'goto'. 

`setjmp`/`longjmp` will jump out _without_ releasing the lock, because automatically scoped objects are not destructed.

http://www.cplusplus.com/reference/csetjmp/longjmp/

If `goto` is evil, then `setjmp`/`longjmp` is the distilled vehemence of all daemons in the Realm of Hades. ;-)

technobly:
Ahh, good to know.  I must admit I've never used `longjmp` in practice so I originally missed that point about non-trivial destructors.

PR #839, opened 2016-01-25
Servo HAL: Do not disable timer if some of its channels are still in use

### LINKED ISSUE #745: Possible flaw in `Servo.detach()` and other HW timer dependent APIs
Since several servo pins share one hardware timer, I'd see some problem in this implementation of `Servo.detach()`

```cpp
void HAL_Servo_Detach(uint16_t pin)
{
    // TIM disable counter
    STM32_Pin_Info* PIN_MAP = HAL_Pin_Map();
    TIM_Cmd(PIN_MAP[pin].timer_peripheral, DISABLE);
}
```

If there are still other servos depending on this timer, instead of disabling the `timer_peripheral` rather only mask the `timer_ch`.
Only the last "user" of a specific timer should actually be allowed to disable the timer.

But this might call for a centralized way to request access to a HW timer and release it with some kind of reference counter.
  
The OP of following thread also reports issues with re-attaching detached servos
 https://community.particle.io/t/interrupt-and-servos-can-they-work-together/17706

sergeuz:
This PR fixes #745 by tracking what timer channels are used to control attached servos.

Note that hardware timers are shared between different HAL submodules, namely, servo, tone and PWM HALs, and any of these still can override timer settings configured by another submodule. Ideally, we need to report an error when user application attempts to reuse already active timer with different settings.

### DIFF
--- a/hal/src/core/servo_hal.c
+++ b/hal/src/core/servo_hal.c
@@ -112,8 +112,32 @@ void HAL_Servo_Attach(uint16_t pin)
 
 void HAL_Servo_Detach(uint16_t pin)
 {
-  // TIM disable counter
-  TIM_Cmd(PIN_MAP[pin].timer_peripheral, DISABLE);
+  const STM32_Pin_Info* pin_info = HAL_Pin_Map() + pin;
+
+  // Disable timer's channel
+  switch (pin_info->timer_ch)
+  {
+    case TIM_Channel_1:
+      pin_info->timer_peripheral->CCER &= ~TIM_CCER_CC1E;
+      break;
+    case TIM_Channel_2:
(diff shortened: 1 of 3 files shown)


sergeuz:
I've updated this PR to make timer state data stored separately from pins table. The code compiles well for the Core platform, but I'm unable to test it there (wiring test is included).

sergeuz:
Simplified this patch by removing previously introduced additional storage for pin state data.

m-mcgowan:
Even better! :+1:

m-mcgowan:
Ran `wiring/no_fixture` tests on the Core - they passed. (There was a unrelated WiFi failure, and I had to remove EEPROM tests to make it fit, in the process discovered that arm gcc 5.2.1 breaks USB serial on the Core - probably an incorrect optimization, or buggy code.)

PR #850, opened 2016-01-29
Cellular.localIP() API

technobly:
### Example Code
```c++
Serial.println("Get the local (private) IP address for the Electron");
Serial.println(Cellular.localIP());
```


m-mcgowan:
I reworked the WLanConfig so we have distinct config objects per network type. Untested - will test before merging.

m-mcgowan:
The IP addresses I get from a few tries is `10.41.5.183` and `10.41.250.26` which seem correct. (I initially wondered if the 183 should have been the first octet, but with the 2nd address I see that we are in a private network `10.x.x.x`.)

PR #856, opened 2016-02-01
I2C fixes

### LINKED ISSUE #854: I2C does not insure a stop condition completes correctly in endTransmission
At about line 560 of hal/src/stm32f2xx/i2c_hal.c a stop condition is issued but the code does not wait for the stop condition to complete.

This can cause a failure in the next start/send address operation.  The failure depends on the time between the endTransmission() call and the next requestFrom() call.

If the calls are back-to-back, they complete correctly if there is sufficient delay.  This is why adding  SINGLE_THREADED_SECTION() to Wire fixed problems like this issue: https://github.com/spark/firmware/issues/698.

It appears that  SINGLE_THREADED_SECTION()  may add enough delay.

If there is a long delay, the stop condition completes OK and there is no problem.

If there is some delay, the result of the requestFrom() can be a failure with just a start condition followed by a stop condition.

The above issue occurs when these calls are made in a tight loop in the MCP23017 library.

```
Wire.endTransmission();
Wire.requestFrom(MCP23017_ADDRESS | i2caddr, 1);
```
I removed SINGLE_THREADED_SECTION() from spark_wiring_i2c.cpp and ran a test program. It fails in a matter of seconds.

I connected a digital scope and caught the failure
scope_7
The transfers up to the gap in SCL is a successful call to endTransmission()/requestFrom(). It is followed by a successful call to endTransmission() followed by a call to requestFrom() that is only a start condition followed immediately by a stop condition.

After looking at the code in i2c_hal.c, I found that there is a problem at about line 560 where endTransmission() does not wait for the stop condition to complete.

I did a quick hack with this fix from "BEGIN WHG ADD" to "END WHG ADD".
```
/* Send STOP Condition */
if(stop == true)
{
      /* Send STOP condition */
      I2C_GenerateSTOP(i2cMap[i2c]->I2C_Peripheral, ENABLE);
      /* BEGIN WHG ADD  */
      int timeout = 0X1000;
      while (i2cMap[i2c]->I2C_Peripheral->CR1 & I2C_CR1_STOP) {
        if (--timeout <= 0) return 6;
      }
      /*  END WHG ADD */
}
```
I tried the test program again without SINGLE_THREADED_SECTION() and it ran for hours with no failures, more than 40 million times through the loop. This is thousands of times longer than the unmodified code.

The bug is still there even though SINGLE_THREADED_SECTION() reduces its impact for some applications. I was able make it happen even with SINGLE_THREADED_SECTION() at a slower clock.

I have an example that works at 400 kHz but fails at 100 kHz with  SINGLE_THREADED_SECTION().

avtolstoy:
Fixes #854


tlangmo:
I tried the changed file but it actually breaks the i2c communication. I use an MPU9250 Invensense chip on develop branch. 
I commented the changes out one by one, and everything went back to normal when 
line 426  `while(i2cMap[i2c]->I2C_Peripheral->CR1 & I2C_CR1_STOP || I2C_GetFlagStatus(i2cMap[i2c]->I2C_Peripheral, I2C_FLAG_BUSY))`
stays as it was.
Interestingly, this is the only instance where the I2C_GenerateSTOP() call is actually conditional (on the last byte only). Can you please double check that? 

I am excited about this fix (I had a lot of instabilities with I2C) but maybe in that one code instance it breaks things?

avtolstoy:
@tlangmo Thanks for reporting this! I'm still testing this and you are correct, that place looks a bit strange. There is no point in waiting for `I2C_CR1_STOP` to clear after every data byte and waiting on `I2C_FLAG_BUSY` will definitely cause a timeout.

technobly:
The reason we initially started waiting for the STOP bit to be sent, is that a user app may try to start a new transmission before the STOP bit has a change to actually send.  This resulted in I2C slaves not responding anymore IIRC.

Maybe you meant to use `&& busy` instead of `|| busy` in the while() loop?  However it sounds like to fix #854 we only need to add `while(i2cMap[i2c]->I2C_Peripheral->CR1 & I2C_CR1_STOP)` in endTransmission() as you've done.

avtolstoy:
Updated PR.

Apart from fixing #854 by waiting for I2C_CR1_STOP to clear in HAL_I2C_End_Transmission, PR also includes:
- Rewrite of HAL_I2C_Request_Data basing on ST AN2824
- Fix for transactions with repeated START condition between them
- Removed SingleThreadedSection from TwoWire
- Tests for consecutive transactions with various combinations of STOP/repeated START conditions inbetween

avtolstoy:
@technobly 
STM32F205 reference manual states the same thing as AN2824:
>For 2-byte reception:
>• Wait until ADDR = 1 (SCL stretched low until the ADDR flag is cleared)
>• Set ACK low, set POS high
>• Clear ADDR flag

tlangmo:
I am not sure how helpful this is, but I tested the new i2c_hal.c and noticed that I now have  a lot of intermittent 'stalls' when I use i2c in a thread. The stalls are likely related to the EVENT_TIMEOUT, being ~100ms. 
The old code did not have that problem. I already suspend the scheduler for i2c calls, so I have no clue why multithreading would have an impact. There are also no other calls which use i2c, so no race condition. Please let me know if I can do a more specific test...

avtolstoy:
@tlangmo Thanks for testing this. Could you provide a code sample so I can try to replicate this issue?

avtolstoy:
Updated PR. 
- Apparently N > 2 Master receiver sequence is a bit different for STM32F2. 15816d4 should fix the issue with I2C when multi-threading is enabled.

PR #864, opened 2016-02-05
Release/v0.5 beta

m-mcgowan:
Combine all the PRs for 0.5 release


m-mcgowan:
If we can make `cancel()` interrupt only the currently running modem operation rather than all future operations then that would address all our needs I think here. We can continue to have a `cancel everything from here on out` mode too if you feel we need that, but am I'm not 100% sure it's needed.

m-mcgowan:
0.5.x PRs now being merged into develop

PR #867, opened 2016-02-07
WIP/POC System.commit() API

technobly:
This is a WIP, proof of concept.

Had an idea this morning for how to identify exactly which git commit the system firmware was built from.

This could be also returned in Listening Mode after pressing `g` or `c`

Currently it only identifies the commit the system parts were built from, and technically it's only one of the system parts where the object lives.  It might be more useful to save this information for each module, and reply with all of them with the `s` command in Listening Mode.  If all modules had their own git commit, a System.commit() might return an object with 4 strings, one for each module (bootloader, system part1, system part2 and user).

One can make an argument that all 40 chars of the commit SHA are not necessary... to future proof it though, I think at least 12 should be included.  But why not include all 40 if this is a valuable piece of indentifying information.

Thoughts, comments?

```
Serial.printlnf(System.version());
Serial.printlnf(System.commit());
// OUTPUT
// 0.4.9
// a3b44179f7f24ebfb34af08acfb0217f7d0cce11
```


technobly:
Closing due to no interest.

PR #872, opened 2016-02-10
Proof-of-concept to allow delay(1.5s) in user code

monkbroc:
C++11 adds user-defined literals and the C++14 standard library defines types for microseconds, milliseconds, seconds, hours. See http://en.cppreference.com/w/cpp/chrono/duration

It would be cool to allow users to use units in their time literals: `delay(1.5s)` or `System.sleep(SLEEP_MODE_DEEP, 1h);`

This is a proof of concept to modify `delay` to accept durations. It requires compiling the user code in C++14 mode.

### DIFF
--- a/user/applications/time/application.cpp
+++ b/user/applications/time/application.cpp
@@ -1,0 +1,17 @@
+#include "application.h"
+
+void setup()
+{
+    pinMode(D7, OUTPUT);
+}
+
+void loop()
+{
+    static bool state = false;
+
+    // Use literals like 150ms, 1.5s or 1h in delay
+    delay(1.5s);
+
+    digitalWrite(D7, state);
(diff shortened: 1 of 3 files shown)


m-mcgowan:
This is very slick, and fits in well with our plans for better type safety with the API.  :+1:

m-mcgowan:
I'm not sure when we can add C++14 support - I imagine when we can move to the 5.x series. Is it possible to back-port the parts of the c++14 library that we need into our firmware?

monkbroc:
Back port is not possible from what I can tell. This library relies on compile-time parsing that was added in C++14.

technobly:
Closing due to uncertainty around when we can add C++14 support.

PR #876, opened 2016-02-13
Log categories and per-category log level filtering

sergeuz:
This PR is currently work in progress and added to proceed with the discussion started here: https://github.com/spark/firmware/issues/739

---

- [x] Review code
- [x] Test on device
- [ ] Add documentation https://github.com/spark/docs/issues/362
- [ ] Add to CHANGELOG.md


m-mcgowan:
Thanks! Looks good from a quick glance - I'll review in detail next week.  One question comes to mind - when debug logs are compiled in, and no formatter/outputter set, is this short circuited to be fast, avoiding any category scanning and message formatting?

Some items of code should be conditional on logging being enabled for a given category/level. Perhaps we have a function for callers to determine if a given log category/level is enabled in cases where they might log lots of info, but want to skip if the logs would not be output.

services/inc/services_dynalib.h
```diff
@@ -49,6 +49,7 @@ DYNALIB_FN(services,LED_RGB_SetChangeHandler)
 DYNALIB_FN(services, log_print_direct_)
 DYNALIB_FN(services, LED_GetColor)
 //DYNALIB_FN(services, log_level_active)
```
m-mcgowan: The `log_level_active()` is something we'd tried adding but strangely caused the system to crash, so it was pulled (we had no time to investigate.)

Perhaps you could sketch out how we might allow log client code to determine if logging at a given level with a given category would produce any log output? This will allow client code to short circuit any expensive logging operations.

m-mcgowan:
The test cases are beautiful!

services/inc/services_dynalib.h
sergeuz: Sure, it's quite easy to implement, I just need to add another callback to application level that will allow to determine logging level for particular category.

sergeuz:
Guys, I've just pushed backward compatible logger implementation. I couldn't resist introducing a bit nicer system API though :) As usual, any feedback is welcomed!

m-mcgowan:
Please rebase against develop. 

For testing I will check existing logging on the electron continues to work with just system-part1 upgraded.

sergeuz:
Do you want `log_print_()`, `log_print_direct_()` and other existent functions to be redirected to newer logging backend as well? I wasn't sure about this and left implementation of these function intact, so if you rebuild only system-part1, all logging from other modules will remain directed to older `debug_output_` callback.

We have two options here:

1. Invoke newer callbacks in older logging functions that don't support categories (`log_print_()`, etc.).

2. Remove older logging functions entirely replacing their entries in dynalib table with placeholders.

m-mcgowan:
The ideal would be to redirect to new logging if that can be made to work, but equally I'm happy to leave as is if it's risky/difficult.  Redirecting to the new implementation might reduce the code size as well as prove the new implementation?

sergeuz:
It's not risky and quite easy to implement. The only drawbacks are unnecessary grow in code size and empty category names that will be generated by modules that are "linked" to older logging functions.

sergeuz:
Rebased this PR on top of current develop. I also checked Photon and Electron with only system modules updated (system-part1 and system-part2) and with user-part built against current develop that still uses `SerialDebugOutput` with older `debug_output_` callback. Both platforms generate expected debugging output.

Going to investigate why CI builds are failing for this branch. Sometimes CI fails due to boost errors in gcc hal, sometimes the build timeouts silently :(

sergeuz:
Also, older `log_print_()`, `log_print_direct_()` and some other functions are no-op now. It's fine as long as both system parts are updated during firmware update.

m-mcgowan:
Could you update just system part1 and not part2 (or update the system parts via OTA rather than DFU.)
I'm not expecting logging to work during the update, but want to be sure the system doesn't crash. (I guess it won't since the old logging APIs are preserved, but worth checking all the same.)

sergeuz:
The system seems to work normally after updating only system-part1: both Photon and Electron are able to connect to the cloud. In this case I can see only debugging output that comes from system-part1 modules.

I think I need some help with figuring out what's wrong with CI builds, as `run_tests.sh` script completes successfully with both gcc 4.x and 5.x in my local environment. I don't have any substantial experience with Travis, but is it possible that some container was cached in some inconsistent state (e.g. partial boost download), causing broken builds over and over for this particular branch?

hal/src/gcc/core_hal.cpp
```diff
@@ -36,33 +36,61 @@
...
+        const char *func, void *reserved)
 {
-    setLoggerLevel(NO_LOG_LEVEL);
-    if (read_device_config(argc, argv)) {
-        app_setup_and_loop();
+    if (level < log_level) {
+        return;
     }
-    return 0;
+    std::ostringstream strm;
```
m-mcgowan: do we need to use stringstream here vs using sprintf()? Using sprintf() would keep the code size to a minimum.

user/tests/unit/makefile
```diff
@@ -43,15 +43,19 @@ CPPSRC += $(call target_files,$(WIRING_SRC),spark_wiring_random.cpp)
 CPPSRC += $(call target_files,$(WIRING_SRC),spark_wiring_string.cpp)
 CPPSRC += $(call target_files,$(WIRING_SRC),spark_wiring_ipaddress.cpp)
 CPPSRC += $(call target_files,$(WIRING_SRC),spark_wiring_print.cpp)
+CPPSRC += $(call target_files,$(WIRING_SRC),spark_wiring_logging.cpp)
 CPPSRC += $(call target_files,$(WIRING_SRC),string_convert.cpp)
 CPPSRC += $(call target_files,$(SYSTEM)src/,system_utilities.cpp)
 CPPSRC += $(call target_files,$(SYSTEM)src/,system_mode.cpp)
 CPPSRC += $(call target_files,$(SYSTEM)src/,system_string_interpolate.cpp)
+CPPSRC += $(call target_files,$(HAL)src/gcc,timer_hal.cpp)
```
m-mcgowan: This is what is causing the build to fail. It fails for me on linux too, since the unit tests do not depend on boost.

wiring/inc/spark_wiring_logging.h
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SPARK_WIRING_LOGGING_H
+#define SPARK_WIRING_LOGGING_H
+
+#include <initializer_list>
+#include <vector>
```
m-mcgowan: I'm unsure if we can safely use vector since memory allocation failures do not throw an exception in the embedded environment. I'm also concerned about code bloat - the STL classes can add quite a bit of overhead. Is there a simpler solution?

hal/src/gcc/core_hal.cpp
sergeuz: Sure, I can replace C++ streams with series of `printf()` calls here. I just thought we're not that concerned about code size and memory allocation on gcc platform.

m-mcgowan: Oh, my bad - you're totally right - I thought this was in main code. It's fine since it's the gcc build.

wiring/inc/spark_wiring_logging.h
sergeuz: You're right, another issue is heap fragmentation caused by `std::vector` when it grows. The problem is that I need to use some container that is able to grow dynamically during construction of the category mappings. I used `std::vector` mostly as a time-saver here, as all current loggers are intended to be instantiated globally before the application starts.

user/tests/unit/makefile
sergeuz: Thanks! Looks like I have boost headers in my default include paths locally. I can either enable boost for unit tests or provide stubs for hal functions that are required by the logger.

wiring/inc/spark_wiring_logging.h
sergeuz: Definitely it's better to refactor this code, i.e. for cloud-enabled logging, which can be enabled at any time. Direct alternative is to use some simple wrapper over `realloc` instead of `std::vector`.

sergeuz:
ooops, seems I've merged EEPROM changes once again

m-mcgowan:
@monkbroc Please review!

sergeuz:
Rebased on top of current `develop`

services/inc/service_debug.h
```diff
@@ -16,121 +16,53 @@
...
+void log_print_(int level, int line, const char *func, const char *file, const char *msg, ...);
+void log_direct_(const char* s);
+void log_print_direct_(int level, void* reserved, const char *msg, ...);
+int log_level_active(LoggerOutputLevel level, void* reserved);
 
 #ifdef __cplusplus
 }
 #endif
 
+#define DEBUG(fmt, ...) LOG(DEBUG, fmt, ##__VA_ARGS__)
```
monkbroc: Let's add a comment that these macros are deprecated. This is the compatibility layer.

sergeuz:
Here is list of things I prefer to get done **before** merging this PR:

- Rename `Logger` class into `LogHandler` so that former name could be used for new application-level API. The API itself will be done as a separate PR.
- Refactor `LogHandler` to take reference to `Stream` object and use it as a destination for formatted logging output. This will allow to split formatting and backend logging functionality into separate classes and improve overall flexibility.

services/inc/service_debug.h
sergeuz: There is a comment that says the whole header file is now deprecated :) https://github.com/sergeuz/firmware/blob/logging/services/inc/service_debug.h#L23

sergeuz:
Guys, no more commits are pending, so let's do some final testing / code review and merge this PR if everything looks good.

Key things to double-check (feel free to add more):

- [ ] Unit test succeeds (`user/tests/unit/logging.cpp`)
- [ ] Firmware builds and runs normally on all platforms (I have only Photon and Electron). Logging on gcc platform should work normally as well
- [ ] Firmware runs normally with only `system-part1` updated (in this case no logging output will be generated by `system-part2`)

Unit tests cover most of the usage scenarios, but feel free to do some manual testing if necessary, e.g.:

- Compatibility loggers continue to work normally (`SerialDebugOutput`, `Serial1DebugOutput`)
- New log handlers work as expected (`SerialLogHandler`, `Serial1LogHandler`)
- Logging macros work as expected (see `services/inc/logging.h` for internal API docs)
- Category overriding and filtering works as expected in both C and C++ code (module vs file vs scope categories)

monkbroc:
I'm starting to review this PR.

services/src/debug.c
```diff
@@ -5,90 +5,42 @@
...
+    }
+    log_compat_level = level;
 }
 
-void log_print_direct_(int level, void* reserved, const char *msg, ...)
-{
-    if (level<log_level_at_run_time || !debug_output_)
-        return;
+void log_print_(int level, int line, const char *func, const char *file, const char *msg, ...) {
+    // Not implemented, use newer log_message() function instead
```
monkbroc: // Deprecated, use newer log_message() function instead
    // This function may not be removed because it is exported in dynalib

services/src/debug.c
```diff
@@ -5,90 +5,42 @@
...
-    va_start(args, msg);
-    int trunc = vsnprintf(_buffer, arraySize(_buffer), msg, args);
-	debug_output_(_buffer);
-	if (trunc > arraySize(_buffer))
-	{
-		debug_output_("...");
-	}
-	va_end(args);
+void log_print_direct_(int level, void* reserved, const char *msg, ...) {
+    // Not implemented, use newer log_printf() function instead
```
monkbroc: // Deprecated, use newer log_printf() function instead
    // This function may not be removed because it is exported in dynalib

services/src/debug.c
```diff
@@ -5,90 +5,42 @@
...
 }
 
 void log_direct_(const char* s) {
```
monkbroc: Remove from code since it's not used anywhere

services/src/debug.c
```diff
@@ -5,90 +5,42 @@
...
-        return;
-
-	debug_output_(s);
+    // Not implemented, use newer log_write() function instead
 }
 
-int log_level_active(LoggerOutputLevel level, void* reserved)
-{
-	return (level>=log_level_at_run_time && debug_output_);
+int log_level_active(LoggerOutputLevel level, void* reserved) {
```
monkbroc: Remove from code since it's not used anywhere

services/inc/logging.h
```diff
@@ -0,0 +1,369 @@
...
+#endif
+
+typedef enum LogLevel {
+    LOG_LEVEL_ALL = 1, // Log all messages
+    LOG_LEVEL_TRACE = 1,
+    LOG_LEVEL_INFO = 30,
+    LOG_LEVEL_WARN = 40,
+    LOG_LEVEL_ERROR = 50,
+    LOG_LEVEL_PANIC = 60,
+    LOG_LEVEL_NONE = 70, // Do not log any messages
```
monkbroc: There's an implicit dependency between the log level integer definitions and log_level_name. Maybe add a note to review log_level_name if this enum is changed.

services/src/logging.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+    char buf[LOG_MAX_STRING_LENGTH];
+    if (log_msg_callback) {
+        const int n = vsnprintf(buf, sizeof(buf), fmt, args);
+        if (n > (int)sizeof(buf) - 1) {
+            buf[sizeof(buf) - 2] = '~';
+        }
+        log_msg_callback(buf, level, category, time, file, line, func, 0);
+    } else {
+        // Using compatibility callback
```
monkbroc: What is the compatibility callback for? In which cases is the log_msg_callback null but log_compat_callback is set?

services/inc/logging.h
```diff
@@ -0,0 +1,369 @@
...
+    LOG_LEVEL = LOG_LEVEL_TRACE,
+    DEBUG_LEVEL = LOG_LEVEL_TRACE,
+    INFO_LEVEL = LOG_LEVEL_INFO,
+    WARN_LEVEL = LOG_LEVEL_WARN,
+    ERROR_LEVEL = LOG_LEVEL_ERROR,
+    PANIC_LEVEL = LOG_LEVEL_PANIC,
+    NO_LOG_LEVEL = LOG_LEVEL_NONE
+} LogLevel;
+
+// Logger callbacks
```
monkbroc: Could you add a description of the link between the logger callbacks and the log handler/log manager? It's not obvious why logging would need to callback to something else.

log_set_callbacks is the only link between spark_wiring_logging that lives in the user module and logging that lives in the HAL. From looking at both modules it seems the callbacks all define sinks for the logs, right?

services/src/logging.cpp
sergeuz: It's used for compatibility with existent applications that were built against previous version of the firmware and use `SerialDebugOutput` for logging

monkbroc:
Could you also update docs/debugging.md with instructions on how to get logs printed on serial?

services/src/debug.c
```diff
@@ -5,90 +5,42 @@
...
-	}
-    trunc = vsnprintf(_buffer,arraySize(_buffer), msg, args);
-	debug_output_(_buffer);
-	if (trunc > arraySize(_buffer))
-	{
-		debug_output_("...");
-	}
-	debug_output_("\r\n");
-	va_end(args);
+void set_logger_output(debug_output_fn output, LoggerOutputLevel level) {
```
monkbroc: Could you add your comment from below:

> This function remains for compatibility with existent applications that were built against previous version of the firmware and use SerialDebugOutput for logging

services/inc/logging.h
sergeuz: The system allows to register one callback per logging output type: `log_msg_callback` for messages and `log_write_callback` for stream-based (or direct) logging. Application module's `LogManager` then dispatches that output to multiple sinks implemented as  `LogHandler` objects.

monkbroc: I get that, but it took a while to understand that link. A source file comment on the role of the callbacks and who is expected to register them would help make that link more explicit.

wiring/src/spark_wiring_logging.cpp
```diff
@@ -0,0 +1,246 @@
...
+spark::LogHandler::LogHandler(Stream *stream, LogLevel level, const Filters &filters) :
+        stream_(stream),
+        level_(level) {
+    for (auto it = filters.begin(); it != filters.end(); ++it) {
+        const char* const category = it->first;
+        const LogLevel level = it->second;
+        std::vector<FilterData> *filters = &filters_; // Root categories
+        size_t pos = 0;
+        for (size_t i = 0;; ++i) {
+            if (category[i] && category[i] != '.') { // Category separator
```
monkbroc: Can you make the category separator a class constant instead of a literal?

wiring/src/spark_wiring_logging.cpp
```diff
@@ -0,0 +1,246 @@
...
+    std::vector<FilterData> filters; // Subcategories
+
+    FilterData(const char *name, size_t size) :
+            name(name),
+            size(size),
+            level(-1) {
+    }
+};
+
+spark::LogHandler::LogHandler(Stream *stream, LogLevel level, const Filters &filters) :
```
monkbroc: I'd really like to see this constructor refactored into several smaller private methods with descriptive names. It's really dense algorithmic code and it's hard to guess what each piece is trying to accomplish.

wiring/src/spark_wiring_logging.cpp
```diff
@@ -0,0 +1,246 @@
...
+            if (category[i] && category[i] != '.') { // Category separator
+                continue;
+            }
+            const size_t size = i - pos;
+            if (!size) {
+                break; // Invalid category name
+            }
+            const char* const name = category + pos;
+            bool found = false;
+            auto it = std::lower_bound(filters->begin(), filters->end(), std::make_pair(name, size),
```
monkbroc: This seems like a good candidate for a method extraction: `bool doStuff(const Filters *filters, const std::pair<const char*, size_t> &value) {...}` where `doStuff` is an appropriate name for the stuff this block does.

wiring/src/spark_wiring_logging.cpp
```diff
@@ -0,0 +1,246 @@
...
+        if (category[i] && category[i] != '.') { // Category separator
+            continue;
+        }
+        const size_t size = i - pos;
+        if (!size) {
+            break; // Invalid category name
+        }
+        const char* const name = category + pos;
+        bool found = false;
+        auto it = std::lower_bound(filters->begin(), filters->end(), std::make_pair(name, size),
```
monkbroc: Reuse method extracted from constructor.

user/tests/unit/makefile
```diff
@@ -65,7 +69,19 @@ INCLUDE_DIRS += $(HAL)inc
 INCLUDE_DIRS += $(COMMUNICATION)src
 INCLUDE_DIRS += dynalib/inc
 
-CFLAGS += $(patsubst %,-I$(SRC_ROOT)%,$(INCLUDE_DIRS)) -I.
+# prefix $(SRC_ROOT)
+ABS_INCLUDE_DIRS += $(patsubst %,$(SRC_ROOT)/%,$(INCLUDE_DIRS)) 
+
+
+ifeq ("$(BOOST_ROOT)","")
+$(error BOOST_ROOT not defined)
```
monkbroc: The unit tests no longer run because of this. I suppose I need to install Boost. Can you tell me how you set it up so I don't end up with a different setup than yours?

m-mcgowan:
run `./ci/install_boost.sh` from the root

monkbroc:
Running `./ci/install_boost.sh` then running `make` for the unit tests causes a bunch of deprecation errors in Boost. The unit tests don't compile.

```
In file included from /home/monkbroc/.ci/boost/boost_1_59_0/boost/shared_ptr.hpp:17:0,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/gregorian/greg_month.hpp:14,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/gregorian/greg_ymd.hpp:16,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/gregorian/greg_calendar.hpp:16,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/gregorian/gregorian_types.hpp:19,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/posix_time/posix_time_config.hpp:18,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/posix_time/posix_time_system.hpp:13,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/posix_time/ptime.hpp:12,
                 from /home/monkbroc/.ci/boost/boost_1_59_0/boost/date_time/posix_time/posix_time.hpp:15,
                 from ../../../hal/src/gcc/timer_hal.cpp:4:
/home/monkbroc/.ci/boost/boost_1_59_0/boost/smart_ptr/shared_ptr.hpp: In member function ‘boost::shared_ptr<T>& boost::shared_ptr<T>::operator=(std::auto_ptr<_Up>&&)’:
/home/monkbroc/.ci/boost/boost_1_59_0/boost/smart_ptr/shared_ptr.hpp:549:38: error: ‘template<class> class std::auto_ptr’ is deprecated [-Werror=deprecated-declarations]
         this_type( static_cast< std::auto_ptr<Y> && >( r ) ).swap( *this );
```

monkbroc:
There isn't a LOG_MODULE_CATEGORY set in the wiring module.

Let's add
```
# wiring/src/build.mk
LOG_MODULE_CATEGORY = wiring
```

monkbroc:
Awesome work @sergeuz! I'm done with my review. I'll give you time to look at my comments. My strongest feedback would be to break up the complex `spark::LogHandler::LogHandler` methods.

I ran the unit tests. I wrote a little test application using the logger with handlers and filtering. Worked great!

Let me know @sergeuz when you've addressed the comments you think need to be addressed. I'll take a quick look to the changes and this will be good to merge.

monkbroc:
Last thing, do you mind if I run `git clang-format` before merging?

wiring/src/spark_wiring_logging.cpp
sergeuz: Despite looking similarly, the code in `categoryLevel()` uses const iterators while constructor deals with non-const ones. This makes refactoring that you're mentioning a bit tricky. I suggest to leave this part of the code as it is for now, as it's going to be refactored for cloud-enabled logging anyway (it's better to get rid of vectors in case of dynamically allocated loggers).

sergeuz:
Hey @monkbroc, thanks for the review! I'll add source code comments and docs that you mentioned.

wiring/src/spark_wiring_logging.cpp
monkbroc: Could you put a few comments that say what different blocks of code are doing? I'd still favor extracting a few methods with descriptive names over comments but it's your call.

sergeuz:
Added comments and other stuff.

I would leave `docs/debugging.md` as it is for now. The new code doesn't break anything from what is described there, and majority of the information about newer logging (usage examples, etc) is going to be covered in the firmware reference. Once we'll have all the pieces we'll replace whole "Logging" section in that file with couple of links.

PR #880, opened 2016-02-19
Add System.benchmark to measure durations in microseconds

monkbroc:
I found myself writing code over and over to read System.ticks() before and after a task to benchmark it so I suggest adding `System.benchmark` to do just that.

Example usage:

```
uint32_t duration = System.benchmark([] {
  logFile.flush();
});
Serial.printlnf("SD card flush took %d us", duration);
```

### DIFF
--- a/user/tests/wiring/api/platform.cpp
+++ b/user/tests/wiring/api/platform.cpp
@@ -24,5 +24,9 @@ test(system_ticks)
     API_COMPILE(value=System.ticks());
     API_COMPILE(value=System.ticksPerMicrosecond());
     API_COMPILE(System.ticksDelay(30));
+    API_COMPILE(value = System.benchmark([] {
+      for(volatile int i = 0; i < 10000; i++) {}
+    }));
+
     (void)value;
 }
--- a/user/tests/wiring/api/system.cpp
+++ b/user/tests/wiring/api/system.cpp
@@ -126,6 +126,7 @@ void handler_event_data_param(system_event_t event, int data, void* param)
 test(system_events)
 {
 	int clicks = system_button_clicks(123);
+	(void)clicks;
 
 	system_event_t my_events =
 			wifi_listen_begin+wifi_listen_end+wifi_listen_update+
--- a/wiring/inc/spark_wiring_system.h
+++ b/wiring/inc/spark_wiring_system.h
@@ -87,6 +87,13 @@ public:
         uint32_t start = ticks();
         while ((ticks()-start)<duration) {}
     }
+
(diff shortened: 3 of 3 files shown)


m-mcgowan:
I like this, but I'm wondering if it would be better placed in a benchmarking library?

I can imagine wanting to print benchmark results to serial, running the benchmark code multiple times and taking the mean/median, having checkpoints in application code that can form the basis of benchmarking.  

Checkpoints have been discussed between @andyw-lala  and me as a way to trace application execution to help diagnose application behavior prior to a crash. 

Adding this one function to the System library does no harm, but if we find we need more benchmarking/checkpoint support in future then it might make sense to break it out into a new library. 

@technobly @jenesaisdiq  any thoughts here?

monkbroc:
This is meant more as a simple tool to help users learn how long their own code take to execute, especially when they use libraries. The example above is a real one. I added an SD card library and wondered how long it took to append a few lines to a file. I was surprised to find it took ~150ms! That's really all I needed to know.

A full benchmark library is not a priority in my opinion.

The checkpoint feature would serve a different purpose of much deeper debugging/tracing. It will also be longer than the 5 lines of code of this implementation :smiley:

technobly:
@monkbroc @m-mcgowan I do like this addition and it won't add much code at all.  I'd like to see the option of using millis() or micros() as the timer as well.  So maybe:

```c++
uint32_t System.benchmarkTicks()
uint32_t System.benchmarkMillis()
uint32_t System.benchmarkMicros()
```

I would also want to be able to pass in a C function or Class function instead of only a lambda, so we should add those tests to `wiring/api` and we can easily add tests for `wiring/no_fixture` just to ensure times reported stay consistent with that expected.

jenesaisdiq:
Going to second @monkbroc on this one- if it does what you need, great. Benchmarking is def interesting, but not a priority. Now, you guys know how much I want checkpoints/breakpoints and stepping through live code on hardware... but sadly, even that wouldn't be top of the stack.

monkbroc:
So I thought I'd use this code to make a Benchmark library and use it to test the happy path of contributing a new lib. I'm writing notes in the libs feedback doc.

monkbroc:
https://github.com/monkbroc/particle-benchmark

I'm OK closing this PR

technobly:
Excellent!  Don't forget to publish your Library :)

monkbroc:
It's published as a V2 lib.

```
$ particle library search benchmark -v
Searching for libraries matching benchmark
> Found 1 library matching benchmark
> benchmark@1.0.0 - A simple code execution time library
```

Issue #814, opened 2016-01-15
SPI Slave Support in Firmware


harrisonhjones:
Posting here for a customer. Would like to add firmware support for using the Photon/Spark Core as a SPI slave. 

This [gist](https://gist.github.com/chrismeyersfsu/3317769) might help. It's written for the arduino. 

This [page](http://www.gammon.com.au/spi) has a bit more in-depth information. 

Feel free to discuss this feature here.

AronRubin:
https://github.com/btmcmahan/Teensy-3.0-SPI-Master---Slave

dseemakurty:
I would find this extremely helpful also!

m-mcgowan:
@avtolstoy  Please add tests to verify that SPI.transfer() with a `nullptr` callback does block until the transfer is complete.

avtolstoy:
#919 fixes SPI.transfer() behavior and adds necessary tests.

PR #882, opened 2016-02-21
SPI Slave implementation for Photon and Electron (STM32F2)

avtolstoy:
Implements #814 on Photon and Electron

Wiring SPI API changes:
- `void SPIClass::begin(SPI_Mode mode, uint16_t);`
- `void SPIClass::onSelect(wiring_spi_select_callback_t user_callback);`
- `void SPIClass::transferCancel();`
- `int32_t SPIClass::available();`

Basic usage example:
https://gist.github.com/avtolstoy/1bbd34a7e69f563d5d37

There are also two test applications in `user/tests/wiring/spi_master_slave` `spi_master` and `spi_slave` that showcase usage.

A couple of notes:
- `transfer()` can be safely called before a Slave Select event occurs
- `transfer()` can be safely provided with larger buffer size (transfer length). When Master deselects the Slave or application calls `transferCancel()` SPI peripheral will be disabled, transfer user callback will still be called to indicate end of DMA transfer, application can then check the actual transfer length by calling `SPIClass::available()`
- Application can also track the state of pending transfer by calling `SPIClass::available()`, which will report actual number of bytes clocked in/out.
- `byte SPIClass::transfer(byte)` is not supported in Slave mode


hal/inc/spi_hal.h
```diff
@@ -84,6 +91,9 @@ void HAL_SPI_DMA_Transfer(HAL_SPI_Interface spi, void* tx_buffer, void* rx_buffe
 bool HAL_SPI_Is_Enabled_Old();
 bool HAL_SPI_Is_Enabled(HAL_SPI_Interface spi);
 void HAL_SPI_Info(HAL_SPI_Interface spi, hal_spi_info_t* info, void* reserved);
+void HAL_SPI_Set_Callback_On_Select(HAL_SPI_Interface spi, HAL_SPI_Select_UserCallback cb);
```
m-mcgowan: Please add void* as a reserved parameter to these so we can extend later.

hal/src/stm32f2xx/spi_hal.c
```diff
@@ -147,7 +159,7 @@ static void HAL_SPI_DMA_Config(HAL_SPI_Interface spi, void* tx_buffer, void* rx_
     }
     else
     {
-        DMA_InitStructure.DMA_Memory0BaseAddr = (uint32_t)(&tempMemory);
+        DMA_InitStructure.DMA_Memory0BaseAddr = (uint32_t)(&tempMemoryTx);
```
m-mcgowan: omg! Nice catch. How did this ever work before? using a stack variable as the DMA address!

m-mcgowan:
Could you please rebase on develop and resolve the conflicts.

avtolstoy:
@m-mcgowan Done!

rpbissonnette:
Hello.

My naive attempts at using the Photon SPI as a slave seem to stall after the first transfer.  I see in the wiring tests I ought to be able to learn from that example.  Using this invocation:

`cd main
make  -d clean program-dfu PARTICLE_DEVELOP=1 PLATFORM=photon USE_SWD_JTAG=y MODULAR=y USE_SPI=SPI USE_CS=A2 TEST=wiring/spi_master_slave`

My "USE_SPI=SPI USE_CS=A2 " didn't go where they should because it fails:

> tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp:11:2: error: #error Define USE_SPI
>  #error Define USE_SPI

When I just edit the two test programs to force in the defines:
`
#define USE_SPI SPI
#define USE_CS  A2

#ifndef USE_SPI
#error Define USE_SPI
#endif

#ifndef USE_CS
#error Define USE_CS
#endif`

The sources compile without error - but the link phase fails with a multiple define problem:
`../../../build/target/user/platform-6-m/tests/wiring/spi_master_slave//libuser.a(application.o):(.bss.SystemMode+0x0): multiple definition of `SystemMode'
../../../build/target/user/platform-6-m/tests/wiring/spi_master_slave//libuser.a(application.o):(.bss.SystemMode+0x0): first defined here
`
Could this because the Master/Slave are both trying to use the same libuser.a?

It should not matter, but I am using SuSE LEAP and my make is
GNU Make 4.0
which is breaking the build/test/build.bats expectation of 3.8.1

If anyone can help me get this test running I would be happy to add a setup readme to the directory.

I THINK this ought to be a very useful example as there seems to be more than enough confusion about using the Photo in slave mode.

Thanks.

PR #883, opened 2016-02-21
Give a compiler error for long names in Particle.variable/function

monkbroc:
A common error is to call `Particle.function("mylongfunction", ...)` with a name longer than 12 characters and overlook to check the return value. It will compile and silently ignore the call at run time.

This PR adds a compile-time check for the length of the name parameter of `Particle.function()` and `Particle.variable()` when the name is a static string literal (type `char [N]`). If the parameter is a dynamic string (type `char *`) or `String` then no check will be done at compile time.


m-mcgowan:
Thanks!  This looks good. It's been on my backlog for a while so very happy to see this resolved!

I'll build a small test framework for capturing compiler errors so that we can test expected compilation failures and the messages produced.

dmiddlecamp:
just got bit by this again today, :+1: to a fix on the horizon :)

PR #884, opened 2016-02-22
Increased stack size and stack overflow detection for Electron

### LINKED ISSUE #821: System threading on Electron sometimes crashes
System threading on the electron fails just after the cloud connection has been established. 

I've investigated briefly, changing thread stack sizes to rule out stack overflow. The error happens in the FreeRTOS queue code, so looks like one of the thread queues is being corrupted.

I'll follow up with steps to reproduce the issue.

sergeuz:
This PR fixes #821 by increasing stack size for threads created via `std::thread` wrapper and adds FreeRTOS hook for stack overflow detection -- latter is enabled only for debug builds on Electron currently (thanks to @avtolstoy for pointing me out).

### DIFF
--- a/system/src/system_threading.cpp
+++ b/system/src/system_threading.cpp
@@ -7,6 +7,15 @@
 
 
 #if PLATFORM_THREADING
+
+#if HAL_PLATFORM_CLOUD_UDP
+// Electron uses larger stack size to workaround stack overflow problem that occurs during
+// handshake in multithreaded configuration
+#define THREAD_STACK_SIZE 4*1024
+#else
+#define THREAD_STACK_SIZE 3*1024
+#endif
+
 void system_thread_idle()
 {
     Spark_Idle_Events(true);
@@ -16,7 +25,7 @@ ActiveObjectThreadQueue SystemThread(ActiveObjectConfiguration(system_thread_idl
 			100, /* take timeout */
 			0x7FFFFFFF, /* put timeout - wait forever */
 			50, /* queue size */
-			3*1024 /* stack size */));
+			THREAD_STACK_SIZE /* stack size */)); // TODO: Use this value for threads spawned by ActiveObjectBase
 
 /**
  * Implementation to support gthread's concurrency primitives.
@@ -94,7 +103,7 @@ namespace std {
         thread_startup startup;
         startup.call = base.get();
         startup.started = false;
-        if (os_thread_create(&_M_id._M_thread, "std::thread", OS_THREAD_PRIORITY_DEFAULT, invoke_thread, &startup, 1024*4)) {
+        if (os_thread_create(&_M_id._M_thread, "std::thread", OS_THREAD_PRIORITY_DEFAULT, invoke_thread, &startup, THREAD_STACK_SIZE)) {
             PANIC(AssertionFailure, "%s %s", __FILE__, __LINE__);
         }
         else {  // C++ ensure the thread has started execution, as required by the standard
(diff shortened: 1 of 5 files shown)


system/src/system_threading.cpp
```diff
@@ -94,7 +94,7 @@ namespace std {
         thread_startup startup;
         startup.call = base.get();
         startup.started = false;
-        if (os_thread_create(&_M_id._M_thread, "std::thread", OS_THREAD_PRIORITY_DEFAULT, invoke_thread, &startup, 1024*3)) {
+        if (os_thread_create(&_M_id._M_thread, "std::thread", OS_THREAD_PRIORITY_DEFAULT, invoke_thread, &startup, 1024*4)) {
```
m-mcgowan: This will increase the stack size on the Photon too. I think that until we know the size increase is needed there, that this should be conditional. You can check `HAL_PLATFORM_CLOUD_UDP` which will be 1 when UDP is enabled. 

I'd like to know what is requiring so much stack size in the UDP handshake, and I'm curious why we get corruption since the heap grows from the bottom of memory while the stack grows from the top. Ah...but the system thread stack is allocated on the heap so that's why any overflow corrupts the heap.

m-mcgowan:
Nice work, particularly enabling the stack overflow detection and glad you narrowed it down to the overflow. I did test overflow, but that was before I fixed the multithreading allocation problem, and didn't get around to retesting with a larger stack with that fix in place.

At some point it would be nice to know what is taking up so much stack (are there profilers that can tell is this?). In the meantime, bumping it up 1K for UDP comms is fine.  :+1:

sergeuz:
I'm not aware of any suitable profilers that we can use on microcontrollers, but we can try to use function call instrumentation provided by gcc -- it allows to specify couple of functions that should be automatically called after entering and before leaving of any other function. Within of those "injected" functions we can dump stack trace and stack usage, for example.

sergeuz:
Btw, I'd like to enable similar stack overflow detection on Photon, but WICED already has `vApplicationStackOverflowHook()` defined and exposes `wiced_rtos_check_stack()` function instead, which is quite useless as it needs to be called periodically. What is the best way to override `vApplicationStackOverflowHook()` with our own implementation in this case? Change link order?

sergeuz:
Added similar FreeRTOS hook to Photon. Increased stack size is used only if `HAL_PLATFORM_CLOUD_UDP` is defined.

m-mcgowan:
Just doing a final check here, looks good. The electron firmware is built with debug enabled by default - will there be any performance or other impact from enabling stack overflow detection?  (I'm asking since I don't know how it's implemented and don't know what any affects of having this enabled would be.)

sergeuz:
Yes, there's an impact definitely. The check is performed before the context switches to another task:
> Stack overflow checking introduces a context switch overhead so its use is only recommended during the development or testing phases (http://www.freertos.org/Stacks-and-stack-overflow-checking.html).

Note that Photon's WICED is also built with stack overflow detection enabled.

m-mcgowan:
Ah yes, I forgot it was in WICED by default, good call! :+1:  That makes me feel better about including it.

sergeuz:
Also, overflow detection requires threading enabled at runtime -- otherwise everything will be running in the main thread without FreeRTOS involved. Correct me if I'm wrong here, but this check doesn't seem to affect much of our users at the moment.

m-mcgowan:
That's correct. Apart from software timers, the default single threaded application mode doesn't use threads. But threading will become the default for the 1.0 release so then it will be used a lot!

PR #889, opened 2016-02-26
Cellular: Add basic support for ublox File System AT-commands

schneider-iosphere:
Export file read/write/del functions from `mdm_hal` through the `cellular` dynalib, as the AT commands are quite cumbersome to work with (especially `+UDWNFILE` with its prompt)

---
- [ ] Check CLA for signature
- [ ] Review code
- [ ] Test on device
- [ ] Add documentation
- [ ] Add to CHANGELOG.md

### DIFF
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -182,8 +182,8 @@ int cellular_file_write(const char* filename, const char* buf, int len, void* re
  * Read a file from the cellular module
  *
  * @param filename: the filename of the file to be downloaded.
- *        buf: buffer into which the file is written
- *        len: lnumber of bytes that should be read. 
+ *        buf: buffer into which the file is written.
+ *        len: number of bytes that should be read.
  *        reserved: pass NULL. Allows future expansion.
  *
  * @returns number of bytes read
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -158,9 +158,49 @@ cellular_result_t cellular_signal(CellularSignalHal &signal, void* reserved);
 
 /**
  * Send an AT command and wait for response, optionally specify a callback function to parse the results
+ *
+ * @note maxmimal length of the resulting command is currently 256 bytes
  */
 cellular_result_t cellular_command(_CALLBACKPTR_MDM cb, void* param,
                          system_tick_t timeout_ms, const char* format, ...);
+
+/** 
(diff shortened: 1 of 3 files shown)


schneider-iosphere:
@m-mcgowan: is there any interest in this? If so, anything else I should add (Wiring, test coverage, documentation)? I intentionally left out the wiring parts, as I think the File API is a somewhat niche feature.

technobly:
@schneider-iosphere sorry for the delay!  I wanted to ask what your use case is with exposing these functions?  Are you using the Particle SIM or 3rd party SIM?  Have you experimented with these functions with the local build already?

Also I don't know why it's not spelled out clearly as a feature of the module, but I searched all of the datasheets, system integration manual and AT command manual for what this memory is; RAM or NVM (NOR)?  It appears to be on-module memory. Sara-G350 = 1MB and Sara-U260/270 = 5MB.

We'll have a discussion about adding this and what things might look like.  At first glance it looks like a useful feature, but appears to need more error checking added and possibly create an object type for the `filename, buf and len` that can be passed to the HAL.  Something like CellularFile perhaps.

schneider-iosphere:
@technobly no worries. You have guessed correctly, I'm using a 3rd party SIM. It is working locally and I'm using it to do HTTPS POST requests via the ublox AT commands, which might not be a core priority for you :wink:. Apart from that it is only used for managing certificates (AFAIK useful for TLS only) doing fw updates, FTP and playing audio.

Regarding the storage type: it is NVM, and I think the limits you mentioned are correct. The whole feature is only mentioned in the AT-Commands Manual ([UBX-13002752](https://www.u-blox.com/sites/default/files/u-blox-ATCommands_Manual_%28UBX-13002752%29.pdf)). You can see supported file sizes in chapter 21.7.

I will add proper input checking, I somehow (wrongfully, despite reading it) assumed that it is handled in `MDMParser`. By exposing this, you can avoid increasing the tx buffer size, as you circumvent the `printf`you normal would call through cellular_command. On the way back I'm currently using `+URDBLOCK` to achieve that. 

On `CellularFile`: I can add something here, but do you want to expose this to the general population? Apart from HTTP/FTP there is probably no sense in exposing it. And even then it probably might sense to expose an HTTP API. As I'm writing this, I think it might make more sense to make an AT-interface that works with user supplied buffers.

technobly:
Added to Clubhouse 1665.

PR #898, opened 2016-03-10
wifi connect timeout

m-mcgowan:
adds a 60 second timeout to wlan_connect() to attempt to regain control after a runaway wifi connect

### DIFF
--- a/hal/src/photon/wlan_hal.cpp
+++ b/hal/src/photon/wlan_hal.cpp
@@ -37,6 +37,7 @@
 #include "wwd_sdpcm.h"
 #include "delay_hal.h"
 #include "dct_hal.h"
+#include "concurrent_hal.h"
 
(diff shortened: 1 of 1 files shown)


m-mcgowan:
@monkbroc Please review.  I have not been able to test the case when wifi takes longer than 60 seconds to connect, but we do know that the `wlan_cancel_connect()` call is good - we can emulate the same by pressing the setup button for 3s while wifi is connecting.

markterrill:
Does this mean it will go to listening mode?

I'd previously setup an ISR timer to alternate (poorly) between listen and connect mode and timeout if either hadn't succeeded within a given timeframe. IE like https://community.particle.io/t/listen-network-wont-shut-down-in-semi-automatic/18449/3

I suppose the question is whether either connecting or listening mode will still stick.

PR #905, opened 2016-03-14
don't block application when OTA happens and updates are disabled. 

m-mcgowan:
extend system event notification to allow a data argument to be passed to the func…tion. This is used to stall the system thread waiting for the application (but not indefinitely, as that would create a potential deadlock.)  The system thread only blocks until the application event has been handled. Previously it would block until system updates were enabled, for a maximum of 30 seconds.

https://community.particle.io/t/dashboard-disableupdates-causes-user-code-to-stop-running/19731/7

### DIFF
--- a/system/inc/system_event.h
+++ b/system/inc/system_event.h
@@ -113,4 +113,4 @@ void system_unsubscribe_event(system_event_t events, system_event_handler_t* han
  * @param data
  * @param pointer
  */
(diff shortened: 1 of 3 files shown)


m-mcgowan:
The test for this is to setup single-threaded blinky with a call to `System.disableUpdates()` in the setup method, and then attempt to send an update via OTA.  Before this fix, the LED would hang. Now the LED continues blinking. 

@sergeuz Please review and merge if it checks out!

Issue #894, opened 2016-03-07
Sleep lockup issue


jamlamberti:
The photon board will lock up if a sleep occurs in a loop and there is not enough time to re-establish the wifi connection. This prevents over-the-air programming with the IDE and requires normal flashing. I would be nice to prevent this or at least warn the user. 

To replicate (tested on latest and develop branches):

```Arduino
int test = D2;
int blink = D7;
int nope=D1;

void setup() {
    pinMode(test,INPUT);
    pinMode(blink,OUTPUT);
    pinMode(nope,OUTPUT);
    
    digitalWrite(blink,HIGH);
    digitalWrite(nope,HIGH);
    delay (1000);
    digitalWrite(blink,LOW);
    digitalWrite(nope,LOW);
}

void loop() {
    digitalWrite(blink,HIGH);
    delay(2000);
    if (digitalRead(test)==HIGH){
        digitalWrite(nope,HIGH);
        System.sleep(10);
    }
    digitalWrite(blink,LOW);
    digitalWrite(nope,LOW);
}
```

Even if the delay is increased to 10 seconds, the device will still lock up. Delaying 30 seconds gives the device enough time to re-connect.

m-mcgowan:
I'm not sure what we can do about this without interfering with application behavior. 

A workaround is to ensure the app doesn't put the device to sleep until the device has connected to the cloud, e.g.

```
static bool sleep_when_connected = false;
if (digitalRead(test)==HIGH){
        digitalWrite(nope,HIGH);
        sleep_when_connected = true;
}
if (sleep_when_connected && Particle.connected()) {
   sleep_when_connected = false; 
   System.sleep(10);
}
```

jamlamberti:
I agree, but it might be worth having a warning appear if someone tries to verify their code. 

Thanks for the suggestion on checking connection status!

m-mcgowan:
I'm afraid that kind of code introspection isn't possible in the general case (it would requite the machine to predict the result of your code, which is often dependent on external factors, such as a button press.) It's not possible to know that the code will not be able to re-establish the wifi connection, or specifically that this is undesirable. In other cases, the application code may truly want to sleep for 10 seconds.

Another alternative is to explicitly connect to the cloud on waking up again:

```
System.sleep(10);
Particle.connect();
waitFor(Particle.connected, 60000);
```

[...]

PR #909, opened 2016-03-15
cloud sleep - waits for all outgoing confirmable UDP messages to be sent

m-mcgowan:
protocol sleep/wake commands which allow the caller to wait for all coap confirmable requests to be acknowledged, and allows the comms layer to take action when waking up after sleep.

This is used as part of system sleep to ensure confirmable messages are confirmed before entering sleep.

### DIFF
--- a/communication/src/build.mk
+++ b/communication/src/build.mk
@@ -22,6 +22,7 @@ CPPSRC += $(TARGET_SRC_PATH)/eckeygen.cpp
 CPPSRC += $(TARGET_SRC_PATH)/lightssl_message_channel.cpp
 CPPSRC += $(TARGET_SRC_PATH)/dtls_message_channel.cpp
 CPPSRC += $(TARGET_SRC_PATH)/dtls_protocol.cpp
+CPPSRC += $(TARGET_SRC_PATH)/lightssl_protocol.cpp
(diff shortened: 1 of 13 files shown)


m-mcgowan:
https://github.com/spark/firmware/commit/3014dd53bee31f89222d25276504657e608b575d addresses issue #894 - the system will reconnect to the cloud before returning control back to the application loop after waking up.

m-mcgowan:
Test app:

```
void setup()
{
	Particle.publish("started","",PRIVATE);
}
int count = 0;
void loop()
{
	Serial.print("loop "); Serial.println(count);
	Particle.publish(Particle.deviceID(), String("sleeping ")+String(count),PRIVATE);
	Serial.println("sleeping.");
	Serial.flush();
	System.sleep(D3, RISING, 30);

	Particle.publish(Particle.deviceID(), String("wakeup ")+String(count),PRIVATE);
	count++;
	delay(1000);
}
```
Expected output:
loop 0
sleeping 0
... device sleeps
wakeup 0
loop 1
sleeping 1
... device sleeps
wakeup 1
```

The test can then be changed to sleep for (30*60) seconds to test long term sleep.

I have completed these tests and verified success.

technobly:
Closing so we don't try to merge, see PR #918 for replacement to this one.

PR #912, opened 2016-03-16
Feature/ymodem listen mode

m-mcgowan:
28800 baud puts the device in listening mode. https://github.com/spark/firmware/issues/865
From there, ymodem can function directly, also it's possible to reset the system into safe listening mode.

### DIFF
--- a/bootloader/src/core/bkpreg_hal.c
+++ b/bootloader/src/core/bkpreg_hal.c
@@ -1,0 +1,3 @@
+
+
+#include "../src/core/bkpreg_hal.c"
--- a/bootloader/src/core/dfu_usb.c
+++ b/bootloader/src/core/dfu_usb.c
@@ -87,35 +87,6 @@ void HAL_DFU_USB_Init(void)
     USB_Init();
 }
 
-int32_t HAL_Core_Backup_Register(uint32_t BKP_DR)
-{
-    switch(BKP_DR)
-    {
-        case BKP_DR_01: return BKP_DR1;  break;
-        case BKP_DR_10: return BKP_DR10; break;
-    }
-    return -1;
-}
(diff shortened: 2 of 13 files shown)


m-mcgowan:
The test here is to use `ymodem.py` or the particle CLI to update serial.  

`stty -f /dev/tty.usbXXX 28800` will put the device in listening mode rather than ymodem mode (so LED is blue rather than magenta) but ymodem functions as before. The client can start a ymodem transfer immediately or send the `f` command from listening mode.

monkbroc:
I'm not familiar with how system flags work but it doesn't look right that a bunch of special cases get added in system_update.cpp. Is this to be improved later?
https://github.com/spark/firmware/pull/912/files#diff-e4a74a9fe08bb33770aceac34b157e46R75
https://github.com/spark/firmware/pull/912/files#diff-e4a74a9fe08bb33770aceac34b157e46R101

m-mcgowan:
The system flags as they were are just booleans, but we can add change handler support to implement backing storage, such as writing a value to DCT or backup registers. That's what the code does. It's special cases, but only because we don't need the overhead of a generic registration mechanism - the behavior is intrinsic to the meaning of the flag.

m-mcgowan:
While waiting for the modem to connect, I connected serial at 28.8k and the device hung. I don't believe the modem hangups that we've seen are a result of #863 but of some general instability in the modem that we are only now seeing with more advanced functionality. (The 28.8k line interrupt is present in factory firmware, so that could be tested also.)

m-mcgowan:
The CLI crash happens with factory electron firmware too. I believe this is unrelated to this PR. As a test, you can use the Device Updater to flash firmware.

Issue #888, opened 2016-02-26
Electron cannot publish with longer stop mode sleep periods > 25 minutes


bprobbins:
I'm finding that electron doesn't publish as expected if sleep mode period is above about 25 minutes.
Is this due to the 23 min ping event setup? Is there a solution?
```
//develop branch with PR#845 fix
#include "application.h"

FuelGauge fuel;
char publishStr[20];
uint32_t start;

void setup() {
  pinMode(D1, INPUT_PULLDOWN);
}//setup()

void loop() {

  sprintf(publishStr, "%.2f %.2f", fuel.getVCell(), fuel.getSoC());
  Particle.publish("e1", publishStr, 60, PRIVATE);

  start = millis();
  while (millis() - start < 1000UL) {Particle.process();}

//will briefly breathe cyan upon awaken but not publish
//unless wakeup < 26 minutes
  System.sleep(D1, RISING, 30 * 60);

  start = millis();
  while (millis() - start < 1000UL) {Particle.process();}
}//loop

```

bprobbins:
And then with a slightly different approach I get a stranger still result:
```
//develop branch with PR#845 fix
#include "application.h"

FuelGauge fuel;
char publishStr[20];
uint32_t start;
int sleepInterval = 21; 

void setup() {
  pinMode(D1, INPUT_PULLDOWN);
}//setup()

void loop() {

  sleepInterval++;

  sprintf(publishStr, "%i", sleepInterval);
  Particle.publish("e1", publishStr, 60, PRIVATE);

  start = millis();
  while (millis() - start < 1000UL) {Particle.process();}

//will briefly breathe cyan upon awaken but not publish
//unless wakeup < 25 minutes
  System.sleep(D1, RISING, sleepInterval * 60);

  start = millis();
  while (millis() - start < 1000UL) {Particle.process();}
}//loop
```
Produces: (electron id obscured)
{"data":"22","ttl":"60","published_at":"2016-02-26T05:42:12.289Z","coreid":"560045000b..............","name":"e1"}
{"data":"85CC212725A957348F63008F18EB4EA3A96CF200D1F6B49DFF5C59D7AD2E74C3","ttl":"60","published_at":"2016-02-26T05:42:13.241Z","coreid":"560045000b..............","name":"spark/device/app-hash"}
{"data":"23","ttl":"60","published_at":"2016-02-26T06:04:16.563Z","coreid":"560045000b..............","name":"e1"}
[...]

PR #913, opened 2016-03-17
configurable keepalive

m-mcgowan:
adds `Particle.keepAlive(seconds)` for UDP comms platforms (Electron) to allow the application to set the frequency that ping keep-alives are sent.

Example:

```
void setup()
{
   Particle.keepAlive(60*23);  // send keep alive ping every 23 minutes
}
```

A keepalive ping uses ca. 100 bytes of data. 

--

Provides a solution for issue that users with 3rd party SIM's were seeing, certain carriers may have slightly quicker network timeouts, requiring a more frequent ping interval than every 23 minutes (which is the default). 

- [x] Review code
- [x] Test on device
- [x] Add documentation https://github.com/spark/docs/commit/df650146
- [x] Add to changelog

### DIFF
--- a/communication/src/communication_dynalib.h
+++ b/communication/src/communication_dynalib.h
@@ -63,12 +63,13 @@ DYNALIB_FN(BASE_IDX + 0, communication, spark_protocol_remove_event_handlers, vo
 #if defined(PARTICLE_PROTOCOL) && HAL_PLATFORM_CLOUD_UDP
 DYNALIB_FN(BASE_IDX + 1, communication, gen_ec_key, int(uint8_t*, size_t, int(*)(void*, uint8_t*, size_t), void*))
 DYNALIB_FN(BASE_IDX + 2, communication, extract_public_ec_key, int(uint8_t*, size_t, const uint8_t*))
-#define BASE_IDX2 (BASE_IDX+3)
+#define BASE_IDX2 (BASE_IDX + 3)
 #else
-#define BASE_IDX2 (BASE_IDX+1)
+#define BASE_IDX2 (BASE_IDX + 1)
 #endif
 
-DYNALIB_FN(BASE_IDX2 + 0, communication, spark_protocol_set_connection_property, int(ProtocolFacade*, unsigned, unsigned, void*, void*))
+DYNALIB_FN(BASE_IDX2 + 0, communication, spark_protocol_set_connection_property,
+           int(ProtocolFacade*, unsigned, unsigned, void*, void*))
 
 DYNALIB_END(communication)
 
--- a/communication/src/communication_dynalib.h
+++ b/communication/src/communication_dynalib.h
@@ -63,11 +63,17 @@ DYNALIB_FN(BASE_IDX + 0, communication, spark_protocol_remove_event_handlers, vo
 #if defined(PARTICLE_PROTOCOL) && HAL_PLATFORM_CLOUD_UDP
 DYNALIB_FN(BASE_IDX + 1, communication, gen_ec_key, int(uint8_t*, size_t, int(*)(void*, uint8_t*, size_t), void*))
 DYNALIB_FN(BASE_IDX + 2, communication, extract_public_ec_key, int(uint8_t*, size_t, const uint8_t*))
+#define BASE_IDX2 (BASE_IDX+3)
+#else
+#define BASE_IDX2 (BASE_IDX+1)
 #endif
+
+DYNALIB_FN(BASE_IDX2 + 0, communication, spark_protocol_set_connection_property, int(ProtocolFacade*, unsigned, unsigned, void*, void*))
 
 DYNALIB_END(communication)
 
 #undef BASE_IDX
+#undef BASE_IDX2
 
 #ifdef	__cplusplus
 }
--- a/communication/src/ping.h
+++ b/communication/src/ping.h
@@ -22,7 +22,13 @@ public:
 		this->ping_timeout = timeout;
 	}
 
-	void reset() {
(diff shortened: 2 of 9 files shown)


elijthomas:
If this uses 100bytes of data , that means the 1mb data cap will be used in approx 10000 keep alive calls . Which if set at 30 seconds will only be about 3 days. Is this correct ??

m-mcgowan:
1Mb isn't a hard cap, but yes, that's the ballpark for the data use.

m-mcgowan:
It is possible to reduce the data usage considerably - possibly by a factor of 5 by sending a plain UDP packet rather than an encrypted DTLS packet.  I am on vacation for 2 weeks this weekend, but will look into this when I return.

bprobbins:
Very nice! Seems to solve issue #888 (electron stop mode sleep) for me - just set Particle.keepAlive(1320);//22 minutes

technobly:
I edited the example to keep people from immediately copy/pasting such a short ping rate into their apps.  I realize this is not the Docs, but could easily make it's way there as well.

technobly:
I rebased against latest develop and forced pushed, so that we could test with NETWORK_STANDBY.

bprobbins:
In my very unsystematic testing so far, I've found that the longer the stop mode sleep interval the more frequent I have to ping to enable publishing beyond 3- 6 hrs, or so it seems,  With sleep interval of 30 mins, pings every 10 mins allows continued publishing  longer than 12 hrs. However, if I ping every 19 mins, publishing fails in about 3 hrs
With sleep interval of 2 hrs and pinging every 10 mins, the startup publish and first wakeup publish at 2 hrs succeeds but fails after that (4 hrs).
This is the test app I used:
```
#include "Particle.h"

char publishStr[20];
volatile uint32_t start;
volatile int sleepInterval = 120; // 30;

void setup() {
  pinMode(D1, INPUT_PULLDOWN);
  Particle.keepAlive(600);//10 minutes
}//setup()

void loop() {

  sprintf(publishStr, "%i", sleepInterval);
  Particle.publish("e1", publishStr, 60, PRIVATE);

  start = millis();
  while (millis() - start < 6000UL) {Particle.process();}

  System.sleep(D1, RISING, sleepInterval * 60); 

  start = millis();
  while (millis() - start < 6000UL) {Particle.process();}

}//loop

```

bprobbins:
Particle.keepAlive(300) , pinging every 5 mins did work overnight to allow publishing every 2 hrs, but that's (12 * 100 * 2) data bytes to keep it going, a little more than I'd like

technobly:
@bprobbins are you using a Particle SIM or a 3rd Party SIM?

When you sleep, no pings will occur while the processor is halted.  Pinging really should only keep the Server to Device connection active (i.e. Particle.function() calls, and Particle.subscribe() events).  When you wake up, and publish... it should always be able to get through if your cellular data connection is still active.  The data connection can timeout after about an hour (but could be longer as well), and we will be forcing it to disconnect if sleeping for longer than 1 hour in the future.  Therefore if you currently sleep for more than 1 hour, when you wake up you should disconnect and reconnect to ensure your data connection is active.  Technically if you know you are sleeping for more than 1 hour, you should make sure you turn off the modem to conserve power and network resources.  It will undoubtedly not be connected when you wake back up anyway.

```c++
bool disconnected() {
    return !Particle.connected();
}

void loop() {
    Particle.disconnect();
    waitFor(disconnected, 60000);
    Particle.connect();
    waitFor(Particle.connected, 60000);

    // Read sensors here

    // Publish events here

    // No need for delays anymore (as of 0.5.x)

    // Sleep for 2 hours with network on (don't do this...
    // but as an example, loop() will force a disconnect 
    // and reconnect upon waking back up)
    System.sleep(D1, RISING, 60*60*2, NETWORK_STANDBY);
}
```

bprobbins:
@technobly , I have definitely been on the wrong track then. So is the fact that the keepAlive frequency *seemed* to affect publish success just a strange anomaly? When I increased the keepAlive to every 5 minutes I was able to publish every 2 hrs last night without fail when in a previous test with a 10 minute ping it failed after 1 publish ( http://community.particle.io/t/electron-not-publishing-after-sleep-stop-mode/20348/49 ).  It's all over my head. Is issue #888 a whole different matter? Sorry if I've been misconstruing things!

bprobbins:
So, I finally must have invoked the right deity. My code above (but without keepAlive since as you say it's inactive) worked the past 24 hrs no problem (no change in firmware). I'm back in the right universe!

elijthomas:
How do you compile it with keepalive, I keep getting ''class CloudClass' has no member named 'keepAlive''. My Electron is running 0.5.0-rc.1. I have tried compiling using build.particle and the CLI, however build.particle says its still running 0.4.8, but if i check the device in the terminal using 'v' i get 0.5.0

m-mcgowan:
The version on the device and the version you compile with aren't the same thing. You'll need to select the 0.5.0 release against your device in the web ide. On the CLI, you pass a `--target 0.5.0-rc.1` flag.

elijthomas:
Thankyou @m-mcgowan, super fast response and greatly appriciated ! That is compiling for the correct release now. On a side note, you wouldnt know why two of my electrons show up on build.particle as photons ? can i force flash the 0.5.0-rc.1 onto them as electrons somehow ? if you can tell me that ill owe you a beer when I come to san-fran next.

m-mcgowan:
If you send me a PM or email with the device IDs of the electrons that have an identity crisis I'll get it sorted. It may take a few hours.

elijthomas:
Thanks I have PM you the device IDs

PR #917, opened 2016-03-18
Fix return value from WiFi.scan() and WiFi.getCredentials() from exce…

jamesr66a:
…eding passed in count.

Current examples in the docs for [getCredentials()](https://docs.particle.io/reference/firmware/photon/#getcredentials-) and [scan()](https://docs.particle.io/reference/firmware/photon/#scan-) will overrun the array of WiFiAccessPoint structs since the return value of these functions can exceed the passed in `result_count` parameter. The intuitive behavior (in my opinion, i.e. what I interpreted it would be from the docs before I read the HAL driver source) would be for the `scan()` and `getCredentials()` calls to limit the return value to min(actual number of APs, result_count).

### DIFF
--- a/wiring/src/spark_wiring_wifi.cpp
+++ b/wiring/src/spark_wiring_wifi.cpp
@@ -37,7 +37,7 @@ namespace spark {
     class APArrayPopulator
     {
         WiFiAccessPoint* results;
-        int count;
+
         int index;
 
         void addResult(WiFiAccessPoint* result) {
@@ -52,6 +52,7 @@ namespace spark {
             ((APArrayPopulator*)cookie)->addResult(result);
         }
 
+        int count;
     public:
         APArrayPopulator(WiFiAccessPoint* results, int size) {
             this->results = results;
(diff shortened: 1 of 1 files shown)


monkbroc:
Thanks for the contribution. I'll make a clarification to the docs too.

m-mcgowan:
I'm not sure about this - the return value indicates how many networks are available, so the caller can choose to size the array accordingly. 

So, if the return value is <= the size of the array then the array was large enough, if the value is > than the size of the array, then the caller knows that there were some networks not copied to the array.

m-mcgowan:
I can see that mixing the "how many networks are there" with "get me X networks" isn't all that comprehensible.  It might help make these more distinct if we have a no-arg version of `WiFi.scan()` that scans for networks without storing to the array, and returns the number of networks found, to give the caller some idea of how large the array should be.

jamesr66a:
The main concern was that the example code on the reference documentation would overrun the buffer and didn't give any information about the actual semantics of the return value. So long as the docs are consistent, pretty much all designs would work imo.

m-mcgowan:
yep, totes, they should be consistent. In this case, I feel the code was correct, and the documentation inconsistent with the behavior. Totally my own doing - I must have written the docs after writing the code.

PR #919, opened 2016-03-20
SPI.transfer() synchronous operation when callback is NULL

avtolstoy:
Fixes behavior of SPI.transfer with NULL callback. According to docs:
> If a user callback function is passed then it will be called after completion of the DMA transfer. This results in asynchronous filling of RX buffer after which the DMA transfer is disabled till the transfer function is called again. **If NULL is passed as a callback then the result is synchronous i.e. the function will only return once the DMA transfer is complete.**

PR also includes tests in `user/tests/wiring/spi_master_slave` for the synchronous behavior.

---

- [x] Review code
- [x] Test on device
- [ ] Add documentation
- [ ] Add to changelog


technobly:
Tested at default speeds: 15Mhz and also 1MHz and all three CI tests passed.

PR #933, opened 2016-04-01
Changed DFU baud rate from 14400 to 19200

nrobinson2000:
Linux Distributions only allow a specific range of baud rates for serial.  14400 baud is not accepted.  I propose that the baud rate to trigger DFU mode on the Photon and Electron to be 19200 baud, an acceptable rate for Linux Distributions that also works on OSX.

To trigger DFU mode on Linux, simply do:
```
stty -F /dev/ttyACM0 19200
```
Likewise on OSX, you would do:
```
stty -f /dev/cu.usbmodemXXXX 19200
```


technobly:
Duplicate of #934

PR #934, opened 2016-04-01
Changed DFU baud rate from 14400 to 19200

nrobinson2000:
Linux Distributions only allow a specific range of baud rates for serial.  14400 baud is not accepted.  I propose that the baud rate to trigger DFU mode on the Photon and Electron to be 19200 baud, an acceptable rate for Linux Distributions that also works on OSX.

To trigger DFU mode on Linux, simply do:

```
stty -F /dev/ttyACM0 19200
```

Likewise on OSX, you would do:

```
stty -f /dev/cu.usbmodemXXXX 19200
```


m-mcgowan:
Thanks for the contribution. I would prefer to leave the default as is (since changing it will break any existing solutions using that) and instead provide a way to easily change the baudrate trigger via  configuration value. 

btw, on linux, you can use Python to open the serial port at 14400 baud, so it's not impossible as is, just inconvenient.

nrobinson2000:
I have tried that.  It does not work.

technobly:
Added to backlog in Clubhouse 1650

>**As a Linux user, I would like to be able to remotely activate DFU mode on my device.**
Currently Linux does not allow 14400 baud rate change as an option. We could either add an additional baud rate such as 19200 to avoid breaking existing use cases, or solve this by making it possible to enable DFU mode via USB device descriptor on Linux. This may already be possible in device firmware, but just needs a command added to the Particle CLI for example.

nrobinson2000:
@technobly I've found a way for linux to open devices on non-standard baud rates.

Here is the topic on stackoverflow:
https://stackoverflow.com/questions/12646324/how-to-set-a-custom-baud-rate-on-linux

And here is the solution:
https://gist.github.com/sentinelt/3f1a984533556cf890d9

monkbroc:
Interesting finding.

I tried to go down this route as well. What I found is that it sometimes works and sometimes doesn't. Linux seems to remember the baud rate of a port when you reopen it. So if you opened the port and switch the baud rate to the special DFU rate, flashed some firmware then did `cat /dev/ttyACM0` it would go back to DFU mode. I tried to reopen the serial port with 9600 baud after opening it with the special baud rate to avoid the issue, but the serial port is not available in DFU mode so the reopen doesn't work.

Here's [my code for a serial port switcher in Go](https://github.com/monkbroc/serial_switcher). The goal was to build executables that would work on all OS from one code base.

Do you see the same issues with the DFU mode "sticking" with the code sample you linked above? If you want to tweak the Go code go ahead. Feel free to include it in po-util if that's useful.

nrobinson2000:
I still see the DFU mode sticking.  It's very annoying.  It goes back into DFU mode with even the slightest interaction with the device file.

However, I was able to unstick a photon from listening mode (28800) by running the sample with 9600, rebooting the photon after it entered listening mode, and then running with it again with 9600.

*I was unsuccessful in using this against DFU mode, as I kept getting `ioctl: Bad file descriptor`.  It seems like the the photon goes into DFU mode way too quickly.*

I would really like to solve this DFU sticking problem for Linux so I can integrate it into po-util, but I'm also creating another utility for a client that needs smooth Linux operation for putting a photon into dfu mode with similar reliability to how it is done on a Mac.

I've been familiar with DFU sticking since the early days of po-util and it is really starting to bug me, as it is a major roadblock for making po-util and my utility for my client, and I imagine it is a big issue for you as well.

>https://github.com/nrobinson2000/po-util/issues/7
>At this point I don't have a good solution other than moving away from the "magic baud rate as a control message" which is what we'll be doing in a future firmware.

How soon until this feature is released? :smile:

nrobinson2000:
@monkbroc I've integrated the sample into po-util so Linux can now put devices into DFU mode with 14400.  I'm still looking for a fix for the DFU sticking.

awardblvr:
OTHERS ARE WAITING AS WELL... I'm currently seeing Particle Photon (using po-util methods)  go into DFU mode each and every time I open any serial communications to the USB-connected Photon to monitor debug messages.

Looking forward to a solution.

m-mcgowan:
The long term solution is to use USB vendor commands to put the device in a given mode. These won't suffer from the stickiness or OS limitations of using given baudrates, and will work more or less independent of the mode the device is currently in.

nrobinson2000:
And how would one do this?

m-mcgowan:
We will provide a node library.  Potentially the CLI could incorporate this, as well as other development tools such as the Desktop IDE.

nrobinson2000:
Would this only work with newer firmware versions?

m-mcgowan:
Yes

nrobinson2000:
Update:
Recently I've no longer been experiencing DFU mode sticking on Linux, which is nice.

monkbroc:
That's great news! Can you provide details? Do you think it is a change in your po-util tool or a change in the Particle firmware?

nrobinson2000:
On my devices I'm using 0.6.3.

To change the baud rate on Linux I am using this tool from within po.

https://github.com/nrobinson2000/po/blob/master/depend/custom-baud/custom-baud.c

```
custom-baud /dev/ttyACM0 14400
```

Perhaps you could try `custom-baud` to see if you experience this too.

monkbroc:
Thanks. I still get the sticking behavior if I open the port normally after doing a baud rate switch:
```
./custom-baud /dev/ttyACM0 14400
cat /dev/ttyACM0
```

Doing a second switch to 9600 still opens the device in DFU that second time :weary: 
```
./custom-baud /dev/ttyACM0 14400
./custom-baud /dev/ttyACM0 9600
```

nrobinson2000:
That's really weird. Because I can do the following without any DFU sticking at all:

```
./custom-baud /dev/ttyACM0 14400
# Press reset button
./custom-baud /dev/ttyACM0 28800
particle serial wifi
```

What version of Particle firmware are you using and what does `uname -a` give you?

shivams:
@nrobinson2000 Could this be because of some upgrades in Linux Kernel? Because, my `uname -a` gives:

    Linux shivam-hp 4.4.0-130-generic #156-Ubuntu SMP Thu Jun 14 08:53:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Clearly, my kernel is older than yours. 

Also, the issue doesn't show up in my Mac OS Sierra at all.

nrobinson2000:
>Could this be because of some upgrades in Linux Kernel?
@shivams It must be.

>the issue doesn't show up in my Mac OS Sierra at all.
This was never an issue on macOS.

jtzemp:
I know this has been closed, but I wanted to add that it's still unreliable for me on Ubuntu 18.04. I'm using @nrobinson2000's `custom-baud` on an Electron E Series LTE running `0.8.0-rc11`. My kernel is: `4.15.0-42-generic #45-Ubuntu SMP Thu Nov 15 19:32:57 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

Plasmarobo:
I can also attest to the fact that this seems to be a kernel or module bug:

On raspberrypi I encountered this behavior with Kernel 4.9, upgrading to Kernel 4.19 seems to fix the sticking issue.

PR #940, opened 2016-04-06
private mbedtls in communications library

m-mcgowan:
the mbedtls headers are private to the communications module so applications can include their own version of mbedtls

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation


m-mcgowan:
Testing (mdma):
- [ ] compile system firmware and verify Electron can connect to the cloud.

technobly:
@m-mcgowan wiring/no_fixture tests pass on Electron U260 as well as connecting and restoring sessions with the tinker usb debugging v0.4.8-rc.6 app.

technobly:
It might be nice to add some kind of bare minimum example to the Docs of what someone can do with this now in User firmware.

m-mcgowan:
That's not quite how it is - it's not in user firmware, it's still private - the changet allows developers to include their own copy of mbedtls in user firmware without conflicts from the system copy.

technobly:
Oh is that kind of beyond the scope of our Docs?  They would have to acquire a copy of mbedtls, and can't duplicate anything used in our stack?

rayjanwilson:
When we include our own, this allows us to push data to services like AWS that require an SSL/TLS connection? Essentially bypassing webhooks?

m-mcgowan:
@rayjanwilson, yes potentially.

PR #942, opened 2016-04-07
wifi country

m-mcgowan:
adds support for setting the wifi country code. This is used to set the available channels, and in the case of japan, the transmit power.

This changes the default country setting to UK, which gives channels 1-13 (previously the default was japan.) This change is so that the default is still full power. 

The country code `JP2` can be used to set the region to Japan, with lower transmit power for TELEC conformance. 

```
# set to japan
echo -n "JP2" > jp2
dfu-util -d 2b04:d006 -a 1 -s 1758:leave -D jp2

# set to us
echo -n "US4" > us
dfu-util -d 2b04:d006 -a 1 -s 1758:leave -D us


# reset to default
echo "0: 00000000" | xxd -r > null
dfu-util -d 2b04:d006 -a 1 -s 1758:leave -D null

```

In theory, the JP2 setting should have 25% less transmit power than the default setting. Looking at the Signal/Noise and SNR on my router, I could see some correlation, although I would welcome further testes to be certain the power level is being altered.

---

- [x] Review code
- [x] Test on device
- [ ] Add documentation
- [ ] Add to CHANGELOG.md

### DIFF
--- a/hal/src/photon/platforms/BCM9WCDUSI09/wifi_nvram_image.h
+++ b/hal/src/photon/platforms/BCM9WCDUSI09/wifi_nvram_image.h
@@ -76,18 +76,18 @@ static const char wifi_ltxp_nvram_image[] =
         "ag0=2"                                                     "\x00"
 
 #if !WIFI_NVRAM_LTXP
-		"maxp2ga0=74"                                               "\x00"
+        "maxp2ga0=74"                                               "\x00"
         "ofdm2gpo=0x44111111"                                       "\x00"
         "mcs2gpo0=0x4444"                                           "\x00"
         "mcs2gpo1=0x6444"                                           "\x00"
 #else
-		"maxp2ga0=62"                                               "\x00"
+        "maxp2ga0=62"                                               "\x00"
         "ofdm2gpo=0xCCCCCCCC"                                       "\x00"
         "mcs2gpo0=0xCCCC"                                           "\x00"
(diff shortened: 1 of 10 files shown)


m-mcgowan:
OMG, this was meant to be in 0.5.0.  Scheduling for 0.5.1 as a bugfix release.

monkbroc:
Code look good. I'll flash it and see if I see the transmit power reduction.

monkbroc:
I can't really see any difference in the transmit power. My laptop and phone are not able to pick up that kind of difference.

I see 2 areas of the DCT that change when I poke values with DFU. I don't know yet how to decode the DCT so I assume that's correct.

Setting JP2:
```
08005d70: 4a50 0002 0000 0000 0000 0000 c0ac de1a  JP..............
08006450: ffff ffff ffff ffff ffff 4a50 32ff 0078  ..........JP2..x
```

Setting US4:
```
08005d70: 5553 0004 0000 0000 0000 0000 c0ac de1a  US..............
08006450: ffff ffff ffff ffff ffff 5553 34ff 0078  ..........US4..x
```

Setting erased (all 0xFF):
```
08005d70: 4742 0000 0000 0000 0000 0000 c0ac de1a  GB..............
08006450: ffff ffff ffff ffff ffff ffff ffff 0078  ...............x
```

monkbroc:
This is ready to merge.

m-mcgowan:
Where should the documentation go? It's not really part of our firmware API.

Issue #925, opened 2016-03-26
[Photon] millis() jumps suddenly


pomplesiegel:
Yesterday we noticed an issue with our lighting fades (based on usage of `millis()`) wherein short fades were jumping suddenly to the end. After a deeper dive, it became clear that the fade was working properly, but the value returned by millis() would suddenly jump.

At first, I figured this was caused by #916, but that issue would cause millis values returned to simply not increment for a long period of time, correct? This looks as if there's suddenly a block when millis() is called, as without warning, the value will jump, without showing the same value over and over, as it would if that value were simply not incremented for a while. From our experience this jump can be up to 300ms, which will make for a very jarring transition. 

This can be seen by a simple application which prints the current `millis()` value:

#### Program to reproduce

Compiled against the tip of develop (as of today) for Photon, this easily reproduces the issue

````cpp
#include "application.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup()
{
  Serial.begin(9600); //Begin USB serial session
  Serial.blockOnOverrun(false); //don't let serial block

  delay(2000); //start-up delay
  WiFi.off(); // no wifi distractions
}

void loop()
{
  Serial.println("Millis: " + String( millis() ) ); 
}
````

#### Sample output

Jump example 1 (small-ish):
````
Millis: 10430
Millis: 10430
Millis: 10430
Millis:Millis: 10431
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
Millis: 10432
MMillis: 10478 <<<<  46 ms jump
Millis: 10478
Millis: 10478
Millis: 10478
````

Jump example 2 (very large): 
````
Millis: 10502
Millis: 10502
Millis: 10502
Millis: 10502
Millis:Millis: 10503
Millis: 10503
Millis: 10504
Millis: 10504
Millis: 10504
Millis: 10504
Millis: 100824 <<<< that's a 320ms jump !!
Millis: 10824
Millis: 10824
Millis: 10824
````
....

#### Details

- Issue occurs with the WiFi chip ON or OFF
- Issue occurs with `SYSTEM_THREAD(ENABLED)` or `DISABLED`
-  Issue occurs with `Serial.blockOnOverrun(false)` or `true`

#### Issue

[...]

PR #943, opened 2016-04-07
millis()/micros() hardening

### LINKED ISSUE #916: millis() stops updating when interrupts are disabled/postponed
The current implementation of `millis()` increments a counter on each SysTIck interrupt (1ms). This works fine for the majority of cases, but there are some corner cases that cause millis to misbehave:

- high priority interrupts or disabled interrupts
- system stop-mode sleep

We may be able to use the `System.ticks()` API to compute `millis()` even when interrupts are disabled.  When sleeping, the RTC can be used to measure the elapsed time between sleep and wakeup, and update `millis()` accordingly.

m-mcgowan:
makes millis()/micros() operations atomic - an interrupt could cause the computed value to be off. Fixes #916 and #925 

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### DIFF
--- a/platform/MCU/shared/STM32/src/hw_ticks.c
+++ b/platform/MCU/shared/STM32/src/hw_ticks.c
@@ -56,8 +56,8 @@ void System1MsTick(void)
  */
 system_tick_t GetSystem1MsTick()
 {
-	int is = __get_PRIMASK();
-	__disable_irq();
+    int is = __get_PRIMASK();
+    __disable_irq();
 
     system_tick_t millis = system_millis + (DWT->CYCCNT-system_millis_clock) / SYSTEM_US_TICKS / 1000;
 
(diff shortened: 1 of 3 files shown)


platform/MCU/shared/STM32/src/hw_ticks.c
```diff
@@ -39,8 +39,15 @@ static volatile system_tick_t system_millis_clock = 0;
  */
 void System1MsTick(void)
 {
+	int is = __get_PRIMASK();
```
technobly: Any reason not to use HAL_disable_irq() and HAL_enable_irq() in interrupts_hal.c?

technobly: Or just wrapping in the ATOMIC_BLOCK() {} ?

m-mcgowan: This is in the platform layer - platform doesn't depend on hal or wiring, so those options are not available.

avtolstoy:
New test `ticks_millis_and_micros_along_with_high_priority_interrupts` passes on both Photon and Electron with this PR, develop branch fails with:
````
Assertion failed: (now_micros=3656010) >= (last_micros=3656346), file tests/wiring/no_fixture_long_running/ticks.cpp, line 66.
```

PR #944, opened 2016-04-07
Store last reset reason and publish it to the cloud

### LINKED ISSUE #403: proposal - store last reset reason / error code / debugging info on reset, send to cloud on reconnect?
It'd be really interesting to know why a device reset, or about where in the firmware things crashed or went wrong.  How hard would it be to catch this, and make it available to users?

sergeuz:
Fixes #403

---

- [x] Review code
- [x] Test on device
- [x] Add documentation
- [ ] Add to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Can set Claim Code via the Serial interface (for use by the CLI.) [#602](https://github.com/spark/firmware/issues/602) 
 - Device ID available via dfu-util. [#949](https://github.com/spark/firmware/pull/949)
 - [Electron] Firmware Reset now available. [#975](https://github.com/spark/firmware/pull/975) and  [Docs](https://docs.particle.io/guide/getting-started/modes/electron/#firmware-reset)
+- [System reset reporting](https://docs.particle.io/reference/firmware/core/#reset-reason) [#403](https://github.com/spark/firmware/issues/403)
 
 ### ENHANCEMENTS
 
--- a/docs/reference/firmware.md
+++ b/docs/reference/firmware.md
@@ -7617,9 +7617,23 @@ void loop() {
 
 The system can track the hardware and software resets of the device.
 
+```
+// EXAMPLE
+// Restart in safe mode if the device previously reset due to a PANIC (SOS code)
+STARTUP(System.enableFeature(FEATURE_RESET_INFO));
+
+void setup() {
+   if (System.resetReason() == RESET_REASON_PANIC) {
+       System.enterSafeMode();
+   }
+}
+```
+
+You can also pass in your own data as part of an application-initiated reset:
+
 ```cpp
 // EXAMPLE
-STARTUP(System.enableFeature(FEATURE_RESET_INFO))
+STARTUP(System.enableFeature(FEATURE_RESET_INFO));
 
 void setup() {
     // Reset the device 3 times in a row
@@ -7629,12 +7643,11 @@ void setup() {
             System.reset(data + 1);
         }
     } else {
+		// This will set the reset reason to RESET_REASON_USER
         System.reset(1);
     }
 }
 
-void loop() {
-}
 ```
 
 **Note:** This functionality requires `FEATURE_RESET_INFO` flag to be enabled in order to work.
@@ -7660,11 +7673,11 @@ Returns a code describing reason of the last device reset. The following codes a
 
 `resetReasonData()`
 
-Returns a user-defined value that has been previously specified for the `reset()` call.
+Returns a user-defined value that has been previously specified for the `System.reset()` call.
 
 `reset(uint32_t data)`
 
-This overloaded method accepts an arbitrary 32-bit value, stores it to the backup register and resets the device. The value can be retrieved via `resetReasonData()` method after the device is restarted.
+This overloaded method accepts an arbitrary 32-bit value, stores it to the backup register and resets the device. The value can be retrieved via `resetReasonData()` method after the device has restarted.
 
 
 ## OTA Updates
--- a/docs/reference/firmware.md
+++ b/docs/reference/firmware.md
@@ -7612,6 +7612,36 @@ void loop() {
 ```
 
 
+### resetReason()
+
+*Since 0.6.0*
+
+Returns code describing reason of the last device reset. Possible codes are:
+
+- `RESET_REASON_PIN_RESET`: Reset button or reset pin
+- `RESET_REASON_POWER_MANAGEMENT`: Low-power management reset
+- `RESET_REASON_POWER_DOWN`: Power-down reset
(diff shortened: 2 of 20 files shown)


sergeuz:
@m-mcgowan Could you review and test this patch on the Core?

m-mcgowan:
Gents, should we do this for the electron? Just thinking about data usage. Perhaps we should have a system flag to enable/disable publishing to the cloud. 

Is there a way the application can get hold of the last reset reason?

m-mcgowan:
Some suggestions:

- rather than publish 2 events, one for the reason and one for the data, it would be more efficient to encode both those values in the one event.

- for readability, we can present the reset reason as a short string rather than a number (or both).

- we'll need docs for the various reset reasons and accompanying data.

m-mcgowan:
Here's a test app that exposes functions for various actions.

```
#include "Particle.h"

int dfu(String s)
{
	System.dfu();
	return 0;
}

int safe(String s)
{
	System.enterSafeMode();
	return 0;
}

int panic(String s)
{
	int value = atoi(s.c_str());
	panic_((ePanicCode)value, nullptr, HAL_Delay_Microseconds);
}

int sleep(String s)
{
	int value = atoi(s.c_str());
	if (!value) value = 1;
	System.sleep(SLEEP_MODE_DEEP, value);
}

void setup()
{
    Particle.function("dfu", dfu);
    Particle.function("safe", safe);
    Particle.function("panic", panic);
    Particle.function("sleep", sleep);
}

void loop()
{
}
```

sergeuz:
Moved flags retrieval and resetting closer to system module's entry point – DFU reset should be recognized properly now (tested on the Photon). Reset code and data are now combined into single cloud event with identifier names used instead of numeric codes whenever possible.

I didn't manage to make deep sleep reset recognized yet – going to check the reference manual further. As a quick workaround we probably can introduce some flag stored in backup register.

m-mcgowan:
Something to try: can an application grab the RCC flags before they are cleared in a startup macro? This could be used as a workaround in cases where applications are already reading the RCC flags, allowing the flags to be retrieved before the system clears them.

Going one step further, it would be good to have  feature flag to disable the system reset syndrome so that applications that don't want it can disable it. (It should be enabled by default.)

sergeuz:
> Something to try: can an application grab the RCC flags before they are cleared in a startup macro? This could be used as a workaround in cases where applications are already reading the RCC flags, allowing the flags to be retrieved before the system clears them.

You're right, it's better to clear flags after user constructors are called, I'll update this PR.

> Going one step further, it would be good to have feature flag to disable the system reset syndrome so that applications that don't want it can disable it. (It should be enabled by default.)

Should such flag disable only cloud notifications or the entire functionality?

m-mcgowan:
I'm thinking perhaps both - a feature flag for the whole thing, and a feature flag for sending the data to the cloud or not. Is there an API for the app to retrieve the reset syndrome.

sergeuz:
> Is there an API for the app to retrieve the reset syndrome.

Yes, though it doesn't expose 'data' field for simplicity: https://github.com/spark/firmware/blob/feature/reset-reason/docs/reference/firmware.md#resetreason

I've added deep sleep handling and moved clearing of the RCC flags to system's main, so that those could be checked in `STARTUP()` code if necessary.

Also, reset info handling now can be configured via `system_flag_t::SYSTEM_FLAG_PUBLISH_RESET_INFO` and `HAL_Feature::FEATURE_RESET_INFO` flags (latter is stored in DCT).

m-mcgowan:
> Yes, though it doesn't expose 'data' field for simplicity:
Would it be much work to expose the data field?  something like `System.resetReasonData()`. Although we have added code to publish the reset reason event, with the data value, I feel app devs should also be able to code their own responses to the reset syndrome, such as logging, or performing conditional action depending upon the type of SOS. For that they'll need the full set of info - both the reason and the data element.

sergeuz:
How about updating `System::resetReason()` in the following way:
```
// Get only reason code
int reason = System.resetReason();

// Get reason code and additional data
uint32_t data;
int reason = System.resetReason(&data);
```

I think we can also allow to store data for user-requested resets, e.g.:
```
enum State {
    STATE_A,
    STATE_B
};

void setup() {
    uint32_t data;
    int reason = System.resetReason(&data);
    if (reason == RESET_REASON_USER && data == STATE_A) {
        System.reset(STATE_B);
    }
}
```

m-mcgowan:
I'm definitely on board with the setting a user defined data when calling System.reset(). 👍 

I'm ambivalent about overloading `resetReason()` to optionally fetch the data.  In some ways, that kind of syntax makes it more of an advanced API, passing in variables by reference, to capture an associated value, compared to having a separate `resetReasonData()` call.  Are there system considerations that would require both values fetched as a single API call?

In terms of code, there's not much in it:

```
 uint32_t data;
 int reason = System.resetReason(&data);
 ```

vs

```
uint32_t data = System.resetReasonData();
int reason = System.resetReason();
```

The 2nd one, while being more to type, is more explicit about what's going on and imho embodies the simplicity of our existing arduino-like APIs, but I open the floor for discussion.

sergeuz:
Sure, let's add `System::resetReasonData()`. Internally, both reason code and additional data are fetched via single `HAL_Core_Get_Last_Reset_Info()` call, but the operation is cheap.

There seems to be a problem with overloading of the `System::reset()` method. It is documented that `System.reset` can be used as a function object for the application watchdog: https://github.com/spark/firmware/blob/develop/docs/reference/firmware.md#application-watchdog, and I've started to get "unresolved overloaded function type" errors from gcc while trying to compile existent API tests.

What could be an alternative name for such extended system reset method? `System.resetEx()`, `System.resetWithData()`? :)

m-mcgowan:
Looks good!

sergeuz:
Not ready yet, updating the documentation :)

sergeuz:
ok, should be ready for the final review

m-mcgowan:
Cool. If the reset reason isn't logged yet (still working through the code) please add that too!

m-mcgowan:
Also I'm wondering if this feature should be enabled by default, since it's the kind of thing you need but only after the firmware is deployed.

The downside is increased bandwidth use (by a few hundred bytes for each cloud connect) but since this can be turned off if the application really doesn't need it then data use can be optimized. What do you think @technobly?

Once the data is published to the cloud, is it republished next time the cloud is connected? It looks like it to me from the code, so it might be a good idea to clear the publish to cloud flag after publishing, so the data is only published once on the first cloud connection.

sergeuz:
> If the reset reason isn't logged yet (still working through the code) please add that too!

> it might be a good idea to clear the publish to cloud flag after publishing, so the data is only published once on the first cloud connection

Done. I've tested the second update just in case – seems to work as expected.

PR #945, opened 2016-04-07
SYSTEM_FLAG_WIFITEST_OVER_SERIAL1

m-mcgowan:
adds `SYSTEM_FLAG_WIFITEST_OVER_SERIAL1` which is disabled by default. Tinker enables this so that the WiFi tester is available during  manufacturing.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### DIFF
--- a/system/inc/system_update.h
+++ b/system/inc/system_update.h
@@ -109,15 +109,15 @@ typedef enum
      */
     SYSTEM_FLAG_STARTUP_SAFE_LISTEN_MODE,
 
-    /**
-     * When 0, the application code is not paused.
-     * When 1, the application code is paused.
-     */
-    // SYSTEM_FLAG_APPLICATION_PAUSED=4,
+	/**
+	 * Enable/Disable use of serial1 during setup.
+	 */
+	SYSTEM_FLAG_WIFITESTER_OVER_SERIAL1,
 
     SYSTEM_FLAG_MAX
 
 } system_flag_t;
+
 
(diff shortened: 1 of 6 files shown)


m-mcgowan:
Tests:
- set TX as input
- enter setup mode
- and verify it's still an input
- set the flag SYSTEM_FLAG_WIFITEST_OVER_SERIAL1
- set TX as input
- enter setup mode
- verify TX is no longer an input

technobly:
@m-mcgowan I edited the above test procedure and executed it.

m-mcgowan:
Does this need adding to the docs? This feels like an internal feature.

technobly:
Product creators will possibly want to take advantage of this right?

markterrill:
Hi, a while back I was emailed the PDF with how to run wifi testing for lab certification. Involved a complicated JTAG reprogramming then using UART to send custom commands to the board to try specific bands at n power.
Is this now in tinker? because that is massive! would love the most basic docu so I can then use tinker instead of the face north, jtag program, etc etc via serial.
it'd be perfection if we could access those tests via standard code methods, ie test wifi b at power n, and run standard serial (usb) writes for debugging notices. that way it'd be really easy to send to a test lab, instructions simply would be, 'it'll do b for 2 minutes at US limits/chans, 2 minutes at UK, 2 mins at japan, then g for ..'
anyways, starting with not having to jtag reprogram is a big step.

PR #948, opened 2016-04-08
WKP pin should not be enabled as a wakeup source unconditionally for STOP mode

### LINKED ISSUE #938: tone on A7 available after stop-mode sleep
https://community.particle.io/t/does-pwm-stop-on-a7-after-stop-mode/21621

avtolstoy:
Fixes #938 

Confirmed that after exiting STOP mode with some other pin specified as a wakeup source, WKP pin retains its configuration (e.g. tone() continues working).

If WKP is specified as wakeup source for System.sleep(), its pinMode will be changed to either INPUT, INPUT_PULLDOWN or INPUT_PULLUP depending on trigger condition. The pin will have to be reconfigured.

---

Doneness:

- [x] Code peer reviewed
- N/A - API tests compiled
- N/A Add documentation

### DIFF
--- a/hal/src/stm32f2xx/core_hal_stm32f2xx.c
+++ b/hal/src/stm32f2xx/core_hal_stm32f2xx.c
@@ -480,9 +480,6 @@ void HAL_Core_Enter_Stop_Mode(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long
 
 void HAL_Core_Execute_Stop_Mode(void)
 {
-    /* Enable WKUP pin */
-    PWR_WakeUpPinCmd(ENABLE);
-
     /* Request to enter STOP mode with regulator in low power mode */
     PWR_EnterSTOPMode(PWR_Regulator_LowPower, PWR_STOPEntry_WFI);
 


m-mcgowan:
Could you write a test to validate the behavior?

m-mcgowan:
I'm uncertain if we should change the default behavior (so A7 is no longer available as a wakeup source after stop mode sleep.)

While that would be the simplest fix, it may break existing applications, although in the long term I feel this is the best solution. We could provide documentation for when A7 wakeup is required then `PWR_WakeUpPinCmd(ENABLE);` should be called before putting the device to sleep.

avtolstoy:
How about we go with the second approach: save the A7 configuration, call `PWR_WakeUpPinCmd(ENABLE);` on exit from STOP mode and restore A7 configuration? I believe it should work just fine, I'm not sure why we dismissed it.

m-mcgowan:
ok, if we can preserve the configuration, that's also a good approach, although I do feel using WKP in stop mode is unexpected, and I can imagine situations where this might cause unexpected wakeups, depending upon what else A7 is being used for. 

I feel your initial PR is best - remove wakeup by default, and let application developers add this if they need it.

monkbroc:
Is this changing documented behavior so that WKP will no longer wake a device up from sleep?

https://docs.particle.io/reference/firmware/photon/#backup-ram-sram-
> Power Down Method: System.sleep(SLEEP_MODE_DEEP)
> Power Up Method	: Rising edge on WKP pin, or Hard Reset

https://docs.particle.io/datasheets/photon-datasheet/#pin-description
> WKP: Active-high wakeup pin, wakes the module from sleep/standby modes

m-mcgowan:
This is stop mode sleep, not deep sleep. As far as I know, being able to wake up stop mode sleep from A7/WKP is undocumented (and imho unepxected) behavior.

monkbroc:
I appears that the WKP pin doesn't wake the micro in STOP mode `System.sleep(D0, RISING);` either before or after this change.

So I'd say this is ready to merge.

avtolstoy:
Correct, according to the reference manual WKP doesn't wakeup CPU from STOP mode:
>If WFI or return from ISR was used for entry:
Any EXTI lines configured in Interrupt mode (the corresponding EXTI
Interrupt vector must be enabled in the NVIC). The interrupt source can
be external interrupts or peripherals with wakeup capability. Refer to
Table 20: Vector table on page 163.

This PR mainly fixes the issue that WKP pin is fixed in INPUT_PULLDOWN configuration after exiting STOP mode and its configuration cannot be changed anymore.

ScruffR:
Just to be awkward ;-)

https://docs.particle.io/reference/firmware/photon/#sleep-sleep-
currently does say
[quote]
wakeUpPin: the wakeup pin number. supports external interrupts on the following pins:
D0, D1, D2, D3, D4, A0, A1, A3, A4, A5, A6, A7
[/quote]

Easily fixed, but contradicts this statement made earlier
[quote]
As far as I know, being able to wake up stop mode sleep from A7/WKP is undocumented (and imho unepxected) behavior.
[/quote]

m-mcgowan:
What I meant was that the WKP pin isn't documented to function _in addition_ to the specified wakeup pin in stop mode. (Currently the code is enabling WKP as a deep-sleep wakeup source, for stop mode sleep.)

danpe:
I'm still having the issue while using the [AssetTracker WakeOnMove](https://build.particle.io/libs/57c486cdb900af2aed000640/tab/3_WakeOnMove.cpp) example.

`System.sleep(WKP, RISING, TIME_PUBLISH_BATTERY_SEC, SLEEP_NETWORK_STANDBY);`

When I used the 0.5.3 electreon firmware version it stopped waking up the device after an hour.
Now after updating to 0.6.0-RC2 it takes about 3-4 hours of no movement to stop waking up the device.

PR #949, opened 2016-04-10
feature/dfu_device_id

m-mcgowan:
Makes the device ID available via dfu, via the DCT, offset 3618, length 12. 

The device ID is copied to persistent storage if the first byte is not 0xFF. 

Works on the P1/Photon/Electron. 
Core TBD. 

The DCT contained a field "device_id" which stored a 6 digit random code. This was renamed "device_code", so we can use "device_id" for the 12 byte stm32 id.

---

Doneness:

- [x] Code peer reviewed
- N/A API tests compiled
- N/A Run unit/integration/application tests on device
- [ ] Add documentation: (Need a new DFU section for the DFU memory map for the DCT region.)

### DIFF
--- a/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/dct.h
+++ b/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/dct.h
@@ -62,7 +62,7 @@ typedef struct __attribute__((packed)) application_dct {
     uint8_t claim_code[63];             // claim code. no terminating null.
     uint8_t claimed[1];                 // 0,0xFF, not claimed. 1 claimed.
     uint8_t ssid_prefix[26];            // SSID prefix (25 chars max). First byte is length.
-    uint8_t device_id[6];               // 6 suffix characters (not null terminated))
+    uint8_t device_code[6];               // 6 suffix characters (not null terminated))
     uint8_t version_string[32];         // version string including date
     uint8_t dns_resolve[128];           // DNS names to resolve.
     uint8_t reserved1[64];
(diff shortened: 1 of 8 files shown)


m-mcgowan:
Tests:

- [ ] - `dfu-util -d 2b04:d006 -a 1 -s 3618:12 -U device_id` matches the device ID printed by `particle serial identify`.
 - [ ] - Photon
 - [ ] - Electron

platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/dct.h
```diff
@@ -75,8 +75,8 @@ typedef struct __attribute__((packed)) application_dct {
     uint8_t alt_device_private_key[192];	// alternative device private key
     uint8_t alt_server_public_key[192];
     uint8_t alt_server_address[DCT_SERVER_ADDRESS_SIZE];		// server address info
-
-    uint8_t reserved2[640];
+    uint8_t device_id[12];									// the STM32 device ID
```
monkbroc: Wouldn't it be more accurate to say "the Particle device ID"? It's not an STM code, right?

avtolstoy: They are actually the same. We use standard STM32 _Unique device ID register (96 bits)_.

monkbroc: I added "Device electronic signature" in a comment in the appropriate HAL function since that's the STM name for the unique device ID.

monkbroc:
Reviewed. Good to merge.

PR #951, opened 2016-04-10
Free memory allocated for previous system interrupt handler

### LINKED ISSUE #927: Memory leak in attachSystemInterrupt
`attachSystemInterrupt` allocates memory for a handler but doesn't delete the previous handler. `detachSystemInterrupt` doesn't delete the previous handler either.

https://github.com/spark/firmware/blob/7c5bfd5d009cfd53d496f05eab05bc92b1bd2794/wiring/src/spark_wiring_interrupts.cpp#L153

sergeuz:
Fixes #927 

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### DIFF
--- a/hal/src/stm32/system_interrupts.cpp
+++ b/hal/src/stm32/system_interrupts.cpp
@@ -21,8 +21,10 @@ uint8_t HAL_Set_System_Interrupt_Handler(hal_irq_t irq, const HAL_InterruptCallb
         *previous = cb;
     if (callback)
         cb = *callback;
-    else
+    else {
         cb.handler = 0;
+        cb.data = 0;
+    }
 
     return true;
 }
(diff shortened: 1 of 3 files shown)


m-mcgowan:
how about adding some tests for this? registering thousands of successive handlers to the same interrupt should be sufficient to show that there is no memory leak and that deleting the old handlers works without causing a fault.

sergeuz:
Added slightly less involved test case that I believe has similar coverage

m-mcgowan:
Tested on the Photon.

Made the tests compile on the core by commenting out GPIO, PWM and Servo tests, which are unrelated. I think we will split the `no_fixture` tests into Peripheral tests and System tests on the core via a conditional define. so we can continue to compile and run the tests on the core.

PR #954, opened 2016-04-12
makes the MAC address available after `WiFi.on()` in all modes.

### LINKED ISSUE #879: WiFi.macAddress() is 00:00:00:00:00:00 when using `SYSTEM_MODE(MANUAL)`
When using `SYSTEM_MODE(MANUAL)`, calling `WiFi.macAddress()` within `setup` or the first `loop` iteration results in an all-zeros MAC address being returned.

Test program:

```
#include "spark_wiring.h"
#include "application.h"

SYSTEM_MODE(MANUAL);

void setup() {
	WiFi.on();
	WiFi.connect();

	while(!WiFi.ready())
		;
}

void loop() {
    byte mac[6];
    WiFi.macAddress(mac);
    Serial.printf("WiFi MAC address: %02x:%02x:%02x:%02x:%02x:%02x\n",
        mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
}
```

Invariably, the output from the first `loop` is `WiFi MAC address: 00:00:00:00:00:00`; the output is correct from the second loop onwards.

It can be fixed by using `AUTOMATIC` mode, but for our application this introduces a significant overhead which we don't need. We could also work around it by skipping the first loop but this seems like a bit of a hack.

Is there a function we can call from user (application) code that will refresh the MAC address, preferably callable from `setup`? Alternately, is there a way the firmware can be fixed so that the MAC address is always made available in `MANUAL` mode?

m-mcgowan:
fixes #879 

Tested on the Core/P1
---

Doneness:

- [x] Code peer reviewed
- N/A  API tests compiled
- [ ] Run unit/integration/application tests on device. Really need the test framework to support reset before each test and setting device modes. 
- [ ] Add documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,10 @@
 - [Electron] Reduced data consumption connecting to the cloud with deep sleep. (NB: see the docs for how to gain the full data reduction.) [#953](https://github.com/spark/firmware/pull/953)
 - Can set Claim Code via the Serial interface (for use by the CLI.) [#602](https://github.com/spark/firmware/issues/602) 
 
-
 ### ENHANCEMENTS
 
 - Local build warns if crc32 is not present. [#941](https://github.com/spark/firmware/issues/941)
+- [Photon/Core] MAC address is available immediately after `WiFi.on()` [#879](https://github.com/spark/firmware/issues/879)
 
 ### BUGFIXES
 
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -345,6 +345,7 @@ public:
         {
             config_clear();
             on_now();
+            update_config(true);
             SPARK_WLAN_STARTED = 1;
             SPARK_WLAN_SLEEP = 0;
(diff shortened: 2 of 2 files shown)


m-mcgowan:
Test:

```
#include "application.h"

SYSTEM_MODE(MANUAL);
void print_address();
void setup() {
    while (!Serial.available());
    WiFi.on();
    print_address();
}

void print_address() {
    byte mac[6];
    WiFi.macAddress(mac);
    Serial.printlnf("WiFi MAC address: %02x:%02x:%02x:%02x:%02x:%02x\n",
        mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
}

void loop()
{
   print_address();
    delay(500);  
}
```

App above verifies that the mac address is available immediately after wifi is enabled. 

Tests:
 - [ ] Core (mdma)
 - [ ] Photon

sergeuz:
The patch looks good and works as expected on the Photon

PR #956, opened 2016-04-13
softap claim code fix

### LINKED ISSUE #937: HTTP server does not handle a request across multiple packets sent without ACK in between
Original issue #680

I've confirmed this is still an issue with at least the `/set` HTTP POST request for setting the claim code. Below is the request body. It works correctly when contained in a single packet (tested via echoing to `nc`), but when the body and headers are separate packets the claim code is not set in the DCT (confirmed via reading the memory over DFU), even though success is returned.

```
POST /set HTTP/1.1
Host: 192.168.0.1
Proxy-Connection: keep-alive
Content-Length: 80
Pragma: no-cache
Cache-Control: no-cache
Origin: null
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.87 Safari/537.36
content-type: text/plain
Accept: */*
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8

{"k":"cc","v":"111111111111111111111111111111111111111111111111111111111111111"}
```

### Wireshark capture of 2 packet request
Packet 1
```
POST /set HTTP/1.1
Host: 192.168.0.1
Connection: keep-alive
Content-Length: 80
Pragma: no-cache
Cache-Control: no-cache
Origin: null
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.87 Safari/537.36
content-type: text/plain
Accept: */*
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8

```
Packet 2
```
{"k":"cc","v":"111111111111111111111111111111111111111111111111111111111111111"}
```

m-mcgowan:
The problem and steps to reproduce it are described in issue #937.

fixes #937. The memory passed from the parse_json() call is only valid during that call, yet the SetValueCommand was caching these for use later. The fix is to copy the data rather than store the pointer.

---

Doneness:

- [x] Code peer reviewed
- N/A - API tests compiled
- N/A - Add documentation

### DIFF
--- a/hal/src/photon/softap.cpp
+++ b/hal/src/photon/softap.cpp
@@ -201,7 +201,21 @@ class JSONRequestCommand : public JSONCommand {
 
 protected:
 
+	/**
+	 * Template method allowing subclasses to handle the parsed json keys.
+	 * @param index The index into the array of keys passed to parse_json_requet of the key that has been matched.
+	 */
     virtual bool parsed_key(unsigned index)=0;
+
+    /**
+     * Template methods allowing subclasses to handle the parsed JSON values.
+     * @param index	the key index this value belongs to
+     * @param t		the jsmn token
+     * @param value	The string value.
+     *
+     * Note that the t and value parameters have a lifetime only for the duration of the method.
+     * They should not be stored for later use.
+     */
     virtual bool parsed_value(unsigned index, jsmntok_t* t, char* value)=0;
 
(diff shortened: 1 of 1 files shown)


m-mcgowan:
I tested using the node app linked in the issue. I could reproduce the problem, seeing corruption. After the fix claim codes work as expected.

PR #959, opened 2016-04-14
HAL_IsISR

### LINKED ISSUE #673: RGB.onChange with delay(1) and stalled RGB-LED
I know it's not a good idea to call delay in an ISR or the `RGB.onChange()` handler, but I did it anyway - just for fun (actually to test `delay()` vs. `Particle.process()` to service `Particle.subscribe()` and `Particle.function()`).

And so I found that without `SYSTEM_THREAD(ENABLED)` even a call of `delay(1)` brings the Photon down to a SOS hard fault (1x red).
Additionally a blocking handler also stalls the onboard RGB-LED (even with multithreading on).

```
#include "InternetButton/InternetButton.h"

//SYSTEM_THREAD(ENABLED)

InternetButton b = InternetButton();

void setup()
{
  b.begin();
  
  RGB.onChange(doRing);
}

void doRing(uint8_t R, uint8_t G, uint8_t B)
{
  static uint32_t ms;
  static int l = 1;
  
  if (b.buttonOn(1) || b.buttonOn(2) || b.buttonOn(3) || b.buttonOn(4))
  {
    l %= 11;
    l++;
    b.allLedsOff();
    b.ledOn(l, 50, 0, 0);
  }

  if (b.buttonOn(1))       // without multithreading
      delay(1);            //   this crashes immediately 
                           // with MF it survives
      
  if (b.buttonOn(2))
    while(b.buttonOn(2));  // this stalls onboard LED
    
  if (b.buttonOn(3))
    while(b.buttonOn(3))   // this one too
      Particle.process();
}
```

m-mcgowan:
adds HAL_IsISR() and uses this to skip calling the background loop from delay(). fixes #673

---

Doneness:

- [x] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/hal/inc/core_hal.h
+++ b/hal/inc/core_hal.h
@@ -211,21 +211,45 @@ int HAL_System_Backup_Restore(size_t offset, void* buffer, size_t max_length, si
 #ifdef USE_STDPERIPH_DRIVER
 #if defined(STM32F10X_MD) || defined(STM32F10X_HD)
 #include "stm32f10x.h"
-inline bool HAL_IsISR()
-{
-	return (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) != 0;
-}
 #elif defined(STM32F2XX)
 #include "stm32f2xx.h"
+#endif // defined(STM32F10X_MD) || defined(STM32F10X_HD)
+
+#if defined(STM32F10X_MD) || defined(STM32F10X_HD) || defined(STM32F2XX)
 inline bool HAL_IsISR()
 {
 	return (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) != 0;
 }
 
+inline int32_t HAL_ServicedIRQn()
+{
+  return (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) - 16;
+}
+
+static inline bool HAL_WillPreempt(int32_t irqn1, int32_t irqn2)
+{
+  if (irqn1 == irqn2)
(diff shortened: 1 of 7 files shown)


avtolstoy:
@m-mcgowan As discussed earlier, I propose to add HAL functions to get currently serviced IRQn and to perform interrupt priority comparison to check whether irqn1 will preempt irqn2 which might be useful under certain circumstances.
Here is the patch that implements the proposed functions `int32_t HAL_ServicedIRQn()` and `HAL_WillPreempt(int32_t irqn1, int32_t irqn2)`:
https://gist.github.com/avtolstoy/affa55f36520721797239eee7fbbd50b

m-mcgowan:
@avtolstoy  - would you write some tests for this please. You can start with basic tests that call `HAL_WillPreempt()` with values that have a known pre-empting outcome. Please document why the subpriority isn't compared (I'm guessing because they don't pre-empt, but would be good to know for certain.

For testing `HAL_ServicedIRQn()` from a real interrupt, perhaps using a timer interrupt is simplest?

In the original PR, I'd not added a test for `HAL_IsISR()`. Please add a test for that too!

avtolstoy:
@m-mcgowan Sure, will do. The subpriority is not affecting the preemption and only comes into play when there are several pending interrupts with the same priority, the one with the lowest subpriority will get serviced first.

avtolstoy:
Test passes on Electron and Photon.

Issue #896, opened 2016-03-09
TCPClient on Electron not receiving all of the data for small files


technobly:
Issue status

- [x] Fixed on U260
- [x] Fixed on G350

---

As reported here in the community:
https://community.particle.io/t/electron-tcpclient-weirdness/20771

Test code, press 'y' to read an 8 byte file.  Typically the header is not even completely printed in the serial console, let alone the 8 bytes of data.  That said, when looking at the logs we do see that all of the data is read, but the system is doing some strange things.  See logs below:

## Code
```cpp
#include "application.h"

// ALL_LEVEL, TRACE_LEVEL, DEBUG_LEVEL, INFO_LEVEL, WARN_LEVEL, ERROR_LEVEL, PANIC_LEVEL, NO_LOG_LEVEL
SerialDebugOutput debugOutput(9600, ALL_LEVEL);

SYSTEM_MODE(AUTOMATIC);

void setup()
{
    Serial.begin( 9600 );
}

/* This function loops forever --------------------------------------------*/
void loop()
{
    if (Serial.available() > 0)
    {
        char c = Serial.read();
        Serial.printf("Hey, you said \'%c\', so I'm gunna: ", c);
        if (c == 't') {
            uint32_t startTime = millis();
            uint32_t endTime = 0;
            Serial.printlnf("Read a text file\r\n[ START TIME ]: %.2f seconds", startTime/1000.0);
            TCPClient client;
            char host[] = "www.textfiles.com";

            Serial.println("connecting...");

            if (client.connect(host, 80))
            {
                Serial.println("connected");
                /* SMALL FILE */
                client.println("GET /100/hack7.txt HTTP/1.0"); // 3988 bytes
                /* LARGE FILE */
                // client.println("GET /100/phrack.29.phk HTTP/1.0"); // 235777 bytes
                client.println("Host: www.textfiles.com");
                client.println("Content-Length: 0");
                client.println();
            }
            else
            {
                Serial.println("connection failed");
            }

            while (client.available() || client.connected())
            {
              if (client.available())
              {
                char c = client.read();
                Serial.print(c);
              }

              if (!client.connected())
              {
                Serial.println();
                Serial.println("disconnecting.");
                client.stop();
              }
            }
            endTime = millis();
            Serial.printlnf("[ END / TOTAL TIME ]: %.2f / %.2f seconds", endTime/1000.0, ((endTime-startTime)/1000.0));
        }
        else if (c == 'y') {
            uint32_t startTime = millis();
[...]

PR #962, opened 2016-04-14
Electron G350 TCP Client fix for #896

technobly:
Part of #896 was previously solved for U260's with #900, but G350's were still not working.  The complete message was not being received by the TCPClient for short messages.

The issue was that U260 and G350 both were receiving a +UUSOCL URC for a remotely closed socket. The U260 would receive this later in time and therefor did not see any issue.  The G350 would receive the URC earlier which caused the socketReadable() routine to throw an unexpected error by setting OK to false.  This was keeping the `cnt` from being returned for the last bytes received.  When not `readable` now, we ensure `len` is set to 0 and OK is set to `true`.

---

Doneness:

- [ ] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1761,6 +1761,7 @@ int MDMParser::socketRecv(int socket, char* buf, int len)
 */
     system_tick_t start = HAL_Timer_Get_Milli_Seconds();
     while (len) {
+        // DEBUG_D("socketRecv: LEN: %d\r\n", len);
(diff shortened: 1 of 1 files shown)


m-mcgowan:
@technobly  - can you please confirm the latest changes were also tested on the U2x0?

technobly:
I tested latest changes on U260 and G350, running both the Y and T commands for small and large files.  The U270 does not work in the US so I have no way to test that one.  If you'd like to give it a go @m-mcgowan that would be helpful.

m-mcgowan:
Verified working on U270.

PR #966, opened 2016-04-18
Fail the build early if crc32 tool is not available

### LINKED ISSUE #941: Warn when crc32 isn't installed
`crc32` is an essential part of the compilation process. If the tool is not installed on the system, `make` currently silently ignore the error and keeps going.

sergeuz:
Fixes #941

---

Doneness:

- [x] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/build/common-tools.mk
+++ b/build/common-tools.mk
@@ -21,5 +21,10 @@ CRC = crc32
 XXD = xxd
 SERIAL_SWITCHER = $(COMMON_BUILD)/serial_switcher.py
 
+crc32_path := $(shell which $(CRC))
+ifeq ("$(crc32_path)", "")
+    $(error "$(CRC) tool is not found")
+endif
+
 CPPFLAGS +=
 


monkbroc:
`which` is not available on Windows. This may break the ability to build on Windows. I don't have a good suggestion right now for portable solution for Windows.

sergeuz:
hmm, I was under impression that when we say "on Windows" in regard to the firmware development we usually mean some environment like MinGW or Cygwin

m-mcgowan:
in fact, on windows crc32 is included in the build, so we can avoid the check on windows. In practice, it's only linux that we need to specifically test, since windows and OSX have crc32 available.

sergeuz:
I made some tests on clean Windows installation. First of all, our makefiles require bash shell (along with some regular UNIX utilities, such as `head` and `tr`), and standalone GNU make quickly fails trying to execute bash statements with batch interpreter. After installing Cygwin or MinGW's MSYS environment (either separately or as part of git for Windows package) the firmware builds normally, however, neither MinGW nor Cygwin include `crc32` by default, so the check introduced by this PR seems to be relevant for Windows as well.

monkbroc:
Great. Let's merge it then.

m-mcgowan:
Tests:
- builds on OSX - builds fine.
- on linux without crc32 - 
```
mat@dev:~/dev/firmware/main$ make
../build/common-tools.mk:26: *** "crc32 tool is not found".  Stop.
```
Error message printed as expected. After installing crc32, `$ sudo apt-get install libarchive-zip-perl` make works as expected.
- windows - (testing now - want to close the browser, so saving this...)

PR #980, opened 2016-04-24
Suggested sudo for Linux users and fix "makes silences" to just "silences".

gl2748:
Suggested sudo for Linux users and fix "makes silences" to just "silences".

---

Doneness:
- [x] Contributor has signed CLA (N/A docs only change)
- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### DIFF
--- a/docs/gettingstarted.md
+++ b/docs/gettingstarted.md
@@ -42,7 +42,7 @@ Navigate to the `modules` folder under firmware
 
     make clean all PLATFORM=photon -s program-dfu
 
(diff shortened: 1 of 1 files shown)


nrobinson2000:
Hi gl

nrobinson2000:
@monkbroc has created a udev rules file for using dfu without sudo. [here it is](https://gist.github.com/monkbroc/b283bb4da8c10228a61e)

nrobinson2000:
It would be good to reference it here.

Or we should mention the utility I created to assist in installing dependencies, using the toolchain and dfu-util. [po-util: the Particle Offline Utility](https://github.com/nrobinson2000/po-util)

monkbroc:
Hi, I work for Particle now so I'll see about including the udev rules in the firmware repository and including instructions to install those: ` sudo cp 50-particle.rules /etc/udev/rules.d/50-particle.rules`

gl2748:
Hi nrob and monk, wherever is easiest, following the instructions without mentioning the need to be root, does not work.

Issue #877, opened 2016-02-17
xQueueGenericReceive deadlock


kiik:
I am experiencing cases on the photon where a deadlock occurs, namely JTAG shows that the device never exits 'xQueueGenericReceive' method. During this both the user application and particle firmware execution are blocked.

gdb frame stack:
```
#0  xQueueGenericReceive (xQueue=0x856f78f8, pvBuffer=pvBuffer@entry=0x0, xTicksToWait=xTicksToWait@entry=1, xJustPeeking=xJustPeeking@entry=0) at WICED/RTOS/FreeRTOS/ver7.5.2/Source/queue.c:1165
#1  0x0803138e in sys_sem_new (sem=sem@entry=0x20005bdc <memp_memory_NETCONN_base+828>, count=count@entry=0 '\000') at WICED/network/LwIP/WWD/FreeRTOS/sys_arch.c:239
#2  0x080267e2 in netconn_alloc (t=<optimized out>, callback=callback@entry=0x0) at WICED/network/LwIP/ver1.4.0.rc1/src/api/api_msg.c:598
#3  0x08025eac in netconn_new_with_proto_and_callback (t=t@entry=NETCONN_TCP, proto=proto@entry=0 '\000', callback=callback@entry=0x0) at WICED/network/LwIP/ver1.4.0.rc1/src/api/api_lib.c:73
#4  0x0802d822 in wiced_tcp_create_socket (socket=0x0, interface=<optimized out>) at WICED/network/LwIP/WICED/tcpip.c:295
#5  0x080401ac in socket_create (family=<optimized out>, type=<optimized out>, protocol=<optimized out>, port=<optimized out>, nif=0) at src/photon/socket_hal.cpp:960
#6  0x0804597c in Spark_Connect () at src/system_cloud_internal.cpp:856
#7  0x08044f14 in establish_cloud_connection () at src/system_task.cpp:215
#8  0x08045174 in manage_cloud_connection (force_events=<optimized out>) at src/system_task.cpp:302
#9  0x080451ae in Spark_Idle_Events (force_events=<optimized out>) at src/system_task.cpp:335
#10 0x08044000 in ActiveObjectBase::run (this=0x200096b8 <SystemThread>) at src/active_object.cpp:49
#11 0x08044020 in ActiveObjectBase::run_active_object (object=<optimized out>) at src/active_object.cpp:92
#12 0x08043968 in std::invoke_thread (ptr=0x2000c498) at src/system_threading.cpp:85
[...]

PR #984, opened 2016-04-27
Patch WICED sys_sem_new to avoid a deadlock with malloc

monkbroc:
Opening a WICED network connection in one thread while a second thread was paused in the middle of memory allocation would lead to a deadlock of the system, with interrupts off.

The symptoms are a solid cyan LED and user code not running.

This PR patches WICED to remove the deadlock possibility in sys_sem_new.

Program to reproduce the issue. Before the fix the blue LED doesn't blink showing a deadlock. After the fix the blue LED blinks quickly.
```
#include "application.h"

Thread allocatorThread;

void allocatorFn() {
  for(;;) {
    void *ptr = malloc(100);
    free(ptr);
  }
}

void setup() {
  pinMode(D7, OUTPUT);
  allocatorThread = Thread("allocator", allocatorFn);
}

void loop() {
  static bool state = false;
  digitalWrite(D7, state);
  state = !state;
  WiFi.resolve("www.particle.io");
}
```

The new unit test deadlocks with the previous WICED library and completes with the patched WICED library.

---

Doneness:

- [x] Code peer reviewed
- N/A API tests compiled
- [x] Run integration tests on device
- [x] Close associated issues and post to community forum (add issue# and posts here)


user/tests/wiring/no_fixture/concurrent.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    // Call malloc in one thread
+    allocatorThread = Thread("allocator", [&]
+    {
+        while(run)
+        {
+            void *ptr = malloc(100);
+            free(ptr);
+        }
+        allocatorThread.dispose();
```
m-mcgowan: afaict, this line will never be reached. It would be better to dispose the allocator thread on line 50, after the WICED connections loop has exited.

m-mcgowan:
@monkbroc - did you see my comment above? I'll merge the fix into the 0.5.1 branch. Could you review the code to be sure the thread is properly cleaned up in the test case.

m-mcgowan:
Fixes #877

user/tests/wiring/no_fixture/concurrent.cpp
monkbroc: I did miss the comment.

I verified that the code works as is: the thread cleanup line is reached when `run` becomes false. `Thread::dispose` doesn't return when called for the current thread.

The reason letting the allocator thread dispose of itself is a better implementation than having the application thread dispose of the allocator thread is the same argument as why `Thread.stop()` is discouraged in Java: you don't know what the thread is doing when you forcibly abort it. In this case it could be in the middle of a call to `malloc` or `free` where aborting the thread could have weird consequences. It's safer to set a `run = false` flag to let the thread exit by itself.

user/tests/wiring/no_fixture/concurrent.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    // Create WICED connections in one thread
+    unsigned long start = millis();
+    while(millis() - start < 2000)
+    {
+        WiFi.resolve("www.particle.io");
+        Particle.process();
+    }
+
+    run = false;
```
m-mcgowan: Ah, I must have missed this on first review, didn't see run=false. (I specifically looked for it too...it must have been hiding behind a semi-colon. ;-)

PR #987, opened 2016-04-28
Fire Hazard, Wrong bitmask is provided for 4208

developingchris:
In the Power Management, there is a function for setting the cut off voltage which takes an integer and converts it to the appropriate bit mask.

When sending in 4.208 volts (the upper bound on the LiPo battery provided with spark kits) it sets the voltage to 4.4 volts. This has caused units on our test bench to fail, and could cause fires if using any other standard LiPo battery that doesn't have a limit circuit installed in the battery.

I am not sure how to complete all of the required steps to get this through your processes, but see this bug as a danger, and would appreciate any help in getting it implemented. Or if someone else can do the unit testing required to get it passed I would not mind them taking over the issue.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation
When passing in 4208, 4400 is actually set to the register.
4.2 is the uppper bound on the operating parameters for the battery that
is provided with the electron.
It would be nice if this function exposed a wider range of voltages,
however this change to perform the function as advertised needs to be
implmented more quickly as it is a fire hazard.

### DIFF
--- a/wiring/src/spark_wiring_power.cpp
+++ b/wiring/src/spark_wiring_power.cpp
@@ -629,7 +629,7 @@ bool PMIC::setChargeVoltage(uint16_t voltage) {
         break;
 
         case 4208:
-        writeRegister(CHARGE_VOLTAGE_CONTROL_REGISTER, (mask | 0b11100000));
+        writeRegister(CHARGE_VOLTAGE_CONTROL_REGISTER, (mask | 0b10110000));
         break;
 
         default:


technobly:
@developingchris thanks for the catch!  We originally set the default limit to 4.112V for battery safety when exposing the battery to extended temperatures.  Not sure how that 4.208V bit mask got to be wrong, but thanks for the catch!

developingchris:
We really like that this could be variable, as we have some HV batteries and other sources we want to try in our systems, that the electron will be a part of. So I would love to expand this function to accept more values between 3504 and 4300  but 4400 is outside of any LiPo we can source.

developingchris:
@technobly what is the process for getting this change up to the point where it's part of the binary I'm putting on my electron?

technobly:
@developingchris this should be available soon.  Keep an eye on the milestone tag for this PR.

PR #995, opened 2016-05-05
TCP src_ports should be unique, in case we make more than one connection to the same host IP/dest_port

gruvin:


gruvin:
My memory doesn't extend back six months, sorry. No one else has had any trouble, apparently. So let's just drop this one. Thanks.

PR #1022, opened 2016-06-09
Additional attributes for log messages

sergeuz:
This PR implements `LOG_ATTR()` macro which allows to specify additional attributes for log messages. For example:

```cpp
wiced_result_t result = wiced_init();
if (result != WICED_SUCCESS) {
    LOG_ATTR(ERROR, (code = result, detail = "wiced"), "Initialization error");
}

int error = mbedtls_rsa_gen_key(&rsa_ctx, rng_func, 1024, 65537);
if (error) {
    LOG_ATTR(ERROR, (code = error, detail = "mbedtls"), "Unable to generate keys");
}
```

Only `code` and `detail` attributes are supported at the moment. New attributes (e.g. thread ID), can be easily added by extending `LogAttributes` structure.

The implementation involves a bit of preprocessor magic in order to support named parameters syntax in both C and C++, so this patch also adds small header file with some typical preprocessor primitives (see `services/inc/preprocessor.h`). In principle, the same could be easily implemented using Boost.Preprocessor, I just wasn't sure if we want to introduce additional compile-time dependency for the firmware build.

Note that this patch is based on the `feature/logger_api` branch and introduces ABI breaking changes to the existent dynalib interface of the logging library. It is likely too late to include this patch into upcoming 0.6.0, so I would mark current logging API as "experimental and subject to change" in the `develop` and postpone publishing of the logging API reference until next release.

------------------

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation


m-mcgowan:
I took a quick look at the code, namely the LogAttributes and the new interface. It would be good to get this in since it's quite extensible, and provides a good stable base for the logging API. Anyone have bandwidth to review in detail?

m-mcgowan:
@sergeuz - any thoughts on how the log attributes can be added to the Wiring logging API?

PR #1026, opened 2016-06-11
Additional attributes for log messages

sergeuz:
*Github doesn't allow to change base branch for existing PRs, so I'm recreating https://github.com/spark/firmware/pull/1022 after base branch of the latter PR has been merged to `develop`.*

*This PR is rebased on top of current `develop`.*

*Original message:*

This PR implements `LOG_ATTR()` macro which allows to specify additional attributes for log messages. For example:

```cpp
wiced_result_t result = wiced_init();
if (result != WICED_SUCCESS) {
    LOG_ATTR(ERROR, (code = result, detail = "wiced"), "Initialization error");
}

int error = mbedtls_rsa_gen_key(&rsa_ctx, rng_func, 1024, 65537);
if (error) {
    LOG_ATTR(ERROR, (code = error, detail = "mbedtls"), "Unable to generate keys");
}
```

Only `code` and `detail` attributes are supported at the moment. New attributes (e.g. thread ID), can be easily added by extending `LogAttributes` structure.

The implementation involves a bit of preprocessor magic in order to support named parameters syntax in both C and C++, so this patch also adds small header file with some typical preprocessor primitives (see `services/inc/preprocessor.h`). In principle, the same could be easily implemented using Boost.Preprocessor, I just wasn't sure if we want to introduce additional compile-time dependency for the firmware build.

Note that this patch is based on the `feature/logger_api` branch and introduces ABI breaking changes to the existent dynalib interface of the logging library. It is likely too late to include this patch into upcoming 0.6.0, so I would mark current logging API as "experimental and subject to change" in the `develop` and postpone publishing of the logging API reference until next release.

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation


sergeuz:
From the original PR:

> any thoughts on how the log attributes can be added to the Wiring logging API?

We have multiple options here, starting with the similar named parameters syntax implemented via overloaded `operator=` (unusual and requires quite a bit of boilerplate code). More traditional Java-like approach could look like the following:

```cpp
Log.info(Log.attributes().setCode(-1).setDetails("details"), "message");
```

I'm kind of ambivalent about adding such functionality though. Most of the attributes are automatic and cannot be extended easily by user. As for the `code` and `detail` attributes, those are going to be used for CLI-assisted log analysis (at least we had some plans for that) and we probably don't want application logging to interfere with it. Also, low-level logging API which supports attributes, source code info and some other stuff is still available for applications, and it's not that scary to use if necessary :)

m-mcgowan:
Ok, we could make it a parameter, which get's a bit confusing with nesting...how about using a more fluid style of coding?

```
Log.attributes().setCode(-1).setDetails("details").info("message");
```

So the attributes class for a log subsumes the logger and exposes the same interface.

sergeuz:
Looks good 👍

sergeuz:
Added user API with the following syntax:
```cpp
Log.code(-1).details("details").info("message");
```

docs/reference/firmware.md
```diff
@@ -8675,12 +8711,25 @@ Parameters:
 
   * format : format string
 
-`void operator()(const char *format, ...)`  
+`void log(const char *format, ...)`  
+`void operator()(const char *format, ...)`
+
+```cpp
+// EXAMPLE
+Log("This is %s message", "info");
```
m-mcgowan: I would change the "info" parameter in case someone confuses it with actually setting the logging level. Something like

```
Log("The secret of everything is %d", 42);
```

docs/reference/firmware.md
```diff
@@ -8675,12 +8711,25 @@ Parameters:
...
+
+  * format : format string
+
+`void log(LogLevel level, const char *format, ...)`  
 `void operator()(LogLevel level, const char *format, ...)`
 
 ```cpp
 // EXAMPLE
-Log("This is %s message", "info"); // Uses default logging level
 Log(LOG_LEVEL_INFO, "This is %s message", "info");
```
m-mcgowan: Same here - let's use a different format param other than "info" to avoid confusion with logging levels.

hal/src/gcc/core_hal.cpp
```diff
@@ -52,25 +52,32 @@ void setLoggerLevel(LoggerOutputLevel level)
...
     if (level < log_level) {
         return;
     }
     std::ostringstream strm;
     // Timestamp
-    strm << std::setw(10) << std::setfill('0') << time << ' ';
-    // Category (optional)
-    if (category && category[0]) {
+    if (attr->has_time) {
+        strm << std::setw(10) << std::setfill('0') << attr->time << ' ';
```
m-mcgowan: we currently don't use the stream classes, since I thought they were quite heavy...do you know what overhead they bring?

services/inc/logging.h
```diff
@@ -202,6 +236,13 @@ extern void HAL_Delay_Microseconds(uint32_t delay);
 #define LOG_MAX_STRING_LENGTH 160
 #endif
 
+// Sets log message attribute
+#define LOG_ATTR_SET(_attr, _name, _val) \
```
m-mcgowan: nice! :-)

hal/src/gcc/core_hal.cpp
sergeuz: This is the gcc platform)

services/inc/services_dynalib.h
```diff
@@ -50,8 +50,8 @@ DYNALIB_FN(17, services, LED_RGB_SetChangeHandler, void(led_update_handler_fn, v
 DYNALIB_FN(18, services, log_print_direct_, void(int, void*, const char*, ...)) // Deprecated
 DYNALIB_FN(19, services, LED_GetColor, uint32_t(uint32_t, void*))
 
-DYNALIB_FN(20, services, log_message, void(int, const char*, const char*, int, const char*, void*, const char*, ...))
-DYNALIB_FN(21, services, log_message_v, void(int, const char*, const char*, int, const char*, void*, const char*, va_list))
+DYNALIB_FN(20, services, log_message, void(int, const char*, LogAttributes*, void*, const char*, ...))
+DYNALIB_FN(21, services, log_message_v, void(int, const char*, LogAttributes*, void*, const char*, va_list))
```
m-mcgowan: how stable is `va_list`? Does it have any provision for binary compatibility between compiler revisions?  My concern is breaking logging when upgrading system part1 to a newer version that has a different layout for va_list (say after we upgrade to a new compiler version.)

wiring/src/spark_wiring_logging.cpp
```diff
@@ -156,50 +173,69 @@ LogLevel spark::LogHandler::categoryLevel(const char *category) const {
...
-    if (info.file) {
-        write(info.file); // File name
-        write(":");
-        snprintf(buf, sizeof(buf), "%d", info.line); // Line number
-        write(buf);
-        if (info.function) {
+    // Source file
+    if (attr.has_file) {
+        // Strip directory path
+        const char *s = extractFileName(attr.file);
```
m-mcgowan: I wonder if LogHandler is doing too much - both formatting and output.  Could we add a `LogFormatter` that takes care of formatting a given log entry? This is separate from a `LogHandler` which outputs the formatted entry to some destination.

services/inc/services_dynalib.h
sergeuz: I doubt that `va_list` definition may change between the compiler versions for the same target platform, as it's an ABI breaking change affecting a lot of existent libraries, including libc itself (though libc always can provide multiple variants of the same function for each ABI version if necessary).

The sole purpose of the `*_v` functions exported by the logging library is to support compile-time checking of the `Logger` printf-alike methods – not a critical functionality, but still it's nice to have it.

m-mcgowan:
If the `_v` functions are only used at compile time, do we need them exposed as a dynalib API? If not, I'd like to remove it so we can save the space used!

sergeuz:
> If the _v functions are only used at compile time, do we need them exposed as a dynalib API? If not, I'd like to remove it so we can save the space used!

There's no other way to forward C-style `...` variable arguments to another function other than using `va_list`. Please check the `services/src/logging.cpp` file, where both regular and `_v` functions are implemented – basically, variants of the functions taking variable arguments via `...` are simple wrappers over the same functions taking `va_list`, so the space overhead is small.

wiring/src/spark_wiring_logging.cpp
sergeuz: I was thinking along the same lines initially, but decided to simplify the implementation to avoid unnecessary dynamic allocations, such as in the example below:

```cpp
LogHandler handler(new DefaultLogFormatter(), &Serial1, ...)
```

C++ mixins are also not an option, as they cannot be configured and instantiated dynamically, which is essential for the yet-to-be-developed dynamically enabled logging.

In principle, we can move the formatting code to the existent `StreamLogHandler` to split the functionality more evenly.

sergeuz:
I think this should be ready for merging, @m-mcgowan?

I've moved the formatting code in question to the `StreamLogHandler` class for now. In general, `LogHandler` and `StreamLogHandler` classes are never used directly in application code (unless it implements its own log handler), nor documented in the firmware reference, so we still have some space for refactoring if it will be necessary.

PR #1030, opened 2016-06-13
Fix soft AP suffix broken by the addition of device id in DCT

monkbroc:
The soft AP name was broken by the renaming of the `device_id` field in DCT to `device_code` (that's the soft AP suffix) and the addition of a new `device_id` field (that's the 24 char device id). The soft AP code was still using the `device_id` field.

Test:
- Flash develop, put the device in soft AP and see that the soft AP name is Photon-$ (where $ is some weird suffix)
- Flash fix, see the soft AP name is Photon-AW23 (where AW23 is 4 device specific characters).

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### DIFF
--- a/hal/src/photon/softap.cpp
+++ b/hal/src/photon/softap.cpp
@@ -727,31 +727,31 @@ void random_code(uint8_t* dest, unsigned len) {
     bytesToCode(value, (char*)dest, len);
 }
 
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Looks good! :+1:

Issue #1024, opened 2016-06-09
Revisit cloud connection behavior in case of DNS failure


monkbroc:
Currently if DNS resolution of `device.spark.io` fails, the Photon will attempt to connect to 54.208.229.4. Use of this hard-coded IP address should be removed and instead resolution attempted again with using a known good DNS server.

This is based on a discussion with the cloud team.

---

Completeness:
- [ ] Minimum test case added
- [ ] Device, system and user firmware versions stated
- [ ] Particle confirmed

technobly:
We used to have a hard coded IP fallback, but it has since been removed.  Now it will keep retrying over and over to resolve the IP address of the server, but there is no logic that attempts to change the DNS server and try again.

Should this issue be converted into an enhancement request for this behavior?  What DNS should be used? 8.8.8.8?

Issue #139, opened 2014-03-03
Persist cache of last successfully connected Cloud server IP


towynlin:
As of 2 March 2014 with latest firmware, we can store Cloud server IP or domain name on external flash.  Fallbacks are hard coded, first to "device.spark.io" then to 54.208.229.4.

An additional layer of robustness suggested by @andyw-lala is a persistent cache of an IP to which the Core recently successfully connected.

This datum is probably a good candidate for the CC3000 EEPROM.

m-mcgowan:
The present connection logic first determines which IP to connect to and then tries to connect to that. In pseudo code:

- is there server address info stored in flash?
 - if it's a domain name, resolve to an IP (which may fail)
 - else use the IP as is
- if no IP is set, use the default (54.208.229.4)

The biggest weakness here is if the DNS lookup fails, in which case I feel we should not fallback to the default IP, but rather fail the connection, or the device may unexpectedly switch servers due to errors in the external environment (DNS).

cc: @joegoggins

joegoggins:
Thank you for clarifying this logic @m-mcgowan and proposing that we eliminate the default IP, I completely agree with you.  

In addition to the benefit this would have to the firmware and device behavior, this would also simplify cloud-side requirements and allow us to do better, more efficient balancing, server-localization, and failover of device connections via DNS. Whereas doing these same things with a static IP are much more difficult and less robust to failure.

Strong   :+1: :shipit: from me on this one.

bkobkobko:
So my only concern is that DNS fails in a bad way on certain TI CC3000 chips with certain links.  And once it fails, it can only be recovered (as far as I can tell) by re-flashing the latest TI patch.

It is easy to tell when the TI part has failed since the IP address of the DNS host is always 76.83.0.0.

You guys should factor that into your thinking on this one since today a core with a bad DNS host can connect to the cloud and be reflashed but you will not be able to do that after a change like this.

m-mcgowan:
Thanks @bkobkobko for articulating the worries that I had in the back of my mind! 

What we can do in this instance is in the `inet_gethostname()` function, check if the hostname matches our primary cloud hostname, and if the lookup fails, and the DNS host is 76.83.0.0 then fallback to the default IP.  This will keep open the cloud connection for devices with failed DNS resolution.

This is a lot nicer than the current default IP logic since it only comes into play during a known error condition with the cc3000.

m-mcgowan:
[...]

PR #1048, opened 2016-06-28
remove IP fallback

m-mcgowan:
Removes the fallback IP that is used when DNS resolution fails. Instead, the cloud connection is failed and the system will have to retry.  This means DNS lookup failure is now consistent with other modes of connection failure. 

Addresses #139 

Related to #1024

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### Internal

- Removed hardcoded server IP that was used when DNS resolution fails. Instead, the cloud connection is failed and the system will have to retry.  This means DNS lookup failure is now consistent with other modes of connection failure.  Addresses #139 Related to #1024

### DIFF
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -955,14 +955,6 @@ int determine_connection_address(IPAddress& ip_addr, uint16_t& port, ServerAddre
     }
 #endif
 
-	if (ip_address_error && (!HAL_PLATFORM_CLOUD_UDP || !udp))
-	{
-		// TCP - final fallback in case where flash invalid
-		ip_addr = (54 << 24) | (208 << 16) | (229 << 8) | 4;
-		//ip_addr = (52<<24) | (0<<16) | (3<<8) | 40;
-		ip_address_error = false;
-	}
-
 	return ip_address_error;
 }
 


m-mcgowan:
## Acceptance testing

Just an outline. This is pretty tricky since it requires co-ordination of several systems:

- create a server address block binary image corresponding to a DNS address that is administered via the local router
- ensure the local DNS server doesn't resolve the fqdn
- flash the server address block to the device via DFU
- restart the device and verify that it doesn't connect to the cloud. (USB vendor request for current system state? query the API to determine if the device is online?)
- add the hostname to the DNS server so that the address resolves to the particle cloud IP
- verify the device connects to the cloud within 2 minutes (the maximum exponential backoff.)
- replace the server address block with the regular public cloud address block

The tricky part is automating the interactions with the local router and the DNS server.  This may be possible via the `wl` command on some routers.

zsup:
Instead of the IP fallback, can we fall back to a known DNS server (like Google's 8.8.8.8)?

m-mcgowan:
We've gotten rid of the IP fallback so we don't have hard-coded IPs in firmware. 

We don't have much influence over the DNS server, since this is obtained via DHCP.  On the Photon, the DNS resolution does use 8.8.8.8 as a backup server internally (that's in WICED) but we have no control over what happens on the Electron (or Core.)

My feeling here is to keep the DNS server as is and not change it in the event of errors, since that could lead to some hard to track changes in behavior that are triggered only relatively infrequently by DNS failure. By not changing the DNS server from the one configured by the access point, we keep the system behavior predictable. @joegoggins: what do you think?

The primary source of DNS failure (apart from network connectivity) when the code was first written was the CC3000 suddenly reporting a bogus value for all hostname lookup queries. This has been taken care of, and with that fix in place, it makes sense now to treat DNS lookup failure the same way as other failures when attempting to connect to the cloud  - by marking the connection attempt as failed, meaning the system will retry again soon.   This allows transient errors to be recovered from (when the system retries the connection) and permanent infrastructure errors (such as firewall, misconfigured DNS, unreachable IP, mismatched keys, etc...) to result in predictable, repeatable connection failure until the source of the error is fixed.

joegoggins:
Very well said @m-mcgowan, I found these points very compelling:

> ...fallback ip...**reason the code was first written was the CC3000** suddenly reporting a bogus value for all hostname lookup queries.

> ...keep the DNS server as is and not change it in the event of errors, **since that could lead to some hard to track changes in behavior that are triggered only relatively infrequently by DNS failure**.

> ... result in **predictable, repeatable connection failure** until the source of the error is fixed.

towynlin:
Let's keep in mind that, down the road, we will definitely need some router manipulation as part of our long term automated hardware testing. Much of the testing requested by product creators involves questions like, "What happens when the Wi-Fi is gone and later comes back?"

Issue #211, opened 2014-05-27
Add support for daylight savings time


wgbartley:
Add support for daylight savings time to Time library, notably `Time.zone()`.

https://community.spark.io/t/time-zone-and-daylight-savings-and-spark-synctime/4568

m-mcgowan:
Are you sure `Time.zone()` solves the problems requested in the thread above? It appears that at the very least a time-zone descriptor is needed, and that the cloud returns the current time zone offset for that descriptor and the time of the next change.

andyw-lala:
Seems to me that the timezone should be a persistent customer settable
attribute of a core. The cloud should always just deal with UTC. That
implies that the core needs to calculate local time, there are table-driven
shortcuts that can be used and/or refreshed from the cloud.

m-mcgowan:
Moving the timezone calculations locally with updates from the cloud is a significant undertaking,  considerably more than having the cloud perform these operations. What's the benefit of having this done locally when we already make use of a cloud connection to retrieve the current time?

I agree the cloud always works in UTC, and that doesn't preclude it also providing a timezone offset for a given time zone descriptor, which is then used by the core to calculate local time. The timezone descriptor may be stored locally, or the cloud may be able to provide the correct timezone via GeoIP services.

andyw-lala:
I guess if the core spits out the (locally configurable, and persistant) TZ
choice as part of the initial handshake, that would be perfectly fine,
except that doesn't handle transitions across clock forward/backward
events, so that would require the cloud to be able to push a new local time.

It is the corner cases that hurt with TZ implementations.

On Sun, Apr 5, 2015 at 10:09 AM, Matthew McGowan <notifications@github.com>
wrote:

> Moving the timezone calculations locally with updates from the cloud is a
> significant undertaking, considerably more than having the cloud perform
> these operations. What's the benefit of having this done locally when we
> already make use of a cloud connection to retrieve the current time?
>
> I agree the cloud always works in UTC, and that doesn't preclude it also
> providing a timezone offset for a given time zone descriptor, which is then
> used by the core to calculate local time. The timezone descriptor may be
> stored locally, or the cloud may be able to provide the correct timezone
> via GeoIP services.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spark/firmware/issues/211#issuecomment-89789522>.
>



-- 
Andy

andyw-lala:
But the offset isn't set and forget unless you have local tables to handle
the transitions.

[...]

PR #1058, opened 2016-07-10
Daylight Saving Time support

avtolstoy:
Implements initial Daylight Saving Time (DST) API proposed in #211.

Added Wiring functions:
```cpp
  /* Retrieve the current DST offset that is added to the current local time when
   * Time.beginDST() has been called.
   * The default is 1 hour.
   */
  float TimeClass::getDSTOffset();
  /* Set a custom DST offset */
  void TimeClass::setDSTOffset(float offset);
  /* Add the offset from getDSTOffset() to the current time */
  void TimeClass::beginDST();
  /* Do not add the offset from getDSTOffset() to the current time */
  void TimeClass::endDST();
  /* Returns true if DST is in effect (beginDST() was called previously) */
  uint8_t TimeClass::isDST();
```

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

FEATURE
- added Daylight Saving Time support [#1058](https://github.com/spark/firmware/pull/1058) per proposed [#211](https://github.com/spark/firmware/issues/211)

### DIFF
--- a/docs/reference/firmware.md
+++ b/docs/reference/firmware.md
@@ -5673,6 +5673,9 @@ Retrieve the current time in the configured timezone as seconds since January 1,
 
 Note that the functions in the `Time` class expect times in UTC time, so the result from this should be used carefully.
 
+_Since 0.6.0_
+
+Local time is also affected by the Daylight Saving Time (DST) settings.
 
 ### zone()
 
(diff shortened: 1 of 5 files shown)


ScruffR:
I'm not sure how this will actually address the issue of switching DST on and off :confused: 

In #211 there was talk about a way to set the switching pattern

> TimeChangeRule usEDT = {"EDT", Second, Sun, Mar, 2, -240};  //UTC - 4 hours
> TimeChangeRule usEST = {"EST", First, Sun, Nov, 2, -300};   //UTC - 5 hours
> Timezone usEastern(usEDT, usEST);
> 
> // calls Time.zone(), Time.enableDST(), Time.setDSTOffset() using a DST setting determined from
> // the current UTC time and the data in the Timezone instance.  
> Time.timezone(usEstern);

But I can't see any way to set an auto-switching pattern in the docs or in this PR, but that would seem to be the only thing that we are actually missing. 
If I still have to decide to switch DST on or off in code, I could do that just as much via `Time.zone()` as usual - so what?

BTW, these values (zone(), DSTOffset and the swtiching rules) should be made sticky too, to add value, otherwise we could dump (`setDSTOffset(float)`) in favour of `beginDST(float)` - since we'd have to code both lines ourselves anyway, what's the benefit?

Without the ability to set switching rules and persisting the settings the whole "enhancement" here is worth little - IMHO.

pomplesiegel:
Oh, i didn't realize that this DST implementation wouldn't automatically switch on and off based on the calendar! Unfortunately I have to agree - an implementation for adding or subtracting amounts of an hour from the current time won't help us very much as product developers. 

Right now we're manually mimicking DST by changing the `Time.zone()` offset as an OTA function call. I don't think this feature in its current state will be be much more elegant in final application implementation. 

Is there a gameplan for an automatically adjusting time based on DST calendar rules? Virtually every ioT product in the world needs this.

ScruffR:
For keeping some things on the radar, I carried over some bits and bobs to issue  #1161

PR #932, opened 2016-03-31
Fixes flush() semantics for UDP and TCPClient

avtolstoy:
Closes #469 

---

- [x] Review code
- [ ] Test on device
- [x] Add documentation
- [x] Add to CHANGELOG.md

PR #1064, opened 2016-07-24
discard buffer still needed in a few places in UDP library

technobly:
After recent changes in #932, `TEST=wiring/networking` was stalling on electron and photon.  Cause was due to buffer not being cleared (i.e. `flush_buffer()`) for `parsePacket()`.  Also made this change in `beginPacket()`... and a couple other places which are probably not critical.  This is in contrast to just using `flush()` which will block while sending data in the future, in functions like endPacket().

---

Doneness:

- [ ] Code peer reviewed ( @avtolstoy please review )
- [ ] API tests compiled
- [x] Add documentation (N/A, already in #932)
- [x] Add to CHANGELOG.md after merging (add links to docs and issues) (N/A, already in CHANGELOG for #932)

### DIFF
--- a/wiring/src/spark_wiring_udp.cpp
+++ b/wiring/src/spark_wiring_udp.cpp
@@ -67,7 +67,7 @@ void UDP::releaseBuffer()
     _buffer = NULL;
     _buffer_allocated = false;
     _buffer_size = 0;
-    flush_buffer();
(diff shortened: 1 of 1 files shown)


avtolstoy:
Yes, these changes look correct, @technobly . I thought I changed those though!

PR #1065, opened 2016-07-25
Feature/electron/system module reorder

avtolstoy:
- Adds Electron system-part1 dependency on system-part3
- Electron system parts renaming:
  - system-part3 -> system-part1 (module index unchanged: 3)
  - system-part1 -> system-part2 (module index unchanged: 1)
  - system-part2 -> system-part3 (module index unchanged: 2)

---

Doneness:

- [x] Code peer reviewed
- [X] API tests compiled
- [X] Run unit/integration/application tests on device
- [ ] Add documentation

Need to update Electron Datasheet to include alternate memory map for post v0.6.0 electrons.

ENHANCEMENT

- [Electron] System parts reordered from 1,2,3 to 3,1,2.  However with renaming, it's kept in 1,2,3 order for flashing. [#1065](https://github.com/spark/firmware/pull/1065)


technobly:
Tested with Tinker, verified 1,2,3 is really the old 3,1,2 behind the scenes memory map wise.

Tested with `make clean all PLATFORM_ID=10 -s program-cloud SPARK_CORE_ID=123412341234123412341234 SPARK_ACCESS_TOKEN=345634563456345634563456` and while it's in the correct order, the script currently doesn't wait long enough for each part to complete, so that needs work (I'll add an issue for that).

PR #1078, opened 2016-08-02
get the current version of gcc-arm-non-eabi from Homebrew

### LINKED ISSUE #1077: gcc-arm-none-eabi-49 is no longer available from homebrew; using gcc-arm-none-eabi
The build instructions specify prerequisites for Mac OS X, that include installing `gcc-arm-none-eabi-49` via a Homebrew tap. That version is no longer labelled as such, but it's possible to install `gcc-arm-none-eabi`.

When I do that, I get this version string:

```
Edwards-MacBook-Air:firmware emv$ arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for ARM Embedded Processors) 4.9.3 20150529 (release) [ARM/embedded-4_9-branch revision 227977]
```

which I think is correct.

---

Completeness:
- [x] Minimum test case added
- [x] Particle confirmed

vielmetti:
< Please summarize the Problem and Solution this PR addresses as clearly as possible. This facilitates speedy peer review and testing :) >

Homebrew has updated its tap and the current version of the compiler dependencies to install is "gcc-arm-non-eabi" not "gcc-arm-non-eabi-49". You end up with the same system, just a different name.

Closes #1077 .

---

Doneness:
- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -22,7 +22,7 @@ message if the version is older than this.
 **OS X** users can install the toolchain with [Homebrew](http://brew.sh/):
 - `brew tap PX4/homebrew-px4`
 - `brew update`
-- `brew install gcc-arm-none-eabi-49`
+- `brew install gcc-arm-none-eabi`
 - `arm-none-eabi-gcc --version` (should now say v4.9.x)
 
 If you are upgrading an existing installation you will have to unlink and link your symblinks:


technobly:
Closed by #1183

Issue #1055, opened 2016-07-09
Run event loop in delay on app thread only


monkbroc:
I have an application running without system thread with the hardware logic running in a separate wiring `Thread` and only cloud related things in `loop()`.

When calling `delay()` in the hardware thread, it processes cloud events. Since the main thread also processes cloud events this leads to random crashes.

The workaround is to call `HAL_Delay_Milliseconds()`.

The fix is to ensure that `delay()` only processes cloud events when called on the main thread.

This issue applies with system thread enabled our disabled.

I'll add a test case when my internet comes back at home.

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated
- [x] Particle confirmed

m-mcgowan:
Let's try to work this through so the code is covered by a unit test that exercises the different cases/emulates different calling threads, as well as a suite of acceptance test apps to actually test the desired outcome for real. This is particularly difficult with the way the code is presently structured.

PR #1094, opened 2016-08-14
Feature/delay event processing

avtolstoy:
Fixes issue #1055

`system_delay_ms` should not use `!system_thread_get_state(NULL)` to determine whether to run event loop or not, as it is still possible to create a thread with `SYSTEM_THREAD(DISABLED)` and execute `delay()` in that thread, which will cause the event loop to be processed from that thread:
https://github.com/spark/firmware/compare/feature/delay-event-processing?expand=1#diff-de0a60700cd1e842d6a89faaba614c8cL446

This PR:
- Changes `system_delay_ms` condition to `if ((!PLATFORM_THREADING || APPLICATION_THREAD_CURRENT()) && !HAL_IsISR())`
  - `!system_thread_get_state(NULL)` -> `!PLATFORM_THREADING`
- Exposes `APPLICATION_THREAD_CURRENT()`/`SYSTEM_THREAD_CURRENT()` macros as  `application_thread_current`/`system_thread_current` dynalib functions
- Adds `wiring/threading_no_threading_delay` test

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### BUGFIX

- Do not run the event loop from delay() when threading is enabled. Fixes [#1055](https://github.com/spark/firmware/issues/1055)

### DIFF
--- a/system/inc/system_dynalib.h
+++ b/system/inc/system_dynalib.h
@@ -59,6 +59,9 @@ DYNALIB_FN(16, system, Spark_Prepare_For_Firmware_Update, int(FileTransfer::Desc
 DYNALIB_FN(17, system, Spark_Save_Firmware_Chunk, int(FileTransfer::Descriptor&, const uint8_t*, void*))
 DYNALIB_FN(18, system, Spark_Finish_Firmware_Update, int(FileTransfer::Descriptor&, uint32_t, void*))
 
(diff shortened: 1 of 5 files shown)


monkbroc:
Fantastic. Code looks good to me. Nice work coming up with such a detailed test.

Issue #1032, opened 2016-06-13
when flashing an invalid firmware binary (that the device ignores) it should post an event describing why the binary was not applied


m-mcgowan:
E.g. flashing system binaries in the wrong order, or flashing a binary with an invalid CRC.

The event will help provide some visibility into the system, and help troubleshoot cases when binaries are not being applied, and conversely for success cases, will provide confirmation that a flashed binary will be applied on next reboot.

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated
- [x] Particle confirmed

technobly:
Spec from a discussion we had about this issue:

- publish an event when the system has rx'd an OTA image
    - success (include parseable module info)
    - failure (reason crc, module dependency (include parseable module info), wrong device type)
    - publish to cloud or reply to YMODEM host
- add to reset syndrome for when the system reboots and applies an OTA image
- HELLO message contains OTA_FLASHED_Status_SysFlag, so theoretically the cloud already knows if an OTA update was applied, and can generate an event based on this.

PR #1097, opened 2016-08-21
OTA/YModem update result

avtolstoy:
Implements #1032

When updating the device via YModem:
- After receiving `EOT` from Host, device doesn't immediately respond with `ACK`. Instead binary validation is performed.
- If successful, device notifies the Host with `ACK` and prints received file name and size.
- If validation fails, the device notifies the Host with two `CA` (`CA`, `CA`) as to indicate the error and prints hexademical error code, which is a bitmask of failed checks (https://github.com/spark/firmware/blob/c916faafaea75f802a521d32ca4353aba812dfe2/hal/inc/ota_flash_hal.h#L52)

When updating OTA:
- After successfully receiving the binary, the device performs validation and before rebooting publishes an event `spark/device/ota_result`, which contains the following JSON object:

``` javascript
{"r": "ok", "u": {JSON ota module info, same format as particle serial inspect}}
```
- In case of an error, the same object is published. JSON ota module info might not be present

``` javascript
{"r": "error", "u": {JSON ota module info, same format as particle serial inspect, might not be present}}
```
- A full system module info object + ota module info don't fit into 255 byte publish message limit, so only ota module info is included.

---

Doneness:
- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### Enhancements

- When flashing (OTA/YModem) an invalid firmware binary (that the device ignores) it will post an event describing why the binary was not applied. [[Implements #1032]](https://github.com/spark/firmware/issues/1032)

### DIFF
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -55,6 +55,7 @@ int userVarType(const char *varKey);
 const void *getUserVar(const char *varKey);
 int userFuncSchedule(const char *funcKey, const char *paramString, SparkDescriptor::FunctionResultCallback callback, void* reserved);
 
+static int finish_ota_firmware_update(FileTransfer::Descriptor& file, uint32_t flags, void* module);
 static void formatResetReasonEventData(int reason, uint32_t data, char *buf, size_t size);
 
 static sock_handle_t sparkSocket = socket_handle_invalid();
@@ -605,7 +606,8 @@ void Spark_Protocol_Init(void)
         		callbacks.transport_context = nullptr;
         }
 		callbacks.prepare_for_firmware_update = Spark_Prepare_For_Firmware_Update;
-        callbacks.finish_firmware_update = Spark_Finish_Firmware_Update;
+        //callbacks.finish_firmware_update = Spark_Finish_Firmware_Update;
+        callbacks.finish_firmware_update = finish_ota_firmware_update;
         callbacks.calculate_crc = HAL_Core_Compute_CRC32;
         callbacks.save_firmware_chunk = Spark_Save_Firmware_Chunk;
         callbacks.signal = Spark_Signal;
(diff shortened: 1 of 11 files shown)


m-mcgowan:
We should try to write tests for this:

- unit tests in the user/test/unit package to validate the ymodem code (we can stub most of the system functionality). focus here is on testing tiny pieces of the code.

- integration tests that save the file transmitted via ymodem to the filesystem or RAM so we can verify the file saved matches the one sent

- acceptance tests that use the CLI to flash a file to the device (both a valid and invalid one) and verifies the result.

system/src/system_cloud_internal.cpp
```diff
@@ -1122,6 +1124,42 @@ inline uint8_t spark_cloud_socket_closed()
...
+int finish_ota_firmware_update(FileTransfer::Descriptor& file, uint32_t flags, void*)
+{
+    hal_module_t module;
+    uint8_t buf[512];
+
+    int result = Spark_Finish_Firmware_Update(file, flags, &module);
+
+    formatOtaUpdateStatusEventData(flags, result, &module, buf, sizeof(buf));
+
+    Particle.publish("spark/device/ota_result", (const char*)buf, 60, PRIVATE);
```
technobly: Add logs for this event ( as in PR #1169 ), dumping the buffer would be a quick way to tell in logs what the result of the OTA/YModem was.

system/src/system_cloud_internal.cpp
```diff
@@ -1122,6 +1124,42 @@ inline uint8_t spark_cloud_socket_closed()
...
+    HAL_Delay_Milliseconds(1000);
```
technobly: I'm assuming we need this to ensure the publish is sent before we soft reset to apply the OTA update.  Can we make this more synchronous, as implemented in PR #1144 ?  We can do this as a separate PR after these get merged.  I'll take notes on all of these review comments to make sure we don't forget :)

system/src/system_update.cpp
```diff
@@ -443,6 +448,42 @@ const char* module_name(uint8_t index, char* buf)
...
 }
 
+bool module_info_to_json(appender_fn append, void* append_data, const hal_module_t* module, uint32_t flags)
+{
+    AppendJson json(append, append_data);
+    char buf[65];
+    bool result = true;
+    const module_info_t* info = module->info;
+
+    buf[64] = 0;
```
technobly: This is fine, but I would like to see this as `buf[sizeof buf - 1] = '\0';`

system/src/system_update.cpp
```diff
@@ -451,41 +492,33 @@ bool system_info_to_json(appender_fn append, void* append_data, hal_system_info_
...
 
     result &= json.write(']');
     return result;
 }
 
+bool ota_update_info(appender_fn append, void* append_data, void* mod, bool full, void* reserved)
+{
+    bool result = true;
+    AppendJson json(append, append_data);
+    //result &= json.write('{');
```
technobly: Do these `{}` make the output exceed 255? Because it would be nice to have a properly formatted JSON object by default.

PR #1105, opened 2016-08-29
Fix for listening mode delay

### LINKED ISSUE #1013: Delay before entering listen mode with multithreading
Entering listen mode through `WiFi.listen()` in multithreading has a delay to wait for the current connection attempt to finish. The system firmware should cancel the connection attempt before entering listen mode.

Calling `WiFi.disconnect()` first from the app isn't sufficient because the system thread races to reconnect before the app has time to call `WiFi.listen()`.

A workaround is to call `Particle.disconnect()` before `WiFi.disconnect()` and `Particle.connect()` after listen mode is done.

Test case to reproduce behavior (tested with Photon 0.5.1):

- Configure a wifi network with a bogus name into the Photon
- Ground D0 to call WiFi.listen() while the connection attempt is ongoing

```
#include "application.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(AUTOMATIC);

void reconnectAfterListen() {
  Particle.connect();
}

void setup() {
  System.on(wifi_listen_end, reconnectAfterListen);
  pinMode(D0, INPUT_PULLUP);
}

void loop() {
  if(digitalRead(D0) == LOW) {
    // Delay of ~20 seconds to enter listen mode if these 2 disconnect calls are missing
    Particle.disconnect();
    WiFi.disconnect();
    WiFi.listen(true);
  }
}

/*
Main source of delay:

While connecting to the WiFi, wlan_connect_finalize calls
wiced_interface_up -> wiced_join_ap -> wiced_join_ap_specific ->
wwd_wifi_join -> host_rtos_get_semaphore (aka xQueueGenericReceive)
which blocks the system thread until connection attempt finishes.

Reason delay cannot be stopped by WiFi.disconnect():

With multithreading on, calling WiFi.disconnect() in the app calls
network.disconnect() to the system thread, then the system background
task is run (Spark_Idle_Events) which runs manage_network_connection
which immediately calls network.connect() which blocks the system thread
while the WiFi connects.
*/
```

---

Completeness:
- [x] Minimum test case added
- [x] Particle confirmed

sergeuz:
This PR fixes #1013 by canceling current connection attempt before entering the listening mode.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### Bugfix

- Cancel current connection attempt before entering the listening mode with WiFi.listen(true) and also WiFi.off(). Fixes [#1013](https://github.com/spark/firmware/issues/1013)

### DIFF
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -266,10 +266,10 @@ public:
         if (stop) {
             WLAN_LISTEN_ON_FAILED_CONNECT = 0;  // ensure a failed wifi connection attempt doesn't bring the device back to listening mode
             WLAN_SMART_CONFIG_START = 0; // Cancel pending transition to listening mode
-        } else {
+            WLAN_SMART_CONFIG_ACTIVE = 0; // Break current listening loop
+        } else if (!WLAN_SMART_CONFIG_ACTIVE) {
             WLAN_SMART_CONFIG_START = 1;
         }
-        WLAN_SMART_CONFIG_ACTIVE = 0; // Break current listening loop
     }
 
     void listen_command() override
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -279,8 +279,9 @@ public:
 
     void connect(bool listen_enabled=true) override
     {
-        INFO("ready():%s,connecting():%s,listening():%s",(ready())?"true":"false",(connecting())?"true":"false",(listening())?"true":"false");
-        if (!ready() && !connecting() && !listening())
+        const bool listening = WLAN_SMART_CONFIG_START;
+        INFO("ready():%s,connecting():%s,listening():%s",(ready())?"true":"false",(connecting())?"true":"false",(listening)?"true":"false");
+        if (!ready() && !connecting() && !listening)
         {
             bool was_sleeping = SPARK_WLAN_SLEEP;
 
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -117,7 +117,8 @@ class ManagedNetworkInterface : public NetworkInterface
 {
     volatile uint8_t WLAN_DISCONNECT;
     volatile uint8_t WLAN_DELETE_PROFILES;
-    volatile uint8_t WLAN_SMART_CONFIG_START;
+    volatile uint8_t WLAN_SMART_CONFIG_START; // Set to 'true' when listening mode is pending
+    volatile uint8_t WLAN_SMART_CONFIG_ACTIVE;
     volatile uint8_t WLAN_SMART_CONFIG_STOP;
     volatile uint8_t WLAN_SMART_CONFIG_FINISHED;
     volatile uint8_t WLAN_CONNECTED;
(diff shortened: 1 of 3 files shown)


system/src/system_network_internal.h
```diff
@@ -279,8 +279,10 @@ class ManagedNetworkInterface : public NetworkInterface
 
     void connect(bool listen_enabled=true) override
     {
-        INFO("ready():%s,connecting():%s,listening():%s",(ready())?"true":"false",(connecting())?"true":"false",(listening())?"true":"false");
-        if (!ready() && !connecting() && !listening())
+        // Do not try to connect if listening mode is active or pending
+        const bool listening = WLAN_SMART_CONFIG_START;
+        INFO("ready():%s,connecting():%s,listening():%s",(ready())?"true":"false",(connecting())?"true":"false",(listening)?"true":"false");
+        if (!ready() && !connecting() && !listening)
```
sergeuz: This change allows to skip a connection attempt not only when listening mode is active, but also when it's pending. Otherwise, a connection initiated from the listening loop (https://github.com/spark/firmware/blob/feature/listen_mode_delay_fix/system/src/system_network_internal.h#L212) cannot be cancelled quickly via `network_listen()` call, due to the event loop entering `manage_network_connection()` before `manage_smart_config()` on its next iteration and blocking the system thread for another connection attempt.

sergeuz:
I'm not sure how I can make a good automated test for this fix. Just in case, here is how I tested it manually:

* Configure the device with some bogus network credentials.

* Flash the following application to the device:
```
#include "application.h"

SYSTEM_MODE(AUTOMATIC)
SYSTEM_THREAD(ENABLED)

static void onButtonClick(system_event_t, int, void*) {
    WiFi.listen();
}

void setup() {
    System.on(button_click, onButtonClick);
}

void loop() {
}
```

* At this point, the device blinks green and repeatedly tries to connect to a network.

* Click setup button. The device should enter listening mode immediately. In this case, `WiFi.listen()` cancels a connection initiated by the event loop: https://github.com/spark/firmware/blob/feature/listen_mode_delay_fix/system/src/system_task.cpp#L118

* Provide the device with some bogus credentials once again. The device should start blinking green.

* Click setup button. The device should enter listening mode immediately. This time, `WiFi.listen()` cancels a connection initiated by the listening loop: https://github.com/spark/firmware/blob/feature/listen_mode_delay_fix/system/src/system_network_internal.h#L212

* Provide the device with valid credentials. The device should connect to the cloud normally.

technobly:
This works really well on Photon and Electron (Electron has one issue though - described below).  3 clicks to enter listening mode, and it will exit after 4 seconds. Tested with and without credentials on Photon.

```
#include "application.h"

SYSTEM_MODE(AUTOMATIC)
SYSTEM_THREAD(ENABLED)

void listening_mode_handler(system_event_t event, int elapsed_ms, void*) {
    // After 4 seconds in Listening Mode, manually force an exit and reset (optional)
    int timeout = 4 * 1000;
    // Serial1.printlnf("%d %d", elapsed_ms, timeout);
    if (elapsed_ms > timeout) {
        #if Wiring_WiFi
            WiFi.listen(false);
        #elif Wiring_Cellular
            Cellular.listen(false);
        #endif
    }
}

// Attach the listening_mode_handler at boot, to prevent being stuck in Listing Mode
STARTUP(System.on(setup_update, listening_mode_handler));

static void onButtonClick(system_event_t, int clicks, void*) {
    if (clicks == 3) {
        #if Wiring_WiFi
            WiFi.listen(true);
        #elif Wiring_Cellular
            Cellular.listen(true);
        #endif
    }
}

void setup() {
    System.on(button_final_click, onButtonClick); // 3 clicks to enter listening mode
}

void loop() {
}
```

A good automated test might include a test where the time between WiFi.listen() is called and WiFi.listening() = true.  And we'd say it's less than 2 seconds on Photon and Electron.

When testing the electron, I noticed if I cold booted it to force a longer connection process it would immediately enter Listening Mode after 3 clicks the first time.  However after it automatically exited and continued the connection process, if I tried 3 clicks a second time there was no response until after it connected.  Then the 3rd and 4th and 5th attempts all entered and exited Listening Mode as expected.

Issue #1059, opened 2016-07-11
Can TESTMODE pin 33 (PA8) on P1 be used as an extra P1S6 pin?


technobly:
It looks like this pin was overlooked as a useable pin, given that it's name was indicated as TESTMODE it was deemed a reserved pin.

A new pin `P1S6 - 30` definition would have to be added here:

https://github.com/spark/firmware/blob/develop/hal/src/stm32f2xx/pinmap_hal.c#L103

A definition here:

https://github.com/spark/firmware/blob/develop/hal/inc/pinmap_hal.h#L161

And TOTAL_PINS bumped to 31 for P1 here:

https://github.com/spark/firmware/blob/develop/hal/inc/pinmap_hal.h#L77

That should allow use of pinMode(P1S6, OUTPUT) and digitalWrite(P1S6, LOW)

This should ensure the pin gets set as an INPUT floating by default. change to **pin<=30**

https://github.com/spark/firmware/blob/develop/hal/src/stm32f2xx/core_hal_stm32f2xx.c#L314-L315

Since a TIM1_CH1 is on PA8, we will want to explore all of the timer based peripherals this may affect as well (PWM, Tone, Servo, Gen purpose timers, etc..)
## The P1 datasheet would need to be updated as well, or at minimum an issue logged to update it.

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated
- [x] Particle confirmed

technobly:
**Added above:**
This should ensure the pin gets set as an INPUT floating by default. change to **pin<=30**

https://github.com/spark/firmware/blob/develop/hal/src/stm32f2xx/core_hal_stm32f2xx.c#L314-L315

technobly:
After a deeper dig in implementing the above changes.  Using this application:

```
#include "application.h"

void doit() {
	pinMode(P1S6, OUTPUT);
	digitalWrite(P1S6, LOW);
}

STARTUP(doit())

void setup() {
	doit();
}

void loop() {
	digitalWrite(P1S6, !digitalRead(P1S6));
	delay(1000);
}
```
[...]

PR #1120, opened 2016-09-19
Added extra spare pin to P1 (P1S6) with GPIO and PWM support.

technobly:
Added extra spare pin to P1 (P1S6) with GPIO and PWM support.  Implements #1059 

>This is the TESTMODE pin 33 on the P1 module, and to use it properly the Wi-Fi Powersave Clock must be disabled.  See docs below.

---

Doneness:

- [ ] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation (for 0.7.0-rc.1)

https://github.com/spark/docs/commit/c63c4f0e03b6fce0361aa221ac8c6ddb329ab19a

### FEATURES

- [P1] Added extra spare pin to P1 (P1S6) with GPIO and PWM support.  Implements [#1059](https://github.com/spark/firmware/issues/1059)


hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -314,6 +319,11 @@ void HAL_Core_Config(void)
 #if PLATFORM_ID==8 // Additional pins for P1
     for (pin_t pin=24; pin<=29; pin++)
         HAL_Pin_Mode(pin, INPUT);
+    const uint8_t* data = (const uint8_t*)dct_read_app_data(DCT_RADIO_FLAGS_OFFSET);
+    uint8_t current = (*data);
+    if ((current&3) == 0x2) {
```
m-mcgowan: I know this is my (shoddy!) code, and felt the same way when I wrote it....it's probably best to pull these magic numbers into a function like

```
bool isWiFiPowersaveClockDisabled() {
 const uint8_t* data = (const uint8_t*)dct_read_app_data(DCT_RADIO_FLAGS_OFFSET);
  uint8_t current = (*data);
  return ((current&3) == 0x2);
}
```

user/tests/wiring/no_fixture/pwm.cpp
```diff
@@ -29,15 +29,26 @@
 
 uint8_t pwm_pins[] = {
 
-#if defined(STM32F2XX)
+#if (PLATFORM_ID == 0) // Core
```
m-mcgowan: lovely! I'm surprised we didn't hit the need to specify the PWM pins per platform before :-)

technobly: Yeah I don't know why, but we were not testing all PWM pins on all platforms before this change.  I was pleased to see it pass first time on all platforms though! :)

hal/src/stm32f2xx/core_hal_stm32f2xx.c
technobly: See commit https://github.com/spark/firmware/pull/1120/commits/b3f8160de6428b597b2d35f576303dd46677abfc for resurrection of a fairy ;-)

technobly:
Here's also the result of the power comparison when running the TEST=wiring/no_fixture test while measuring current on VUSB, with powersave clock enabled and disabled.  As seen there are no notable differences.
powersave

PR #1125, opened 2016-09-23
Uninitialize WICED's WLAN module when turning off the WiFi

### LINKED ISSUE #1098: P1 Stop mode using too much power
Im using a P1 with the 0.6.0rc1 firmware, system threading enabled and semi-automatic and am having an issue with the stop mode sleep drawing too much power. When stop mode is working as expected the power draw of my system is 2mA, but when its not working the power draw is 17mA.

I think I have narrowed down the issue to a particular case. Stop mode works as expected when particle.connect is called and actually connects to the cloud. But when particle.connect is called and the device is not within the wifi network, the system tires to connect (flashing green) and never connects to the wifi (because it cant). In this case when it goes into stop mode is when the power draw is 17mA.

I do call particle.disconnect, wifi.disconnect and the wifi.off prior to going into stop mode, but it seems like something is not getting powered down correctly due to the P1 trying to connect to wifi that has creds but it out of the wifi network.

Is there a way to force stopping the P1 from trying the wifi.creds?

does wifi.disconnect do anything if wifi.connecting is true?

---

Completeness:
- [x] Minimum test case added
- [X] Device, system and user firmware versions stated
- [x] Particle confirmed

sergeuz:
Fixes #1098

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### BUGFIX

- [Photon/P1] Previously, when entering Sleep-stop mode: `System.sleep(D1, RISING, 60);` while in the process of making a Wi-Fi connection resulted in some parts of the radio still being initialized, consuming about 10-15mA more than normal. Fixes [#1098](https://github.com/spark/firmware/issues/1098)

### DIFF
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -23,6 +23,7 @@
 #include "system_task.h"
 #include "system_cloud.h"
 #include "system_cloud_internal.h"
+#include "system_network_internal.h"
 #include "system_threading.h"
 #include "rtc_hal.h"
 #include "core_hal.h"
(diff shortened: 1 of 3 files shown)


system/src/system_sleep.cpp
```diff
@@ -43,10 +44,12 @@ WakeupState wakeupState;
 
 static void network_suspend() {
     // save the current state so it can be restored on wakeup
-    wakeupState.wifi = !SPARK_WLAN_SLEEP;
-    wakeupState.wifiConnected = wakeupState.cloud | network_ready(0, 0, NULL) | network_connecting(0, 0, NULL);
 #ifndef SPARK_NO_CLOUD
     wakeupState.cloud = spark_cloud_flag_auto_connect();
+#endif
+    wakeupState.wifi = !SPARK_WLAN_SLEEP;
+    wakeupState.wifiConnected = wakeupState.cloud || network_ready(0, 0, NULL) || network_connecting(0, 0, NULL);
```
sergeuz: I've noticed that `wakeupState.cloud` was used before initialization previously.

Issue #1037, opened 2016-06-17
CLI-assisted logging


sergeuz:
As we discussed previously, it would be nice to have possibility to enable logging dynamically, via the cloud or over the wire. Taking into account that all currently implemented [log handlers](https://github.com/spark/firmware/blob/develop/docs/reference/firmware.md#log-handlers) already require serial connection with the host, it seems natural to start with the USB logging first, and then reuse the same configuration code for cloud logging.

The plan is to use [vendor-specific USB requests](https://github.com/spark/firmware/pull/1010) (already in `develop`) to configure logging dynamically, without any prior configuration done at application side, and without entering the listening mode, which affects normal workflow of the system and application code greatly. The CLI then can be used to attach to already running system, for example:

```
# Start logging with default settings (uses USBSerial1)
particle log

# Enable verbose logging
particle log all

# Enable verbose logging for application code only
particle log warn app=all

# Use hardware serial and UART to USB converter for logging
particle log -s serial1 -d /dev/ttyUSB0
```

On any of the above commands the CLI sends necessary configuration requests to the device and then starts to read and display a nicely formatted and highlighted logging output.

Implementation-wise, the `LogManager` class needs to be extended to support named and dynamically allocated log handlers. Few factory methods also need to be added in order to allocate log handlers based on their requested type and settings. A trivial JSON formatter for generated logs would also simplify parsing at the CLI side.

Configuration request might look like the following:

``` js
{
  "op": "addLogHandler",
  "handler": { // Handler settings
    "type": "JSONLogHandler"
  },
  "stream" : { // Output stream
    "type": "Serial1",
    "baud": 57600
  },
  "level": "warn", // Default logging level
  "filters": {
    "app": "trace" // Category-specific logging level
  }
}
```

I used JSON here for simplicity, though @avtolstoy confirmed that we can use it for vendor-specific requests with no problems, assuming that request size limitations are met.

---

Completeness:
- [ ] Minimum test case added
- [ ] Device, system and user firmware versions stated
- [ ] Particle confirmed

m-mcgowan:
This is a great start! For the CLI interface, I suggest we follow the same syntax as used for the `flash` command to support both local and remote devices:

```
particle log --usb  ...
```
and
```
particle log my_device ...
```

[...]

PR #1126, opened 2016-09-23
Common tools for the Wiring library and unit testing

sergeuz:
This PR is a part 1 of 2 of the code adding support for [runtime logging configuration](https://github.com/spark/firmware/pull/1127), and contains various common tools which I think can be reviewed separately, in order to simplify overall review process. Some background on the topic can be found here: #1037.

This PR adds the following:

**Vector container** ([spark_wiring_vector.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/wiring/inc/spark_wiring_vector.h#L50))

The problem is that we can't really use STL containers in our firmware (though we do use `std::vector` in some places), as none of them are designed to work with disabled exceptions: it's not possible to write custom allocator that doesn't use exceptions, and standard APIs do not provide any alternative mechanism for error handling. Logging framework uses `spark::Vector` extensively, e.g. to maintain a prefix tree for category-based filtering, list of active log handlers, etc.

**Application-side JSON parser/writer classes** ([spark_wiring_json.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/wiring/inc/spark_wiring_json.h#L53))

These classes are mostly wrappers for jsmn library, which is a part of the system module, with added support for type conversions and escaped strings (latter is essential for [JSON log message formatter](https://github.com/spark/firmware/blob/feature/usb_logging_2_of_2/wiring/src/spark_wiring_logging.cpp#L517)). It's also quite handy to use JSON for parameterized [acceptance tests](https://github.com/spark/firmware/blob/feature/usb_logging_2_of_2/user/tests/accept/features/logging/log_config.feature#L107). Usage example can be found [here](https://github.com/spark/firmware/blob/feature/usb_logging_2_of_2/user/tests/accept/features/logging/log_config_app/app.cpp).

**Helper tools for unit testing**
- [tools/random.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/random.h) - random number/string generation.
- [tools/string.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/string.h) - string utilities.
- [tools/stream.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/stream.h) - Wiring-compatible output stream.
- [tools/buffer.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/buffer.h) - wrapper class for a buffer with padding bytes.
- [tools/alloc.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/alloc.h) - allocator class allowing to detect memory usage errors.
- [tools/check.h](https://github.com/spark/firmware/blob/feature/usb_logging_1_of_2/user/tests/unit/tools/check.h) - mixin for "checkable" types. Example usage:

``` cpp
#include "tools/stream.h"

TEST_CASE() {
    test::OutputStream stream;
    uint8_t buf[] = { 1, 2, 3, 255 };
    stream.write(buf, sizeof(buf));
    check(stream).hex().equals("010203ff");
}

TEST_CASE() {
    std::string s = "one,two,three";
    check(s).matches("(.+),(.+),(.+)")
            .at(0).equals("one")
            .at(1).equals("two")
            .at(2).equals("three");
}
```

---

Doneness:
- [x] Code peer reviewed
- [x] API tests compiled

N/A see part 2 of 2 #1127 
- [x] Add documentation

### DIFF
--- a/wiring/src/spark_wiring_string.cpp
+++ b/wiring/src/spark_wiring_string.cpp
@@ -72,6 +72,12 @@ String::String(const char *cstr)
 	if (cstr) copy(cstr, strlen(cstr));
 }
 
+String::String(const char *cstr, unsigned int length)
+{
+	init();
+	if (cstr) copy(cstr, length);
+}
+
 String::String(const String &value)
 {
 	init();
--- a/ci/build_boost.sh
+++ b/ci/build_boost.sh
@@ -1,4 +1,4 @@
 pushd $BOOST_ROOT
 ./bootstrap.sh
-./b2  --with-thread  --with-system --with-program_options --with-random  --threading=single
+./b2  --with-thread  --with-system --with-program_options --with-random --with-regex --threading=single
 popd
--- a/ci/install_boost.sh
+++ b/ci/install_boost.sh
@@ -9,4 +9,5 @@ test -d $BOOST_ROOT || (
    mv boost_$BOOST_VERSION  $BOOST_HOME
 ) 
 test -d $BOOST_ROOT || ( echo "boost root $BOOST_ROOT not created." && exit 1)
-export DYLD_LIBRARY_PATH=$BOOST_ROOT/stage/lib
+export DYLD_LIBRARY_PATH="$BOOST_ROOT/stage/lib:$DYLD_LIBRARY_PATH"
+export LD_LIBRARY_PATH="$BOOST_ROOT/stage/lib:$LD_LIBRARY_PATH"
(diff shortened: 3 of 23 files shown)


wiring/src/spark_wiring_string.cpp
```diff
@@ -72,6 +72,12 @@ String::String(const char *cstr)
 	if (cstr) copy(cstr, strlen(cstr));
 }
 
+String::String(const char *cstr, unsigned int length)
```
technobly: Please add a `wiring/no_fixture/string.cpp` test for String::String(const char *cstr, unsigned int length)

PR #1127, opened 2016-09-23
Runtime logging configuration

sergeuz:
This PR adds support for runtime logging configuration, which allows to enable logging on already running system via USB control requests. Some background on the topic can be found here: #1037. Note that this PR is based on the common code from another branch which is being reviewed here: https://github.com/spark/firmware/pull/1126

Parameters that can be configured dynamically: destination stream (`Serial`, `Serial1`, `USBSerial1`), message format (plain text, JSON), message filtering (level and category-based).

This PR also adds basic tools for acceptance testing, such as handling of serial IO, USB requests, application flashing, etc. The test utilizing above functionality can be found [here](https://github.com/spark/firmware/blob/feature/usb_logging_2_of_2/user/tests/accept/features/logging/log_config.feature).

I'll be adding further comments directly in the code to provide more details on the implementation.

---

Doneness:
- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### Features

- Added support for runtime logging configuration, which allows to enable logging on already running system via USB control requests.


user/tests/accept/init_env
```diff
@@ -0,0 +1,76 @@
...
+      USB_VENDOR_ID=2b04
+      USB_PRODUCT_ID=c006
+      ;;
+    p1)
+      USB_VENDOR_ID=2b04
+      USB_PRODUCT_ID=d008
+      ;;
+    electron)
+      USB_VENDOR_ID=2b04
+      USB_PRODUCT_ID=d00a
```
avtolstoy: This should be c00a (otherwise it targets Electron in DFU mode)

user/tests/accept/init_env
```diff
@@ -0,0 +1,76 @@
...
+      USB_VENDOR_ID=1d50
+      USB_PRODUCT_ID=607d
+      ;;
+    photon)
+      USB_VENDOR_ID=2b04
+      USB_PRODUCT_ID=c006
+      ;;
+    p1)
+      USB_VENDOR_ID=2b04
+      USB_PRODUCT_ID=d008
```
avtolstoy: This should be `c008` (otherwise it targets P1 in DFU mode)

user/tests/accept/init_env
sergeuz: Thanks! Honestly, I don't remember from where I copied these IDs. Since you're working with this branch already, would you mind committing correct IDs?

PR #1130, opened 2016-09-30
Bugfix for subscriptions checksum calculation

sergeuz:
`handler.scope` is an enum, and we want its address to be casted to `uint8_t*`, not its value

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### BUGFIX

- Particle.subscribe() used with same events but changing scope between PUBLIC and PRIVATE or vice versa would potentially result in non-registered subscriptions.  This was also crashing the GCC virtual device with a segfault when subscription checksums were calculated.


monkbroc:
This also crashed the virtual device with a segfault

technobly:
Thanks, I added that to the bugfix note.

PR #1135, opened 2016-10-06
Feature/time isvalid

### LINKED ISSUE #965: current time not valid in `setup()`
The cloud handshake sends a request for the current time from the cloud, but doesn't wait for the response, so that any code in `setup()` that attempts to access the current time may not see the correct time. 

This is particularly true with UDP comms, where the time request may need to be resent until a response is received. 

Tests are needed to also validate that the timezone set before the cloud connection is maintained after the cloud connection.

---

Completeness:
- [x] Minimum test case added
- [x] Particle confirmed

### LINKED ISSUE #1116: As a developer, I can determine if the device's current time is correct
When a device boots up, the current time is set to some default, which of course is not correct at all.

Once the device receives the current time from the cloud, the time is then correct (to within a few seconds, due to network latency.)  The time typically remains correct while the device is powered, or sleeps. 

The system firmware should track if the current time is valid:
- when the time is received from the cloud it is valid
- when the device sleeps and the RTC maintains the time. it is valid

IIRC the default time was some time in the past. The test for validity may be as simple as the time being later than the year 2015, or it may be involved, tracking when the time is set and that it has been maintained. 
## Related to #965 

Completeness:
- [x] Minimum test case added
- [x] Particle confirmed

avtolstoy:
Fixes issues #1116 and #965

Wiring functions added:
- `Time.isValid()`
- `Particle.syncTimePending()`
- `Particle.syncTimeDone()`
- `Particle.timeSyncedLast()`

This PR also fixes an issue with how RTC on Photon and Electron is configured. The year may only be double digit (00 - 99).

Tests added to `wiring/no_fixture`:
- TIME_13_syncTimePending_syncTimeDone_when_disconnected
- TIME_14_timeSyncedLast_works_correctly
- TIME_15_SyncTimeInAutomaticMode
- TIME_16_SyncTimeInManualMode
- TIME_17_RestoreSystemMode
- TIME_18_TimeChangedEvent

New functions documentation copied from `firmware.md` :

> ### isValid()
> 
> Used to check if current time is valid. This function will return `true` if:
> - Time has been set manually using [`Time.setTime()`](#settime-)
> - Time has been successfully synchronized with the Particle Cloud. The device synchronizes time with the Particle Cloud during the handshake. The application may also manually synchronize time with Particle Cloud using [`Particle.syncTime()`](#particle-synctime-)
> - Correct time has been maintained by RTC.{{#unless core}} See information on [`Backup RAM (SRAM)`](#backup-ram-sram-) for cases when RTC retains the time. RTC is part of the backup domain and retains its counters under the same conditions as Backup RAM.{{/unless}}
> 
> **NOTE:** When {{device}} is running in `AUTOMATIC` mode {{#unless core}}and threading is disabled {{/unless}} this function will block if current time is not valid and there is an active connection to Particle Cloud. Once {{device}} synchronizes the time with Particle Cloud or the connection to Particle Cloud is lost, `Time.isValid()` will return its current state. This function is also implicitly called by any `Time` function that returns current time or date (e.g. `Time.hour()`/`Time.now()`/etc).
> ### Particle.syncTimeDone()
> 
> Returns `true` if there is no `syncTime()` request currently pending or there is no active connection to Particle Cloud. Returns `false` when there is a pending `syncTime()` request.
> ### Particle.syncTimePending()
> 
> Returns `true` if there a `syncTime()` request currently pending. Returns `false` when there is no `syncTime()` request pending or there is no active connection to Particle Cloud.
> ### Particle.timeSyncedLast()
> 
> Used to check when time was last synchronized with Particle Cloud.
> 
> ``` C++
> // SYNTAX
> Particle.timeSyncedLast();
> Particle.timeSyncedLast(timestamp);
> ```
> 
> Returns the number of milliseconds since the device began running the current program when last time synchronization with Particle Cloud was performed.
> 
> This function takes one optional argument:
> - `timestamp`: `time_t` variable that will contain a UNIX timestamp received from Particle Cloud during last time synchronization
> 
> ``` C++
> // EXAMPLE
> #define ONE_DAY_MILLIS (24 * 60 * 60 * 1000)
> 
> void loop() {
>   time_t lastSyncTimestamp;
>   unsigned long lastSync = Particle.timeSyncedLast(lastSyncTimestamp);
>   if (millis() - lastSync > ONE_DAY_MILLIS) {
>     unsigned long cur = millis();
>     Serial.printlnf("Time was last synchronized %lu milliseconds ago", millis() - lastSync);
>     if (lastSyncTimestamp > 0)
>     {
>       Serial.print("Time received from Particle Cloud was: ");
>       Serial.println(Time.timeStr(lastSyncTimestamp));
>     }
>     // Request time synchronization from Particle Cloud
>     Particle.syncTime();
>     // Wait until {{device}} receives time from Particle Cloud (or connection to Particle Cloud is lost)
>     waitUntil(Particle.syncTimeDone);
>     // Check if synchronized successfully
>     if (Particle.timeSyncedLast() >= cur)
>     {
>       // Print current time
>       Serial.println(Time.timeStr());
>     }
>   }
> }
> ```

EDIT:
- `time_changed` event added
- When resuming session on Electron if for some reason time was not synchronized during handshake, send time request again

---

Doneness:
- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### Features

- New Time API's! `Time.isValid()` | `Particle.syncTimePending()` | `Particle.syncTimeDone()` | `Particle.timeSyncedLast()` [[Fixes #1116]](https://github.com/spark/firmware/issues/1116) [[Fixes #965]](https://github.com/spark/firmware/issues/965)


m-mcgowan:
Very thoroughly documented!

Some thoughts:

- Is there a system event for time changes?  That way, application code can be actively notified when the time changes during a sync. 

- `isValid()` is blocking - what's the rationale for that? I would have expected it not to block since it's only a sniff of the current status.  I would have imagined `getTime()` to block since then we are trying to get the active time, which might be in flux.  We might also add an overload that disables any blocking and just returns the current time as it is. 

-  What's the distinguishing factor between syncTimeDone and syncTimePending? As far as I can tell, they are inverses of each other.  That's also fine, since`waitUntil` doesn't have an easy way to invert predicates, but a description of the expected use cases would be helpful. 

- On the electron, it's possible that `Particle.syncTime()` never syncs if all server CoAP responses are lost. The system should try to do what it can to ensure the sync will eventually happen.  This is possibly outside the scope of this PR, and something to address elsewhere.

avtolstoy:
@m-mcgowan 
Thanks for comments!

> Is there a system event for time changes? That way, application code can be actively notified when the time changes during a sync.
I like the idea of the system event for time changes. Should an event be fired if time is set manually using `Time.setTime()`?

> `isValid()` is blocking - what's the rationale for that? I would have expected it not to block since it's only a sniff of the current status. I would have imagined `getTime()` to block since then we are trying to get the active time, which might be in flux. We might also add an overload that disables any blocking and just returns the current time as it is.
`Time.isValid()` is only blocking in AUTOMATIC mode with threading disabled. The rationale was that it's internally used by `Time.now()` (and as a consequence by all the other functions like `Time.hour()` etc) and in AUTOMATIC mode we want to ensure that when `Time` functions are accessed, the time is already valid. It also only blocks _if there is a pending_ `syncTime()` request.

> What's the distinguishing factor between syncTimeDone and syncTimePending? As far as I can tell, they are inverses of each other. That's also fine, since`waitUntil` doesn't have an easy way to invert predicates, but a description of the expected use cases would be helpful.
Exactly as you say, they are inverses of each other to make it easier to use them with `waitFor()/waitUntil()`. The documentation lists an example usage for both of those functions.

> On the electron, it's possible that Particle.syncTime() never syncs if all server CoAP responses are lost. The system should try to do what it can to ensure the sync will eventually happen. This is possibly outside the scope of this PR, and something to address elsewhere.
According to how the reliable channel is implemented (regarding confirmable requests), when several retransmissions occur and there is no reply the cloud connection is reset. I think the only issue is that when Electron tries to connect again  and resumes the session instead of performing a handshake it will not issue a new `syncTime()` request. Perhaps we should add a check there that if time is not valid and `Particle.timeLastSynced()` reports 0, a time synchronization request should be sent again? If the application is manually calling `Particle.syncTime()` and wants to ensure that the time request succeeded, `Particle.timeLastSynced()` can be used to check for that. There is an example in the docs.

m-mcgowan:
> I like the idea of the system event for time changes. Should an event be fired if time is set manually using Time.setTime()
Yes, I think so. The important thing is that time as changed. We might also include a reason for change as part of the event so the application can distinguish external changes from internal updates via `Time.setTime()`. 

> I think the only issue is that when Electron tries to connect again and resumes the session instead of performing a handshake it will not issue a new syncTime() request. Perhaps we should add a check there that if time is not valid and Particle.timeLastSynced() reports 0, a time synchronization request should be sent again?
I like the idea of synching the time if it's not valid once the cloud is connected.

system/src/system_cloud.cpp
```diff
@@ -87,6 +87,18 @@ bool spark_sync_time(void *reserved)
...
 
+bool spark_sync_time_pending(void* reserved)
+{
+    SYSTEM_THREAD_CONTEXT_SYNC(spark_sync_time_pending(reserved));
+    return spark_protocol_time_request_pending(sp, nullptr);
+}
+
+system_tick_t spark_sync_time_last(time_t* tm, void* reserved)
+{
+    SYSTEM_THREAD_CONTEXT_SYNC(spark_sync_time_pending(reserved));
```
technobly: Should this be `SYSTEM_THREAD_CONTEXT_SYNC(spark_sync_time_last(time_t* tm, void* reserved));`

communication/src/spark_protocol_functions.cpp
```diff
@@ -252,4 +263,15 @@ int spark_protocol_command(ProtocolFacade* protocol, ProtocolCommands::Enum cmd,
...
+{
+    (void)reserved;
+    return protocol->time_request_pending();
+}
+system_tick_t spark_protocol_time_last_synced(ProtocolFacade* protocol, time_t* tm, void* reserved)
+{
+    (void)reserved;
+    return protocol->time_last_synced(tm);
+}
+
```
technobly: Should this be `SparkProtocol` instead of `ProtocolFacade` here?

avtolstoy: Good catch. The two functions above should also be changed:

```c++
int spark_protocol_set_connection_property(ProtocolFacade* protocol, unsigned property_id,
                                           unsigned data, void* datap, void* reserved)
{
```
```c++
int spark_protocol_command(ProtocolFacade* protocol, ProtocolCommands::Enum cmd, uint32_t data, void* reserved)
```

system/src/system_cloud.cpp
avtolstoy: It should! We should also update #1146 after merging, so that this would have been caught.

PR #1140, opened 2016-10-17
Electron modem USART and buffer handling fixes

### LINKED ISSUE #1104: Electron socketSend max buffer size logic 1 byte too small for 1024 byte chunks
Please see community discussion here: https://community.particle.io/t/solved-tcpclient-data-transmission-rates/25255/7

``` c++
int MDMParser::socketSend(int socket, const char * buf, int len)
{
    int cnt = len;
    while (cnt > 0) {
        int blk = USO_MAX_WRITE;
        if (cnt < blk)
            blk = cnt;
        bool ok = false;
        {
            LOCK();
            if (ISSOCKET(socket)) {
                sendFormated("AT+USOWR=%d,%d\r\n",_sockets[socket].handle,blk);
                if (RESP_PROMPT == waitFinalResp()) {
                    HAL_Delay_Milliseconds(50);
                    send(buf, blk);
                    if (RESP_OK == waitFinalResp())
                        ok = true;
                }
            }
            UNLOCK();
        }
        if (!ok)
            return MDM_SOCKET_ERROR;
        buf += blk;
        cnt -= blk;
    }
    return (len - cnt);
}
```

From the community:

> Basically, USO_MAX_WRITE is 1024, so if you send a chunk of 1024 bytes, it fails the cnt < blk test. Things go bad from there. So use 1023. Or 512. Or anything less than 1024 (but not equal to).

When 1024 is used, blk gets set to USO_MAX_WRITE and failing the cnt < blk test is inconsequential.

Note that USOWR max buffer sizes are:

```
Number of data bytes to write:
• Base syntax normal mode: range 0-1024
• Base syntax HEX mode: range 0-512
• Binary extended syntax: range 0-1024
```

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (electron, 0.6.0-rc.1 or 0.5.3-rc.2 would have this issue)
- [x] Particle confirmed

### LINKED ISSUE #1138: [Electron] AT commands with long responses not completing
Please refer to this community post: https://community.particle.io/t/electron-serious-issue-with-ublox-at-communication-for-retrieving-large-result-sets/22131

Tracking internally: FIRM-333, AE-193

// List all available AT commands with AT+CLAC
Use the AT command mode of the troubleshooting app to send AT+CLAC ... wait a bit and send AT to get the result of the buffer.
https://gist.github.com/technobly/5ff967632eeaab19128d84c805f20f10

Or with debugging enabled, try adding this to the end of `MDMParser:_powerOn()` to have more direct control over URCs:

``` c++
    // List all available AT commands with AT+CLAC
    sendFormated("AT+CLAC\r\n");
    if(RESP_OK != waitFinalResp(NULL,NULL,60000);
        goto failure;
```

Response should be seen in the log output.

Suspected issue with electronserialpipe_hal buffer size or hardware interaction with modem USART.

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (all)
- [x] Particle confirmed

avtolstoy:
Closes #1138 and #1104.

Fixes several issues:
- `USART_FLAG_TXE` was not checked when writing into DR register from system thread which caused an unsent data byte to be overwritten
- RX flow control was not correctly handled. When an internal `ElectronSerialPipe` buffer overflows, a data byte should be left in DR register to de-assert RTS line to ask modem to pause transmission for a while
- `ElectronSerialPipe` circular buffer was initialized to 1024 bytes, but the implementation allowed only 1023 bytes to be held inside the buffer
- `AT+CLAC`-like commands weren't parsed correctly (thanks @technobly)

Relevant tests:
- `MDM_01_socket_writes_with_length_more_than_1023_work_correctly`
- `MDM_02_at_commands_with_long_response_are_correctly_parsed_and_flow_controlled`

---

Doneness:
- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### Bugfixes

- [Electron] Fixed modem USART and buffer handling [[Fixes #1138]](https://github.com/spark/firmware/issues/1138) [[Fixes #1104]](https://github.com/spark/firmware/issues/1104)

### DIFF
--- a/hal/src/electron/modem/electronserialpipe_hal.cpp
+++ b/hal/src/electron/modem/electronserialpipe_hal.cpp
@@ -170,7 +170,7 @@ int ElectronSerialPipe::put(const void* buffer, int length, bool blocking)
 
 void ElectronSerialPipe::txCopy(void)
 {
-    if (_pipeTx.readable()) {
+    if (_pipeTx.readable() && USART_GetFlagStatus(USART3, USART_FLAG_TXE)) {
         char c = _pipeTx.getc();
         USART_SendData(USART3, c);
     }
@@ -214,11 +214,17 @@ int ElectronSerialPipe::get(void* buffer, int length, bool blocking)
 
 void ElectronSerialPipe::rxIrqBuf(void)
 {
-    char c = USART_ReceiveData(USART3);
-    if (_pipeRx.writeable())
+    if (_pipeRx.writeable()) {
+        char c = USART_ReceiveData(USART3);
         _pipeRx.putc(c);
-    else
-        /* overflow */;
+    } else {
+        /* TODO: Handle overflow?
+         * We should not read USART3->RDR register in order to
+         * signal to the modem that we are not ready to receive,
+         * then try to process and free the buffer.
+         */
+        (void)USART_ReceiveData(USART3);
+    }
 }
 
 /*******************************************************************************
@@ -238,5 +244,9 @@ extern "C" void HAL_USART3_Handler(void)
     if(USART_GetITStatus(USART3, USART_IT_TXE) != RESET)
     {
         electronMDM.txIrqBuf();
+    }
+
+    if (USART_GetFlagStatus(USART3, USART_FLAG_ORE) != RESET) {
+        (void)USART_ReceiveData(USART3);
     }
 }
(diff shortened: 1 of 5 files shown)


technobly:
@avtolstoy any comment on my question above?  Let me know if you need me to verify tests on a particular electron.

PR #1160, opened 2016-11-07
[Electron] Modem serial pausing before going into sleep with SLEEP_NETWORK_STANDBY

avtolstoy:
With introduction of proper flow control handling via RTS line in PR #1140, modem should be unconditionally paused (by deasserting RTS line) before going into sleep with `SLEEP_NETWORK_STANDBY`. This should allow Electron not to lose any URCs from the modem while it is sleeping.

This PR includes a cucumber-based acceptance test which demonstrates how a sleepy Electron can receive a publish posted while it was sleeping:
- `$ cd user/tests/accept`
- `$ bundler install`
- `$ PLATFORM=electron ./run electron_modem_pause`

Includes PR #1097

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### Enhancements

- [Electron] Modem USART paused via HW_FLOW_CONTROL (RTS) before going into sleep with SLEEP_NETWORK_STANDBY. Receives and buffers small messages while system sleeping.


system/src/system_sleep.cpp
```diff
@@ -150,12 +155,28 @@ int system_sleep_pin_impl(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long sec
     {
         network_suspend();
     }
+
+#if PLATFORM_ID==PLATFORM_ELECTRON_PRODUCTION
+    if (!network_sleep_flag(param)) {
+        // Pause the modem Serial
+        electronMDM.pause();
```
technobly: This should be called via a wrapper function in cellular_hal

system/src/system_sleep.cpp
```diff
@@ -150,12 +155,28 @@ int system_sleep_pin_impl(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long sec
...
     HAL_Core_Enter_Stop_Mode(wakeUpPin, edgeTriggerMode, seconds);
     if (network_sleep)
     {
         network_resume();   // asynchronously bring up the network/cloud
     }
+
+#if PLATFORM_ID==PLATFORM_ELECTRON_PRODUCTION
+    if (!network_sleep_flag(param)) {
+        // Pause the modem Serial
+        electronMDM.resume();
```
technobly: This should also be called via a wrapper function in cellular_hal

technobly:
Since this does change the current behavior, we should add some Docs that explain what happens with received data now when the modem remains powered up while the system sleeps.  The system has the ability to receive a message that was buffered by the modem during STM32 sleep.  If too much data is received (how much?), what happens? (excessive data is lost or does it overwrite in circular fashion?).

avtolstoy:
Updated

PR #1167, opened 2016-11-11
Workaround for 100% CPU usage problem on the virtual device

### LINKED ISSUE #1036: gcc virtual device runs at 100% cpu
The gcc virtual device emulates an embedded device, where the CPU is being given work to do all the time. This carries over to the cross compiled version, where it consumes 100% of the executing cpu core on a desktop machine.

Some steps to mitigate this will make the cross compile device more scalable and environmentally friendly.

---

Completeness:
- [x] Minimum test case added
- [x] Particle confirmed

sergeuz:
This PR fixes #1036 by suspending main thread for some minimum time on every Nth iteration of the application loop. That's probably not the best solution possible, but it allowed to reduce CPU usage from 100 to 10-15% on my machine (Intel Core i7), while keeping reasonable performance comparing to the real devices.

Note that calling `std::this_thread::yield()` has appeared to be insufficient to workaround the problem, also a simple unconditional sleep on each iteration makes an application significantly slower comparing to the same application running on the real device (at least 7 times slower in my tests).

Below is the simple test application that can be used to check how many times application loop is being invoked per second:

```cpp
#include "application.h"

SYSTEM_MODE(MANUAL)

SerialLogHandler logHandler(LOG_LEVEL_WARN, {
    { "app", LOG_LEVEL_ALL }
});

uint32_t t = 0;
unsigned n = 0;

void setup() {
    t = millis();
}

void loop() {
    ++n;
    if (millis() - t >= 1000) {
        Log.info("%u", n);
        t = millis();
        n = 0;
    }
}
```

---

Doneness:

- [x] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation
…atform

### DIFF
--- a/services/inc/config.h
+++ b/services/inc/config.h
@@ -37,4 +37,10 @@
 #define DEFAULT_SEC_INACTIVITY          0
 #define DEFAULT_SEC_NETOPS              20
 
+// Number of application loop iterations after which the main thread is suspended for some short time.
+// This is primarily used to workaround 100% CPU usage on the virtual device platform
+#ifndef SUSPEND_APPLICATION_THREAD_LOOP_COUNT
+#define SUSPEND_APPLICATION_THREAD_LOOP_COUNT 25
+#endif
+
 #endif /* CONFIG_H_ */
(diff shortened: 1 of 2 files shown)


m-mcgowan:
Great writeup and good work around for the time being. 👍 

So...can we remove the hard-coded value of 25 loops and push that into a define? then we have a way of tweaking the number for different use cases.

PR #1169, opened 2016-11-14
update comms logs with new logging API

technobly:
Updated system communication logging with new logging API
    
- makes logs available by default without compiling with DEBUG_BUILD=y
- improves logging in MBEDTLS to include \r\n after messages
- added INFO logs around main parts of comms process

Comments welcome on changes to logging categories.  Some categories were used in a more local scope such as "handshake" to reduce the size of the log messages (each had "Handshake: " on them before).

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### Enhancement

- Updated system communication logging with new logging API


communication/lib/mbedtls/library/debug.c
```diff
@@ -70,7 +70,7 @@ static inline void debug_send_line( const mbedtls_ssl_context *ssl, int level,
      */
 #if defined(MBEDTLS_THREADING_C)
     char idstr[20 + DEBUG_BUF_SIZE]; /* 0x + 16 nibbles + ': ' */
-    mbedtls_snprintf( idstr, sizeof( idstr ), "%p: %s", ssl, str );
+    mbedtls_snprintf( idstr, sizeof( idstr ), "%p: %s\r\n", ssl, str );
```
technobly: I did originally do that via `debug_send_line`, but there are some cases where it's not appropriate, like a buffer dump.  So the higher level API needed modification instead.  The above change to line 73 I believe was residual testing that should have been removed (MBEDTLS_THREADING_C is not currently defined).  If we add `\r\n` to `my_debug` most of the log messages would end with `\n\r\n` adding a bunch of space to the log output.  I agree we should try to minimize changes to 3rd party libraries, but we also need to have effective logs.  I think these changes should merge with upstream MBEDTLS library changes without conflict.  I'll revert the change on line 73.

communication/src/dtls_message_channel.cpp
```diff
@@ -17,17 +17,19 @@
  ******************************************************************************
  */
 
+#include "logging.h"
+LOG_SOURCE_CATEGORY("message_channel.dtls")
```
technobly: I was hoping to not have to hard code the parent category, but I do like the idea of knowing the logs are a sub category of the communication library. Is it fair to say all files within the communication/ folder will have a default parent category of "comm" ?  That's probably a better way of keeping log messages filterable by "comm" as well ;-)  I'll make this change in a couple places.

communication/src/protocol_defs.h
```diff
@@ -18,20 +18,33 @@ namespace particle { namespace protocol {
...
+	/* 17 */ IO_ERROR_GENERIC_ESTABLISH,
+	/* 18 */ IO_ERROR_GENERIC_RECEIVE,
+	/* 19 */ IO_ERROR_GENERIC_SEND,
+	/* 20 */ IO_ERROR_GENERIC_MBEDTLS_SSL_WRITE,
+	/* 21 */ IO_ERROR_DISCARD_SESSION,
+	/* 21 */ IO_ERROR_LIGHTSSL_BLOCKING_SEND,
+	/* 22 */ IO_ERROR_LIGHTSSL_BLOCKING_RECEIVE,
+	/* 23 */ IO_ERROR_LIGHTSSL_RECEIVE,
+	/* 24 */ IO_ERROR_LIGHTSSL_HANDSHAKE_NONCE,
+	/* 25 */ IO_ERROR_LIGHTSSL_HANDSHAKE_RECV_KEY,
```
technobly: Thanks for this!  I'll add that with a note that if the ProtocolError's get expanded they also need to be added to the `system_error particle::protocol::toSystemError(ProtocolError error) {` switch.

communication/src/dtls_message_channel.cpp
sergeuz: Yes, if category name is not defined explicitly for a log statement's scope, upper level category name is used, up to category specified at module level (which is "comm" for communication/ directory). I didn't find a way to make categories easy to define and override, and also maintain nesting automatically at compile time.

communication/lib/mbedtls/library/debug.c
sergeuz: Ideally, new line sequence should be specific to a log handler, as "\r\n", for example, makes sense only for serial logging. We can make the logging framework perform new line conversions on the fly, but, surely, that's not something that needs to be discussed in the scope of this PR.

PR #1182, opened 2016-11-21
Listening Mode Timeout defaults and set|get API [ch318]

### LINKED ISSUE #687: Electrons flashing blue.
Flashing blue is supposed to mean no SIM card, but when running 0.0.2 and tinker with taoglas patch antenna both 2G & 3G units appear to enter this mode. Note that devices running identical code, but connected to external antennas do not exhibit this problem.

One or more resets (no power cycle, no reseat of SIM card) brings the device back to life.

Will try and capture serial logs to enable analysis.

technobly:
This PR addresses an issue with Electrons primarily where they can start flashing blue (Listening Mode) due to reconnects that result in no SIM found.  This can happen if there is significant shock and vibration on a loose or oxidized SIM contact during a power cycle.  Upon exiting Listening Mode, a power cycle and reset of the modem will occur if the SIM is not found.  If it continues to be unresponsive, Listening Mode will be re-entered.  Additionally, previously once in Listening Mode required user intervention to exit.  Now there is a default 5 minute timeout applied to Cellular devices only.  Wi-Fi devices continue to have the same behavior with no timeout from Listening Mode, and both Cellular and Wi-Fi timeouts are user configurable.

Also added an API `WiFi.set|getListenTimeout()` | `Cellular.set|getListenTimeout()` to override the automatic new Listening Mode timeout (Wi-Fi = no timeout by default, Cellular = 5 minute timeout by default).  [[Fixes #687]](https://github.com/spark/firmware/issues/687) [[Docs]](https://prerelease-docs.particle.io/reference/firmware/electron/#setlistentimeout-)

See included Docs included in PR for more info (link above will not be live until 0.7.0-rc.1 is released)

Tests added `TEST=wiring/threading` and `TEST=wiring/api`

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### Features

- Added `WiFi.set|getListenTimeout()` | `Cellular.set|getListenTimeout()` to override the automatic new Listening Mode timeout (Wi-Fi = no timeout by default, Cellular = 5 minute timeout by default).  [[Fixes #687]](https://github.com/spark/firmware/issues/687) [[Docs]](https://prerelease-docs.particle.io/reference/firmware/electron/#setlistentimeout-)


user/tests/wiring/threading/application.cpp
```diff
@@ -23,45 +23,14 @@
...
-	app->invoke_async(fn);
-
-	// test value not incremented
-	assertEqual((int)test_val, 0);
-
-    Particle.process();
-
-    // validate the function was called.
-    assertEqual((int)test_val, 1);
-}
```
sergeuz: These 2 test cases are not related to listening mode, so instead of moving them to `listening.cpp` I would prefer them to be defined in some separate file.

PR #1184, opened 2016-11-27
Move cellular HAL to system-part3

sergeuz:
This PR moves cellular HAL and its direct dependencies from part2 to part3 on Electron. On my system (gcc 5.3.1) this allows to reduce part2 by ~18K:

Before applying this patch (with debugging enabled):
```
  13724	    304	   1460	  15488	   3c80	../../../build/target/system-part1/platform-10-m/system-part1.elf
 127348	    192	    448	 127988	  1f3f4	../../../build/target/system-part2/platform-10-m/system-part2.elf
 125828	   2496	   3768	 132092	  203fc	../../../build/target/system-part3/platform-10-m/system-part3.elf
```

After:
```
  37260	    320	   2140	  39720	   9b28	../../../build/target/system-part1/platform-10-m/system-part1.elf
 127348	    192	    448	 127988	  1f3f4	../../../build/target/system-part2/platform-10-m/system-part2.elf
 107108	   2496	   3272	 112876	  1b8ec	../../../build/target/system-part3/platform-10-m/system-part3.elf
```

P.S. Don't get confused by the file names :)

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation


hal/src/electron/cellular_hal.cpp
```diff
@@ -7,6 +8,36 @@
 
 static CellularCredentials cellularCredentials;
 
+static HAL_NET_Callbacks netCallbacks = { 0 };
+
```
m-mcgowan: I wonder if the callbacks should take a reserved parameter so we can add more info later?

m-mcgowan: On reflection, I see that's not needed - the only eventuality is that the caller (Celluar HAL) passes more data than the callee is expecting to the callback. This is fine with the C calling convention.

m-mcgowan:
I wonder why we gain 25k in one part, but lose only 18k in another.  Is there duplication happening?

hal/src/electron/cellular_hal.cpp
m-mcgowan: If we need new callbacks yes...but if we need to extend the existing callbacks with additional parameters, we can just add them to the callback. Older system modules won't see the extra parameters, so that has to be kept in mind, but since they are informational/status that's not a concern - they only mean something if the receiver can understand them.

sergeuz:
I think that's something from libc that now happens to be in both part2 and part3. I'll check if there are any suspicious symbols though.

hal/src/electron/cellular_hal.cpp
sergeuz: hm, does that mean that it's not necessary to add our typical `void* reserved` to newly added dynalib functions? That trick when a regular function is called with more arguments than it expects isn't formally allowed in C as far as I know, though it seems to be supported by all major compilers.

sergeuz: disregard the above question. While it's okay to pass more arguments than necessary to a function, it's certainly bad idea to pass fewer arguments, which is the case for dynalib functions without `void* reserved`.

m-mcgowan: you got it! 👍

PR #1187, opened 2016-11-29
First implementation of I2C for RPi

monkbroc:
I'd like someone to review the container <-> array buffer code for improvement suggestions.

Specifically how `read` expects an array, but the I2C HAL interface must be able to provide bytes one by one later.
https://github.com/spark/firmware/pull/1187/files#diff-b8a60c5ba8245093dafd297716cb879aR115

And how `write` expects an array, but the data is accumulated through the I2C HAL interface one by one.
https://github.com/spark/firmware/pull/1187/files#diff-b8a60c5ba8245093dafd297716cb879aR148

Questions:

- What's I2C Stop?
- Does I2C only work inside transactions?


monkbroc:
Reviewed with Brett.

PR #1216, opened 2016-12-22
Improve Arduino compatibility

monkbroc:
- Simulate latest Arduino version: The latest Arduino version is 1.8.0 and uses C++11. Some libraries check this constant to know if they can use C++11 syntax.

- Ensure Arduino compatibility macros are defined even when Particle.h has already been included. This happens when compiling a .ino file on the compile server since the preprocessor includes Particle.h before the application has a chance of including Arduino.h.  Also happens for all existing Particle Libraries since Arduino.h was replaced typically with application.h or Particle.h.

- Define additional hardware serial defines.

- Added optional command line parameter to disable arduino compatibility.  Duplicated docs for this below.

---

## Compilation without Arduino Compatibility Support

Arduino compatibility defines are included by default to support hundreds of libraries that have been ported from Arduino to Particle.  If those defines should cause an issue with your build, you may remove them by adding

PARTICLE_NO_ARDUINO_COMPATIBILITY=y to the make command line. This requires a clean build.

There should be no difference in FLASH or RAM allocation with or without these defines.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation

### Enhancements

- [[PR #1216]](https://github.com/spark/firmware/pull/1216) Improved Arduino Compatibility (now supported by default, added PARTICLE_NO_ARDUINO_COMPATIBILITY=y command line option for disabling)


monkbroc:
Thanks you sir :)

user/inc/Arduino.h
```diff
@@ -4,6 +4,7 @@
 
 #define PARTICLE_ARDUINO_COMPATIBILITY
 #include "Particle.h"
+#include "spark_wiring_arduino.h"
```
technobly: This change may break existing libraries, since when original Arduino libraries were ported to Particle, Arduino.h was typically replaced with application.h or Particle.h.  I would suggest leaving #include "spark_wiring_arduino.h" in application.h and keeping the removal of #ifdef PARTICLE_ARDUINO_COMPATIBILITY in spark_wiring_arduino.h so that arduino is just supported by default.

technobly: Changing my mind on how we include arduino support by default, but we still need it.  I'll follow up with a commit.

monkbroc: I want to make sure your removal of this change still keeps the intent that:
```
#include "application.h"
#include "Arduino.h"
// compatibility macros should be defined here
```

It wasn't the case due to `Arduino.h` loading `Particle.h` loading `application.h` which had already been loaded without compatibility macros and wasn't loaded a second time due to the guard clauses.

technobly: Yep, by default now application.h will define PARTICLE_ARDUINO_COMPATIBILITY = 1 unless PARTICLE_NO_ARDUINO_COMPATIBILITY is explicitly defined as =y on the make command line.  Also Arduino.h just includes Particle.h which is an alias for application.h.

PR #1222, opened 2016-12-29
string buffer overflow

m-mcgowan:
fixes bug in `String(const char* str, int len)` constructor when the string is longer than the specified length. Instead of extracting just the specified substring, it copies the full string to the smaller buffer, corrupting the heap beyond the buffer.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### Bugfixes

- [[PR #1222]](https://github.com/spark/firmware/pull/1222) Fixed bug in `String(const char* str, int len)` constructor when the string is longer than the specified length.

### DIFF
--- a/wiring/src/spark_wiring_string.cpp
+++ b/wiring/src/spark_wiring_string.cpp
@@ -217,7 +217,8 @@ String & String::copy(const char *cstr, unsigned int length)
 		return *this;
 	}
 	len = length;
-	strcpy(buffer, cstr);
+	memcpy(buffer, cstr, length);
+	buffer[len] = 0;
 	return *this;
 }
 
--- a/user/tests/wiring/api/string.cpp
+++ b/user/tests/wiring/api/string.cpp
@@ -1,9 +1,131 @@
-
-
 #include "testapi.h"
 
-test(string_constructor_printable)
-{
+test(api_string) {
+    // String()
+    String s;
+    API_COMPILE(String());
+    API_COMPILE(String("abc"));
+    API_COMPILE(String("abc", 3));
+    API_COMPILE(String(s)); // Copy constructor
+    API_COMPILE(String(std::move(s))); // Move constructor
+    API_COMPILE(String('a'));
+    API_COMPILE(String(123, (unsigned)10 /* base */)); // FIXME: Ambiguous call
+    API_COMPILE(String(123.0, 2 /* decimalPlaces */));
+
+    // reserve()
+    bool b = false;
+    API_COMPILE(b = s.reserve(10));
+
+    // length()
+    unsigned u = 0;
+    API_COMPILE(u = s.length());
+
+    // concat()
+    String s2;
+    API_COMPILE(b = s.concat(s2));
+    API_COMPILE(b = s.concat("abc"));
+    API_COMPILE(b = s.concat('a'));
+    API_COMPILE(b = s.concat(123));
+    API_COMPILE(b = s.concat(123.0));
+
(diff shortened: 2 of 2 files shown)


wiring/src/spark_wiring_string.cpp
```diff
@@ -217,7 +217,8 @@ String & String::copy(const char *cstr, unsigned int length)
 		return *this;
 	}
 	len = length;
-	strcpy(buffer, cstr);
+	memcpy(buffer, cstr, length);
+	buffer[len] = 0;
```
harrisonhjones: Might it not be clearer if it was `buffer[len] = '\0';`? My C/C++ knowledge is woefully behind y'all's so sorry if the suggestion is silly.

technobly: They are the same thing, but I agree most people see a null terminator and know what's going on without thinking about it.  Sometimes the way to go though is to match the style used in the existing file instead of applying your own style here and there.  In that light, most of the file uses `= 0;` and there are only two instances of `= '\0';`

technobly:
Nice catch Mat! Looks good and fixed and I'd love to see a test added to keep it that way.

wiring/src/spark_wiring_string.cpp
m-mcgowan: The problem isn't null terminating the local buffer, but that it's copying too much data into the buffer.  We can't null terminate the `cstr` at the given length since that's input data (and const).

m-mcgowan: Oh, I see what you mean, Using a char instead of an int. Sure we can do that.

m-mcgowan:
I was unsure how to test it when I found the bug (how to test corrupt heap?) but on reflection it's simple. 

If we can pre-size the internal buffer to be larger than the input string (say 10 chars) and then provide a string that is 7 chars long, with a length of 5, then we verify that only 5 chars were copied to the buffer, and not 7.  (And with the buffer being larger, there is no memory corruption so we don't corrupt the test runner.)

I'll add this test when I'm back from vacation (if someone doesn't beat me to it.)

sergeuz:
Nice catch, Mat. I'll add the test, since, after all, I was the one who added that constructor to the `String` class. Couldn't imagine `String::copy()` was implemented in such a weird way though.

m-mcgowan:
Thank you for for stepping up! Yes, it's surprising that the provided `String` implementation makes a textbook error, but these things are easy to do, and easy to overlook in a review too. It took me several hours to track down the bug too, was far from trivial even though the fix is. I'm glad we provide the String class so that the insidiousness of c strings is encapsulated and wrapped to save users from shooting themselves in the foot. 

While you're working on this, could you please check with the arduino String implementation to see if there are any bugfixes made since our code was forked. There might be other bombs lurking waiting to go off :-)

sergeuz:
Hey @technobly, I've added API tests for the `String` class, as per your 0.6.1-rc.1 merging notes.

As for the unit test discussed above, at closer look I don't see how it can be implemented it without changing `String` API, since the bug is in `String::copy()` method, which is not public, nor can be tested indirectly via other methods in this particular case.

Considering that current coverage of `String` class by unit tests is very limited anyway, maybe let's merge this PR as it is? It still would be good to review the implementation later, and, as Mat suggested, backport possible fixes from Arduino later, after releasing rc2.

technobly:
Sounds good, I captured this in the merging notes for 0.6.1-rc.2

PR #1223, opened 2017-01-01
Only remake $(TARGET_BASE).elf el al. if necessary

jablko:
Commit ee1d29e made it possible to run program-dfu and friends without
first running `make all`, but it also made it so the $(TARGET_BASE).elf
recipe is constantly executed, whether it needs it or not.

The problem is that currently, $(MAKE_DEPENDENCIES) is phony, so real
targets shouldn't depend on that [1]. Phony prerequisites are never up
to date, so neither is anything that depends on them, and make rebuilds
anything that isn't up to date.

One solution is to move the phony prerequisites from the real targets to
program-dfu, which is itself phony. That's how the `all` target worked
before ee1d29e.

[1] https://www.gnu.org/software/make/manual/html_node/Phony-Targets

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/build/module.mk
+++ b/build/module.mk
@@ -82,9 +82,7 @@ ALLDEPS += $(addprefix $(BUILD_PATH)/, $(patsubst $(COMMON_BUILD)/arm/%,%,$(ASRC
 CLOUD_FLASH_URL ?= https://api.spark.io/v1/devices/$(SPARK_CORE_ID)\?access_token=$(SPARK_ACCESS_TOKEN)
 
 # All Target
-all: $(TARGET) postbuild
-
-build_dependencies: $(MAKE_DEPENDENCIES)
+all: $(MAKE_DEPENDENCIES) $(TARGET) postbuild
 
 elf: $(TARGET_BASE).elf
 bin: $(TARGET_BASE).bin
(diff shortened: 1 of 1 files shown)


technobly:
@jablko I'm curious what make commands you are running, and what this change helps you achieve.  I'm following what you are saying in the problem statement, and I'm also wondering if there is something we might break by making this change.

@m-mcgowan would you please review the make changes you made which were referenced in the past vs. this current proposal?

I currently use only make commands that start with `make clean all -s` and occasionally `make all -s` when building just user modules in `firmware/main`, so I'd be curious to know what other use cases are.

I ran a test between these two sets of commands and see a speed increase using the latter, but no difference between this PR and `release/v0.6.1`:
```
// 126 seconds
make clean all -s COMPILE_LTO=n PLATFORM_ID=10 DEBUG_BUILD=y

// 78 seconds
make clean -s PLATFORM=electron
make all -s COMPILE_LTO=n PLATFORM_ID=10 DEBUG_BUILD=y
```

jablko:
> @jablko I'm curious what make commands you are running, and what this change helps you achieve. I'm following what you are saying in the problem statement, and I'm also wondering if there is something we might break by making this change.

I'm running `make program-dfu` -- I'm following the getting started guide, [here](https://github.com/spark/firmware/blob/develop/docs/gettingstarted.md#4-flash-it).

I find that it runs the $(TARGET_BASE).elf recipe every time -- even though $(TARGET_BASE).dfu is up to date. Do you see the same thing?

It didn't used to do this -- and it's a mistake for a real target to depend on a phony target:

> A phony target should not be a prerequisite of a real target file; if it is, its recipe will be run every time make goes to update that file. https://www.gnu.org/software/make/manual/html_node/Phony-Targets

I don't expect this change will break anything -- it's the same way things worked before ee1d29e85f5514f333e713c0d673729c3b63bb8c -- @m-mcgowan?

jablko:
Sweet, thank you for merging!

m-mcgowan:
I did test it and merged, but was clearly not wide enough awake..because `program-dfu` fails with 

```
dfu-util -d 0x1D50:0x607F -a 0 -s 0x08005000:leave -D newlib_nano
dfu-util 0.8

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2014 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to dfu-util@lists.gnumonks.org

dfu-util: Could not open file newlib_nano for reading: No such file or directory
```

So we will have to back out the commit until we can fix it.  I'm puzzled why this doesn't happen for you, but perhaps changes later in the `develop` branch have caused it to break?

jablko:
Oops, I didn't notice that @sergeuz already fixed this. Thank you!

PR #1228, opened 2017-01-04
Added more description to tone()

jenesaisdiq:
Slightly more description on what it's used for, just to make it more searchable. "buzzer" previously returned no results.

< Please summarize the Problem and Solution this PR addresses as clearly as possible. This fascilitates speedy peer review and testing :) >

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/docs/reference/firmware.md
+++ b/docs/reference/firmware.md
@@ -2797,7 +2797,7 @@ void loop()
 
 ### tone()
 
-Generates a square wave of the specified frequency and duration (and 50% duty cycle) on a timer channel pin which supports PWM. Use of the tone() function will interfere with PWM output on the selected pin.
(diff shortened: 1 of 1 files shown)


technobly:
@jenesaisdiq thanks for the docs edit. Would you please move this to spark/docs/master though? Edits in the spark/firmware repo only get pulled into the docs/prerelease during firmware releases. So it will be a while before this edit makes its way to docs/master if done here.

jenesaisdiq:
Ahh, good to know @technobly , my first firmware docs edit after the restructuring. When I glanced in the docs repo I just saw the symlink to here. Was going all cowboy and making the edit right on github. It's ok if this only gets pulled in at the next release.

technobly:
Committed to [spark/docs/master](https://github.com/spark/docs/commit/ff89e3b521f780705e39a9e5dfed5bbafc3d4266) directly.

PR #1247, opened 2017-02-05
Add return value to WiFi.setCredentials()

sergeuz:
Currently, `WiFi.setCredentials()` may fail silently depending on specified security settings, it also doesn't provide any mechanism for error reporting, which may be confusing for users. This PR makes `WiFi.setCredentials()` return `true` if credentials has been stored successfully, or `false` otherwise.

I also noticed that the API allows to use `NULL` as a password argument, however, the implementation lacks necessary null pointer checks.

*TODO: Update firmware reference.*

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

### Enhancements

- [[PR #1247]](https://github.com/spark/firmware/pull/1247) Adds error checking to `WiFi.setCredentials()`, will return `true` if credentials has been stored successfully, or `false` otherwise.

### Bug fixes

- [[PR #1247]](https://github.com/spark/firmware/pull/1247) Previously no null pointer checks on password argument of `WiFi.setCredentials()`.


technobly:
Since we will have a return type now, should we enforce 8-63 character passwords for WPA?  I know from testing in the past that any password < 8 for WPA will not be saved by WICED.  I'm not sure how this differs for WEP though.

sergeuz:
I'll make some tests to see if WICED returns an error in this case, and if it does, then the error should be propagated to `WiFi.setCredentials()` already.

sergeuz:
Added "bug" label since the PR also fixes potential null pointer access.

sergeuz:
@technobly You're right, since we're only storing WiFi credentials to DCT rather than trying to connect to a specified AP directly, we're bypassing WICED's preliminary checks for password validity and thus `Particle.setCredentials()` returns `true` for passwords that don't meet requirements for a specified security type.

I've added password length checks for WPA/WPA2 and WEP (Photon-only).

PR #1248, opened 2017-02-05
Avoid division by zero when map() is called with incorrect parameters

### LINKED ISSUE #1193: Protect `map()` against Usage Fault when `fromEnd == fromStart`
This is the current implementation (as used on AVR Arduinos without issue)
```cpp
long map(long value, long fromStart, long fromEnd, long toStart, long toEnd)
{
  return (value - fromStart) * (toEnd - toStart) / (fromEnd - fromStart) + toStart;
}
```

But on STM32F µC an unhandled DIV/0 exception will cause a Usage Fault.
So checking for that and acting gracefully would be good.
e.g. like this
```cpp
#include <math.h>

long map(long value, long fromStart, long fromEnd, long toStart, long toEnd)
{
  if (fromEnd == fromStart) return NAN;  // to prevent DIV/0 exception bail out with NaN
  return (value - fromStart) * (toEnd - toStart) / (fromEnd - fromStart) + toStart;
}
```

And if at that, maybe also add a `double` overload
```cpp
double map(double value, double fromStart, double fromEnd, double toStart, double toEnd)
{
  if (fromEnd == fromStart) return NAN;  // to prevent DIV/0 exception bail out with NaN
  return (value - fromStart) * (toEnd - toStart) / (fromEnd - fromStart) + toStart;
}
```

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (as of 0.6.0)

sergeuz:
This PR fixes #1193 by adding a check into `map()` function that ensures that incorrect parameters won't cause division by zero error. The PR also overloads `map()` to take `double` arguments.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation (move docs to `docs/prerelease`)

### Bug fixes

- [[PR #1248]](https://github.com/spark/firmware/pull/1248) [[Fixes #1193]](https://github.com/spark/firmware/issues/1193) Fixes divide by zero on incorrect parameters of `map()` function.

### Enhancements

- [[PR #1248]](https://github.com/spark/firmware/pull/1248) Added an overload to `map()` function that takes `double` arguments.

### DIFF
--- a/docs/reference/firmware.md
+++ b/docs/reference/firmware.md
@@ -7569,9 +7569,9 @@ The function also handles negative numbers well, so that this example
 
 is also valid and works well.
 
-The `map()` function uses integer math so will not generate fractions, when the math might indicate that it should do so. Fractional remainders are truncated, and are not rounded or averaged.
+When called with integers, the `map()` function uses integer math so will not generate fractions, when the math might indicate that it should do so. Fractional remainders are truncated, not rounded.
 
-*Parameters:*
+*Parameters can either be integers or floating point numbers:*
 
 - `value`: the number to map
 - `fromLow`: the lower bound of the value's current range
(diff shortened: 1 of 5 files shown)


sergeuz:
Thanks for the docs, @monkbroc! I wasn't sure if we're still committing them directly to the firmware repo or not.

PR #1250, opened 2017-02-13
fix crash in tcp_server_t due to incorrect tcp_server_t::clients[]

tlangmo:
Fixes a crash related to memory corruption. I think it is the reason for the problem mentioned here:
https://community.particle.io/t/active-tcpserver-client-appears-to-crash-photon-when-wifi-disconnects/27130

The crash shows when following event sequence occurs
1. TCPServer is started
2. TCPClient is created for connection
3. TCPClient connection is closed gracefully
4. TCPServer is destroyed

Bug:
The client list of TCPServer is not updated correctly when a client disconnects. Subsequently, on destruction of TCPServer, an already destroyed client object is accessed. 

I confirmed the events with the debugger, but unfortunately do not have the stack traces at hand anymore.

Fix:
In TCPServer::disconnect, set correct pointer in clients[] to NULL. The object itself is destroyed in socket_t::dispose()

---

Doneness:

- [ ] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation (N/A)

### DIFF
--- a/hal/src/photon/socket_hal.cpp
+++ b/hal/src/photon/socket_hal.cpp
@@ -289,6 +289,12 @@ struct tcp_server_t : public wiced_tcp_server_t
      */
     wiced_result_t disconnect(wiced_tcp_socket_t* socket) {
         wiced_tcp_disconnect(socket);
+        // remove from client array as this socket is getting closed and
+        // subsequently destroyed
+	    int idx = index(socket);
+	    if (idx >= 0) {
+		  clients[idx] = NULL;
+	    }
         wiced_result_t result = wiced_tcp_server_disconnect_socket(this, socket);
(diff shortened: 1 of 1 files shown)


technobly:
Hi @tlangmo thanks for the PR!

I wonder if something along these lines is appropriate to close the individual client before setting it to NULL: 
https://github.com/Motesque/photon-firmware/blob/b6b3b0c24217d09f6855d807d20e5852b04a3789/hal/src/photon/socket_hal.cpp#L305-L309

sergeuz:
Thanks for the patch, @tlangmo. I've included your commits into a larger PR that addresses various TCP server issues: https://github.com/spark/firmware/pull/1286

PR #1258, opened 2017-02-22
use 'using std::**' instead of define * std::*

kreopt:
**Summary of Commits:**
Fix C++ build using std::isnan or std::isinf with explicit namespace specification

**Problem:** 
Using define for using c++ namespaces breaking compilation when names are used with explicit namespace specification. So std::isnan converts to std::std::isnan.

**Solution:** 
use c++ _using_ keyword insead of define.

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation


sergeuz:
@m-mcgowan Mat, could you review this? I remember you were working on something related to math.h and Arduino compatibility.

PR #1260, opened 2017-02-26
Interrupt cloud connection when entering listening mode or powering off the device via setup button

### LINKED ISSUE #1067: as a user, I can interrupt the blinking cyan cloud connection with the setup button
At present, pressing the setup button to enter listening mode, or a double tap to powerdown, doesn't interrupt the cloud connection process. (It does interrupt the network connection process.)

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (as of 0.5.3-rc.1)

sergeuz:
This PR fixes #1067 by introducing a flag that can be set from an ISR and cause read/write operations on the cloud socket to fail with a generic error, effectively breaking any busy loop within the comms library.

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/system/inc/system_task.h
+++ b/system/inc/system_task.h
@@ -90,6 +90,13 @@ uint8_t main_thread_current(void* reserved);
 
 uint8_t application_thread_invoke(void (*callback)(void* data), void* data, void* reserved);
 
+/**
+ * Cancels current network connection attempt and aborts cloud connection. This function can be
+ * called from an ISR and is used to unblock the system thread in order to perform some other
+ * operation immediately.
+ */
+void cancel_connection();
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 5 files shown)


m-mcgowan:
This is a good attempt at interrupting the cloud connection.  I wonder about some of the edge cases though...such was when the wifi network disappears or a socket doesn't timeout.  Can we push the "pull the plug" functionality down into the networking layers too to the socket connect/read/write implementation in our various HALs?  I know this is a big ask - async interruption of I/O is tricky - just would like to discuss and see what possibilities there might be.  (For example, I believe the Electron HAL supports interruptible socket operations. )

sergeuz:
@technobly will correct me, but uBlox manual says that socket AT commands cannot be aborted, so if we want to interrupt a blocking socket call on Electron and keep the modem and the parser in a consistent state the only option is probably to reset the modem. As for the Photon, any blocking socket calls are aborted as part of the LwIP shutdown process which in turn is performed during WICED's WLAN uninitialization.

PR #1263, opened 2017-03-01
Add solution for this 'while' clause does not guard error

ulikoehler:
I've installed arm-none-eabi-gcc version
```gcc version 6.2.1 20161205 (release) [ARM/embedded-6-branch revision 243739] (GNU Tools for ARM Embedded Processors)```
which causes this error message when building:

```
MCU/STM32F2xx/SPARK_Firmware_Driver/src/system_stm32f2xx.c: In function 'SetSysClock':
MCU/STM32F2xx/SPARK_Firmware_Driver/src/system_stm32f2xx.c:394:5: error: this 'while' clause does not guard... [-Werror=misleading-indentation]
     while ((RCC->CFGR & (uint32_t)RCC_CFGR_SWS ) != RCC_CFGR_SWS_PLL);
```

This pull request adds a new solution to the Getting Started page that addresses this issue by disabling this specific type of warning in the compiler options.

This is a doc-only change, therefore I haven't run any tests. I'm not sure if I need to fill the doneness section below or if it will be filled by you, so I've left it empty.

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/docs/gettingstarted.md
+++ b/docs/gettingstarted.md
@@ -54,6 +54,16 @@ The [makefile documentation](build.md) describes the build options supported and
 
 * You get `make: *** No targets specified and no makefile found.  Stop.`
   Solution: `cd firmware/main`
+  
+* Newer GCCs produce this error message:
+  ```
+MCU/STM32F2xx/SPARK_Firmware_Driver/src/system_stm32f2xx.c: In function 'SetSysClock':
+MCU/STM32F2xx/SPARK_Firmware_Driver/src/system_stm32f2xx.c:394:5: error: this 'while' clause does not guard... [-Werror=misleading-indentation]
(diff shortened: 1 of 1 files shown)


ulikoehler:
Strange, even fixing this issue caused many missing include directories etc. to pop up.

I can build the firmware without any issues when running `make` from inside `firmware` instead of inside `firmware/modules`. The binary that is produced looks OK.

Maybe this is supposed to work that way and the documentation is just not up to date ?!!?

sergeuz:
I believe invoking `make` from the firmware's root directory can only be used to build the firmware for the Core platform. We already have a PR that attempts to fix GCC 6.x build for all other platforms as well, and slightly more work than just disabling the warnings is necessary there. Feel free to check it out: https://github.com/spark/firmware/pull/1218

Issue #1165, opened 2016-11-10
Clarification about pending Particle.publish() delivery on cloud disconnections


ScruffR:
In connection with [this thread](http://community.particle.io/t/connecting-to-cloud-publishing-variable-and-going-offline-again/27164) I had a short discussion with Mat (@m-mcgowan) whether an introduced fix (maybe #469 or PRs #909/#918) which addressed the issue of events not being delivered without a 5sec delay between publish and sleep, also applies to `WiFi.off()` (and also `Particle.disconnect()` or other commands that would cut the connection).
The outcome of that discussion wasn't really clear, so I decied to file this issue for safe keeping ;-)

```cpp
  Particle.connect();
  waitFor(Particle.connected, 10000);
  Particle.publish("event", "1", PRIVATE);

  // this used to be required, but isn't anymore for 0.5.3+for sleep()
  //for (uint32_t ms=millis(); millis()-ms < 5000; Particle.process());

  // this definetly works on Electron and probably on Photon too
  //System.sleep(SLEEP_MODE_DEEP, 60);

  // but how about these?
  WiFi.off();
  // WiFi.disconnect();
  // WiFi.listen(); // ???
  // Cellular.off();
  // Cellular.disconnect();
  // Particle.disconnect();
  // and whatever I've forgotten
```

It would be good if a pending publish would first get delivered (maybe even blocking, or with an additional parameter `boolean force = false` to give the user contol) before this action cuts the connection.

Maybe in the same context keeping these and `sleep()` from cutting the connection during an OTA update might save useres some headache why their code just won't stick. There are multiple threads about this issue in a way or another. That even (or especially) seems to be a problem for Console users who rely on pending updates to be delivered correctly without too much fuss in their code.
(respective issue #1166)

---

Completeness:
- [X] Minimum test case added
- [X] Device, system and user firmware versions stated

ScruffR:
Two months and no Particle confirmation???

This problem still seems to exist in v0.6.1-rc.1 on Photons even for `System.sleep()`

When entering Stop Mode Sleep immediately after a `Particle.publish()` the event never reaches the console and on next wake the device doesn't seem to connect properly either.
With a `delay(5000)` between publish and sleep the code works for a Photon.
But for Electrons the very same code works now even without the delay.

technobly:
[...]

PR #1269, opened 2017-03-06
Graceful cloud disconnect

avtolstoy:
Firmware should terminate cloud connection in a graceful manner when possible to ensure that confirmable messages have been received by the server. Relevant issue: #1165

TODO: This PR could make use of #1245 (mainly the `Particle.publish()` change that makes it returns a `Future`) for on-device tests.

## Review only. Do not merge. Already in #1289

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [x] Add documentation


system/src/system_cloud_internal.cpp
```diff
@@ -1047,24 +1047,41 @@ int spark_cloud_socket_connect()
...
+                    system_tick_t start = millis();
+                    while (millis() - start < 5000) {
+                        if (!Spark_Communication_Loop())
+                            break;
+                    }
+                }
+            } else {
+                spark_protocol_command(sp, ProtocolCommands::DISCONNECT, 0, nullptr);
+            }
+        }
```
sergeuz: It looks like application [ACK handlers](https://github.com/spark/firmware/blob/feature/graceful_disconnect/communication/src/dtls_protocol.h#L79) don't get cleared in case of a non-graceful disconnect. Should we add a separate command which would tear down a protocol's session without waiting for unconfirmed messages?

system/src/system_network_internal.h
```diff
@@ -387,7 +386,7 @@ class ManagedNetworkInterface : public NetworkInterface
             WLAN_CONNECTED = 0;
             WLAN_DHCP = 0;
 
-            cloud_disconnect();
+            cloud_disconnect_graceful();
```
sergeuz: I don't think the socket should be closed gracefully here. If an application (or the system) attempts to turn off the network without disconnecting from the cloud explicitly (either gracefully or not) then it most likely wants the network to be turned off immediately, without waiting for any unconfirmed messages.

communication/src/spark_protocol.cpp
```diff
@@ -1411,10 +1423,8 @@ bool SparkProtocol::handle_message(msg& message, token_t token, CoAPMessageType:
     case CoAPMessageType::DESCRIBE:
     {
         int desc_flags = DESCRIBE_ALL;
-        if (message.len > 8 && queue[8] <= DESCRIBE_ALL) {
+        if (message_padding_strip(queue, message.len) > 8 && queue[8] <= DESCRIBE_ALL) {
```
avtolstoy: I should have probably added a comment that this strips PKCS#7 padding. This is correctly handled in newer protocol implementation, see https://github.com/spark/firmware/blob/develop/communication/src/lightssl_message_channel.cpp#L76

It actually does fix an issue with https://github.com/spark/firmware/pull/1226 not working correctly as `message.len` is always `> 8` due to the padding.

system/src/system_cloud_internal.cpp
avtolstoy: :+1:

system/src/system_network_internal.h
avtolstoy: I don't remember exactly what we decided on when discussing futures and graceful disconnection, but if I remember correctly the idea was that `WiFi.disconnect()` and `Cellular.disconnect()` calls should close the connection sanely and disconnect from WiFi or cellular network, compared to `WiFi.off()` and `Cellular.off()` which would terminate the connection and turn off wireless module 'immediately'.

What would be the cons of closing cloud connection normally before disconnecting from wireless network?

sergeuz: Yeah, I meant `WiFi.off()` and `Cellular.off()` specifically, which I think should turn off the networking immediately (at least by default). The problem with current code is that `ManagedNetworkInterface::off()` calls `ManagedNetworkInterface::disconnect()` first and thus always waits for unconfirmed messages.

technobly:

PR #1270, opened 2017-03-06
[ch1741] OTA'd module info sent as a payload in UpdateDone message or as an ACK to UpdateDone

### LINKED ISSUE #1240: TCP Firmware sends ACKs and separate response for every Chunk in Fast OTA mode
In Fast OTA mode, the firmware should not respond to every Chunk message (especially because they are NON confirmable messages). To compensate for this, it should respond to the UpdateDone with whether it needs to request additional missed/invalid chunks. Those missing/invalid chunks are then requested via the ChunkMissed message.

The current firmware follows the correct procedure for requesting missing chunks in Fast OTA mode, but it **incorrectly [sends ACKs](https://github.com/spark/firmware/blob/dfdadfb3e3509bf3e98a28f64bd3917caca1ffa2/communication/src/spark_protocol.cpp#L1074) and [separate responses](https://github.com/spark/firmware/blob/dfdadfb3e3509bf3e98a28f64bd3917caca1ffa2/communication/src/spark_protocol.cpp#L1122) for each individual Chunk message**.

Firmware version: up to and including 0.6.1-rc.2
Device: Photon/P1/Core (any TCP device)

---

Completeness:
- [x] Minimum test case added

avtolstoy:
Original PR #1097 for issue #1032 introduced `spark/device/ota_result` event that was posted after receiving a binary OTA and before performing a reboot, which carried OTA'd module info after validity has been checked. This PR removes that separate event and instead sends OTA'd module info as a payload in UpdateDone message or as an ACK to UpdateDone as disccussed in _ch1741_.

This PR also fixes #1240.

TODO: Update acceptance test (#1153) if possible.

---

Doneness:

- [x] Code peer reviewed
- [x] API tests compiled
- [ ] Add documentation

---

### Enhancements

- [`[PR #1270]`](https://github.com/spark/firmware/pull/1270)  Removes `spark/device/ota_result` event and instead sends OTA'd module info as a payload in UpdateDone message, or as an ACK to UpdateDone.

### Bugfix

- [`[PR #1270]`](https://github.com/spark/firmware/pull/1270) [`[Fixes #1240]`](https://github.com/spark/firmware/issues/1240) TCP Firmware will not ACK every chunk in Fast OTA mode now.

### DIFF
--- a/communication/makefile
+++ b/communication/makefile
@@ -7,6 +7,6 @@ TARGET_TYPE = a
 
 BUILD_PATH_EXT=$(COMMUNICATION_BUILD_PATH_EXT)
 
-DEPENDENCIES = hal dynalib services wiring
+DEPENDENCIES = hal dynalib services wiring system
 
 include ../build/arm-tlm.mk
--- a/communication/makefile
+++ b/communication/makefile
@@ -7,6 +7,6 @@ TARGET_TYPE = a
 
 BUILD_PATH_EXT=$(COMMUNICATION_BUILD_PATH_EXT)
 
-DEPENDENCIES = hal dynalib services wiring system
+DEPENDENCIES = hal dynalib services wiring
 
 include ../build/arm-tlm.mk
--- a/communication/src/messages.cpp
+++ b/communication/src/messages.cpp
@@ -132,7 +132,7 @@ size_t Messages::hello(uint8_t* buf, message_id_t message_id, uint8_t flags,
 	return len;
 }
 
-size_t Messages::update_done(uint8_t* buf, message_id_t message_id, uint8_t* result, size_t result_len, bool confirmable)
+size_t Messages::update_done(uint8_t* buf, message_id_t message_id, const uint8_t* result, size_t result_len, bool confirmable)
 {
 	size_t sz = 6;
 	buf[0] = confirmable ? 0x40 : 0x50; // confirmable/non-confirmable, no token
--- a/communication/src/messages.cpp
+++ b/communication/src/messages.cpp
@@ -132,15 +132,26 @@ size_t Messages::hello(uint8_t* buf, message_id_t message_id, uint8_t flags,
 	return len;
 }
 
-size_t Messages::update_done(uint8_t* buf, message_id_t message_id, bool confirmable)
+size_t Messages::update_done(uint8_t* buf, message_id_t message_id, uint8_t* result, size_t result_len, bool confirmable)
 {
+	size_t sz = 6;
 	buf[0] = confirmable ? 0x40 : 0x50; // confirmable/non-confirmable, no token
 	buf[1] = 0x03; // PUT
 	buf[2] = message_id >> 8;
 	buf[3] = message_id & 0xff;
 	buf[4] = 0xb1; // Uri-Path option of length 1
 	buf[5] = 'u';
-	return 6;
+	if (result && result_len) {
+		buf[sz++] = 0xff; // payload marker
+		memcpy(buf + sz, result, result_len);
+		sz += result_len;
+	}
+	return sz;
+}
+
+size_t Messages::update_done(uint8_t* buf, message_id_t message_id, bool confirmable)
+{
+	return update_done(buf, message_id, NULL, 0, confirmable);
 }
 
 size_t Messages::function_return(unsigned char *buf, message_id_t message_id, token_t token, int return_value, bool confirmable)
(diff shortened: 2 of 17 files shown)


technobly:
@avtolstoy would you please add examples of what the payload looks like based on the different responses?  Perhaps this isn't necessary if the data is just forwarded to the event stream as-is, but would be useful to document here.

communication/src/messages.cpp
```diff
@@ -132,15 +132,26 @@ size_t Messages::hello(uint8_t* buf, message_id_t message_id, uint8_t flags,
 	return len;
 }
 
-size_t Messages::update_done(uint8_t* buf, message_id_t message_id, bool confirmable)
+size_t Messages::update_done(uint8_t* buf, message_id_t message_id, uint8_t* result, size_t result_len, bool confirmable)
```
technobly: Any reason not to make the `result`, `result_len`, `data`, and `data_len` const?

communication/src/spark_protocol.cpp
```diff
@@ -1198,6 +1200,33 @@ void SparkProtocol::set_chunks_received(uint8_t value)
...
+size_t SparkProtocol::notify_update_done(uint8_t* msg, token_t token, uint8_t code)
+{
+    size_t msgsz = 0;
+    char buf[255];
+    size_t data_len = 0;
+
+    memset(buf, 0, sizeof(buf));
+    if (code != ChunkReceivedCode::BAD) {
+        callbacks.finish_firmware_update(file, UPDATE_FLAG_SUCCESS | UPDATE_FLAG_VALIDATE_ONLY, buf);
+        data_len = strnlen(buf, sizeof(buf) - 1);
```
technobly: Do we include the length of the extra payload info in `buf`? If not, do we just rely on the `size_t` of the packet to know where the payload ends, since we are not sending the null terminator `sizeof(buf) - 1`.

system/src/system_update.cpp
```diff
@@ -318,7 +318,12 @@ int Spark_Finish_Firmware_Update(FileTransfer::Descriptor& file, uint32_t flags,
 
     hal_module_t mod;
 
-    if (flags & 1) {    // update successful
+    if ((flags & (UPDATE_FLAG_VALIDATE_ONLY | UPDATE_FLAG_SUCCESS)) == (UPDATE_FLAG_VALIDATE_ONLY | UPDATE_FLAG_SUCCESS)) {
```
technobly: Just a question about `(UPDATE_FLAG_VALIDATE_ONLY | UPDATE_FLAG_SUCCESS)) == (UPDATE_FLAG_VALIDATE_ONLY | UPDATE_FLAG_SUCCESS)`.  Can this end up not being 1 somehow, what is the purpose of this change?

avtolstoy: `(UPDATE_FLAG_VALIDATE_ONLY | UPDATE_FLAG_SUCCESS) != 1`. The first `if` branches us into a validate-only case when both `UPDATE_FLAG_VALIDATE_ONLY` and `UPDATE_FLAG_SUCCESS` are set. The one after is the original `flags & 1` which is equivalent to `flags & UPDATE_FLAG_SUCCESS`.

communication/src/spark_protocol.cpp
avtolstoy: We don't include the length of the payload and we don't include `\0`. The payload size is easily calculated at the receiving side when parsing the packet.

I've checked and we don't include `\0` for publishes either.

communication/src/messages.cpp
avtolstoy: Sure, they could be const.

communication/makefile
```diff
@@ -7,6 +7,6 @@ TARGET_TYPE = a
 
 BUILD_PATH_EXT=$(COMMUNICATION_BUILD_PATH_EXT)
 
-DEPENDENCIES = hal dynalib services wiring
+DEPENDENCIES = hal dynalib services wiring system
```
m-mcgowan: We've been very conscious to avoid coupling `communication` with `system`. For any external notifications/behaviour, please add a callback to avoid a direct coupling.

system/inc/system_update.h
```diff
@@ -54,6 +54,12 @@ typedef enum {
     MODULE_INFO_JSON_INCLUDE_PLATFORM_ID = 0x0001
 } module_info_json_flags_t;
 
+typedef enum {
```
m-mcgowan: I would move these down into the comms layer as part of the definition of the notify_update_done callback definition so this layer isn't dependent upon system.

PR #1273, opened 2017-03-09
Feature/publish private default

m-mcgowan:
When an event visibility isn't specified, uses `PRIVATE` by default. Was previously `PUBLIC`. 

---

Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### DIFF
--- a/user/tests/unit/catch.hpp
+++ b/user/tests/unit/catch.hpp
@@ -1,6 +1,6 @@
 /*
- *  CATCH v1.1 build 6 (develop branch)
- *  Generated: 2014-10-02 19:13:03.375275
+ *  Catch v1.8.1
+ *  Generated: 2017-03-01 16:04:19.016511
  *  ----------------------------------------------------------
  *  This file has been merged from multiple headers. Please don't edit it directly
  *  Copyright (c) 2012 Two Blue Cubes Ltd. All rights reserved.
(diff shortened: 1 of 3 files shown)


brandoaire:
@m-mcgowan is the default scope of Particle.subscribe() also changed? If not, should it? My thought is that if a user specifies no public or private scope, both publish() and subscribe() should  interact using the same event stream.

m-mcgowan:
That's a good question - I was wondering the same when I made the change. Should the default for subscriptions then be MY_DEVICES, with the option to specify ALL_DEVICES?  I believe it should change, since public/private events are mutually exclusive (iirc - I will double check this.) so that subscribing to ALL_DEVICES for PRIVATE events does't receive the private events even though you might think so from the ALL_DEVICES name.

brandoaire:
We also need to remember to change the examples in documentation when we make this change
https://docs.particle.io/reference/firmware/core/#particle-publish-

technobly:
Yes they are mutually exclusive.  
https://github.com/spark/docs/issues/560

I was thinking maybe we should standardize the scope flags for Particle.subscribe and Particle.publish to both use PRIVATE and PUBLIC.  Would that be less confusing than remembering MY_DEVICES / ALL_DEVICES?  So MY_DEVICES would be deprecated and instead PRIVATE would be used.  PUBLIC would also be new for Particle.subscribe.

m-mcgowan:
Some open questions:

- what impact will this have on webhooks
- what impact will this have on IFTTT integration

cc: @dmiddlecamp

technobly:
We had some great discussions today about this change, and how it relates to subscriptions and webhooks.

To summarize, what we'd like to move forward with for 0.7.0 is the following:

Particle.publish() PRIVATE by default, with the option of being explicitly set to PRIVATE or PUBLIC.

Particle.subscribe() to be explicitly set by default.  

- If the default case is compiled, it will present a nice error describing that one or more of the following flags should be used.
- Even though the default case would not be supported, we'd have to maintain the dynalib signature for supporting updating system firmware before user firmware is updated.

MY_DEVICES - still means MY_PRIVATE (but is depreciated)
ALL_DEVICES - still means ALL_PUBLIC (but is depreciated)
MY_PRIVATE - my private events
MY_PUBLIC - my public events
ALL_PUBLIC - all public firehose events

MY_PUBLIC currently doesn't have any CoAP setup to support it, although there is a Particle API endpoint for this.  This would require a new CoAP message and device service changes.

These flags should be able to be OR'd together to get a desired combination, e.g.
MY_PRIVATE | ALL_PUBLIC
MY_PRIVATE | MY_PUBLIC

Some related Docs issues:
- https://github.com/spark/docs/issues/611
- https://github.com/spark/docs/issues/560

towynlin:
I'm reviewing the CoAP messages for subscriptions and will write a brief spec for the protocol changes.

technobly:
I updated the above summary with our updates:

- pub/sub require scope to be explicitly set
- MY_PRODUCT subscription flag (feel free to change the name if desired)
- desire to reintroduce default APIs at v1.0.0 after the majority of users have adopted the changes to pub/sub APIs.

Issue #1278, opened 2017-03-16
Arduino Compatibility for All


technobly:
Re: https://community.particle.io/t/fastled-not-compiling-to-photon/30031/13?u=bdub

## Legacy
To provide maximum compatibility for Arduino libraries, and still support hundreds of libraries that were ported from Arduino to Particle, we include Arduino defines by default for 0.6.1 firmware now.

In the past <= 0.6.0, users would get Arduino defines without the need to add `#include "Arduino.h"` to their code.  And there was no Arduino.h, so that had to be commented out in libraries for them to compile.

## Present 0.6.1
Now Arduino.h just includes Particle.h (which in turn includes the legacy application.h).  Also legacy libraries would include application.h directly in many cases.  If application.h is included, [PARTICLE_ARDUINO_COMPATIBILITY will be defined.  And by default it will be included and set to 1](https://github.com/spark/firmware/blob/develop/user/inc/application.h#L90).  If that is the case, all of the defines in [`spark_wiring_arduino.h`](https://github.com/spark/firmware/blob/develop/wiring/inc/spark_wiring_arduino.h#L23) will be included in your code.  So [RwReg](https://github.com/spark/firmware/blob/develop/wiring/inc/spark_wiring_arduino.h#L152-L154) will be defined.

There is no way to get ahead of the application.h include if it is included in a library and that is added to your app in the Web IDE.  It appears to compile the library first.

If we could get the app to compile first, and compile the library once the include for the library is found, we could juggle something like this perhaps:

```
#include "Particle.h"

#ifdef RwReg
  #undef RwReg
#endif

#include "FastLED.h"
FASTLED_USING_NAMESPACE;
#define NUM_LEDS 60
CRGB leds[NUM_LEDS];
void setup() { FastLED.addLeds<NEOPIXEL, 6>(leds, NUM_LEDS); }
void loop() { 
    leds[0] = CRGB::White; FastLED.show(); delay(30); 
    leds[0] = CRGB::Black; FastLED.show(); delay(30);
}
```

But that's going to be a mess I think because there's bound to be a bunch of things to redefine.

## Suggested Way Forward (for 0.6.1)
I think the proper thing to do here to maintain compatibility with the library and older AND newer firmware, is to wrap the arduino-based defines embedded in the library with

```
#ifndef MY_ARDUINO_SYMBOL 
  #define MY_ARDUINO_SYMBOL 
#endif
```
[...]

PR #1283, opened 2017-03-27
Feature/arduino compatibility regression

m-mcgowan:
Attempts to restore 0.6.0-style arduino compatibility as outlined by issue #1278 

In a nutshell:

- the default is 0.6.0 Arduino compatibility - let's call this partial compatibility. 
- applications and libraries include "Arduino.h" to define `ARDUINO` and bring in many more additional defines to complete the Arduino API.  We'll call this full compatibility. 

## Mixing libraries that require partial and full Arduino compatibility

Libraries that require full Arduino compatibility may include Arduino.h in their library header file, while libraries that depend on partial compatibility (because they themselves define missing symbols) don't include Arduino.h.  If a library requiring partial compatibility is included after one requiring full compatibility then the partial compatibility library may break, since Arduino symbols that it's not expecting will be defined. 

The workaround for this is to include all libraries requiring partial compatibility before libraries requiring full compatibility in the application file. 

---
Doneness:

- [ ] Code peer reviewed
- [ ] API tests compiled
- [ ] Add documentation

### ENHANCEMENT / BUG FIX

- [[PR #1283]](https://github.com/spark/firmware/pull/1283) [[Implements #1278]](https://github.com/spark/firmware/issues/1278) Restores 0.6.0-style Arduino compatibility by default, full Arduino compatibility when including Arduino.h


m-mcgowan:
## Internal Design

- The partial Arduino API as it was in 0.6.0 is defined in `spark_wiring_arduino.h`. This file has been restored to it's former 0.6.0 state without the new Arduino APIs. This file is always included with Particle.h/application.h. 

- The full Arduino API is declared in "Arduino.h". By keeping it separate we make a clear distinction between which symbols are included in the partial API and the full API. 

- There is some overlap of the partial API and the full API - `constrain` and `range` are examples. In the partial API, these are declared as templates. In the full API, they are macros. Hence, the usual trick of using `#ifndef constrain` won't work in the full API since it's defined as a template and not a macro. The workaround is to use `PARTICLE_WIRING_ARDUINO_TEMPLATE` as a define when these few common symbols have already been defined in the partial API, so they are not re-defined in the full API.  If we are splitting hairs, this may mean that some arduino libraries fail to compile in cases where the template does not behave the same as the macro (or that they are relying upon double-evaluation side-effects of the macro!!).  In this rare (hopefully non-existent) case, it is possible to get the arduino version of these defines, by disabling the preprocessor, and including `Arduino.h` at the top of the sketch.

monkbroc:
Did you check what happens when multiple include files are included? Are the Arduino symbols defined "as expected"?

For example:
```
#include "some_library.h"
#include "Arduino.h"
```
and `some_library.h` includes `Particle.h`.

Or 
```
#include "library1.h" // includes "Particle.h"
#include "library2.h" // includes "Arduino.h"
```

It can get hairy with symbol redefinitions in libraries so it would be fine if some cases don't work, but it would be nice if the easy case (both Arduino.h and Particle.h included, libraries don't redefine Arduino symbols) works.

wiring/inc/spark_wiring_spi.h
```diff
@@ -137,7 +137,7 @@ class SPISettings : public Printable {
 }
 
 #ifdef PARTICLE_WIRING_ARDUINO_COMPATIBILITY
-typedef particle::SPISettings SPISettings;
+typedef particle::__SPISettings __SPISettings;
```
avtolstoy: Should this be `typedef particle::__SPISettings SPISettings`?

m-mcgowan: yes...good catch. I had the IDE do the renaming...

technobly:
Github can't tell if the changes I requested were implemented... but I think they were?

m-mcgowan:
Once the arduino header is included, the symbols remain visible until the end of the compilation unit (and for all subsequent headers.)

For example, in this case 

```
#include "libraryExpecting062ArduinoSymbols.h"  // library 1
#include "libraryExpecting060Symbols.h"  // library 2
```

The second library is not expecting the arduino symbols to be defined (e.g. it might declare it's own copy of `SPISettings`) but these symbols are defined if library 1 includes `Arduino.h` in its header. To work around this, includes should be done in libraries an applications in a way thats all the headers not expecting arduino support first, and then all arduino-dependent headers afterwards.  

In the example above, reversing the order of the headers would fix the problem.

PR #1295, opened 2017-04-10
v2 dcd format and tests

m-mcgowan:
### Problem

Hypothesised write errors in flash that appear only after power cycle can corrupt the configuration data.
Flash access is not thread-safe.

### Solution

Adds CRC checking to the Electron DCD implementation so that write errors are detected.
Add a critical section around flash operations and around DCD operations.

### Steps to Test

Run the firmware unit tests.

Manual testing:

- flash system firmware to an electron that was connecting to the cloud
- verify that the device continues to connect to the cloud
- make a change in DCD via DFU
- verify the device still connects to the cloud (the DCD contents will have been upgraded to the V2 format.)

### Example App

This is an internal system change. No example application.

---

### Completeness

- [ ] Added documentation


hal/src/stm32f2xx/bootloader_platform_10.c
```diff
@@ -1,37 +1,37 @@
 const unsigned char bootloader_platform_10_bin[] = {
```
avtolstoy: If a new bootloader binary is bundled here, it should probably have a bumped module version, otherwise it is not automatically upgraded, which causes a device to enter an endless loop, as well as DCT corruption.

m-mcgowan: Totes, I probably tested against an older version so got the upgrade anyway. Although I did also see the endless loop when developing this initially. That's a bit worrying - anything we can to do help folks out of that if things do go wrong?

services/inc/dcd.h
```diff
@@ -194,45 +493,100 @@ class DCD
...
-        Header header;
-        header.make_valid();
-        error = store.write(destination, &header, sizeof(header));
+        uint8_t counter = 0;
+        if (existing) {
+        		counter = uint8_t((existingFooter.counter() + 1) & 3);
+        }
+        error = _write_v2_footer(newSector, existing ? &existingFooter : nullptr, counter);
+        if (error) return error;
+        typename Footer::crc_type crc = computeSectorCRC(newSector);
```
avtolstoy: Why not CRC the header as well?

m-mcgowan: The values in the header is checked fully so no need to CRC it.

technobly:

PR #1297, opened 2017-04-11
Generalize FuelGauge to also use alternative I2C interfaces

ScruffR:
To enable Particle Power Shield to use spark_wiring_fuel without extra
library as mentioned in connection with this issue
https://github.com/spark/firmware/issues/1275

### DIFF
--- a/wiring/src/spark_wiring_fuel.cpp
+++ b/wiring/src/spark_wiring_fuel.cpp
@@ -33,8 +33,22 @@ FuelGauge::FuelGauge()
 
 boolean FuelGauge::begin()
 {
-	// this should be unecessary since, begin is already called from pmic setup
-	return 1;
+#if (PLATFORM_ID == PLATFORM_ELECTRON_PRODUCTION) // what PRODUCT_ID will E0 get?
+  return this->begin(Wire3);  // with Electron the fuel gauge will be attached to Wire3
+#else                                             
+  return this->begin(Wire);   // otherwise we assume it to be on default Wire
+#endif
+}
+
+boolean FuelGauge::begin(TwoWire& i2c)
+{
+  _i2c = &i2c;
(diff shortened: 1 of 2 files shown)


ScruffR:
@mohitbhoite or @technobly, can you check why this keeps failing? 
The error message in the travis log doesn't make sense for me (and my local build passed, only the test seems to fail)

wiring/src/spark_wiring_fuel.cpp
```diff
@@ -175,23 +189,23 @@ void FuelGauge::readConfigRegister(byte &MSB, byte &LSB) {
...
 void FuelGauge::writeRegister(byte address, byte MSB, byte LSB) {
-#if Wiring_Wire3
-	Wire3.beginTransmission(MAX17043_ADDRESS);
-    Wire3.write(address);
-    Wire3.write(MSB);
-    Wire3.write(LSB);
-    Wire3.endTransmission(true);
-#endif
-}
+  if (_i2c) {
```
m-mcgowan: are the pointer checks necessary? static analysis shows that they are always initialized from a reference and can never be null.

m-mcgowan:
The unit tests are failing - you can build these locally by running make from the `user/tests/unit` directory. Also please rebase on develop, thanks :-)

wiring/src/spark_wiring_fuel.cpp
ScruffR: maybe not for `FuelGauge::begin()` but if `FuelGauge::begin()` wasn't called prior any function that uses the private `_i2c` pointer it'll fail.
I'll update my PR accordingly (or file a new one, since I'm not sure how to update ;-))

ScruffR:
This PR has been superceded by https://github.com/spark/firmware/pull/1443

PR #1288, opened 2017-04-04
Support for Twilio SIMs by default in system firmware [ch1537]

technobly:
### Problem

Support for Twilio SIMs by default in system firmware was required: to use a custom port, avoid requiring a custom user app and maintain connectivity in Safe Mode.

### Solution

- Twilio SIM have a sequential range of IMSI (SIM serial number) numbers that can be detected.
- Port 4500 is used for Twilio vs. the default port 5864.
- UDP Server port is set at runtime based on Network Provider.
- Keep alive ping defaults are set at runtime based on Network Provider, although they are currently still 23 minutes for all.  They can always be overridden in a user app with Particle.keepAlive(seconds);
- Tests included

### Steps to Test

- [x] Run Unit Tests from the root directory (firmware $):
```
source ci/install_boost.sh
ci/build_boost.sh
cd user/tests/unit && make -s clean all
```
- [x] Build tinker with Serial1 debugging (included)
	- watch logs for correct setup of network provider and port used.
	- swap between SIMs under different methods and ensure proper connection to server.
		- hot swapping (no reset)
		- hot swapping (reset after)
		- holding reset
		- during soft power down mode
		- from powered off state
		- booting into Safe Mode
		- using the 3rd party SIM API (tested with AT&T)
- [x] Point Electron at test server with port 4500 enabled and test different SIMs (TODO: Monday)

### Example App

```c
/*
 ******************************************************************************
  Copyright (c) 2015 Particle Industries, Inc.  All rights reserved.

  This program is free software; you can redistribute it and/or
  modify it under the terms of the GNU Lesser General Public
  License as published by the Free Software Foundation, either
  version 3 of the License, or (at your option) any later version.

  This program is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  Lesser General Public License for more details.

  You should have received a copy of the GNU Lesser General Public
  License along with this program; if not, see <http://www.gnu.org/licenses/>.
  ******************************************************************************
 */

/* Includes ------------------------------------------------------------------*/
#include "application.h"
#include "stdarg.h"
#include "cellular_hal.h"

PRODUCT_ID(PLATFORM_ID);
PRODUCT_VERSION(3);

Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL);

// 3rd Party SIM credentials
[...]

PR #1300, opened 2017-04-16
Restores public server key and server address if missing

technobly:
### Problem

There have been a number of community reports of devices losing their Public Server Key and/or Server Address.  Typically this is due to a power loss or reset during a critical flash write operation.

### Solution

- Public Server Key and Server Address are baked into System Firmware with this PR, and will be restored in RAM before a connection is attempted if either is found to be missing.
- This data will not be automatically written to FLASH when found missing to prevent further corruption.
- A cloud event will generated by the device alerting the cloud to which error is currently being patched by system firmware. (currently this does not hit the user's event stream)
- A subscription to a cloud event has been added to system firmware that will allow the user to restore the data to FLASH remotely when deemed appropriate. (currently there is no mechanism for the user to do this)
- The TCP server address in PSK is NULL terminated (backup copy in system firmware matches)
- The UDP server address in PSK is NOT NULL terminated (backup copy in system firmware will now include a NULL terminated server address).
- pubkey and private_key arrays in system_cloud_internal.cpp were not initialized and would contain garbage wherever key data was not written to, they are not initiallized to 0xFF to mimick erased flash memory.
- PR #1295 offers hardening and error checking to the FLASH read/write routines to help prevent this PR from being necessary, but just in case we will still be able to recover.

### Steps to Test

- All features have been manually tested on Photon/Electron/Core.
- Use dfu-util to erase [D] and verify [U] data at the appropriate address
	- Photon - dfu-util -d 2b04:d006 -a 1 -s 2082:768 -[D|U] blank.der
	- Electron - dfu-util -d 2b04:d00a -a 1 -s 3298:320 -[D|U] blank.der
	- Core - dfu-util -d 1d50:607f -a 1 -s 4096:768 -[D|U] blank.der
- Change system subscription/publish event name from `spark` to `spork` and publish PUBLIC event `spork/device/key/restore` to restore PSK and Server Address.

### Example App

This is an internal system change. No example application.

### References

- This will likely have a conflict with PR #1288 due to logging code where the public server address is read (prefer #1288's logging changes)
- Related to PR #1295

---

### Completeness

- [x] Added documentation (N/A)

### Enhancement

- [[PR #1300]](https://github.com/spark/firmware/pull/1300) Restores public server key and server address if missing


system/src/system_cloud_internal.cpp
```diff
@@ -475,6 +478,24 @@ void SystemEvents(const char* name, const char* data)
                 System.reset();
         }
     }
+    if (!strncmp(name, KEY_RESTORE_EVENT, strlen(KEY_RESTORE_EVENT))) {
```
avtolstoy: What's the usecase for this? If your key/server address information was corrupted/removed then you won't be able to connect to server at all. And if you are already connected and you are able to receive this event from the server, it means that you probably have valid public key and server address :sweat_smile:

system/inc/system_cloud.h
```diff
@@ -179,6 +197,58 @@ ProtocolFacade* system_cloud_protocol_instance(void);
 
 int spark_set_connection_property(unsigned property_id, unsigned data, void* datap, void* reserved);
 
+// minimal udp public server key
+const unsigned char backup_udp_public_server_key[] = {
```
avtolstoy: These should be defined in a source file instead of a header.

system/src/system_cloud_internal.cpp
technobly: The PSK and server address can be restored in RAM from system flash... so you can connect again. We decided not to automatically restore to DCT/DCD in case of conditions still present that may have caused the first corruption; perhaps they may further corrupt something else during the next flash write. We will track metrics on units reporting corruption, and in the future enable a way for users to send the restore event.

system/inc/system_cloud.h
technobly: Good point, I'll make that change.

system/src/system_cloud_internal.cpp
```diff
@@ -68,7 +68,7 @@ static sock_handle_t sparkSocket = socket_handle_invalid();
 
 ProtocolFacade* sp;
 
-static Flags<ParticleKeyErrorFlag> particle_key_errors;
+static uint32_t particle_key_errors;
```
avtolstoy: Just to be on a safe side I would initialize it to `0`.

technobly: Done!

PR #1289, opened 2017-04-05
WPA/WPA2 Enterprise support [ch1317]

avtolstoy:
### Problem

WPA Enterprise support was required :)

### Solution

Apart from implementing WPA Enterprise support (with EAP-TLS and PEAP/MSCHAPv2 authentication), this PR includes a newer WICED: 3.7.0-7. See the relevant `photon-wiced` branch for a list of WICED fixes/enhancements.

### Steps to Test

https://github.com/avtolstoy/particle-wpa-enterprise-docker

### Example App

TODO

### References

- CH1317
- Closes #914 
- spark/particle-cli#358
- spark/softap-setup-js#34

---

### Completeness

- [X] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [X] Problem and Solution clearly stated
- [ ] Add additional tests
- [ ] Run unit/integration/application tests on device
- [ ] Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

### Feature

- [[PR #1289]](https://github.com/spark/firmware/pull/1289) [[Implements #914]](https://github.com/spark/firmware/issues/914) `[Photon/P1]` WPA/WPA2 Enterprise support added (PEAP/MSCHAPv2 and EAP-TLS)!
- [[PR #1269]](https://github.com/spark/firmware/pull/1269) [[Closes #1165]] Cloud connection can be closed gracefully allowing confirmable messages to reach the cloud before the connection is terminated
- BUGFIX Fixes a stack overlap with system-part2 static RAM on Photon/P1
- BUGFIX Fixes a memory leak when Thread is terminated
- BUGFIX Fixes a deadlock in SoftAP, when connection is terminated prematurely
- [Photon/P1] Automatic cipher/security detection when configuring WiFi settings over Serial

PR #1307, opened 2017-04-23
Change DCT read access to lock()/unlock() or read-with-copy

avtolstoy:
### Problem

When using `dct_read_app_data` the returned pointer might become invalid in case a different thread writes into DCT causing a DCT swap.

Photon/P1 should have a critical section around flash operations and around DCT operations.

### Solution

This branched is based on #1295 and then rebased on #1289.

- Changes DCT read access to lock()/unlock() or read-with-copy. New methods introduced: `dct_read_app_data_lock()`, `dct_read_app_data_unlock()`, `dct_read_app_data_copy()`.
- Adds atomic flag mutex for DCT read/write operations on Photon/P1 (see PR#8 in WICED repo)
- Adds atomic flag mutex for flash operations on Photon/P1 (see PR#8 in WICED repo)

**NOTE: Photon/P1 bootloader builds should overflow. It's intended, as we are removing DCT implementation out of it and importing it from system-part2 instead in https://github.com/spark/firmware/tree/feature/bootloader_dct**

### Steps to Test

N/A

### Example App

N/A

### References

- photon-wiced: PR#8
- #1295
- #1289

---

### Completeness

- [x] Added documentation

### Enhancement

- [`[PR #1307]`](https://github.com/spark/firmware/pull/1307) `[Photon/P1]` New version of WICED adds CRC checking to the DCT implementation so that write errors are detected. Added a critical section around flash operations and around DCT operations to make them thread safe.


services/inc/atomic_flag_mutex.h
```diff
@@ -0,0 +1,19 @@
...
+	void lock() {
+	    while (flag.test_and_set(std::memory_order_acquire)) {
+	        spin();
+	    }
+	}
+
+	void unlock() {
+	    flag.clear(std::memory_order_release);
+	}
+};
```
sergeuz: I would like to see this `AtomicFlagMutex` implementation supporting recursive locking, possibly in next versions of the firmware. For example:
```cpp
void lock() {
    acquire_flag();
    if (!mutex_is_initialized && mutex_can_be_initialized) { // Heap is initialized, etc.
        init_recursive_mutex();
        mutex_is_initialized = true;
    }
    if (mutex_is_initialized) {
        lock_mutex();
        release_flag();
    }
}

void unlock() {
    if (mutex_is_initialized) {
        unlock_mutex();
    } else {
        release_flag();
    }
}
```

avtolstoy: How about something like this?
```c++
template <typename R, R(*spin)()>
class AtomicRecursiveMutex {
    std::atomic<std::thread::id> owner;
    std::atomic_int refcount{0};
public:
    int lock() {
        SPARK_ASSERT(!HAL_IsISR());
        std::thread::id no_id = std::thread::id();
        std::thread::id self = std::this_thread::get_id();
        while ((no_id != std::thread::id() && no_id != self) || !owner.compare_exchange_weak(no_id, self)) {
            if (no_id != std::thread::id() && no_id != self) {
                no_id = std::thread::id();
            }
            spin();
        }
        return ++refcount;
    }

    int unlock() {
        SPARK_ASSERT(!HAL_IsISR());
        SPARK_ASSERT(owner.load() == std::this_thread::get_id());
        int c = --refcount;
        SPARK_ASSERT(c >= 0);
        if (c == 0) {
            // Unlock
            owner.store(std::thread::id());
        }
        return c;
    }
};
```

hal/src/photon/softap.cpp
```diff
@@ -973,9 +973,11 @@ class SoftAPController {
         if (result == WICED_SUCCESS)
         {
             if (memcmp(&expected, soft_ap, sizeof(expected))) {
+                wiced_dct_read_unlock( soft_ap, WICED_FALSE );
                 result = wiced_dct_write(&expected, DCT_WIFI_CONFIG_SECTION, OFFSETOF(platform_dct_wifi_config_t, soft_ap_settings), sizeof(wiced_config_soft_ap_t));
+            } else {
+                wiced_dct_read_unlock( soft_ap, WICED_FALSE );
```
sergeuz: Nice! 👍 I've just checked other places where we use non-copying variant of `wiced_dct_read_lock()`, and haven't found anything suspicious.

technobly:

PR #1309, opened 2017-04-23
[Core] Fixes I2C slave mode

avtolstoy:
### Problem

Core crashes when I2C is configured in slave mode.

### Solution

- Fixes incorrectly named interrupt handlers: `HAL_I2C1_ER_Handler`, `HAL_I2C1_EV_Handler`
- Copied relevant I2C slave code from Photon/Electron implementation

### Steps to Test

`wiring/i2c_master_slave`

### Example App

N/A

### References

Supersedes #1305

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/core/i2c_hal.c
+++ b/hal/src/core/i2c_hal.c
@@ -83,6 +83,8 @@ typedef struct STM32_I2C_Info {
     void (*callback_onReceive)(int);
 
     I2C_Mode mode;
+
+    uint8_t clkStretchingEnabled;
 } STM32_I2C_Info;
 
 /*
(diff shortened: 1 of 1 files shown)


geekbozu:
So I pulled and tested this, I no longer can get valid I2C communication with these changes. Where as with my PR https://github.com/spark/firmware/pull/1305 (which I updated as I did not perform unit testing), I am able to properly send and receive data. I will take a look into this with my logic analyzer tomorrow just wanted to chime in before bed.

geekbozu:
So I think this block is not needed, and removing it fixes the problem.

https://github.com/spark/firmware/blob/57fd7cbc3a16ec6cd2e4829eb62a72800fd1a7cd/hal/src/core/i2c_hal.c#L660-L666

The comment claims that with clock stretching enabled it will pull SCL low if DR is empty. However this is only partially true.

The stm will only stretch the clock if it believes it has more data to send. It decides if there is more data to send based on the Master's ACK/NACK of the last transfered byte. If it recieves an ACK it will stretch the clock until the next byte is ready to be shifted out. How-ever if it recieves a nack it will not hold the clock as according to the [I2C specification](http://cache.nxp.com/documents/user_manual/UM10204.pdf) that is how the master signals to the slave that the current transfer is over. 

Source: 
http://www.st.com/content/ccc/resource/technical/document/reference_manual/59/b9/ba/7f/11/af/43/d5/CD00171190.pdf/files/CD00171190.pdf/jcr:content/translations/en.CD00171190.pdf

PR #1311, opened 2017-04-26
Support for Twilio SIMs by default in system firmware [ch1537]

technobly:
Supersedes PR #1288 so we can release in 0.6.2-rc.2 instead of 0.7.0-rc.1.

---

### Feature

[[PR#1311]](https://github.com/spark/firmware/pull/1311) `[Implements CH1537] [Electron]` Added support for Twilio SIMs by default in system firmware.

### DIFF
--- a/hal/inc/cellular_hal_constants.h
+++ b/hal/inc/cellular_hal_constants.h
@@ -1,0 +1,126 @@
+/*
+ ******************************************************************************
+  Copyright (c) 2015 Particle Industries, Inc.  All rights reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation, either
+  version 3 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, see <http://www.gnu.org/licenses/>.
+  ******************************************************************************
+ */
+
+#ifndef CELLULAR_HAL_CONSTANTS_H
+#define CELLULAR_HAL_CONSTANTS_H
+
+#include "inet_hal.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct __attribute__((__packed__))  _CellularConfig_t {
+    uint16_t size;
+    NetworkConfig nw;
+} CellularConfig;
+
+typedef int cellular_result_t;
+
+typedef int (*_CALLBACKPTR_MDM)(int type, const char* buf, int len, void* param);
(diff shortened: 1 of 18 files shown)


hal/inc/cellular_hal_constants.h
```diff
@@ -0,0 +1,126 @@
...
+
+typedef int cellular_result_t;
+
+typedef int (*_CALLBACKPTR_MDM)(int type, const char* buf, int len, void* param);
+
+typedef void (*_CELLULAR_SMS_CB_MDM)(void* data, int index);
+
+typedef enum { CELLULAR_NETPROV_TELEFONICA =   0,
+               CELLULAR_NETPROV_TWILIO     =   1,
+               CELLULAR_NETPROV_UNKNOWN    = 999 } CellularNetProv;
```
m-mcgowan: why 999? why  not make it the largest value possible?

hal/src/electron/cellular_internal.cpp
```diff
@@ -0,0 +1,26 @@
...
+        if (imsi && strlen(imsi) > 0) {
+            // convert to unsigned long long (imsi can be 15 digits)
+            unsigned long long imsi64 = strtoull(imsi, NULL, 10);
+            // LOG(INFO,"IMSI: %s %lu%lu", imsi, (uint32_t)(imsi64/100000000), (uint32_t)(imsi64-310260800000000));
+            // set network provider based on IMSI range
+            if (imsi64 >= 310260859000000 && imsi64 <= 310260859999999) {
+                return CELLULAR_NETPROV_TWILIO;
+            }
+            else {
+                return CELLULAR_NETPROV_TELEFONICA;
```
m-mcgowan: do we have a imsi range for telefonica? would be good to tighten up this code for future forward compatibility.

system/src/system_task.cpp
```diff
@@ -285,6 +286,16 @@ void establish_cloud_connection()
             return;
         }
 
+#if PLATFORM_ID==PLATFORM_ELECTRON_PRODUCTION
```
m-mcgowan: rather than hard-coding the provider port, APN and other details. it would be more flexible and less code to create structs in the HAL that describe these things as data, keyed on the provider ID, so that we don't have any conditional logic for each provider.

hal/inc/cellular_hal_constants.h
technobly: 1000 providers seemed reasonable, but I'll change it to UINT_MAX.

hal/src/electron/cellular_internal.cpp
technobly: We do not, but that's something we can work on.

system/src/system_task.cpp
technobly: I started there, but opted out because it seemed like more complexity at the time.  I'll look at refactoring it.

hal/inc/cellular_hal_constants.h
technobly: I ended up changing this to a CELLULAR_NETPROV_MAX enum, to be used only in the future if we need to cycle through providers.  Right now we test and set one of two.

PR #1317, opened 2017-05-02
Changed branch to checkout from 'latest' to 'release/stable'

RaemondBW:
### Problem
Documentation still points at using the 'latest' branch which no longer exists.

### Solution
Switched from 'latest' to 'release/stable'

### References
https://docs.particle.io/faq/particle-tools/local-build/photon/


technobly:
@RaemondBW thanks for that catch!

PR #1320, opened 2017-05-03
DCT locking with recursive mutex

avtolstoy:
### Problem

Same thing as #1307, but we are switching from an atomic spinlock to recursive FreeRTOS mutex.

### Solution

Switch to recursive FreeRTOS mutex. Apart from that we are also upgrading to FreeRTOS v9.0.0. See also spark/photon-wiced#10

### Steps to Test

N/A

### Example App

N/A

### References

Supersedes #1307

---

### Completeness

- [ ] Added documentation


hal/inc/concurrent_hal.h
```diff
@@ -179,6 +179,7 @@ int os_mutex_trylock(os_mutex_t mutex);
 int os_mutex_unlock(os_mutex_t mutex);
 
 int os_mutex_recursive_create(os_mutex_recursive_t* mutex);
+int os_mutex_recursive_create_static(os_mutex_recursive_t* mutex);
```
avtolstoy: Yes, that was my intention initially. I'd like to see static variants for `os_primitive_create` as well in the future, but for the intents and purposes of this PR, adding them here seems like an overkill.

hal/src/stm32f2xx/concurrent_hal.cpp
```diff
@@ -489,3 +496,47 @@ int os_timer_is_active(os_timer_t timer, void* reserved)
...
+    flash_lock.lock();
+}
+
+void __flash_release() {
+    flash_lock.unlock();
+}
+
+#define DCT_LOCK_TIMEOUT 10000
+
+static RecursiveMutex dctLock(DCT_LOCK_TIMEOUT);
```
avtolstoy: Seem useful at least for debugging. There is an assert that checks whether the lock was successfully acquired or not.

services/inc/mutex.h
```diff
@@ -0,0 +1,47 @@
...
+    }
+    int unlock() {
+        return xSemaphoreGiveRecursive(mutex_);
+    }
+
+private:
+    SemaphoreHandle_t mutex_ = NULL;
+    StaticSemaphore_t buffer_;
+    uint32_t blockTime_ = 0;
+};
```
avtolstoy: +1 about relying on HAL, but I think we should implement static initializers for `os_` as a separate feature.

hal/inc/hal_dynalib_concurrent.h
```diff
@@ -61,6 +61,8 @@ DYNALIB_FN(24, hal_concurrent, os_queue_destroy, int(os_queue_t, void*))
 DYNALIB_FN(25, hal_concurrent, os_queue_put, int(os_queue_t, const void* item, system_tick_t, void*))
 DYNALIB_FN(26, hal_concurrent, os_queue_take, int(os_queue_t, void* item, system_tick_t, void*))
 DYNALIB_FN(27, hal_concurrent, os_thread_exit, os_result_t(os_thread_t))
+DYNALIB_FN(28, hal_concurrent, dct_lock, int(int, void*))
+DYNALIB_FN(29, hal_concurrent, dct_unlock, int(int, void*))
```
avtolstoy: This is mainly for electron, where DCT (or DCD) implementation is in `libplatform`, which gets linked into `system-part1`, causing linking error because FreeRTOS functions are not available there. Another solution for this could be implementing `os_mutex_recursive_create_ex` and using HAL functions for `RecursiveMutex` implementation.

avtolstoy:
Regarding  https://github.com/spark/firmware/pull/1320/commits/19eacffb20f40ca3275c86ba863e268a727489c5:

_(parts here are module indices, not filenames)_

| develop                                                                                                                 | here                                                                                                                   |
|-------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------|
| electron_before | electron_after |

Regarding https://github.com/spark/firmware/pull/1320/commits/b721db35f19bdf0f75189f8a4034c08aa9c3a6fa:

| develop                                                                                                                | feature/photon/wiced-3.7.0-7                                                                                            | here                                                                                                                 |
|------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------|
| photon_develop | photon_wiced370 | photon_after |

PR #1321, opened 2017-05-04
Fix misleading indent making PLATFORM_ID=3 fail

monkbroc:
Reformat socket_hal.cpp with clang-format to fix this error when compiling with PLATFORM_ID=3:
```
src/gcc/socket_hal.cpp: In function ‘sock_result_t socket_sendto(sock_handle_t, const void*, socklen_t, uint32_t, sockaddr_t*, socklen_t)’:
src/gcc/socket_hal.cpp:366:5: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
     if (result == boost::asio::error::would_block)
     ^~
src/gcc/socket_hal.cpp:369:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
  return result ? result : count;
```

### Steps to reproduce

```
firmware/main$ make all PLATFORM_ID=3
```

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/gcc/socket_hal.cpp
+++ b/hal/src/gcc/socket_hal.cpp
@@ -42,75 +42,65 @@
 namespace ip = boost::asio::ip;
 
 const sock_handle_t SOCKET_COUNT = sock_handle_t(8);
-const sock_handle_t SOCKET_MAX =  SOCKET_COUNT*2;
+const sock_handle_t SOCKET_MAX = SOCKET_COUNT * 2;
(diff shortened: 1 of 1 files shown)


m-mcgowan:
That's a whole bunch of changes just to fix a whitespace issue :-)

monkbroc:
This was back in the day when we tried using clang-format. You can remove all the reformatting and just fix the indent of the line with `if (result == boost::asio::error::would_block)`

PR #1323, opened 2017-05-08
Process USB vendor requests in system thread

avtolstoy:
### Problem

Most of the currently supported vendor requests should be executed on system thread instead of being processed in ISR.

### Solution

Utilize `SystemISRTaskQueue` for vendor requests.

### Steps to Test

N/A

### Example App

N/A

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/inc/system_control.h
+++ b/system/inc/system_control.h
@@ -55,7 +55,8 @@ typedef enum USBRequestType {
   USB_REQUEST_DFU_MODE = 50,
   USB_REQUEST_SAFE_MODE = 60,
   USB_REQUEST_LISTENING_MODE = 70,
-  USB_REQUEST_LOG_CONFIG = 80
+  USB_REQUEST_LOG_CONFIG = 80,
+  USB_REQUEST_MODULE_INFO = 90
 } USBRequestType;
 
 typedef enum USBRequestResult {
@@ -66,6 +67,7 @@ typedef enum USBRequestResult {
 typedef struct USBRequest {
   size_t size; // Structure size
   int type; // Request type (as defined by USBRequestType enum)
+  uint16_t value; // wValue field
   char* data; // Data buffer
   size_t request_size; // Request size
   size_t reply_size; // Reply size (initialized to 0)
--- a/system/inc/system_control.h
+++ b/system/inc/system_control.h
@@ -67,11 +67,11 @@ typedef enum USBRequestResult {
 typedef struct USBRequest {
   size_t size; // Structure size
   int type; // Request type (as defined by USBRequestType enum)
-  uint16_t value; // wValue field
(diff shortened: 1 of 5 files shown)


system/src/system_control.cpp
```diff
@@ -181,10 +170,13 @@ uint8_t SystemControlInterface::handleVendorRequest(HAL_USB_SetupRequest* req) {
         } else {
           // Return as string
           String id = System.deviceID();
```
sergeuz: Unused variable

system/inc/system_control.h
```diff
@@ -66,6 +67,7 @@ typedef enum USBRequestResult {
 typedef struct USBRequest {
   size_t size; // Structure size
   int type; // Request type (as defined by USBRequestType enum)
+  uint16_t value; // wValue field
```
sergeuz: This change breaks ABI (new fields should be added at the end of the structure)

system/src/system_network_internal.h
```diff
@@ -279,6 +279,7 @@ class ManagedNetworkInterface : public NetworkInterface
                 console.loop();
             }
 #if PLATFORM_THREADING
+            SystemISRTaskQueue.process();
```
sergeuz: 👍

system/src/system_control.cpp
```diff
@@ -292,15 +286,51 @@ void SystemControlInterface::setRequestResult(USBRequest* req, USBRequestResult
...
+    break;
+  }
+
+  case USB_REQUEST_DFU_MODE: {
+    System.dfu(false);
+    break;
+  }
+
+  case USB_REQUEST_LISTENING_MODE: {
+    setRequestResult(req, result);
```
sergeuz: Is there any problem with setting a result after calling `WiFi.listen()`? I just would like to see error handling unified in this function (if possible), i.e. make all case blocks to either call `setRequestResult()` and return, or assign to `result` variable.

system/src/system_control.cpp
```diff
@@ -292,15 +286,51 @@ void SystemControlInterface::setRequestResult(USBRequest* req, USBRequestResult
...
+    break;
+  }
+
   default:
     if (usbReqAppHandler) {
       processAppRequest(data); // Forward request to the application thread
     } else {
       setRequestResult(req, USB_REQUEST_RESULT_ERROR);
     }
+    return;
```
sergeuz: Same here:
```cpp
if (usbReqAppHandler) {
  processAppRequest(data);
  return;
} else {
  result = USB_REQUEST_RESULT_ERROR;
}
```

avtolstoy:
Thanks for the comments! Updated PR.

PR #1324, opened 2017-05-09
Verify the bootloader was flashed correctly and retry if not

avtolstoy:
### Notes

_Targets `feature/dct-locking-recursive-mutex` for now because it's based on it. Makes it easier to review the changes. Should be retargeted to `feature/photon/wiced-3.7.0-7` later._

### Problem

As a safety measure we should verify that when updating bootloader OTA or over YModem the copy operation from ota_module area to the bootloader area has succeeded and the bootloader image in bootloader_module area passes the checks. If not, there is nothing we can do but retry until it does.

### Solution

Verify the bootloader was flashed correctly and retry if not.

### Steps to Test

N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/photon/core_hal.c
+++ b/hal/src/photon/core_hal.c
@@ -122,14 +122,14 @@ void HAL_Core_Setup_finalize(void)
     // in the eeprom region.
     const module_info_t* app_info = FLASH_ModuleInfo(FLASH_INTERNAL, app_backup);
     if (app_info->module_start_address==(void*)0x80A0000) {
-    		LED_SetRGBColor(RGB_COLOR_GREEN);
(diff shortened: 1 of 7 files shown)


technobly:
@avtolstoy have you force tested this works well with interrupts disabled? https://github.com/spark/firmware/blob/47c429555ea3a549e4927e5f5d05c2597191e072/hal/src/stm32f2xx/ota_flash_hal_stm32f2xx.cpp#L235  I don't see why it shouldn't, but would be good to test.

avtolstoy:
@technobly Yeah I did. Added a code block to `bootloader_update` to fail first 10 times and succeed on 11th.

PR #1325, opened 2017-05-10
Minimize DCT usage in bootloader

sergeuz:
### Problem

- Bootloader doesn't fit into its flash region on P1 with the newer WICED.
- Bootloader unnecessarily stores and updates system flags in DCT, potentially increasing chance of a DCT corruption.

### Solution

- Load DCT functions dynamically in bootloader on P1 and Photon.
- Use backup registers to store system flags.

### Steps to Test

1. Build 0.6.2 and flash it to the P1 or Photon (downgrade bootloader if necessary).
2. Build this branch with the following changes: bump firmware version to 0.7.0-rc.1 (expected module version is 200; use [this commit](https://github.com/spark/firmware/commit/ed6f66405afe17dee90a14ae3e81599fed1627da) as a reference), bump bootloader version, run `hal/src/stm32f2xx/image_bootloader.sh` (optional).
3. Flash the newer part1, part2 and bootloader to the device OTA. Ensure that all modules have expected version numbers via `particle serial inspect`. **Note:** `feature/photon/wiced-3.7.0-7`, which this branch is based on, doesn't update bootloader automatically.
4. Check that the bootloader's DCT functionality works as expected: device/server keys can be downloaded/uploaded, LED mirroring and themes work, etc.
5. Downgrade part2 back to 0.6.2. Any attempts to write to the DCT while the device is in DFU mode should fail now. Read operations should return null data (TODO: would be nice to make them fail explicitly as well).
**Note:** At this point, attempt to downgrade part1 will fail silently, since the bootloader can't access DCT and read module update info, thus, in order to downgrade from 0.7.0-rc.1, firmware modules should be flashed in the following order: part2 -> bootloader -> part1. **Make sure release notes are updated accordingly.**
6. Downgrade bootloader and part1 to 0.6.2. Use `particle serial inspect` to check that all modules have correct version numbers.
7. Run unit tests.

(feel free to extend the list!)

### Example App

N/A

### References

This PR supersedes #1320 and #1307.

---

### Completeness

- [ ] Added documentation


hal/src/stm32f2xx/static_recursive_mutex.h
```diff
@@ -0,0 +1,50 @@
...
+        if (mutex_) {
+            vSemaphoreDelete(mutex_);
+        }
+    }
+
+    bool lock() {
+        if (!mutex_) {
+            init();
+        }
+        return (xSemaphoreTakeRecursive(mutex_, portMAX_DELAY) == pdTRUE);
```
avtolstoy: I still think we should add a timeout here at least in debug builds to make it easier to spot incorrect usage of DCT. That way the timeout will trigger an assert in `dct_lock` instead of a deadlock.

avtolstoy:
> Note: At this point, attempt to downgrade part1 will fail silently, since the bootloader can't access DCT and read module update info, thus, in order to downgrade from 0.7.0-rc.1, firmware modules should be flashed in the following order: part2 -> bootloader -> part1. Make sure release notes are updated accordingly.

We could also add a dependency to the new bootloader on system-part1. That way after downgrading system-part2 and attempting to downgrade system-part1 the dependency check will fail (even if we are already running pre-0.7.0 system-part2) and the device will not even attempt to replace the module. After the bootloader is downgraded (to pre-0.7.0 which doesn't have any dependencies), downgrade of system-part1 (and system-part3 (module index) on Electron) should succeed.

platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/system_flags_impl.c
```diff
@@ -0,0 +1,79 @@
+#include "system_flags_impl.h"
+
+#include "stm32f2xx_rtc.h"
+
+#include <string.h>
+
+#define SYSTEM_FLAGS_MAGIC_NUMBER 0x1ADEACC0u
+
+int Load_SystemFlags_Impl(platform_system_flags_t* f) {
```
sergeuz: Fixed in `Load_SystemFlags()` and `Save_SystemFlags()` which call these `_Impl` functions.

PR #1328, opened 2017-05-10
Revert "Verify the bootloader was flashed correctly and retry if not"

technobly:
Reverts spark/firmware#1324

missed this step:
>Targets feature/dct-locking-recursive-mutex for now because it’s based on it. Makes it easier to review the changes. Should be retargeted to feature/photon/wiced-3.7.0-7 later.


technobly:
Didn't need this after all...

PR #1337, opened 2017-05-18
Listening mode fixes

sergeuz:
### Problem

The system attempts to enter listening mode every 1ms when the SETUP button is pressed.

### Solution

Restore pre-0.6.0 behavior and make `ManagedNetworkInterface::listening()` return `true` not only when listening mode is already active, but also if it's pending for execution.

### Steps to Test

Add some logging before `network.listen()` gets called in `HAL_SysTick_Handler()`.

### Example App

N/A

### References

Original issue due to which `WLAN_SMART_CONFIG_START` flag was separated into `WLAN_SMART_CONFIG_START` and `WLAN_SMART_CONFIG_ACTIVE` flags: https://github.com/spark/firmware/issues/1013

---

### Completeness

- [ ] Added documentation

---

### Bug fix

- [`[PR #1337]`](https://github.com/spark/firmware/pull/1337) Fixed system attempting to enter listening mode every 1ms when the SETUP button is pressed.


m-mcgowan:
how about using 2 flags - listening (for real) and listening pending?

sergeuz:
There are 2 flags already: `WLAN_SMART_CONFIG_START` (listening mode is pending) and `WLAN_SMART_CONFIG_ACTIVE` (listening loop actually is running) – the latter one was introduced to fix https://github.com/spark/firmware/issues/1013. In this PR I just updated `network_listening()` to return `true` if any of the flags is set – previously it was checking only `WLAN_SMART_CONFIG_ACTIVE`.

PR #1338, opened 2017-05-18
Do not read or write feature flags from an ISR

sergeuz:
### Problem

Device panic (assertion failure) occurs when DFU mode is triggered via baud rate change.

### Solution

Do not read or write feature flags, which are stored in DCT, from an ISR. Users are still allowed to call such methods as `System.reset()` and `System.dfu()` from an ISR to preserve backwards compatibility.

### Steps to Test

- Try to enable DFU mode via baud rate change.
- Check what happens with System.resetReason(), characterize the behavior.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/stm32f2xx/core_hal_stm32f2xx.c
+++ b/hal/src/stm32f2xx/core_hal_stm32f2xx.c
@@ -263,33 +263,44 @@ static void Init_Last_Reset_Info()
 
 static int Write_Feature_Flag(uint32_t flag, bool value, bool *prev_value)
 {
+    if (HAL_IsISR()) {
+        return -1; // DCT cannot be accessed from an ISR
+    }
     uint32_t flags = 0;
-    dct_read_app_data_copy(DCT_FEATURE_FLAGS_OFFSET, &flags, sizeof(flags));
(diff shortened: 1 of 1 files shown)


hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -1270,10 +1280,9 @@ int HAL_Feature_Set(HAL_Feature feature, bool enabled)
                 current |= 2;   // 0bxxxxxx10 to disable the clock, any other value to enable it.
             }
             dct_write_app_data(&current, DCT_RADIO_FLAGS_OFFSET, 1);
+            return 0;
```
technobly: Any reason we don't use the same error checking pattern here?
```c++
int result = dct_read_app_data_copy(DCT_RADIO_FLAGS_OFFSET, &current, sizeof(current));
if (result != 0) {
    return result;
}
//...
result = dct_write_app_data(&current, DCT_RADIO_FLAGS_OFFSET, 1);
if (result != 0) {
    return result;
}
```

PR #1342, opened 2017-06-01
Remove the firmware docs from the repo

monkbroc:
Make docs changes directly in https://github.com/spark/docs/blob/master/src/content/reference/firmware.md


monkbroc:
I'll let you merge this. I don't want to disrupt other PRs that might have docs still in them.

technobly:
FYI, to remove the subtree from local `spark/firmware` use `git remote rm firmware-docs`

PR #1343, opened 2017-06-07
Fix endless IWDG reset loop

sergeuz:
### Problem

In order to minimize usage of the DCT in the bootloader, the firmware now uses backup registers to store system flags (see https://github.com/spark/firmware/pull/1325). This poses a problem when firmware is downgraded from 0.7.0-rc.1: bootloader enables IWDG and sets a flag in the backup registers, but an older firmware is unable to detect that IWDG is running since it expects IWDG flag to be set in the DCT. This causes the device to reset every 5 seconds.

### Solution

Do not enable IWDG if currently flashed firmware has a version number lower than 0.7.0-rc.1. This gives user a possibility to downgrade the bootloader properly.

### Steps to Test

1. Set all module versions to 200 (0.7.0-rc.1) in [system_module_version.mk](https://github.com/spark/firmware/blob/fix/disable_iwdg/modules/shared/system_module_version.mk).
2. Compile and flash all system parts **and bootloader** to the device.
3. Flash 0.6.2 firmware via DFU. The device should work normally without resetting unexpectedly.

### Example App

N/A

---

### Completeness

- [ ] Added documentation


bootloader/src/main.c
```diff
@@ -414,8 +417,19 @@ int main(void)
...
             __set_MSP(*(__IO uint32_t*) ApplicationAddress);
 
-            // Set IWDG Timeout to 5 secs based on platform specific system flags
-            IWDG_Reset_Enable(5 * TIMING_IWDG_RELOAD);
+            uint8_t disable_iwdg = 0;
+#ifdef CHECK_FIRMWARE
+            // Pre-0.7.0 firmwares were expecting IWDG flag to be set in the DCT, now it's stored in
+            // the backup registers. As a workaround, we disable IWDG if an older firmware is detected
+            const uint16_t module_ver = get_main_module_version();
+            if (module_ver >= 0 && module_ver < SYSTEM_MODULE_VERSION_0_7_0_RC1) {
```
technobly: Should this be MIN_MODULE_VERSION_SYSTEM_PART2 instead of SYSTEM_MODULE_VERSION_0_7_0_RC1?

Also should `module_ver` be declared as `const int16_t` to match return sign from `get_main_module_version()`? Otherwise default returned value of -1 when module version is not identifiable will be greater than 200 and result in IWDG not being disabled.  Part of this question is correct logic, and other part what should the default be in case of module corruption?  Seems like we'd need to disable the IWDG to be able to DFU new firmware, so we'd want to always default to IWDG disabled.

sergeuz: > Should this be MIN_MODULE_VERSION_SYSTEM_PART2 instead of SYSTEM_MODULE_VERSION_0_7_0_RC1?

No, `MIN_MODULE_VERSION_SYSTEM_PART2` is internal for `bootloader_dct.c`.

> Also should module_ver be declared as const int16_t to match return sign from get_main_module_version()?

Yeah, `module_ver` should be a signed integer, thanks. I first used 0 as error code, but then it appeared that 0 is actually used as a version number for all current monolithic builds.

> Part of this question is correct logic, and other part what should the default be in case of module corruption? Seems like we'd need to disable the IWDG to be able to DFU new firmware, so we'd want to always default to IWDG disabled.

This code actually doesn't check integrity of any system modules, it only checks version of the main system module (which is either part2 or monolithic firmware module) to workaround particular issue with downgrading from 0.7.0. There's an existing logic in the bootloader, which turns the device into DFU mode once a specific number of IWDG resets is reached, though I never seen it in action.

technobly: Re: corrupted module, worst case example: if a module was simply and completely erased, would `get_main_module_version()` return -1?  Basically if the module header is corrupted/erased, will get_module_info() return NULL?  I guess it's also possible that the header is corrupted and get_main_module_version() returns something > 200 by mistake.

sergeuz: Right, `get_module_info()` (and thus `get_main_module_version()`) returns data straight from the module's header on flash, so if the header is corrupted in a weird way it's possible that the IWDG won't be enabled. We can add an integrity check here, but I'd like to ensure that it won't affect the boot time. There's a similar check already, which is performed when DCT functions are loaded, but it doesn't happen on every boot, only on demand.

technobly: @sergeuz Can we bound it also to the current SYSTEM_PART1_MODULE_VERSION, so if the header was corrupt and greater than SYSTEM_PART1_MODULE_VERSION (at the time bootloader was compiled), it would disable the IWDG as well.  That might mean updating to a new system firmware would disable the IWDG though... hmm, how about just giving it some kind of sane upper bounds for the time being like (SYSTEM_PART1_MODULE_VERSION + 500) to allow for some upgrade buffer but still protect against 99% of corrupt values `((65536-500)/65536)*100`.  If nothing else, we could protect against the erased state (0xFFFF) if that doesn't result in get_module_info() returning NULL but instead returning 65535 for `module_ver`.

sergeuz: It's the opposite: we don't want to enable IWDG for module versions < 200 (0.7.0-rc.1), since older firmwares won't be able detect if IWDG is running. In other words, we're enabling IWDG for any firmware version in the range 200-65535, which includes most of the potentially corrupted values and erased state.

technobly: Ah true, and in that case the bootloader will end up in DFU mode after a few tries and resets due to IWDG, yeah?

sergeuz: Yep. Ideally, bootloader should always check integrity of a newly flashed firmware with all its dependencies, and then save a flag to the DCT to avoid checking it until next update. This will allow bootloader to enter DFU mode instantly in case of a corrupted firmware, without relying on IWDG (which, in turn, ideally should be kept enabled all the time).

PR #1367, opened 2017-08-15
Use the device ID as the USB serial number for Core

monkbroc:
### Problem

When connecting a Core to USB, the serial number reported is a strange number instead of the device ID. The Photon and Electron report their device ID.

```
$ lsusb -v | grep Particle -A 2
  iManufacturer           1 Particle          
  iProduct                2 Spark Core with WiFi    
  iSerial                 3 6D75569D4957
```

This is blocking the CLI from using the USB serial number to detect the device ID of any connected device.

### Solution

Port USB serial number code from Photon/Electron to Core.

Note: DFU mode still reports the original serial number instead of the device ID. I'm not sure how to recompile and flash the piece of software that contains DFU to a Core. It would be good to update in the long run, but it's not crucial to do it right away.

### Steps to Test

Connect Core to USB, flash system firmware from PR.

```
$ lsusb -v | grep Particle -A 2
  iManufacturer           1 Particle          
  iProduct                2 Spark Core with WiFi    
  iSerial                 3 54FF72011111111111111111
```

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/src/core/dfu_usb.c
+++ b/bootloader/src/core/dfu_usb.c
@@ -31,6 +31,9 @@
 #include "usb_pwr.h"
 #include "dfu_mal.h"
 #include "core_hal.h"
+#include "deviceid_hal.h"
+#include "bytes2hexbuf.h"
+#include <string.h>
 
 /* Private typedef -----------------------------------------------------------*/
 /* Private define ------------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


monkbroc:
Sure for the bootloader. I had not found the place where the bootloader
logic lived. It's not worth the trouble of updating the bootloader just for
this but when the next strong reason to update the bootloader comes this
fix will be ready.

monkbroc:
Reading this back it didn't sound clear. I think fixing the bootloader code now is fine, but if my understanding that actually updating the bootloader on devices out in the wild is difficult this change is not enough to justify asking people to update their bootloader.

m-mcgowan:
@avtolstoy - ok to merge this one in 0.8.0-rc.2?

avtolstoy:
@m-mcgowan If we are not going to update the same code in the bootloader (https://github.com/particle-iot/firmware/blob/60f703bd4165a292f89124a7e3755e4eb37b1a03/bootloader/src/core/dfu_usb.c#L96), then yes.

PR #1369, opened 2017-08-17
[CH6178] Checkpoints/diagnostics

avtolstoy:
### Problem

When a device hard faults, crashes, freezes or is otherwise misbehaving, it is presently difficult to know what the last action the device took is and whether the fault lies in application code, system code, peripheral code etc..

A checkpoints API would allow application and system firmware to record their current execution progress and provide an indication of where execution halted prior to the crash. This information is then published to the cloud on next successful connection.

It is also possible to get stack trace of all the threads using a naive stack unwinder, that just scans through the thread stack, looking for flash addresses (within system and user part bounds) and checks if there is a branching instruction at that location.

It would be useful to store per-thread stack traces in addition to checkpoint info whenever the device enters a hardfault, a panic state, or on demand. This will provide a better overview of the system state and help understand the cause of the crash or deadlock for difficult to reproduce issues.

### Solution

System code may use the following macros defined in `diagnostic.h`:
- `DIAGNOSTIC_CHECKPOINT()` - regular instruction address type checkpoint
- `DIAGNOSTIC_PANIC_CHECKPOINT()` - special instruction address type checkpoint which should only be called from a panic handler, which will prevent further modification of diagnostic data
- `DIAGNOSTIC_CRASH_CHECKPOINT(pc)` - special instruction address type checkpoint which should only be called from a crash (hardfault) handler, which will prevent further modification of diagnostic data. An instruction address needs to be manually passed as an argument.
- `DIAGNOSTIC_UPDATE()` - forces a full system state update (including stacktraces).

All `LOG` statements, except for `LOG_DUMP`, make a call to `DIAGNOSTIC_CHECKPOINT()`.
Both hardfault and panic handlers make a call to their respective macros:  `DIAGNOSTIC_CRASH_CHECKPOINT()`, `DIAGNOSTIC_PANIC_CHECKPOINT()`

Application code may use the following macros defined in `spark_wiring_diagnostic.h`:
- `CHECKPOINT()` - a standard variant of `CHECKPOINT()` macro. Implementation depends on a preprocessor macro `DIAGNOSTIC_ELF_AVAILABLE`.
  - When defined and equal to 1, `CHECKPOINT()` macro internally calls `DIAGNOSTIC_CHECKPOINT()`, making it an instruction address type checkpoint
  - When undefined or set to 0, `CHECKPOINT()` macro internally makes a call to `DIAGNOSTIC_TEXT_CHECKPOINT()`, which saves the location of the checkpoint in textual format: `__FILENAME__:__LINE__`
- `CHECKPOINT(text)` - a specialized variant of `CHECKPOINT()` macro, which forces a textual checkpoint with the text passed as an argument

All `Logger` class calls, except for `print`, `printf`, `write` and `dump`, internally include a call to `_LOG_CHECKPOINT()` macro defined in spark_wiring_diagnostic.h, which is only enabled when `DIAGNOSTIC_ELF_AVAILABLE == 1`. It acquires an address of the calling function and uses that as the checkpoint instruction address.

See CH6178 for implementation details.

### Steps to Test

- unit tests
- app/checkpoint (see README.md)

### Example App

- app/checkpoint

### References

- CH6178

---

### Completeness

- [ ] Added documentation (required)

---

### Feature

- [`[PR #1369]`](https://github.com/spark/firmware/pull/1369) `[Photon/P1/Electron]` Added Checkpoints/diagnostics


hal/inc/concurrent_hal.h
```diff
@@ -128,6 +131,25 @@ os_result_t os_thread_cleanup(os_thread_t thread);
...
+typedef struct {
+  uint8_t reserved;
+
+  os_thread_t thread;
+  const char* name;
+  os_unique_id_t id;
+  void* stack;
+  void* stack_start;
+  void* stack_end;
+} os_thread_dump_info_t;
```
technobly: If we wanted to expand this structure later while retaining backwards compatibility, what would we need to add/change now?  A reserved void * at the end?

avtolstoy: It's currently only used within a single module and is not exported, so that shouldn't be a concern at the moment. If we later decide to export it, the first `reserved` becomes a `version` field as usual.

technobly: Thanks for pointing that out @avtolstoy, sounds good.

technobly:
This great feature is also going to need some docs to ensure users understand how to use it.

m-mcgowan:
Andrey, please fix up the conflicts so we can rebase for 0.8.0-rc.2 release.

avtolstoy:
We also need to rename all the mentions of *diagnostic* to trace/tracing as originally mentioned in [CH8404] to avoid the confusion.

PR #1373, opened 2017-08-21
wiring++

m-mcgowan:
adds a PoC for pin objects, so you can write

```
if (D0.read()) {   
  D1.write(LOW);
}
```

We might later extend to allow direct conversion and assignment, e.g.

```
if (D0) {
    D1 = HIGH;
}
```

Having objects as pins allows for many possibilities, including traits - preventing a pin from being used in a way that is not compatible with that pin.

### DIFF
--- a/hal/inc/pinmap_defs.h
+++ b/hal/inc/pinmap_defs.h
@@ -1,0 +1,71 @@
+#pragma once
+
+#include <stdint.h>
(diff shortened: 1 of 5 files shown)


m-mcgowan:
I'm going to close this since it was there for experimentation and to see if it raised any interest.

PR #1374, opened 2017-08-23
Added info on the two major things that prevented me from successfull…

jenschr:
When setting up for local compilation the first time, there was two things that caused me problems. I believe all others will experience the same unless they have a solid C++ background, so I suggest adding these two bits of info to the Getting Started docs.

Ref: https://community.particle.io/t/memory-profiling/31817/23

PS: the link to https://www.particle.io/resources at the bottom of the document is dead. I don't know where to point it to, so I'm just mentioning it here.

### DIFF
--- a/docs/gettingstarted.md
+++ b/docs/gettingstarted.md
@@ -55,6 +55,12 @@ The [makefile documentation](build.md) describes the build options supported and
 * You get `make: *** No targets specified and no makefile found.  Stop.`
   Solution: `cd firmware/main`
 
+* GCC does not know what an .ino file is, so your application is empty.
+  Solution: rename yourapp.ino file to yourapp.cpp
+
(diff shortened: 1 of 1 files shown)


m-mcgowan:
Thank you for this! You are totally right, that the local compile option does assume a lot of background knowledge at present. These chances will help make that clearer.

Our longer-term plan is to make the local compile available via our Desktop IDE which will make setup almost trivial, and hide some of these "pitfalls".   Thanks again!

jenschr:
Glad to help. It does take quite some time until one can submit something useful when this isn't all you do. I'm hoping to put together the tutorial I myself would have needed with hints like this https://community.particle.io/t/memory-profiling/31817/26

jenschr:
Done :-)

Issue #1304, opened 2017-04-21
Cellular resolve does not return 0 / false when it receives bad DNS resolution related to bad cell service


hwestbrook:
### Expected Behavior

Based on the documentation, I would expect IP address to be 0.0.0.0 and an if statement pertaining to the IP address to resolve to false in the scenario that the modem cannot resolve the hostname's ip address.

### Observed Behavior

I am running cellular resolve in my code and when my device has bad connectivity, the IP address resolves to the wrong address without throwing an error. The below snippet recreates the issue. 

For google.com the expected result is: 172.217.23.46. I've received: 
- 0.0.174.11
- 0.0.0.8

A tangential and possibly related issue is that when the device is trying to resolve hostname and does not have cell signal it attempts to reset the WLAN and acquire a cell signal again, meanwhile blocking the loop. Is this expected behavior? How might I exit from this block if I need to run critical application code? It blocks for quite a while...

### Steps to Reproduce

Run the below test app and then detach the cellular antenna to your device.

### Test App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);
SerialLogHandler logHandler(LOG_LEVEL_INFO);

void setup() {
	Serial.begin(9600);
	while(!Serial.available()) {}
	Serial.read();
	Cellular.on();
	Cellular.connect();
	waitFor(Cellular.ready, 120000);
}

void loop() {
	CellularSignal sig = Cellular.RSSI();
	IPAddress ip = Cellular.resolve("google.com");
	Serial.printlnf("Google IP Address: %s. This is a %s bool. Cell reception is %i / %i.", ip.toString().c_str(), ip ? "true" : "false", sig.rssi, sig.qual);
	delay(3000);
}
```

### References

Here is some logging:

```
0000004523 [system] INFO: ready(): 0; connecting(): 0; listening(): 0; WLAN_SMART_CONFIG_START: 0
0000004523 [system] INFO: Sim Ready
0000004523 [system] INFO: ARM_WLAN_WD 1
0000004995 [system] INFO: ARM_WLAN_WD 2
0000004995 [system] INFO: CLR_WLAN_WD 1, DHCP success
Google IP Address: 172.217.23.46. This is a true bool. Cell reception is -111 / 19.
Google IP Address: 172.217.23.46. This is a true bool. Cell reception is -111 / 19.
Google IP Address: 172.217.23.46. This is a true bool. Cell reception is -111 / 19.
[...]

PR #1375, opened 2017-08-25
Fixes incorrect handling of inet_gethostbyname() return values in Cellular.resolve() and WiFi.resolve()

avtolstoy:
### Problem

See issue #1304

`Cellular.resolve()` and `WiFi.resolve()` functions that internally use `inet_gethostbyname()` incorrectly handle `inet_gethostbyname()` return values.

### Solution

- Treat any `inet_gethostbyname()` return value other than `0` as an error.
- Fix tests to check all octets of the IP address, instead of just the first one.

### Steps to Test

- `wiring/no_fixture`
- `wiring/wifi`

### References

- Closes issue #1304
- CH7267

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_wifi.h
+++ b/wiring/inc/spark_wiring_wifi.h
@@ -200,7 +200,7 @@ public:
 
     IPAddress resolve(const char* name)
     {
-        HAL_IPAddress ip;
+        HAL_IPAddress ip = {0};
         return (inet_gethostbyname(name, strlen(name), &ip, *this, NULL) != 0) ?
                 IPAddress(uint32_t(0)) : IPAddress(ip);
     }
--- a/wiring/inc/spark_wiring_wifi.h
+++ b/wiring/inc/spark_wiring_wifi.h
@@ -201,7 +201,7 @@ public:
     IPAddress resolve(const char* name)
     {
         HAL_IPAddress ip;
-        return (inet_gethostbyname(name, strlen(name), &ip, *this, NULL)<0) ?
+        return (inet_gethostbyname(name, strlen(name), &ip, *this, NULL) != 0) ?
                 IPAddress(uint32_t(0)) : IPAddress(ip);
     }
 
--- a/user/tests/wiring/no_fixture/cellular.cpp
+++ b/user/tests/wiring/no_fixture/cellular.cpp
@@ -113,6 +113,12 @@ test(CELLULAR_05_resolve) {
     checkIPAddress("www.google.com", Cellular.resolve("www.google.com"));
 }
 
+test(CELLULAR_06_resolve) {
+    connect_to_cloud(6*60*1000);
+    IPAddress addr = Cellular.resolve("this.is.not.a.real.host");
+    assertEqual(addr, 0);
+}
+
 int how_many_band_options_are_available(void)
 {
     CellularBand band_avail;
--- a/user/tests/wiring/no_fixture/wifi.cpp
+++ b/user/tests/wiring/no_fixture/wifi.cpp
@@ -25,7 +25,7 @@
 test(WIFI_01_resolve_3_levels)
 {
     IPAddress address = WiFi.resolve("pool.ntp.org");
-    assertNotEqual(address[0], 0);
+    assertNotEqual(address, 0);
 
     // ensure the version field is set
     assertNotEqual(address.version(), 0);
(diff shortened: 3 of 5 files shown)


wiring/inc/spark_wiring_wifi.h
```diff
@@ -201,7 +201,7 @@ class WiFiClass : public NetworkClass
     IPAddress resolve(const char* name)
     {
         HAL_IPAddress ip;
-        return (inet_gethostbyname(name, strlen(name), &ip, *this, NULL)<0) ?
+        return (inet_gethostbyname(name, strlen(name), &ip, *this, NULL) != 0) ?
```
m-mcgowan: can you please explain this? I thought the default comparison is not equal to 0?

avtolstoy: Electron's `inet_gethostbyname` returns a positive value in case of an error, which fails the `<0` check. It sounds also reasonable to make sure that the return value is negative in all the HALs, but the check for `!= 0` sounds good as well, as that mimics the behavior of `gethostbyname_r`.

m-mcgowan: oh, I didn't see the <0 when I originally looked (text wrapping.)  If we change it to != 0 then we should check no other HALs are using >0 codes for success. (I have a feeling the CC3000 does this...) so it might be better to make the Electron HAL conform to the established <0 convention.

avtolstoy: I don't think that's the case on the Core either: see https://github.com/spark/firmware/blob/develop/hal/src/core/inet_hal.c#L39

technobly:
- [ ] Perhaps we should fix this in `inet_gethostbyname` directly.  If you see the user's logs when resolve returns an obvious failure there is still a non-zero IP address.  

- [ ] If it fails, perhaps we should also ensure the IP address is 0.0.0.0 as well as returning a proper error code (for users that don't bother to add error checking).

- [ ] Unclear why we (and by we I mean me) went with a positive return value, looks like an error.  If we made cellular return -1 when an error occurs we'd have to check the places inet_gethostbyname is called in system firmware for cellular and possibly flip the logic.  

- [ ] We should double check that either way.

## cellular
```c++
int inet_gethostbyname(const char* hostname, uint16_t hostnameLen, HAL_IPAddress* out_ip_addr,
		network_interface_t nif, void* reserved)
{
	uint32_t result = electronMDM.gethostbyname(hostname);
	if (result > 0) {
		out_ip_addr->ipv4 = result;
		return 0;
	}
    return 1;
}
```

## wifi 
```c++
int inet_gethostbyname(const char* hostname, uint16_t hostnameLen, HAL_IPAddress* out_ip_addr, network_interface_t nif, void* reserved)
{
    wiced_ip_address_t address;
    wiced_result_t result = wiced_hostname_lookup (hostname, &address, 5000);
    if (result == WICED_SUCCESS) {
    		HAL_IPV4_SET(out_ip_addr, GET_IPV4_ADDRESS(address));
    }
    return -result;
}
```

avtolstoy:
How about we do both then? :) Core also returns positive error:
```c++
int inet_gethostbyname(const char* hostname, uint16_t hostnameLen, HAL_IPAddress* out_ip_addr,
    network_interface_t nif, void* reserved)
{
    int attempts = 5;
    out_ip_addr->ipv4 = 0;
    int result = 0;
    while (!out_ip_addr->ipv4 && attempts --> 0) {
        HAL_Delay_Milliseconds(1);
        result = gethostbyname(hostname, hostnameLen, &out_ip_addr->ipv4)<0;
    }
    return result;
}
```

`!= 0` check will also ensure that we return `0.0.0.0` IP address in case of any non-zero return value from `inet_gethostbyname`

technobly:
where are you referring to for the `!=0` check?  I'm not seeing how that would force the ip to be 0.0.0.0.

`electronMDM.gethostbyname()` will return 0 for an error, so in that case we can set the ip object to 0.0.0.0 (or 0), but perhaps it's better to initialize the `HAL_IPAddress ip;` object to zero in Cellular/Wifi.resolve().

For cellular, we should be setting the ip with `HAL_IPV4_SET(out_ip_addr, result);` as well.

m-mcgowan:
Cherry-picked to 0.7.0-rc.7 branch, and that merged to develop for 0.8.0-rc.2 release. (I probably should have rebased instead so Github maintained the tracking, but that's hindsight.)

PR #1390, opened 2017-09-18
Diagnostics service

sergeuz:
### Problem

This PR implements a simple service allowing the system and applications to register diagnostic data sources.

### Steps to Test

Run unit tests.

### Example App

```cpp
#include "application.h"

enum DiagnosticId {
    UPTIME = DIAG_ID_USER,
    COUNTER,
    STATE
};

enum State {
    CONNECTING = 1,
    CONNECTED = 2,
    DISCONNECTED = 3
};

class UptimeDiagnosticData: public AbstractIntegerDiagnosticData {
public:
    UptimeDiagnosticData() :
            AbstractIntegerDiagnosticData(UPTIME, "app.uptime") {
    }

protected:
    virtual int get(int32_t& val) override {
        val = millis();
        return 0; // OK
    }
};

UptimeDiagnosticData uptime;
AtomicIntegerDiagnosticData counter(COUNTER, "app.counter");
AtomicEnumDiagnosticData<State> state(STATE, "app.state", DISCONNECTED);

void setup() {
    ++counter;
    state = CONNECTING;
}

void loop() {
}
```

### References

* [ch7575]

---

### Completeness

- [ ] Added documentation


modules/photon/system-part1/src/newlib.cpp
```diff
@@ -1 +1,4 @@
 #include "../../../../hal/src/stm32/newlib.cpp"
+
+// Defining this in hal/src/stm32/newlib.cpp causes linker errors
+void *__dso_handle = NULL;
```
m-mcgowan: Do you know what has changed to make this necessary?

services/inc/diagnostics.h
```diff
@@ -0,0 +1,72 @@
...
+    DIAG_CMD_GET = 4
+} diag_cmd;
+
+typedef struct diag_source diag_source;
+
+typedef int(*diag_source_cmd_callback)(const diag_source* src, int cmd, void* data);
+typedef void(*diag_enum_sources_callback)(const diag_source* src, void* data);
+
+typedef struct diag_source {
+    size_t size; // Size of this structure
```
m-mcgowan: should we make `size` a 16 bit quantity so it packs nicely into memory better?

modules/photon/system-part1/src/newlib.cpp
sergeuz: `part1` now has a global variable (`Diagnostics`) with a non-trivial destructor

services/inc/diagnostics.h
sergeuz: 👍

services/src/diagnostics.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <algorithm>
+
+namespace {
+
+using namespace spark;
+
+class Diagnostics {
+public:
+    Diagnostics() :
+            enabled_(0) { // The service is disabled initially
```
m-mcgowan: what's the rationale for having enable/disable for the service? If it's to save expensive computation then perhaps we should move the enable flag to the diag_source?

wiring/inc/spark_wiring_diagnostics.h
```diff
@@ -0,0 +1,367 @@
...
+    IntType operator--();
+    IntType operator--(int);
+    IntType operator+=(IntType val);
+    IntType operator-=(IntType val);
+
+    IntegerDiagnosticData<AtomicLockingPolicy>& operator=(IntType val);
+    operator IntType() const;
+
+private:
+    std::atomic<IntType> val_;
```
m-mcgowan: What's the overhead of using std:atomic here with SimpleIntegerDiagnosticData?  The majority of cases the diagnostic data doesn't need to be thread safe, only a few cases need it, so the non-atomic version should be the more efficient case ideally.

wiring/inc/spark_wiring_diagnostics.h
```diff
@@ -0,0 +1,367 @@
...
+
+template<typename LockingPolicyT>
+inline particle::IntegerDiagnosticData<LockingPolicyT>::IntegerDiagnosticData(uint16_t id, const char* name, IntType val) :
+        AbstractIntegerDiagnosticData(id, name),
+        val_(val) {
+}
+
+template<typename LockingPolicyT>
+inline particle::AbstractIntegerDiagnosticData::IntType particle::IntegerDiagnosticData<LockingPolicyT>::operator++() {
+    LockingPolicyT::lock();
```
m-mcgowan: RAII opportunity here?

services/src/diagnostics.cpp
sergeuz: The rationale is to avoid a race condition, which is possible when the system receives a request or a call querying diagnostic data but data sources are still being registered. The system explicitly enables the diagnostic service after invoking all global constructors in all system parts [here](https://github.com/spark/firmware/blob/feature/diagnostics/system/src/main.cpp#L657)

wiring/inc/spark_wiring_diagnostics.h
sergeuz: `SimpleIntegerDiagnosticData` uses a dummy [`NoLockingPolicy`](https://github.com/spark/firmware/blob/feature/diagnostics/wiring/inc/spark_wiring_thread.h#L236), so there shouldn't any overhead. The class in question is a specialization of `IntegerDiagnosticData` for `AtomicLockingPolicy`, i.e. it's up to a developer to decide if a particular diagnostic data needs to be updated atomically.

services/src/diagnostics.cpp
m-mcgowan: Thanks, that makes sense. The name `enabled` is slightly counter-intuitive and that you cannot add sources once it is enabled. On first reading, I was sure I understood it as the opposite - the error is raised when adding sources and the service is not enabled. 

Perhaps we should call it running, or started?  Alternatively, you could add a comment to explain, since normally things are not usable while disabled, while here it's the opposite.

sergeuz: 👍

wiring/inc/spark_wiring_diagnostics.h
sergeuz: 👍

PR #1396, opened 2017-09-26
Simple pool allocator

avtolstoy:
<details>

```

```
</details>

### Problem

Implements a simple pooled memory allocator to simplify communications between interrupt handlers and system threads.

Currently, passing data from an ISR to the system thread is complicated and requires implementing feature-specific object pools, which are not always helpful (for example, when it is necessary to pass some data to the system thread from an ISR and forget about it) and have code size implications.

Having a small general purpose memory pool supporting typical malloc/free operations in the context of an ISR would be helpful.

### Steps to Test

Unit tests: `user/tests/unit/pool_allocator.cpp`

### Example App

N/A

### References

- [CH7595]

---

### Completeness

- [ ] Added documentation


system/inc/simple_pool_allocator.h
```diff
@@ -0,0 +1,159 @@
...
+    uint8_t* ptr_;
+
+    BlockHeader* freeList_ = nullptr;
+    BlockHeader* freeListEnd_ = nullptr;
+};
+
+class SimpleAllocedPool : public SimpleBasePool {
+public:
+    SimpleAllocedPool(size_t size) :
+        SimpleBasePool(reinterpret_cast<void*>(new uint8_t[size]), size) {
```
sergeuz: I think `SimpleAllocedPool` (and its base class) should be tolerant to memory allocation errors, given that it can be requested to allocate a relatively large buffer and that the exact size of the heap depends on a user application.

avtolstoy: 👍

PR #1397, opened 2017-09-26
"net.rssi" diagnostic data

sergeuz:
### Problem

The system should expose RSSI diagnostic info on platforms that have a wireless network module.

### Solution

Add `net.rssi` diagnostic data source.

### Example App

```cpp
#include "application.h"

namespace {

    { "app", LOG_LEVEL_ALL }
});

bool logDiagData(void* appender, const uint8_t* data, size_t size) {
    Log.write(LOG_LEVEL_INFO, (const char*)data, size);
    return true;
}

} // namespace

void setup() {
    uint16_t id = DIAG_ID_NETWORK_RSSI;
    system_format_diag_data(&id, 1, 0, logDiagData, nullptr, nullptr);
    Log.print("\r\n");
}

void loop() {
}
```

### References

* [ch7642]
* #1390 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_network.cpp
+++ b/system/src/system_network.cpp
@@ -18,6 +18,8 @@
  */
 
 #include "spark_wiring_ticks.h"
+#include "spark_wiring_diagnostics.h"
+#include "spark_wiring_fixed_point.h"
 #include "system_setup.h"
 #include "system_network.h"
 #include "system_network_internal.h"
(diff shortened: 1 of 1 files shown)


sergeuz:
The registration happens in the constructor of `AbstractDiagnosticData`: https://github.com/spark/firmware/blob/feature/diagnostics/wiring/inc/spark_wiring_diagnostics.h#L170

PR #1401, opened 2017-10-04
Move global declarations to a namespace

### LINKED ISSUE #1267: Move global defines to a namespace.
Hello!

After the last release, `Promise` has now been declared globally. The project I'm currently working on has it's own `Promise` implementation which, of course, causes the entire thing to break.

My options are:

- Rename my own `Promise` implementation and every reference to it to something less descriptive
- Compile using an earlier release, which will not be sufficient long term
- Post an issue here to stop new releases from clobbering defines in existing projects

My feature request is to add everything defined globally into it's own namespace and have `.ino` files `using namespace spark` by default. As it stands now, to my knowledge, there is no way to override the globals defined in the firmware so, in cases like these where there is a new naming conflict, the end user is forced to modify and obfuscate a large portion of their code for it to build. By adding these globals to a namespace, the user can choose when and when not to use spark defines. 

Of course this may be a breaking change to many projects and libraries already existing, but without this, every new feature added to the firmware is potentially a breaking change.

I'm currently using:

Firmware version: `0.6.1`
Particle: `photon`

sergeuz:
### Problem

Particle APIs get pulled into the global namespace and cause conflicts with an application code. This behavior cannot be overridden by application developers.

### Solution

Add `PARTICLE_NO_GLOBAL_NAMESPACE` macro that can be defined by applications to avoid having Particle APIs declared in the global namespace.

**Note:** This PR doesn't attempt to move all current APIs to the `particle` namespace, and some classes and functions are still declared in the global namespace directly. I suggest we start moving such declarations to namespaces iteratively, on occasion.

### Steps to Test

`wiring/api` test should compile successfully.

### Example App

```cpp
#define PARTICLE_NO_GLOBAL_NAMESPACE

#include "application.h"

// I just need my own WiFi class
class WiFi {
};

void test(WiFi&) {
}
```

### References

* fixes #1267
* [ch8449]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/inc/active_object.h
+++ b/system/inc/active_object.h
@@ -31,6 +31,8 @@
 #include "channel.h"
 #include "concurrent_hal.h"
 
+namespace particle {
+
 /**
  * Configuratino data for an active object.
  */
@@ -436,9 +438,11 @@ public:
 
 };
 
+#else
 
+namespace particle {
 
-#endif // PLATFORM_THREADING
+#endif // !PLATFORM_THREADING
 
 /**
  * This class implements a queue of asynchronous calls that can be scheduled from an ISR and then
@@ -466,3 +470,5 @@ private:
     Task* firstTask_; // Task queue
     Task* lastTask_;
 };
+
+} // namespace particle
--- a/system/inc/system_threading.h
+++ b/system/inc/system_threading.h
@@ -20,7 +20,8 @@
 #define	SYSTEM_THREADING_H
 
 #include "active_object.h"
-extern ISRTaskQueue SystemISRTaskQueue;
+
+extern particle::ISRTaskQueue SystemISRTaskQueue;
 
 #if PLATFORM_THREADING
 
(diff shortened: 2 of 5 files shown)


monkbroc:
That's a good idea. :+1: 

Due to how the wiring preprocessor works, it needs to be documented that both the new define and the Particle include need to be added. For a refresher on how the preprocessor works, see [the preprocessor description in the docs](https://docs.particle.io/reference/firmware/photon/#preprocessor).

For example this wouldn't work in the Web IDE:
```
#define PARTICLE_NO_GLOBAL_NAMESPACE
class WiFi {};
```

This would work:
```
#define PARTICLE_NO_GLOBAL_NAMESPACE
#include "Particle.h"
class WiFi {};
```

We also document using `#include "Particle.h"` not `#include "application.h"`. I know they are equivalent but that's the documentation choice right now.

sergeuz:
That's a good point, Julien. Good to know that the Web IDE is smart enough to not pre-include any header files unconditionally.

m-mcgowan:
It would be nice to see tests affirming that global symbols such as `Message` and `Promise` are no longer taken by default.

m-mcgowan:
Is this a safe change? is it something we should queue up for the 2.x release in case there are breaking changes?  I would like to see the API test suite expanded to cover the entire API and become our definitive reference of what is in the global and particle namespaces.

monkbroc:
Let's document the macro in https://docs.particle.io/reference/device-os/firmware/photon/#preprocessor

PR #1402, opened 2017-10-05
Add the String::split Wiring API

monkbroc:
### Problem

As a firmware developer I often need to split a string on a delimiter like when I receive an event I subscribed to. Splitting a String is not documented in our firmware docs and using the C function `strtok` is difficult because it modifies the actual string.

### Solution

Add an easy to use Wiring API `String::split(delimiter)`

### Steps to Test

Compilation API and unit tests included

Run `firmware/user/tests/unit$ make`

Note that I had to fix `services/inc/dcd.h` to make the unit tests compile. I didn't include it in this PR because it was unrelated and it's easy to do in a separate PR.
```
		bool isValid() const {
			return watermark==WATERMARK;
		}
```

I also had to comment out the last 2 unit tests from `user/tests/unit/string_interpolate.cpp` because they were segfaulting the unit tests on my machine before I added my new code.

### Example App

```c
#include "Particle.h"

void myHandler(const char *event, const char *data) {
  String str(data);
  auto parts = str.split("=");
  String name = parts.next();
  if (!parts.hasNext()) {
    Serial.println("Couldn't find =");
  }
  String value = parts.next();
  Serial.printlnf("name=%s value=%s", name, value);
}

void setup() {
  Particle.subscribe("testsplit", myHandler, MY_DEVICES);
  Particle.publish("testsplit", "D7=HIGH", PRIVATE);
}
```

### References

---

### Completeness

- [ ] Added documentation


sergeuz:
Given that it's anyway tricky to split a string without copying, how about using [`particle::Vector`](https://github.com/spark/firmware/blob/develop/wiring/inc/spark_wiring_vector.h#L50)?

```cpp
#include "Particle.h"

void setup() {
    Serial.begin();
    String data("D7=HIGH");
    auto parts = data.split("="); // Returns Vector<String>
    if (parts.size() < 2) {
        Serial.println("Couldn't find =");
        return;
    }
    Serial.printlnf("name=" + parts[0] + " value=" + parts[1]);
}
```

monkbroc:
I like that idea.

technobly:
👍 +1 for vectors. This would have been super helpful just recently :)

PR #1408, opened 2017-10-20
Persistent diagnostic data

sergeuz:
### Problem

This PR implements diagnostic data sources with persistence to backup RAM.

### Steps to Test

- Run unit tests.
- Run `wiring/persistence` test.

### Example App

**Note:** GCC's `section` attribute (and its `retained` and `retained_system` wrappers) can be applied only to global variables, so the implementation requires a separate storage variable to be declared for each persistent diagnostic source (_TODO: Add a convenience macro declaring both a diagnostic source and a storage for its data_).

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC)

enum Id {
    ID_1 = DIAG_ID_USER,
    ID_2
};

enum Enum {
    ONE = 1,
    TWO = 2
};

retained RetainedIntegerDiagnosticDataStorage<ID_1> storage1;
PersistentIntegerDiagnosticData<decltype(storage1)> data1(&storage1, ID_1, 1);

retained RetainedEnumDiagnosticDataStorage<Enum, ID_2> storage2;
PersistentEnumDiagnosticData<Enum, decltype(storage2)> data2(&storage2, ID_2, ONE);

    { "app", LOG_LEVEL_ALL }
});

void setup() {
    Log.info("data1: %d; data2: %d", (int)data1, (int)data2);
    data1 = 2;
    data2 = TWO;
}

void loop() {
}
```

### References

- [ch8408]

---

### Completeness

- [ ] Added documentation


services/inc/util.h
```diff
@@ -0,0 +1,43 @@
...
+    typedef T Type;
+};
+
+template<typename T>
+struct UnderlyingType<T, typename std::enable_if<std::is_enum<T>::value>::type> {
+    typedef typename std::underlying_type<T>::type Type;
+};
+
+template<typename T>
+inline void combineHash(size_t& seed, const T& value) {
```
m-mcgowan: These look like good additions, but I am always a bit uneasy about 'util" packages/headers etc.  Naming is hard! but let's try to put these  in more tightly-named headers. E.g. `combine_hash.h` for hashing and `underlying_type_enum.h` for the underlying type tweaks.

wiring/inc/spark_wiring_diagnostics.h
```diff
@@ -57,120 +59,422 @@ class AbstractIntegerDiagnosticData: public AbstractDiagnosticData {
...
+
+    operator EnumT() const {
+        const auto lock = ConcurrencyT::lock();
+        const EnumT v = (EnumT)storage_->value();
+        ConcurrencyT::unlock(lock);
+        return v;
+    }
+
+private:
+    StorageT* storage_;
```
m-mcgowan: Would it be safer to use a reference rather than a pointer here?

wiring/inc/spark_wiring_diagnostics.h
```diff
@@ -57,120 +59,422 @@ class AbstractIntegerDiagnosticData: public AbstractDiagnosticData {
...
+
+    virtual int get(IntType& val) override { // AbstractIntegerDiagnosticData
+        const auto lock = ConcurrencyT::lock();
+        val = (IntType)storage_->value();
+        ConcurrencyT::unlock(lock);
+        return SYSTEM_ERROR_NONE;
+    }
+};
+
+template<typename ValueT, DiagnosticDataId id>
```
m-mcgowan: I wonder how much flash space is taken each time this template is instantiated with a different ID.  Would making the ID member field be worth it?  This could totally be premature optimization that we can do later.

sergeuz: yeah, I had the same concern but decided in favour of the lower backup RAM usage. At second thought, the overhead of storing an ID in the backup RAM is pretty minimal and as a bonus we're getting a slightly more reliable integrity check

wiring/inc/spark_wiring_diagnostics.h
sergeuz: 👍

services/inc/util.h
sergeuz: 👍

wiring/inc/spark_wiring_diagnostics.h
m-mcgowan: I don't mean that the ID would be stored in backup ram, but that it is stored in SRAM, so we don't need a new template instantiation for each ID.

sergeuz:
Updated the API according to the feedback and added a couple of macros to simplify declaration of the persistent diagnostic data sources:
```cpp
enum Id {
    INT_1 = DIAG_ID_USER,
    ENUM_1
};

enum Enum {
    ONE = 1
};

PARTICLE_RETAINED_INTEGER_DIAGNOSTIC_DATA(
    int1, /* Variable name */
    INT_1, /* Data source ID */
    "int1", /* Data source name */
    1 /* Initial value */,
    AtomicConcurrency /* Concurrency mode (optional) */
);

PARTICLE_RETAINED_ENUM_DIAGNOSTIC_DATA(enum1, ENUM_1, "enum1", ONE); // NoConcurrency
```

PR #1413, opened 2017-10-24
Fix caching of the description CRCs in the backup RAM

### LINKED ISSUE #1172: [Electron] Application State CRC always mismatched when building with DEBUG_BUILD=y
Flashing the electron with the same app over and over again, I observe different application checksums, but they remain consistent between DFU attempts.

    TRACE: application state checksum: 8458b248, expected: 2a92a22f 
    TRACE: application state checksum: 8458b248, expected: 2a92a22f

A hard reset yeilds the same thing:

    TRACE: application state checksum: 8458b248, expected: 2a92a22f 

If I change the app slightly (log message) I see:

    TRACE: application state checksum: 8458b248, expected: e1f6f230

If I change the app slightly (add another function key for one of the same functions) I see:

    TRACE: application state checksum: 8458b248, expected: 62fde692

- `DEBUG("application state checksum: %x, expected: %x", actual, app_state_crc);` Note: actual on left, and app_state_crcc on right of logs message.  This seems to contradict the results above when the actual app flashed to the device is changing (notice the left crc value is constant).  Perhaps the logs need to be flipped?  Expected should be what is stored in backup SRAM, and if it is... then the results above are odd.

- This only seems to be the case when `DEBUG_BUILD=y` during compile time.  If `DEBUG_BUILD` is not defined, application checksums appear to match after first reset (via button, DFU same app, etc..)

- This is important because the electron system firmware is compiled by default with `DEBUG_BUILD=y`, and when the application checksum is the same (as is the case when sleeping and waking up), the handshake process can be VERY short which means little data used.  When the application checksum is different, unnecessary data usage is occuring for each re-connection to the cloud.

---

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (v0.6.0)

sergeuz:
### Problem

This PR fixes #1172 by ensuring that the application/system description checksums stored in the backup RAM don't get overwritten by invalid values when DTLS context is updated.

### Steps to Test

1. Flash a blank application with enabled logging to an Electron.
2. Toggle the power off and on to ensure that the persistent session data is invalidated.
3. Check that the device sends the HELLO message.
4. Flash the same application binary to the device OTA.
5. Check that the device skips sending the HELLO message.

### References

- [CH8911]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/communication/src/dtls_message_channel.cpp
+++ b/communication/src/dtls_message_channel.cpp
@@ -348,9 +348,9 @@ ProtocolError DTLSMessageChannel::establish(uint32_t& flags, uint32_t app_state_
 				sessionPersist.out_ctr[4],sessionPersist.out_ctr[5],sessionPersist.out_ctr[6],
 				sessionPersist.out_ctr[7], sessionPersist.next_coap_id);
 		sessionPersist.make_persistent();
-		uint32_t actual = sessionPersist.application_state_checksum(this->callbacks.calculate_crc);
-		LOG(INFO,"app state crc: %x, expected: %x", actual, app_state_crc);
-		if (actual==app_state_crc) {
+		const uint32_t cached = sessionPersist.application_state_checksum(this->callbacks.calculate_crc);
+		LOG(INFO,"app state crc: cached: %x, actual: %x", (unsigned)cached, (unsigned)app_state_crc);
+		if (cached==app_state_crc) {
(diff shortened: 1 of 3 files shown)


sergeuz:
> The original report mentioned this only happened with debug builds - were you able to confirm that and find out why?

In my tests the issue was reproducible regardless of `DEBUG_BUILD`.

PR #1415, opened 2017-10-26
Guard cellular_command() with a global lock

### LINKED ISSUE #1298: Publishing in a multi-thread application running on UDP device during OTA results in hard fault
```
Given I have a UDP device running application firmware with multithreading enabled
And the application firmware publishes events to the cloud at regular intervals
When I OTA flash the device with another application firmware
And the device publishes an event while the OTA is in progress
Then the OTA will be irreparably disrupted and the UDP device will hard fault until a manual reset
```

Application firmware running on the Electron was:

```
// This #include statement was automatically added by the Particle IDE.
#include <google-maps-device-locator.h>
GoogleMapsDeviceLocator locator;

SerialLogHandler logHandler;
SYSTEM_THREAD(ENABLED);

void setup() {
    Serial.begin(9600);
    locator.withLocatePeriodic(10);
}


void loop() {
    locator.loop();
}
```

_when_ this firmware is running on the device, _and_ I OTA a new application firmware to the electron, the OTA fails mid-way through and my electron starts hard faulting. Serial logs for the device during the OTA:

```
0000218892 [comm.protocol] INFO: rcv'd message type=13
0000219614 [comm.protocol] INFO: rcv'd message type=13
0000226230 [comm.protocol] INFO: rcv'd message type=5
0000226484 [comm.protocol] INFO: rcv'd message type=7
0000226738 [comm.protocol] INFO: rcv'd message type=7
0000226992 [comm.protocol] INFO: rcv'd message type=7
0000227246 [comm.protocol] INFO: rcv'd message type=7
0000227500 [comm.protocol] INFO: rcv'd message type=7
publishLocation
0000228099 [comm] WARN: handle received message failed - aborting transfer
0000228100 [system] INFO: Send spark/device/ota_result event
0000229100 [comm] WARN: handle received message failed - aborting transfer
0000229100 [system] INFO: Send spark/device/ota_result event
0000230101 [comm] WARN: handle received message failed - aborting transfer
0000230101 [system] INFO: Send spark/device/ota_result event
0000231102 [comm] WARN: handle received message failed - aborting transfer
```

So looks like to me the device is receiving UDP packets, but for some reason, the application code on the device is still running while receiving the OTA. The `publishLocation` is logged every time network data is sent up to the cloud. Once this happened, the following system messages suggest an issue, and the OTA is aborted. Then the device starts hard-faulting (blinking red).

Note that this is only observable on an Electron (UDP) not Photon (TCP). Running the same experiment on a Photon produces the following device logs:

```
0000055938 [comm.sparkprotocol] INFO: chunk idx=22 crc=1 fast=1 updating=1
publishLocation
0000056043 [comm.sparkprotocol] INFO: chunk
0000056043 [comm.sparkprotocol] INFO: chunk idx=23 crc=1 fast=1 updating=1
0000056146 [comm.sparkprotocol] INFO: chunk
0000056146 [comm.sparkprotocol] INFO: chunk idx=24 crc=1 fast=1 updating=1
0000056249 [comm.sparkprotocol] INFO: chunk
0000056249 [comm.sparkprotocol] INFO: chunk idx=25 crc=1 fast=1 updating=1
0000056352 [comm.sparkprotocol] INFO: chunk
0000056352 [comm.sparkprotocol] INFO: chunk idx=26 crc=1 fast=1 updating=1
0000056455 [comm.sparkprotocol] INFO: chunk
```

sergeuz:
### Problem

This PR fixes #1298 by guarding the implementation of `cellular_command()` with the Cellular HAL's global lock.

### Steps to Test

Flash the following application to an Electron, and ensure that repeated attempts to update the device OTA don't cause communication errors and hard faults.

```cpp
#include "application.h"

SYSTEM_THREAD(ENABLED)

namespace {

const Serial1LogHandler logHandler(115200, LOG_LEVEL_INFO, {
    { "app", LOG_LEVEL_ALL }
});

const unsigned INTERVAL = 500;
unsigned lastMillis = 0;

void onFirmwareUpdate() {
    lastMillis = millis();
}

} // namespace

void setup() {
    System.on(firmware_update_pending, onFirmwareUpdate);
}

void loop() {
    if (lastMillis != 0) {
        if (!Particle.connected()) {
            Log.info("disconnected");
            lastMillis = 0;
        } else if (millis() - lastMillis >= INTERVAL) {
            Log.info("sending command");
            Cellular.command("AT\r\n");
            lastMillis = millis();
        }
    }
}
```

### References

- [CH8908]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -376,6 +376,15 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
     return WAIT;
 }
 
+int MDMParser::command(const char* cmd, _CALLBACKPTR cb, void* param, system_tick_t timeout)
+{
+    LOCK();
+    send(cmd, strlen(cmd));
+    const int ret = waitFinalResp(cb, param, timeout);
+    UNLOCK();
+    return ret;
+}
+
 int MDMParser::_cbString(int type, const char* buf, int len, char* str)
 {
     if (str && (type == TYPE_UNKNOWN)) {
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -405,6 +405,16 @@ int MDMParser::sendCommandWithArgs(const char* fmt, va_list args, _CALLBACKPTR c
     return ret;
 }
 
+void MDMParser::lock()
+{
+    mdm_mutex.lock();
+}
+
+void MDMParser::unlock()
+{
+    mdm_mutex.unlock();
+}
+
 int MDMParser::_cbString(int type, const char* buf, int len, char* str)
 {
     if (str && (type == TYPE_UNKNOWN)) {
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -214,14 +214,34 @@ int MDMParser::send(const char* buf, int len)
 }
 
 int MDMParser::sendFormated(const char* format, ...) {
-    if (_cancel_all_operations) return 0;
-
-    char buf[MAX_SIZE];
     va_list args;
     va_start(args, format);
-    int len = vsnprintf(buf,sizeof(buf), format, args);
+    const int ret = sendFormattedWithArgs(format, args);
     va_end(args);
-    return send(buf, len);
+    return ret;
+}
+
+int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
+    if (_cancel_all_operations) {
+        return 0;
+    }
+    va_list argsCopy;
+    va_copy(argsCopy, args);
+    char buf[128];
+    int n = vsnprintf(buf, sizeof(buf), format, args);
+    if (n >= 0) {
+        if ((size_t)n < sizeof(buf)) {
+            n = send(buf, n);
+        } else {
+            char buf[n + 1]; // Use larger buffer
+            n = vsnprintf(buf, sizeof(buf), format, argsCopy);
+            if (n >= 0) {
+                n = send(buf, n);
+            }
+        }
+    }
+    va_end(argsCopy);
+    return n;
 }
 
 int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
@@ -376,10 +396,10 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
     return WAIT;
 }
 
-int MDMParser::command(const char* cmd, _CALLBACKPTR cb, void* param, system_tick_t timeout)
+int MDMParser::sendCommandWithArgs(const char* fmt, va_list args, _CALLBACKPTR cb, void* param, system_tick_t timeout)
 {
     LOCK();
-    send(cmd, strlen(cmd));
+    sendFormattedWithArgs(fmt, args);
     const int ret = waitFinalResp(cb, param, timeout);
     UNLOCK();
     return ret;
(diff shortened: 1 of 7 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -375,6 +375,15 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
     return WAIT;
 }
 
+int MDMParser::command(const char* cmd, _CALLBACKPTR cb, void* param, system_tick_t timeout)
+{
+    LOCK();
+    send(cmd, strlen(cmd));
```
technobly: This should use `MDMParser::sendFormated` to ensure it is cancellable with `if (_cancel_all_operations) return 0;`

sergeuz: hm, but why is there a semantic difference between `send()` and `sendFormatted()` in regards to the `_cancel_all_operations` flag? Is it an oversight or a designed behavior?

technobly: all send operations in mdm_hal start with sendFormatted(), so that's where I targeted the cancel operation, instead of down the line in the lower level send() function.

sergeuz: ok, so every logical modem operation starts with a call to `sendFormatted()` and `send()` is mostly used to send remaining data, such as the contents of a datagram buffer. If so, could we reduce the size of [the buffer](https://github.com/spark/firmware/blob/fix/issue_1298/hal/src/electron/modem/mdm_hal.cpp#L218) used by `sendFormatted()`? Currently it allocates 1K on stack just to format an AT command and I couldn't find any place in the code where such a large buffer might be needed. Most of the commands sent via `sendFormatted()` don't even use any formatting at all and are as short as e.g. `sendFormated("AT+CGSN\r\n")`.

sergeuz: I suggest we use a similar approach as in `Print::printf()`, i.e. allocate a small buffer initially and then use a variable length array if it appears the formatted data doesn't fit into the original buffer: https://github.com/spark/firmware/blob/develop/wiring/src/spark_wiring_print.cpp#L271

technobly: That method looks fine, and we have another place in cellular_command with `char buf[256];` that could use a similar treatment.

sergeuz: I'll add an overloaded `sendFormatted()` taking `va_list` as an argument in order to not duplicate the code.

technobly:
Just remembered we have this scheme as well, but I do prefer the global lock so you don't have to know if a peripheral is thread safe, or is a shared resource and think about trying to acquire a lock before using it.  https://prerelease-docs.particle.io/reference/firmware/electron/#synchronizing-access-to-shared-system-resources

@m-mcgowan @sergeuz I'm needing this change to be merged asap :)

towynlin:
👍  Let's get this released ASAP.

sergeuz:
Good point, @technobly. It would be nice to have a scoped locking for `Cellular` as well, because for certain commands the modem can reply with a prompt asking the caller to provide a continuation of a command. In this case the atomicity provided by `Cellular` at the method level is not enough. However, I think that this PR can still be merged as is, since it addresses a somewhat separate issue, where sending of a command and receiving of a response are not done atomically even for a single and complete command.

technobly:
Good point as well @sergeuz.  The specific case I need involves using `Cellular.command` for one command that does not involve a prompt.  So I believe this PR will fix that situation as it currently stands.  I want to make sure that when the system makes a call for something like a `MDMParser::socketSendTo` that requires a prompt, that it will not allow `Cellular.command()` to interrupt it.  I think it should, because those calls are wrapped with a `LOCK()`.  But if I wanted to do the same thing with `Cellular.command()` it would require something like the `WITH_LOCK()` API.  I can see this becoming a future requirement quickly, so please prioritize that feature.

sergeuz:
@technobly Added support for `WITH_LOCK(Cellular)` and rebased this branch on `develop`.

PR #1434, opened 2017-11-16
Remove dependency check in bootloader (0.8.x)

sergeuz:
### Problem

Starting with 0.8.0-rc.1 firmware, part1 on Photon has a dependency on part2 of a specific minimum version. The bootloader doesn't expect that part1 can have any dependencies and fails to import DCT functions, making it impossible to access the DCT when the device is in DFU mode.

### Solution

Remove the failing dependency check from the bootloader on Photon. The fix requires making an intermediate release (0.7.0-rc.5).

### Steps to Test

Prepare 3 sets of the firmware binaries:

- 0.7.0-rc.4 (`release/v0.7.0-rc.4` branch).
- 0.7.0-rc.5 (`fix/bootloader_dep_check_0.7.x` branch). Make sure to set [the module versions](https://github.com/spark/firmware/blob/fix/bootloader_dep_check_0.7.x/modules/shared/system_module_version.mk#L3) to 204.
- 0.8.0-rc.1 (`fix/bootloader_dep_check_0.8.x` branch).

Test steps:

1. Flash 0.7.0-rc.4 binaries (including bootloader) to a Photon device.
2. Try to update the firmware to 0.8.0-rc.1 OTA. The attempt to flash part1 should fail (an intermediate update to 0.7.0-rc.5 is required).
3. Update the firmware to 0.7.0-rc.5 OTA. The device should boot into safe mode due to unmet dependency on the new bootloader. Flash 0.7.0-rc.5 bootloader OTA.
4. Update the firmware to 0.8.0-rc.1.
5. Check that DCT operations in DFU mode work as expected (e.g. by invoking `particle keys doctor`).

### References

- 0.7.x fixes: https://github.com/spark/firmware/pull/1433
- [CH9466]

---

### Completeness

- [ ] Added documentation


sergeuz:
No need to merge this branch since we have merged 0.7.0-rc.5 (which includes the same changes) into develop

m-mcgowan:
In develop, the system-part1 dependency is still pointing to 0.7.0-rc.4 - I don't see the need to bump this to 0.7.0-rc.5 since rc.4 will work fine (it includes the wifi resource function that we need.)

m-mcgowan:
I'm not sure we need this in develop any longer - the main bootloader change has already been merged. The photon/p1 wifi dependency on 0.7.0-rc.4 doesn't need bumping, since 0.7.0-rc.4 is still a valid dependency.

sergeuz:
I don't quite get why we would want part1 to depend on 0.7.0-rc.4 instead of 0.7.0-rc.5. This fix has nothing to do with the WiFi resources, it prevents the system from updating a pre-0.7.0-rc.5 firmware straight to 0.8.0-rc.1, because an intermediate update is required in order to fetch the fixed bootloader.

m-mcgowan:
This has already been merged into develop, so closing this one.

PR #1439, opened 2017-11-22
Feature/go faster stripes

m-mcgowan:
### Problem

The CI build is slower than a herd of snails travelling though peanut butter, taking around 50 minutes to complete.

### Solution

- Build multiple platforms in parallel
- Report all failed tests (not stop at the first, reducing the number of failed repeat builds)
- use Ccache ompiler
- cache the built Boost binaries

NB: For tagged builds, each individual platform docker image will be uploaded when the build for that platform succeeds, even when some of the others fail. This will be addressed in another PR. 

### Completeness

- [ ] Added documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,20 @@ sudo: required
 services:
   - docker
 install:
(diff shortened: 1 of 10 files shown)


suda:
From **33 min 34 sec** to **12 min 46 sec** 😱  Noice!

PR #1440, opened 2017-12-01
Device configuration requests

sergeuz:
### Problem

This PR implements the following control requests:

- Firmware factory reset (`CTRL_REQUEST_FACTORY_RESET`)
- Setting the claim code (`CTRL_REQUEST_SET_CLAIM_CODE`)
- Checking if the device is claimed (`CTRL_REQUEST_IS_CLAIMED`)
- Setting/getting device/server keys (`CTRL_REQUEST_SET_SECURITY_KEY`, `CTRL_REQUEST_GET_SECURITY_KEY`)
- Setting/getting server address (`CTRL_REQUEST_SET_SERVER_ADDRESS`, `CTRL_REQUEST_GET_SERVER_ADDRESS`)
- Setting/getting the cloud transport protocol (`CTRL_REQUEST_SET_SERVER_PROTOCOL`, `CTRL_REQUEST_GET_SERVER_PROTOCOL`)
- Setting prefix/suffix of the Soft AP's SSID (`CTRL_REQUEST_SET_SOFTAP_SSID`)
- Starting/stopping nyan LED indication (`CTRL_REQUEST_START_NYAN_SIGNAL`, `CTRL_REQUEST_STOP_NYAN_SIGNAL`)

### Steps to Test

I suggest we introduce a convenient API for the above requests into the [USB library](https://github.com/sergeuz/particle-usb) first, and then test the requests using the library.

### References

- [CH9456]

---

### Completeness

- [ ] Added documentation


system/src/system_control_internal.cpp
```diff
@@ -152,6 +157,81 @@ int decodeRequestMessage(ctrl_request* req, const pb_field_t* fields, void* dst)
     return ret;
 }
 
+// Helper classes for working with nanopb's string fields
```
m-mcgowan: Some more comments about what this does would be helpful.

system/src/system_control_internal.cpp
```diff
@@ -259,6 +344,150 @@ void particle::SystemControl::processRequest(ctrl_request* req, ControlRequestCh
...
+        particle_ctrl_SetClaimCodeRequest pbReq = {};
+        int ret = decodeRequestMessage(req, particle_ctrl_SetClaimCodeRequest_fields, &pbReq);
+        if (ret == 0) {
+            ret = HAL_Set_Claim_Code(pbReq.code);
+        }
+        setResult(req, ret);
+        break;
+    }
+    case CTRL_REQUEST_IS_CLAIMED: {
+        const int ret = (HAL_IsDeviceClaimed(nullptr) ? SYSTEM_ERROR_NONE : SYSTEM_ERROR_NOT_FOUND);
```
m-mcgowan: Is an error code the correct response for not claimed? I would imagine SYSTEM_ERROR_NONE to be returned in each case and the logical result returned as a response.

system/src/system_control_internal.cpp
sergeuz: 👍

system/src/system_control_internal.cpp
sergeuz: I decided to save some space on simple getter requests returning `true` or `false`. Here is how the client side is implemented for this request: https://github.com/particle-iot/particle-usb/blob/master/src/cloud-device.js#L35

Let me know if you'd prefer this request to be reworked though.

m-mcgowan: I'd prefer it to be reworked, since I feel it's semantically wrong to use an error code to indicate a negative result to a successful request. We can centralize the method for returning a boolean result, so the overhead is incurred only once.

PR #1452, opened 2017-12-21
Feature/heap fragmentation

m-mcgowan:
### Problem

The application does not know when the system has ran out of memory, and cannot determine if it is due to total exhaustion of the heap or due to a fragmented heap. 

### Solution

Adds a new system event, `out_of_memory`, which is fired when a heap request fails. The `flags` parameter for the posted event is the size of the block requested. The event is posted on the application thread - recursive invocation of this event (in the case where attempting to post the event itself causes an out of memory error) raises a `PANIC`.

The `HAL_Core_Runtime_Info` function has the struct extended with an additional field `largest_free_block_heap` which is the size of the largest free block in the heap (i.e. the largest block that can be allocated.)  With a fragmented heap, the largest block may be significantly less than the amount of free memory. 

The integration tests verify the event is fired, and verify how the heap behaves when it is fragmented. 

### Steps to Test

- `wiring/no_fixture`  SYSTEM_06*, SYSTEM_07*, built with USE_THREADING=y && n

### Example App

todo - do not merge until the docs PR is linked here. 

### Completeness

- [ ] Added documentation


hal/src/electron/rtos/FreeRTOSv8.2.2/FreeRTOS/Source/portable/MemMang/heap_4_lock.c
```diff
@@ -324,8 +363,8 @@ void *pvReturn = NULL;
     {
         if( pvReturn == NULL )
         {
-            extern void vApplicationMallocFailedHook( void );
-            vApplicationMallocFailedHook();
+            extern void vApplicationMallocFailedHook( size_t );
+            vApplicationMallocFailedHook(xWantedSize);
```
sergeuz: Shouldn't we report the originally requested size here (`xWantedSize - xHeapStructSize`) similarly to the Photon implementation?

Another problem here is that FreeRTOS invokes `vApplicationMallocFailedHook()` when the size passed to `pvPortMalloc ()` is 0. Given that our current `malloc()` is a trivial wrapper over `pvPortMalloc()` this would cause the `out_of_memory` event to be generated for the `void* ptr = malloc(0)` statement, which is perfectly valid in terms of the standard library.

m-mcgowan: yes, in fact, I wanted to move that file to a common location and avoid the duplication. I will do that.  And I will add a 0 size check so that the system event is not fired. Thanks for picking that up - really good catch!

hal/src/portable/FreeRTOS/heap_4_lock.c
```diff
@@ -0,0 +1,573 @@
...
+            mtCOVERAGE_TEST_MARKER();
+        }
+
+        traceMALLOC( pvReturn, xWantedSize );
+    }
+    __malloc_unlock(NULL);
+
+    #if( configUSE_MALLOC_FAILED_HOOK == 1 )
+    {
+        if( pvReturn == NULL && (xWantedSize-xHeapStructSize)>0)
```
sergeuz: `(xWantedSize-xHeapStructSize)>0` is always `true` given that both operands are unsigned integers, and `xWantedSize` is adjusted only if it's larger than 0

m-mcgowan: strange - the tests show that the behaviour is as expected.

sergeuz: hm, `SYSTEM_08*` fails for me

PR #1456, opened 2018-01-08
Firmware update and access to internal flash via USB requests

sergeuz:
### Problem

This PR is based on [feature/usb-requests](https://github.com/particle-iot/firmware/tree/feature/usb-requests) and introduces the following changes:

* The firmware now can be updated via USB requests (see `START_FIRMWARE_UPDATE`, `FINISH_FIRMWARE_UPDATE`, `CANCEL_FIRMWARE_UPDATE` and `FIRMWARE_UPDATE_DATA` requests specifically).

* The system now exposes its internal storage (firmware modules, factory backup, DCT and EEPROM) in a self-descriptive manner via `DESCRIBE_STORAGE`, `READ_SECTION_DATA`, `WRITE_SECTION_DATA` and `CLEAR_SECTION_DATA` requests.

* `system_ctrl_set_result()` now accepts an optional completion callback making it possible for the system to, for example, reset the device only after a reply to the host is sent completely. This helps to avoid USB errors at the client side.

* `system_ctrl_alloc_reply_data()` now acts as `realloc()` and, consequently, `system_ctrl_free_reply_data()` has been removed.

* The `LISTENING_MODE` request has been split into two requests: `START_LISTENING` and `STOP_LISTENING`.

* Protocol files have been moved to a git submodule ([firmware-protobuf](https://github.com/particle-iot/firmware-protobuf)). The firmware repo still contains the protocol files in a precompiled form, so, normally, there's no need for our users or CI to perform any additional steps in order to build the firmware.

### Steps to Test

As of now, the easiest way to test the USB requests is to use the [particle-usb](https://github.com/particle-iot/particle-usb) library

### References

* [CH10193]
* [CH10383]

---

### Completeness

- [ ] Added documentation


hal/src/stm32f2xx/ota_flash_hal_stm32f2xx.h
```diff
@@ -40,9 +42,22 @@ typedef enum server_protocol_type {
 
 extern const module_bounds_t* module_bounds[];
 extern const unsigned module_bounds_length;
-extern const module_bounds_t module_ota;
+
 extern const module_bounds_t module_bootloader;
+
+// Modular firmware
+extern const module_bounds_t module_system_part1;
```
m-mcgowan: we should pull these into an internal HAL header

services/inc/eeprom_emulation.h
```diff
@@ -226,7 +226,7 @@ class EEPROMEmulation
 
     // Returns number of bytes that can be stored in EEPROM
     // The actual capacity is set to 50% of the records that fit in the smallest page
-    constexpr size_t capacity() const
+    static constexpr size_t capacity()
```
m-mcgowan: why remove the `const` declaration?

system/inc/system_dynalib.h
```diff
@@ -91,12 +91,11 @@ DYNALIB_FN(BASE_IDX + 8, system, system_format_diag_data, int(const uint16_t*, s
...
 DYNALIB_FN(BASE_IDX + 10, system, system_ctrl_alloc_reply_data, int(ctrl_request*, size_t, void*))
-DYNALIB_FN(BASE_IDX + 11, system, system_ctrl_free_reply_data, void(ctrl_request*, void*))
-DYNALIB_FN(BASE_IDX + 12, system, system_ctrl_free_request_data, void(ctrl_request*, void*))
-DYNALIB_FN(BASE_IDX + 13, system, system_ctrl_set_result, void(ctrl_request*, int, void*))
+DYNALIB_FN(BASE_IDX + 11, system, system_ctrl_free_request_data, void(ctrl_request*, void*))
+DYNALIB_FN(BASE_IDX + 12, system, system_ctrl_set_result, void(ctrl_request*, int, ctrl_completion_handler_fn, void*, void*))
 
-DYNALIB_FN(BASE_IDX + 14, system, system_pool_alloc, void*(size_t, void*))
-DYNALIB_FN(BASE_IDX + 15, system, system_pool_free, void(void*, void*))
+DYNALIB_FN(BASE_IDX + 13, system, system_pool_alloc, void*(size_t, void*))
```
m-mcgowan: any reason not to make the other alloc/free pairs function like realloc?

system/src/control/storage.cpp
```diff
@@ -0,0 +1,496 @@
...
+    HAL_EEPROM_Put(address, data, size); // FIXME: Result code?
+    return 0;
+}
+
+int eepromClear(uintptr_t address, size_t size) {
+    HAL_EEPROM_Clear(); // FIXME: Result code?
+    return 0;
+}
+
+const Section INTERNAL_STORAGE_SECTIONS[] = {
```
m-mcgowan: Could we add a storage for the entire flash space? It's low level, but might be useful in some situations.

system/src/system_control_internal.cpp
```diff
@@ -77,29 +78,36 @@ particle::SystemControl::SystemControl() :
 void particle::SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
     switch (req->type) {
     case CTRL_REQUEST_RESET: {
-        System.reset();
-        setResult(req, SYSTEM_ERROR_NONE);
+        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
```
m-mcgowan: really nice - I realized that before the system would reset before sending the response. This way is much better.

system/src/usb_control_request_channel.cpp
```diff
@@ -183,36 +183,18 @@ particle::UsbControlRequestChannel::~UsbControlRequestChannel() {
...
-    }
-    // Try to allocate the buffer from the pool
-    if (size <= USB_REQUEST_MAX_POOLED_BUFFER_SIZE &&
-            (req->reply_data = (char*)system_pool_alloc(size, nullptr))) {
-        req->flags |= RequestFlag::POOLED_REP_DATA;
-    } else {
-        // Allocate the buffer dynamically
-        req->reply_data = (char*)t_malloc(size);
-        if (!req->reply_data) {
+    if (size > 0) {
```
m-mcgowan: are we not using the system pool any more? what was the reason for using it to begin with?

PR #1462, opened 2018-01-16
Fixes GCC virtual device running with UDP protocol

avtolstoy:
<details>

```

```
</details>

### Problem

There are two issues that prevent GCC virtual device to successfully connect to the cloud with UDP:
1. (main) The UDP socket is being bound to a multicast address 224.0.0.251, instead of 0.0.0.0 or some interface address
2. Cloud sockets use the same source UDP port, which prevents from running several virtual devices on the same host.

### Solution

1. Bind to 0.0.0.0
2. Use ephemeral source port when creating UDP cloud socket

### Steps to Test

Compile `PLATFORM=gcc` and test whether it can connect to the cloud with UDP protocol.

### Example App

N/A

### References

- [CH10641]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -1104,8 +1104,13 @@ int spark_cloud_socket_connect()
     ip_address_error = determine_connection_address(ip_addr, port, server_addr, udp);
     if (!ip_address_error)
     {
-    		uint8_t local_port_offset = (PLATFORM_ID==3) ? 100 : 0;
-        sparkSocket = socket_create(AF_INET, udp ? SOCK_DGRAM : SOCK_STREAM, udp ? IPPROTO_UDP : IPPROTO_TCP, port+local_port_offset, NIF_DEFAULT);
+#if PLATFORM_ID == 3
+        // Use ephemeral port
+        uint16_t bport = 0;
+#else
+        uint16_t bport = port;
+#endif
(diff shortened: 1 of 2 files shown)


system/src/system_cloud_internal.cpp
```diff
@@ -1104,8 +1104,13 @@ int spark_cloud_socket_connect()
     ip_address_error = determine_connection_address(ip_addr, port, server_addr, udp);
     if (!ip_address_error)
     {
-    		uint8_t local_port_offset = (PLATFORM_ID==3) ? 100 : 0;
-        sparkSocket = socket_create(AF_INET, udp ? SOCK_DGRAM : SOCK_STREAM, udp ? IPPROTO_UDP : IPPROTO_TCP, port+local_port_offset, NIF_DEFAULT);
+#if PLATFORM_ID == 3
```
m-mcgowan: can we push this down to the HAL to avoid platform specifics in non-HAL code

avtolstoy: I'd disagree. In that case all UDP sockets would use ephemeral ports. For some specific use-cases we might need to use a specific UDP port.

We have something similar on Electron: https://github.com/particle-iot/firmware/blob/develop/system/src/system_task.cpp#L292

Might be a good idea to move this block there.

m-mcgowan: What I mean is an option that lets the platform HAL decide the source port specifically for the cloud socket. Perhaps an option "UseEphemeralPortForCloud" that is sourced from the HAL, so there is no platform conditionality in the non-HAL code.

PR #1465, opened 2018-01-17
Serialize access to the CRC peripheral (STM32F2xx)

sergeuz:
### Problem

The CRC peripheral is not thread-safe and thus concurrent hardware-accelerated CRC calculations should be serialized. 

While there's no evidence that the system and application threads perform CRC calculations concurrently, a synthetic test shows that such a situation can cause invalidation of the DCT.

### Solution

Guard the platform's `Compute_CRC32()` function with a global lock.

### Steps to Test

- `wiring/no_fixture`: `CONCURRENT_02*`

### References

- [CH10651]

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/src/stm32f2xx/concurrent_hal.c
+++ b/bootloader/src/stm32f2xx/concurrent_hal.c
@@ -1,12 +1,15 @@
-
-int os_thread_yield() {
+int os_thread_yield(void) {
 	return 0;
 }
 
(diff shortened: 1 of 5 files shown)


sergeuz:
> Would be good with some tests too to demonstrate the thread safety

I'll strip that DCT stress test application which I used to reproduce the issue and turn it into a test app 👍

PR #1466, opened 2018-01-18
Decouple recipes from order of their prerequisites

jablko:
Fix breakage in 9aa4b8f: `$<` is the first prerequisite.


m-mcgowan:
Thank you for this, although as you noted the issue has already been fixed!

PR #1477, opened 2018-01-23
heap upperbound overlaps system-part1 (module index 3) static RAM

m-mcgowan:
### Problem

SYSTEM_07_* memory stress test was failing on the Electron. This was because the heap extended into the region of static RAM for system-part1 (module index 3).

There were no assertions for the symbol `link_heap_location_end` and so this issue went undetected at build time.

### Solution

Move the upper bound of the heap to before the static RAM block.  The symbol `system_heap_end` was already pointing to the correct location. 

### Steps to Test

`SYSTEM_07*` in `wiring/no_fixture`

---

### Completeness

- [ ] Added documentation


towynlin:
👏

PR #1478, opened 2018-01-23
fix heap bounds build for system part1

m-mcgowan:


m-mcgowan:
I will apply this fix also to 0.7.0-rc.7

Issue #1461, opened 2018-01-15
Photon TCP can blocks indefinitely with new wiced 3.7 (release 0.7x and 0.8x)


tlangmo:
## Bug Report
Whenever actively sending over tcp, the application thread can stall indefinitely long in the new releases. This causes the application watchdog to timeout and generally stops any use code from executing.

### Expected Behavior
On Wiced SDK 3.3 (particle release 0.6.x and prior), sending tcp data returned WICED_RESULT_WOULD_BLOCK (-16) whenever the network stack was not able to send the data. This typically happens when the connection quality is poor, for example if the device moves out of the wifi range. The calling thread continued executing and the user code can handle this condition properly (such as sending at a later point)

### Observed Behavior
In Wiced SDK 3.7, this behavior was changed according to its changelog.txt.
`LwIP network_tcp_send_packet() no longer sets NETCONN_DONTBLOCK when calling netconn_write()`

Now, TCPClient::write and everything else which uses wiced_tcp_send_buffer will simply block.
This causes user code to stall whenever having a bad wifi signal during an active tcp connection. I have seen stalls of many minutes until either the Wifi connection reset itself, the photon crashed or the photon resets itself, similar to watchdog (although I have not activated one)

### Steps to Reproduce
The key is to have a weak WiFi signal. This can be achieved by walking out of range or shielding the signal. Using the external antenna and unplugging it worked in many cases as well. 

Compiling against releases < 0.7.x shows that the app loop never stalls

### Test App  (application.cpp)
The app monitors the app thread by blinking and serial output every 2 seconds. 
1. Connect to the photon via the python script
2. Reduce wifi signal and perceive the blocking

```c
#include "application.h" 

PRODUCT_ID(PLATFORM_ID);

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);
STARTUP(WiFi.selectAntenna(ANT_EXTERNAL));
STARTUP(System.enableFeature(FEATURE_RESET_INFO));

SerialLogHandler logHandler(LOG_LEVEL_INFO);

#define WIFI_SSID  "ssid"
#define WIFI_PWD  "password"

TCPClient tcpConn;
TCPServer  tcpServer(5000);

void destroyNetworkResources() {
	Log.info("destroyNetworkResources");
	tcpServer.stop();
	tcpConn.stop();
}
void createNetworkResources() {
	Log.info("createNetworkResources");
	tcpServer.begin();
}

[...]

PR #1485, opened 2018-01-31
Photon/P1: TCPClient: non-blocking, blocking, blocking with timeout writes support

avtolstoy:
### Problem

Unlike `TCPClient::read()`, `TCPClient::write()` is a blocking call and whenever TCP send queue gets completely filled (see #1461 for a particular case on Photon/P1 with low wifi signal), the application thread may be blocked for an indefinite amount of time.

There needs to be a way to specify the maximum amount of time the write should take.

### Solution

New `TCPClient::write()` overloads are added:
- `size_t write(uint8_t b, system_tick_t timeout);`
- `size_t write(const uint8_t *buf, size_t size, system_tick_t timeout);`

### Additional notes

`write()` methods return `size_t`, which is an unsigned integer, however in case of an error, `TCPClient::write()` method may return a negative error code cast to `size_t`. We are still keeping this behavior, with a thought of deprecating it in the next release.

Applications should use `TCPClient::getWriteError()/clearWriteError()` methods to check for errors, while `TCPClient::write()` should only return the number of bytes written.

### Steps to Test

- `wiring/api`

### Example App

```c++
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

Serial1LogHandler dbg(115200, LOG_LEVEL_ALL);

TCPServer tcpServerNonBlocking(5000);
TCPServer tcpServerBlockingTimeout(5001);
TCPServer tcpServerBlocking(5555);

void setup() {
    WiFi.on();
    WiFi.connect();
    WiFi.selectAntenna(ANT_EXTERNAL);
}

static void sendBunchOfData(TCPClient&& client, system_tick_t timeout) {
    while (client.connected()) {
        uint8_t buf[500];
        memset(buf, 'a', 500);
        int rc = client.write(buf, sizeof(buf), timeout);
        LOG(TRACE, "%d", rc);
    }
}

void loop() {
    sendBunchOfData(tcpServerNonBlocking.available(), 0);
    sendBunchOfData(tcpServerBlockingTimeout.available(), 1000);
    sendBunchOfData(tcpServerBlocking.available(), SOCKET_WAIT_FOREVER);
}
```

Connect to the appropriate port using telnet or something similar, while looking at the log place a finger onto the external antenna connector. Device should be unable to send data, causing TCP send queue to fill up. The behavior on each of the ports should be the following:
- 5000: as soon as the tcp queue is full, the log should be filled with a constant flow of -16 error codes
- 5001: as soon as the tcp queue is full, the log should show -16 error codes every second
- 5555: as soon as the tcp queue is full, there should be no log entries, indicating that the application thread is blocked

### References

- #1461
- particle-iot/photon-wiced#25
- [CH10639]

---

### Completeness

- [ ] Added documentation **NEEDS TO BE UPDATED**


wiring/inc/spark_wiring_print.h
```diff
@@ -66,6 +67,9 @@ class Print
     }
     virtual size_t write(const uint8_t *buffer, size_t size);
 
+    virtual size_t write(uint8_t b, system_tick_t timeout) { return write(b); }
```
m-mcgowan: I would prefer to be careful here and add the new `write` methods to TCPClient only. The reason is that there is only one stream implementation that supports it, and I imagine users will be confused with the presence of the new methods that don't do anything useful for most streams.

wiring/src/spark_wiring_tcpclient.cpp
```diff
@@ -110,12 +110,30 @@ int TCPClient::connect(IPAddress ip, uint16_t port, network_interface_t nif)
...
+}
+
+size_t TCPClient::write(uint8_t b, system_tick_t timeout)
+{
+    return write(&b, 1, timeout);
+}
+
+size_t TCPClient::write(const uint8_t *buffer, size_t size, system_tick_t timeout)
+{
+    int ret = status() ? socket_send_ex(d_->sock, buffer, size, 0, timeout, nullptr) : -1;
```
m-mcgowan: should we call `clearWriteError()` here?

elcojacobs:
How would you feel about being able to set a timeout value on the TcpClient class instance, instead of passing it with every function call?
That avoids the function overloading and I think will be even easier to use.

avtolstoy:
Sounds like a good idea analogous to how it's done with regular BSD-socket interface: `setsockopt()` with `SO_RCVTIMEO` and `SO_SNDTIMEO`. Having both options available also doesn't sound too bad :)

I also agree that `size_t write(const char *str, system_tick_t timeout)` needs to be added. It was an oversight on my part.

elcojacobs:
Are you going to implement this? I can have a go at it too if you want.

PR #1492, opened 2018-02-28
Photon/P1: A number of networking-related fixes

avtolstoy:
<details>
  <summary><i>submission notes</i></summary>

```
**Important:** Please sanitize/remove any confidential info like usernames, passwords, org names, product names/ids, access tokens, client ids/secrets, or anything else you don't wish to share.

Please Read and Sign the Contributor License Agreement ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md)).

You may also delete this submission notes header if you'd like. Thank you for contributing!
```
</details>

### Problem

1. WICED makes a write to DCT while DCT is locked for reading causing an assertion failure (particle-iot/photon-wiced#27)
2. There is a race condition between HAL and WICED in DHCP handling: DHCP may be started non-synchronously under certain conditions (and we assume that it's synchronous in wlan_hal)
3. There is a bug with tcp server resource cleanup in socket_hal: when `wiced_tcp_server_start()` reports an error, we do not call `wiced_tcp_server_stop()`, causing an internal tcp server list in WICED to get completely filled. As a result, we are unable to create any more tcp servers
4. There is a race condition with `SYSTEM_THREAD(ENABLED)` between `WiFi.ready()` and when the config is updated in system thread (#1371)
5. LwIP `tcp_tmr` no longer functions after network stack is reinitialized (particle-iot/photon-wiced#27)
6. `socket_close_all()` causes a hardfault due to socket destruction when it should simply be closing them and leaving the resource cleanup up to the socket 'user'
7. `Pinger` is not functioning properly, hence #1479

### Solution
- Photon: `socket_create_tcp_server()` needs to `close()` server instance in case of an error before deleting it
- wlan_hal: call `wiced_network_down()` in `wlan_connect_finalize()` to prevent DHCP from being started asynchronously under certain conditions
- system_network_internal: ip config needs to be updated in `notify_dhcp()` in order to mitigate an issue when `network_ready() == true` but the config has not been updated yet
- Photon: socket_hal: Fixes a hardfault due to socket destruction in `socket_close_all()`
- communication: Fixes broken `Pinger` (#1479)
- particle-iot/photon-wiced#27

### Steps to Test

N/A

### Example App

N/A

### References

- Closes #1479
- Closes #1490
- Closes #1371
- particle-iot/photon-wiced#27

---

### Completeness

- [x] User is totes amazing for contributing!
[...]

Issue #1371, opened 2017-08-20
Photon/P1 does not always recover from a WiFi channel switch / WiFi.ready() can return true with IP 0.0.0.0


elcojacobs:
## Bug Report
According to the docs, WiFi.ready() returns true only if the device is connected to WiFi and has a valid IP address. My P1 based board connects to WiFi without problems initially. Then it loses network, and the code I have to restore WiFi is not triggered, because WiFi.ready() stil returns true.

I have narrowed down the cause to WiFi channel switching. Auto channel seems to be the default on many routers, so this affects many setups.

Initially, this bug report was about WiFi.ready() returning 'true' while the IP address was `0.0.0.0`, but this seems to be one of the symptoms of WiFi channel switching not being handled correctly.

#### WiFi.ready()
WiFi.ready() returns true only if the P1 has a valid IP address, as described here:
https://docs.particle.io/reference/firmware/photon/#ready-

What should it do?
Return `false` if the IP address is `0.0.0.0`.

#### Channel switching
The photon should be able to switch seamlessly to a new WiFi channel if the router changes the channel.

### Observed Behavior

When I switch the WiFi channel on my router, the behavior of the photon is not consistent.
I see one of the following behaviors:

* Photon starts blinking green. Display shows WiFi.ready() returning false and IP address becoming 0.0.0.0. A short blue blink is followed by a white blink and then green blinking again and a successful reconnect.
* Photon starts flashing green, then switches to breathing green. WiFi.ready() stays true, but IP address becomes 0.0.0.0. Doesn't recover from this state. It's in limbo until a reboot.
* Hard fault SOS (1 blink).

### Analysis of the problem

In my application, I display WiFi status on a TFT display. It displays the IP address in light grey when connected and in dark grey when not connected (code below).

![2017-08-20 12 51 47](https://user-images.githubusercontent.com/1708921/29494587-dfb846a0-85ad-11e7-830b-b1f1f5ec7425.jpg)

```
void ControllerWiFiView::update(bool wifiConnected, char * const ipAddress)
{
    if(obj != nullptr){
        obj->clrScheme->fore = wifiConnected ? D4D_COLOR_LIGHT_GREY : D4D_COLOR_GREY;
        D4D_SetText(obj, ipAddress);
    }
}

void ControllerWiFiPresenter::update(){
#if BREWPI_USE_WIFI
    char ipAddressString[16];
    bool wifiConnected = WiFi.ready();
[...]

PR #1500, opened 2018-03-14
Photon/P1: A number of networking-related fixes [v2]

avtolstoy:
### Solution

1. Fixes a socket resource leak in `TCPServer::available()`
2. Default write timeouts added to `TCPClient` (30s) and cloud socket (20s)
3. Removed `socket_close_all()` from `wlan_disconnect_now()`
4. Makes `socket_xxx()` calls thread safe
5. particle-iot/photon-wiced#29

This PR is based on #1492 and targets its branch for ease of review.

### Steps to Test

N/A

### References

- #1492
- particle-iot/photon-wiced#29

---

### Completeness

- [ ] Added documentation


elcojacobs:
Good to see you're working on this again.

Any other steps needed to test this than merging the branch?
Any tests you want me to re-run?

Could/should this branch be rebased on 0.8.0-rc.2?

avtolstoy:
Hey @elcojacobs.

Apart from fixing the hardfault you were seeing in #1371, there are a couple of preventive measures implemented here that should guard the system thread from being blocked (at the very least for a substantial amount of time) due to socket and networking operations.

I did test this with your `wifi-test` application, but I'd appreciate if you could try these changes in your environment.

The only issue that this PR does not address (because I couldn't reproduce it) is the latest blinking green reported in #1492, but I'm hoping it will no longer present itself.

Let me know how it goes, and I'm hoping we can release these changes in 0.8.0-rc.3 as soon as possible.

avtolstoy:
> Could/should this branch be rebased on 0.8.0-rc.2?

You may merge this branch into `0.8.0-rc.2` if needed, but it's based on #1492 originally, which in turn should be based on `develop`, which should be at or around 0.8.0-rc.2. So in theory you can use it as-is as well.

elcojacobs:
Allright, I'll skip the 0.8.0-rc.2 rebase. The changes seem to mostly in version flags.
I think the fast blinking green could be explained by a hanging system thread. If that was caused by a hanging TCP write, this would indeed prevent that.

As a last failsafe watchdog, I would like to reset the network stack if no communication has taken place for 15 minutes.
What would be the safest and most complete way to do that without rebooting? I don't want to reset the user thread, but I do want to reset everything related to network and TCP.

avtolstoy:
Depending on the meaning of "no communication has taken place for 15 minutes", I'd say something like running a separate thread (or using an application watchdog) and doing `WiFi.off()` and `WiFi.on()` in it should do it, but it's not a bullet-proof solution, and might not help in certain cases.

elcojacobs:
Ok, based on your edits, I guess WiFi.off() and Wifi.on() are causing problems? What did you encounter?

I will stop TCP communication before taking down WiFi on purpose.

The user thread is not blocked, so I can add the check there. I already have a hard reboot watchdog for when the user thread stalls.
I can wrap parts of the code in SINGLE_THREADED_BLOCK(), as long as the code inside that block has no potential to block.

Ideally, I would be able to set a flag that causes the system thread to restart WiFi in the background, without blocking the user thread.
If there are flags I need to monitor to prevent problems, what would be no problem either.

Pseudo code example:

```
if(no communication since 15 min){
   if (wifi enabled){
      tcpClient.stop();
      tcpServer.stop();
      Particle.disconnect();
      wifi.disconnect();
      wifi.off()
   }
   if(wifi fully shut down){
       wifi.on();
       wifi.connect(WIFI_SKIP_LISTEN);
   }
}
```

elcojacobs:
wlan_restart(nullptr) results in hard fault and wrapping it in a SINGLE_THREADED_BLOCK() results in going from breathing cyan to breathing green, without connectivity (ping).
```
                    SINGLE_THREADED_BLOCK(){
                        // wlan_restart is not thread safe
                        wlan_restart(nullptr);
                    }
```

avtolstoy:
@elcojacobs Wrapping `wlan_restart()` in `SINGLE_THREADED_BLOCK()` will definitely not work, as some of the internal WICED threads need to cleanly exit during that call.

If your system thread is working normally (and I hope that now we've eliminated any potential network-related issues/lockups), then as I originally suggested `WiFi.off()/on()` should work just fine. Otherwise it's a bug that should be fixed.

elcojacobs:
```
WiFi.off();
WiFi.on();
```
Seems to cause no crashes. It does result in a strange status LED when I turn off WiFi on the router.

State: breathing cyan, connected to cloud, everyone is happy.

Action: turn off wifi on router
New state: breathing dark blue

Action: re-enable WiFi
New state: fast blinking green (connecting to WiFi).

It remains in this state, even after the WiFi connection has been established. It responds to ping and local TCP traffic. WiFi.ready() returns false.
If I reset the device, it connects to the cloud and breaths cyan again.


Alternative:
```
WiFi.off();
WiFi.on();
Particle.connect();
```
State: breathing cyan, connected to cloud, everyone is happy.

Action: turn off wifi on router
New state: blinking green

Action: re-enable WiFi
New state: fast blinking cyan, followed by breathing cyan.

----------
Seems like the second solution works. Leaving out Particle.connect() seems to trigger a bug.

elcojacobs:
Upon further testing, first behavior also occurred with the code below to reset:

```
                    Particle.disconnect();
                    WiFi.off();
                    WiFi.on();
                    Particle.connect();
```

Going to slow breathing dark blue on WiFi loss, staying in fast blinking green on WiFi reconnect.

It is probably timing related. Maybe the WiFi.off() call is made just when the wifi is reconnecting.

elcojacobs:
Wrapping in if(WiFi.connecting()) seems to be a good workaround.

```
            ticks_seconds_t sinceReceive = ticks.timeSinceSeconds(lastReceive);
            ticks_seconds_t sinceReset = ticks.timeSinceSeconds(lastReset);
            ticks_seconds_t now = ticks.seconds();
            if(sinceReceive > TCP_RESTART_ON_NO_RECEIVE_INTERVAL && sinceReset > TCP_RESTART_ON_NO_RECEIVE_INTERVAL){
                // no receive for the past 10 minutes
                if(tcpServerRunning){
                    stopTcp();
                    lastReset = now;
                    // restart will be handled next time this function is called if WiFi is ready.
                }
                if(sinceReceive > WIFI_RESTART_ON_NO_RECEIVE_INTERVAL){
                    // Restarting TCP didn't help
                    stopTcp();
                    if(!WiFi.connecting()){
                        Particle.disconnect();
                        WiFi.off();
                        WiFi.on();
                        Particle.connect();
                    }
                    lastReceive = now; // reset last receive to delay next restart
                    lastReset = now;
                }
            }
            else if(!WiFi.ready() || WiFi.RSSI() >= 0){
                // WiFi is in error state, stop TCP server
                if(tcpServerRunning){
                    stopTcp();
                }
            }
            else{
                if(!tcpServerRunning){
                    startTcp();
                }
            }
```

elcojacobs:
I found a new issue.

When building with MODULAR=n

* The photon does not connect to WiFi, it keeps just blinking green
* When I enter listening mode and use the Particle app to set up the photon, I get a hard fault SOS (1 blink)

The same code works with a modular build.

I wasn't sure whether to list this as a new issue or comment here.

elcojacobs:
To confirm that this is a bug in Particle code:

Just run in the /main directory of this repo (on this branch):
```
make3 clean all PLATFORM=p1 WARNINGS_AS_ERRORS=n MODULAR=n program-dfu
```

And confirm that the device does not connect to WiFi anymore.
This is with the default Particle app!

elcojacobs:
In addition to the bug above, please do some testing of listening mode with these changes. A user testing this for me experienced issues setting up WiFi with the Particle app.

PR #1505, opened 2018-03-23
Re-create pull request: Add new platform Duo running Particle firmware

XuGuohui:
This pull request enables the platform Duo to run the Particle firmware.

Key notes to be highlighted:

- Memory map differs to Photon
- Wi-Fi firmware, OTA and FAC images are placed in external flash
- Bootloader: shares the STM32 stdperiphdriver libraries with Photon
- HAL: shares the same HAL libraries with Photon, both based on the same WICED SDK.
- Modular: share the same modular source code with Photon
- Supports BLE functionality
- Enables BLE setup during Listening mode

Please help review. Any comments are welcomed.

Best regards,
Guohui


aronsemle:
Can anyone report the status of this? We're experiencing connection reliability issues with the current redbear firmware 0.3.1. We get WiFi disconnects that leave the redbear blinking GREEN or LITE BLUE and never connecting without a power cycle/reset. This is hurting us as we scale up, because we need to call customers and have them reset the device. We're assuming that this merge might help with the particle connectivity stuff by getting on the latest particle firmware? Maybe?

We don't mind testing/helping move this along.

XuGuohui:
Hi @aronsemle , please try the latest released firmware v0.3.3: https://github.com/redbear/firmware/releases, which is based on particle firmware v0.7.0-rc7.

aronsemle:
@XuGuohui we've been testing v0.3.3 and so far so good! However, it doesn't look like it's supported on the Particle side. In our Particle Product, for a device running 0.3.3 we just see.. "Device OS:
unknown". Any guidance on this? Is this safe? Can it be supported in our Particle Product?

XuGuohui:
Currently this version is not supported , but we will add it later. For now, you can develop your project using Arduino. We have released a new board package version for Arduino.

PR #1517, opened 2018-04-05
[Photon/P1/Electron] Recursive logging freezes the application thread while the LogHandler is trying to acquire a lock on the resource.

### LINKED ISSUE #1497: subclassing LogHandler and implementing logMessage with code that generates logs blocks the application thread
### Expected Behavior

[subclassing LogHandler and implementing logMessage with code that generates logs](https://github.com/barakwei/ParticlePapertrail/blob/master/src/papertrail.cpp#L83) should behave sanely.

### Observed Behavior

The logging within the logger freezes the application thread while the LogHandler is trying to acquire a lock on the resource.

### Steps to Reproduce

Using system firmware 0.8.0-rc.2 and the following app

One fix tested that worked, recompiling user firmware locally changing `Mutex mutex_` to `RecursiveMutex mutex_` from [here](https://github.com/particle-iot/firmware/blob/develop/wiring/inc/spark_wiring_logging.h#L587) and also including a flag to prevent re-entry to the logMessage function (so logs generated from calls within the logMessage function don't cause a stack overflow due to recursion).

### Test App

```c
#include "Particle.h"

class UdpThing : LogHandler
{
private:

    UDP Udp;
    volatile bool active;

public:

    UdpThing(LogLevel level = LOG_LEVEL_INFO) :
            LogHandler(level),
            active(false) {
    }

    void begin()
    {
        Udp.begin(8888);
    }

    void addHandler() {
        LogManager::instance()->addHandler(this);
    }

    void logMessage(const char *msg, LogLevel level, const char *category, const LogAttributes &attr)
    {
        if (active) {
            return;
        }
        active = true;

        Log.info("logger: %s", msg);
        if (millis() > 15000) {
            Udp.beginPacket({8, 8, 8, 8}, 5000);
            Udp.write("testing ");
            Udp.endPacket();
            Log.info("application thread hangs here!!");
        }

        active = false;
    }

};

SerialLogHandler logHandler(LOG_LEVEL_ALL);  // all level over usb

UdpThing udpThing;

void setup()
{
    udpThing.begin();
    udpThing.addHandler();
}

void loop()
{
    Log.info("%lu ping", millis());
    delay(500);
}
```

### References

[issue 1](https://github.com/barakwei/ParticlePapertrail/issues/8)
[issue 2](https://github.com/barakwei/ParticlePapertrail/issues/7)
[issue 3](https://community.particle.io/t/papertrail-for-electron/37508)
[issue 4](https://community.particle.io/t/new-library-papertrail/28105/18?u=bdub)
[issue 5](https://community.particle.io/t/help-udp-send-in-loghandler-hangs-electron-w-0-6-1/30792)

technobly:
### Problem

Recursive logging freezes the application thread while the LogHandler is trying to acquire a lock on the resource (#1497).

### Solution

1. Use a `RecursiveMutex` in `spark_wiring_logging.h`
2. Prevent re-entry into subclassed `logHandler` functions `logMessage` and `logWrite` from the `LogManager` instance

### Steps to Test

Run `TEST=wiring/no_fixture`

### References

- Closes #1497

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_logging.h
+++ b/wiring/inc/spark_wiring_logging.h
@@ -577,7 +577,7 @@ private:
 
     Vector<LogHandler*> activeHandlers_;
 
-    volatile bool output_active_;
+    bool outputActive_;
 
 #if Wiring_LogConfig
     Vector<FactoryHandler> factoryHandlers_;
--- a/wiring/inc/spark_wiring_logging.h
+++ b/wiring/inc/spark_wiring_logging.h
@@ -577,6 +577,8 @@ private:
 
     Vector<LogHandler*> activeHandlers_;
 
+    volatile bool output_active_;
+
 #if Wiring_LogConfig
     Vector<FactoryHandler> factoryHandlers_;
     LogHandlerFactory *handlerFactory_;
@@ -584,7 +586,7 @@ private:
 #endif
 
 #if PLATFORM_THREADING
-    Mutex mutex_; // TODO: Use read-write lock?
+    RecursiveMutex mutex_; // TODO: Use read-write lock?
 #endif
 
     // This class can be instantiated only via instance() method
@@ -602,6 +604,9 @@ private:
     static void logMessage(const char *msg, int level, const char *category, const LogAttributes *attr, void *reserved);
     static void logWrite(const char *data, size_t size, int level, const char *category, void *reserved);
     static int logEnabled(int level, const char *category, void *reserved);
+
+    bool isActive() const;
+    void setActive(bool output_active);
 };
 
 #if Wiring_LogConfig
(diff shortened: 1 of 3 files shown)


wiring/inc/spark_wiring_logging.h
```diff
@@ -577,14 +577,16 @@ class LogManager {
 
     Vector<LogHandler*> activeHandlers_;
 
+    volatile bool output_active_;
```
sergeuz: I know we tend to mix different styles in the code, but let's at least make sure that the code within a single file follows the same naming convention :) Also, `volatile` isn't really necessary here, since the access to this member variable is already guarded by a mutex.

technobly: I started with the word `active_` and then thought it would be better to say what exactly is active.  It wasn't my intention to mix `snake_case` with `camelCase`.  I'll fix it :)

PR #1524, opened 2018-04-11
fixes communication library unit tests [ch13052]

m-mcgowan:
### Problem

The unit tests for the communications library were not run as part of the CI build. 

### Solution

Build and run the catch tests as part of the CI unit tests. 

The communications catch tests rely upon mbedtls, which has been moved to the `crypto` module since the catch tests were originally developed. The ad-hoc makefile has been changed to a regular module. so that we can use other modules as dependencies.

Regular modules will pull in `build.mk` files anywhere under the module root. Here, the catch unit tests module should not be built as part of the communications module.  To exclude the `build.mk` in the tests, the make system has a new variable `MODULE_BUILD_PATH` which is the folder to search for `build.mk` files. It defaults to the `MODULE_PATH` so that existing modules continue working as usual. The communications module defines `MODULE_BUILD_PATH` as the `src` subdirectory so that the `tests` subdirectory is ignored when building the communications library. 

*NB!!* I had to change code in the Pinger to make the tests work.  This code has been changed recently, so this change should be reviewed thoroughly, and if necessary discussed and new tests added to capture the problem that was recently fixed. 

### Steps to Test

run `ci/unit_test.sh`

---

### Completeness

- [ ] Added documentation


communication/src/ping.h
```diff
@@ -44,6 +50,10 @@ class Pinger
 				// timed out, disconnect
 				return PING_TIMEOUT;
 			}
+			else
+			{
+				expecting_ping_ack = false;
```
technobly: This shouldn't be here and was specifically removed in 0.8.0-rc.3 as a bug, which now makes me naturally very suspicious of all changes that are not in the tests directly :)

m-mcgowan: Yes, it was this change that I wanted us to review. I don't have any context for the bug or how the behaviour is not correct, and what the desired behaviour should be.

technobly: I see your original comments now.  It's possible the tests need to be updated and not the code.  But definitely some tests need to be added.

m-mcgowan:
Closing this since it is made obsolete by https://github.com/particle-iot/device-os/pull/1869

PR #1528, opened 2018-04-17
Wait for Wiring Thread to start

### LINKED ISSUE #1527: Wiring Thread causes crash (Regression in 0.7.0)
## Bug Report

Using the Wiring `Thread` API to start a new user thread leads to a crash (SOS HardFault).

### Expected Behavior

Programs using `Thread` that work in 0.6.3 should still work in 0.7.0

### Observed Behavior

Crash as soon as the Wiring thread runs.

### Steps to Reproduce

Flash the app below. It works on a Photon running 0.6.3 (prints increasing numbers on USB serial) but crashes with SOS on 0.7.0.

```c
#include "application.h"

Thread t;

int count = 0;

void run() {
  while(true) {
    count++;
    delay(1);
  }
}

void setup() {
  Serial.begin();
  t = Thread("count", run);
}

void loop() {
  Serial.printlnf("count=%d", count);
  delay(500);
}
```

Note: using the other `Thread` constructor (accepting a `wiring_thread_fn_t`) also leads to a crash.

### Debugging notes

Running this app on the debugger gave some hints of what's going on.

The crash occurs in [`Thread::run`](https://github.com/particle-iot/firmware/blob/647fd310718831ebb631d2cb6c7a9a59054955ba/wiring/inc/spark_wiring_thread.h#L159). The `Thread* th` parameter contains garbage instead of the target function so the MCU jumps to a junk address leading to a HardFault.

I see the root cause by looking at the code now.

The constructor passes `this` as the "thread function parameter" to `os_thread_create`. When the `Thread::run` static method runs, it uses the "thread function parameter" to refer to the instance of the `Thread` class.

However in my example code above, the `Thread` instance was created on the stack and assigned to a global statically-allocated variable. If `Thread::run` runs after the assignment takes places, the "thread function parameter" will point to the old instance on the stack instead of the new global instance so the pointer will contain junk.

### Proposed solution

The solution is to wait the new thread to run before returning from the `Thread` constructor, the same way [the C++ thread constructor does](https://github.com/particle-iot/firmware/blob/ad64959eac6e3e3e68c1e17ef817e5bbbf1dca29/system/src/system_threading.cpp#L101).
* Add a `Thead::started` boolean defaulting to false
* Set `Thread::started` to true in `Thread::run`
* Add `while (!started) os_thread_yield();` to both `Thread` constructors.

### Workaround

Avoid the [`Thread assignment operator`](https://github.com/particle-iot/firmware/blob/647fd310718831ebb631d2cb6c7a9a59054955ba/wiring/inc/spark_wiring_thread.h#L142) to make sure the instance of `Thread` will be the same when the constructor is called and when `Thread::run` is called.

```
#include "application.h"

Thread *t;

int count = 0;

void run() {
  while(true) {
    count++;
    delay(1);
  }
}

void setup() {
  Serial.begin();
  t = new Thread("count", run);
}

void loop() {
  Serial.printlnf("count=%d", count);
  delay(500);
}
```

monkbroc:
Wait for Wiring `Thread` to reach `Thread::run` before returning from the constructor to avoid HardFault crash when the Thread instance is not available anymore.

Fixes #1527. See that issue for details, steps to reproduce and example app.

This is a regression that was introduced in 0.7.0. It should be backported to the 0.7.x line.

### References

- [CH17452]

### Completeness

- [X] Run unit/integration/application tests on device
- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_thread.h
+++ b/wiring/inc/spark_wiring_thread.h
@@ -81,6 +81,8 @@ private:
         }
     };
 
+    // The thread context is allocated dynamically in order to keep all pointers to it valid, even if
+    // the original wrapper object has been moved and subsequently destroyed
     std::unique_ptr<Data> d_;
 
 public:
--- a/wiring/inc/spark_wiring_thread.h
+++ b/wiring/inc/spark_wiring_thread.h
@@ -67,6 +67,7 @@ private:
     mutable wiring_thread_fn_t *wrapper = NULL;
     os_thread_fn_t func_ = NULL;
     void* func_param_ = NULL;
+    bool started_ = false;
     bool exited_ = false;
 
 public:
@@ -79,6 +80,7 @@ public:
           func_param_(function_param)
     {
         os_thread_create(&handle, name, priority, &Thread::run, this, stack_size);
+        while (!started_) os_thread_yield();
     }
 
     Thread(const char *name, wiring_thread_fn_t function,
@@ -88,6 +90,7 @@ public:
         if(function) {
             wrapper = new wiring_thread_fn_t(function);
             os_thread_create(&handle, name, priority, &Thread::run, this, stack_size);
+            while (!started_) os_thread_yield();
         }
     }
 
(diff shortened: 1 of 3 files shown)


wiring/inc/spark_wiring_thread.h
```diff
@@ -67,6 +67,7 @@ class Thread
     mutable wiring_thread_fn_t *wrapper = NULL;
     os_thread_fn_t func_ = NULL;
     void* func_param_ = NULL;
+    bool started_ = false;
```
m-mcgowan: doesn't this need to be volatile since it's accessed by multiple threads without any memory guard?

PR #1530, opened 2018-04-18
Fixes missing URCs for received data during TX or RX socket operations

technobly:
### Problem

We've been getting reports of times when devices seem to stop being able to receive data.  They seem to recover after the device exhausts its retries (about 78 seconds) and reconnects to the Cloud.  This has been a tricky bug to nail down, but it appears to be a corner case in how the cellular modem reports URCs, or in this case, doesn't report them.

The error condition occurs with the following sequence.  We start with the RX buffer is empty.  We receive some data, and the modem notifies us of this pending data via URC (Unsolicited Response Code). When we go to read that data from the modem, if we received more data concurrently, no further URC is generated after the data is received, or after the read completes.  The error case can occur during a write operation as well (receiving data while we are sending data).  We normally rely on URCs to allow us to receive status from the modem asynchronously and avoid polling it for status, which allows it to enter low power mode as well.

### Solution

After extensive testing it was discovered that a forced poll of the socket (AT+USORF=0,0 for UDP and AT+USORD=0,0 for TCP) after all read and write operations was necessary to mitigate this corner case when we don't receive URCs naturally for received data.

There will be a 40 ms delay to poll the modem for pending data after every packet sent and received.  This cost in time comes with the reward of reliability.

### Steps to Test

- `TEST=wiring/no_fixture`
- Extensive internal testing has shown this fix is very reliable. In general what is needed is a fast echo server to send lots of traffic to the device, and for the device to also echo that data back to the server. The server should also let the device repeatedly catch-up and process all messages so that the RX buffer can start from 0 many times.

---

### Completeness

- [N/A] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -311,8 +311,20 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                         else CLR_GPRS_TIMEOUT(); // else if re-attached clear WDT.
                     }
                 // Socket Specific Command ---------------------------------
+                // +USORD: <socket>,<length>
+                } else if ((sscanf(cmd, "USORD: %d,%d", &a, &b) == 2)) {
+                    int socket = _findSocket(a);
+                    DEBUG_D("Socket %d: handle %d has %d bytes pending\r\n", socket, a, b);
+                    if (socket != MDM_SOCKET_ERROR)
+                        _sockets[socket].pending = b;
                 // +UUSORD: <socket>,<length>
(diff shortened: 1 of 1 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1764,6 +1776,12 @@ int MDMParser::socketSend(int socket, const char * buf, int len)
         buf += blk;
         cnt -= blk;
     }
+    LOCK();
+    if (_sockets[socket].pending == 0) {
```
sergeuz: I would add `ISSOCKET(socket)` check here and in similar places in the new code

sergeuz:
Well done! 👍

PR #1532, opened 2018-04-25
Add support for SARA-R410 to the Cellular HAL

sergeuz:
### Problem

This PR adds support for U-blox SARA-R410 modules to the Cellular HAL.

The major difference between SARA-U2/G3 and SARA-R4 modules is the network registration process: in case of SARA-R4 the module automatically establishes a PSD (Packed Switched Data) connection as part of the network registration process, so there's no need to activate a PDP context/EPS bearer explicitly. Cellular HAL has been refactored accordingly (and in a backward-compatible way) in order to support different network registration flows.

Additionally, SARA-R4 doesn't currently support the RTS/CTS flow control and this poses a problem for the runtime modem type detection, since we first need to initialize UART without any flow control enabled, detect the modem type and then reinitialize UART if it appears that the modem does support the hardware flow control. Due to that, in my tests the modem (SARA-U270) or the AT command parser sometimes get into a weird unresponsive state (typically after a system reset), which gets resolved only after an automatic modem reset performed by the `MDMParser`. I can see a similar problem when I run current `develop` with hardware flow control disabled (i.e. with `USE_USART3_HARDWARE_FLOW_CONTROL_RTS_CTS` set to 0). I'd like to invite everyone to try to debug this problem. One workaround, as suggested by Andrey, is to store the modem type in a backup register to avoid repeated initialization of the UART after a system reset.

This PR also adds a custom DNS client implementation based on LwIP, which is used specifically for SARA-R4 since the latter doesn't support DNS lookup commands.

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/electronserialpipe_hal.cpp
+++ b/hal/src/electron/modem/electronserialpipe_hal.cpp
@@ -26,17 +26,98 @@
 #include "gpio_hal.h"
 #include "mdm_hal.h"
 
-static USART_InitTypeDef USART_InitStructure;
-
 ElectronSerialPipe::ElectronSerialPipe(int rxSize, int txSize) :
     _pipeRx( rxSize ),
     _pipeTx( txSize ),
-    pause_(false)
+    baud_(0),
+    pause_(false),
+    hwFlowCtrl_(false)
 {
 }
 
 ElectronSerialPipe::~ElectronSerialPipe(void)
 {
+    end();
+}
+
+void ElectronSerialPipe::begin(unsigned baud, bool hwFlowCtrl)
+{
+    //HAL_USART_Begin(HAL_USART_SERIAL3, baud);
+    //USART_DeInit(USART3);
+    end();
+
+    hwFlowCtrl_ = hwFlowCtrl;
+    baud_ = baud;
+
+    if (hwFlowCtrl) {
+        // Configure USART RTS and CTS as alternate function push-pull
+        HAL_Pin_Mode(RTS_UC, AF_OUTPUT_PUSHPULL);
+        HAL_Pin_Mode(CTS_UC, AF_OUTPUT_PUSHPULL);
+    } else {
+        HAL_Pin_Mode(RTS_UC, OUTPUT);
+        HAL_GPIO_Write(RTS_UC, 0); // VERY IMPORTANT FOR CORRECT OPERATION W/O HW FLOW CONTROL!!
+    }
+    // Configure USART Rx and Tx as alternate function push-pull, and enable GPIOA clock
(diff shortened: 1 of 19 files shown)


hal/src/electron/modem/electronserialpipe_hal.cpp
```diff
@@ -135,6 +112,52 @@ void ElectronSerialPipe::begin(unsigned int baud)
...
+{
+    if (baud_ == 0) {
+        return;
+    }
+
+    // wait for transmission of outgoing data
+    while (_pipeTx.readable())
+    {
+        char c = _pipeTx.getc();
+        USART_SendData(USART3, c);
```
avtolstoy: This will overflow the TX buffer in USART peripheral. I see two possible options here:
```c++
txStart();
while (_pipeTx.readable()) {
  os_thread_yield();
}
```

Or if the idea is not to use interrupts here this loop should instead be transformed into:
```c++
USART_ITConfig(USART3, USART_IT_TXE, DISABLE);
while (_pipeTx.readable()) {
  txCopy();
}
```

hal/src/electron/modem/electronserialpipe_hal.cpp
```diff
@@ -26,59 +26,37 @@
...
-    HAL_Pin_Mode(RTS_UC, AF_OUTPUT_PUSHPULL);
-    HAL_Pin_Mode(CTS_UC, AF_OUTPUT_PUSHPULL);
-#endif
+    if (hwFlowCtrl) {
+        // Configure USART RTS and CTS as alternate function push-pull
+        HAL_Pin_Mode(RTS_UC, AF_OUTPUT_PUSHPULL);
+        HAL_Pin_Mode(CTS_UC, AF_OUTPUT_PUSHPULL);
+    } else {
+        HAL_Pin_Mode(RTS_UC, OUTPUT);
+        HAL_GPIO_Write(RTS_UC, 0); // VERY IMPORTANT FOR CORRECT OPERATION W/O HW FLOW CONTROL!!
```
avtolstoy: :+1:

sergeuz:
c3a2400cacec1520d894d052505bb1506590bf38 reverts part of the UART initialization logic introduced by dff02e7edf113a0b33b1036cec433ba38e88b098: now the firmware initializes UART with hardware flow control enabled, then performs a runtime modem type detection and, depending on the modem type, reinitializes the UART to disable the flow control. The patch assumes that the modem still keeps the CTS pin in a correct state even if doesn't support the CTS/RTS flow control

sergeuz:
The latest changes look good, although it took some time to understand all the RSRP/RSRQ range conversions :) Well done :+1:

PR #1547, opened 2018-06-09
Bug fixes for SARA_R410 LTE E Series [ch16140]

technobly:
### Problem

- GPRS URCs (CIEV: 9,x) were causing SARA_R410 to disconnect
- Fast OTA is not working correctly for SARA_R410
- Kore SIMs on SARA_R410 experience a UDP NAT timeout after 60s

### Solution

- GPRS URCs disabled on SARA_R410
- disables Fast OTA for SARA_R410
- enables 30 second ping for Kore SIMs on SARA_R410

### Steps to Test

- Test with Kore SIM on SARA_R410 E Series, should exhibit 30 second ping, port 5684 and use Slow OTA updates.
- Test with any SIM on SARA_R410 E Series with 3rd Party credentials API with no Particle.keepAlive() API, should exhibit 23 minute ping, port 5684 and use Slow OTA updates (should connect using custom APN).
- Test with Twilio SIM on U260 Electron, should exhibit 23 minute ping, port 4500 and use Fast OTA updates.
- Test with Telefonica SIM on U260 Electron, should exhibit 23 minute ping, port 5684 and use Fast OTA updates.

---

### Completeness

- [x] Added documentation (N/A)

### DIFF
--- a/communication/src/chunked_transfer.cpp
+++ b/communication/src/chunked_transfer.cpp
@@ -33,6 +33,16 @@ ProtocolError ChunkedTransfer::handle_update_begin(
     if (actual_len >= 20 && queue[7] == 0xFF)
     {
         flags = decode_uint8(queue + 8);
+
+        if (fast_ota_override) {
+            if (fast_ota_value) {
+                flags |= (1<<0); // enabled
+            } else {
+                flags &= ~(1<<0); // disabled
+            }
+            DEBUG("Fast OTA: %s", fast_ota_value?"enabled":"disabled");
+        }
+
         file.chunk_size = decode_uint16(queue + 9);
         file.file_length = decode_uint32(queue + 11);
         file.store = FileTransfer::Store::Enum(decode_uint8(queue + 15));
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -318,7 +318,9 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     DEBUG_D("New SMS at index %d\r\n", a);
                     if (sms_cb) SMSreceived(a);
                 }
-                else if ((sscanf(cmd, "CIEV: 9,%d", &a) == 1)) {
+                // TODO: This should include other LTE devices,
+                // perhaps adding _net.act < 7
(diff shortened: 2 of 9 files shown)


communication/src/chunked_transfer.cpp
```diff
@@ -33,6 +33,18 @@ ProtocolError ChunkedTransfer::handle_update_begin(
     if (actual_len >= 20 && queue[7] == 0xFF)
     {
         flags = decode_uint8(queue + 8);
+
+#if PLATFORM_ID == 10
```
avtolstoy: Could we possibly do this in a more generic way? Bringing platform-specific stuff into communication library doesn't sound like a good idea.

One possible approach I can think of from the top of my head is adding a new protocol command (`ProtocolCommands` / `spark_protocol_command`) that would allow to force `ChunkedTransfer` instance to use non-Fast-OTA mode. Having such a command available we could call it either from cellular_hal, once device type is detected, or at least from the `system`.

technobly: This is only a temporary solution while we spend some more time troubleshooting Fast OTA on LTE.  We could also evolve to a more generic strategy if we determine Slow OTA has to stay.

technobly: I ended up moving this code to system_task.cpp and pushing the FAST_OTA override into the communications library as we discussed.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -318,7 +318,9 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     DEBUG_D("New SMS at index %d\r\n", a);
                     if (sms_cb) SMSreceived(a);
                 }
-                else if ((sscanf(cmd, "CIEV: 9,%d", &a) == 1)) {
+                // TODO: This should include other LTE devices,
+                // perhaps adding _net.act < 7
+                else if ((_dev.dev != DEV_SARA_R410) && (sscanf(cmd, "CIEV: 9,%d", &a) == 1)) {
```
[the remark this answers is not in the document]
technobly: My intention here was that if the device is the `SARA_R410`, it will evaluate to `false` and short-circuit the rest of the `&&` expression, not running the `sscanf()`.  As far as I can tell it works because I don't see the `CIEV matched: 9,x` debugging messages with `SARA_R410`.

Issue #1570, opened 2018-08-27
system event network_credentials not triggered via firmware 


fitchn:
<details>
  <summary><i>submission notes</i></summary>

```
If wifi credentials are added or cleared in firmware (e.g. WiFi.clearCredentials()), this system event is not generated/triggered.  Clearing the wifi credentials using a long hold of the setup button does trigger this event.  
```
</details>

### Expected Behavior

Calling WiFi.clearCredentials() should trigger the network_credentials event.  

### Observed Behavior

Nothing!

### Steps to Reproduce

P1 on 0.7.0, asking system to do some logging and reset when this event is triggered.  Never happens.  

### Test App

```
void credsChangedHandler() {
    Serial.println("creds changed");
    System.reset();
}

SYSTEM_MODE(AUTOMATIC);

// in automatic mode, setup is blocked until connection to cloud is made
void setup() {
    Serial.begin(9600);
    Serial.println("starting");
    System.on(network_credentials, credsChangedHandler);
}

void loop() {
    delay(1000);
    Serial.println("clearing creds");
    WiFi.clearCredentials();
}
```

PR #1580, opened 2018-10-09
Fix WiFi credentials clear event not being fired

garzdin:
### Problem

A WiFi credentials event should be fired when a user calls WiFi.clear_credentials(). It doesn't currently happen.

### Solution

A system event is triggered, when a user clears WiFi credentials.

### Example App

```c
void credentialsChanged() {
    Serial.println("Credentials changed");
}

SYSTEM_MODE(AUTOMATIC);

void setup() {
    Serial.begin(9600);
    System.on(network_credentials, credentialsChanged);
}

void loop() {
    Serial.println("Clearing credentials");
    WiFi.clearCredentials();
    delay(1000);
}
```

### References

[#1570](https://github.com/particle-iot/firmware/issues/1570)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_network_wifi.h
+++ b/system/src/system_network_wifi.h
@@ -184,6 +184,8 @@ public:
 
     bool clear_credentials() override
     {
+        system_notify_event(network_credentials, network_credentials_cleared);
+
         return wlan_clear_credentials() == 0;
     }
 


system/src/system_network_wifi.h
```diff
@@ -184,6 +184,8 @@ class WiFiNetworkInterface : public ManagedIPNetworkInterface<WLanConfig, WiFiNe
 
     bool clear_credentials() override
     {
+        system_notify_event(network_credentials, network_credentials_cleared);
```
m-mcgowan: I feel I would like to make this conditional on the result of `wlan_clear_credentials`.  We should also check if other Radio interfaces need a similar change.  It might be better to add it to the `network_clear_credentials()` function so that it is ubiquitous?

PR #1581, opened 2018-10-25
[Electron/LTE] Power Manager Watchdog Timer Fix

### LINKED ISSUE #1548: Electron power-on connectivity issues with 0.8.0-rc
### Expected Behavior

Stay connected without an external battery.

### Observed Behavior

Upgrading to 0.8.0-rc7 has made it so our units no longer stay connected when booted off the power supply. We are using a regulated lab 5V supply rated well in excess of 5A.

This was reproduced across multiple Electrons.

Another user reported similar issues with 0.8.0-rc3.

### Steps to Reproduce

The Electrons boot and stay connected when a battery is present; but without the battery, on the first boot they do not stay connected. What’s weird is that they do complete the handshake and deliver an initial diagnostics message (we can see it in the console), but then they go silent and become unreachable. The cyan LED starts blinking rapidly.

It’s also weird that the battery can be removed and then the Electron rebooted without reproducing this problem. It really is only on initial power-on.

Using particle device doctor to downgrade to 0.7.0 solves the problem.

This occurs whether the firmware was compiled for 0.7.0 or 0.8.0-rc4 (-rc7 is not available right now as an option in Particle Dev).

This is using a 3rd party AT&T SIM, with KeepAlive set at 60 seconds.

### Test App

None yet.

### References

https://community.particle.io/t/electron-power-on-connectivity-issues-with-0-8-0-rc/42391/

technobly:
### Problem

Modem on Electron/LTE turn off unexpectedly during initialization when powered from a USB Host only (no battery).

After an investigation it was discovered that the PMIC watchdog causes current limit to drop to 100mA temporarily which causes the modem to stop responding to AT commands.

### Solution

Power Manager no longer using PMIC watchdog because it causes current limit to drop to 100mA temporarily when it expires

**Workaround:** For 0.8.0-rc.1 ~ 0.8.0-rc.11, add `PMIC().disableCharging()` to `setup()` when no battery is used.

### Steps to Test

1. Program an Electron with this PR's firmware
2. Leave the battery disconnected, powered via USB only
3. The system will check for a disconnected battery every 40 seconds
4. The device should stay connected, AT commands should continue to work.
5. Compared to 0.8.0-rc.11 system firmware, after 40 seconds of the battery being disconnected, the PMIC watchdog causes current limit to drop to 100mA temporarily which causes the modem to stop responding to AT commands.

### References

Fixes #1548 

---

### Completeness

### DIFF
--- a/system/src/system_power.cpp
+++ b/system/src/system_power.cpp
@@ -16,6 +16,9 @@ BatteryChargeDiagnosticData::BatteryChargeDiagnosticData(uint16_t id, const char
 }
 
 int BatteryChargeDiagnosticData::get(IntType& val) {
+    if (g_batteryState == BATTERY_STATE_DISCONNECTED) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
     FuelGauge fuel(true);
     float soc = fuel.getNormalizedSoC();
     val = particle::FixedPointUQ<8, 8>(soc);
--- a/system/src/system_power.cpp
+++ b/system/src/system_power.cpp
@@ -42,6 +42,9 @@ void system_power_management_sleep(bool sleep) {
 
 #else /* Wiring_Cellular != 1 */
 
+void system_power_management_init() {
+}
+
 void system_power_management_sleep(bool sleep) {
 }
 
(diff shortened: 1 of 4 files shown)


elijthomas:
This was an issue for a customer identified on ALL there LTE devices without batteries, PMIC().disableCharging() (as mentioned above) didn't fix the issue but PMIC.disableWatchdog(); IS the fix that allows these units to continue to send and receive AT commands.

PR #1587, opened 2018-11-06
[Mesh] build speedup

avtolstoy:
### Problem

Build process takes too much time. Even with a completely built firmware, rerun of `make` takes at least 8+ seconds.

### Solution
This PR:
1. Removes recursive inclusion of `include.mk` in `third_party` dependencies. There is only a single `include.mk` in a known location.
2. Disables implicit builtin rules

### References

- [CH22668]

---

### Completeness

- [ ] Added documentation


m-mcgowan:
I was curious about the built-in rules. Disabling them doesn't seem to do any harm since we define our own rules for compiling and linking.  https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

PR #1588, opened 2018-11-06
Feature/hal/spi slave

YutingYou:
### Problem

Implement SPI Slave

### Steps to Test

1. Change `TEST_MASTER_APP` to `1`, flash the application to Xenon-SPI-Master device
2. Change `TEST_MASTER_APP` to `0`, flash the application to Xenon-SPI-Slave device
3. Connect serial dongle to TX pin on Xenon to observe the log, SPI master will send and receive data from 1 byte to 10 bytes.

### TODOs
1. D2 D3 D4 are assigned to SPI1, but they are used by ethernet wing, we could change these pins to A0, A1, A2
2. SPI3 doesn’t support SPI slave mode and only SPI3 supports up to 32MHz clock
3. This SPI driver **CANNOT** pass `wiring` test. On STM32 we can set slave buffer after CS pin is pulled low. But on nRF52, we have to set SPI slave buffer before the transition started.
https://devzone.nordicsemi.com/f/nordic-q-a/39387/spis-set-buffer

### Example App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);
Serial1LogHandler   logHandler(115200, LOG_LEVEL_ALL);

#define CS_PIN              A5

#define TEST_MASTER_APP     1

#if TEST_MASTER_APP
volatile uint8_t tx_buffer[100] = {'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I'};
volatile uint8_t rx_buffer[100] = {0};
volatile static uint32_t transfer_state = 0x00;

void onTransferFinished() {
    transfer_state = 1;
    // Log.printf(">> trans done!\r\n");
}

void setup() {

}

void loop() {
    static int tx_counter = 0;
    SPI1.begin(SPI_MODE_MASTER, CS_PIN);
    SPI1.setDataMode(SPI_MODE3);
    Log.printf("\r\n");
    digitalWrite(CS_PIN, 0);
    transfer_state = 0;
    SPI1.transfer((void *)tx_buffer, (void *)rx_buffer, tx_counter++ % 10 + 1, onTransferFinished);
    while(transfer_state == 0) {
        ;
    }
    digitalWrite(CS_PIN, 1);
    SPI1.end();
    Log.printf("master read: %s\r\n", rx_buffer);
    if (SPI1.available() > 0) {
        Log.printf("Received %d bytes ", SPI1.available());
        for (int i = 0; i < SPI1.available(); i++) {
            Log.printf("%02x ", rx_buffer[i]);
        }
        Log.printf("\r\n");
    }
    memset((void *)rx_buffer, 0, 100);
    
    delay(1000);
}

#else

volatile uint8_t tx_buffer[100] = {'1', '2', '3', '4', '5', '6', '7', '8', '9'};
volatile uint8_t rx_buffer[100] = {0};
static volatile uint32_t select_state = 0x00;
static volatile uint32_t transfer_state = 0x00;

void onTransferFinished() {
    transfer_state = 1;
    Log.printf(">> trans done!\r\n");
}

void onSelect(uint8_t state) {
    if (state) {
        select_state = state;
    }

    Log.printf(">> cs pin: %d!\r\n", state);
}

void setup() {
    SPI1.onSelect(onSelect);
    SPI1.begin(SPI_MODE_SLAVE, CS_PIN);

    // TODO: add a function to set tx rx buffer for SPI slave
    SPI1.transfer((void *)tx_buffer, (void *)rx_buffer, sizeof(rx_buffer), onTransferFinished);
}

void loop() {
    while (1) {
        delay(1000);
        if (SPI1.available() > 0) {
            Log.printf("Received %d bytes ", SPI1.available());
            for (int i = 0; i < SPI1.available(); i++) {
                Log.printf("%02x ", rx_buffer[i]);
            }
            Log.printf("\r\n");
            memset((void *)rx_buffer, 0, 100);
        }
    }
}
#endif
```

### References

[CH19818]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/spi_hal.h
+++ b/hal/inc/spi_hal.h
@@ -123,7 +123,6 @@ int32_t HAL_SPI_DMA_Transfer_Status(HAL_SPI_Interface spi, HAL_SPI_TransferStatu
 // HAL_SPI_Set_Bit_Order, HAL_SPI_Set_Data_Mode and HAL_SPI_Set_Clock_Divider in one go
 // to avoid having to reconfigure SPI peripheral 3 times
 int32_t HAL_SPI_Set_Settings(HAL_SPI_Interface spi, uint8_t set_default, uint8_t clockdiv, uint8_t order, uint8_t mode, void* reserved);
-int32_t HAL_SPI_Set_Slave_Buffers(HAL_SPI_Interface spi, void* tx_buffer, void* rx_buffer, uint32_t length);
 
 int32_t HAL_SPI_Acquire(HAL_SPI_Interface spi, void* reserved);
 int32_t HAL_SPI_Release(HAL_SPI_Interface spi, void* reserved);
--- a/hal/inc/spi_hal.h
+++ b/hal/inc/spi_hal.h
@@ -123,6 +123,7 @@ int32_t HAL_SPI_DMA_Transfer_Status(HAL_SPI_Interface spi, HAL_SPI_TransferStatu
 // HAL_SPI_Set_Bit_Order, HAL_SPI_Set_Data_Mode and HAL_SPI_Set_Clock_Divider in one go
 // to avoid having to reconfigure SPI peripheral 3 times
 int32_t HAL_SPI_Set_Settings(HAL_SPI_Interface spi, uint8_t set_default, uint8_t clockdiv, uint8_t order, uint8_t mode, void* reserved);
+int32_t HAL_SPI_Set_Slave_Buffers(HAL_SPI_Interface spi, void* tx_buffer, void* rx_buffer, uint32_t length);
 
 int32_t HAL_SPI_Acquire(HAL_SPI_Interface spi, void* reserved);
 int32_t HAL_SPI_Release(HAL_SPI_Interface spi, void* reserved);
--- a/hal/src/nRF52840/spi_hal.cpp
+++ b/hal/src/nRF52840/spi_hal.cpp
@@ -89,6 +89,7 @@ void spi_slave_event_handler(nrfx_spis_evt_t const * p_event, void * p_context) 
 
     if (p_event->evt_type == NRFX_SPIS_XFER_DONE) {
         m_spi_map[spi].transfer_length = p_event->rx_amount;
+        m_spi_map[spi].transmitting = false;
 
         if (p_event->rx_amount) {
             if (m_spi_map[spi].spi_dma_user_callback) {
@@ -296,7 +297,18 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
         spi_uninit(spi);
     }
 
-    m_spi_map[spi].ss_pin = (pin == SPI_DEFAULT_SS) ? PIN_INVALID : pin;
+    if (pin == SPI_DEFAULT_SS) {
+        if (spi == HAL_SPI_INTERFACE1) {
+            m_spi_map[spi].ss_pin = SS;
+        } else if (spi == HAL_SPI_INTERFACE2) {
+            m_spi_map[spi].ss_pin = D5;
+        } else {
+            m_spi_map[spi].ss_pin = PIN_INVALID;
+        }
+    } else {
+        m_spi_map[spi].ss_pin = pin;
+    }
+
     m_spi_map[spi].spi_mode = mode;
     spi_init(spi, mode);
     m_spi_map[spi].enabled = true;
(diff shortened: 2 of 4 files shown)


avtolstoy:
> This SPI driver CANNOT pass wiring test. On STM32 we can set slave buffer after CS pin is pulled low. But on nRF52, we have to set SPI slave buffer before the transition started. https://devzone.nordicsemi.com/f/nordic-q-a/39387/spis-set-buffer

I'm fine with this platform-specific caveat of requiring to initiate `transfer()` on slave before `onSelect()` event and leaving `onSelect()` informational only, as long as we properly document this. However, this is not the main issue preventing the test from passing. `transfer_ongoing` is not being correctly reported in `HAL_SPI_DMA_Transfer_Status()` while in slave mode. With the following patch everything works as expected.

```
diff --git a/hal/src/nRF52840/spi_hal.cpp b/hal/src/nRF52840/spi_hal.cpp
index 26884f1e2..8adb2fe56 100644
--- a/hal/src/nRF52840/spi_hal.cpp
+++ b/hal/src/nRF52840/spi_hal.cpp
@@ -90,6 +90,8 @@ void spi_slave_event_handler(nrfx_spis_evt_t const * p_event, void * p_context)
     if (p_event->evt_type == NRFX_SPIS_XFER_DONE) {
         m_spi_map[spi].transfer_length = p_event->rx_amount;
 
+        m_spi_map[spi].transmitting = false;
+
         if (p_event->rx_amount) {
             if (m_spi_map[spi].spi_dma_user_callback) {
                 (*m_spi_map[spi].spi_dma_user_callback)();
@@ -484,6 +486,7 @@ int32_t HAL_SPI_Set_Slave_Buffers(HAL_SPI_Interface spi, void* tx_buffer, void*
         // LOG_DEBUG(WARN, "nrfx_spis_buffers_set, invalid state");
     } else {
         SPARK_ASSERT(err_code == NRF_SUCCESS);
+        m_spi_map[spi].transmitting = true;
     }
 
     return 0;
diff --git a/user/tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp b/user/tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp
index 647b56be3..80851b53f 100644
--- a/user/tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp
+++ b/user/tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp
@@ -119,9 +119,11 @@ test(SPI_Master_Slave_Slave_Transfer)
         /* Preload SLAVE_TEST_MESSAGE_1 reply */
         memcpy(SPI_Slave_Tx_Buffer, SLAVE_TEST_MESSAGE_1, sizeof(SLAVE_TEST_MESSAGE_1));
 
-        /* Wait for Master to select us */
-        while(SPI_Selected_State == 0);
-        SPI_Selected_State = 0;
+        /* IMPORTANT: NOT waiting for Master to select us, as some of the platforms
+         * require TX and RX buffers to be configured before CS goes low
+         */
+        // while(SPI_Selected_State == 0);
+        // SPI_Selected_State = 0;
 
         /* Receive up to TRANSFER_LENGTH_2 bytes */
         SPI_Transfer_DMA(SPI_Slave_Tx_Buffer, SPI_Slave_Rx_Buffer, TRANSFER_LENGTH_2, count % 2 == 0 ? &SPI_DMA_Completed_Callback : NULL);
@@ -141,7 +143,7 @@ test(SPI_Master_Slave_Slave_Transfer)
             if (((switchMode >> 16) & 0xffff) == 0x8DC6) {
                 uint8_t mode = switchMode & 0xff;
                 uint8_t bitOrder = (switchMode >> 8) & 0xff;
-                
+
                 assertMoreOrEqual(mode, SPI_MODE0);
                 assertLessOrEqual(mode, SPI_MODE3);
 
```

Master: `TEST=wiring/spi_master_slave/spi_master USE_SPI=SPI1`
Slave:  `TEST=wiring/spi_master_slave/spi_slave USE_SPI=SPI1`

hal/inc/spi_hal.h
```diff
@@ -123,6 +123,7 @@ int32_t HAL_SPI_DMA_Transfer_Status(HAL_SPI_Interface spi, HAL_SPI_TransferStatu
 // HAL_SPI_Set_Bit_Order, HAL_SPI_Set_Data_Mode and HAL_SPI_Set_Clock_Divider in one go
 // to avoid having to reconfigure SPI peripheral 3 times
 int32_t HAL_SPI_Set_Settings(HAL_SPI_Interface spi, uint8_t set_default, uint8_t clockdiv, uint8_t order, uint8_t mode, void* reserved);
+int32_t HAL_SPI_Set_Slave_Buffers(HAL_SPI_Interface spi, void* tx_buffer, void* rx_buffer, uint32_t length);
```
avtolstoy: In light of https://github.com/particle-iot/firmware/pull/1588#issuecomment-436582634 I don't think there is any need in a separate function.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -205,210 +274,216 @@ void HAL_SPI_Init(HAL_SPI_Interface spi)
...
-    if (m_spi_map[spi].enabled)
-    {
+    if (m_spi_map[spi].enabled) {
         spi_uninit(spi);
     }
 
-    m_spi_map[spi].ss_pin = pin;
-
-    // TODO: need to support spi slave
+    m_spi_map[spi].ss_pin = (pin == SPI_DEFAULT_SS) ? PIN_INVALID : pin;
```
avtolstoy: This needs to work correctly. If `SPI_DEFAULT_SS`, the SS pin needs to be set to `SS` for `HAL_SPI_INTERFACE1` and `D5` for `HAL_SPI_INTERFACE2`.

avtolstoy:
> D2 D3 D4 are assigned to SPI1, but they are used by ethernet wing, we could change these pins to A0, A1, A2
In this case we should add `HAL_SPI_INTERFACE3` same as on Electron.

hal/inc/spi_hal.h
YutingYou: Thanks for the fix, finally got it passed unit test! Have removed `HAL_SPI_Set_Slave_Buffers `.

hal/src/nRF52840/spi_hal.cpp
YutingYou: Fixed.

YutingYou:
If we add `HAL_SPI_INTERFACE3` then we need to consider the resource conflict with TWI. We have 4 peripherals for SPI and TWI, SPI gets two and TWI gets two.

avtolstoy:
Correct, the conflict will need to be handled. We can leave this out of the scope of this PR though.

PR #1591, opened 2018-11-07
Implement I2C Slave

YutingYou:
### Problem

Implement I2C slave

### Steps to Test

1. Flash `firmware-private/user/tests/wiring/i2c_master_slave/i2c_master` to Xenon-Master
2. Flash `firmware-private/user/tests/wiring/i2c_master_slave/i2c_slave` to Xenon-Slave
3. Watch the output from terminal

### Test Result
Overview:
image

Master Write:
image

Master Read:
image

> Note: The low level delay is clock stretch test

Output log:
image

### References

[CH19818]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/i2c_hal.h
+++ b/hal/inc/i2c_hal.h
@@ -40,12 +40,8 @@ typedef enum
 /* Exported types ------------------------------------------------------------*/
 typedef enum HAL_I2C_Interface {
     HAL_I2C_INTERFACE1 = 0,
-#if PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON
-    HAL_I2C_INTERFACE2 = 1
-#else
     HAL_I2C_INTERFACE2 = 1,
     HAL_I2C_INTERFACE3 = 2
-#endif
 } HAL_I2C_Interface;
 
 /* Exported constants --------------------------------------------------------*/
--- a/hal/inc/i2c_hal.h
+++ b/hal/inc/i2c_hal.h
@@ -29,6 +29,7 @@
 
 /* Includes ------------------------------------------------------------------*/
 #include "pinmap_hal.h"
+#include "platforms.h"
 
 /* Exported types ------------------------------------------------------------*/
 typedef enum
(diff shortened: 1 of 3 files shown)


hal/inc/i2c_hal.h
```diff
@@ -39,8 +40,12 @@ typedef enum
 /* Exported types ------------------------------------------------------------*/
 typedef enum HAL_I2C_Interface {
     HAL_I2C_INTERFACE1 = 0,
+#if PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON
```
avtolstoy: This is a rebase artifact. There should no longer be a conditional here.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -54,37 +59,80 @@ typedef struct {
 
     uint8_t                     address;    // 7bit data
 
-    uint8_t                     rx_buf[BUFFER_LENGTH];
-    uint8_t                     rx_buf_index;
-    uint8_t                     rx_buf_length;
-    uint8_t                     tx_buf[BUFFER_LENGTH];
-    uint8_t                     tx_buf_index;
-    uint8_t                     tx_buf_length;
+    volatile uint8_t            rx_buf[BUFFER_LENGTH];
```
avtolstoy: I don't see any reason to make all of these volatile. In slave mode, they are going to be accessed only from interrupt context in `onRequest()` and `onReceive()` callbacks.

hal/inc/i2c_hal.h
YutingYou: Do you mean we've define all I2C instance in `hal_platform.h`, so `HAL_I2C_Init()` won't get nonexistent parameter such as `HAL_I2C_INTERFACE3`?

hal/src/nRF52840/i2c_hal.cpp
YutingYou: Removed `volatile` for `rx_buf_index` and `rx_buf_length`

hal/inc/i2c_hal.h
avtolstoy: It shouldn't, because we are correctly defining capabilities in `spark_wiring_platform.h` based on `HAL_PLATFORM_I2CX` defines in `hal_platform.h`

YutingYou: That make sense.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -60,8 +60,8 @@ typedef struct {
     uint8_t                     address;    // 7bit data
 
     volatile uint8_t            rx_buf[BUFFER_LENGTH];
-    volatile uint8_t            rx_buf_index;
-    volatile uint8_t            rx_buf_length;
+    uint8_t                     rx_buf_index;
+    uint8_t                     rx_buf_length;
     volatile uint8_t            tx_buf[BUFFER_LENGTH];
```
avtolstoy: I meant that we don't really need `volatile` for all of `rx_buf`, `rx_buf_index`, `rx_buf_length`, `tx_buf`, `tx_buf_index`, `tx_buf_length`.

YutingYou: I see, we use block sending, and the buffers seem not to be optimized. Have removed.

PR #1597, opened 2018-11-13
Add BLE bare HAL APIs

XuGuohui:
The file name and API name may be changed once we have finished the BLE HAL APIs set that can fully support the BLE control request in the listening mode. Just want to make the BLE control request work using the pre-existing BLE HAL APIs before that happens.


filipnavara:
Is there going to be new pull request or some other place I can track?

XuGuohui:
Yup. I just open a new pull request that contains both BLE central and peripheral features.

filipnavara:
Great, thanks!

PR #1598, opened 2018-11-13
system event on isr

m-mcgowan:
### Problem

The malloc implementation in release 0.8.0-rc.11 includes an explicit check to ensure malloc is not called from an ISR. 

The button events are sometimes handled from an ISR, and consequently when the system event is marshaled to the application thread (using malloc) the assertion fails and the system enters a SOS. 

### Solution

The button handlers use a new function `system_notify_event_isr` instead of `system_notify_event`.  The new ISR-aware function checks if the call is actually from an ISR - if not, it simply delegates to the original `system_notify_event`.  Otherwise, the event parameters are stored in a Task that is enqueued on the System ISR queue and the system event then later notified from the system thread.  This ensures that malloc is not called from the button handler ISR. 

### Steps to Test

[CH24165](https://app.clubhouse.io/particle/story/24165) contains test applications to trigger the issue.  After applying this PR the SOS no longer occurs. 

The example app below demonstrates that button clicks are still correctly handled on the application thread.  Note that the delay in the application thread means the button handlers sometimes take some time to execute.  

### Example App

```c

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

void setup()
{
  Particle.connect();
  System.on(button_click, []() {
	 RGB.control(true);
	 RGB.color(255,128, 0);
	 delay(500);
	 RGB.control(false);
  });
}

void loop()
{
	delay(1000);
}

```

### References

- [CH24165](https://app.clubhouse.io/particle/story/24165)
---

### Completeness

- [ ] Added documentation

### DIFF
--- a/services/inc/panic_codes.h
+++ b/services/inc/panic_codes.h
@@ -9,21 +9,23 @@
 // followed by 300 blinks of the value of code
 /// ...---...    code   ...---...
 
-def_panic_codes(Faults,RGB_COLOR_RED,HardFault)
-def_panic_codes(Faults,RGB_COLOR_RED,NMIFault)
-def_panic_codes(Faults,RGB_COLOR_RED,MemManage)
-def_panic_codes(Faults,RGB_COLOR_RED,BusFault)
-def_panic_codes(Faults,RGB_COLOR_RED,UsageFault)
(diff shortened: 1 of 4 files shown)


m-mcgowan:
Stale - closing out.

PR #1599, opened 2018-11-14
_CONST was dropped in Newlib 3.0.0

jablko:
### Problem

If eventually the firmware is built with Newlib 3.0.0, then without this change there's an `error: unknown type name '_CONST'` because `_CONST` was [dropped in Newlib 3.0.0](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;a=commit;h=0bda30e1ffd23488aa4a9b73f228089463fbee1a).

### Solution

This patch follows the same pattern as the change in Newlib and replaces `_CONST` with `const`. Building with previous version of Newlib is unaffected. The firmware already employs the `const` type qualifier.

### Completeness

- [ ] Added documentation

### DIFF
--- a/services/inc/printf_export.h
+++ b/services/inc/printf_export.h
@@ -51,12 +51,12 @@ _printf_float (struct _reent *data,
            struct _prt_data_t *pdata,
            FILE *fp,
            int (*pfunc)(struct _reent *, FILE *,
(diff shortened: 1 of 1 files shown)


jablko:
Thanks!

PR #1600, opened 2018-11-14
fixes SOS in 0.8.0-rc.11

m-mcgowan:
Backport of PR #1598 for non-mesh hardware.

---

### Completeness

- [x] Added documentation

### DIFF
--- a/system/src/system_event.cpp
+++ b/system/src/system_event.cpp
@@ -24,6 +24,7 @@
 #include <stdint.h>
 #include <vector>
 
+namespace {
 
 struct SystemEventSubscription {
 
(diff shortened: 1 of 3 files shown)


sergeuz:
I agree with @jlovinger-particle. It's a little bit confusing that `system_notify_event_isr()`, which is a more specialized function, can be called in any context, while `system_notify_event()` can be used only in a thread. I'll refactor this PR to make `system_notify_event()` work both in an ISR or thread if nobody minds

sergeuz:
@jlovinger-particle Could you please re-review this PR?

m-mcgowan:
The intent is that `system_notify_event()` can never be on an ISR, while `send_system_event_isr()` may be on an ISR so that the known context is explicitly called out rather than being left to guesswork. I suppose it's ok to change the logic - the isISR() check is presumably fairly performant, so could be a case of premature optimization if that were the only consideration. Also I wanted to go for minimum change. If this isn't clear, happy to have `system_notify_event()` work for all cases.

When coding this I wanted a way to have functions describe their calling context (ISR/thread), I didn't come up with anything, but would be great to have the compiler check that only ISR-allowed functions are called from an ISR context.  Could perhaps be a custom extension with some special attributes defined on ISR-allowed functions?

jlovinger-particle:
Whatever you guys decide on I'm sure will be OK and all will be equally functional. Improvement on the status quo.

We're kind of getting into the weeds because system_notify_event with this Pull request hides three kinds of behaviors.

Defer initial execution into System thread if running from an interrupt.
Defer final execution into Application thread if running from the System Thread.
Execute immediately if running from the Application Thread (but not an ISR!).

Without really ripping it apart or refactoring it will be hard to make explicitly clear the execution paths without also parsing through the code. For example...

1. ISR interrupts Application Thread and triggers an event
2. Defer event into the System Thread
3. Drop back into the Application Thread
4. Preempt/etc into the System Thread
5. Begin processing event in the System Thread
6. Defer event into the Application Thread
7. Preempt/etc into the Application Thread
8. Process event

None of which is made clear by the old/new names of system_notify_event/system_notify_event_isr OR the new/new with system_notify_event/system_notify_impl.

And on top of that, previously it would have just been...

1. ISR interrupts Application Thread and triggers an event
2. Process event immediately since detects Application Thread context.

Which was likely bad in its own way as the ISR would have been executing like it *was* the Application Thread introducing all sorts of concurrency issues I wonder if people were handling. But if they were it sure would have been responsive...

Unexpected complexity, for sure.

sergeuz:
@m-mcgowan My reasoning is that this is the system code, which is (or should be) pretty well abstracted from the low-level platform code. If some handler runs in an ISR it doesn't mean that it will be running in an ISR on another platform as well, and by hiding these details we're improving portability of the system code. Also, most of our current APIs perform the ISR check internally rather than provide separate functions that can be used in ISRs

sergeuz:
@jlovinger-particle That's correct. Ideally, there should be a separate `ISRTaskQueue` for the application thread. I think that the current solution is fine for now though

sergeuz:
@m-mcgowan I doubt it's possible to determine the context of a function at compile time, but what we can do is to introduce a custom attribute (e.g. "isr_safe") and require that all functions called from such an ISR-safe function should also be ISR-safe – similarly to how const methods work in C++

jlovinger-particle:
> introduce a custom attribute (e.g. "isr_safe") and require that all functions called from such an ISR-safe function should also be ISR-safe

Famous last words :)

sergeuz:
@jlovinger-particle Could you please approve this PR if the changes look good to you?

system/src/system_event.cpp
```diff
@@ -87,17 +138,16 @@ void system_unsubscribe_event(system_event_t events, system_event_handler_t* han
...
-    }
-    if (fn)
-        fn(fndata);
+void system_notify_event(system_event_t event, uint32_t data, void* pointer, void (*fn)(void* data), void* fndata) {
+  if (HAL_IsISR()) {
+      void* space = (system_pool_alloc(sizeof(SystemEventTask), nullptr));
+      if (space) {
+          auto task = new (space) SystemEventTask(event, data, pointer, fn, fndata);
+          SystemISRTaskQueue.enqueue(task);
+      };
```
avtolstoy: nit: extra `;`

PR #1604, opened 2018-11-20
Escape special characters in SSIDs and passwords when sending them to ESP32

sergeuz:
### Problem

Argon fails to connect to a WiFi network if the network's password contains special characters (`,`, `"`, `\`).

### Solution

Escape special characters in SSIDs and passwords as described in the AT command set manual (see the `+CWJAP` command).

### Steps to Test

- Change the password of your network to some string containing special characters, and make sure an Argon is able to connect to the network.

- Run [unit tests](https://github.com/particle-iot/firmware/tree/fix/escape_ssid/test/unit_tests).

### Example App

N/A

### References

- [ESP32 AT command set manual](
https://www.espressif.com/sites/default/files/documentation/ESP32_AT_Instruction_Set_and_Examples__EN.pdf)

---

### Completeness

- [ ] Added documentation


services/src/str_util.cpp
```diff
@@ -0,0 +1,38 @@
...
+namespace particle {
+
+char* escape(const char* src, size_t srcSize, const char* spec, size_t specSize, char esc, char* dest, size_t destSize) {
+    size_t i = 0;
+    size_t j = 0;
+    while (i < srcSize && j < destSize) {
+        const char c = src[i++];
+        if (memchr(spec, (unsigned char)c, specSize) != nullptr) {
+            if (j == destSize - 1) {
+                break; // Avoid generating an invalid escaped string
```
m-mcgowan: How can the caller know the buffer was too small?

services/src/str_util.cpp
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "str_util.h"
+
+namespace particle {
+
+char* escape(const char* src, size_t srcSize, const char* spec, size_t specSize, char esc, char* dest, size_t destSize) {
```
m-mcgowan: Could you add unit tests for this?

services/inc/str_util.h
```diff
@@ -87,4 +87,34 @@ inline bool endsWith(const char* str, const char* suffix) {
...
+ * @param esc Escape character.
+ * @param dest Destination buffer.
+ * @param destSize Size of the destination buffer.
+ *
+ * @return Destination buffer.
+ */
+char* escape(const char* src, size_t srcSize, const char* spec, size_t specSize, char esc, char* dest, size_t destSize);
+
+/**
+ * Escapes a set of characters by prepending them with an escape character.
```
m-mcgowan: Nice to see the method is well documented.  How does the caller know how many characters are added to the buffer. Is the output guaranteed to be null-terminated or only operates on the number of characters given?
 
For methods like this that are relatively simple, no side effects etc, I really liked doctest in python that allows you to write inline tests in documentation.  Seems like there is an equivalent for C++. https://github.com/onqtam/doctest

services/src/str_util.cpp
sergeuz: There's no way to know that. We can make this function fill the entire output buffer and return the number of characters required to store the complete escaped string, which can be larger than the size of the output buffer – similarly to how `printf()` family of functions work. I just wasn't sure if it's necessary, given that the caller always knows the size of the buffer in the worst case (`strlen(src) * 2 + 1`)

sergeuz:
Updated the implementation according to the suggestions and added unit tests: https://github.com/particle-iot/firmware/tree/fix/escape_ssid/test/unit_tests

PR #1607, opened 2018-11-28
Release timer 2, use systick to calculate debounce time

YutingYou:
### Problem

Button driver occupied a high precision to calculate debounce time,  it is wasteful, we can use systick handler to calculate debounce time.

### Steps to Test
1. Flash example application
2. Press the mode button, observe the press count and debounce time.

### Example App

```c
 #include "application.h"
 #include "button_hal.h"
 
SYSTEM_MODE(MANUAL);

Serial1LogHandler logHandler(115200);

static volatile int counter = 0;

void setup() {
  System.on(button_click, [](system_event_t ev, int clicks) -> void {
    counter++;
  });
}

void loop() {
    Log.info("counter: %d, debounced time: %d", counter, BUTTON_GetDebouncedTime(BUTTON1));
    delay(1000);
}
```

### References

---

### Completeness

- [ ] Added documentation


system/src/main.cpp
```diff
@@ -415,6 +415,8 @@ extern "C" void HAL_SysTick_Handler(void)
     }
 #endif
 
+    BUTTON_Timer_Handler();
```
avtolstoy: Please add stub `BUTTON_Timer_Handler()` for all the other platforms.

hal/src/nRF52840/button_hal.c
```diff
@@ -49,14 +39,19 @@ button_config_t HAL_Buttons[] = {
     }
 };
 
+static struct {
```
avtolstoy: Let's make this `volatile` just in case, as it will be accessed from two different ISRs.

system/src/main.cpp
YutingYou: Can I use `if PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON` here?

If I add stub `BUTTON_Timer_Handler()`, I just need to add it to below file, right?
```
platform/MCU/STM32F1xx/SPARK_Firmware_Driver/inc/hw_config.h
platform/MCU/STM32F1xx/SPARK_Firmware_Driver/inc/hw_config.c
platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/hw_config.h
platform/MCU/STM32F2xx/SPARK_Firmware_Driver/inc/hw_config.c
```

hal/src/nRF52840/button_hal.c
YutingYou: Added

system/src/main.cpp
avtolstoy: Or even better let's add a new feature flag to `hal_platform.h` and set it in Gen 3-specific `hal_platform_config.h` and `#if` on it. Let's say `HAL_PLATFORM_BUTTON_DEBOUNCE_IN_SYSTICK`?

YutingYou: Sounds good, better than adding stub function to other platforms. I'll add `HAL_PLATFORM_BUTTON_DEBOUNCE_IN_SYSTICK` to `hal/src/nRF52840/hal_platform_nrf52840_config.h`

hal/src/nRF52840/hal_platform_nrf52840_config.h
```diff
@@ -51,3 +51,5 @@
 #define HAL_PLATFORM_IFAPI (1)
 
 #define HAL_PLATFORM_ETHERNET (1)
+
+#define HAL_PLATFORM_BUTTON_DEBOUNCE_IN_SYSTICK (1)
```
avtolstoy: Please add the default to `hal_platform.h` as well.

YutingYou: Added

avtolstoy:
@YutingYou Please rebase on `mesh-develop`.

avtolstoy:
Superseded by #1590

PR #1609, opened 2018-11-28
[network] nat64: randomize base source port on boot

avtolstoy:
### Problem

There is a known issue in Device Service sometimes preventing Gen 3 devices from connecting to a different node from the same source IP and port pair (see [CH23915]). We've introduced a workaround for this in rc.25 by using ephemeral ports for connections over non-mesh network interfaces.

For connections over mesh we use a fixed source port (5684), because we don't want to easily exhaust the gateway NAT64 translation table, however after the gateway resets, first connection from within the mesh will always get 40000 source port on the outside interface after NAT64, which may or may not be the cause of the same issue we've seen on gateway devices in [CH23915].

### Solution

Randomize initial base NAT64 UDP source port.

### Steps to Test

N/A

### Example App

N/A

### References

- [CH25381]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/nat64.cpp
+++ b/hal/network/lwip/nat64.cpp
@@ -32,6 +32,7 @@ LOG_SOURCE_CATEGORY("net.nat64")
 #include <lwip/inet_chksum.h>
 #include <lwip/timeouts.h>
 #include "lwiplock.h"
+#include "random.h"
 
 using namespace particle::net;
 using namespace particle::net::nat;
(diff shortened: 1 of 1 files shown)


hal/network/lwip/nat64.cpp
```diff
@@ -90,9 +91,9 @@ static_assert(MEMP_NUM_SYS_TIMEOUT > LWIP_NUM_SYS_TIMEOUT_INTERNAL, "An extra ti
 
 } /* anonymous */
 
-Nat64::Nat64()
-        : udpNextPort_(DEFAULT_UDP_NAT_MIN_PORT) {
+Nat64::Nat64() {
     IP6_ADDR(&pref64_, PP_HTONL(0x64ff9b), 0, 0, 0);
+    udpNextPort_ = rand() % (DEFAULT_UDP_NAT_MAX_PORT - DEFAULT_UDP_NAT_MIN_PORT) + DEFAULT_UDP_NAT_MIN_PORT;
```
sergeuz: `rand()` is not thread-safe, let's use `rand_r()` instead

PR #1617, opened 2018-12-03
Fixes user part update procedure in hybrid build

avtolstoy:
### Problem

When flashing a user-part that depends on the newer version of system-part1, hybrid firmware will overwrite that user part with the embedded tinker due to failed dependency check (https://github.com/particle-iot/firmware/blob/mesh-develop/hal/src/nRF52840/core_hal.c#L352), preventing safe mode from triggering and as a consequence preventing safe mode healer from applying newer system-part1 to resolve the dependencies.

### Solution

Flash the embedded tinker only if there is no valid user module or it fails to pass the integrity check. Dependencies shouldn't be checked.

This PR also fixes safe mode LED indication priority issue.

### Steps to Test

1. Build with `HYBRID_BUILD=y INCLUDE_APP=y`
2. Flash `system-part1`
3. Erase user-part
4. Reset
5. The device should boot into tinker
6. Bump module versions, rebuild normal modular firmware + some different app
7. Flash user-part OTA
8. The device should accept the new user firmware, reset and go into safe mode
9. Flash bumped system-part1 OTA
10. The device should accept the new system-part1, reset and go into normal breathing cyan running the app that was flashed at step 8.

### Example App

N/A

### References

- [CH25698]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/core_hal.c
+++ b/hal/src/nRF52840/core_hal.c
@@ -293,7 +293,7 @@ void HAL_Core_Config(void) {
     HAL_RNG_Configuration();
 
     HAL_RTC_Configuration();
-    
+
 #if defined(MODULAR_FIRMWARE)
     if (HAL_Core_Validate_User_Module()) {
         new_heap_end = module_user_pre_init();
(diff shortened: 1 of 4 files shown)


hal/src/nRF52840/core_hal.c
```diff
@@ -304,9 +308,6 @@ void HAL_Core_Config(void) {
     else {
         // Set the heap end to the stack start to make most use of the SRAM.
         malloc_set_heap_end(&_Stack_Init);
-
-        // Try copying the embedded user image to user application address
-        user_update();
```
sergeuz: Any reason not to call `user_update_if_needed()` here? It looks like we'd avoid checking the module's integrity twice on every boot in this case

PR #1623, opened 2018-12-05
Do not set the sticky skip hello after handshake

monkbroc:
### Problem

After OTA the device should send the HELLO message so it gets prompted to send its describe message.

Occasionally after sending the HELLO message, the cloud communications fail. After reconnecting, it does not send the HELLO message again so the version of firmware running on the device is not updated in the database.

### Solution

The correct logic is in `DTLSMessageChannel`, but this sticky flag that gets set before we hear anything back from the cloud in fact breaks it in case the cloud disconnects before the describe message is sent since the HELLO won't be sent after reconnecting.

### Steps to Test

* `$ particle flash tinker-serial1-debugging-0.8.0-rc.25-xenon.bin  --usb`
* Pull power
* Verify in the database that current_build_target is 0.8.0-rc.25
* `$ particle flash x2 system-part1-0.8.0-rc.26-xenon.bin`
* Expect in the database that current_build_target is 0.8.0-rc.26

### Completeness

- [ ] Added documentation

### DIFF
--- a/communication/src/protocol.cpp
+++ b/communication/src/protocol.cpp
@@ -335,7 +335,6 @@ int Protocol::begin()
 	}
 	LOG(INFO,"Handshake completed");
 	channel.notify_established();
-	flags |= SKIP_SESSION_RESUME_HELLO;
 	return error;
 }
 


avtolstoy:
@sergeuz This bugfix should be backported into Gen 2 firmware as well.

PR #1638, opened 2018-12-14
Fixes to the Argon in Listening Mode showing wrong MAC address with the command `m`.

Cheong2K:
### Problem

The Argon shows MAC address with "00:00:00:00:00:00" in Listening Mode with command `m` which is incorrect.

### Solution

Allows to read MAC address from NCP even the interface is not up yet.

### Steps to Test

1. Compile and load system-part1 using DFU to an Argon.
2. Enter Listening Mode and press `m`, it should show the Wi-Fi MAC address correctly.
3. Using `$ particle serial mac` should also show it correctly.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/esp32/esp32ncpnetif.cpp
+++ b/hal/network/lwip/esp32/esp32ncpnetif.cpp
@@ -147,9 +147,7 @@ int Esp32NcpNetif::queryMacAddress() {
 void Esp32NcpNetif::loop(void* arg) {
     Esp32NcpNetif* self = static_cast<Esp32NcpNetif*>(arg);
     unsigned int timeout = 100;
-    auto r = self->queryMacAddress();
-    if (r) // Failed to query MAC Address
-        return;
+    self->queryMacAddress();
     self->wifiMan_->ncpClient()->off();
     while(!self->exit_) {
         self->wifiMan_->ncpClient()->enable(); // Make sure the client is enabled
--- a/hal/network/lwip/esp32/esp32ncpnetif.cpp
+++ b/hal/network/lwip/esp32/esp32ncpnetif.cpp
@@ -124,9 +124,32 @@ err_t Esp32NcpNetif::initInterface() {
     return ERR_OK;
 }
 
+int Esp32NcpNetif::queryMacAddress() {
+    auto r = wifiMan_->ncpClient()->on();
+    if (r) {
+        LOG(TRACE, "Failed to initialize ESP32 NCP client: %d", r);
+        return r;
+    }
+    MacAddress mac = {};
+    if (!memcmp(interface()->hwaddr, mac.data, interface()->hwaddr_len)) {
+        // Query MAC address
+        r = wifiMan_->ncpClient()->getMacAddress(&mac);
+        if (r) {
+            LOG(TRACE, "Failed to query ESP32 MAC address: %d", r);
+            return r;
+        }
+        memcpy(interface()->hwaddr, mac.data, interface()->hwaddr_len);
+    }
+
+    return r;
+}
+
 void Esp32NcpNetif::loop(void* arg) {
     Esp32NcpNetif* self = static_cast<Esp32NcpNetif*>(arg);
     unsigned int timeout = 100;
+    auto r = self->queryMacAddress();
+    if (r) // Failed to query MAC Address
+        return;
     self->wifiMan_->ncpClient()->off();
     while(!self->exit_) {
         self->wifiMan_->ncpClient()->enable(); // Make sure the client is enabled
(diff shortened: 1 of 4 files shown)


system/src/system_setup.cpp
```diff
@@ -317,12 +317,17 @@ template<typename Config> void SystemSetupConsole<Config>::handle(char c)
...
         print("Your device MAC address is\r\n");
+    #if !HAL_PLATFORM_WIFI    
         IPConfig config = {};
         auto conf = static_cast<const IPConfig*>(network_config(0, 0, 0));
         if (conf && conf->size) {
             memcpy(&config, conf, conf->size);
         }
         const uint8_t* addr = config.nw.uaMacAddr;
+    #else
+        auto conf = static_cast<const IPConfig*>(network_config(NETWORK_INTERFACE_WIFI_STA, 0, 0));
```
avtolstoy: You still need to check `conf && conf->size` for this case. Why not:
```cpp
        IPConfig config = {};
#if !HAL_PLATFORM_WIFI
        auto conf = static_cast<const IPConfig*>(network_config(0, 0, 0));
#else
        auto conf = static_cast<const IPConfig*>(network_config(NETWORK_INTERFACE_WIFI_STA, 0, 0));
#endif // HAL_PLATFORM_WIFI
        if (conf && conf->size) {
            memcpy(&config, conf, conf->size);
        }
        const uint8_t* addr = conf->nw.uaMacAddr;
```

hal/src/argon/wlan_hal.cpp
```diff
@@ -192,12 +192,12 @@ int wlan_fetch_ipconfig(WLanConfig* conf) {
     CHECK_TRUE(iface, SYSTEM_ERROR_INVALID_STATE);
     unsigned flags = 0;
     CHECK(if_get_flags(iface, &flags));
-    CHECK_TRUE((flags & IFF_UP) && (flags & IFF_LOWER_UP), SYSTEM_ERROR_INVALID_STATE);
     // MAC address
     sockaddr_ll hwAddr = {};
     CHECK(if_get_lladdr(iface, &hwAddr));
     CHECK_TRUE((size_t)hwAddr.sll_halen == MAC_ADDRESS_SIZE, SYSTEM_ERROR_UNKNOWN);
     memcpy(conf->nw.uaMacAddr, hwAddr.sll_addr, MAC_ADDRESS_SIZE);
+    CHECK_TRUE((flags & IFF_UP) && (flags & IFF_LOWER_UP), SYSTEM_ERROR_INVALID_STATE);
```
avtolstoy: This is not enough. You still need [this](https://github.com/particle-iot/firmware/blob/mesh-develop/hal/network/lwip/esp32/esp32ncpnetif.cpp#L205-L219) block of code to be executed before https://github.com/particle-iot/firmware/blob/mesh-develop/hal/network/lwip/esp32/esp32ncpnetif.cpp#L130 (it should still be left in `upImpl()` as well), otherwise the MAC will only be available after the network interface is brought up at least once. As discussed that block of code should be decoupled into a separate method and should be called from both places.

system/src/system_setup.cpp
Cheong2K: Fixed.

system/src/system_setup.cpp
```diff
@@ -318,11 +318,15 @@ template<typename Config> void SystemSetupConsole<Config>::handle(char c)
...
+    #if !HAL_PLATFORM_WIFI    
         auto conf = static_cast<const IPConfig*>(network_config(0, 0, 0));
+    #else
+        auto conf = static_cast<const IPConfig*>(network_config(NETWORK_INTERFACE_WIFI_STA, 0, 0));
+    #endif
         if (conf && conf->size) {
             memcpy(&config, conf, conf->size);
         }
-        const uint8_t* addr = config.nw.uaMacAddr;
+        const uint8_t* addr = conf->nw.uaMacAddr;
```
avtolstoy: This should be left as `config.`, otherwise if `network_config()` returns `nullptr` you'll have a null pointer dereference here.

hal/src/argon/wlan_hal.cpp
Cheong2K: Added.

hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -124,9 +124,32 @@ err_t Esp32NcpNetif::initInterface() {
...
+
+    return r;
+}
+
 void Esp32NcpNetif::loop(void* arg) {
     Esp32NcpNetif* self = static_cast<Esp32NcpNetif*>(arg);
     unsigned int timeout = 100;
+    auto r = self->queryMacAddress();
+    if (r) // Failed to query MAC Address
+        return;
```
avtolstoy: Error should be ignore here. Otherwise there is a possibility that due to some possibly intermittent communication error with ESP32 NCP, this whole thread will stop forever.

system/src/system_setup.cpp
Cheong2K: Is here something wrong?
```
        if (conf && conf->size) {
            memcpy(&config, conf, conf->size);
        }
```
Should be the following, the size of `conf->size` is bigger than `sizeof(config)`?
```
        if (conf && conf->size) {
            memcpy(&config, conf, sizeof(config));
        }
```

avtolstoy: Or better yet just in case `std::min(sizeof(config), (size_t)conf->size)`

hal/network/lwip/esp32/esp32ncpnetif.cpp
Cheong2K: Fixed.

system/src/system_setup.cpp
Cheong2K: Changed as you said.

PR #1651, opened 2018-12-20
Basic implementation for cellular command

YutingYou:
### Problem

Providing Cellular command on Gen 3 Devices to allow testing customer AT commands.

### Solution

Given `AT Parser` has helped do some formatting work, for compatibility, we add some workarounds to make sure the behavior has the same result as it on Gen 2 devices.

#### TODO
**unsupported feature:**
 - Not support SMS, need to support `>` AT response int AT Parser
 - Not support Socket, need to support `@` AT response in AT Parser

### Steps to Test

1. Flash unit test application `user/tests/hal/cellular` to Electron and Boron

### Example App

`user/tests/hal/cellular/test_cellular_command.cpp`

### References

[CH26290]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/boron/cellular_hal.cpp
+++ b/hal/src/boron/cellular_hal.cpp
@@ -24,8 +24,9 @@
 
 #include "system_network.h" // FIXME: For network_interface_index
 
-#include "scope_guard.h"
+#include "str_util.h"
 #include "endian_util.h"
+#include "scope_guard.h"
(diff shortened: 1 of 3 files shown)


YutingYou:
Thank you Sergey, thanks for all the suggestions and tips, it would be really helpful for the following work!

PR #1652, opened 2018-12-21
Fixes an assertion failure (SOS 10) with Mesh.subscribe() and threading enabled

avtolstoy:
### Problem

The following app will trigger an [assertion failure](https://github.com/particle-iot/openthread/blob/release/20181217-particle/src/core/net/netif.cpp#L124) in OpenThread:

```cpp
SYSTEM_THREAD(ENABLED);
void setup() {
    Mesh.subscribe("test", [](const char* a, const char* b) {
    });
}
```

### Solution

This is a side-effect of OpenThread upgrade. It was previously possible to manage multicast subscriptions even if OpenThread netif was down or not fully up. In order to resolve this issue, we are adding a check for `otIp6IsEnabled()` in `OpenThreadNetif::mldMacFilterCb()` that manages multicast subscription state between LwIP and OpenThread. If OpenThread netif is down, we will keep them stored and will synchronize the state once it comes up.

This PR also enables wiring `UDP` logs only in `DEBUG_BUILD=y` builds, otherwise they pollute the logging output too much.

### Steps to Test

Flash the test application, confirm that the device does not go into SOS 10 (assertion failure).

### Example App

N/A

### References

- [CH26693]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/openthread/lwip_openthreadif.cpp
+++ b/hal/network/lwip/openthread/lwip_openthreadif.cpp
@@ -305,15 +305,24 @@ err_t OpenThreadNetif::mldMacFilterCb(netif* netif, const ip6_addr_t *group,
     otIp6Address addr = {};
     ip6AddrToOtIp6Address(*group, &addr);
 
-    int ret = OT_ERROR_FAILED;
+    int ret = OT_ERROR_NONE;
+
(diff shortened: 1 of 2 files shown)


avtolstoy:
> If OpenThread netif is down, we will keep them stored and will synchronize the state once it comes up.

PR #1658, opened 2018-12-25
CH26714: Arduino compatibility for mesh platforms.

XuGuohui:
```
#include "application.h"
#include "Arduino.h"

SYSTEM_MODE(AUTOMATIC);

volatile uint32_t *portreg;

/* This function is called once at start up ----------------------------------*/
void setup()
{
    // Configure the port to output.
    portreg = portModeRegister(digitalPinToPort(D2));
    *portreg = 0xFFFFFFFF;

    portreg = portOutputRegister(digitalPinToPort(D2));
    *portreg = 0xFFFFFFFF;
}

/* This function loops forever --------------------------------------------*/
void loop()
{
    // Port output high
    *portreg = 0xFFFFFFFF;
    delay(1000);

    // Port output low
    *portreg = 0x00000000;
    delay(1000);
}
```

### DIFF
--- a/user/inc/Arduino.h
+++ b/user/inc/Arduino.h
@@ -120,7 +120,7 @@ inline void yield() {
 # define digitalPinHasPWM(P)        ( HAL_Validate_Pin_Function(P, PF_TIMER) == PF_TIMER )
 # endif
 
-#elif HAL_PLATFORM_MESH
+#elif HAL_PLATFORM_NRF52840
 
 # ifndef digitalPinToPort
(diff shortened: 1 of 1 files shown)


user/inc/Arduino.h
```diff
@@ -120,6 +120,32 @@ inline void yield() {
 # define digitalPinHasPWM(P)        ( HAL_Validate_Pin_Function(P, PF_TIMER) == PF_TIMER )
 # endif
 
+#elif HAL_PLATFORM_MESH
```
avtolstoy: Since this is nRF52840-specific, let's instead use `HAL_PLATFORM_NRF52840`.

PR #1662, opened 2018-12-28
Feature/mesh som

XuGuohui:
Add Argon SoM, Boron SoM and Xenon SoM platforms to Device OS.


hal/src/argon/network/network.cpp
```diff
@@ -129,7 +129,11 @@ int if_init_platform(void*) {
     }
 
     if (HAL_Feature_Get(FEATURE_ETHERNET_DETECTION)) {
+#if PLATFORM_ID == PLATFORM_ARGON
         en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
+#else // Argon SoM
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, D8, 0xFF, D22, mac);
```
avtolstoy: We are setting `RESET` pin to `PIN_INVALID` here for some reason. Did we want to have `CS` pin to be set to the default instead?
```c++
WizNetif(HAL_SPI_Interface spi, pin_t cs, pin_t reset, pin_t interrupt, const uint8_t mac[6]);
```

hal/src/boron/network/network.cpp
```diff
@@ -131,7 +131,11 @@ int if_init_platform(void*) {
     }
 
     if (HAL_Feature_Get(FEATURE_ETHERNET_DETECTION)) {
+#if PLATFORM_ID == PLATFORM_BORON
         en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
+#else // Boron SoM
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, D8, 0xFF, D22, mac);
```
avtolstoy: Same here.

hal/src/nRF52840/rgbled_hal.c
```diff
@@ -6,23 +6,32 @@
...
 #define LED4_GPIO_PIN                       14                      //GREEN Led
+#endif
 
 
 led_config_t HAL_Leds[] = {
     {
         .version = 0x00,
         .pin = LED1_GPIO_PIN,
-        .is_inverted = 1
+        .is_inverted = 0
```
avtolstoy: Will this break non-SoM devices? Should there be a conditional instead here?

hal/src/xenon/network/network.cpp
```diff
@@ -65,7 +65,11 @@ int if_init_platform(void*) {
     }
 
     if (HAL_Feature_Get(FEATURE_ETHERNET_DETECTION)) {
+#if PLATFORM_ID == PLATFORM_XENON
         en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
+#else // Xenon SoM
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, D8, 0xFF, D22, mac);
```
avtolstoy: Same here.

hal/src/nRF52840/rgbled_hal.c
XuGuohui: This will also work for the mesh dev kits, since it was the bug in mesh-develop.

hal/src/argon/network/network.cpp
XuGuohui: We are going to connect A7 to the Ethernet reset pin on the final mesh EVB. Will change it here in advance. It's not very meaningful to set the `cs` pin to a default, cos `D5` is not the default `CS` pin for mesh dev kits. Also for future platforms, it might not use the default `CS` pin for the Ethernet.

PR #1663, opened 2018-12-29
[Gen 3] Fix on-device unit tests

sergeuz:
### Problem

This PR is a WIP, and at least the following tests are still failing:

#### wiring/no_fixture

- `GPIO_05`, `GPIO_06` take too long and eventually fail.
- `GPIO_08` expects `HAL_ADC_Read()` to change the pin mode to `AN_INPUT`.
- `WIFI_04` depends on `wlan_fetch_ipconfig()` which doesn't set DNS and DHCP server addresses at the moment.
- ~`TIMER_07`, `TIME_13`, `WIFI_12` cause system panic either immediately or shortly after the test.~
- ~`APPLICATION_WATCHDOG_02` fails intermittently.~
- `LED_11` fails to validate the state of the RGB LED (possibly, due to `HAL_Led_Rgb_Get_Max_Value()` returning a value that doesn't work well with the test logic).
- `INTERRUPTS_02`, `INTERRUPTS_03`, `SYSTEM_04`, `SYSTEM_05` don't compile due to missing interrupt HAL definitions or dependency on STM32-specific functions.

### References

- [ch19780]

---

### Completeness

- [ ] Added documentation


sergeuz:
This PR is ready for review (remaing tests will be addressed in separate PRs). Besides of the tests, this PR fixes the following issues:
- An attempt to unitialize an SPI interface causes a hardfault if the interface is not initialized.
- Default SPI settings are not recognized correctly.
- Fixed a possible race condition in the Timer's uninitialization code.
- `random()` is now properly seeded on application startup.

PR #1664, opened 2019-01-02
Bypass git dependency in build

zfields:
### Problem

`git` should not be a build dependency.

It is possible for the firmware to build correctly without having `git` installed. Currently, `git` is used to test which branch is being built and offer a warning message if you are working on the `develop` branch, without the having specified the `PARTICLE_DEVELOP` environment variable.

The branch check forces a dependency on Git and spews error messages in its absence. The false positive error messages obfuscate true errors and warnings, which make it harder for a developer to monitor and catch warnings and errors.

### Solution

Gate the logic in `checks.mk` behind a check to see if `git` is installed

### Steps to Test

Build the firmware

### References

N/A

--------

### Completeness

### DIFF
--- a/build/checks.mk
+++ b/build/checks.mk
@@ -1,10 +1,11 @@
-
-git_branch := $(shell git rev-parse --abbrev-ref HEAD)
-ifeq ("$(git_branch)","develop")
-    ifndef PARTICLE_DEVELOP
-       $(error Please note the develop branch contains untested, unreleased code. \
-        We recommend using the 'release/stable' branch which contains the latest released firmware code. \
-        To build the develop branch, please see the the build documentation at \
-        https://github.com/spark/firmware/blob/develop/docs/build.md#building-the-develop-branch)
+ifneq (, $(shell /bin/bash -c "command -v git"))
+    git_branch := $(shell git rev-parse --abbrev-ref HEAD)
+    ifeq ("$(git_branch)","develop")
+        ifndef PARTICLE_DEVELOP
+            $(error Please note the develop branch contains untested, unreleased code. \
+            We recommend using the 'release/stable' branch which contains the latest released firmware code. \
+            To build the develop branch, please see the the build documentation at \
+            https://github.com/spark/firmware/blob/develop/docs/build.md#building-the-develop-branch)
+        endif
     endif
 endif
--- a/build/checks.mk
+++ b/build/checks.mk
@@ -1,4 +1,4 @@
-ifneq (, $(shell sh -c "command -v git"))
+ifneq (, $(shell command -v git;))
     git_branch := $(shell git rev-parse --abbrev-ref HEAD)
     ifeq ("$(git_branch)","develop")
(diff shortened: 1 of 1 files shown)


kennethlimcp:
`git` is also used to pull `submodules` for the eventual codebase with mesh firmware so removing the dependency check might end up tripping many.

zfields:
Git is an will always be an important part of our workflow, but "compiling the code" is a separate task from "acquiring the code" and this change will allow us to better separate our concerns.

An immediate benefit that can be realized by separating concerns is the ability to minimize the space requirements of a container housing the build dependencies (which so happens to be the motivation for this PR).

zfields:
After a conversation with @m-mcgowan and @sergeuz, we have decided on an alternative approach.

First, we will perform a simple check to see if `git` is installed on the host machine. If not, the check will not be performed and no warning will be generated. Second, this check will be performed upon the conclusion of the build (as opposed to the beginning) where the warning will be displayed.

The reasoning behind this approach is as follows:
- A warning of this nature should not block the build
- The warning will get lost in the build spew if it is performed early on and the build is successful

zfields:
This PR now reflects part 1 (checking for git) of the complete solution described above.

zfields:
Need to confirm on all platforms prior to merging...

- [x] Linux
- [ ] OSX
- [ ] Windows

build/checks.mk
```diff
@@ -1,4 +1,4 @@
-ifneq (, $(shell /bin/bash -c "command -v git"))
+ifneq (, $(shell sh -c "command -v git"))
```
avtolstoy: Why not simply `$(shell command -v git)`?

zfields: That was my first effort, it returns the following errors...

```bash
make: command: Command not found
make[1]: command: Command not found
make[2]: command: Command not found
...
```

In order for it to work correctly, it requires [modifying the SHELL variable in the Makefile](https://stackoverflow.com/questions/40182887/make-command-not-found). I was afraid of collateral damage, but I can avoid it by invoking the local shell.

avtolstoy: `$(shell command -v git 2> /dev/null)` ?

avtolstoy: That's weird. `command` should be supported everywhere. But if it doesn't work for you, it might not work for somebody else, so `sh -c command` sounds safer. :+1:

zfields: Thanks @avtolstoy for sticking to your guns! I learned something new, thanks to the [link you shared with](https://stackoverflow.com/questions/17547625/how-to-use-shell-builtin-function-from-a-makefile) me about makefile shell execution.
I have updated the script according to your feedback, it looks much cleaner! Thanks for hanging with me. :smile:

zfields:
I have confirmed the fix works for Windows!

sergeuz:
Works on macOS as well

zfields:
What else needs to be done before we merge this?

technobly:
Rebase and merge :)

PR #1665, opened 2019-01-03
Dependency clarification

zfields:
### Problem

The dependency information is incomplete, and the available information is unorganized and hard to process.

### Solution

Introduced markdown tables to help standardize the dependencies and where to find them. Also created new section to describe using the toolchain via Docker container (which manages all dependencies for the end-user).

I have also supplied the Dockerfile used to create the image that supports containerized toolchain.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -46,7 +46,6 @@ docker run --rm --volume <local firmware repository>:/particle-iot/firmware --vo
 
 - [`arm-none-eabi-gcc`](#gcc-for-arm-cortex-processors)
 - [`make`](#gnu-make)
-- [`dfu-util`](#device-firmware-upgrade-utilities)
 - [`crc32`](#32-bit-cyclic-redundency-check-utility)
 - [`xxd`](#xxd-hex-dump-utility)
 - [Zadig (Windows only)](#zadig-windows-only)
@@ -72,6 +71,11 @@ message if the version is older than this.
 **Linux _(non-Ubuntu)_:**
 
 Install from the download page.
+
+Additional 32-bit dependencies:
+
+- [`lib32ncurses5`](https://packages.debian.org/stretch/lib32ncurses5) - The ncurses library routines are a terminal-independent method of updating character screens with reasonable optimization.
+- [`libc6-i386`](https://packages.debian.org/stretch/libc6-i386) - This package includes shared versions of the standard C library and the standard math library, as well as many others.
 
 **OSX:**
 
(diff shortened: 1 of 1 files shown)


docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
+There are two options for building the Particle firmware.
 
-1. [GCC for ARM Cortex processors](#1-gcc-for-arm-cortex-processors)
-2. [Make](#2-make)
-3. [Device Firmware Upgrade Utilities](#3-device-firmware-upgrade-utilities)
-4. [Zatig](#4-zatig) (for Windows users only)
-5. [Git](#5-git)
-6. [Command line tools](#6-command-line-tools)
+1. You can download and install the dependencies natively
+1. You can install Docker and run the toolchain in a container _(with all dependencies supplied therein)_
```
avtolstoy: 2?

docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
+
+Both options are presented below, and only one is required to build the Particle firmware.
+
+## Docker build container
+
+The Docker build container will execute `make`, and any parameters you pass behind it will be passed directly to `make`. Although more verbose, the Docker build container has a couple of advantages over native installation.
+
+1. Isolation of build dependencies required by the Particle firmware
+1. Dependencies are always in sync and are managed by Particle
```
avtolstoy: 2?

docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
 
-#### 1. GCC for ARM Cortex processors
 The Core/Photon uses an ARM Cortex M3 CPU based microcontroller. All of the code is built around the GNU GCC toolchain offered and maintained by ARM.
 
 The build requires version 5.3.1 20160307 or newer of ARM GCC and will print an error
 message if the version is older than this.
 
+| OS | Distribution | Package Manager | Command |
+|:-- |:------------ |:--------------- |:------- |
+| Linux | Ubuntu | apt-get | `sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa && sudo apt-get update && sudo apt-get install gcc-arm-embedded` |
```
avtolstoy: This will install gcc 8 though, which we do not support.

docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
+- [`make`](#gnu-make)
+- [`crc32`](#32-bit-cyclic-redundency-check-utility)
+- [`xxd`](#xxd-hex-dump-utility)
+- [Zadig (Windows only)](#zadig-windows-only)
+
+### GCC for ARM Cortex processors
+
+| Confirmed version | Download page |
+|:-----------------:|:-------------:|
+| 5.3.1 | [launchpad.net](https://launchpad.net/gcc-arm-embedded) |
```
avtolstoy: AFAIK ARM toolchains can no longer be downloaded from Launchpad. This link should instead be https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads

docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
+| OSX | n/a | [Homebrew](http://brew.sh/) | `# See detailed instructions below` |
+| Windows | n/a | n/a | `# See detailed instructions below` |
+
+**Linux _(non-Ubuntu)_:**
+
+Install from the download page.
 
-**Linux and Windows**:
-- Download and install version 5.3.1 from: https://launchpad.net/gcc-arm-embedded
+Additional 32-bit dependencies:
```
avtolstoy: I think we should generalize this section and remove Debian-specific references.

docs/dependencies.md
```diff
@@ -1,80 +1,196 @@
...
+| Windows | n/a | n/a | `# See detailed instructions below` |
 
-#### 5. Git
+**Windows:**
 
-Download and install Git: http://git-scm.com/
+Install `MinGW` and add it to your path.
 
-#### 6. Command line tools
+### Zadig (Windows only)
```
avtolstoy: In theory this should no longer be required as we implement proper WCID descriptors for Windows on both Gen 2 and Gen 3 devices (but not for Core). So we might add something like 'generally this should not be required for recent releases'.

build/checks.mk
```diff
@@ -1,10 +0,0 @@
-
```
avtolstoy: As far as I understand this change was reverted in https://github.com/particle-iot/device-os/pull/1664. I'd remove https://github.com/particle-iot/device-os/pull/1665/commits/f77ae441d6250b0a91371adba6739aff8c1b86d4 from this branch altogether in order to not accidentally overwrite #1664.

docs/dependencies.md
zfields: In markdown you just use the number 1 over and over again, and it will render it correctly. This allows you to rearrange steps without having to renumber the whole list.

avtolstoy: While I agree that it renders correctly, markdown is still a human-readable format and I don't see any reason why the enumeration shouldn't be correct even in plain-text.

PR #1667, opened 2019-01-07
Implement standby mode

YutingYou:
### Problem

Implement standby mode.

### Solution

This lower power mode can be implemented using nRF52840 System OFF sleep mode. Unfortunately, it is not possible to keep RTC running while in this state, so only wake up by WKP pin is supported.

### API Test
Test case 1: Enter sleep mode
Test case 2: Doesn't enter sleep mode, RTC seconds should be 0s
Test case 3: Doesn't enter sleep mode, have to support wakeup pin

```cpp
#include "application.h"

SYSTEM_MODE(MANUAL);

void setup() {
    Serial.begin(115200);
}

void loop() {
    while (Serial.available()) {
        char ch = Serial.read();
        switch (ch) {
            // Handle terminal operation characters
            case '1': {
                Serial.printf("enter sleep");
                System.sleep(SLEEP_MODE_DEEP, 0);
                break;
            }
            case '2': {
                Serial.printf("enter sleep, 10s");
                System.sleep(SLEEP_MODE_DEEP, 10);
                break;
            }
            case '3': {
                Serial.printf("enter sleep, 0s, disable WKP_PIN");
                System.sleep(SLEEP_MODE_DEEP, 0, SLEEP_DISABLE_WKP_PIN);
                break;
            }
            default:
                break;
        }
    }
}
```
### Power test
With below test application, got the power consumption result. This test is rough, just make sure the power consumption is greatly reduced.

1. Running:  7.82~8.35mA
2. Deep sleep: 25uA

```cpp
#include "application.h"

SYSTEM_MODE(MANUAL);

static volatile bool s_sleep = false;

void setup(void) {
    System.on(button_click, [](system_event_t event, int data) -> void {
        s_sleep = true;
    });
}

void loop(void) {
    if (s_sleep) {
        System.sleep(SLEEP_MODE_DEEP, 0);
    }
}
```
### References

- [CH26710]
- [CH27717]

---

### Completeness

- [ ] Added documentation


EchoBravoVictor:
Finally !!

gothamsound:
On the Gen3 hardware, how would one specify WKP_PIN, or is it a fixed pin like previous gen hw?

Thanks,
Peter

YutingYou:
Hi Peter @gothamsound, it is a fixed pin like Gen2 device for now.

gothamsound:
@YutingYou Eugene - Thank you! Which physical pin would that be on the Boron?
Also, is there any way I could implement some of this functionality now, with the currently released firmware?
Thanks and I apologize if this isn't the right place for this question.

Peter

YutingYou:
@gothamsound, I afraid you need to wait for next release, the WKP pin on Boron is D8.

PR #1681, opened 2019-01-25
[Boron] Ensure that the modem is on before trying to retrieve IMEI and ICCID

avtolstoy:
### Problem

Boron in Listening Mode does not show IMEI and ICCID.
```console
i
Device ID: XXXXXXXXXXXXXXXXXXXXXX
Device Secret: XXXXXXXXXXXXXX
Serial Number: XXXXXXXXXXXXXX
IMEI:
ICCID:

v
system firmware version: 0.8.0-rc.26
```

### Solution

Ensure that the modem is on before trying to retrieve IMEI and ICCID in `cellular_device_info()`.

### Steps to Test

Put the Boron into listening mode, open its serial port, send 'i', it should correctly output IMEI and ICCID.

### Example App

N/A

### References

- [CH26968]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/boron/cellular_hal.cpp
+++ b/hal/src/boron/cellular_hal.cpp
@@ -136,6 +136,10 @@ int cellular_device_info(CellularDevice* info, void* reserved) {
     CHECK_TRUE(mgr, SYSTEM_ERROR_UNKNOWN);
     const auto client = mgr->ncpClient();
     CHECK_TRUE(client, SYSTEM_ERROR_UNKNOWN);
+
+    const NcpClientLock lock(client);
+    // Ensure the modem is powered on
+    CHECK(client->on());
     CHECK(client->getIccid(info->iccid, sizeof(info->iccid)));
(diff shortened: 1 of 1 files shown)


hal/src/boron/cellular_hal.cpp
```diff
@@ -136,6 +136,13 @@ int cellular_device_info(CellularDevice* info, void* reserved) {
     CHECK_TRUE(mgr, SYSTEM_ERROR_UNKNOWN);
     const auto client = mgr->ncpClient();
     CHECK_TRUE(client, SYSTEM_ERROR_UNKNOWN);
+
+    client->lock();
+    SCOPE_GUARD({
+        client->unlock();
+    });
```
sergeuz: There's a lock guard for NCP client classes: https://github.com/particle-iot/device-os/blob/mesh-develop/hal/network/ncp/ncp_client.h#L110

avtolstoy: Completely forgot about it. Fixed.

Issue #1654, opened 2018-12-22
[Mesh rc.27] Regression causing PietteTech_DHT library to fail


ScruffR:
As discussed in this thready
https://community.particle.io/t/argon-dht11-nothing-but-zeros/46108/31?u=scruffr

While the library sample [`dht_simple.ino`](https://github.com/eliteio/PietteTech_DHT/tree/master/examples/DHT_simple) works perfectly fine with rc.26 for DHT11 & DHT22 sensors, rc.27 has introduced major problems for that library that worked flawless on Gen1&2 for years.
With a DHT22 I get max 3 to 4 good readings followed by a couple of errors and with a DHT11 no good readings at all.
Since this library is interrupt based and there were some "fixes" in connection with interrupts I'm inclined some of these fixes have actually unfixed something else.

Tested on Argon 0.8.0-rc.26 & -rc.27

YutingYou:
Hi @ScruffR, I can't reproduce this issue on Argon-RC27 modular firmware with DHT11. I used manual mode.

```
Sat Jan  1 00:00:32 2000
Humid: 46.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 10.75*C - HeatI: 25.11*C
Sat Jan  1 00:00:39 2000
Humid: 49.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 11.70*C - HeatI: 25.09*C
Sat Jan  1 00:00:41 2000
Humid: 61.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 15.07*C - HeatI: 24.71*C
Sat Jan  1 00:00:43 2000
Humid: 74.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 18.11*C - HeatI: 23.76*C
Sat Jan  1 00:00:46 2000
Humid: 83.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 19.95*C - HeatI: 22.78*C
Sat Jan  1 00:00:48 2000
Humid: 89.00% - Temp: 23.00*C 73.40*F 296.15*K - DewP: 21.08*C - HeatI: 21.98*C
Sat Jan  1 00:00:50 2000
Humid: 93.00% - Temp: 24.00*C 75.20*F 297.15*K - DewP: 22.79*C - HeatI: 23.32*C
Sat Jan  1 00:00:52 2000
[...]

PR #1682, opened 2019-01-28
[Gen 3] STOP sleep mode support

YutingYou:
### Problem

STOP mode is not implemented for Gen 3 devices.

### Solution

Well, implement it.

### Tests to run

- :heavy_check_mark: `app/sleep_multiple_pins`
- :heavy_check_mark: `wiring/sleep`

### References

- [CH26710]
- [CH27767]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/core_hal.c
+++ b/hal/src/nRF52840/core_hal.c
@@ -53,6 +53,7 @@
 #include "system_error.h"
 #include <nrfx_rtc.h>
 #include "gpio_hal.h"
+#include "exflash_hal.h"
 
 #define BACKUP_REGISTER_NUM        10
 static int32_t backup_register[BACKUP_REGISTER_NUM] __attribute__((section(".backup_registers")));
@@ -99,6 +100,9 @@ extern void* malloc_heap_end();
 #if defined(MODULAR_FIRMWARE)
 void* module_user_pre_init();
 #endif
+
+extern void nrf5AlarmInit(void);
+extern void nrf5AlarmDeinit(void);
 
 __attribute__((externally_visible)) void prvGetRegistersFromStack( uint32_t *pulFaultStackAddress ) {
     /* These are volatile to try and prevent the compiler/linker optimising them
@@ -508,7 +512,8 @@ static void fpu_sleep_prepare(void) {
     SPARK_ASSERT((fpscr & 0x07) == 0);
 }
 void HAL_Core_Enter_Stop_Mode(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long seconds) {
-
+    InterruptMode m = (InterruptMode)edgeTriggerMode;
+    HAL_Core_Enter_Stop_Mode_Ext(&wakeUpPin, 1, &m, 1, seconds, NULL);
 }
 
 static void wakeup_rtc_handler(nrfx_rtc_int_type_t int_type) {
@@ -569,6 +574,12 @@ int32_t HAL_Core_Enter_Stop_Mode_Ext(const uint16_t* pins, size_t pins_count, co
     // Detach USB
     HAL_USB_Detach();
 
+    // Disable RTC2
+    nrf5AlarmDeinit();
+
+    // Disable external flash
+    hal_exflash_uninit();
+
     // Flush all USARTs
     for (int usart = 0; usart < TOTAL_USARTS; usart++) {
         if (HAL_USART_Is_Enabled(usart)) {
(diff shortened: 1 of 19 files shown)


hal/src/nRF52840/exflash_hal.c
```diff
@@ -150,6 +150,25 @@ int hal_exflash_init(void)
...
+    nrfx_qspi_uninit();
+    // PATCH: Initialize CS pin, external memory discharge 
+    nrf_gpio_cfg_output(QSPI_FLASH_CSN_PIN);
+    nrf_gpio_pin_set(QSPI_FLASH_CSN_PIN);
+    nrf_gpio_cfg_input(QSPI_FLASH_IO1_PIN, NRF_GPIO_PIN_PULLDOWN);
+    // The nrfx_qspi driver doesn't clear pending IRQ
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+    SPARK_ASSERT(sd_nvic_ClearPendingIRQ(QSPI_IRQn) == NRF_SUCCESS);
+#else
+    NVIC_ClearPendingIRQ(QSPI_IRQn);
```
avtolstoy: This conditional should be unnecessary as `sd_nvic_ClearPendingIRQ()` in bootloader will map to `NVIC_ClearPendingIRQ()` directly.

hal/src/nRF52840/core_hal.c
```diff
@@ -464,11 +486,229 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+        }
+    }
+
+    hal_exflash_init();
+
+    nrf5AlarmInit();
+
+    HAL_USB_Attach();
+
+    SysTick->CTRL |= SysTick_CTRL_ENABLE_Msk;
```
avtolstoy: We should block all interrupts until this point, otherwise a user interrupt handler will not be executed correctly if the wakeup pin was configured with `attachInterrupt()` in user firmware.

hal/src/nRF52840/core_hal.c
```diff
@@ -464,11 +486,229 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+    SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
+
+    // Detach USB
+    HAL_USB_Detach();
+
+    // Disable RTC2
+    nrf5AlarmDeinit();
+
+    // Disable external flash
+    hal_exflash_uninit();
```
avtolstoy: This should be performed only after the interrupts are disabled, otherwise a context switch might happen and some other thread might attempt to read/write the QSPI flash.

hal/src/nRF52840/core_hal.c
```diff
@@ -464,11 +486,229 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+        }
+    }
+
+    SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
+
+    // Detach USB
+    HAL_USB_Detach();
+
+    // Disable RTC2
+    nrf5AlarmDeinit();
```
avtolstoy: This should probably also be performed after the interrupts are disabled just in case.

hal/src/nRF52840/core_hal.c
```diff
@@ -464,11 +486,229 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+
+    // Configure RTC wake-up
+    if (seconds > 0) {
+        wakeup_from_rtc(seconds);
+
+        exit_conditions |= STOP_MODE_EXIT_CONDITION_RTC;
+    }
+
+    // Enable all interrupt, __enable_irq()
+    err_code = sd_nvic_critical_region_exit(dummy);
```
avtolstoy: Should not be enabled here. They should be disabled almost until exiting this function after the wakeup.

hal/src/nRF52840/core_hal.c
```diff
@@ -464,11 +486,229 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+    APP_ERROR_CHECK(err_code);
+
+    fpu_sleep_prepare();
```
avtolstoy: If PWM keeps running while in STOP mode, I think we need to suspend all 4 PWM instances before going to sleep here (while keeping their configuration).

hal/src/nRF52840/core_hal.c
```diff
@@ -470,13 +485,450 @@ void HAL_Core_Enter_Bootloader(bool persist) {
...
+
+        // Temporarily enable SoftDevice API interrupts just in case
+        uint32_t base_pri = __get_BASEPRI();
+        // We are also allowing IRQs with priority = 5, because that's what
+        // OpenThread uses for its SWI3
+        __set_BASEPRI(_PRIO_APP_LOW << (8 - __NVIC_PRIO_BITS));
+        sd_nvic_critical_region_exit(st);
+        {
+
+            nrf_drv_clock_hfclk_request(NULL);
```
avtolstoy: Remove this.

avtolstoy:
This PR also increases micros() resolution and should resolve #1654

dmiddlecamp:
just curious what's the current state of sleep behavior on gen3 devices, are we targeting a particular release for this PR?

PR #1685, opened 2019-01-30
Maintain parity with Arduino IDE

zfields:
Maintain parity with Arduino IDE

### Problem

Workbench does not recognize `uint32_t`, but it does _NOT_ result in a compile error

### Solution

`#include <stdint.h>` in `Particle.h` to ensure `stdint.h` is in workbench's include path.

### Steps to Test

Create an app in workbench and specify `unit32_t` type

### Example App

```c
uint32_t var;

void setup() {}
void loop() {}
```

### References

https://community.particle.io/t/issue-uint32-t-is-not-a-type-name-when-configured-for-mesh/47359

---

### Completeness

- [ ] Added documentation


zfields:
PR against wrong base

PR #1686, opened 2019-01-30
Enable workbench to recognize `uint32_t`

zfields:
Maintain parity with Arduino IDE

### Problem

Workbench does not recognize `uint32_t`, but it does _NOT_ result in a compile error

### Solution

`#include <stdint.h>` in `Particle.h` to ensure `stdint.h` is in workbench's include path.

### Steps to Test

Create an app in workbench and specify `unit32_t` type

### Example App

```c
uint32_t var;

void setup() {}
void loop() {}
```

### References

https://community.particle.io/t/issue-uint32-t-is-not-a-type-name-when-configured-for-mesh/47359

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/inc/Particle.h
+++ b/user/inc/Particle.h
@@ -2,7 +2,8 @@
 #ifndef PARTICLE_H
 #define	PARTICLE_H
 
+#include <stdint.h>
+
 #include "application.h"
 
 #endif	/* PARTICLE_H */
-


user/inc/Particle.h
```diff
@@ -2,7 +2,8 @@
 #ifndef PARTICLE_H
 #define	PARTICLE_H
 
+#include <stdint.h>
```
avtolstoy: I have no objections towards merging this, but the simple question is: are we going to add duct tape for all the cases where vscode intellisense fails? :) I'm pretty confident that `stdint.h` or `cstdint` is included somewhere in the hal or platform headers, hence
> but it does NOT result in a compile error

zfields: Apparently, the inclusion bubbles all the way to the top on some platforms and not for others. I spoke with JBeri and he believes we have set an expectation that we need to maintain.

beriberikix: I wouldn't suggest we add Band-Aids or hacks across our platform but we should evolve our systems to better intergrate with tools, whether that's CI, CLI or Workbench.

Is there a better way to fix this issue in the longer-term?

zfields: Mirande said this does not fix his problem. That blew my mind, I need to circle the wagons. I've changed the status to `do not merge`

busticated: fwiw, this is what i did to work-around the issue:

https://github.com/particle-iot/workbench/pull/54/commits/5ff49b73ac8987ddc88f79a54a447f1bb0b680ed

also, here's the story i put together for the Device OS side:

https://app.clubhouse.io/particle/story/28581/uint32-t-is-not-defined-visible-to-intellisense

cc @avtolstoy @zfields

PR #1688, opened 2019-02-01
[LTE/Electron] Fixes software Timer()'s halting after millis() overflows (every 49.7 days)

monkbroc:
### Problem

The [Particle software timers aka `Timer`](https://docs.particle.io/reference/device-os/firmware/photon/#software-timers) use the FreeRTOS timer feature. FreeRTOS version 8.2.2 used on the Electron introduced a bug where timers stop triggering after an overflow of the tick count, which happens after ~49 days.

### Solution

Rather than upgrade the whole FreeRTOS this PR backports the bugfix from https://github.com/cjlano/freertos/commit/f780d64f5244431c609696f7ca1c3e06ea268a76#diff-ed9ddddbc8c65ef44735bc5950056ef7R474

### Steps to Test

Steps to validate the issue and fix:
1. Modify https://github.com/particle-iot/device-os/blob/0b83dbf09fb9c98d910c91ac39d0a75c983fecad/hal/src/electron/rtos/FreeRTOSv8.2.2/FreeRTOS/Source/tasks.c#L1593 to trigger the timer overflow soon after boot

```
		xTickCount = ( TickType_t ) 0xFFFFD000U;
```
1. Add the example application below as `~Downloads/timer/timer.cpp`
1. Compile and flash with `modules$ make clean all -s COMPILE_LTO=n PLATFORM=electron DEBUG_BUILD=y APPDIR=~/Downloads/timer program-dfu`
1. Connect to the USB serial port and see the printed numbers go from 1 to 11 then stop.
1. Apply the changes in this PR
1. Recompile (clean build important!) and flash
1. Connect to the USB serial port and see the printed numbers go from 1 upwards without stopping.

### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);

void print_every_second()
{
    static int count = 0;
    Serial.println(count++);
}

Timer timer(1000, print_every_second);

void setup()
{
    Serial.begin(9600);
    timer.start();
}
```

### Completeness

### DIFF
--- a/hal/src/electron/rtos/FreeRTOSv8.2.2/FreeRTOS/Source/timers.c
+++ b/hal/src/electron/rtos/FreeRTOSv8.2.2/FreeRTOS/Source/timers.c
@@ -233,7 +233,7 @@ static TickType_t prvGetNextExpireTime( BaseType_t * const pxListWasEmpty ) PRIV
  * If a timer has expired, process it.  Otherwise, block the timer service task
  * until either a timer does expire or a command is received.
  */
-static void prvProcessTimerOrBlockTask( const TickType_t xNextExpireTime, const BaseType_t xListWasEmpty ) PRIVILEGED_FUNCTION;
+static void prvProcessTimerOrBlockTask( const TickType_t xNextExpireTime, BaseType_t xListWasEmpty ) PRIVILEGED_FUNCTION;
 
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


MrZANE42:
@technobly When did you start using FreeRTOS 8.2.2? Ie. What (electron) System OS versions are affected by this issue?

hwestbrook:
@MrZANE42 -- This bug looks like it exists in 0.6.1. Not sure how much further back it might go.

jlovinger-particle:
The version of FreeRTOS with this bug was introduced to the repo on 9/22/2015 in the following commit.

https://github.com/particle-iot/device-os/commit/ba342c3530a37ca9da1ae4be045f5fc5f294c147

Based on the timing (didn't test personally) looks like everything from 0.4.6 onwards.

PR #1689, opened 2019-02-04
[LTE/Electron] bug fixes and enhancements

technobly:
### Problem

Several LTE bugs have been found that all revolve around socket AT commands being prone to various types of failure.  See below.

#### Fixes incorrect handling of socket send when lost signal recovers during 2nd try
    
- If signal was lost before socket send, first command would timeout, during retry if signal was restored the first command would send and return OK instead of the second command requiring a PROMPT.  We would return a socket error instead of signaling data was sent OK.  This caused a network teardown and reconnection unnecessarily.
- Also adjusts the timeouts for various commands used during socket send operations.

#### Changes eDRX mode command so that it fully indicates it is disabled
    
- Despite eDRX mode being disabled via CEDRXS=0, it would still indicate \r\n+CEDRXS: 2, ... \r\n
- Now with CEDRXS=3,2 it will respond as fully disabled \r\n+CEDRXS:\r\n

#### Fixes LTE AT command interface failing when sending data with no cell signal, and more... [ch26915]
    
- If signal was lost before data sent, the AT interface would hang until the modem was reset.  This process would take around 10 minutes.
- Added checks for network registration before certain commands: USOST, USOCO, USOWR, USOCL to prevent AT interface hanging.  Unfortunately it still takes 20-30s before network registration updates, so we can still stall on these commands at times.
- Timeouts were reduced as much as possible per AT command manuals to speed up recovery process when AT interface fails.  Some timeouts were optimized for UDP usage.
- Additional checks for AT interface being available were added before long timeout commands.
- Improvements to power off routine added by using the PWR_UC pin and checking for V_INT voltage to drop.
- 10ms delay changed to 1ms delay in main parsing loop to speed up processing of many commands.
- Removed 50ms delay in socket send functions (TCP & UDP) which should also speed up communication.
- CREG?/CGREG?/CEREG? polling sped up considerably from 15s delays between tries to 10ms (1500x faster in certain cases). Time to register now on 2G/3G/LTE can be <4s for a warm boot and <9s for a cold boot.

#### Fixes LTE devices remaining disconnected until reset when connecting with no signal [ch26919]
    
- If the modem starts up with the factory default URAT setting of 7,8 (Cat-M1 & Cat-NB1) and is not able to initially connect to a Cat-M1 network, it will start scanning for Cat-NB1 networks and not return to scanning for Cat-M1 networks. In this case, it will be unable to connect to a Cat-M1 network until the cellular modem is power cycled after 5 minutes.
- Solution is to disable Cat-NB1 RAT and only use Cat-M1 RAT by default.

#### Fixes LTE/Electron AT command replay issue [ch26922]
    
- If the AT interface fails, commands sent to the modem can be replayed when it is power cycled and reset due to the circular buffer not being cleared completely.
- This resolves the issue by resetting all indexes used in the circular buffer to 0 when the serial pipe is effectively closed.

### Steps to Test

- Run `wiring/no_fixture` tests
- Simulate no signal by removing the antenna
    - and sending data after 5 seconds, wait for AT commands to fail, replace antenna after 30 seconds, unit should recover after about 60 seconds from when data was sent.
    - and sending data after 5 seconds, wait for AT commands to fail, reset the device, unit should boot up again and try to register.  (failing case would get stuck in listening mode repeatedly due to the AT command replay issue)
    - and sending data after 40 seconds (unit should already be indicating no signal and no network connection), AT commands should not fail as all offending commands should be avoided with CEREG? checks, replace antenna, unit should recover in about 30 seconds from when data was sent.
    - and sending data immediately, and replace antenna after first USOST command observed, unit should recover without issue and keep communicating.
    - and sending data immediately, and replace antenna after second USOST command observed, unit should usually recover without issue and keep communicating.
    - and cold boot the unit, and observe it will now toggle between CEREG: 2,0 and CEREG: 2,2.  After replacing the antenna it will connect almost immediately.
- Ensure URAT? responds with +URAT: 7 instead of +URAT: 7,8
- Attempt to force URAT=7,8 to check again that system will fix this default situation
```c
SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);
void setup() {
    cellular_on(NULL);
    cellular_command(NULL, NULL, 1000, "AT+URAT=7,8\r\n");
    Particle.connect();
}
```

---

### Completeness

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -62,12 +62,25 @@ std::recursive_mutex mdm_mutex;
 // data read from a socket if the data contains a null byte
 // #define SOCKET_HEX_MODE
 
-// Timeouts for various AT commands
-#define USOWR_TIMEOUT   (120 * 1000) /* 120 for R4, 1 for U2/G3 */
-#define USORD_TIMEOUT   ( 1 * 1000)
-#define USOST_TIMEOUT   (10 * 1000)  /*  10 for R4, 1 for U2/G3 */
-#define USORF_TIMEOUT   ( 1 * 1000)
-#define COPS_TIMEOUT    ( 3 * 60 * 1000)
+// Timeouts for various AT commands based on R4 & U2/G3 AT command manual as of Jan 2019
+#define AT_TIMEOUT      (  1 * 1000)
+#define USOCL_TIMEOUT   ( 10 * 1000) /* 120s for R4 (TCP only, optimizing for UDP with 10s), 1s for U2/G3 */
+#define USOCO_TIMEOUT   ( 10 * 1000) /* 120s for R4 (TCP only, optimizing for UDP with 10s), 1s for U2/G3 */
+#define USOCR_TIMEOUT   (  1 * 1000)
+#define USOCTL_TIMEOUT  (  1 * 1000)
+#define USOWR_TIMEOUT   ( 10 * 1000) /* 120s for R4 (optimizing for non-blocking behavior with 10s), 1s for U2/G3 */
+#define USORD_TIMEOUT   ( 10 * 1000) /* FIXME: 1s for R4/U2/G3, but longer timeouts required in deployments */
+#define USOST_TIMEOUT   ( 10 * 1000) /*  10s for R4, 1s for U2/G3 */
+#define USORF_TIMEOUT   ( 10 * 1000) /* FIXME: 1s for R4/U2/G3, but longer timeouts required in deployments */
+#define CEER_TIMEOUT    (  1 * 1000)
+#define CEREG_TIMEOUT   (  1 * 1000)
+#define CGDCONT_TIMEOUT (  1 * 1000)
+#define CGREG_TIMEOUT   (  1 * 1000)
+#define COPS_TIMEOUT    (180 * 1000)
+#define CPWROFF_TIMEOUT ( 40 * 1000)
+#define CREG_TIMEOUT    (  1 * 1000)
+#define CSQ_TIMEOUT     (  1 * 1000)
+#define URAT_TIMEOUT    (  1 * 1000)
 
 // num sockets
 #define NUMSOCKETS      ((int)(sizeof(_sockets)/sizeof(*_sockets)))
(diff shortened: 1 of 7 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -838,37 +864,65 @@ bool MDMParser::powerOff(void)
...
-            int ret = waitFinalResp(NULL,NULL,40*1000);
-            if (RESP_OK == ret) {
-                _pwr = false;
-                // todo - add if these are automatically done on power down
-                //_activated = false;
-                //_attached = false;
-                // Give the modem some time to switch off cleanly
-                HAL_Delay_Milliseconds(1000);
-                ok = true;
+            if(!_atOk()) {
```
jlovinger-particle: If we only have RI_UC connected to V_INT on the SARA R410 then no way after the PWR_UC pulse to validate modem powered off if AT interface is blocked. Suggest putting the entire section behind the DEV_SARA_410 check (i.e. `if(!_atOK() && _dev.dev == DEV_SARA_R410)`.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -838,37 +864,65 @@ bool MDMParser::powerOff(void)
...
+            } else {
+                sendFormated("AT+CPWROFF\r\n");
+                int ret = waitFinalResp(nullptr, nullptr, CPWROFF_TIMEOUT);
+                if (RESP_OK == ret) {
+                    _pwr = false;
+                    // todo - add if these are automatically done on power down
+                    //_activated = false;
+                    //_attached = false;
+                    ok = true;
+                    check_ri = true;
```
jlovinger-particle: This should be behind an if(_dev.dev == DEV_SARA_R410) check.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -838,37 +864,65 @@ bool MDMParser::powerOff(void)
...
         }
     }
 
+    if (check_ri) {
+        system_tick_t t0 = HAL_Timer_Get_Milli_Seconds();
+        while (HAL_GPIO_Read(RI_UC) && !TIMEOUT(t0, 10000))
+        {
+            HAL_Delay_Milliseconds(1); // just wait
+        }
+        if(!HAL_GPIO_Read(RI_UC))
```
jlovinger-particle: We should set `ok = true` and `ok = false` depending on whether the RI_UC/V_INT indicates modem is actually powered off within the timeout or not. Probably worked in my functional tests since _pwr was still set to false...

hal/src/electron/modem/mdm_hal.cpp
technobly: It likely will power off though, and for now we could add a busy wait of 1s if we know it's not a R410 just to give it some time to do so (similar to the code we've had in there for CPWROFF).

hal/src/electron/modem/mdm_hal.cpp
technobly: 👍

PR #1583, opened 2018-10-31
System API for mesh network management

sergeuz:
This PR implements initial system API for mesh network management.

PR #1693, opened 2019-02-13
Mesh: implement function Mesh.selectAntenna().

XuGuohui:


wiring/inc/spark_wiring_mesh.h
```diff
@@ -38,6 +38,11 @@
 
 namespace spark {
 
+typedef enum {
+    MESH_ANT_INTERNAL = 0,
```
avtolstoy: Let's reserve `0` for default.

wiring/inc/spark_wiring_mesh.h
```diff
@@ -182,6 +187,8 @@ class MeshClass : public NetworkClass, public MeshPublish {
         return network_ready(*this, 0,  NULL);
     }
 
+    void selectAntenna(MeshAntennaType antenna);
```
avtolstoy: Should return `int` (and the error codes from `system_error_t`).

XuGuohui:
So I'll close this PR and export this function in https://github.com/particle-iot/device-os/pull/1583.

XuGuohui:
It needs to change the Boron system part1 linker script as well.

PR #1694, opened 2019-02-13
Fix/gen3 fix no fixture unit test

YutingYou:
### Problem

This PR is additional fixes for some issue in [[Gen 3] Fix on-device unit tests](https://github.com/particle-iot/device-os/pull/1663)

> `GPIO_05`, `GPIO_06` take too long and eventually fail.

This is because PWM function is not available on D1 pin, use D2 instead

> `GPIO_08` expects `HAL_ADC_Read()` to change the pin mode to `AN_INPUT`.

`AN_INPUT` is only for Gen2 device

> `LED_11` fails to validate the state of the RGB LED (possibly, due to `HAL_Led_Rgb_Get_Max_Value()` returning a value that doesn't work well with the test logic).

A7 is not available on Gen3 device and `HAL_Led_Rgb_Get_Max_Value()` is fixed in [mirror button and led](https://github.com/particle-iot/device-os/pull/1590)

### Solution

N/A

### Steps to Test
`user/tests/wiring/no_fixture`

wait until  [mirror button and led](https://github.com/particle-iot/device-os/pull/1590) merged

### References

[CH28410]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/pinmap_hal.h
+++ b/hal/inc/pinmap_hal.h
@@ -212,8 +212,6 @@ STM32_Pin_Info* HAL_Pin_Map(void);
 #define RGBG            22
 #define RGBB            23
 
-#define A7              A5   // FIXME: A7 is used in spark_wiring_wifitester.cpp
-
 #else // Mesh SoMs
 
 // button pin
@@ -230,7 +228,7 @@ STM32_Pin_Info* HAL_Pin_Map(void);
 #endif
 
 // WKP pin
-#define WKP             D8   // FIXME:
+#define WKP             D8
 
 #if PLATFORM_ID == PLATFORM_XENON // Xenon
 #define TOTAL_PINS      (31)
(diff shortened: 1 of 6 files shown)


sergeuz:
> `AN_INPUT` is only for Gen2 device

Should we rework the test, or mixing `analogRead()` and `digitalRead()` without reconfiguring the pin is not supported on gen 3 devices?

YutingYou:
For Gen3 device, `analogRead()` and `digitalRead()` can work concurrently without changing pin mode. So this test case is not needed on Gen3 device.

sergeuz:
> For Gen3 device, `analogRead()` and `digitalRead()` can work concurrently without changing pin mode. So this test case is not needed on Gen3 device.

That's exactly what the test attempts to verify though

YutingYou:
The GPIO hardware structure is different on Gen2 and Gen3 device.

On Gen2 device, in `analogRead()`, the driver saves current pin mode and configure the pin mode to `AN_INPUT` to read analog value. Then in `digitalRead()`, the driver restores pin mode to read digital value. Because there is the pin mode transform operation, the unit test wants to make sure the save/restore operation is correct.

But on Gen3 device, even if you configure it to OUTPUT mode, we can read the analog value. we don't need the save/restore pin mode operation.

PR #1695, opened 2019-02-14
BLE HAL implementation

XuGuohui:
There are examples provided under "./user/tests/app/ble" and units tests under "./user/tests/hal".

**Note: BLE HAL APIs are only supported for monolithic firmware. Bond Manager and Muti-link is not supported for now.**

Build BLE examples:
1. `cd main`
2. `make PLATFORM=xenon MODULAR=n DEBUG_BUILD=y TEST=app/ble/peripheral`

Flash to mesh device:
1. Make the device in DFU mode.
2. `dfu-util -d 2b04:d00e -a 0 -s 0x30000 -D ../build/target/main/platform-14/peripheral.bin`

Please help review and pay more attention to:
1. What BLE HAL APIs are needed but not exposed, e.g. callbacks.
2. The implemented BLE HAL APIs naming convention.
3. BLE HAL related data structures and members.
4. BLE events dispatching scheme.


XuGuohui:
https://github.com/particle-iot/device-os/pull/1740

PR #1710, opened 2019-03-04
Feature/ble/wiring bare api

XuGuohui:


user/tests/wiring/ble/heartrate/application.cpp
```diff
@@ -20,21 +20,16 @@
...
+BLEAttribute heartrate("heartrate", READ | NOTIFY);
+int newHr = 1234;
 
 void setup() {
-    BLE.on();
-
     BLE.advertise();
-
-    BLEDevice local;
-    myConn = BLE.connect();
```
towynlin: There should be no need to call `connect()` here.
`advertise()` should set up everything that is required.

user/tests/wiring/ble/heartrate/application.cpp
```diff
@@ -20,21 +20,16 @@
...
 }
 
 void loop() {
-    uint32_t newHr = 1234;
-    if (BLE.connected(myConn)) {
-        heartrate.setValue((const uint8_t*)&newHr, sizeof(uint32_t));
+    if (BLE.connected()) {
+        heartrate.setValue(newHr);
```
towynlin: We might need to include the size as a second argument. If possible, I'd love to allow users to omit it.

XuGuohui: The second argument is required if user wants to pass in a pointer to an array as the first argument.

user/tests/wiring/ble/heartrate/application.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this program; if not, see <http://www.gnu.org/licenses/>.
+  ******************************************************************************
+ */
+
+/* Includes ------------------------------------------------------------------*/
+#include "application.h"
+
+BLEAttributeClass heartrate("heartrate", READ | NOTIFY);
```
towynlin: Please remove "Class" from the class name — it should just be `BLEAttribute`  like you had it before. I'm curious why you added "Class".

user/tests/wiring/ble/heartrate_c/application.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void setup() {
+}
+
+void loop() {
+    if (BLE.connected()) {
+        uint8_t newHr[4];
+        uint16_t len = 4;
+        heartrate->getValue(newHr, &len);
```
towynlin: I don't think we should have a separate C API. If there's some reason to keep the C API that I don't currently understand, then this second argument should not be a reference. Just make it a `size_t` passed by value.

user/tests/wiring/ble/heartrate_c/application.cpp
```diff
@@ -0,0 +1,63 @@
...
+                bool found = results[i].find(BLE_SIG_AD_TYPE_FLAGS);
+                if (found) {
+                    BLE.connect(results[i].address());
+
+                    if (BLE.connected()) {
+                        peer = BLE.peer(results[i].address());
+
+                        peer->attribute("heartrate", &heartrate);
+
+                        heartrate->onDataReceived(heartrateUpdated);
```
towynlin: This API seems mostly fine, but using it this way looks error prone. The name `onDataReceived` implies you're setting up an event handler that will asynchronously call `heartrateUpdated`, but then we're going to keep iterating through the scan results and overwrite peer, heartrate, and the event handler.

user/tests/wiring/ble/uart/application.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this program; if not, see <http://www.gnu.org/licenses/>.
+  ******************************************************************************
+ */
+
+/* Includes ------------------------------------------------------------------*/
+#include "application.h"
+
+void onDataReceived(uint8_t* data, uint16_t len);
```
towynlin: second arg should be type `size_t`

user/tests/wiring/ble/heartrate/application.cpp
XuGuohui: I defined a new type `typedef BLEAttribute* BLEAttribute;` to maintain an attribute in peer device using a pointer. An peer attribute is associated with connection status, when disconnected the attribute should not be valid any more. If let user have a copy of peer attribute, then the copied attribute cannot be updated internally.

user/tests/wiring/ble/heartrate_c/application.cpp
XuGuohui: I have defined three kinds of `getValue()`:
```
    int getValue(uint8_t* buf, uint16_t* len) const;
    int getValue(String& str) const;

    template<typename T> int getValue(T* val) const {
        uint16_t len = sizeof(T);
        return getValue(static_cast<uint8_t*>(val), &len);
    }
```

user/tests/wiring/ble/heartrate_c/application.cpp
XuGuohui: Once the target device being found, it then breaks the scan results iteration.

wiring/inc/spark_wiring_ble.h
```diff
@@ -301,12 +306,6 @@ class BLEClass {
     BLEClass();
     ~BLEClass();
 
-    /* Initialize BLE stack and add pre-defined local attributes. */
-    int on(void);
-
-    /* Turn BLE off. */
-    int off(void);
```
towynlin: Sorry for the confusion! Please KEEP `on()` and `off()` defined. I only removed them because in our simple example apps, there should be no need to call `on()` since BLE will be on by default. Big 👍 that users may want to turn the radio off in some low power situations. On the other hand, if it simplifies things to not have to implement this now, we could save on and off for a later release.

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+void setup() {
+}
+
+void loop() {
+    if (BLE.connected()) {
+        int newHr;
+        heartrate->getValue(&newHr);
+    }
+    else {
+        size_t count = BLE.scan(results, 10);
```
towynlin: This should not be `size_t`. `size_t` represents the size of an object/array/allocation in _bytes_. I assume `BLE.scan()` is returning the number of scan results, and that their size in bytes is much bigger.

avtolstoy: Sorry for barging in, but `size_t` is a valid type to return for the number of items. stdlib uses it for its container types (see https://en.cppreference.com/w/cpp/types/size_t).

avtolstoy: One suggestion here is that if `BLE.scan()` may fail, I'd advise returning a signed type (such as simply `int` or perhaps `ssize_t`) to be able to retrieve a `system_error_t`.

towynlin: Interesting! I don't actually read that cppreference.com page as saying what you're saying, but I believe you — you're by far the more knowledgeable here.

I'd agree in recommending `int` for the return type of `BLE.scan()`.

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+                if (found) {
+                    peer = BLE.connect(results[i].address());
+
+                    if (BLE.connected(peer)) {
+                        peer->attribute("heartrate", &heartrate);
+                        heartrate->onDataReceived(heartrateUpdated);
+                    }
+                }
+
+                break;
```
towynlin: Either the break is misplaced or there's no need for a `for` loop. As currently written, we always break on the first iteration.

XuGuohui: Oh, this is a mistype. the `break` should be placed at the end of `if(found)` case.

XuGuohui:
New PR here: https://github.com/particle-iot/device-os/pull/1728

PR #1712, opened 2019-03-06
[CI] Build time optimizations

avtolstoy:
### Problem

CI builds take too long.

### Solution

I don't want them to take that long to finish. This PR introduces some optimizations to the CI builds:
1. The number of concurrent jobs is set to 5, with the job build times balanced. You might ask why 5: unfortunately for some reason most of the times Travis doesn't start more than 5 build jobs at a time.
2. `enumerate_build_matrix.sh` now uses `parallel` tool to utilize both cores available on Travis workers
3. It is now possible to build DeviceOS out of tree (using `BUILD_PATH_BASE`). This feature is used in `enumerate_build_matrix.sh` where each build job has its own build directory
4. Fixes `newhal` platform builds
5. Fixes compilation issues with new gcc versions

---

- [Internal] [CI] Build time optimizations [#1712](https://github.com/particle-iot/device-os/pull/1712)
- [Bug fix] [wiring] Fixed a potential (but unlikely due to bounds checking) buffer overflow in time formatting function [#1712](https://github.com/particle-iot/device-os/pull/1712)


wiring/src/spark_wiring_time.cpp
```diff
@@ -343,13 +343,13 @@ String TimeClass::timeFormatImpl(tm* calendar_time, const char* format, int time
     strcpy(format_str, format);
     size_t len = strlen(format_str);
 
-    char time_zone_str[10];
+    char time_zone_str[16];
```
technobly: These changes to spark_wiring_time.cpp seem out of place for this PR... was this part of `Fixes compilation issues with new gcc versions`?  What was the issue here? Just making sure we don't need to include a bug fix note in the changelog.

ci/enumerate_build_matrix.sh
```diff
@@ -12,7 +12,28 @@ set -x # be noisy + log everything that is happening in the script
...
+  make_directory="$PWD/$1"
+  echo
+  echo '-----------------------------------------------------------------------'
+  echo "Running ${@:3} in $1, build directory ${build_directory}"
+  mkdir -p "${build_directory}"
+  cd "${BUILD_JOBS_DIRECTORY}"
+  eval ${@:3} -C "${make_directory}" BUILD_PATH_BASE="${build_directory}"
+  # I didn't want to deal with another build type, so we are specifically dealing
+  # with newhal build here
+  if echo "${@:3}" | grep --quiet "newhal"; then
```
technobly: Can you describe why so many changes were necessary for `hal/newhal` and is the above just moving the `newhal` build to the top of the process before the jobs run in parallel?  Are we still building the newhal platform?

FYI, this test for `has no sections` as a way to gauge pass/fail seems to have broken down... looking back at a previous build:

```
 make -s clean all PLATFORM=newhal COMPILE_LTO=n
runing matrix PLATFORM= MODULAR_PLATFORM= PLATFORM_BOOTLOADER=
-----------------------------------------------------------------------
In file included from ../wiring/inc/spark_wiring_cloud.h:32:0,
                 from ../wiring/inc/spark_wiring.h:51,
                 from ./inc/application.h:40,
                 from src/application.cpp:21:
../wiring/inc/spark_wiring_async.h:32:2: error: #error "std::atomic is not always lock-free for required types"
 #error "std::atomic is not always lock-free for required types"
  ^
make[1]: *** [../build/target/user/platform-60000/src/src/application.o] Error 1
../build/module.mk:277: recipe for target '../build/target/user/platform-60000/src/src/application.o' failed
../build/recurse.mk:11: recipe for target 'user' failed
make: *** [user] Error 2
++ echo 2
++ grep 'has no sections'
+ HAS_NO_SECTIONS=
+ [[ ! -z HAS_NO_SECTIONS ]]
+ testcase
+ '[' 0 -eq 0 ']'
 ✓ SUCCESS 
```

Clearly it failed to build, but we passed it.

wiring/src/spark_wiring_time.cpp
avtolstoy: Potential buffer overflow detected by newer gcc resulting in a warning, resulting in an error and a failed build.

ci/enumerate_build_matrix.sh
avtolstoy: It's become out of date with the reset of the HALs (no implementation headers etc.) and we've simply never actually checked that it builds.

I agree that this whole `has no sections` thing seems weird and probably never worked. For some reason we are grepping the result code for that string, instead of the stdout/stderr.

avtolstoy: And yes, we are building `newhal` normally like all the other things. You can see that it's added to the list of jobs [here](https://github.com/particle-iot/device-os/pull/1712/files/a7215e55fe1d29e858e9bae3196fbcd42d2ef26c#diff-a3c50403e095e3bff90f2e7ed13f36daR85) which are then later run in parallel [here](https://github.com/particle-iot/device-os/pull/1712/files/a7215e55fe1d29e858e9bae3196fbcd42d2ef26c#diff-a3c50403e095e3bff90f2e7ed13f36daR166)

wiring/src/spark_wiring_time.cpp
technobly: I see, is that if time_zone happened to be INT_MAX and the format string grew beyond the 3 and 2 placeholders?  Could that happen if a user accidentally supplied a really large time zone value?

avtolstoy: I didn't get into too much details. The compiler was saying that at maximum that string can be `11`, so I grew the buffer to `16` just in case to make it happy. This was gcc `7.3.0` on my host machine.

ci/enumerate_build_matrix.sh
technobly: `has no sections` was an additional check for error, since the result code of 0 with `has no sections` is not a passing case: https://github.com/particle-iot/device-os/commit/b0dde1cb0d0faa54c1f4dfef1b1c22a62b6daead#diff-a3c50403e095e3bff90f2e7ed13f36da

Looks like Mat chopped out the pass/fall portion later during the parallel builds (not sure why).

technobly: ^ Unless I'm misinterpreting how `testcase` works... 

From the build logs... I'm not sure newhal is compiling (notice no output under it):
```
-----------------------------------------------------------------------
Running runmake PLATFORM="newhal" COMPILE_LTO="n" in main, build directory /firmware/build/jobs/0
-----------------------------------------------------------------------
Running runmake DEBUG_BUILD="n" PLATFORM="xenon" COMPILE_LTO="n" SPARK_CLOUD="y" in main, build directory /firmware/build/jobs/1
Creating /firmware/build/jobs/1/user-part/platform-14-m/module_user_memory.ld ...
Creating /firmware/build/jobs/1/user-part/platform-14-m/module_user_memory.ld ...
  11012	    152	   1060	  12224	   2fc0	/firmware/build/jobs/1/user-part/platform-14-m/user-part.elf
 ✓ SUCCESS 
```

avtolstoy: The builds are run in parallel and the stdout/stderr is interleaved. It's there somewhere: https://travis-ci.org/particle-iot/device-os/jobs/503331630#L1367

wiring/src/spark_wiring_time.cpp
technobly: Ooh fancy! gcc version what?  _Shhh..._ xD

Just looking for potential reason to add something helpful in the changelog that we fixed some potential Panic SOS some user could be seeing when manipulating Time zone info.  If you think that's useful please suggest one.

avtolstoy: > [wiring] Fixed a potential buffer overflow in time formatting function
?

ci/enumerate_build_matrix.sh
avtolstoy: I've fixed the check in this commit: https://github.com/particle-iot/device-os/pull/1712/commits/a7c21d5f55af7c32f127b11aa581727df014670a

wiring/src/spark_wiring_time.cpp
technobly: 👍

ci/enumerate_build_matrix.sh
technobly: Still seems to be failing (and passing anyway), but we can fix this one later since there are more important things to test!!  I'll create a story for this one.
```
-----------------------------------------------------------------------
Running runmake PLATFORM="newhal" COMPILE_LTO="n" in main, build directory /firmware/build/jobs/0
arm-none-eabi-objcopy: error: the input file '/firmware/build/jobs/0/main/platform-60000/main.elf' has no sections
../build/module.mk:216: recipe for target '/firmware/build/jobs/0/main/platform-60000/main.bin' failed
make: *** [/firmware/build/jobs/0/main/platform-60000/main.bin] Error 1
 ✓ SUCCESS 
```

PR #1719, opened 2019-03-11
[Gen 3] Fix/wiring tests

technobly:
### Problem

Many tests needed updates and reliability fixes

### Solution

- [gen 3] increases pwr thread stack size to prevent SOS 10 on boron-som
- [test] fixes TEST=wiring/no_fixture_long_running PWM test
    Test was broken due to a bad ABS macro that was spitting out mostly negative values, also refactored to iterate from 20-90% Duty Cycle PWM in 5% steps, and made adjustments to keep it from hanging (20-30us pulses still too low to measure on Gen 3), and lightened the rigid accuracy calculations `Pulse width should be within 15% error margin resolution > 6, else 40%` and finally we don’t bother to test 2 and 3 bit PWM anymore since it’s very crude.
- [test] move TEST=wiring/no_fixture cellular Band Select tests to a dedicated test app
- [test] update upgrade-downgrade.sh to 1.1.0-rc.1
- [test] update country-updown.sh to 1.1.0-rc.1
- [ci] skip gen 3 modular debug builds
- [gen 3][test] fixes TEST=wiring/api
- Revert "travis: disabling tests and DEBUG_BUILD=y for now"
- [wiring] pmic: avoid calling Wire.begin() if it's already been initialized
- [system] power: do not report battery soc when there is no battery present or its state is unknown
- [wiring] network: Network.resolve() for platforms not-supporting getaddrinfo() (Gen 2)
- [wiring] network: pass hints with protocol = AF_UNSPEC and AI_ADDRCONFIG flag to getaddrinfo
- [system] network: some fixes in network_ready()
- [wiring] ipaddress: fixes an annoying pragma message on Gen 2 platforms by basing Wiring_IPv6 on HAL_PLATFORM_IPV6 instead of HAL_IPv6
- [wiring] network: fixes Network.resolve() to not return IPv4/IPv6 addresses when there is no IPv4/IPv6 connectivity respectively
- [system] network: extends network_ready() on gen 3 platforms to query for IPv4 or IPv6 connectivity specifically
- [test] fixes TEST=wiring/networking and adds smaller version TEST=app/networking for Core
- [test] adds Gen 3 support to SPI tests
- [Gen 3][SoM][test] no_fixture: adds more dummy reads after PWM period updates
- Fixes tolerance of DELAY01 test for Gen 3

### Steps to Test

Run all tests!

---

### Completeness

---

- [internal] [Gen 3] Fix/wiring tests [#1719](https://github.com/particle-iot/device-os/pull/1719)


avtolstoy:
PWM tests are still broken and are to be fixed separately in #1590

PR #1724, opened 2019-03-14
Enable on demand vitals publishing

zfields:
- [x] Run `.clang-format` to homogenize new code
- [x] Complete the unit-tests
- [ ] Draw up a test plan for end-to-end validation

**Related Stories:**

- [ch29075](https://app.clubhouse.io/particle/story/29075/allow-device-os-to-publish-vitals-as-single-shot)
- [ch29208](https://app.clubhouse.io/particle/story/29208/allow-device-os-to-publish-vitals-repeated)
- [ch29926](https://app.clubhouse.io/particle/story/29926/create-a-test-suite-for-wiring-api)
- [ch29930](https://app.clubhouse.io/particle/story/29930/create-doxygen-documents-for-the-api-and-support-class)


system/inc/system_dynalib_cloud.h
```diff
@@ -49,6 +49,7 @@ DYNALIB_FN(12, system_cloud, spark_sync_time_pending, bool(void*))
 DYNALIB_FN(13, system_cloud, spark_sync_time_last, system_tick_t(time_t*, void*))
 DYNALIB_FN(14, system_cloud, spark_set_connection_property, int(unsigned, unsigned, particle::protocol::connection_properties_t*, void*))
 DYNALIB_FN(15, system_cloud, spark_set_random_seed_from_cloud_handler, int(void (*handler)(unsigned int), void*))
+DYNALIB_FN(16, system_cloud, spark_send_description, bool(void*))
```
m-mcgowan: Let's paramaterize this with the description type so we can send application, system or vitals (and other types of descriptions.)

wiring/inc/spark_wiring_cloud.h
```diff
@@ -75,6 +75,8 @@ class CloudClass {
 
 
 public:
+    static const size_t PUBLISH_VITALS_DISABLE;
```
m-mcgowan: Why `size_t`?  Seems a little strange since the expected units is typically bytes,  but here we are expressing a duration. Is there a better type? In the HAL we define `system_tick_t` which might be a closer fit.

zfields: This is a bizarre copy/paste error or something. It shouldn't be all caps and the data type was meant to be boolean. The variable below it is the `period` and it was type `bool` instead of `size_t`. Who knows? I was probably cross-eyed while typing.

system/inc/system_dynalib_cloud.h
zfields: sounds good!

wiring/inc/spark_wiring_cloud.h
m-mcgowan: I see, so a mixup of the types. But is `size_t` the best type to use for duration? The units for `size_t` are bytes normally which doesn't seem to fit well with duration. On 64-bit platforms the duration will also be 64-bits which seems excessive! :-)

zfields: `size_t` is the platform-native unsigned integer. It will be the most performant. Any other size type would bear the unnecessarily overhead of being masked when moving in and/or out of the device registers. From an efficiency perspective, as long as it is large enough, `size_t` seems like the right type to me.

system/src/system_cloud.cpp
```diff
@@ -38,8 +42,37 @@
...
 
+namespace {
+
+using namespace particle::system;
+
+#if PLATFORM_ID == PLATFORM_SPARK_CORE
+  particle::NullTimer _vitals_timer;
+  VitalsPublisher<std::function<int(ProtocolFacade *, int, void *)>, particle::NullTimer> _vitals(spark_protocol_post_description, _vitals_timer);
+#else // not PLATFORM_SPARK_CORE
+  std::function<void(void)> _periodic_callback = []() -> void {
```
zfields: `&sp`

zfields: `spark_protocol_instance()`

zfields: The rest of this file is using `sp` an externally defined global from `system_cloud_internal.h`, which in turn sets it equal to `spark_protocol_instance()` in it's implementation.

Probably best to follow the pattern utilized in rest of the file. If the value of the `sp` abstraction is changed, then the behavior of the lambda will follow that of the file.

zfields:
Having issues with Electron...

communication/src/protocol.cpp
```diff
@@ -501,16 +501,66 @@ void Protocol::build_describe_message(Appender& appender, int desc_flags)
...
+		// what's going on.
+		SPARK_ASSERT(!appender.overflowed());
+	}
+
+	LOG(INFO,"Posting '%s%s%s' describe message", desc_flags & DESCRIBE_SYSTEM ? "S" : "",
+												  desc_flags & DESCRIBE_APPLICATION ? "A" : "",
+												  desc_flags & DESCRIBE_METRICS ? "M" : "");
+
+	const ProtocolError error = channel.send(message);
+	if (error == NO_ERROR && descriptor.app_state_selector_info
```
zfields: What is the `descriptor.app_state_selector_info`? This was copied from the other describe message and is not currently being used when posting the `DESCRIBE_METRICS` message.

I'm guessing it should probably be eliminated, but I wasn't sure if it needs to happen each time we send `(desc_flags & DESCRIBE_APPLICATION || desc_flags & DESCRIBE_SYSTEM)`.

@m-mcgowan Can you shine some light on this?

communication/src/build.mk
```diff
@@ -13,7 +13,7 @@ TARGET_SRC_PATH = src
 # C++ source files included in this build.
 CPPSRC += $(TARGET_SRC_PATH)/coap.cpp
 CPPSRC += $(TARGET_SRC_PATH)/handshake.cpp
-CPPSRC += $(TARGET_SRC_PATH)/spark_protocol.cpp
+CPPSRC += $(TARGET_SRC_PATH)/core_protocol.cpp
```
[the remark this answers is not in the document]
m-mcgowan: That implementation only used in the Core. It can be confusing to see SparkProtocol and Protocol instances and not know which one to use. (Zak hit this issue.) . Naming it CoreProtocol seemed to be more descriptive, although a comment in the class header would also have been useful. :-)

communication/src/core_protocol.cpp
```diff
@@ -815,74 +815,107 @@ void SparkProtocol::update_ready(unsigned char *buf, unsigned char token, uint8_
...
+    BufferAppender appender(buf+offset, QUEUE_SIZE-(offset + 2));
+    // diagnostics must be requested in isolation to be a binary packet
+    if (descriptor.append_metrics && (desc_flags == DESCRIBE_METRICS))
+    {
+        appender.append(char(0));	// null byte means binary data
+        appender.append(char(DESCRIBE_METRICS)); 	// uint16 describes the type of binary packet
+        appender.append(char(0));	//
+        const int flags = 1;		// binary
+        const int page = 0;
+        descriptor.append_metrics(append_instance, &appender, flags, page, nullptr);
```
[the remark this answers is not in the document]
m-mcgowan: Nope.  I suggested this should return a not implemented error.

wiring/inc/spark_wiring_cloud.h
```diff
@@ -242,6 +239,28 @@ class CloudClass {
...
+     * @arg \p s - Publish an initial message and subsequent messages every \p s seconds thereafter
+     * @param[in,out] reserved Reserved for future use (default value: \p nullptr).
+     *
+     * @returns \p system_error_t result code
+     * @retval \p system_error_t::SYSTEM_ERROR_NONE
+     * @retval \p system_error_t::SYSTEM_ERROR_IO
+     *
+     * @note The periodic functionality is not available for the Spark Core
+     */
+    int publishVitals(size_t period = particle::NOW);
```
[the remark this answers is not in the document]
m-mcgowan: I have raised this issue previously.  The implied units of `size_t` is bytes. I would change to `system_tick_t` if the raw units are milliseconds.

communication/src/build.mk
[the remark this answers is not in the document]
sergeuz: But the class is still called `SparkProtocol`, it's just the file that got renamed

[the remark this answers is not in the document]
m-mcgowan: Yeah...we should probably rename the class too and add a comment about how it's used and that new functionality should not be added.

communication/src/protocol.cpp
m-mcgowan: Although it's unused, I have no problem with it being there, apart from that it is copied and pasted. I'd prefer the code to be factored out.

system/inc/system_cloud.h
```diff
@@ -44,6 +44,10 @@ typedef enum
 	CLOUD_VAR_BOOLEAN = 1, CLOUD_VAR_INT = 2, CLOUD_VAR_STRING = 4, CLOUD_VAR_DOUBLE = 9
 } Spark_Data_TypeDef;
 
+namespace particle {
+    static const size_t PUBLISH_VITALS_NOW = static_cast<size_t>(-1);
```
m-mcgowan: I think as a general design preference, I would make a typedef that is named after the feature, e.g. `vitalsperiod_t`. This means the code becomes more descriptive and any arguments over what the actual type should be requires just one change :-)

system/src/system_cloud.cpp
```diff
@@ -38,8 +44,36 @@
...
+      if (!task) {
+          return;
+      }
+      task->func = [](ISRTaskQueue::Task* task) {
+          delete task;
+          spark_protocol_post_description(spark_protocol_instance(), particle::protocol::DESCRIBE_METRICS, nullptr);
+      };
+      SystemISRTaskQueue.enqueue(task);
+  }, false);
+  VitalsPublisher<Timer> _vitals(std::bind(spark_protocol_post_description, spark_protocol_instance(), particle::protocol::DESCRIBE_METRICS, nullptr), &_vitals_timer);
```
sergeuz: I suggest to use a C++ singleton here:
```cpp
VitalsPublisher<Timer>* vitalsPublisher() {
    static VitalsPublisher<Timer> vitals(...);
    return &vitals;
}
```
That would make the initialization independent of the order in which globals are initialized across multiple translation units in C++. As a bonus, the resources, such as a timer instance, will be allocated on demand, i.e. only if the application does publish device vitals.

test/unit_tests/cloud/publish_vitals.cpp
```diff
@@ -0,0 +1,247 @@
...
+    fakeit::When(Method(mock_timer, changePeriod)).AlwaysReturn(true);
+    fakeit::When(Method(mock_timer, dispose)).AlwaysReturn();
+    fakeit::When(Method(mock_timer, isActive)).AlwaysReturn(false);
+    fakeit::When(Method(mock_timer, reset)).AlwaysReturn();
+    fakeit::When(Method(mock_timer, start)).AlwaysReturn();
+    fakeit::When(Method(mock_timer, stop)).AlwaysReturn();
+
+    SECTION("When disabled, then the timer will be disposed.")
+    {
+        // Arrange
```
m-mcgowan: The comments separating the setup/execute/validate sections are good,  although Catch supports Given, When, Then blocks to make this explicit in the code, with the added benefit that these items can be nested for reuse. (Often individual tests have the common setup or execution steps.)  Here's an example: https://github.com/particle-iot/device-os/pull/1732/files#diff-deee133a44ce35e2149c0ac5feddc545R26

communication/src/build.mk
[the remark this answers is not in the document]
sergeuz: I'd suggest to keep everything under original names, and once the Core is finally and officially EOL'ed remove the old implementation from the source tree altogether

system/src/system_publish_vitals.h
```diff
@@ -0,0 +1,123 @@
...
+     */
+    typedef std::function<int(void)> publish_fn_t;
+
+    /**
+     * @brief Constructor
+     *
+     * @param[in] publish_fn The function used to send cloud messages
+     * @param[in] timer The timer used to schedule the period
+     */
+    VitalsPublisher(publish_fn_t publish_fn, Timer* timer);
```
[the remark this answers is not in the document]
zfields: That was the original implementation, but was modified to use a pointer for dependency injection to facilitate unit-testing.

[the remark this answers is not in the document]
sergeuz: So what was the problem with testing in the original code? You can easily say that there's something wrong with the design by looking at how this class is instantiated: https://github.com/particle-iot/device-os/pull/1724/files#diff-58e1af22bce41fb40b15b3822e26ce28R69 – note how the same `spark_protocol_post_description()` function gets passed to the constructor twice, as an `std::function` and indirectly via the timer

services/inc/logging.h
```diff
@@ -384,6 +384,8 @@ static const char* const _log_category = NULL;
 
 #else // LOG_DISABLE
 
+#warning  "Logging MACROs have been disabled! (-DLOG_DISABLE)"
```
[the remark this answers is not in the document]
zfields: It is there for the opposite reason. As a confirmation you had indeed enabled the option.

I had tried to compile with `-DDISABLE_LOG` and it was not working as expected.

I agree is does make a lot of noise, and I will remove it.

wiring/inc/spark_wiring_cloud.h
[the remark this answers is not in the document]
zfields: Yes, I agree. I switched to `system_tick_t` in other places. I must have gotten distracted and missed this API.

communication/src/core_protocol.cpp
```diff
@@ -815,74 +815,107 @@ void SparkProtocol::update_ready(unsigned char *buf, unsigned char token, uint8_
 }
 
 int SparkProtocol::description(unsigned char *buf, unsigned char token,
-                               unsigned char message_id_msb, unsigned char message_id_lsb, int desc_flags)
-{
-    buf[0] = 0x61; // acknowledgment, one-byte token
-    buf[1] = 0x45; // response code 2.05 CONTENT
+      unsigned char message_id_msb, unsigned char message_id_lsb, int desc_flags) {
```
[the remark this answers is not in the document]
zfields: Reverting this bit of formatting back to it's original form.

communication/src/protocol.cpp
zfields: The duplicate code has been refactored into a function to be called aposteriori.

communication/src/build.mk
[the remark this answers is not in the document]
zfields: I can confirm this creates a lot of friction for a new person. It should be easier for those familiar with the code base to follow the change, than those who are unfamiliar with the code base to understand the ambiguous naming scheme.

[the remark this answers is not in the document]
zfields: I will be extending the changes to make the `SparkProtocol` --> `CoreProtocol`.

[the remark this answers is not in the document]
sergeuz: Let's keep only relevant changes in this PR. The class can be renamed in a separate PR if necessary

system/inc/system_cloud.h
zfields: This was intended to be a `system_tick_t` and was an oversight. The PRD specifies the name as `particle::NOW`. Both of these changes have been applied.

communication/src/core_protocol.cpp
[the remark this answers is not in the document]
zfields: For completeness...

`Timer` is not implemented in the Core, and therefore the periodic functionality related to it was not implemented (system layer).

However, the technical foundation for the immediate invocation of these messages had already existed in the Core and simply needed to be invoked. No real work was done here.

Furthermore, I probably would not have done this minor work had I have understood the fact that `SparkProtocol` was exclusively for the Core (which is the inspiration of the refactoring of the name to `CoreProtocol`).

system/src/system_publish_vitals.h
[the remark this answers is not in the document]
zfields: @sergeuz Shouldn't the implementation of `publishVitals` incorporate the `ISRTaskQueue`?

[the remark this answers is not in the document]
sergeuz: I don't have any experience with FakeIt, but it does seem to allow modifying the behavior of existing objects: https://github.com/eranpeer/FakeIt/wiki/Quickstart#spying

I've modified one of your test cases in the following way and it seems to yield the same results:
```
particle::system::VitalsPublisher<particle::mock_type::Timer> vp;

fakeit::Mock<particle::mock_type::Timer> mock_timer(*vp.timer());

// Configure Mock
fakeit::When(Method(mock_timer, changePeriod)).AlwaysReturn(true);
fakeit::When(Method(mock_timer, dispose)).AlwaysReturn();
fakeit::When(Method(mock_timer, isActive)).AlwaysReturn(false);
fakeit::When(Method(mock_timer, reset)).AlwaysReturn();
fakeit::When(Method(mock_timer, start)).AlwaysReturn();
fakeit::When(Method(mock_timer, stop)).AlwaysReturn();

vp.disablePeriodicPublish();

fakeit::Verify(Method(mock_timer, stop));
```

The problem I have with the current code is that it still doesn't eliminate that weird decoupling and duplication of the functionality. The class still expects the external code to provide a publishing function via the Timer instance, and implements its own publishing function internally as well.

I don't see a unit test in this PR that would verify that the publisher function does indeed get called on a timer event, but besides of being essential for the functionality in this PR such a test would be a good illustration to the problem

system/src/system_publish_vitals.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "logging.h"
+#include "spark_protocol_functions.h"
+#include "system_threading.h"
+
+using namespace particle::system;
+
+template <class Timer>
+const typename VitalsPublisher<Timer>::publish_fn_t VitalsPublisher<Timer>::_publishVitals =
+    std::bind(spark_protocol_post_description, spark_protocol_instance(),
+              particle::protocol::DESCRIBE_METRICS, nullptr);
```
sergeuz: This can be a regular static function, there's no need to wrap `spark_protocol_post_description()` into an `std:function`

system/src/system_publish_vitals.cpp
```diff
@@ -0,0 +1,129 @@
...
+inline int spark_protocol_post_description(ProtocolFacade*, int, void*)
+{
+    spark_protocol_post_description_called = true;
+    return spark_protocol_post_description_result;
+}
+inline void ISRTaskQueue::enqueue(ISRTaskQueue::Task*)
+{
+}
+ISRTaskQueue SystemISRTaskQueue;
+#endif // UNIT_TEST
```
sergeuz: Can we implement these stubs in the unit test instead?

zfields: I tried that at first, it wouldn't link. I'll fiddle with CMake a little more and see what I can do.

system/src/system_publish_vitals.cpp
zfields: good point

zfields: good point, it has been removed

system/src/system_publish_vitals.h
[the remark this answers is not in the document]
zfields: To address a couple of points:

First, I was able to reproduce the example you posted above, but it is only possible when the Timer is a pointer and not a first class data member - I have no idea why. I have updated the PR to allocate the Timer on the heap, and then destroy it in the destructor if it was created and not passed in.

> The class still expects the external code to provide a publishing function via the `Timer` instance

I think this is better phrased, "_The class allows..._", because the pointer is not used in production - it is only there to allow for mocking and unit testing.

> I don't see a unit test in this PR that would verify that the publisher function does indeed get called on a timer event.

That sounds more like an end-to-end or integration test, because it is really testing that the `Timer` is functioning correctly in concert with the `VitalsPublisher`. To me, it would be more interesting to know the callback was set correctly. At that point, I would expect the timer to do it's job.

That being said, it's a fair point and that relationship needs to be proven one way or another. I'll do more research to figure out what this will take to implement.

user/tests/app/particle_publish_vitals/particle_publish_vitals.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "Particle.h"
+
+static const uint16_t ALT_LED = D7;
+size_t setup_ms = 0;
+bool publish = true;
+bool final_publish = false;
+
+// setup() runs once, when the device is first turned on.
+void setup()
```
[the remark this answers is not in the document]
zfields: I love that idea!

user/tests/app/particle_publish_vitals/particle_publish_vitals.cpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * Project: particle_publish_vitals
+ * Description: Confirm the functionality of the `Particle.publishVitals()` API
+ * Author: Zachary J. Fields
+ * Date: 13th March, 2019
```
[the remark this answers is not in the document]
zfields: That's the boilerplate from Workbench (used to create the test application). I will replace with something more general to Particle from elsewhere in the code.

PR #1725, opened 2019-03-14
Power leakage from external flash

YutingYou:
### Problem

1. We don't send enter sleep command to external flash before uninitializing it
1. We don't uninitialize external flash before entering STANDBY mode.

### Steps to Test
Test the fix on Xenon-SoM, the power consumption decrease from 8uA to 1uA in STANDBY mode.

### Example App

```c
void setup() {
}
void loop() {
    System.sleep(SLEEP_MODE_DEEP, 0);
}
```

### References

[CH29844]
[CH29689]

---

### Completeness

- [Bugfix] [Gen 3] Fixes a deadlock in `system_power_manager` and `i2c_hal` when exiting the sleep mode  [1725](https://github.com/particle-iot/device-os/pull/1725)
- [Enhancement] [Gen 3] QSPI flash is put into sleep mode and is deinitialized when entering STANDBY or STOP sleep mode [1725](https://github.com/particle-iot/device-os/pull/1725)

### DIFF
--- a/hal/src/nRF52840/core_hal.c
+++ b/hal/src/nRF52840/core_hal.c
@@ -56,6 +56,7 @@
 #include "exflash_hal.h"
 #include "flash_common.h"
 #include <nrf_pwm.h>
+#include "concurrent_hal.h"
 
 #define BACKUP_REGISTER_NUM        10
 static int32_t backup_register[BACKUP_REGISTER_NUM] __attribute__((section(".backup_registers")));
@@ -554,6 +555,9 @@ int HAL_Core_Enter_Stop_Mode_Ext(const uint16_t* pins, size_t pins_count, const 
     // Make sure we acquire exflash lock BEFORE going into a critical section
     hal_exflash_lock();
 
+    // Disable thread scheduling
+    os_thread_scheduling(false, NULL);
+
     // This will disable all but SoftDevice interrupts (by modifying NVIC->ICER)
     uint8_t st = 0;
     sd_nvic_critical_region_enter(&st);
@@ -563,7 +567,6 @@ int HAL_Core_Enter_Stop_Mode_Ext(const uint16_t* pins, size_t pins_count, const 
 
     // Disable external flash
     hal_exflash_uninit();
-    hal_exflash_unlock();
 
     uint32_t exit_conditions = STOP_MODE_EXIT_CONDITION_NONE;
 
@@ -916,6 +919,12 @@ int HAL_Core_Enter_Stop_Mode_Ext(const uint16_t* pins, size_t pins_count, const 
     // Re-enable USB
     HAL_USB_Attach();
 
+    // Unlock external flash
+    hal_exflash_unlock();
+
+    // Enable thread scheduling
+    os_thread_scheduling(true, NULL);
+
     return reason;
 }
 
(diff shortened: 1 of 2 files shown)


hal/src/nRF52840/core_hal.c
```diff
@@ -939,11 +939,20 @@ int HAL_Core_Execute_Standby_Mode_Ext(uint32_t flags, void* reserved) {
...
     // This will disable all but SoftDevice interrupts (by modifying NVIC->ICER)
     uint8_t st = 0;
     sd_nvic_critical_region_enter(&st);
 
+    // Disable SysTick
+    SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
+
+    // Disable external flash
+    hal_exflash_uninit();
+    hal_exflash_unlock();
```
avtolstoy: I think it should be fine to keep the lock here.

YutingYou: Yeah, the MCU won't execute code after entering standby mode. Removed it.

YutingYou:
To be doubly sure, I'll run the EEPROM test in stop mode, in case the sleep command impacts external flash.

YutingYou:
The external flash on dev board is different from the one on SoM, we need an additional read id command to wake up flash. And we should ensure the flash operation is alright after resuming from stop/standby mode.

Test device: Xenon, Xenon-SoM
Test application:
```
void setup() {
}

void loop() {
    delay(5000);
    Log.info("about to SLEEP_MODE_DEEP");

    // System.sleep(SLEEP_MODE_DEEP);
    System.sleep(D8, FALLING, 10, SLEEP_NETWORK_OFF);
    static uint32_t value = 1;
    EEPROM.write(5, value++);
    Log.info("data: %d", EEPROM.read(5));
}
```

YutingYou:
Disabling thread scheduler is also needed in stop/standby mode. The power thread has high priority than system thread, we disable some interrupts in stop mode function, if the thread scheduler occurs before we enable I2C interrupt, then the power thread will get stuck in I2C transaction.

hal/src/nRF52840/core_hal.c
```diff
@@ -938,6 +945,9 @@ int HAL_Core_Execute_Standby_Mode_Ext(uint32_t flags, void* reserved) {
     if (flags & HAL_STANDBY_MODE_FLAG_DISABLE_WKP_PIN) {
         return SYSTEM_ERROR_NOT_SUPPORTED;
     }
+ 
+    // Disable thread scheduling
+    os_thread_scheduling(false, NULL);
```
avtolstoy: This shouldn't be done. You are disabling scheduling and later on attempt to acquire the exflash lock which is implemented using a recursive mutex. If by this point it's being held by some other thread, it will never be released and everything will deadlock. `sd_nvic_critical_region_enter` will also disable thread scheduling, disabling the scheduler is unnecessary.

hal/src/nRF52840/core_hal.c
```diff
@@ -541,6 +542,9 @@ int HAL_Core_Enter_Stop_Mode_Ext(const uint16_t* pins, size_t pins_count, const
         }
     }
 
+    // Disable thread scheduling
+    os_thread_scheduling(false, NULL);
```
avtolstoy: This shouldn't be done. You are disabling scheduling and later on attempt to acquire the exflash lock which is implemented using a recursive mutex. If by this point it's being held by some other thread, it will never be released and everything will deadlock. `sd_nvic_critical_region_enter` will also disable thread scheduling, disabling the scheduler is unnecessary.

PR #1728, opened 2019-03-15
BLE wiring implementation.

XuGuohui:


user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+}
+
+void setup() {
+}
+
+void loop() {
+    if (BLE.connected()) {
+        int newHr;
+        heartrate->getValue(&newHr);
```
towynlin: Better would be `int newHr = heartrate.getValue();` Is that possible?

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+    else {
+        size_t count = BLE.scan(results, 10);
+
+        if (count > 0) {
+            for (uint8_t i = 0; i < count; i++) {
+                bool found = results[i].find(BLE_SIG_AD_TYPE_FLAGS);
```
towynlin: What does `find` do? What would the documentation say? Does it filter the results? Does it get the first result that matches? What are all the options that can be passed to it as a parameter?

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+                if (found) {
+                    peer = BLE.connect(results[i].address());
+
+                    if (BLE.connected(peer)) {
+                        peer->attribute("heartrate", &heartrate);
```
towynlin: Can this be `heartrate = peer.attribute("heartrate");` ?

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+                        heartrate->onDataReceived(heartrateUpdated);
```
towynlin: Can we use a `.` instead of `->`?

towynlin: Also, this looks to me like a bad pattern, leaving open timing issues. This event handler should get set up _before_ the attribute gets set, or else data could be sent by the peer and lost before the event handler is ready.

user/tests/wiring/ble/heartrate_central/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+                    }
+                }
+
+                break;
```
towynlin: This always breaks when `i` is `0`. The code should be better. In this implementation there is no need for a `for` loop. If you intend to only use the first search result, then the code should make that clear. If, on the other hand, you want to iterate through multiple matching results then you need a `for` loop.

user/tests/wiring/ble/ibeacon/application.cpp
```diff
@@ -0,0 +1,38 @@
...
+void setup() {
+    iBeacon beacon(1, 2, beaconUuid, -30);
+    BLE.advertisementData(beacon);
+
+    BLE.advertise();
+}
+
+void loop() {
+    delay(5000);
+    BLE.stopAdvertise();
```
towynlin: change to `stopAdvertising`

user/tests/wiring/ble/ibeacon/application.cpp
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+/* Includes ------------------------------------------------------------------*/
+#include "application.h"
+
+uint8_t beaconUuid[16] = {0x00,0x11,0x22,0x33,0x44,0x55,0x66,0x77,0x88,0x99,0xaa,0xbb,0xcc,0xdd,0xee,0xff};
+
+void setup() {
+    iBeacon beacon(1, 2, beaconUuid, -30);
+    BLE.advertisementData(beacon);
```
towynlin: This should work more like the heartrate peripheral — set up the iBeacon, no need for `advertisementData`, `advertise()` just does the right thing.

user/tests/wiring/ble/scanner/application.cpp
```diff
@@ -0,0 +1,62 @@
...
+    Serial.begin();
+}
+
+void loop() {
+    int count = BLE.scan(results, SCAN_RESULT_COUNT);
+
+    if (count > 0) {
+        for (int i = 0; i < count; i++) {
+            uint8_t buf[BLE_ADV_DATA_MAX];
+            size_t len = results[i].advData(buf, sizeof(buf));
```
towynlin: `advData` is not a wiring style name — "advertisementData"? I don't know what it stands for.

user/tests/wiring/ble/scanner/application.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+            Serial.println("Advertising data:");
+            for (size_t j = 0; j < len; j++) {
+                Serial.printf("0x%02x, ", buf[j]);
+            }
+            Serial.println("");
+
+            len = results[i].srData(buf, sizeof(buf));
```
towynlin: `srData` is not a wiring style name — what does it stand for?

XuGuohui: It stands for scan response data. How about `results[i].scanResponseData.data(buf, sizeof(buf))`?

user/tests/wiring/ble/scanner/application.cpp
XuGuohui: I'd prefer `AdvertisingData` as stated in BLE spec. How about `results[i].advertisingData().data(buf, sizeof(buf))`?

user/tests/wiring/ble/scanner/application.cpp
XuGuohui: We can also provide an API that can dump both advertising data and scan response data: `results[i].broadcastData(buf, sizeof(buf))`.

user/tests/wiring/ble/ibeacon/application.cpp
XuGuohui: The advertising data is not just an array. It must follow a particular pattern so that the BLE stack will adopt it. Here are the APIs to let user the advertising data safely: https://github.com/particle-iot/device-os/pull/1728/files?#diff-75b7ed256204b5221da93d699fee0429R219

user/tests/wiring/ble/ibeacon/application.cpp
XuGuohui: Will do. Same to `stopScanning`?

user/tests/wiring/ble/heartrate_central/application.cpp
XuGuohui: If the `heartrate` is a reference, then we can use `.`. But it's user's responsibility to check whether it is a valid reference or not, otherwise, it will result failure.

user/tests/wiring/ble/heartrate_central/application.cpp
XuGuohui: Yes, it can be. But the `heartrate` might be invalid if the specific attribute is not found. By the way, I just replace `attribute` with `characteristic, since attribute is a very under layer concept to user.

user/tests/wiring/ble/heartrate_central/application.cpp
XuGuohui: It checks whether the advertising contain the specific type of data. I have changed this API: https://github.com/particle-iot/device-os/pull/1728/files?#diff-75b7ed256204b5221da93d699fee0429R205

user/tests/wiring/ble/heartrate_central/application.cpp
XuGuohui: It worths a try.

user/tests/wiring/ble/heartrate_central/application.cpp
towynlin: The point here is that we need to provide the best practice example in these tests that everyone can copy-paste and follow, so we don't want to show users a buggy pattern that they will copy repeatedly.

user/tests/wiring/ble/ibeacon/application.cpp
towynlin: I see you've changed the underlying API :+1: — let's change this app and any other usage then resolve this conversation.

user/tests/wiring/ble/heartrate_central/application.cpp
towynlin: I see in the SDD you moved the `break` up so it breaks after the first relevant result is found. 👍 Do that here, and probably add a comment like "find first matching result"

user/tests/wiring/ble/scanner/application.cpp
towynlin: 👍 to "advertising". Let's avoid the user having to write the word "data" twice. How about `results[i].advertisingData(buf, sizeof(buf));`?

user/tests/wiring/ble/scanner/application.cpp
towynlin: Interesting idea! Except I think that would make the user have to figure out how to parse the two, and it would be better for us to handle that for them. As above, let's avoid them having to type the word "data" twice, like this: `results[i].scanResponseData(buf, sizeof(buf));`

user/tests/wiring/ble/scanner/application.cpp
XuGuohui: We use `results[i].advertisingData.get(buf, sizeof(buf));` instead.

XuGuohui:
https://github.com/particle-iot/device-os/pull/1740

PR #1729, opened 2019-03-15
fix mismatched types in arguments to std::min and std::max

elcojacobs:
<details>
  <summary><i>fixes type mismatch in std::min and std::max</i></summary>
</details>

### Problem

A modern compiler doesn't accept mismatched arguments to std::min and std::max. I'm crosscompiling (PLATFORM_GCC) with gcc 8.
You should try it. Some interesting warnings pop up for device-os.

### Solution

Use size_t for both.

### Completeness

- [ ] Added documentation

### DIFF
--- a/communication/src/chunked_transfer.cpp
+++ b/communication/src/chunked_transfer.cpp
@@ -279,8 +279,8 @@ ProtocolError ChunkedTransfer::handle_update_done(token_t token, Message& messag
     {
         updating = 2;       // flag that we are sending missing chunks.
         DEBUG("update done - missing chunks starting at %d", index);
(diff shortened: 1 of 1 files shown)


elcojacobs:
sorry, didn't see that this was fixed in develop with a typecast of the second argument
```
        chunk_index_t increase = std::max(unsigned(chunk_count*0.2), (unsigned)MINIMUM_CHUNK_INCREASE);	// ensure always some growth
        chunk_index_t resend_chunk_count = std::min(unsigned(chunk_count+increase), (unsigned)MISSED_CHUNKS_TO_SEND);
```

PR #1734, opened 2019-03-26
Docs for the check and scope guard macros

sergeuz:
### Problem

This PR add Doxygen documentation for the `CHECK()` and `SCOPE_GUARD()` macros.

### References

- [ch26297]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/services/inc/check.h
+++ b/services/inc/check.h
@@ -15,6 +15,107 @@
  * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
+/**
+ * @file check.h
+ *
+ * @section check_macros Check macros
+ *
+ * This header file defines a set of macros that simplify error handling in the Device OS code.
+ *
+ * Device OS doesn't use C++ exceptions, and, by convention, its internal and public API functions
+ * return a negative value in case of an error. All error codes are defined in the system_error.h
+ * file.
+ *
+ * Consider the following example:
+ * ```
+ * int foo() {
+ *     void* buffer = malloc(1024);
+ *     if (!buffer) {
+ *         return SYSTEM_ERROR_NO_MEMORY; // -260
+ *     }
+ *     ...
+ *     free(buffer);
+ *     return 0; // No error
+ * }
+ * ```
+ * When invoking such a function, the calling function should properly handle all possible errors.
+ * In the most typical scenario, it simply forwards the error to its own calling function:
+ * ```
+ * int fooBarBaz() {
+ *     int result = foo();
+ *     if (result == 0) {
+ *         result = bar();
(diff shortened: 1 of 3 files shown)


services/inc/check.h
```diff
@@ -15,6 +15,107 @@
...
+ * int fooBarBaz() {
+ *     // This macro makes fooBarBaz() return SYSTEM_ERROR_NO_MEMORY if foo() fails to allocate its
+ *     // internal buffer
+ *     CHECK(foo());
+ *     CHECK(bar());
+ *     CHECK(baz());
+ *     return 0;
+ * }
+ * ```
+ * The `CHECK()` macro expands to a code that evaluates the expression passed to the macro, and if
```
zfields: Please refactor this sentence. The meaning of "calling code" vs "function that invokes the macro" were not immediately obvious. I had to read it twice to understand the MACRO forces an early return from the calling function.

services/inc/check.h
```diff
@@ -15,6 +15,107 @@
...
+ * int receiveMessage(int socket) {
+ *     char buffer[128];
+ *     // It is an error to use non-Device OS functions with CHECK()
+ *     size_t size = CHECK(lwip_read(socket, buffer, sizeof(buffer)));
+ *     LOG(TRACE, "Read %u bytes", (unsigned)size);
+ *     ...
+ *     return 0;
+ * }
+ * ```
+ * This header file also defines the `CHECK_TRUE()` and `CHECK_FALSE()` macros, which can be used
```
zfields: Please add the following phrase to begin this section.
> "For checking non-Device OS functions, this header also defines..."

Otherwise it seems to contradict the note above about not using our MACROs against non-Particle code.

services/inc/scope_guard.h
```diff
@@ -15,16 +15,124 @@
...
+ *     CHECK(stream->read(buffer, bufferSize));
+ *     ...
+ *     return 0;
+ * }
+ * ```
+ * Internally, `SCOPE_GUARD()` declares an object that schedules an arbitrary user-provided code
+ * for execution at the end of the enclosing scope of the object. In the above example, the buffer
+ * allocated via `malloc()` will be automatically freed by the scope guard when the function returns
+ * either successfully or with an error.
+ *
```
zfields: If there are any ways that this macro syntax could be abused or used incorrectly, it should be noted here.

services/inc/check.h
sergeuz: That note was about the `CHECK()` macro specifically though. Also, Device OS still has a lot of functions that return `bool`, or otherwise don't follow the convention.

PR #1748, opened 2019-04-12
Runtime logging configuration

sergeuz:
### Problem

This PR reimplements the control requests for runtime logging configuration using protobuf (the original implementation used JSON).

The most part of the logging framework is implemented in the user module. In order to reduce the overhead for user applications, the parsing of protobuf messages is performed in the system module. A deserialized configuration is then passed to the user module via a thin system API layer (see `system_logging.h`).

Note that the capability to configure the logging at runtime is disabled by default. The application needs to set a feature flag in order to enable it (`FEATURE_CONFIGURABLE_LOGGING`). Unfortunately, we can't enable this feature by default and provide an option to disable it, because in that case the linker won't be able to optimize the unused code out. A solution would be to move the logging framework to the system module entirely.

### Steps to Test

**TODO**

### References

- [ch31052]


wiring/src/spark_wiring_logging.cpp
```diff
@@ -472,6 +284,12 @@ int spark::detail::LogFilter::nodeIndex(const Vector<Node> &nodes, const char *n
 
 // spark::StreamLogHandler
 void spark::StreamLogHandler::logMessage(const char *msg, LogLevel level, const char *category, const LogAttributes &attr) {
+    // TODO: Move this check to a base class (see also JSONStreamLogHandler::logMessage())
+#if PLATFORM_ID != PLATFORM_GCC
+    if (stream_ == &Serial && Network.listening()) {
```
monkbroc: Will this also quiet logs sent to Serial1? I know Brett relies on Serial1 logging for deep dive into connectivity issues, even in listening mode.

sergeuz: No, this change will only affect logs sent to `Serial`

m-mcgowan: Bravo, this is a good bandaid to fixing broken listening mode!

wiring/src/spark_wiring_logging.cpp
```diff
@@ -629,7 +568,8 @@ Print* spark::DefaultOutputStreamFactory::createStream(const char *type, const J
 void spark::DefaultOutputStreamFactory::destroyStream(Print *stream) {
 #if PLATFORM_ID != 3
     if (stream == &Serial) {
-        Serial.end();
+        // FIXME: Uninitializing the primary USB serial detaches a Gen 3 device from the host
```
m-mcgowan: Given it's a Gen3 issue, should we make it platform conditional?

wiring/inc/spark_wiring_logging.h
```diff
@@ -451,15 +453,13 @@ class LogHandlerFactory {
 public:
     virtual ~LogHandlerFactory() = default;
 
-    virtual LogHandler* createHandler(const char *type, LogLevel level, LogCategoryFilters filters, Print *stream,
-            const JSONValue &params) = 0; // TODO: Use some generic container or a buffer instead of JSONValue
-    virtual void destroyHandler(LogHandler *handler);
+    virtual LogHandler* createHandler(log_handler_type type, LogLevel level, LogCategoryFilters filters, Print* stream = nullptr) = 0;
```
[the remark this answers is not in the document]
sergeuz: It's technically a breaking change, however, the API was marked as experimental in the code and was never documented: [spark_wiring_logging.h#L449 (develop)](https://github.com/particle-iot/device-os/blob/develop/wiring/inc/spark_wiring_logging.h#L449)

wiring/src/spark_wiring_logging.cpp
```diff
@@ -345,7 +95,158 @@ const char* extractFuncName(const char *s, size_t *size) {
...
+    };
+    std::unique_ptr<Result> r(new(std::nothrow) Result());
+    CHECK_TRUE(r, SYSTEM_ERROR_NO_MEMORY);
+    r->result.version = LOG_SYSTEM_API_VERSION; // API version
+    const auto mgr = LogManager::instance();
+    const bool ok = mgr->enumFactoryHandlers([](const char* id, void* data) {
+        const auto r = (Result*)data;
+        Handler h = {};
+        h.handlerId = id;
+        if (!h.handlerId) {
```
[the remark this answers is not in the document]
sergeuz: `CHECK_TRUE()` returns an error code while this callback is expected to return `bool`

wiring/src/spark_wiring_logging.cpp
sergeuz: :+1:

system/inc/system_logging.h
```diff
@@ -0,0 +1,204 @@
...
+    const char* category; ///< Category name.
+    uint8_t level; ///< Logging level (a value defined by the `LogLevel` enum).
+} log_filter_list_item;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Set the command handler callback.
```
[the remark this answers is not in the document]
sergeuz: Will add a comment :+1:

PR #1749, opened 2019-04-12
Fix/unit tests pre-releases

technobly:
### Problem

Unit Tests do not expect Alpha or Beta pre-release strings in addition to RC.

### Solution

Give the Unit-Test some help in determining which type of string format is is testing

### Steps to Test

Run Unit Tests from the root directory (firmware $):
```
source ci/install_boost.sh
ci/build_boost.sh
cd user/tests/unit && make -s clean all
```

All tests should pass with these version strings:
```
1.2.0-alpha.1
1.2.0-beta.1
1.2.0-rc.1

===============================================================================
All tests passed (144673 assertions in 152 test cases)
```

---

### Completeness

- [x] N/A Added documentation


sergeuz:
The test will still fail if the a given Device OS version has alpha, beta and candidate releases. How about changing our internal version numbering scheme instead? For example:
```cpp
#define SYSTEM_VERSION_1_2_0_ALPHA_1 0x010200a1
#define SYSTEM_VERSION_1_2_0_BETA_1  0x010200b1
#define SYSTEM_VERSION_1_2_0_RC_1    0x010200c1
#define SYSTEM_VERSION_1_2_0         0x010200ff
```

technobly:
@sergeuz can you explain what you mean by this "The test will still fail if a given Device OS version has alpha, beta and candidate releases."

I don't see how a single release can have a version string that is `-alpha.x` AND `-beta.x` AND `-rc.x`.  It's mutually exclusive.

The plan is to treat the lowest byte as a counter starting at 1 for alpha.1, and incrementing for as many alpha's, then beta's then rc's as necessary in that order only.  We talked about changing the default release to use 0xFF to ensure it was greater than all pre-release types.

I also don't see a reason to change the formatting of the SYSTEM_VERSION defines.

### Example

```c
#define SYSTEM_VERSION_v120ALPHA1 0x01020001
#define SYSTEM_VERSION_v120ALPHA2 0x01020002
#define SYSTEM_VERSION_v120BETA1  0x01020003
#define SYSTEM_VERSION_v120BETA2  0x01020004
#define SYSTEM_VERSION_v120RC1    0x01020005
#define SYSTEM_VERSION_v120RC2    0x01020006
#define SYSTEM_VERSION_v120       0x010200ff
#define SYSTEM_VERSION            SYSTEM_VERSION_v120
```

sergeuz:
> can you explain what you mean by this "The test will still fail if a given Device OS version has alpha, beta and candidate releases."

@technobly Given the following defines, your test will incorrectly expect the version string to be `1.2.0-beta.3`:
```cpp
#define SYSTEM_VERSION_v120ALPHA1 0x01020001
#define SYSTEM_VERSION_v120ALPHA2 0x01020002
#define SYSTEM_VERSION_v120BETA1  0x01020003
#define SYSTEM_VERSION            SYSTEM_VERSION_v120BETA1
```

technobly:
Thanks for the clarification. Hmm... on one hand I'd like to keep the door open to easily incrementing the lowest byte so we have plenty of room to do so.  0.8.0 made it up to rc.27 which wouldn't have room with the above scheme.  If we reserved the TWO MSB's for 4 variations of pre-releases that would allow for 64 different types of alpha, beta and rc.  Just a bit more to think about when crafting these version defines (but the test would catch any errors in encoding and it has in the past!), and you wouldn't be able to definitively say BETA1 came after the last ALPHAx for example, but I'm not sure when you'd need to make a test like that.  Being able to test in user firmware if the system version is definitively alpha, beta, rc or default seems very powerful though for obvious reasons.

### How about:

```c
#define SYSTEM_VERSION_v120ALPHA1 0x01020001
#define SYSTEM_VERSION_v120ALPHA2 0x01020002
#define SYSTEM_VERSION_v120BETA1  0x01020041
#define SYSTEM_VERSION_v120BETA2  0x01020042
#define SYSTEM_VERSION_v120RC1    0x01020085
#define SYSTEM_VERSION_v120RC2    0x01020086
#define SYSTEM_VERSION_v120       0x010200ff
// We can reserve 0xXXYYZZC0 ~ 0x......FE for a different type of pre-release,
// or let RC's use 127, where default is 0x......FF
#define SYSTEM_VERSION            SYSTEM_VERSION_v120
```

sergeuz:
Using bits 6-7 of a version number to encode the version type sounds good to me as well. Let's also introduce a convenience macro to encode those version numbers, for example:
```cpp
#define SYSTEM_VERSION_v120ALPHA1 SYSTEM_VERSION_ALPHA(1, 2, 0, 1)
#define SYSTEM_VERSION_v120ALPHA2 SYSTEM_VERSION_ALPHA(1, 2, 0, 2)
#define SYSTEM_VERSION_v120BETA1  SYSTEM_VERSION_BETA(1, 2, 0, 1)
```

PR #1754, opened 2019-04-22
Shared Gen3 module code

monkbroc:
### Problem

Compiling Gen 3 Device OS would put intermediate build files outside of the `build/target` directory or `$APPDIR/target` directory (when compiling a customer application).

The source code for system-part1 and user-part modules for Gen 3 devices was in `../../../modules/shared/nRF52840/src/system-part1` and `../../../modules/shared/nRF52840/src/user-part` which led to the build files being put in a relative folder `modules/shared/nRF52840` instead of `target/system-part1/platform-X-m`.

### Solution

Include source files for `system-part1` and `user-part` modules for `argon`, `boron`, `xenon` that are only stubs that include the actual sources from the `modules/shared/nRF52840/inc/system-part1` and `modules/shared/nRF52840/inc/user-part` directories.

### Steps to Test

- In Workbench, open a directory with a Particle application (say `blink`)
- In Workbench, run command `Particle: Launch Compiler Shell`
- In the shell, run `DEVICE_OS_PATH=$HOME/device-os make -f $PARTICLE_MAKEFILE compile-all PLATFORM=xenon`
- Notice that the `target` directory contains `blink.bin` and `target/obj/src` contains some `.o` and `.o.d` files.
- Notice that there is no unexpected `shared` directory inside the Particle application directory

### References
- [CH31009]

### Completeness

- [INTERNAL] [Gen 3] Fixes a build system issue that caused object files to be created outside build directory [#1754](https://github.com/particle-iot/device-os/pull/1754)


avtolstoy:
[CH31009]

PR #1757, opened 2019-04-29
Sets C++ and C standard versions in a single location (C++14 and C11)

avtolstoy:
### Problem

1. We'd like to upgrade the C++ standard we use to C++14
2. Almost every module seems to set its own C++ and/or C standard version

### Solution

This PR removes all mentions of `-std=xxx` from the makefiles and adds a single `build/lang-std.mk` file setting C++ and C standard version (C++14 with GNU extensions, C11 with GNU extensions), which is then included by the other makefiles.

As we also have CMake-based unit-tests, exactly the same versions are separately set in `test/CMakeLists.txt` with notes in both `lang-std.mk` and `CMakeLists.txt` about keeping the standard versions in sync.

### Steps to Test

N/A

### Example App

N/A

### References

- [CH31589]

---

### Completeness

- [ ] Added documentation

---

- [enhancement] Sets C++ and C standard versions in a single location (C++14 and C11) [#1757](https://github.com/particle-iot/device-os/pull/1757)


avtolstoy:
`lang-std.mk` is always included by `arm-tlm.mk`, which is always included when building DeviceOS. Things that don't use regular DeviceOS build system (e.g. unit test) specifically include `lang-std.mk`.

PR #1758, opened 2019-04-29
[Gen 3] Always check ongoing RX DMA transaction when reporting number of bytes available in RX buffer

avtolstoy:
### Problem

`SerialX.available()` may report smaller number of bytes available in RX buffer than actually received, until the data is read out. There is no data loss though.

This is a "bug" in how we handle partial DMA transfers. We commit pending DMA data only if the RX buffer is empty, so you would see a smaller number in `available()` that wouldn't change until read out and then partially completed DMA transfer would be commited.

Originally issue was reported in https://community.particle.io/t/error-serial1-available-argon/48380/4

### Solution

Always check ongoing RX DMA transaction when reporting number of bytes available in RX buffer.

### Steps to Test

Run the `wiring/serial_loopback2` test.

### Example App

N/A

### References

- https://community.particle.io/t/error-serial1-available-argon/48380/4
- [CH31890]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/usart_hal.cpp
+++ b/hal/src/nRF52840/usart_hal.cpp
@@ -266,13 +266,13 @@ public:
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         RxLock lk(uarte_);
         ssize_t d = rxBuffer_.data();
-        if (d == 0 && receiving_) {
+        if (receiving_) {
             const ssize_t toConsume = timerValue() - rxConsumed_;
             if (toConsume > 0) {
                 rxBuffer_.acquireCommit(toConsume);
                 rxConsumed_ += toConsume;
+                d += toConsume;
             }
-            d += toConsume;
         }
         return d;
     }
--- a/user/tests/wiring/serial_loopback2/README.md
+++ b/user/tests/wiring/serial_loopback2/README.md
@@ -1,0 +1,12 @@
+## Serial testing hardware requirements
+
+Connect a jumper to TX and RX pins i.e. TX-RX lines should be shorted.
+```
+TX <-------> RX
+```
+
+## Flashing the wiring/serial_loopback2 test application
+
+```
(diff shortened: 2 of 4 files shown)


avtolstoy:
> So as you can see, something much more problematic is going on …
I honestly don't. Could you create an issue and provide some example of how this can be reproduced?

Are you using flow control? Is your Xenon doing something else other than reading the Serial port? Do you use threading? If no flow control is enabled and Xenon also does other things you may simply be not reading out the data the internal RX buffer fast enough causing overflows and errors you are seeing. You could try using `SYSTEM_MODE(MANUAL)` and not establishing the cloud connection and see if that resolves the issue for you.

We have a set of tests that verify the USART implementation and also on both the Boron and Argon the connection to the NCP (cellular modem or ESP32 for WiFi) relies on USART, so I would expect we would have caught any issues causing data loss/corruption.

jimmie11:
@avtolstoy 

I added in the community the post below but not in this forum.

To answer your questions,  I am not using flow control.  The Xenon is only reading from the serial sensor, but I keep getting the panic resets per my post below.   I have tried SYSTEM_MODE(MANUAL) and SYSTEM_THREAD(ENABLED); but it did not help.

I think that most likely reason is buffer overflow.  BTW, when the same code reads from an i2C sensor (that is updating at 80 Hz), the code NEVER crashes.  This problem occurs when interfacing to a fast serial sensor (190 Hz in my case).

I hope this helps.
================================================================

_I moved the blocking loop in my code to the loop() routine from setup(). When I did that, I found that I am reading the full sensor update speed. So it looks like the OS is doing something during setup() which takes some processor time even with SYSTEM_THREAD(ENABLED);. I want to thank Particle for the 1.2Beta which is a major improvement.

I am still getting some panic, hard_fault every about 15-25 minutes (sometimes after a couple of hours). There remain issues (apparently due to UART buffer overflow) during setup() where the sensor readings are missing/incorrect. In my application, this is serious because some major operational parameters are re-calculated during each restart. I would appreciate the community’s help for a workaround.

Hopefully these remaining issues will be resolved as 1.2 matures._

PR #1759, opened 2019-04-29
Integrate cellular network vitals data into `DESCRIBE_x` message

zfields:
[ch29061/cellular vitals](https://app.clubhouse.io/particle/story/29061/merge-cellular-network-vitals-data-into-develop-branch)


hal/src/boron/cellular_hal.cpp
```diff
@@ -96,8 +96,9 @@ hal_net_access_tech_t fromCellularAccessTechnology(CellularAccessTechnology rat)
         return NET_ACCESS_TECHNOLOGY_UTRAN;
     case CellularAccessTechnology::LTE:
     case CellularAccessTechnology::E_UTRAN:
-    case CellularAccessTechnology::EC_GSM_IOT: // FIXME
         return NET_ACCESS_TECHNOLOGY_LTE;
+    case CellularAccessTechnology::EC_GSM_IOT:
```
technobly: I think we may want to move `CellularAccessTechnology::E_UTRAN:` down here as well in the  NET_ACCESS_TECHNOLOGY_LTE_IOT result, since it's Cat-NB1.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,68 @@ int SaraNcpClient::initParser(Stream* stream) {
...
         CHECK_TRUE(r >= 1, SYSTEM_ERROR_UNKNOWN);
         // Home network or roaming
-        if (val[r - 1] == 1 || val[r - 1] == 5) {
+        if (val[0] == 1 || val[0] == 5) {
             self->creg_ = RegistrationState::Registered;
         } else {
             self->creg_ = RegistrationState::NotRegistered;
         }
         self->checkRegistrationState();
+        // Cellular Global Identity (partial)
```
technobly: May want to add a if (r >= 4) check here, and/or ensure int val[4] is memset to 0 or give val[2] and val[3] some proper `unknown` defaults.

hal/src/electron/cellular_internal.cpp
```diff
@@ -54,9 +54,11 @@ cellular_result_t cellular_signal_impl(CellularSignalHal* signal, cellular_signa
             signalext->rat = NET_ACCESS_TECHNOLOGY_UTRAN;
             break;
         case ACT_LTE:
+            signalext->rat = NET_ACCESS_TECHNOLOGY_LTE;
```
technobly: These changes go along with the change above to move Cat-NB1 down into NET_ACCESS_TECHNOLOGY_LTE_IOT

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -845,6 +847,11 @@ bool MDMParser::init(DevStatus* status)
     sendFormated("AT+CIMI\r\n");
     if (RESP_OK != waitFinalResp(_cbString, _dev.imsi))
         goto failure;
+    // Reformat the operator string to be numeric
+    // (allows the capture of `mcc` and `mnc`)
+    sendFormated("AT+COPS=3,2\r\n");
```
technobly: Does this command affect the response of AT+COPS?  Does it still return <mode> 0, 2, or 4 if those are set and not 3?

system/src/system_network_compat.cpp
```diff
@@ -255,6 +255,86 @@ class NetworkAccessTechnologyDiagnosticData: public AbstractIntegerDiagnosticDat
...
+    NetworkCellularCellGlobalIdentityMobileCountryCodeDiagnosticData()
+        : AbstractIntegerDiagnosticData(
+              DIAG_ID_NETWORK_CELLULAR_CELL_GLOBAL_IDENTITY_MOBILE_COUNTRY_CODE,
+              DIAG_NAME_NETWORK_CELLULAR_CELL_GLOBAL_IDENTITY_MOBILE_COUNTRY_CODE)
+    {
+    }
+
+    virtual int get(IntType& val)
+    {
+        const int some_value = 310;
```
technobly: I'm guessing these 4 get() methods are WIP

hal/src/boron/cellular_hal.cpp
zfields: I was reading and reading, and the EUTRA has been around so long, I couldn't tack it down as to what it was. I will gladly move it down.

system/src/system_network_compat.cpp
zfields: Yes, using them to get the framing out there. They will be changed.

hal/src/boron/cellular_hal.cpp
technobly: Cat-M1 is somewhat hard to know based on name alone, since there are some footnotes in the R4 AT command manual, but Cat-NB1 is a little clearer:

### COPS
Screen Shot 2019-04-29 at 12 51 46 PM

Screen Shot 2019-04-29 at 12 53 27 PM

zfields: Fix is in.

hal/src/boron/network/sara_ncp_client.cpp
zfields: Initialized `val` array...
```cpp
int val[4] = {-1,-1,-1,-1};
```

hal/src/electron/cellular_internal.cpp
zfields: :+1:

hal/src/electron/modem/mdm_hal.cpp
zfields: Here is a sample AT response:
```none
// +COPS: <mode>[,<format>,<oper>[,<AcT>]]
+COPS: 0,2,"310410",8
```

If you are asking what the behavior is like when you force set your <oper> using `AT+COPS=1,2`, then I don't know the answer. I would appreciate it greatly if you know how and are comfortable testing.

I don't have multiple pieces of hardware, and I'm afraid to twiddle with those for fear of messing up a device.

hal/src/boron/network/sara_ncp_client.cpp
technobly: 👍

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,68 @@ int SaraNcpClient::initParser(Stream* stream) {
...
     parser_.destroy();
     CHECK(parser_.init(std::move(parserConf)));
+    // +CREG: <stat>[,<lac>,<ci>[,<AcTStatus>]]
     CHECK(parser_.addUrcHandler("+CREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %d,%d", &val[0], &val[1]));
+        int val[4];
+        // Parse as direct response (ignoring mode)
+        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %*d,%d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3]));
```
technobly: Looking at this more, <stat> would be val[1] for this direct response with verbose output <n>=2, whereas <stat> would be val[0] for the URC.

hal/src/electron/modem/mdm_hal.cpp
technobly: This looks to be ok then with a AT+COPS=0 automatic mode selected.  I'm assuming if you tried manual AT+COPS=4,2,"310410" it would also return similarly +COPS: 4,2,"310410",8 but with the mode noted as 4 (manual).  Was just wanting to check that it didn't return mode = 3.

hal/src/boron/network/sara_ncp_client.cpp
zfields: I'm specifically ignoring _<mode>_ `%*d`, so `val[0]` must always be _<stat>_.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,68 @@ int SaraNcpClient::initParser(Stream* stream) {
...
+    // +CREG: <stat>[,<lac>,<ci>[,<AcTStatus>]]
     CHECK(parser_.addUrcHandler("+CREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %d,%d", &val[0], &val[1]));
+        int val[4] = {-1,-1,-1,-1};
+        // Parse as direct response (ignoring mode)
+        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %*d,%d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3]));
+        // Reparse as URC
+        if (r <= 1) { r = CHECK_PARSER_URC(reader->scanf("+CREG: %d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3])); }
```
avtolstoy: Formatting. Let's not put code within the codeblock on the same line.

This `if` will invalidate the correct result of the previous scanf for `+CREG: <n>,<stat>` responses, causing `stat` to be parsed as `<n>` instead in the second `scanf`.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,68 @@ int SaraNcpClient::initParser(Stream* stream) {
...
+    // n=2     +CGREG: <stat>[,<lac>,<ci>[,<AcT>,<rac>]]
     CHECK(parser_.addUrcHandler("+CGREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CGREG: %d,%d", &val[0], &val[1]));
+        int val[4];
+        // Parse as direct response (ignoring mode)
+        int r = CHECK_PARSER_URC(reader->scanf("+CGREG: %*d,%d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3]));
+        // Reparse as URC
+        if (r <= 1) { r = CHECK_PARSER_URC(reader->scanf("+CGREG: %d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3])); }
```
avtolstoy: Formatting. Let's not put code within the codeblock on the same line.

This `if` will invalidate the correct result of the previous scanf for `+CREG: <n>,<stat>` responses, causing `stat` to be parsed as `<n>` instead in the second `scanf`.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,68 @@ int SaraNcpClient::initParser(Stream* stream) {
...
+    // +CEREG: <stat>[,[<tac>],[<ci>],[<AcT>][,<cause_type>,<reject_cause>[,[<Active_Time>],[<Periodic_TAU>]]]]
     CHECK(parser_.addUrcHandler("+CEREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CEREG: %d,%d", &val[0], &val[1]));
+        int val[4];
+        // Parse as direct response (ignoring mode)
+        int r = CHECK_PARSER_URC(reader->scanf("+CEREG: %*d,%d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3]));
+        // Reparse as URC
+        if (r <= 1) { r = CHECK_PARSER_URC(reader->scanf("+CEREG: %d,\"%x\",\"%x\",%d", &val[0], &val[1], &val[2], &val[3])); }
```
avtolstoy: Formatting. Let's not put code within the codeblock on the same line.

This `if` will invalidate the correct result of the previous scanf for `+CREG: <n>,<stat>` responses, causing `stat` to be parsed as `<n>` instead in the second `scanf`.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -198,8 +199,10 @@ MDMParser::MDMParser(void)
     inst = this;
     memset(&_dev, 0, sizeof(_dev));
     memset(&_net, 0, sizeof(_net));
-    _net.lac = 0xFFFF;
-    _net.ci = 0xFFFFFFFF;
+    _net.cgi.mobile_country_code = 0;
```
avtolstoy: zeroing here should be unnecessary, as we've just called `memset`

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,47 @@ int cellular_credentials_clear(void* reserved) {
...
+    const auto client = mgr->ncpClient();
+    CHECK_TRUE(client, SYSTEM_ERROR_UNKNOWN);
+
+    // Validate Argument(s)
+    if (nullptr == cgi_)
+    {
+        (void)reserved_;
+        result = SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+
```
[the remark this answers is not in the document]
zfields: I used the auto-formatting from `.clang-format` file, I'll switch it back to your style.

hal/inc/hal_cellular_global_identity.h
```diff
@@ -0,0 +1,12 @@
+#ifndef HAL_CELLULAR_GLOBAL_IDENTITY_H
+#define HAL_CELLULAR_GLOBAL_IDENTITY_H
+
+typedef struct
+{
+    uint16_t mobile_country_code;
```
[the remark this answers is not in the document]
zfields: This fields is this struct are defined by the 3GPP. It should be akin to a phone number `<country_code>-<area_code>-<prefix>-<customer>`.

I don't expect it to change too much, but I will check the web to see if a version is applied to these fields and sizes.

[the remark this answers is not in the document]
avtolstoy: This was a suggestion for ABI stability as we do for most of the structs nowadays in case we ever want to extend them and not in the sense that there are multiple variants defined by 3GPP.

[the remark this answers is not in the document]
avtolstoy: Actually, I would insist on this change.

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,47 @@ int cellular_credentials_clear(void* reserved) {
...
+    // Load cached data into result struct
+    else if (client->getCellularGlobalIdentity(&cgi))
```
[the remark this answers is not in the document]
zfields: I don't mind to change, but I'll share the motive behind the style...

This is an error waterfall pattern, and it allows you to achieve a similar pattern to your `CHECK()` macros.

It prevents deep conditional nesting while providing blocks to validate parameters and calls. It also keeps the functional logic of your method _(the only part people care about)_ in the `else` at the end.

hal/inc/hal_cellular_global_identity.h
```diff
@@ -0,0 +1,12 @@
+#ifndef HAL_CELLULAR_GLOBAL_IDENTITY_H
```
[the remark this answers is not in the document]
zfields: I started there, but @technobly was concerned that including `cellular_hal_constants.h` in `enums_hal.h`

hal/src/boron/cellular_hal.cpp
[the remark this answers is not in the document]
avtolstoy: I personally first and foremost find it's difficult to comprehend when glancing over the code (as I did during the review), notwithstanding the fact that this is can be considered an antipattern in C++ specifically considering RAII and even in C a simple `goto` would make this code more readable.

hal/inc/hal_cellular_global_identity.h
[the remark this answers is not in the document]
avtolstoy: No strong preferences here, but I don't see any reason why it wouldn't work. The internal modem implementation could also define it's own internal structure for this, but it does sound excessive. If we are leaving this file, let's at least follow the same naming pattern:
`cellular_hal_global_identity.h` and at least add a copyright header.

hal/src/electron/modem/mdm_hal.cpp
zfields: Agreed, but we also set `false` to a bunch of variables, that's what threw me off.

hal/src/boron/cellular_hal.cpp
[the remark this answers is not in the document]
zfields: As I said, I don't mind to change it.

system/src/system_network_compat.cpp
```diff
@@ -255,6 +257,90 @@ class NetworkAccessTechnologyDiagnosticData: public AbstractIntegerDiagnosticDat
...
+
+    virtual int get(IntType& val)
+    {
+        CellularGlobalIdentity cgi;
+        cellular_global_identity(&cgi, nullptr);
+        val = static_cast<IntType>(cgi.mobile_country_code);
+
+        return SYSTEM_ERROR_NONE;
+    }
+} g_networkCellularCellGlobalIdentityMobileCountryCodeDiagnosticData;
```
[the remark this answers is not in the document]
zfields: Actually, since there are so few others (4, maybe 5), I think it may be better to change the rest of the file to follow this.

It keeps each message as a single block and allows for them to be organized much better. For instance, the block makes it easier to wrap such messages in a platform flag (i.e. `HAL_PLATFORM_CELLULAR`).

hal/src/boron/network/sara_ncp_client.cpp
zfields: I forgot to auto format code here, so it will be broken out into multiple lines as soon as I do.

Since, I'm ignoring the first integer, I think the URC will not actually parse any values, so I believe the correct thing to do is check `if (0 == r)` and that should alleviate your concern.

hal/src/boron/network/sara_ncp_client.cpp
zfields: same as above

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: That should probably work, yes. It's not a concern though, this is a breaking change.

system/src/system_network_compat.cpp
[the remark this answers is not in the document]
avtolstoy: No strong preference, but again I personally find it's easy to miss the instantiation :man_shrugging:

hal/src/boron/network/sara_ncp_client.cpp
zfields: What is "this"? The current behavior of the UrcParser, or checking for `r==0`?

avtolstoy: Currently checking for `r <= 1` will introduce a bug where `+CREG: <n>,<stat>` responses would get reparsed as URC, which has a different format (`+CREG: <stat>,...`) and [later](https://github.com/particle-iot/device-os/pull/1759/files/a2cf9b9cdb8a892dbdac2ee0ea5b33573f793a15#diff-b3f74fd1e22ae53aae264584d668ea1cR190) we will be erroneously checking `<n>` instead of `<stat>`.

zfields: This is all fixed by testing against `0 == r`, as we discussed, correct?

avtolstoy: > That should probably work, yes.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -363,13 +400,27 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
     {
         int act;
-        int v;
+        char home_network_identity[7] = {'\0','\0','\0','\0','\0','\0','\0'};
+        char mobile_country_code[4] = {'\0','\0','\0','\0'};
+        char mobile_network_code[4] = {'\0','\0','\0','\0'};
+
         auto resp = parser_.sendCommand("AT+COPS?");
-        int r = CHECK_PARSER(resp.scanf("+COPS: %d,%*d,\"%*[^\"]\",%d", &v, &act));
+        int r = CHECK_PARSER(resp.scanf("+COPS: %*d,%*d,\"%6[0-9]\",%d", home_network_identity, &act));
```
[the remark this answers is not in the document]
zfields: Love this idea. Preliminarily, I didn't trust my `scanf` syntax, but I'm switching over to your suggestion.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -363,13 +400,27 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
         CHECK_TRUE(r == 2, SYSTEM_ERROR_UNKNOWN);
         r = CHECK_PARSER(resp.readResult());
         CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);
 
+        // Parse MCC and MNC, from HNI
+        mobile_country_code[0] = home_network_identity[0];
```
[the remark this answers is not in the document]
zfields: eliminated

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1375,8 +1391,23 @@ int MDMParser::_cbCOPS(int type, const char* buf, int len, NetStatus* status)
 {
     if ((type == TYPE_PLUS) && status){
         int act = 99;
+        char home_network_identity[7] = {'\0','\0','\0','\0','\0','\0','\0'};
+        char mobile_country_code[4] = {'\0','\0','\0','\0'};
+        char mobile_network_code[4] = {'\0','\0','\0','\0'};
+
         // +COPS: <mode>[,<format>,<oper>[,<AcT>]]
-        if (sscanf(buf, "\r\n+COPS: %*d,%*d,\"%[^\"]\",%d",status->opr,&act) >= 1) {
+        if (sscanf(buf, "\r\n+COPS: %*d,%*d,\"%6[0-9]\",%d",home_network_identity,&act) >= 1) {
```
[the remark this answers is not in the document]
zfields: Done

hal/src/boron/network/sara_ncp_client.cpp
zfields: Forgot to apply the auto formatting. Fixed

hal/inc/net_hal.h
```diff
@@ -48,7 +48,8 @@ typedef enum {
     NET_ACCESS_TECHNOLOGY_WCDMA = 4,
     NET_ACCESS_TECHNOLOGY_CDMA = 5,
     NET_ACCESS_TECHNOLOGY_LTE = 6,
-    NET_ACCESS_TECHNOLOGY_IEEE802154 = 7
+    NET_ACCESS_TECHNOLOGY_IEEE802154 = 7,
+    NET_ACCESS_TECHNOLOGY_LTE_IOT = 8,
```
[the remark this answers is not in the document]
zfields: This isn't actually necessary for the work I'm doing, so I will stay with this more coarse resolution. We can improve it as needed.

hal/inc/hal_cellular_global_identity.h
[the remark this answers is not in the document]
zfields: How big should the version member be and what should the number scheme look like?

hal/src/boron/network/sara_ncp_client.cpp
zfields: I have also reduced the tolerance for matching, so this should only parse the appropriate values.

hal/inc/net_hal.h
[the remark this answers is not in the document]
avtolstoy: Can you elaborate on why it's not necessary? Why not do it properly from the get go?

hal/src/boron/network/sara_ncp_client.cpp
zfields: @technobly Would you review the new logic and see if it addresses your concerns?

system/src/system_network_compat.cpp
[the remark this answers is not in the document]
avtolstoy: Also I remember mentioning this previously, but this source is only being compiled for Gen 2. Ideally your PR should also decouple these diagnostic classes into a separate source file to be available on both Gen 2 and Gen 3.

hal/src/boron/network/sara_ncp_client.h
```diff
@@ -17,11 +17,14 @@
 
 #pragma once
 
+#include <stdlib.h>
```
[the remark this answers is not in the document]
zfields: fixed

hal/inc/net_hal.h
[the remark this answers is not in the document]
zfields: I don't have access to networks to test, and the distinction is not detailed in the desired RAT code of the PRD/SDD.

Leave this discussion open, and I will revisit before we finalize the PR.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -326,6 +348,21 @@ int SaraNcpClient::connect(const CellularNetworkConfig& conf) {
...
 
+int SaraNcpClient::getCellularGlobalIdentity(CellularGlobalIdentity* cgi)
+{
+    int result;
+    if ( !cgi ) {
+        result = -1;
+    }
+    else
+    {
+       *cgi = cgi_;
```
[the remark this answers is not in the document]
zfields: Great suggestion, I have isolated the update function and improved the logic flow in the latest PR.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -326,6 +348,21 @@ int SaraNcpClient::connect(const CellularNetworkConfig& conf) {
     return 0;
 }
 
+int SaraNcpClient::getCellularGlobalIdentity(CellularGlobalIdentity* cgi)
+{
+    int result;
+    if ( !cgi ) {
```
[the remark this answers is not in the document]
zfields: This stylistic choices need to be made to the `.clang-format` file, so I can apply them more broadly.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -326,6 +348,21 @@ int SaraNcpClient::connect(const CellularNetworkConfig& conf) {
     return 0;
 }
 
+int SaraNcpClient::getCellularGlobalIdentity(CellularGlobalIdentity* cgi)
+{
```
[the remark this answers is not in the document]
zfields: `.clang-format`

I'm updating as I find them.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -326,6 +348,21 @@ int SaraNcpClient::connect(const CellularNetworkConfig& conf) {
...
+    int result;
+    if ( !cgi ) {
+        result = -1;
+    }
+    else
```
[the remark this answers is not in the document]
zfields: `.clang-format`

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -363,13 +400,27 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
 
     {
         int act;
-        int v;
+        char home_network_identity[7] = {'\0','\0','\0','\0','\0','\0','\0'};
```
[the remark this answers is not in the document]
zfields: excellent.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1375,8 +1391,23 @@ int MDMParser::_cbCOPS(int type, const char* buf, int len, NetStatus* status)
...
+            // Parse MCC and MNC, from HNI
+            mobile_country_code[0] = home_network_identity[0];
```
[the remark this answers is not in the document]
zfields: parsing with `scanf`

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1375,8 +1391,23 @@ int MDMParser::_cbCOPS(int type, const char* buf, int len, NetStatus* status)
 {
     if ((type == TYPE_PLUS) && status){
         int act = 99;
+        char home_network_identity[7] = {'\0','\0','\0','\0','\0','\0','\0'};
```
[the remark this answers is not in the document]
zfields: done

system/src/system_network_compat.cpp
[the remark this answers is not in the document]
zfields: Thank you for reiterating, we went over a lot of topics that day...

Do you have a suggestion as to where it should live?

hal/inc/net_hal.h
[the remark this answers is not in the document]
zfields: I have split it apart (`LTE_CAT_M1` and `LTE_CAT_NB1`) as it leaves the `device-os`, it will be narrowed as it is reported by the `device-service`.

hal/inc/hal_cellular_global_identity.h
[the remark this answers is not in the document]
zfields: In the current iteration we have a `schema` field for describing the data format, as well as a `size` field.
At this time, I think we should choose a single 16-bit field. It would allow for better packing of the data, which would provide better space efficiency as well as faster data access.
As it stands, we are currently wasting 24-bits, and force aligning the `schema` field.

hal/src/boron/network/sara_ncp_client.cpp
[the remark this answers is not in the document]
zfields: field removed during refactoring

system/src/system_network_compat.cpp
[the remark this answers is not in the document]
zfields: This has now been split into `system_network_diagnostics.cpp`

hal/inc/cellular_hal_cellular_global_identity.h
```diff
@@ -0,0 +1,35 @@
...
+#include <stdint.h>
+
+typedef struct __attribute__((__packed__))
+{
+    uint16_t size;
+    uint16_t mobile_country_code;
+    uint16_t mobile_network_code;
+    uint16_t location_area_code;
+    uint32_t cell_id;
+    uint8_t schema;
```
avtolstoy: What is `schema`? We sometimes add a `version` field instead or along with `size`, but that goes at the top of the structure. Fields that are used for alignment are usually called `reservedX`, but if this field is removed your structure is already aligned by `sizeof(uintptr_t)` (12 bytes). If you want to add a reserved field, I'd prefer `uint32_t reserved` instead.

hal/inc/net_hal.h
```diff
@@ -48,7 +48,9 @@ typedef enum {
     NET_ACCESS_TECHNOLOGY_WCDMA = 4,
     NET_ACCESS_TECHNOLOGY_CDMA = 5,
     NET_ACCESS_TECHNOLOGY_LTE = 6,
-    NET_ACCESS_TECHNOLOGY_IEEE802154 = 7
+    NET_ACCESS_TECHNOLOGY_IEEE802154 = 7,
+    NET_ACCESS_TECHNOLOGY_LTE_CAT_M1 = 8,
+    NET_ACCESS_TECHNOLOGY_LTE_CAT_NB1 = 9,
```
avtolstoy: :+1:

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,41 @@ int cellular_credentials_clear(void* reserved) {
     return 0;
 }
 
+cellular_result_t cellular_global_identity(CellularGlobalIdentity* cgi_, void* reserved_)
+{
```
avtolstoy: Nitpicking: Please fix the formatting to follow the same style as the rest of the file.

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,41 @@ int cellular_credentials_clear(void* reserved) {
...
+    CHECK_TRUE(mgr, SYSTEM_ERROR_UNKNOWN);
+    const auto client = mgr->ncpClient();
+    CHECK_TRUE(client, SYSTEM_ERROR_UNKNOWN);
+
+    // Validate Argument(s)
+    if (nullptr == cgi_) {
+        (void)reserved_;
+        result = SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    // Load cached data into result struct
```
avtolstoy: Nitpicking: I find comments between code blocks confusing, but let's at least settle on one way or the other as https://github.com/particle-iot/device-os/pull/1759/files#diff-0db0ed911a9860ed2a4f0420a4dd3eefR244 does the opposite.

hal/inc/net_hal.h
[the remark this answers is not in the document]
avtolstoy: Any reason it should be narrowed by the device service?

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,41 @@ int cellular_credentials_clear(void* reserved) {
...
+    else if ((result = client->getCellularGlobalIdentity(&cgi))) {
```
avtolstoy: https://github.com/particle-iot/device-os/pull/1759#discussion_r279689461

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,86 @@ int SaraNcpClient::initParser(Stream* stream) {
             .commandTerminator(AtCommandTerminator::CRLF);
     parser_.destroy();
     CHECK(parser_.init(std::move(parserConf)));
+    // +CREG: <stat>[,<lac>,<ci>[,<AcTStatus>]]
     CHECK(parser_.addUrcHandler("+CREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %d,%d", &val[0], &val[1]));
+        int val[4] = {-1,-1,-1,-1};
+        char at_response[64]{};
```
avtolstoy: This should be `char atResponse[64] = {}`. Initializer list after `=` and `snake_case` should not be used in a `camelCase` source.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,86 @@ int SaraNcpClient::initParser(Stream* stream) {
...
         return 0;
     }, this));
+    // n={0,1} +CGREG: <stat>
+    // n=2     +CGREG: <stat>[,<lac>,<ci>[,<AcT>,<rac>]]
     CHECK(parser_.addUrcHandler("+CGREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CGREG: %d,%d", &val[0], &val[1]));
+        int val[4] = {-1,-1,-1,-1};
+        char at_response[64]{};
```
avtolstoy: Same here.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,86 @@ int SaraNcpClient::initParser(Stream* stream) {
...
+        self->cgi_.cell_id = val[2];
         return 0;
     }, this));
+    // +CEREG: <stat>[,[<tac>],[<ci>],[<AcT>][,<cause_type>,<reject_cause>[,[<Active_Time>],[<Periodic_TAU>]]]]
     CHECK(parser_.addUrcHandler("+CEREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CEREG: %d,%d", &val[0], &val[1]));
+        int val[4] = {-1,-1,-1,-1};
+        char at_response[64]{};
```
avtolstoy: Same here.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -355,42 +395,65 @@ int SaraNcpClient::getImei(char* buf, size_t size) {
...
 }
 
-int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
-    const NcpClientLock lock(this);
-    CHECK_TRUE(connState_ != NcpConnectionState::DISCONNECTED, SYSTEM_ERROR_INVALID_STATE);
-    CHECK_TRUE(qual, SYSTEM_ERROR_INVALID_ARGUMENT);
-    CHECK(checkParser());
+int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
+    int act;
+    char mobile_country_code[4] = {};
```
avtolstoy: `snake_case` in `camelCase` source.

hal/src/electron/cellular_hal.cpp
```diff
@@ -357,7 +357,36 @@ cellular_result_t cellular_band_available_get(MDM_BandSelect* bands, void* reser
...
+{
+    cellular_result_t result;
+    CellularGlobalIdentity cgi;
+
+    // Validate Argument(s)
+    if (nullptr == cgi_) {
+        (void)reserved_;
+        result = SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    // Load cached data into result struct
```
avtolstoy: Nitpicking: I find comments between code blocks confusing.

hal/src/electron/cellular_hal.cpp
```diff
@@ -357,7 +357,36 @@ cellular_result_t cellular_band_available_get(MDM_BandSelect* bands, void* reser
...
+    else if (!electronMDM.getCellularGlobalIdentity(cgi)) {
```
avtolstoy: https://github.com/particle-iot/device-os/pull/1759#discussion_r279689461

system/src/system_network_diagnostics.cpp
```diff
@@ -0,0 +1,355 @@
+/**
+ ******************************************************************************
+  Copyright (c) 2013-2015 Particle Industries, Inc.  All rights reserved.
```
avtolstoy: Let's use the current copyright header for new files.

system/src/system_network_diagnostics.cpp
```diff
@@ -0,0 +1,355 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ ******************************************************************************
+ */
+
+// Mesh signal strength/quality have not been implemented
+#if (PLATFORM_ID != 14) && (PLATFORM_ID != 24)
```
avtolstoy: Let's prefer to use the macros from `platforms.h`.

We could alternatively change the return values in https://github.com/particle-iot/device-os/blob/develop/wiring/inc/spark_wiring_mesh.h#L59 to negative for now to cause an error to be returned but I agree that it's probably better not to send them at all.

avtolstoy:
:+1: 
I'm going to double check the `scanfs` just in case and run a test before approving this, but I'd also prefer some of the comments to be addressed as well.

hal/inc/net_hal.h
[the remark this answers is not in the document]
zfields: The PRD explicitly requests the values of `network.cellular.radio_access_technology` to be `2G`, `3G`, `LTE` or `Mesh` - as printed on the hardware itself

hal/src/boron/cellular_hal.cpp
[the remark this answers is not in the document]
zfields: Actually, I want to push back on this. This follows the exact same pattern as the `CHECK` macro, except that you can choose how you want to handle your error and it's fully transparent.

hal/src/electron/cellular_hal.cpp
```diff
@@ -357,7 +357,43 @@ cellular_result_t cellular_band_available_get(MDM_BandSelect* bands, void* reser
...
 }
 
-cellular_result_t cellular_sms_received_handler_set(_CELLULAR_SMS_CB_MDM cb, void* data, void* reserved)
+cellular_result_t cellular_global_identity(CellularGlobalIdentity* cgi_, void* reserved_)
+{
+    cellular_result_t result;
+    CellularGlobalIdentity cgi;
+
+    // Validate Argument(s)
+    if (nullptr == cgi_)
```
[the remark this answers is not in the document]
zfields: Again, this follows the same pattern as the `CHECK` macro, except that you can choose how you want to handle your error, and as a benefit, it's fully transparent.

hal/src/boron/cellular_hal.cpp
zfields: There is a difference between the two. The comment above the block describes the attempted action, while the comment inside the block describes the error and its handling.

hal/src/boron/cellular_hal.cpp
zfields: done.

hal/inc/cellular_hal_cellular_global_identity.h
zfields: updated

hal/src/boron/network/sara_ncp_client.cpp
zfields: updated

hal/src/boron/network/sara_ncp_client.cpp
zfields: updated

hal/src/boron/network/sara_ncp_client.cpp
zfields: updated

hal/src/boron/network/sara_ncp_client.cpp
zfields: updated

hal/src/electron/modem/mdm_hal.cpp
zfields: We are now sending `AT+COPS=x,2`, so the format is constantly updated to be numeric.

system/src/system_network_diagnostics.cpp
zfields: Great catch!

system/src/system_network_diagnostics.cpp
zfields: updated

hal/inc/hal_cellular_global_identity.h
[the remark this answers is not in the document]
zfields: updated

hal/src/boron/network/sara_ncp_client.cpp
zfields: please confirm the new logic

zfields:
Pushing refactored code. I have adopted several of the formatting concerns as well as the initialization syntax.

As a heads up, I'm leaving the comments between code blocks in place, but I have attempted to address "why" I have elected to use them as such.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -155,46 +155,86 @@ int SaraNcpClient::initParser(Stream* stream) {
             .commandTerminator(AtCommandTerminator::CRLF);
     parser_.destroy();
     CHECK(parser_.init(std::move(parserConf)));
+    // +CREG: <stat>[,<lac>,<ci>[,<AcTStatus>]]
     CHECK(parser_.addUrcHandler("+CREG", [](AtResponseReader* reader, const char* prefix, void* data) -> int {
         const auto self = (SaraNcpClient*)data;
-        int val[2];
-        int r = CHECK_PARSER_URC(reader->scanf("+CREG: %d,%d", &val[0], &val[1]));
+        int val[4] = {-1,-1,-1,-1};
+        char at_response[64] = {0};
```
avtolstoy: `snake_case` in `camelCase` file, `0` is unnecessary, `{}` is enough. I think this is the third time I'm complaining about `snake_case` in exactly the same place :)

system/src/system_network_diagnostics.cpp
```diff
@@ -0,0 +1,355 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "platforms.h"
+
+// Mesh signal strength/quality have not been implemented
+#if (PLATFORM_ID != PLATFORM_XENON) && (PLATFORM_ID != PLATFORM_XENON_SOM)
```
avtolstoy: FYI we've renamed them in https://github.com/particle-iot/device-os/pull/1774, so after 1.2.0 branch is rebased on develop and this branch is rebase on 1.2.0, this will need to get fixed.

hal/inc/cellular_hal_cellular_global_identity.h
```diff
@@ -0,0 +1,51 @@
...
+ * The Cellular Global Identity is a globally unique identifier for a Base Transceiver Station in
+ * mobile phone networks. It consists of four parts: Mobile Country Code (MCC), Mobile Network Code
+ * (MNC), Location Area Code (LAC) and Cell Identification (CI). It is an integral part of 3GPP
+ * specifications for mobile networks, for example, for identifying individual base stations to
+ * "handover" ongoing phone calls between separately controlled base stations, or between different
+ * mobile technologies.
+ */
+typedef struct __attribute__((__packed__))
+{
+    uint16_t size; /*!< \c size is specified by the user application, to inform the device-os of the
```
avtolstoy: Consider aligning comments.

avtolstoy:
I'm still seeing textual operator representation on U270 in `AT+COPS?` response.

```
[ Modem::getSignalStrength ] = = = = = = = = = =
   119.905 AT send      10 "AT+COPS?\r\n"
   119.906 AT read  +   18 "\r\n+UUSORF: 0,105\r\n"
Socket 0: handle 0 has 105 bytes pending
   119.938 AT read UNK   9 "AT+COPS?\r"
   119.939 AT read  +   26 "\r\n+COPS: 0,0,\"Beeline\",2\r\n"
   119.940 AT read OK    6 "\r\nOK\r\n"
   119.940 AT send       8 "AT+CSQ\r\n"
   119.945 AT read UNK   7 "AT+CSQ\r"
   119.946 AT read  +   14 "\r\n+CSQ: 20,2\r\n"
   119.947 AT read OK    6 "\r\nOK\r\n"
```

avtolstoy:
I suspect we may have to either check `<format>` returned by `AT+COPS?` and reset it to numeric and rerun `AT+COPS?` or (which seems somewhat simpler) to always run `AT+COPS=3,2` before `AT+COPS?`, but I remember `AT+COPS=3,2` wasn't working correctly on some particular modem. :disappointed:

hal/inc/cellular_hal_cellular_global_identity.h
zfields: I used the `.clang-format` file after typing my comments and this is the output.

avtolstoy: We are all aware that our current `.clang-format` needs to be updated, because it was merely a PoC done by Julien a long time ago and doesn't cover all the cases, but it shouldn't mean that we completely forget about formatting and blame everything on `clang-format`.

avtolstoy: And all I'm saying is _consider_.

hal/src/boron/network/sara_ncp_client.cpp
zfields: You finally pierced my skull! `snake_case` is now `camelCase`!

hal/inc/cellular_hal_cellular_global_identity.h
zfields: The current directive is to apply the `.clang-format` to format the entire file you are working on, make a separate commit and start making your changes. Personally, I don't value the beauty of the code as much as the time spent to make it look that way.

I understand `.clang-format` is currently incomplete and I don't wish to be confrontational, but if we keep following the routine above, we will have homogeneous code that is capable of evolving over time.

zfields: For this PR, I have not applied the format to the entire file, because I was already working before we were told to do so.

system/src/system_network_diagnostics.cpp
zfields: updated

hal/inc/cellular_hal_cellular_global_identity.h
technobly: Good luck getting `.clang-format` to align these comments the way you want... I agree over time `.clang-format` used blindly is going to de-beautify areas of the code and make them harder to read.  Consider these two options:
1. https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code
```
    // clang-format off
    code or comments not to format
    // clang-format on
```
2. Further down from above link: AlignTrailingComments (bool)

technobly: `AlignTrailingComments: true` appears to already be set... so I wonder why it's not working.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1234,12 +1240,20 @@ bool MDMParser::getDataUsage(MDM_DataUsage &data)
 
 bool MDMParser::getCellularGlobalIdentity(CellularGlobalIdentity& cgi_) {
     bool result;
+    // Reformat the operator string to be numeric (allows the capture of `mcc` and `mnc`)
+    if (0 == sendFormated("AT+COPS=3,2\r\n")) {
```
avtolstoy: I still strongly discourage such a pattern in C/C++. Let's discuss this in a wider team and come to a unified stance on this if you feel that I'm simply being annoying, but it decreases the readability of code by a lot and also prevents the usage of RAII in-between the actions. cc @sergeuz @technobly @m-mcgowan

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1234,12 +1240,20 @@ bool MDMParser::getDataUsage(MDM_DataUsage &data)
 
 bool MDMParser::getCellularGlobalIdentity(CellularGlobalIdentity& cgi_) {
     bool result;
```
avtolstoy: `LOCK()` / `UNLOCK()` needs to be added.

hal/src/electron/modem/mdm_hal.cpp
zfields: I would love to have this conversation. To be honest, I don't fully understand your concern, and it would be nice to discuss it as opposed to messaging back and forth. I'm always happy to learn something new.

technobly: I agree that an `if-else tree` should not be used as a state machine.  The `goto failure` pattern used in the rest of mdm_hal.cpp is more readable, or you can use a `do { } while (false);` block with `breaks` if you really despise goto's.

avtolstoy: The simplest example:
```cpp
int func() {
    CHECK(doSomething());
    CHECK(doSomethingElse());

    // Only at this point after doing something and something else
    // we want to for example acquire some lock
    RaiiLock lock;

    CHECK(doThirdThing());
    CHECK(doFourthThing());

    return 0;
}
```

How would you do this with your (if-else-waterfall? I don't know how to call it)?

I'm not accentuating the usage of `CHECK` macros here in the example, it can very well be a set of `ifs` with a `return` or a `goto` or the whole thing can be a `do-while` with `breaks`.

And it's not only about locking. With your pattern for example you are almost always required to declare all the variables at the top of the function, you cannot easily do so consecutively as the code flow goes which I find rather inconvenient and an atavism from C89.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -845,6 +846,11 @@ bool MDMParser::init(DevStatus* status)
     sendFormated("AT+CIMI\r\n");
     if (RESP_OK != waitFinalResp(_cbString, _dev.imsi))
         goto failure;
+    // Reformat the operator string to be numeric
+    // (allows the capture of `mcc` and `mnc`)
+    sendFormated("AT+COPS=3,2\r\n");
+    if (RESP_OK != waitFinalResp())
```
technobly: This should have the proper timeout value `if (RESP_OK != waitFinalResp(nullptr, nullptr, COPS_TIMEOUT))`

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1156,6 +1164,12 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
         if (!_atOk()) {
             goto failure;
         }
+        // Reformat the operator string to be numeric
+        // (allows the capture of `mcc` and `mnc`)
+        sendFormated("AT+COPS=3,2\r\n");
+        if (RESP_OK != waitFinalResp()) {
```
technobly: This should have the proper timeout value `if (RESP_OK != waitFinalResp(nullptr, nullptr, COPS_TIMEOUT))`

hal/src/electron/modem/mdm_hal.cpp
zfields: As for `goto`, Bjarne Stroustrup says it best in his talk, [What can C++ do for embedded systems developers?](https://youtu.be/VoHOLDdfDhk?t=3455)

> Why don't we use `goto`? The answer is, it can do _ANYTHING_. We want something better behaved...

avtolstoy: You are derailing conversation somewhat as both Brett and I have mentioned other possible ways to do this, but if you want to use Stroustrup's name as an argument point, please consider reading through [E: Error handling section of C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e-error-handling) also written by Stroustrup and specifically sections:

```
E.25: If you can’t throw exceptions, simulate RAII for resource management
E.26: If you can’t throw exceptions, consider failing fast
E.27: If you can’t throw exceptions, use error codes systematically
E.28: Avoid error handling based on global state (e.g. errno)
```

zfields: I would wager, it is not as difficult to decipher as it is foreign (and therefore not immediately obvious).

1. It has the advantage of not being opaque (unlike the `CHECK` macro)
2. It can handle any result type (not only error values `< 0` or success codes `>= 0`
3. It can clean an non-RAII allocation from a previous step (allocation, lock, etc...)
   - Even if we perfect our codebase, our third-party dependencies can create this need.

Also, the example above may be symptomatic of an overly complex function, but if not, it can still be expressed. It requires one level of nesting at each RAII acquisition.

```cpp
int func() {
    int result;
    if (result = doSomething()) {
        // handle error condition
    } else if (result = doSomethingElse()) {
        // handle error condition and clean resources acquired during previous step(s)
    } else {
        // Only at this point after doing something and something else
        // we want to for example acquire some lock
        RaiiLock lock;

        if (result = doThirdThing()) {
            // handle error condition and clean resources acquired during previous step(s)
        } else if (result = doFourthThing()) {
            // handle error condition and clean resources acquired during previous step(s)
        } else {
            // do stuff
        }
    }
    return result;
}
```

technobly: They can't do _ANYTHING_ in modern compilers.  Try to make a `goto` jump out of a function and into another function... see if that compiles ^_^  

>_Re: goto, The main reason you shouldn’t use them is because they make some people freak out. -Gramshackle_

hal/src/electron/modem/mdm_hal.cpp
zfields: fixed

hal/src/electron/modem/mdm_hal.cpp
zfields: updated

hal/src/electron/modem/mdm_hal.cpp
zfields: updated

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: That actually looks at least somewhat palatable, due to simplicity of the example and lack of comments _between_ the blocks compared to the PR, but still, instead of going downwards, you need to go south-east in order to go through the flow of the function. Why? We can argue about short functions but even with such a simple example it's still apparent.

But again, this is example is trivial, you if add non-trivial C++ objects and other things into the equation, you'll end up nesting most of the times.

hal/src/boron/cellular_hal.cpp
```diff
@@ -221,6 +223,34 @@ int cellular_credentials_clear(void* reserved) {
...
+
+    // Validate Argument(s)
+    (void)reserved_;
+    CHECK_TRUE((nullptr != cgi_), SYSTEM_ERROR_INVALID_ARGUMENT);
+
+    // Load cached data into result struct
+    CHECK(client->getCellularGlobalIdentity(&cgi));
+
+    // Validate cache
+    CHECK_TRUE((0 != cgi.mobile_country_code && 0 != cgi.mobile_network_code &&
```
avtolstoy: Nitpicking (non-actionable): Prefer simple ifs for complex statements.

hal/src/electron/cellular_hal.cpp
```diff
@@ -357,7 +358,31 @@ cellular_result_t cellular_band_available_get(MDM_BandSelect* bands, void* reser
...
+
+    // Validate Argument(s)
+    (void)reserved_;
+    CHECK_TRUE((nullptr != cgi_), SYSTEM_ERROR_INVALID_ARGUMENT);
+
+    // Load cached data into result struct
+    CHECK_TRUE(electronMDM.getCellularGlobalIdentity(cgi), SYSTEM_ERROR_AT_NOT_OK);
+
+    // Validate cache
+    CHECK_TRUE((0 != cgi.mobile_country_code && 0 != cgi.mobile_network_code &&
```
avtolstoy: Nitpicking (non-actionable): Prefer simple ifs for complex statements.

avtolstoy:
@technobly I think this should be considered a feature and not an enhancement.

hal/src/boron/cellular_hal.cpp
m-mcgowan: Given that the expression is a conjunction, we could test each part of it individually. This would also give clearer logging so we can see which expression failed. 

```suggestion
    CHECK_TRUE((0 != cgi.mobile_country_code, SYSTEM_ERROR_BAD_DATA);
    CHECK_TRUE(0 != cgi.mobile_network_code, SYSTEM_ERROR_BAD_DATA);
    CHECK_TRUE(0xFFFF != cgi.location_area_code, SYSTEM_ERROR_BAD_DATA);
    CHECK_TRUE(0xFFFFFFFF != cgi.cell_id, SYSTEM_ERROR_BAD_DATA);
```

m-mcgowan: eh, not sure how I make the suggestion delete more lines...

PR #1762, opened 2019-05-03
[Electron/LTE] disables all eDRX AcT types [ch32051]

technobly:
### Problem

eDRX is enabled by default on production u-Blox modules, and it causes a variable lag in the network data transmissions.  We have previously disabled it, but not entirely.

### Solution

Scan for and disable each eDRX AcT type, and reset the modem if disabling is required or an error occurs.

>**Note:** When disabling eDRX mode, the modem needs to be reset to clear eDRX behavior in the cellular network.  This will cause a one time ~30s reconnection delay after OTA updating the new firmware.

>**Note:** When eDRX mode is enabled, TX on the device temporarily allows the network to RX faster.  For example, if the device published every 2 minutes and we asynchronous tried to read a Particle Variable every ~5 seconds, the device would up timing out 2-3 times in a row and succeeding 2-3 times in a row repeatedly.  This affects OTA updates as well.  An OTA update might never start until the device pings the Cloud and reduces the RX timing in the network.  Because of this, devices with this behavior that are remote in the field might need special care in updating them remotely, Ex. a script that repeatedly tries OTA updating every 30 seconds for 30 minutes.

### Steps to Test

🔗 [Electron/LTE firmware compiled for this PR-1762, v1.1.0-rc.2.  Tinker included but not required.](https://www.dropbox.com/s/gox036vwxp51iok/particle_device-os%401.1.0-rc.2-pr-1762.zip?dl=1)

Manual testing required:
1. Flash LTE device with this system firmware and tinker
2. Make a function call `particle call <device> digitalwrite D7,LOW`
3. It should return 1 if successful
4. Repeat every 30 seconds for 10 or more attempts

⚠️ Should you need to for testing purposes, you may enable eDRX mode with the following commands:

```c
sendFormated("AT+CEDRXS=2,2,\"1001\"\r\n"); waitFinalResp();
sendFormated("AT+CEDRXS=2,3,\"1001\"\r\n"); waitFinalResp();
sendFormated("AT+CEDRXS=2,4,\"1001\"\r\n"); waitFinalResp();
sendFormated("AT+CEDRXS=2,5,\"1001\"\r\n"); waitFinalResp();
```

### References

#1567 
Closes [ch32051]

---

### Completeness

- [x] [n/a] Added documentation

---

- [bugfix] [Electron/LTE] disables all eDRX AcT types [ch32051] [#1762](https://github.com/particle-iot/device-os/pull/1762)

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -480,6 +480,23 @@ void MDMParser::unlock()
     mdm_mutex.unlock();
 }
 
+int MDMParser::_cbCEDRXS(int type, const char* buf, int len, EdrxActs* edrxActs)
+{
+    if (((type == TYPE_PLUS) || (type == TYPE_UNKNOWN)) && edrxActs) {
+        // if response is "\r\n+CEDRXS:\r\n", all AcT's disabled, do nothing
+        if (strncmp(buf, "\r\n+CEDRXS:\r\n", len) != 0) {
+            int a;
+            if (sscanf(buf, "\r\n+CEDRXS: %1d[2-5]", &a) == 1 ||
+                sscanf(buf, "+CEDRXS: %1d[2-5]", &a) == 1) {
+                if (edrxActs->count < MDM_R410_EDRX_ACTS_MAX) {
+                    edrxActs->act[edrxActs->count++] = a;
+                }
+            }
+        }
+    }
+    return WAIT;
+}
+
 int MDMParser::_cbString(int type, const char* buf, int len, char* str)
 {
     if (str && (type == TYPE_UNKNOWN)) {
@@ -807,33 +824,6 @@ bool MDMParser::init(DevStatus* status)
     sendFormated("AT+CCID\r\n");
     if (RESP_OK != waitFinalResp(_cbCCID, _dev.ccid))
         goto failure;
-    // enable power saving
-    if (_dev.lpm != LPM_DISABLED) {
-        // enable power saving (requires flow control, cts at least)
-        sendFormated("AT+UPSV=%d\r\n", _power_mode);
-        if (RESP_OK != waitFinalResp())
-            goto failure;
-        if (_power_mode != 0) {
-            _dev.lpm = LPM_ACTIVE;
-        }
-    }
-    if (_dev.dev == DEV_SARA_R410) {
-        // Force eDRX mode to be disabled
-        // 18/23 hardware doesn't seem to be disabled by default
-        sendFormated("AT+CEDRXS=3,2\r\n");
-        if (RESP_OK != waitFinalResp())
-            goto failure;
-        sendFormated("AT+CEDRXS?\r\n");
-        if (RESP_OK != waitFinalResp())
-            goto failure;
-        // Force Power Saving mode to be disabled for good measure
-        sendFormated("AT+CPSMS=0\r\n");
-        if (RESP_OK != waitFinalResp())
-            goto failure;
-        sendFormated("AT+CPSMS?\r\n");
-        if (RESP_OK != waitFinalResp())
-            goto failure;
-    }
     // Setup SMS in text mode
     sendFormated("AT+CMGF=1\r\n");
     if (RESP_OK != waitFinalResp())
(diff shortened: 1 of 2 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -851,14 +842,67 @@ bool MDMParser::init(DevStatus* status)
...
     return true;
+
 failure:
     UNLOCK();
+    return false;
 
+reset_failure:
+    // Don't get stuck in a reset-retry loop
+    static int resetFailureAttempts = 0;
+    if (resetFailureAttempts++ < 3) {
```
sergeuz: I'd move this condition to where the `resetNeeded` variable is checked. Otherwise there's still a chance for the system to stuck in a modem reinitialization loop, because `MDMParser::init()` returns `false` and causes `cellular_init()` to fail even if the number of resets has exceeded the limit

sergeuz:
I approved this PR, but it looks like the handler for `+CEDRXS` command won't work correctly in case of multiple modem initialization attempts, because it stores the number of AcT types in a static variable which only grows and never gets reset (see [this](https://github.com/particle-iot/device-os/pull/1762#discussion_r284214199) comment)

sergeuz:
@jlovinger-particle Could you test this PR? I don't have an LTE Gen 2 device

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -851,14 +842,67 @@ bool MDMParser::init(DevStatus* status)
...
+        if (RESP_OK != waitFinalResp())
+            goto failure;
+        // Force eDRX mode to be disabled
+        // 18/23 hardware doesn't seem to be disabled by default
+        sendFormated("AT+CEDRXS?\r\n");
+        int edrxActs[4] = { -1, -1, -1, -1 };
+        if (RESP_ERROR == waitFinalResp(_cbCEDRXS, edrxActs)) {
+            goto reset_failure;
+        }
+        int resetNeeded = false;
```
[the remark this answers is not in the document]
technobly: It's not stated in the AT manual, but it has been determined though testing that it is required.  Especially when OTA updating this code to a device that has eDRX enabled currently.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -479,6 +479,24 @@ void MDMParser::unlock()
...
 
+int MDMParser::_cbCEDRXS(int type, const char* buf, int len, int* i)
+{
+    if (((type == TYPE_PLUS) || (type == TYPE_UNKNOWN)) && i) {
+        // if response is "\r\n+CEDRXS:\r\n", all AcT's disabled, do nothing
+        if (strncmp(buf, "\r\n+CEDRXS:\r\n", len) != 0) {
+            int a;
+            if (sscanf(buf, "\r\n+CEDRXS: %1d[2-5]", &a) == 1 ||
+                sscanf(buf, "+CEDRXS: %1d[2-5]", &a) == 1) {
+                static int idx = 0;
```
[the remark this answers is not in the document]
technobly: I was trying to keep the code here as simple as possible due to it being behind a check for the R410 modem type.  How about we just change the magic number 4 to a define specifically for SARA_R410M_02B and revisit this if needed for future expansion later?

hal/src/electron/modem/mdm_hal.cpp
technobly: It's ok if it gets stuck in an initialization loop by default exiting through `failure`, although it still won't do that (read on).  This protects against the specific case where we want to force a modem reset due to CEDRXS commands failing/succeeding, and preventing the cellular registration from continuing to run anyway.  Also if `MDMParser ::init()` returns false, it does not prevent registration from being run because the return value is not checked.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
sergeuz: It's not just about the array size. This index variable never gets reset and the handler for the `AT+CEDRXS` command won't work correctly during a reinitialization attempt

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
sergeuz: I mean let's change the type of this variable from `int` to `bool` :)

hal/src/electron/modem/mdm_hal.cpp
sergeuz: > it does not prevent registration from being run because the return value is not checked

It's an oversight in the implementation of the system layer and not something we should rely on in the HAL code. If that's okay for `AT+CEDRXS` to fail and not cause a modem reset after a certain number of attempts then why not to express that explicitly in the code?

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
technobly: I see what you mean for `idx`, yes that would make things better 👍

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
technobly: ahh yes good catch, I didn't spot that typo twice :)

hal/src/electron/modem/mdm_hal.cpp
technobly: Express it as a comment?  Basically preventing cellular registration just gets us back to retrying init faster, and the timeout of 3 attempts is arbitrary but allows us to register and connect even with an error, since that error is persisting and might just be a fluke or bug in the modem.  Allowing to continue on eventually helps the device recover in odd situations we can't predict.

sergeuz: I was thinking about making `MDMParser::init()` return `true` in case of an eDRX-related error to make it apparent that we don't want that error to prevent the modem from registering in the network, but, yes, at the very least, a comment that we expect the system to ignore the result of `cellular_init()` and proceed with the connection attempt would be helpful

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
technobly: Updated :)

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
technobly: Fixed

hal/src/electron/modem/mdm_hal.cpp
technobly: Added the comments in a shadowy commented code block 🕶

hal/src/electron/modem/enums_hal.h
```diff
@@ -112,6 +112,24 @@ typedef struct {
     unsigned int ci;     //!< Cell ID in hexadecimal format (2 to 4 bytes in hex)
 } NetStatus;
 
+#ifdef __cplusplus
+#define MDM_R410_EDRX_ACTS_MAX (4) //!< maximum number of AcTs for eDRX mode on SARA-R410M-02B
+//! eDRX AcTs
+struct MDM_EdrxActs {
+    uint16_t size;
```
sergeuz: This structure is only being used internally, there's no need for it to support versioning

hal/src/electron/modem/mdm_hal.h
```diff
@@ -557,6 +558,7 @@ class MDMParser
     NetStatus   _net; //!< collected network information
     MDM_IP       _ip;  //!< assigned ip address
     MDM_DataUsage _data_usage; //!< collected data usage information
+    MDM_EdrxActs _edrxActs; //!< detected eDRX AcTs that are enabled
```
sergeuz: We can store this structure on the stack and save some memory:
```
MDM_EdrxActs edrxActs = {};
if (RESP_ERROR == waitFinalResp(_cbCEDRXS, &edrxActs)) {
    ...
}
```

technobly: True, I can replace the call to clear .count with just creating a new object which self memsets to 0 :)

hal/src/electron/modem/enums_hal.h
technobly: I'll reduce it's size, and move it to mdm_hal.h as well.

PR #1767, opened 2019-05-07
Wait for confirmable messages when entering the deep sleep mode

sergeuz:
### Problem

`System.sleep(SLEEP_MODE_DEEP, ...)` doesn't wait for confirmable messages to be processed before entering the deep sleep mode. If the current session data became invalid for some reason, this may put a sleepy device into an endless loop, where it wakes up, restores the session, publishes an event and goes back to sleep without ever noticing that there's no established connection with the cloud.

### Solution

By default, the system should wait for all confirmable messages to be processed before entering a sleep mode. This PR also introduces a flag that allows overriding the default behavior (`SLEEP_NO_WAIT`).

### Steps to Test

1. Flash the following application to an Electron:
```cpp
#include "application.h"

// SYSTEM_THREAD(ENABLED)

namespace {

    { "app", LOG_LEVEL_ALL }
});

} // namespace

void setup() {
    Log.info("setup()");
}

void loop() {
    Log.info("Particle.publish()");
    auto f = Particle.publish("my_event", PRIVATE | WITH_ACK);
    f.onSuccess([](bool) {
        Log.info("ACK received");
    }).onError([](Error) {
        Log.warn("Particle.publish() failed");
    });
    Log.info("System.sleep()");
    System.sleep(SLEEP_MODE_DEEP, 5);
}
```

Verify that the application gets notified about an ACK received for the published message before the device enters the deep sleep mode.

~2. Apply the following patch via `git apply`:~
```diff
diff --git a/communication/src/dtls_message_channel.cpp b/communication/src/dtls_message_channel.cpp
index dec29fb9d..be096f4c2 100644
--- a/communication/src/dtls_message_channel.cpp
+++ b/communication/src/dtls_message_channel.cpp
@@ -112,6 +112,7 @@ auto SessionPersist::restore(mbedtls_ssl_context* context, bool renegotiate, uin
 		context->state = MBEDTLS_SSL_HANDSHAKE_WRAPUP;
 		context->in_epoch = in_epoch;
 		memcpy(context->out_ctr, &out_ctr, sizeof(out_ctr));
+		context->out_ctr[0] = 'X'; // Invalidate the session data
 		memcpy(context->handshake->randbytes, randbytes, sizeof(randbytes));
 
 		context->transform_negotiate->ciphersuite_info = mbedtls_ssl_ciphersuite_from_id(ciphersuite);
```

~Verify that the device running the same application performs the full handshake on every second boot. When the system resumes the session using invalid data, it should take no longer than 1 minute for the communication layer to fail with a timeout error while waiting for ACKs and reset the session.~

**Edit:** A better solution for the second test case is implemented in https://github.com/particle-iot/device-os/pull/1776.

### References

- [ch32341]

---

- [bugfix] Wait for confirmable messages when entering the deep sleep mode [#1767](https://github.com/particle-iot/device-os/pull/1767)

### DIFF
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -116,6 +116,10 @@ int system_sleep_impl(Spark_Sleep_TypeDef sleepMode, long seconds, uint32_t para
     //---- #2
     // SLEEP_NETWORK_STANDBY can keep the modem on during DEEP sleep
     // System.sleep(10) always powers down the network, even if SLEEP_NETWORK_STANDBY flag is used.
+
+    // Make sure all confirmable UDP messages are sent and acknowledged before sleeping
+    Spark_Sleep();
+
     if (network_sleep_flag(param) || SLEEP_MODE_WLAN == sleepMode) {
         network_suspend();
     }
@@ -158,11 +162,9 @@ int system_sleep_impl(Spark_Sleep_TypeDef sleepMode, long seconds, uint32_t para
 int system_sleep_pin_impl(const uint16_t* pins, size_t pins_count, const InterruptMode* modes, size_t modes_count, long seconds, uint32_t param, void* reserved)
 {
     SYSTEM_THREAD_CONTEXT_SYNC(system_sleep_pin_impl(pins, pins_count, modes, modes_count, seconds, param, reserved));
-    // If we're connected to the cloud, make sure all
-    // confirmable UDP messages are sent before sleeping
-    if (spark_cloud_flag_connected()) {
-        Spark_Sleep();
-    }
+
+    // Make sure all confirmable UDP messages are sent and acknowledged before sleeping
+    Spark_Sleep();
 
     bool network_sleep = network_sleep_flag(param);
     if (network_sleep)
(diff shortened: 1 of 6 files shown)


system/src/system_sleep.cpp
```diff
@@ -116,6 +116,12 @@ int system_sleep_impl(Spark_Sleep_TypeDef sleepMode, long seconds, uint32_t para
     //---- #2
     // SLEEP_NETWORK_STANDBY can keep the modem on during DEEP sleep
     // System.sleep(10) always powers down the network, even if SLEEP_NETWORK_STANDBY flag is used.
+
+    // Make sure all confirmable UDP messages are sent and acknowledged before sleeping
+    if (spark_cloud_flag_connected() && !(param & SYSTEM_SLEEP_FLAG_NO_WAIT)) {
+        Spark_Sleep();
```
m-mcgowan: This is a blocking call. Do we want to keep it like that?

sergeuz: For now I'd say yes, because `System.sleep()` is a blocking call as well. Also, we're already waiting for confirmable messages synchronously when entering the stop mode: https://github.com/particle-iot/device-os/blob/5068605f4fb4d80bfc6ba366d1a5da61db60480f/system/src/system_sleep.cpp#L164

PR #1769, opened 2019-05-08
photon UDP support

m-mcgowan:
### Problem

UDP Cloud Communications on the Photon/P1. This was not working correctly after refactoring mbedtls in the system firmware and the Gen2/Gen3 integration.

### Solution

- Allow the standard mbedtls config to be used (not the special photon one) when UDP is enabled.
- Weaken the duplicated mbedtls functions from the BCEL library in WICED so that we don't get duplicate 

### Steps to Test
- Edit `hal_platform_compat.h` to enable `HAL_PLATFORM_CLOUD_UDP` on the Photon/P1. 
- Flash monolithic firmware (It does not build with separate system parts.)
- Power on the photon, verify that it connects to the cloud. 
- Verify the handshake is visible via fast blinking cyan LED
- Hit reset
- Verify the handshake is no longer visible and that the device almost immediately connects to the cloud. This is a symptom of UDP session resumption. 
- NB: I did not test that TCP communications still works. UDP is enabled by default with this build. (Verify!)
- Verify WPA Enterprise authentication still works after this change.

### Example App

```c
void setup() {
   */
}

void loop() {

}
```

### References

---

### Completeness

- [ ] Added documentation


crypto/inc/mbedtls_config_default.h
```diff
@@ -21,10 +21,19 @@
  *  This file is part of mbed TLS (https://tls.mbed.org)
  */
 
-#if PLATFORM_ID == 6 || PLATFORM_ID == 8
+#include "hal_platform.h"
+
+#if PLATFORM_ID==6 || PLATFORM_ID==8
+#define PLATFORM_WICED 1
```
avtolstoy: `HAL_PLATFORM_WICED` in `hal_platform.h` perhaps?

We should also get rid of these `mbedlts_config_default.h`/`mbedtls_config_photon.h` in the future and use the same scheme for Gen 2 devices as for Gen 3: platform-specific `mbedtls_config_platform.h` in platform-specific HAL folder.

hal/src/photon/lib/BESL.ARM_CM3.release.a.weaken
```diff
@@ -37,3 +37,24 @@ rsa_pkcs1_verify
...
 rsa_check_privkey
+mbedtls_ccm_init
+mbedtls_ccm_setkey
+mbedtls_ccm_free
+mbedtls_ccm_encrypt_and_tag
+mbedtls_ccm_auth_decrypt
+mbedtls_cipher_info_from_type
+mbedtls_cipher_info_from_string
+mbedtls_cipher_info_from_values
+mbedtls_cipher_init
```
avtolstoy: We should definitely verify that WPA Enterprise authentication still works after this change.

crypto/inc/mbedtls_config_default.h
```diff
@@ -21,10 +21,19 @@
  *  This file is part of mbed TLS (https://tls.mbed.org)
  */
 
-#if PLATFORM_ID == 6 || PLATFORM_ID == 8
+#include "hal_platform.h"
+
+#if PLATFORM_ID==6 || PLATFORM_ID==8
```
avtolstoy: Nitpicking: spaces around `=`

hal/src/photon/lib/BESL.ARM_CM3.release.a.weaken
technobly: Added to Steps to test

m-mcgowan:
Closing out. Not sure if it is still relevant.

PR #1774, opened 2019-05-13
Rename SoM platform names [ch32184]

technobly:
### Problem

The Device OS `PLATFORM_NAME` definition for platforms 22, 23, 24 started as `argon-som`, `boron-som` and `xenon-som` in 1.1.0-rc.1 and 1.1.0-rc.2. The official platform names for the SoMs are A SoM, B SoM and X SoM so in Device OS the `PLATFORM_NAME` should be renamed to `asom`, `bsom`, `xsom` for compilation.

### Solution

- Update build scripts to reflect the name changes
- Update USB device descriptor info
- Update comments and PLATFORM_ARGON_SOM throughout the code to PLATFORM_ASOM, etc..
- Update previous 1.1.0-rc.x release notes.

### Steps to Test

- If this build compiles, that will test Travis CI
- Local compilation should also pass:
```
time ./make_release.sh --platform=asom --tests --debug --publish=1.1.0-rc.2
time ./make_release.sh --platform=bsom --tests --debug  --publish=1.1.0-rc.2
time ./make_release.sh --platform=xsom --tests --debug  --publish=1.1.0-rc.2
```

---

### Completeness

- [x] N/A Added documentation

---

- [internal] Rename SoM platform names [ch32184] [#1774](https://github.com/particle-iot/device-os/pull/1774)


bootloader/src/nRF52840/usbd_device.cpp
```diff
@@ -143,11 +143,11 @@ class NrfDevice : public Device {
 #elif PLATFORM_ID == 14
 # define USBD_PRODUCT_NAME              "Xenon"
 #elif PLATFORM_ID == 22
-# define USBD_PRODUCT_NAME              "Argon SoM"
+# define USBD_PRODUCT_NAME              "A SoM"
 #elif PLATFORM_ID == 23
-# define USBD_PRODUCT_NAME              "Boron SoM"
+# define USBD_PRODUCT_NAME              "B SoM"
 #else
```
zfields: I know this isn't the change you just made, but shouldn't this be narrowed to `PLATFORM_ID == 24`?

build/release.sh
```diff
@@ -7,8 +7,8 @@ function display_help ()
 {
     echo "\
 usage: release.sh [--output-directory=<binary_output_directory>]
-                  (--platform=<argon|argon-som|boron|boron-som...
-                  |core|electron|p1|photon|xenon|xenon-som>...
+                  (--platform=<argon|asom|boron|bsom...
```
m-mcgowan: I realize this isn't the focus of this PR but it seems there is room for refactoring in these bash scripts to remove the duplication. cc: @zfields

bootloader/src/nRF52840/usbd_device.cpp
technobly: I see what you mean, and I'm not sure what we would default it to then... at least this is a somewhat sane default as there are only 6 nrF52840 platforms.

technobly: I guess we could just throw an #error "platform not supported"

technobly:
> Looks good. How about adding a step to the unit test bash script that greps the codebase for "PLATFORM_XENON_SOM" and "Xenon SoM" and similarly for A and B. and fails if they are found?

I want to say that will unnecessarily slow down the CI job, since there are so many files to search though.  Also I don't think it's necessary to be that strict on comments that we'd fail the build for it.  Also if anyone tries to use PLATFORM_ARGON_SOM, PLATFORM_BORON_SOM, PLATFORM_XENON_SOM in code out of habit it will fail that it's not defined.

build/platform-id.mk
```diff
@@ -333,13 +333,13 @@ PLATFORM_LWIP=posix
 endif
 
 ifeq ("$(PLATFORM_ID)","22")
-PLATFORM=argon-som
+PLATFORM=asom
 PLATFORM_NAME=argon
```
avtolstoy: This is used in https://github.com/particle-iot/device-os/blob/develop/hal/src/nRF52840/device_code.cpp#L93, so despite all the changes the broadcast BLE name will still remain (Argon-XXXXXX/Boron-XXXXXX/Xenon-XXXXXX for SoMs).

technobly: Thank @avtolstoy :) @monkbroc do we want to broadcast a different name than Argon-XXXXXX, etc.. for SoMs?  ASoM-XXXXXX seems a little weird without the space, and I think the space would break things.

monkbroc: That's fine. Leave it as is for now as it would break mobile apps otherwise. We can change it later if needed.

m-mcgowan: We are going to be reworking the mobile apps for the upcoming BLE APIs and changes to how setup mode functions, so there is scope there to rework the apps when we decide on the final name.

PR #1775, opened 2019-05-14
communication: cleans up CoAP timeout/parameters/retransmission stuff

avtolstoy:
### Description

This is something I made some time ago while investigating CoAP issues for Gen 3 devices, so putting it here so it's not lost in time.

This PR:
1. Cleans up (IMHO) CoAP parameter calculations
2. Fixes retransmission logic to exactly follow RFC (random part is calculated only once), additional logging
3. `MAX_TRANSMIT_SPAN` -> `EXCHANGE_LIFETIME` again to follow the RFC
4. Makes `PING` ACK timeout based on `MAX_TRANSMIT_SPAN` instead of arbitrary `30s`

### DIFF
--- a/communication/src/coap_channel.cpp
+++ b/communication/src/coap_channel.cpp
@@ -48,6 +48,7 @@ bool CoAPMessageStore::retransmit(CoAPMessage* msg, Channel& channel, system_tic
 
 void CoAPMessageStore::message_timeout(CoAPMessage& msg, Channel& channel)
 {
+	LOG(TRACE, "timeout waiting for ACK for message id=%x", msg.get_id());
 	g_unacknowledgedMessageCounter++;
 	msg.notify_timeout();
 	if (msg.is_request())
@@ -99,7 +100,7 @@ ProtocolError CoAPMessageStore::send(Message& msg, system_tick_t time)
 		if (coapType==CoAPType::CON)
 			coapmsg->prepare_retransmit(time);
 		else
-			coapmsg->set_expiration(time+CoAPMessage::MAX_TRANSMIT_SPAN);
+			coapmsg->set_expiration(time + CoAPMessage::EXCHANGE_LIFETIME);
 		add(*coapmsg);
 	}
 	return NO_ERROR;
@@ -150,7 +151,7 @@ ProtocolError CoAPMessageStore::receive(Message& msg, Channel& channel, system_t
 				return INSUFFICIENT_STORAGE;
 			// the timeout here is ideally purely academic since the application will respond immediately with an ACK/RESET
 			// which will be stored in place of this message, with it's own timeout.
-			coapmsg->set_expiration(time+CoAPMessage::MAX_TRANSMIT_SPAN);
+			coapmsg->set_expiration(time + CoAPMessage::EXCHANGE_LIFETIME);
 			add(*coapmsg);
 		}
 	}
--- a/communication/src/coap_channel.h
+++ b/communication/src/coap_channel.h
@@ -29,6 +29,8 @@ namespace particle
 namespace protocol
 {
 
+inline bool time_has_passed(system_tick_t now, system_tick_t tick);
+
 /**
  * Decorates a MessageChannel with message ID management as required by CoAP.
  * When a message is sent that doesn't have an assigned ID, it is assigned the next available ID.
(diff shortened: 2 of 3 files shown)


communication/src/coap_channel.h
```diff
@@ -149,20 +164,34 @@ class __attribute__((packed)) CoAPMessage
...
 
 	/**
 	 * The number of outstanding messages allowed.
 	 */
-	static const uint8_t NSTART = 1;
+	static constexpr const auto NSTART = 1;
 
 
-	CoAPMessage(message_id_t id_) : next(nullptr), timeout(0), id(id_), transmit_count(0), delivered(nullptr), data_len(0) {
+	CoAPMessage(message_id_t id_) :
```
m-mcgowan: Please be sure the clang_format file has these formatting changes.

communication/src/coap_channel.h
```diff
@@ -218,25 +249,37 @@ class __attribute__((packed)) CoAPMessage
 	bool prepare_retransmit(system_tick_t now)
 	{
 		CoAPType::Enum coapType = CoAP::type(get_data());
-		if (coapType==CoAPType::CON) {
-			timeout = now + transmit_timeout(transmit_count);
-			transmit_count++;
-			return transmit_count <= MAX_RETRANSMIT+1;
+		if (coapType == CoAPType::CON) {
```
m-mcgowan: While this technically is closer to the spec (taking the timeout with a pre-applied random factor and doubling it vs doubling the timeout and then applying a random factor), the difference in practice is not significant, especially considering the device service uses an entirely different set of parameters.  Both approaches are within the operational constraints of how long the server will wait since the maximum timeout ends up being the same.

I'd like to drive this change from observed problems rather than from a strict adherence to the spec since this is a non-trivial change, so I want to be sure the gain outweighs the risk.  

In terms of gain, I appreciate the inline references to the spec, which adds clarity.  Also welcome other voices here about which is easiest to comprehend since that is a key factor here too. I am on the fence, but can imagine others might find this easier to grasp.

communication/src/dtls_protocol.cpp
```diff
@@ -13,8 +13,13 @@ void DTLSProtocol::init(const char *id,
...
 	memcpy(device_id, id, sizeof(device_id));
-	// send a ping once every 23 minutes
-	initialize_ping(23*60*1000,30000);
+	// Sets ping interval by default to 23 minutes. This is Electron-specific default
+	// value, however all the other platforms will override this currently depending
+	// on various conditions.
+	// IMPORTANT: the second argument is the timeout waiting for the ping
+	// acknowledgment. This SHOULD be set to MAX_TRANSMIT_WAIT, which is 93s
+	// for default transmission parameters defined in RFC7252. If required we might lower it.
+	initialize_ping(23 * 60 * 1000, CoAPMessage::MAX_TRANSMIT_WAIT);
```
m-mcgowan: Bravo, I never liked the arbitrary choice of 30s.   Ideally these magic values and the CoAP retries and timeouts should be parametrizable from the system layer.

communication/src/coap_channel.cpp
```diff
@@ -150,7 +151,7 @@ ProtocolError CoAPMessageStore::receive(Message& msg, Channel& channel, system_t
 				return INSUFFICIENT_STORAGE;
 			// the timeout here is ideally purely academic since the application will respond immediately with an ACK/RESET
 			// which will be stored in place of this message, with it's own timeout.
-			coapmsg->set_expiration(time+CoAPMessage::MAX_TRANSMIT_SPAN);
+			coapmsg->set_expiration(time + CoAPMessage::EXCHANGE_LIFETIME);
```
m-mcgowan: The exchange lifetime is much larger than the max transmit span, meaning we will have to keep many more messages in memory so that their responses can be resent on retransmission from the server.  Just something to thing about - we should ideally run memory profiling over a busy CoAP connection with very slow acknowledgements and see how much memory is required. 

Expecting an application to wait some 300 seconds (5 minutes) for a response seems unrealistic. Reducing the MAX_LATENCY from 100s to 5-10s seems reasonable given our use case. and would reduce the EXCHANGE_LIFETIME to something more manageable. 

Going forward, I propose that we aim for shorter timeouts and more frequent retransmits, which we can have as a negotiation step between the device and the server. That way we can tailor the timeouts to the network characteristics, data cost, and the application needs.

communication/src/coap_channel.h
```diff
@@ -149,20 +164,34 @@ class __attribute__((packed)) CoAPMessage
...
+	// RFC7252 4.8. Transmission Parameters
+	// We are NOT using the default RFC values
+	static constexpr const auto ACK_TIMEOUT = 4000;
+	static constexpr const auto ACK_RANDOM_FACTOR = 1500;
+	static constexpr const auto ACK_RANDOM_DIVISOR = 1000;
+	static constexpr const auto MAX_RETRANSMIT = 3;
+	// RFC7252 4.8.2. Time Values Derived from Transmission Parameters
+	static constexpr const auto MAX_TRANSMIT_SPAN = (ACK_TIMEOUT * ((1 << MAX_RETRANSMIT) - 1) * ACK_RANDOM_FACTOR) / ACK_RANDOM_DIVISOR;
+	static constexpr const auto MAX_TRANSMIT_WAIT = (ACK_TIMEOUT * ((1 << (MAX_RETRANSMIT + 1)) - 1) * ACK_RANDOM_FACTOR) / ACK_RANDOM_DIVISOR;
+	static constexpr const auto MAX_LATENCY = 100000;
```
m-mcgowan: 100s is not realistic for our use case.

PR #1776, opened 2019-05-14
expires a session after 3 unsuccessful attempts at connecting to the cloud

m-mcgowan:
### Problem

The device and cloud may get out of sync when reusing a session. For instance, session data may become corrupted on the device, or session caches on the cloud expired. 

When this occurs the device presently retries indefinitely to use the current session. 

### Solution

The number of times the session is used is counted - each time the session is restored as part of the main handshake, the use count is incremented.  When the use count reaches the expiration count (presently 3), the session is considered invalid, forcing it to be discarded and performing a full handshake. 

The use count is reset after successfully decrypting a message from the cloud. This validates the session data.

### Steps to Test

* Apply the randbytes hack, but rather than invert, we increment the data (or the session becomes valid every 2nd attempt.)
* Allow the device to connect to the cloud and then reset. Every 3rd try the device does a full handshake.

### References

* [ch32368](https://app.clubhouse.io/particle/story/32368/devices-can-expire-a-session-after-a-number-of-unsuccessful-connection-attempts)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/communication/src/dtls_message_channel.cpp
+++ b/communication/src/dtls_message_channel.cpp
@@ -87,14 +87,16 @@ auto SessionPersist::restore(mbedtls_ssl_context* context, bool renegotiate, uin
 		return NO_SESSION;
 	}
 
-    increment_use_count();
-    save(saver);
-
     LOG(WARN, "session has %d uses", use_count());
 	if (has_expired()) {
+	    invalidate();
+	    save(saver);
 	    LOG(WARN, "session has expired after %d uses", use_count());
 	    return NO_SESSION;
 	}
+
+    increment_use_count();
+    save(saver);
 
 	// assume invalid initially. With the ssl context being reset,
 	// we cannot return NO_SESSION from this point onwards.
@@ -412,6 +414,7 @@ ProtocolError DTLSMessageChannel::establish(uint32_t& flags, uint32_t app_state_
 ProtocolError DTLSMessageChannel::notify_established()
 {
 	sessionPersist.make_persistent();
+	sessionPersist.save(callbacks.save);
 	return NO_ERROR;
 }
 
@@ -468,9 +471,11 @@ ProtocolError DTLSMessageChannel::receive(Message& message)
  */
 void DTLSMessageChannel::cancel_move_session()
 {
-    move_session = false;
-    sessionPersist.clear_use_count();
-    command(SAVE_SESSION);
+    if (move_session) {
+        move_session = false;
+        sessionPersist.clear_use_count();
+        command(SAVE_SESSION);
+    }
 }
 
 ProtocolError DTLSMessageChannel::send(Message& message)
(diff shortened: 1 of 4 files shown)


communication/src/dtls_session_persist.h
```diff
@@ -114,6 +114,12 @@ class __attribute__((packed)) SessionPersistOpaque : public SessionPersistData
 
 	void invalidate() { size = 0; }
 
+	void increment_use_count() { use_counter++; }
+	void clear_use_count() { use_counter = 0; }
+	int use_count() { return use_counter; }
+	bool has_expired() { return use_counter >= maximumSessionUses; }
+
+	static const int maximumSessionUses = 3;
```
sergeuz: Let's use all caps for constant names: `MAXIMUM_SESSION_USES`

communication/src/dtls_message_channel.cpp
```diff
@@ -455,6 +456,18 @@ ProtocolError DTLSMessageChannel::receive(Message& message)
...
+/**
+ * Once data has been successfully received we can stop
+ * sending move-session messages.
+ * This is also used to reset the expiration counter.
+ */
+void DTLSMessageChannel::cancel_move_session()
+{
+    move_session = false;
+    sessionPersist.clear_use_count();
+    command(SAVE_SESSION);
```
sergeuz: I'd add a check here and avoid invoking the `SAVE_SESSION` command if `use_counter` is already 0. The channel implementation calls `cancel_move_session()` on every message it receives

communication/src/dtls_message_channel.cpp
```diff
@@ -405,7 +407,6 @@ ProtocolError DTLSMessageChannel::establish(uint32_t& flags, uint32_t app_state_
 ProtocolError DTLSMessageChannel::notify_established()
 {
 	sessionPersist.make_persistent();
-	sessionPersist.save(callbacks.save);
```
sergeuz: So we don't actually need to store the `persistent` flag itself in the session's persistent storage, right?

communication/src/dtls_message_channel.cpp
```diff
@@ -94,9 +87,19 @@ auto SessionPersist::restore(mbedtls_ssl_context* context, bool renegotiate, uin
 		return NO_SESSION;
 	}
 
+    increment_use_count();
+    save(saver);
+
+    LOG(WARN, "session has %d uses", use_count());
+	if (has_expired()) {
```
sergeuz: Mixed usage of spaces and tabs for indentation

communication/src/dtls_message_channel.cpp
```diff
@@ -94,9 +87,19 @@ auto SessionPersist::restore(mbedtls_ssl_context* context, bool renegotiate, uin
 		return NO_SESSION;
 	}
 
+    increment_use_count();
+    save(saver);
```
sergeuz: We should probably check if the session is expired first and then increment the counter (or change the implementation of `has_expired()`), otherwise the session will be invalidated after 2 attempts instead of 3 (not sure if that was the intent)

communication/src/dtls_message_channel.cpp
m-mcgowan: This was an erroneous delete - will fix up, thanks for spotting it!

Issue #1772, opened 2019-05-13
[Gen3] Multiple pinMode() issues


ScruffR:
Testing some combinations of `pinMode()` and `analogWrite()` on Argon (1.1.0-rc.1&2 & 1.2.0-beta.1) I found some enexpected behaviours

 - despite being documented as PWM capable D4 does not work with `analogWrite()`
 - after using `analogWrite()` on any pin `digitalWrite(pin, HIGH)` will never set the GPIO HIGH but LOW. 
 - D7 is by default set to OUTPUT - which is unexected as in previous generations and versions D7 was just the same as all other pins set INPUT (hi-Z) by default. This behaviour cost me some time to realise as I often use the D7 LED for quick testing whether another pin behaves as expected by jumpering to D7 but with that unexpected behaviour I suspected some issue with the tested pin before I realised this break in common practice

YutingYou:
@ScruffR  Thanks for reporting the issue!

D4 pin should be configured to OUTPUT mode before` analogWrite()`
> analogWrite() (PWM)
> ...
> NOTE: pinMode(pin, OUTPUT); is required before calling analogWrite(pin, value); or else the pin will not be initialized as a PWM output and set to the desired duty cycle.

Other issues have been fixed in https://github.com/particle-iot/device-os/pull/1777

ScruffR:
I did have that and I tested with `digitalWrite()` before actually trying `analogWrite()`.

YutingYou:
@ScruffR It worked on my side, can you please share your code?

ScruffR:
This is my code
```
const uint8_t pins[] = { A0, A1, A2, A3, D2, D3, D4, D5, D6, D8 };

int aw(const char* arg) {
  int v = atoi(arg);
  
  for (int i=0; i < sizeof(pins); i++)
    analogWrite(pins[i], v);
    
  return v;
}

int dw(const char* arg) {
  int v = atoi(arg);
  
  for (int i=0; i < sizeof(pins); i++)
    digitalWrite(pins[i], v);
    
  return v;
}

void setup()
{
  pinMode(D7, INPUT);
  
  for (int i=0; i < sizeof(pins); i++)
    pinMode(pins[i], OUTPUT);
  
  Particle.function("analogWrite" , aw);
  Particle.function("digitalWrite", dw);
}
```

[...]

PR #1777, opened 2019-05-15
Fix/hal/gen3 pinmode

YutingYou:
### Problem

1. User LED is configured to OUTPUT mode when initializing.
2. Pin mode is changed to default mode after calling `analogWrite()` and `digitalWrite()`

Reference issue:
https://github.com/particle-iot/device-os/issues/1772

### Steps to Test
1. Flash example
2. If D7 is OUTPUT mode, the device will enter PANIC mode
3. Use a multimeter to measure OUTPUT value from D4 pin, it should be high now

### Example App

```c
void setup(void) {
    SPARK_ASSERT(getPinMode(D7) == PIN_MODE_NONE);

    pinMode(D4, OUTPUT);
    analogWrite(D4, 128);
    digitalWrite(D4, 1);
}
```

### References

[CH32946]

---

### Completeness

- [ ] Added documentation


hal/src/nRF52840/rgbled_hal.c
```diff
@@ -126,7 +126,6 @@ void LED_Init(Led_TypeDef Led) {
         return;
     }
 
-    HAL_Pin_Mode(HAL_Leds[Led].pin, OUTPUT);
```
avtolstoy: Does this affect RGB LED pins in any way?

YutingYou: We use PWM for RGB LED, PWM will overwrite GPIO configuration, so RGB LEDs work correctly.
But it will affect USER LED.

`LED_Init()` is only called once to initialize all LEDs at boot, if we don't configure it output, USER LED should not work.
```
HAL_Core_Led_Mirror_Pin() -> HAL_Led_Init() -> LED_Init()
Set_System() -> LED_Init()
```
Gen2 should have this issue too, I found Gen2 also configures USER LED as output in `LED_Init()`, should we leave the output configuration?

avtolstoy: :+1:

I see no reason why we should by default configure it as output. We should double check the current behavior for Gen 2 though.

YutingYou:
Change the fix for user LED issue.
Fixes PWM channel duplicate assignment.

Need rereview, thanks. cc: @avtolstoy

technobly:
I see the PWM channels were changed in this PR for D5-D7, but I don't see why in the problem/solution.  Does this affect the documentation? @avtolstoy @YutingYou

YutingYou:
@technobly  This PR is the fix for issue #1772, it includes pin mode issue and duplicated PWM channel.

> Does this affect the documentation?

It doesn't. It's for the Nordic driver, we only care about PWM instance in Particle PWM chapter.
image

technobly:
@YutingYou ok thanks, I see now that the change looks to align all Gen 3 platforms to use the same channels for PWM on D5-D7.  I noticed you didn't list the D7 output above, yet PWM is available on D7.

YutingYou:
@technobly D7 works with RGB LED. It should be this part in Particle Docs:
>Group 3: Pins D2, D3, A4, and A5.
>Group 2: Pins A0, A1, A2, and A3.
>Group 1: Pins D4, D5, D6, and D8.
>Group 0: Pin D7 and the RGB LED. This must use the default resolution of 8 bits (0-255) and frequency of 500 Hz.

PR #1778, opened 2019-05-15
[Gen 3] Fixes various issues caused by the gateway reset

avtolstoy:
### Problem

Mesh devices may be unable to connect to the cloud or appear to be connected but can neither send nor receive data to/from the cloud under certain conditions.

### Solution

1. OpenThread version we've upgraded to in the end of January moved SLAAC functionality into its core and enabled it by default, which interfered with our own SLAAC logic. A PR was submitted later to the OpenThread repo to make this configurable, so for now we are simply pulling in that PR and disable SLAAC: https://github.com/particle-iot/openthread/commit/ea382ea6002ce41af4151bac64e26b36a1e3cb49
2. Under certain conditions (see 'Steps to Test' to replicate) mesh devices don't notice that the gateway has been reset. This is resolved by artificially publishing a prefix with `preferred` flag set to 0 (which means that the addresses generated from this prefix should be 'DEPRECATED' but may still be used for communications), waiting for this change to propagate through the network and then re-publishing with proper `preferred` set to 1, so that the device can receive events from OpenThread about a potential change in upstream connection availability and explicitly check their cloud connection state by issuing a ping.
3. When changing the state of IP addresses on OpenThread interface, make sure that events are being fired, so that system level subscribers may understand that there is a potential change in connection availability.
4. Make sure that `Pref::1/PrefLen` address is always generated (instead of a random one) from the prefix announced by the gateway itself no matter the current role (child or router).

### Steps to Test

The easiest way to test this is to create a mesh network of 3 devices.

1. Power down all devices
2. Power up non-gateway devices
3. Make sure that one of them becomes the leader (for now take a look at the logs for that)
4. Power on the gateway device
5. Wait until all three are connected to the cloud and are breathing cyan
6. Power off the gateway device for about a minute or so. The other two devices should stay breathing cyan (occasionally you might see some blinks due to ping timeouts and session resumption)
7. Power the gateway back on BEFORE the other devices start blinking green
8. All three should connect to the cloud. The non-gateway devices should not go through blinking green state.
9. Make sure you may reach your non-gateway devices from the cloud (e.g. `particle nyan`)
10. Wait for some minutes, you should still be able to reach your non-gateway devices and they should be breathing cyan

Steps 9 - 10 should fail on `develop` and should work correctly with this PR.

### Example App

N/A

### References

- [CH29981]

---

### Completeness

- [ ] Added documentation

---

- [bugfix] [gen 3] Fixes various issues caused by the gateway reset [#1778](https://github.com/particle-iot/device-os/pull/1778)

### DIFF
--- a/hal/network/lwip/openthread/lwip_openthreadif.cpp
+++ b/hal/network/lwip/openthread/lwip_openthreadif.cpp
@@ -350,7 +350,78 @@ void OpenThreadNetif::otStateChangedCb(uint32_t flags, void* ctx) {
     self->stateChanged(flags);
 }
 
+void OpenThreadNetif::debugLogOtStateChange(uint32_t flags) {
+    if (flags & OT_CHANGED_IP6_ADDRESS_ADDED) {
+        LOG_DEBUG(TRACE, "OT_CHANGED_IP6_ADDRESS_ADDED");
+    }
+    if (flags & OT_CHANGED_IP6_ADDRESS_REMOVED) {
+        LOG_DEBUG(TRACE, "OT_CHANGED_IP6_ADDRESS_REMOVED");
+    }
+    if (flags & OT_CHANGED_THREAD_ROLE) {
+        LOG_DEBUG(TRACE, "OT_CHANGED_THREAD_ROLE");
+    }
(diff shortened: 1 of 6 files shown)


avtolstoy:
- [CH29981]

PR #1779, opened 2019-05-15
[Electron/LTE] Make sure that the RAT information is actual before calculating signal strength (RSSI) and quality

avtolstoy:
### Problem

Invalid RAT information causes signal strength/quality information to be unavailable. This is caused by `AT+COPS?` in `checkNetStatus()` which invalidates the RAT set by URC handlers as it only handles `GSM` and `UTRAN` for some reason: https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L1380

### Solution

1. Fix `AT+COPS?` handler to correctly parse all RATs
2. Adds explicit `AT+COPS?` call to `getSignalStrength()`

### Steps to Test

N/A

### Example App

N/A

### References

- [CH32382]
- #1759 

---

### Completeness

- [ ] Added documentation

---

- [bugfix] [Electron/LTE] Make sure that the RAT information is actual before calculating signal strength (RSSI) and quality [#1779](https://github.com/particle-iot/device-os/pull/1779)

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1277,6 +1277,12 @@ bool MDMParser::getSignalStrength(NetStatus &status)
     if (_init && _pwr) {
         MDM_INFO("\r\n[ Modem::getSignalStrength ] = = = = = = = = = =");
 
+        // Reformat the operator string to be numeric
+        // (allows the capture of `mcc` and `mnc`)
+        sendFormated("AT+COPS=3,2\r\n");
+        if (RESP_OK != waitFinalResp(nullptr, nullptr, COPS_TIMEOUT))
+            goto cleanup;
+
         // We do receive RAT changes asynchronously via CREG URCs, but
         // just in case we'll update it here explicitly.
         sendFormated("AT+COPS?\r\n");
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -118,6 +118,34 @@ inline void CLR_GPRS_TIMEOUT() {
     DEBUG("GPRS WD Cleared, was %d", gprs_timeout_duration);
 }
 
+namespace {
+
+AcT toCellularAccessTechnology(int v) {
+    switch (v) {
+        case 0: // GSM
+        case 1: // GSM COMPACT
+            return ACT_GSM;
+        case 2: // UTRAN
+            return ACT_UTRAN;
+        case 3: // GSM with EDGE availability
+            return ACT_EDGE;
+        case 4: // UTRAN with HSDPA availability
+        case 5: // UTRAN with HSUPA availability
+        case 6: // UTRAN with HSDPA and HSUPA availability
+            return ACT_UTRAN;
+        case 7: // LTE
+            return ACT_LTE;
+        case 8: // LTE Cat M1
+            return ACT_LTE_CAT_M1;
+        case 9: // LTE Cat NB1
+            return ACT_LTE_CAT_NB1;
+        default:
+            return ACT_UNKNOWN;
+    }
+}
+
+} // anonymous
+
 #ifdef MDM_DEBUG
  #if 0 // colored terminal output using ANSI escape sequences
   #define COL(c) "\033[" c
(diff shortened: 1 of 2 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -118,6 +118,34 @@ inline void CLR_GPRS_TIMEOUT() {
     DEBUG("GPRS WD Cleared, was %d", gprs_timeout_duration);
 }
 
+namespace {
+
+AcT toCellularAccessTechnology(int v) {
```
zfields: What is `v`, "value"? Please change to a more verbose/descriptive variable name. Maybe something like, `rat_code`?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -118,6 +118,34 @@ inline void CLR_GPRS_TIMEOUT() {
...
+    switch (v) {
+        case 0: // GSM
```
zfields: Instead of the magic numbers, can we pull [`CellularAccessTechnology`](https://github.com/particle-iot/device-os/blob/develop/hal/src/boron/network/cellular_ncp_client.h#L51) into a shared header and use the enumeration instead?

zfields:
I rebased on the new `v1.2.0`. I didn't realize you were merging into my branch.

You will need to rebase on me again.

hal/src/electron/modem/mdm_hal.cpp
technobly: I think `v` is ok here, `val` could have been better... but this function essentially tells you what v is... and technically it's just a number, but we do supply it here with input from a sscanf that happens to be an AcTStatus in one case and AcT in another case based on definition from u-blox modem.

zfields: can we call it `act`? It's infinitely more descriptive than `v`.

hal/src/electron/modem/mdm_hal.cpp
technobly: Honestly I think the magic numbers are fine, but if we wanted to supply an enum it should be one for the modems used like `UbloxCellularAccessTechnology`.  Down the road we will have to likely map the modem's codes to `CellularAccessTechnology` and also to this used in enums_hal.h
```
//! Access Technology
typedef enum {
    ACT_UNKNOWN     = 0,
    ACT_GSM         = 1,
    ACT_EDGE        = 2,
    ACT_UTRAN       = 3,
    ACT_CDMA        = 4,
    ACT_LTE         = 5,
    ACT_LTE_CAT_M1  = 6,
    ACT_LTE_CAT_NB1 = 7
} AcT;
```

technobly: And then the enum makes it a bit harder to use `toCellularAccessTechnology()`

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1202,6 +1221,14 @@ bool MDMParser::getSignalStrength(NetStatus &status)
     LOCK();
     if (_init && _pwr) {
         MDM_INFO("\r\n[ Modem::getSignalStrength ] = = = = = = = = = =");
+
+        // We do receive RAT changes asynchronously via CREG URCs, but
+        // just in case we'll update it here explicitly.
+        sendFormated("AT+COPS?\r\n");
+        if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
+            return ok;
```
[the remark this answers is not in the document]
avtolstoy: Oh my, you are right. Honestly, we should refactor locking here to make use of RAII to avoid such mishaps.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: I agree about the magic numbers, might as well fix that. Resolved in https://github.com/particle-iot/device-os/commit/5c168b39a4b5d90a94062d5fdc1410457e909e92

hal/src/electron/modem/mdm_hal.cpp
technobly: ended up with `rat`...

PR #1782, opened 2019-05-16
[wiring] Make sure that `Serial` and `SerialX` methods are in sync with the documentation and don't return unexpected values

### LINKED ISSUE #1737: [Gen3] USARTSerial::read() returns -260 instead of -1 for empty RX buffer
Unlinke the [documentation](https://docs.particle.io/reference/device-os/firmware/xenon/#read-) states 
image

and also in contradiction to the known behaviour of Gen1 & Gen2 devices Gen3 returns -260 instead of -1

Tested with 0.9.0

avtolstoy:
### Problem

#1737

### Solution

Make sure that `available()`, `availableForWrite()`, `peek()` and `read()` don't forward the errors from HAL and only return 0 or -1 in case of errors appropriately.

### Steps to Test

N/A

### Example App

N/A

### References

- Closes #1737
- [CH32372]

---

### Completeness

- [ ] Added documentation

---

- [wiring] Make sure that `Serial` and `SerialX` methods are in sync with the documentation and don't return unexpected values [#1782](https://github.com/particle-iot/device-os/pull/1782)

### DIFF
--- a/wiring/src/spark_wiring_usbserial.cpp
+++ b/wiring/src/spark_wiring_usbserial.cpp
@@ -65,24 +65,23 @@ void USBSerial::end()
 // Read data from buffer
 int USBSerial::read()
 {
-	return HAL_USB_USART_Receive_Data(_serial, false);
+	return std::max(-1, (int)HAL_USB_USART_Receive_Data(_serial, false));
 }
 
 int USBSerial::availableForWrite()
 {
-  return HAL_USB_USART_Available_Data_For_Write(_serial);
+  return std::max(0, (int)HAL_USB_USART_Available_Data_For_Write(_serial));
 }
 
 int USBSerial::available()
 {
-	return HAL_USB_USART_Available_Data(_serial);
+	return std::max(0, (int)HAL_USB_USART_Available_Data(_serial));
 }
 
 size_t USBSerial::write(uint8_t byte)
 {
   if (HAL_USB_USART_Available_Data_For_Write(_serial) > 0 || _blocking) {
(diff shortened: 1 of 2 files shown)


m-mcgowan:
I'm not sure about this - rather than breaking the . behaviour wouldn't;'t it be better to instead update the documentation? That way we make no breaking changes to existing apps.

avtolstoy:
We are not really breaking the behavior, but restoring it here at the very least for Gen 3? :smile: Currently `Serial1.read()` may return `-260` on Gen 3 (`SYSTEM_ERROR_NO_MEMORY`, which is actually somewhat wrong), and I assume many apps simply check for `!= -1` instead of `< 0` for error conditions.

We've discussed this previously and came to a conclusion that as a first step we'd ensure that no error codes are propagating up to the wiring caller and we are in conformance with the documentation and then later would change the behavior, documentation as well as notify about the breaking change in the release notes. I agree though that with 1.2.0 we may go straight to doing that. So, what do we do?

wiring/src/spark_wiring_usbserial.cpp
```diff
@@ -99,7 +98,7 @@ void USBSerial::blockOnOverrun(bool block)
 
 int USBSerial::peek()
 {
-	return HAL_USB_USART_Receive_Data(_serial, true);
+	return std::max(0, (int)HAL_USB_USART_Receive_Data(_serial, true));
```
sergeuz: This should be `std::max(-1, ...)`

PR #1783, opened 2019-05-16
[Photon / P1] Fixes MAC address info not being available in listening mode

### LINKED ISSUE #1781: [P1] Particle CLI returning MAC address 00:00:00:00:00:00 on Device OS 1.1.0
### Expected Behavior
The device should return its real MAC address when requested through Particle CLI. At least, this is the case with Device OS previous to v1.1.0, now it always outputs 00:00:00:00:00:00. I believe it is not a problem with the latest version of Particle CLI (1.41.0) because the MAC is correct if I reflash Device OS v1.0.1.

I have noticed that the ``particle serial mac`` command is no longer listed in the [documentation](https://docs.particle.io/reference/developer-tools/cli/), although it still appears in the Particle CLI command line help. I ignore if this behaviour change is intentional or not. If it is, I would appreciate having an alternative option to locally check the MAC address.

### Steps to Reproduce
1) Flash Device OS 1.1.0
2) Enter listening mode
3) Type ``particle serial mac``

avtolstoy:
### Problem

#1781

This is an artifact of Gen 2 / Gen 3 merge where we've changed the behavior of listening mode: it's no longer blocking and is a part of the system loop. This causes network config to be zeroed out by `manage_ip_config()` -> [`update_config()`](https://github.com/particle-iot/device-os/blob/develop/system/src/system_network_internal.h#L786).

### Solution

A simplest fix is probably just force-filling the configuration in the API `network_config()` call.

### Steps to Test

Put your Photon into listening mode, call `particle serial mac`. The MAC address should be available.

### Example App

N/A

### References

- Closes #1781

---

### Completeness

- [ ] Added documentation

---

- [bugfix] [Photon/P1] Fixes MAC address info not being available in listening mode [#1783](https://github.com/particle-iot/device-os/pull/1783)

### DIFF
--- a/system/src/system_network_compat.cpp
+++ b/system/src/system_network_compat.cpp
@@ -101,6 +101,7 @@ void HAL_NET_notify_dhcp(bool dhcp)
 
 const void* network_config(network_handle_t network, uint32_t param, void* reserved)
 {
+    nif(network).update_config(true);
     return nif(network).config();
 }
 


incorvia:
Is there anywhere I can read more information about this comment: "we've changed the behavior of listening mode: it's no longer blocking and is a part of the system loop".

PR #1785, opened 2019-05-16
[gen 3] update bootloader dependency to v302

technobly:
### Problem

After merging Gen 3 and Gen 2 firmware bases together, there was no way to bump the bootloader dependency separately for each generation of device.

### Solution

Separate the BOOTLOADER_DEPENDENCY for each generation based on a new definition PLATFORM_GEN, which is set to 0, 1, 2, 3, and 60000 (of course).

Backstory:
- v1.1.0 contains bootloader v301, and dependency of v201 for both platforms.  However Gen 3 devices  (except for SoMs) and Electron contained the bootloader in system firmware, so they would apply the bootloader if it was less than v301.  Photon/P1 don't contain bootloaders though, so they rely on the CLI  via `particle update` or the Cloud to update them.  The CLI only gets updated on default releases and did get v301 bootloaders for Photon/P1 added.  That said, the dependency was still v201 so if they get Cloud updated to v1.1.0 they would not request v301 bootloader from the Cloud.  There is only a minor update for Gen 2 bootloaders and the thought is that it's not absolutely required, so let's not risk an update to the bootloader just yet.

- v1.2.0-rc.1 will contain this PR which is still v302 bootloaders (bumped this release due to #1777 ), and also now will have a separate dependency for Gen 2 and Gen 3.  For now, we will leave Gen 2 as is depending on v201, but we will bump Gen 3 to v302 because also in this PR we have removed the bootloaders from Gen 3 devices.  If a user updates from 0.9.0 to 1.2.0-rc.1, they would not get the required updates via the CLI since we only update `particle update` on default releases.  So they will need to get it from the Cloud or via manual Y-Modem protocol update over USB Serial with `particle flash --serial bootloader.bin`.  To force the Cloud update we need to bump the dependency version to v302.

- v1.2.0 default - when this happens, we can update the dependency for all to v302 and Photon/P1 can get the minor update easily via the CLI or via the Cloud if they have an internet connection at the time.  And just to state it one more time, the bootloader can always be manually updated via Y-Modem protocol over USB Serial with `particle flash --serial bootloader.bin`.

### Steps to Test

- Run unit tests
- Clean build with different platforms and add an "error2" or "error3" inserted next to the BOOTLOADER_DEPENDENCY to ensure the build is picking up the correct one.

---

### Completeness

- [x] Added documentation

---

- [gen 3] update bootloader dependency to v302 [#1785](https://github.com/particle-iot/device-os/pull/1785)

### DIFF
--- a/modules/shared/system_module_version.mk
+++ b/modules/shared/system_module_version.mk
@@ -32,7 +32,16 @@ endif
 # Skip to next 100 every v0.x.0 release (e.g. 11 for v0.6.2 to 100 for v0.7.0-rc.1),
 # but only if the bootloader has changed since the last v0.x.0 release.
 # Bump by 1 for every updated bootloader image for a release with the same v0.x.* base.
-BOOTLOADER_VERSION ?= 301
+BOOTLOADER_VERSION ?= 302
 
-# the version of the bootloader that the system firmware requires
+# The version of the bootloader that the system firmware requires
+# NOTE: this will force the device into safe mode until this dependency is met, which is why
+# this version usually lags behind the current bootloader version, to avoid non-mandatory updates.
+ifeq ($(PLATFORM_GEN),2)
 BOOTLOADER_DEPENDENCY = 201
+else ifeq ($(PLATFORM_GEN),3)
+BOOTLOADER_DEPENDENCY = 302
+else
+# Some sensible default
+BOOTLOADER_DEPENDENCY = 0
+endif
--- a/build/platform-id.mk
+++ b/build/platform-id.mk
@@ -2,15 +2,16 @@ ifeq ($(included_productid_mk),)
 included_productid_mk := 1
 
 # defines
-# PLATFORM_NAME - a unique name for the platform, can be used to organise sources
+# PLATFORM_NAME - a unique name for the platform, can be used to organize sources
 #                 by platform
+# PLATFORM_GEN  - a number that identifies the generation of device
 # PLATFORM_MCU  - an identifier for the MCU family
 # PLATFORM_NET  - the network subsystem
 # STM32_DEVICE  - the specific device being targeted for STM32 platform builds
-# ARCH		- architecture (ARM/GCC)
+# ARCH          - architecture (ARM/GCC)
 # PRODUCT_DESC  - text description of the product ID
 # PLATFORM_DYNALIB_MODULES - if the device supports a modular build, the name
-#		- of the subdirectory containing
+#                          - of the subdirectory containing
 
 # Default USB Device Vendor ID for Spark Products
 USBD_VID_SPARK=0x1D50
@@ -112,6 +113,7 @@ ARCH=arm
 ifeq ("$(PLATFORM_ID)","0")
 STM32_DEVICE=STM32F10X_MD
 PLATFORM=core
+PLATFORM_GEN=1
 PLATFORM_NAME=core
 PLATFORM_MCU=STM32F1xx
 PLATFORM_NET=CC3000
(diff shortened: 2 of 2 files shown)


modules/shared/system_module_version.mk
```diff
@@ -34,5 +34,14 @@ endif
...
-# the version of the bootloader that the system firmware requires
+# The version of the bootloader that the system firmware requires
+# NOTE: this will force the device into safe mode until this dependency is met, which is why
+# this version usually lags behind the current bootloader version, to avoid non-mandatory updates.
+# gen 2 devices
+ifneq (,$(filter $(PLATFORM_ID),6 8 10))
 BOOTLOADER_DEPENDENCY = 201
+endif
+# gen 3 devices
+ifneq (,$(filter $(PLATFORM_ID),12 13 14 22 23 24))
```
m-mcgowan: I'm a little uneasy about explicitly listing Gen3 platforms...I feel we should apply the latest bootloader unless otherwise specified? This is the kind of thing that is easy to forget to update.

technobly: Updating the bootloader is kind of a risk vs reward proposition.  Thus the reason why the dependency is currently lagging the actual version.  When we get to the default release though, that seems like a good time to take the minor step to sync everything up as I stated in the PR.  I'm not sure what the issue is with explicitly listing the platforms, we have 9 modular platforms and they are all accounted for.

modules/shared/system_module_version.mk
```diff
@@ -34,5 +34,12 @@ endif
 # Bump by 1 for every updated bootloader image for a release with the same v0.x.* base.
 BOOTLOADER_VERSION ?= 301
 
-# the version of the bootloader that the system firmware requires
+# The version of the bootloader that the system firmware requires
+# NOTE: this will force the device into safe mode until this dependency is met, which is why
+# this version usually lags behind the current bootloader version, to avoid non-mandatory updates.
+ifneq (,$(filter $(PLATFORM_GEN),2))
```
avtolstoy: `ifneq ($(PLATFORM_GEN),2)` ? It's not a list anymore. :smile:

modules/shared/system_module_version.mk
```diff
@@ -34,5 +34,12 @@ endif
...
 BOOTLOADER_DEPENDENCY = 201
+endif
+ifneq (,$(filter $(PLATFORM_GEN),3))
+BOOTLOADER_DEPENDENCY = 301
```
avtolstoy: We should bump this for 1.2.0-rc.1 due to https://github.com/particle-iot/device-os/pull/1777/files#diff-9de5c4685f6a4024eb7b8121590af8b6R136.

technobly: Yes we need a better way to detect when the bootloader needs to be updated :)

modules/shared/system_module_version.mk
technobly: Totally

modules/shared/system_module_version.mk
```diff
@@ -32,7 +32,14 @@ endif
...
 # but only if the bootloader has changed since the last v0.x.0 release.
 # Bump by 1 for every updated bootloader image for a release with the same v0.x.* base.
-BOOTLOADER_VERSION ?= 301
+BOOTLOADER_VERSION ?= 302
 
-# the version of the bootloader that the system firmware requires
+# The version of the bootloader that the system firmware requires
+# NOTE: this will force the device into safe mode until this dependency is met, which is why
+# this version usually lags behind the current bootloader version, to avoid non-mandatory updates.
+ifeq ($(PLATFORM_GEN),2)
```
avtolstoy: Consider changing this to:
```suggestion
ifeq ($(PLATFORM_GEN),2)
BOOTLOADER_DEPENDENCY = 201
else ifeq ($(PLATFORM_GEN),3)
BOOTLOADER_DEPENDENCY = 302
else
# Some sensible default
BOOTLOADER_DEPENDENCY = 0
endif
```

PR #1797, opened 2019-05-28
BLE: to not pollute system listening mode handler class.

XuGuohui:
**This PR targets currently `feature/ble/develop` branch.**

- [CH33672]


system/src/system_listening_mode.h
```diff
@@ -33,7 +33,29 @@
 #include "ble_hal.h"
 #endif /* HAL_PLATFORM_BLE */
 
-namespace particle { namespace system {
+namespace particle { namespace system {\
```
avtolstoy: Stray `\`

system/src/system_listening_mode.cpp
```diff
@@ -395,4 +230,177 @@ int ListeningModeHandler::clearNetworkConfiguration() const {
...
+
+int BleListeningModeHandler::exit() {
+    /*
+     * TODO: Would it be better to call the method of BLE Control Request Channel to isolate the logic?
+     *
+     * FIXME: It's not possible to restore the final state that is changed by user application
+     * during in the listening mode when threading is enabled.
+     */
+    SCOPE_GUARD ({
+        if (preAdvData_) {
```
avtolstoy: Prefer `std::unique_ptr<uint8_t[]>` to just `uint8_t*`.

system/src/system_listening_mode.cpp
```diff
@@ -395,4 +230,177 @@ int ListeningModeHandler::clearNetworkConfiguration() const {
     return r;
 }
 
+#if HAL_PLATFORM_BLE
+BleListeningModeHandler::BleListeningModeHandler()
```
avtolstoy: Let's decouple this class into a separate header and source file.

PR #1798, opened 2019-05-28
[gen 3] hal: added apn user credentails to sara_ncp_client

jaredwolff:
### Problem

Currently the SDK supports using 3rd party SIMs. This includes the ability to enter a custom APN, user and password. The problem? That information doesn't currently get forwarded on to the SARA-R410M module. Plus if a username and password is used, it causes the program to execute an AT+CGDCONT command which outright fails (by using the auth prefix `CHAP:` before the APN name)

### Solution

This addresses the issue by removing the logic for inserting the authentication type into the `AT+CGDCONT` command. 

It also adds an `AT+UAUTHREQ` which indicates not only the authentication type but the username and password. In my testing I noticed that executing these commands on PDP context 1 does authenticate properly. While using PDP context 2 does work. This would require further testing on other SIMs and networks. (I only have one SIM to work with right now)

### Steps to Test

A simple test is to use a Soracom SIM and the example code below. (In my case I have changed the log level in the HAL system code to allow for TRACE logs to get feedback on the connection process). I've also tested with the INTERNAL_SIM to make sure I didn't break underlying functionality.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {

    Serial.begin();

    Cellular.setActiveSim(EXTERNAL_SIM);
    Cellular.setCredentials("soracom.io","sora","sora");

    // This is just so you know the operation is complete
    pinMode(D7, OUTPUT);
    digitalWrite(D#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {

    Serial.begin();

    Cellular.setActiveSim(EXTERNAL_SIM);
    Cellular.setCredentials("soracom.io","sora","sora");

    // Cellular.clearCredentials();

    // This is just so you know the operation is complete
    pinMode(D7, OUTPUT);
    digitalWrite(D7, HIGH);

    Particle.connect();
}

void loop() {

    Cellular.command("AT+CGDCONT?\r\n");
    Cellular.command("AT+CREG?\r\n");

    delay(5000);

}
```

### References

[Here's where I raised the issue yesterday.](https://community.particle.io/t/solved-boron-3rd-party-sim-soracom/)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -889,11 +889,27 @@ int SaraNcpClient::configureApn(const CellularNetworkConfig& conf) {
         netConf_ = networkConfigForImsi(buf, strlen(buf));
     }
     // FIXME: for now IPv4 context only
-    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s%s\"",
-            (netConf_.hasUser() && netConf_.hasPassword()) ? "CHAP:" : "",
+    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s\"",
             netConf_.hasApn() ? netConf_.apn() : "");
-    const int r = CHECK_PARSER(resp.readResult());
+    int r = CHECK_PARSER(resp.readResult());
     CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
+
+    // If we have a password and user, set it to PDP context #2
+    // Set the type of auth (CHAP by default)
+    // Then set username and password
+    if (netConf_.hasPassword() && netConf_.hasUser()) {
+        resp = parser_.sendCommand("AT+CGDCONT=2,\"IP\",\"%s\"",
+                netConf_.hasApn() ? netConf_.apn() : "");
+        r = CHECK_PARSER(resp.readResult());
+        CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
+
+        resp = parser_.sendCommand("AT+UAUTHREQ=2,2,\"%s\",\"%s\"",
+                netConf_.hasPassword() ? netConf_.password() : "",
+                netConf_.hasUser() ? netConf_.user() : "");
+        r = CHECK_PARSER(resp.readResult());
+        CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
+    }
+
     return 0;
 }
 


avtolstoy:
> it causes the program to execute an AT+CGDCONT command which outright fails (by using the auth prefix CHAP: before the APN name)

Is there a specific error or is it just a generic `ERROR`?

> In my testing I noticed that executing these commands on PDP context 1 does authenticate properly. While using PDP context 2 does work.
Am I right in understanding that cid = 1 doesn't work and cid = 2 does?

Unfortunately, I don't think this is the right way to go about this. We do support authentication, which is done at the PPP layer using CHAP (hence `CHAP:%apnname%` when configuring the APN) and the credentials are delivered to the PPP layer here: https://github.com/particle-iot/device-os/blob/develop/hal/src/boron/network/sara_ncp_client.cpp#L988. This supposedly should work on R4 according to the AT reference manual, but at the same time, this may very well be a bug in its firmware, so we may need to use this flow but for *R4 specifically* (see `ncpId()`). If we cannot use cid = 1 and will need to use cid = 2 for APNs that require authentication, I would also pass cid to the PPP layer along with the credentials.

You also have mixed up username and password arguments: https://github.com/particle-iot/device-os/pull/1798/files#diff-b3f74fd1e22ae53aae264584d668ea1cR907

jaredwolff:
Thanks @avtolstoy. I went back and revisited it with your comments. Here's the error:

```
0000007633 [ncp.at] TRACE: > AT+CGDCONT=1,"IP","CHAP:soracom.io"
0000007673 [ncp.at] TRACE: < +CME ERROR: operation not supported
```

Without 
```
(netConf_.hasUser() && netConf_.hasPassword()) ? "CHAP:" : "",
```
it works ok. (and the extra `%s` removed from the line above)

So, these changes are definitely unnecessary. 

To understand on my end it looks like the authentication parameters are being plugged in as a PSD profile instead? It makes a bit more sense now. This is my first foray into the deviceOS.

The main question now is, setting the `CHAP:` prefix is SARA-R410 specific. Will it break any other modes or edge cases? Seems like as long as there is a user name and password it will bork on startup.

avtolstoy:
Thanks for looking into this.

> it works ok. (and the extra %s removed from the line above)

Could you please clarify what happens if you only configure the APN and leave username and password blank for you SIM card. Are you able to register on the network and establish a data connection?

The second test (if the first one doesn't cause an error for `CGDCONT`, but you are still unable to register or establish a data connection) would probably to configure the APN (without `CHAP:`) for cid = 1 and attempt to run `AT+UAUTHREQ` on it setting username, password and CHAP authentication type. If this does in fact work, we might need to add a workaround for R410 specifically to do it this way, instead of `CHAP:%apn%`.

jaredwolff:
When I was testing it would get to the check connection stage

```
0000007391 [ncp.at] TRACE: < OK
0000007392 [ncp.at] TRACE: > AT+CEREG?
0000007393 [ncp.at] TRACE: < +CEREG: 2
0000007441 [ncp.at] TRACE: < +CEREG: 2,2
```

And it would time out and try all over again.  I've just tried reproducing it here it's connecting successfully 😬. I'm not sure if the credentials are being stored on the module or SORACOM has changed something since I chatted with them about the problem. 

The only way I knew I was making progress before was because I looked at the SORACOM console and saw the device create a new connection and then promptly kill it. The firmware tries 3 times before giving up, right?

avtolstoy:
Yeah, the context configuration is persisted:
> Each context is permanently stored so that its definition is persistent over power cycles

You may need to run `AT+CGDCONT=<cid>` to clear it.

> The firmware tries 3 times before giving up, right?

3 times to do what..? We do have a 5 minute network registration timeout, after which the modem is reset and the whole registration is restarted.

jaredwolff:
Alright, I did reset both `<cid>` 1 and 2. I'm still getting a successful connection. I'm going to unplug it. Maybe it's still authenticated from the last session using the username and password. I'm not sure how they have it on the operator side. It does look like it's coming up on context 1. No extra context this time though:

```
+CGDCONT: 1,"IP","soracom.io","<IP ADDR>",0,0,0,0
```

jaredwolff:
@avtolstoy should I close this out? It may make sense considering my original suggestion is not-so-valid. I can open up a corresponding issue for the ERROR when using the `CHAP:` prefix.

I have no SIM cards that **require** username and password for the APN so I have no way to test if removing the `CHAP:` will work in all cases.

avtolstoy:
@jaredwolff Sure, let's close it. :+1: I'd appreciate if you could create an issue about `ERROR` with `CHAP:` in the APN string. Might be something SIM/network-specific, but we'll keep it in mind and try to reproduce in the future. Thank you.

PR #1804, opened 2019-06-03
Add two-digit mnc diagnostic flag

zfields:
<details>
  <summary><i>The two-digit `mnc` diagnostic allows the original `mnc` to be reconstructed accurately.</i></summary>
</details>

### Problem

Currently, the `mnc` is translated to and stored as a decimal value. According to the specification, `40` and `040` are not equal values. Currently when using the `mcc` and `mnc` tuple, there does not appear to be a clash. However, maintaining this distinction is important is important to ensure future bugs do not appear in the system.

### Solution

By sending an indicator as to the original form of the information, we are able to reconstruct the original value in a lossless manner.

### Steps to Test

user/tests/accept/hal-api-test-plan.md

### Example App

user/tests/app/cellular_global_identity/cellular_global_identity.cpp

### Tickets
- ch33501
---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/inc/cellular_hal_cellular_global_identity.h
+++ b/hal/inc/cellular_hal_cellular_global_identity.h
@@ -20,6 +20,8 @@
 
 #include <stdint.h>
 
+#include "static_assert.h"
+
 /*! Cellular Global Identity Structure Version */
 enum __attribute__((__packed__)) CgiVersion
 {
@@ -32,6 +34,8 @@ enum __attribute__((__packed__)) CgiFlags
 {
     CGI_FLAG_TWO_DIGIT_MNC = 0b00000001, /*!< Indicates two-digit format of Mobile Network Code */
 };
+
+PARTICLE_STATIC_ASSERT(CgiFlags_size, sizeof(enum CgiFlags) == 1);
 
 /*!
  * \brief Cellular Global Identity (CGI)
--- a/hal/inc/cellular_hal_cellular_global_identity.h
+++ b/hal/inc/cellular_hal_cellular_global_identity.h
@@ -23,19 +23,21 @@
 #include "static_assert.h"
 
 /*! Cellular Global Identity Structure Version */
-enum __attribute__((__packed__)) CgiVersion
+typedef enum __attribute__((__packed__)) CgiVersion
 {
     CGI_VERSION_1,
     CGI_VERSION_LATEST = CGI_VERSION_1,
-};
+} CgiVersion;
+
+PARTICLE_STATIC_ASSERT(CgiVersion_size, sizeof(CgiVersion) <= sizeof(uint16_t));
 
 /*! Cellular Global Identity Flags */
-enum __attribute__((__packed__)) CgiFlags
+typedef enum __attribute__((__packed__)) CgiFlags
 {
     CGI_FLAG_TWO_DIGIT_MNC = 0b00000001, /*!< Indicates two-digit format of Mobile Network Code */
-};
+} CgiFlags;
 
-PARTICLE_STATIC_ASSERT(CgiFlags_size, sizeof(enum CgiFlags) == 1);
+PARTICLE_STATIC_ASSERT(CgiFlags_size, sizeof(CgiFlags) <= sizeof(uint8_t));
 
 /*!
  * \brief Cellular Global Identity (CGI)
--- a/hal/inc/cellular_hal_cellular_global_identity.h
+++ b/hal/inc/cellular_hal_cellular_global_identity.h
@@ -51,8 +51,8 @@ typedef struct __attribute__((__packed__))
 {
     uint16_t size; /*!< \c size is specified by the user application, to inform the device-os of the
                       amount of space allocated to the structure */
-    uint16_t version;  /*!< \c version is specified by the device-os, to inform the user-application
-                           of the version of the information returned */
+    uint16_t version;  /*!< \c version is specified by the user application, to inform the
+                          device-os of the version of the information to be returned */
     uint8_t reserved;  /*!< \c reserved is allocated for future usage */
     uint8_t cgi_flags; /*!< \c cgi_flags Indicates the structure/configuration of the values in
                            CellularGlobalIdentity */
--- a/hal/inc/cellular_hal_cellular_global_identity.h
+++ b/hal/inc/cellular_hal_cellular_global_identity.h
@@ -20,6 +20,19 @@
 
 #include <stdint.h>
 
+/*! Cellular Global Identity Structure Version */
+enum __attribute__((__packed__)) CgiVersion
+{
+    CGI_VERSION_1,
+    CGI_VERSION_LATEST = CGI_VERSION_1,
+};
+
+/*! Cellular Global Identity Flags */
+enum __attribute__((__packed__)) CgiFlags
+{
+    CGI_FLAG_TWO_DIGIT_MNC = 0b00000001, /*!< Indicates two-digit format of Mobile Network Code */
+};
+
 /*!
  * \brief Cellular Global Identity (CGI)
  *
@@ -35,8 +48,10 @@ typedef struct __attribute__((__packed__))
     uint16_t size; /*!< \c size is specified by the user application, to inform the device-os of the
                       amount of space allocated to the structure */
     uint16_t version;  /*!< \c version is specified by the device-os, to inform the user-application
-                          of the version of the information returned */
-    uint16_t reserved; /*!< \c reserved is allocated for future usage */
+                           of the version of the information returned */
+    uint8_t reserved;  /*!< \c reserved is allocated for future usage */
+    uint8_t cgi_flags; /*!< \c cgi_flags Indicates the structure/configuration of the values in
+                           CellularGlobalIdentity */
     uint16_t mobile_country_code; /*!< \c mobile_country_code consists of three decimal digits and
                                      identifies uniquely the country of domicile of the mobile
                                      subscription */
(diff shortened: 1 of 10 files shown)


hal/inc/cellular_hal_cellular_global_identity.h
```diff
@@ -34,9 +34,12 @@ typedef struct __attribute__((__packed__))
 {
     uint16_t size; /*!< \c size is specified by the user application, to inform the device-os of the
                       amount of space allocated to the structure */
-    uint16_t version;  /*!< \c version is specified by the device-os, to inform the user-application
-                          of the version of the information returned */
-    uint16_t reserved; /*!< \c reserved is allocated for future usage */
+    uint16_t version; /*!< \c version is specified by the device-os, to inform the user-application
+                         of the version of the information returned */
+    uint8_t reserved; /*!< \c reserved is allocated for future usage */
+    uint8_t
```
avtolstoy: Can we make it a flag instead in order not to waste a whole `uint8_t` for just one boolean?

system/src/system_network_diagnostics.cpp
```diff
@@ -308,6 +308,27 @@ class NetworkCellularCellGlobalIdentityMobileNetworkCodeDiagnosticData
...
+              DIAG_ID_NETWORK_CELLULAR_CELL_GLOBAL_IDENTITY_TWO_DIGIT_MOBILE_NETWORK_CODE,
+              DIAG_NAME_NETWORK_CELLULAR_CELL_GLOBAL_IDENTITY_TWO_DIGIT_MOBILE_NETWORK_CODE)
+    {
+    }
+
+    virtual int get(IntType& val)
+    {
+        CellularGlobalIdentity cgi;
+        cellular_global_identity(&cgi, nullptr);
+        val = static_cast<IntType>(cgi.two_digit_mobile_network_code);
```
avtolstoy: Instead of introducing a new diagnostic type to just indicate whether the code is 2 or 3 digits, we could potentially add some kind of a flag to the current reported MNC value. Perhaps offset 2-digit values by something like a `10000` or set the most significant bit? cc @m-mcgowan

avtolstoy: When we initially discussed this, I thought all these are numbers, so to simplify development I suggested using current `IntType`. Perhaps we should revisit that decision, and this could potentially mean that we do need a string diagnostic type, so perhaps we should introduce that instead. :man_shrugging:

hal/inc/cellular_hal_cellular_global_identity.h
zfields: I had considered this myself, but we have the `reserved` space, so I thought I would use it for simplicity; I'm very open to change.

zfields: done

system/src/system_network_diagnostics.cpp
zfields: I believe we arrived at a successful compromise during scrum.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -412,6 +412,15 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
     r = CHECK_PARSER(resp.readResult());
     CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
 
+    // Preserve digit format data
+    const int mnc_digits = ::strnlen(mobileNetworkCode, 4);
```
avtolstoy: Magic number, let's use `sizeof(mobileNetworkCode)` here.

Same here https://github.com/particle-iot/device-os/pull/1804/files#diff-13c08dca4df073df121b6a68266e0105R1520

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -412,6 +412,15 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
...
+    CHECK_TRUE((2 == mnc_digits || 3 == mnc_digits), SYSTEM_ERROR_BAD_DATA);
+    if (2 == mnc_digits) {
+        cgi_.cgi_flags |= CGI_FLAG_TWO_DIGIT_MNC;
+    } else {
+        cgi_.cgi_flags &= ~CGI_FLAG_TWO_DIGIT_MNC;
```
avtolstoy: Let's zero out `cgi_.cgi_flags`, in which case this becomes unnecessary.

Same here https://github.com/particle-iot/device-os/pull/1804/files#diff-13c08dca4df073df121b6a68266e0105R1524

system/src/system_network_diagnostics.cpp
```diff
@@ -302,7 +302,14 @@ class NetworkCellularCellGlobalIdentityMobileNetworkCodeDiagnosticData
     {
         CellularGlobalIdentity cgi;
```
avtolstoy: I've just noticed this. Why aren't we zeroing out this struct and not setting neither size nor version?

hal/src/boron/network/sara_ncp_client.cpp
zfields: Yeah, we talked about it and I forgot... :+1:

hal/src/boron/network/sara_ncp_client.cpp
zfields: :+1:

system/src/system_network_diagnostics.cpp
zfields: Great catch! Thank you!

hal/src/boron/network/sara_ncp_client.cpp
zfields: Correct me if I'm wrong, but I believe the struct _(and subsequently the `cgi_flags`)_ were already zero'd at [mdm_hal.cpp|Line 1212](https://github.com/particle-iot/device-os/blob/mnc2d/hal/src/electron/modem/mdm_hal.cpp#L1212) for Gen2 devices, and is initialized because it is a non-static data member of `SaraNcpClient` at [sara_ncp_client.h|Line 80](https://github.com/particle-iot/device-os/blob/mnc2d/hal/src/boron/network/sara_ncp_client.h#L80) for Gen3 devices.

I'm less confident for the Gen3 devices, but that was my original understanding...

zfields: I'm going to explicitly zero the struct on the `SaraNcpClient` to ensure it is behaving as expected.

avtolstoy:
@zfields Please also set the appropriate milestone for this PR.

system/src/system_network_diagnostics.cpp
```diff
@@ -321,7 +328,7 @@ class NetworkCellularCellGlobalIdentityLocationAreaCodeDiagnosticData
 
     virtual int get(IntType& val)
     {
-        CellularGlobalIdentity cgi;
+        CellularGlobalIdentity cgi{sizeof(CellularGlobalIdentity),CGI_VERSION_LATEST};
```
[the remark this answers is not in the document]
zfields: I actually prefer those, but I thought that was only a `C` thing. I will switch over.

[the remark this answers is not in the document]
avtolstoy: AFAIK it's in C++20 officially, but GCC supported trivial cases since long ago.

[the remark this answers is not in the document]
zfields: I just did a cursory check and I cannot find where designated initialization has made it into the c++ standard.
@avtolstoy Can you please confirm when it was added to the standard and if our compile flags `-std=c++14` support it?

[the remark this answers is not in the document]
avtolstoy: See message above. Yes, it'll work as long as the designators are specified in the same order as the fields during the declaration.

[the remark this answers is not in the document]
zfields: I know it will work, but I'm more curious as to whether this code is inline with the standard we are compiling our code against. While I truly like the clarity of the syntax, I would prefer to maintain standard compliance and not become compiler dependent.

hal/inc/cellular_hal_cellular_global_identity.h
```diff
@@ -33,6 +35,8 @@ enum __attribute__((__packed__)) CgiFlags
     CGI_FLAG_TWO_DIGIT_MNC = 0b00000001, /*!< Indicates two-digit format of Mobile Network Code */
 };
 
+PARTICLE_STATIC_ASSERT(CgiFlags_size, sizeof(enum CgiFlags) == 1);
```
avtolstoy: Prefer `typedef enum XXX {} XXX`

system/src/system_network_diagnostics.cpp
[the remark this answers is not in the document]
avtolstoy: In that case you should consider removing `__attribute__` or any other GNU Extensions as well.

I simply want to improve readability, so you could also consider:
```suggestion
        CellularGlobalIdentity cgi = {};
        cgi.size = sizeof(cgi);
        cgi.version = CGI_VERSION_LATEST;
```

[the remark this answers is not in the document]
zfields: You make a good point about `__attribute__`, seems like we are well committed to GNU at this point. I will follow your first suggestion, and use the designated initializers so we can keep it on one line.

hal/inc/cellular_hal_cellular_global_identity.h
zfields: done

PR #1806, opened 2019-06-05
Improve compatibility with recent versions of GCC (module version should bump to 1320)

sergeuz:
### Ideally we would merge this when we feel 1.3.0 is being finalized (maybe the first RC?)

---

### Problem

This PR is a continuation of the work started in https://github.com/particle-iot/device-os/pull/1742:

- Refactor linker files to eliminate duplication of RAM memory map constants in different files (only for modular builds for now).
- Move the system-part1 SRAM region to another location on Gen 2 platforms (see https://github.com/particle-iot/device-os/pull/1742).
- Increase the size of the system-part1 SRAM from 768 to 1024 bytes on Photon. This is enough to store the statically allocated locale data that was added in newlib bundled with ARM GCC >= 5.4.1.
- Use `os_thread_t` instead of `std::thread` in the active object classes. This decouples our code from the libstdc++ internal API which changes between GCC releases.
- Fix various compiler warnings and errors to make the code buildable with GCC 6-8. Note that correct work with recent versions of GCC is not guaranteed, the default compiler version is still 5.3.1.

### Steps to Test

1. Run `wiring/no_fixture` with enabled threading.
2. Flash Device OS 1.2.1-rc.2 to a Gen 2 device.
3. Bump the module versions to 1320 and flash this branch to the device OTA.
4. Downgrade to 1.2.1-rc.2 OTA.


build/arm/linker/linker_nrf52840_bootloader.ld
```diff
@@ -7,17 +7,7 @@ __Stack_Init = _estack - __Stack_Size;
...
-/* bootloader_ram_length = 16K; */
-/* bootloader_ram_offset = 8K; */
-/* bootloader_ram_origin = user_module_sram_origin + bootloader_ram_offset; */
-/* ASSERT(bootloader_ram_offset + bootloader_ram_length <= user_module_sram_length, */
-/*     "Insufficient space for RAM region"); */
-
+INCLUDE platform_ram.ld
 INCLUDE softdevice.ld
 
 bootloader_ram_origin = 0x20000000 + APP_RAM_BASE;
```
avtolstoy: Should we define this in `platform_ram.ld` ?

wiring/src/spark_wiring_string.cpp
```diff
@@ -729,9 +729,9 @@ String& String::remove(unsigned int index, unsigned int count){
 	if (index + count > len) { count = len - index; }
 	char *writeTo = buffer + index;
 	len = len - count;
-	strncpy(writeTo, buffer + index + count,len - index);
+	memmove(writeTo, buffer + index + count,len - index);
```
avtolstoy: :+1:

newlib_nano/src/malloc.cpp
```diff
@@ -31,11 +37,13 @@ void* _malloc_r(struct _reent *r, size_t s) {
 
 void _free_r(struct _reent* r, void* ptr) {
     panic_if_in_isr();
+#if __GLIBCXX__ < 20160919 // ARM GCC 5.4.1-2016q3
```
avtolstoy: :+1:

build/arm/linker/stm32f2xx/electron/platform_ram.ld
```diff
@@ -0,0 +1,76 @@
...
+| System part 1 static RAM |
++--------------------------+ 0x2001cc00
+| System part 3 static RAM |
++--------------------------+ 0x2001bc00
+|                          |
+| Heap                     |
+|                          |
++~~~~~~~~~~~~~~~~~~~~~~~~~~+
+| User part static RAM     |
++--------------------------+ 0x20000000
```
avtolstoy: :heart:

sergeuz:
> It would be nice to include RAM / flash space difference between gcc 5.3.1 and latest in this PR just for reference.

The difference is somewhat weird:

Electron, GCC 5.3.1:
```
  74852	    556	   2584	  77992	  130a8	../../../build/target/system-part1/platform-10-m/system-part1.elf
 125080	    212	    564	 125856	  1eba0	../../../build/target/system-part2/platform-10-m/system-part2.elf
 119708	   2696	   4328	 126732	  1ef0c	../../../build/target/system-part3/platform-10-m/system-part3.elf
  11092	    156	    432	  11680	   2da0	../../../build/target/user-part/platform-10-m/tinker.elf
```

GCC 8.2.1:
```
  78580	    896	   2788	  82264	  14158	../../../build/target/system-part1/platform-10-m/system-part1.elf
 122664	    208	    524	 123396	  1e204	../../../build/target/system-part2/platform-10-m/system-part2.elf
 123980	   2700	   5848	 132528	  205b0	../../../build/target/system-part3/platform-10-m/system-part3.elf
  15348	    160	    448	  15956	   3e54	../../../build/target/user-part/platform-10-m/tinker.elf
```

As you can see, .text sections of the user part and system part 2 and 3 grew significantly, system part 1 .text became smaller, some .bss sections grew significantly as well. Now when we have a technical possibility to build the firmware with GCC 8 we can investigate the difference further.

build/arm/linker/linker_nrf52840_bootloader.ld
sergeuz: This PR addresses only modular builds of the main firmware to narrow the scope a little bit and keep the changes focused on compatibility with recent versions of GCC. I'd be happy to refactor linker files for the bootloader and monolithic builds as well (and also define flash memory maps in a similar manner), but not in the scope of this PR

wiring/src/spark_wiring_string.cpp
sergeuz: All kudos to GCC 8!

sergeuz:
[ch34561]

PR #1810, opened 2019-06-07
Cache cellular diagnostics

zfields:
### Problem

Current implementation requires multiple calls to `AT+COPS` and could result in an inconsistent cellular global identity.

### Solution

By caching the diagnostic results we only check every 1 second.

### Steps to Test

Watch the modem logs.

### References

- [[CH34335]](https://app.clubhouse.io/particle/story/34335/optimize-diagnostic-message-broadcast-for-cellular-network-traffic)
---

### Completeness

- [x] Added documentation

### DIFF
--- a/system/src/system_network_diagnostics.cpp
+++ b/system/src/system_network_diagnostics.cpp
@@ -56,16 +56,16 @@ namespace
 using namespace particle;
 using namespace spark;
 
-class SignalCache
+class NetworkCache
 {
 public:
-    static const system_tick_t SIGNAL_INFO_CACHE_INTERVAL = 1000;
+    static const system_tick_t NETWORK_INFO_CACHE_INTERVAL = 1000;
 
     const Signal* getSignal()
     {
 #if Wiring_WiFi || Wiring_Cellular
         system_tick_t m = millis();
-        if (ts_ == 0 || (m - ts_) >= SIGNAL_INFO_CACHE_INTERVAL)
+        if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL)
         {
 #if Wiring_WiFi
             sig_ = WiFi.RSSI();
@@ -80,9 +80,28 @@ public:
 #endif
     }
 
+#if HAL_PLATFORM_CELLULAR
+    const CellularGlobalIdentity* getCGI()
+    {
+        system_tick_t m = millis();
+        if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL)
+        {
+            cgi_.size = sizeof(cgi_);
+            cgi_.version = CGI_VERSION_LATEST;
+            cellular_global_identity(&cgi_, nullptr);
+            ts_ = millis();
+        }
+        return &cgi_;
+    }
+#endif
+
 private:
 #if Wiring_Cellular
     CellularSignal sig_;
+#if HAL_PLATFORM_CELLULAR
+    CellularGlobalIdentity cgi_{.size = sizeof(CellularGlobalIdentity),
+                                .version = CGI_VERSION_LATEST};
+#endif
 #elif Wiring_WiFi
     WiFiSignal sig_;
 #elif Wiring_Mesh
@@ -91,7 +110,7 @@ private:
     system_tick_t ts_ = 0;
 };
 
-static SignalCache s_signalCache;
+static NetworkCache s_networkCache;
 
 class SignalStrengthDiagnosticData : public AbstractIntegerDiagnosticData
 {
@@ -104,7 +123,7 @@ public:
 
     virtual int get(IntType& val)
     {
-        const Signal* sig = s_signalCache.getSignal();
+        const Signal* sig = s_networkCache.getSignal();
         if (sig == nullptr)
         {
             return SYSTEM_ERROR_NOT_SUPPORTED;
@@ -133,7 +152,7 @@ public:
 
     virtual int get(IntType& val)
     {
-        const Signal* sig = s_signalCache.getSignal();
+        const Signal* sig = s_networkCache.getSignal();
         if (sig == nullptr)
         {
             return SYSTEM_ERROR_NOT_SUPPORTED;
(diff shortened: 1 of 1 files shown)


system/src/system_network_diagnostics.cpp
```diff
@@ -80,9 +80,28 @@ class SignalCache
 #endif
     }
 
+#if HAL_PLATFORM_CELLULAR
```
technobly: I’m not entirely sure this will catch Gen 3 devices. Check the cellular no fixture tests.

system/src/system_network_diagnostics.cpp
```diff
@@ -80,9 +80,28 @@ class SignalCache
...
+    {
+        system_tick_t m = millis();
+        if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL)
+        {
+            cgi_.size = sizeof(cgi_);
+            cgi_.version = CGI_VERSION_LATEST;
+            cellular_global_identity(&cgi_, nullptr);
+            ts_ = millis();
+        }
+        return &cgi_;
```
technobly: I see this only updates every 1 second now, but couldn’t there still be a potential for a non-atomic update if MNC is just before 1 second and MCC is just after? I think based on how quickly these 4 parameters are updated it might even be easier to make the data non-atomic now since it would have forced the calling system to wait 1 full second between two reads of the same data set. 

Can we just force the original caller of the 4 get() functions to only be one call tot the new getCGI() function and remove the 1 second update mechanism?

system/src/system_network_diagnostics.cpp
zfields: `HAL_PLATFORM_CELLULAR` is working below to send the messages for Gen3 devices.

system/src/system_network_diagnostics.cpp
zfields: The result of an offline conversation _(scrum)_ was, as a team, we are pleased with the atomicity of the data in the `CellularGlobalIdentity` struct returned from the HAL layer. Building on that, we are confident that the four data points `mcc`, `mnc`, `lac` and `ci` will remain coherent, because all of the calls and caching in this PR are happening above the HAL layer and the time period begins at the first message invocation.

We did discuss the notion of a marker that could be used to indicate when a diagnostic group started emitting it's messages and when it was complete. This change could work for both signal strength and quality as well as cgi data. However, it was decided this change was beyond the scope of this PR, and would be best placed into the backlog.

m-mcgowan: The key feature is that all the CGI diagnostics must be retrieved within 1 second of each other, which should be the case since they are registered within the same file.

system/src/system_network_diagnostics.cpp
```diff
@@ -80,9 +80,28 @@ class SignalCache
...
 
+#if HAL_PLATFORM_CELLULAR
+    const CellularGlobalIdentity* getCGI()
+    {
+        system_tick_t m = millis();
+        if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL)
+        {
+            cgi_.size = sizeof(cgi_);
+            cgi_.version = CGI_VERSION_LATEST;
+            cellular_global_identity(&cgi_, nullptr);
```
avtolstoy: Error code is not being checked. `WiFi/Cellular.RSSI()` have special values that indicate an error, here we need to specifically handle errors returned by `cellular_global_identity()` and somehow forward them to the diagnostic sources.

zfields: Updated.

technobly: So am I correct in assuming this is the macro that will be used?  Why is it marked `Deprecated` and in the case of error will it just return `nullptr`?
```
// Deprecated
#define CHECK_RETURN(_expr, _val) \
        ({ \
            const auto _ret = _expr; \
            if (_ret < 0) { \
                _LOG_CHECKED_ERROR(_expr, _ret); \
                return _val; \
            } \
            _ret; \
        })
```

sergeuz: I marked `CHECK_RETURN()`  as deprecated for a couple of reasons:

- It loses the error code info and thus breaks the convention. If some module needs to perform a conversion between system- and module-specific result codes/return values then it should implement its own local conversion function and/or a helper macro.
- Having a lot of `CHECK_SOMETHING()` macros hits the readability of the code in the long run. It's better to avoid introducing new ones in the global namespace unless absolutely necessary.

With that said I won't mind having globally defined `CHECK_RETURN()` and `CHECK_GOTO()`, but they should be implemented in a way that preserves the error code info.

system/src/system_network_diagnostics.cpp
```diff
@@ -298,7 +303,8 @@ class NetworkCellularCellGlobalIdentityMobileCountryCodeDiagnosticData
 
     virtual int get(IntType& val)
     {
-        const CellularGlobalIdentity* cgi = s_networkCache.getCGI();
+        const CellularGlobalIdentity* const cgi = s_networkCache.getCGI();
+        CHECK_TRUE(cgi, SYSTEM_ERROR_NETWORK);
```
technobly: Instead of returning CellularGlobalIdentity, how about passing it into getCGI by reference and returning the system error code directly.  There are a lot of specific errors that could be bubbled up instead of SYSTEM_ERROR_NETWORK, and this is kind of the same way cellular_global_identity() works already.

system/src/system_network_diagnostics.cpp
```diff
@@ -279,11 +305,13 @@ class NetworkCellularCellGlobalIdentityMobileCountryCodeDiagnosticData
 
     virtual int get(IntType& val)
     {
-        CellularGlobalIdentity cgi{.size = sizeof(CellularGlobalIdentity), .version = CGI_VERSION_LATEST};
-        cellular_global_identity(&cgi, nullptr);
+        cellular_result_t result;
+        CellularGlobalIdentity cgi = {.size = sizeof(CellularGlobalIdentity),
```
avtolstoy: I'd prefer this was filled in `getCGI()`.

zfields: done

PR #1816, opened 2019-06-10
[Gen 3] SoftDevice update support

avtolstoy:
### Problem

https://app.clubhouse.io/particle/story/31496/nordic-softdevice-upgrade

Currently, the only way to upgrade the SoftDevice is by using the debugger and manually flashing the SoftDevice binary into the appropriate location.

We've specifically decoupled our Gen 3 bootloader from SoftDevice, so it should be possible for us to employ current OTA update mechanics to update the SoftDevice, where the module is first copied into the OTA region by the DeviceOS, the device is reset and then the bootloader copies the module from the OTA region to the appropriate location in the internal flash. The only caveat is that we'll need to strip off the header/footer.

It should ideally also be possible to flash the SoftDevice over DFU as a means of recovery.

### Solution

This PR:
- adds a new module function definition `MODULE_FUNCTION_RADIO_STACK`
- adds `module_radio_stack` module on Gen 3 devices
- adds `platform_radio_stack_xxx` functions to retrieve version info of the radio stack module and update it
- **changes a reserved field in `module_info_t` into `flags` of type `module_info_flags_t` and introduces a new module flag `MODULE_INFO_FLAG_DROP_MODULE_INFO` which indicates that module info needs to be stripped when this module is flashed into its location**
- adds a simple python script to generate modules out of raw binaries/hex files

sidenote: https://github.com/particle-iot/device-os/compare/feature/softdevice-update?expand=1#diff-3c500a02247f461f914ecdf1380550ebR29

#### Update

1. Generate SoftDevice radio stack module for the platform of interest e.g.:
```console
$ python3 build/create_module.py --platform argon --function radio_stack ../third_party/nrf5_sdk/nrf5_sdk/components/softdevice/s140/hex/s140_nrf52_6.1.1_softdevice.hex argon-softdevice-6.1.1.bin
```
2. By default this module will depend on system-part1 version 1301 and bootloader version 312, so manually bump them and build both out of this branch and flash onto the device
3. Put the device into listening mode and retrieve the module info with `particle serial inspect` (CLI 1.43.2 required). Make sure that there is a radio stack module with version `169` (6.0.0).
4. Flash the SoftDevice module OTA or using serial e.g.:
```console
$ particle flash --serial argon-softdevice-6.1.1.bin
$ particle flash %deviceid% argon-softdevice-6.1.1.bin
```
4. Put the device into listening mode, retrieve module info and make sure that the SoftDevice has been upgraded: the version should change to `182` (6.1.1)

#### MBR read-only

DFU writes into address range `0x0000 - 0x1000` should not be possible.

#### Recovery

The device should still able to boot into the bootloader even if SoftDevice upgrade is interrupted in the middle of the copying process from OTA region into the correct location.

1. Write some garbage data into the SoftDevice area (`0x1000 - 0x30000`)
2. The device will crash on boot
3. Attempt to enter DFU should succeed

#### SoftDevice DFU upgrade

It should be possible to upgrade SoftDevice using DFU. Particle CLI supports `MODULE_INFO_FLAG_DROP_MODULE_INFO` flag since 1.43.2.

2. Flash the SoftDevice e.g.:
```console
$ particle flash --usb argon-softdevice-6.1.1.bin
```
3. The device should function correctly

### Release plan

Introduction of this feature will require two release.

The first release will only include this PR, but will not trigger the SoftDevice upgrade using our standard dependency mechanism. This will be the mandatory release and its system-part1 and bootloader versions will be used as dependencies in generated SoftDevice binaries.

The second release will add a dependency to `system-part1` on a new SoftDevice version, which will trigger the upgrade. For example, a simplified upgrade path from 0.9.0 to 1.4.0 via a mandatory 1.3.0 will look like this:

1. 1.4.0 user-part
2. 1.3.0 bootloader
3. 1.3.0 system-part1
4. 6.1.1 SoftDevice
5. 1.4.0 system-part1

### Example App

N/A

### References

- [CH31496]
- https://github.com/particle-iot/firmware-protobuf/pull/5
- https://github.com/particle-iot/binary-version-reader/pull/24/files
- https://github.com/particle-iot/particle-cli/pull/493

### TODO

- [x] PRs updating `binary-version-reader` and `particle-cli`.
- [x] Update submodule reference once https://github.com/particle-iot/firmware-protobuf/pull/5 is merged
- [x] Update openthread submodule reference once SoftDevice 7.0.0 support is backported

---

### Completeness

- [feature] [Gen 3] Nordic SoftDevice update support


m-mcgowan:
Rather than call this a radio module, perhaps we generalize it as a 3rd party module? Future hardware may have different 3rd party modules not specifically focused on the radio.

avtolstoy:
@m-mcgowan I have no strong preferences towards the naming, but honestly I prefer to call things what they really are. This naming was proposed in https://app.clubhouse.io/particle/story/31496/nordic-softdevice-upgrade and I didn't hear any negative feedback towards it. This naming also somewhat coincides with what we have on Photons and P1s for example, where there is a a proprietary WiFi firmware blob, which could also be considered a 'radio stack` module.

hal/src/nRF52840/platform_radio_stack.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "ota_flash_hal_impl.h"
+#include "flash_mal.h"
+#include "check.h"
+#include "nrf_sdm.h"
+#include "nrf_mbr.h"
+
+int platform_radio_stack_fetch_module_info(hal_system_info_t* sys_info, bool create) {
+    for (int i = 0; i < sys_info->module_count; i++) {
+        hal_module_t* module = sys_info->modules + i;
+        if (memcmp(&module->bounds, &module_radio_stack, sizeof(module_radio_stack))) {
```
m-mcgowan: Doesn't this need to check all the module bounds before deciding to create? At it is it checks just the first.

hal/src/nRF52840/platform_radio_stack.cpp
```diff
@@ -0,0 +1,71 @@
...
+            continue;
+        }
+
+        if (create) {
```
m-mcgowan: Looks like this block needs to be moved out and a flag set if the module was previously found.

platform/MCU/nRF52840/src/flash_mal.c
```diff
@@ -190,6 +190,17 @@ int FLASH_CopyMemory(flash_device_t sourceDeviceID, uint32_t sourceAddress,
     uint8_t data_buf[MAX_COPY_LENGTH];
     uint32_t copy_len = 0;
 
+    // We may only check the module header if we've been asked to verify it
+    if (flags & MODULE_VERIFY_MASK)
+    {
+        const module_info_t* info = FLASH_ModuleInfo(sourceDeviceID, sourceAddress);
+        if (info->flags & MODULE_INFO_FLAG_DROP_MODULE_INFO && (uintptr_t)info == (uintptr_t)sourceAddress)
```
m-mcgowan: why do we compare the result of the structure with the source address?  Won't these be different with OTA'd modules?

system/src/system_update.cpp
```diff
@@ -540,6 +540,7 @@ const char* module_function_string(module_function_t func) {
         case MODULE_FUNCTION_SYSTEM_PART: return "s";
         case MODULE_FUNCTION_USER_PART: return "u";
         case MODULE_FUNCTION_NCP_FIRMWARE: return "c";
+        case MODULE_FUNCTION_RADIO_STACK: return "a";
```
m-mcgowan: A similar change will be needed in https://github.com/particle-iot/binary-version-reader

platform/MCU/nRF52840/src/flash_mal.c
avtolstoy: Because we have corner cases where the module info is not located in the front of the module but for example after the vector table and we only want to enable this feature in the case it it in the front, hence the module_info_t located at the start of the source address check.

hal/inc/hal_platform.h
```diff
@@ -115,8 +115,10 @@
 #define HAL_PLATFORM_NCP_UPDATABLE (0)
 #endif /* HAL_PLATFORM_NCP_UPDATABLE */
 
+// FIXME: why are these here? Are they supposed to be overridable by platform?
```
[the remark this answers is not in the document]
avtolstoy: I will wrap these in `ifndef` checks then.

system/src/system_update.cpp
avtolstoy: https://github.com/particle-iot/binary-version-reader/pull/24
https://github.com/particle-iot/particle-cli/pull/493

hal/src/nRF52840/platform_radio_stack.cpp
avtolstoy: I'm not sure I get this. Will we get multiple modules in `sys_info` with exactly the same module bounds that match `module_radio_stack`?

hal/src/nRF52840/platform_radio_stack.cpp
avtolstoy: I'm not sure I get this either, this code blocks is executed only if `the module->bounds` matches the `module_radio_stack`.

platform/MCU/nRF52840/src/flash_mal.c
technobly: This sounds like a really good comment to add here in the code.

hal/src/nRF52840/platform_radio_stack.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+            info->module_version = SD_FWID_GET(MBR_SIZE);
+            info->platform_id = PLATFORM_ID;
+            info->module_function = MODULE_FUNCTION_RADIO_STACK;
+
+            module->info = info;
+            // FIXME: assuming that all checks passed for now
+            module->validity_result = module->validity_checked;
+        } else {
+            delete module->info;
```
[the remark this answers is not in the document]
avtolstoy: Yeap, that's the standard behavior for these functions. Same can be seen in `platform_ncp_fetch_module_info()`.

platform/MCU/nRF52840/src/flash_mal.c
avtolstoy: Added in https://github.com/particle-iot/device-os/pull/1816/commits/cb5448b7c16aaca3c11b5da81b6dc03d086a18b5

PR #1825, opened 2019-06-14
Particle.connected() should return true only after handshake messages are acknowledged

sergeuz:
### Problem

`Particle.connected()` returns `true` immediately after the session is resumed, even if the session data became invalid for some reason and the device will in fact never receive ACKs for its handshake messages until the next connection attempt.

### Solution

1. Wait for all system- and protocol-level handshake messages to be acknowledged before changing the cloud connection state.
2. Make `Particle.publish()` and other cloud functions in the Wiring API fail early if the device is not connected to the cloud.

### Steps to Test

1. Run the communication unit tests.
2. Flash the example app and verify via the logs that `Particle.connected()` changes its return value only after all expected ACKs are received from the server.

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(MANUAL)

namespace {

const SerialLogHandler logHandler(LOG_LEVEL_INFO);

void cloudStatusChanged(system_event_t event, int param) {
    if (param == cloud_status_connected) {
        Log.info("cloud_status_connected");
    }
}

bool lastConnected = true;

} // namespace

void setup() {
    System.on(cloud_status, cloudStatusChanged);
    Particle.connect();
}

void loop() {
    const bool connected = Particle.connected();
    if (connected != lastConnected) {
        Log.info("Particle.connected(): %d", (int)connected);
        lastConnected = connected;
    }
    Particle.process();
}
```

### References

- [ch32340]

### DIFF
--- a/system/src/system_update.cpp
+++ b/system/src/system_update.cpp
@@ -56,7 +56,7 @@ static uint32_t start_ymodem_flasher_serial_speed = START_YMODEM_FLASHER_SERIAL_
 
 ymodem_serial_flash_update_handler Ymodem_Serial_Flash_Update_Handler = NULL;
 
-// TODO: Replace SPARK_CLOUD_XXX flags with a single state variable
+// TODO: Use a single state variable instead of SPARK_CLOUD_XXX flags
 volatile uint8_t SPARK_CLOUD_SOCKETED;
 volatile uint8_t SPARK_CLOUD_CONNECTED;
 volatile uint8_t SPARK_CLOUD_HANDSHAKE_PENDING = 0;
--- a/system/src/system_update.cpp
+++ b/system/src/system_update.cpp
@@ -56,8 +56,11 @@ static uint32_t start_ymodem_flasher_serial_speed = START_YMODEM_FLASHER_SERIAL_
 
 ymodem_serial_flash_update_handler Ymodem_Serial_Flash_Update_Handler = NULL;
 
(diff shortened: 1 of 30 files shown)


sergeuz:
@m-mcgowan I'd like to merge this PR, could you please take a look at it?

system/src/system_update.cpp
```diff
@@ -56,8 +56,11 @@ static uint32_t start_ymodem_flasher_serial_speed = START_YMODEM_FLASHER_SERIAL_
 
 ymodem_serial_flash_update_handler Ymodem_Serial_Flash_Update_Handler = NULL;
 
+// TODO: Use a single state variable instead of SPARK_CLOUD_XXX flags
```
m-mcgowan: <3

sergeuz:
> I'd like to draw up a more comprehensive test plan prior to release.

What specifically do you think needs to be tested in addition to what is already covered by the unit tests and test app in this PR?

m-mcgowan:
> > I'd like to draw up a more comprehensive test plan prior to release.
> 
> What specifically do you think needs to be tested in addition to what is already covered by the unit tests and test app in this PR?

I'd like to test the edge cases with end to end tests to be sure the integration works as expected. E.g. what happens when none of the system messages are delivered? or just some of them?  I know we expect a timeout, so we should have tests to validate that. 

Are we certain that a race condition doesn't occur where the response is received so quickly that we empty the message queue and consider the handshake complete when it's not?

PR #1824, opened 2019-06-13
[lte] cellular fixes

technobly:
### Problem

There are several issues currently:
- u-blox SARA-R410M-02B firmware 05.06,A.02.00 has a memory housekeeping issue where after every 124 days it does a cleanup.  If the power is forced off while that cleanup is in progress, the modem can enter an unrecoverable state.  u-blox has advised on proper timeouts from power up and registration to mitigate this issue.  Only this particular version of u-blox firmware has this issue.
- u-blox SARA-R410M-02B firmware 05.08,A.02.04 changes the way some line endings are generated for multiple commands.  This breaks the AT command parser on Gen 2 devices.  Gen 3 is unaffected based on how it removes all line endings before parsing commands.
- u-blox SARA-R410M-02B firmware 05.08,A.02.04 requires the AT+UMNOPROF default setting to be changed to a specific carrier, or SIM Select.  We prefer SIM Select as it should allow 3rd party SIMs to have the correct settings applied.

### Solution

This PR solves the above issues by implementing a commit for each one:
- [gen 2] mitigates memory issue in u-blox firmware 05.06,A.02.00 [ch33464]
- [gen 2] fixes AT command parser issues with u-blox firmware 05.08,A.02.04 [ch32049]
- [gen 2] adds AT+UMNOPROF=1 (SIM Select) setting [ch29063]
- [gen 2] adds debugging for Pipe buffer with MDM_DEBUG_RX_PIPE & MDM_DEBUG_TX_PIPE defines
- [gen 3] mitigates memory issue in u-blox firmware 05.06,A.02.00 [ch33464]
- [gen 3] adds AT+UMNOPROF=1 (SIM Select) setting [ch29063]

### Steps to Test

- Run unit tests
- Run TEST=wiring/no_fixture
- Lots of manual testing on U260, G350, R410 modems with new and old u-blox firmware

### References

Closes [ch33464]
Closes [ch32049]
Closes [ch29063]

---

### Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [x] Problem and Solution clearly stated
- [x] Run unit/integration/application tests on device
- [x] (N/A) Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

PR #1788, opened 2019-05-22
allow the bootloader to be flashed over DFU

m-mcgowan:
### Problem

When preparing a new firmware release for Gen2 and Gen3 platforms that don't contain the bootloader, we currently prepare an Ascender app that allows updating the bootloader of the device. This causes unnecessary delay in the release process.

### Solution

This PR allows the bootloader (and other DeviceOS images) to be flashed via DFU.

The image is flashed to the OTA update region, and a flag set in the DCT.  When the system reboots, the image is applied just as if it were sent OTA or OTW.  This enables the bootloader to be flashed via DFU (such as by the CLI.)

### Steps to Test

 1. Flash the Device OS from this PR to your target device
 1.  Use `particle serial inspect` to retrieve the bootloader version of the current bootloader
 1. edit modules/shared/system_module_version.mk, and change `BOOTLOADER_VERSION` to higher than the bootloader version in the previous step.  Build the bootloader for your device.
 1. `echo -e "\xA3" > A3`
 1. `echo -e "\xA5" > A5`
 1.  flash the bootloader to your device via dfu using the address of the OTA region. E.g. 0x80C0000 on Gen2 platforms. `dfu-util -d 2b04:d006 -a 0 -s 0x80C0000 -D bootloader.bin`
 1. Restart the device
 1. Use `particle serial inspect` and verify that the bootloader version has NOT changed
 1. `dfu-util -d 2b04:d006 -a 1 -s 1753:leave -D A3` . (deliberately the incorrect flag value.)
 1. Use `particle serial inspect` and verify that the bootloader version has NOT changed
 1. `dfu-util -d 2b04:d006 -a 1 -s 1753:leave -D A5` . (correct flag value.)
 1. Use `particle serial inspect` and verify that the bootloader version has changed to match `BOOTLOADER_VERSION`.
 1. Verify that the OTA flag has been cleared:
  1. `dfu-util -d 2b04:d006 -a 1 -s 1753:1 -U ota_flag`
  1. xxd ota_flag
```
00000000: ff                         
```
### References

- [[CH32507]](https://app.clubhouse.io/particle/story/32507/investigate-alternative-approaches-to-the-ascender-app)

---

### Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [x] Problem and Solution clearly stated
- [x] Run unit/integration/application tests on device
- [ ] Added documentation
[...]

PR #1826, opened 2019-06-14
Backport fixes for 1.2.1-rc.3

### LINKED ISSUE #1811: 1.2.1-rc.2 regression: Only three of four Particle.subscribe handlers get registered
With this simple test code only the handlers t1..t3 get called t4 doesn't do anything.
```
void setup() {
  Serial.begin();
  Particle.subscribe("t1", test, MY_DEVICES);
  Particle.subscribe("t2", test, MY_DEVICES);
  Particle.subscribe("t3", test, MY_DEVICES);
  Particle.subscribe("t4", test, MY_DEVICES);
}

void test(const char* event, const char* data) {
  Serial.println(event);
}
```
This works as expected (t1..t4) with 1.1.0 but 1.2.1-rc.2 fails (not tried intermediate RCs)
Tested with Photon and Argon.

technobly:
## This PR consists of fixes and enhancements cherry-picked from develop to create 1.2.1-rc.3

- [bugfix] [lte] cellular fixes [#1824](https://github.com/particle-iot/device-os/pull/1824)
- [bugfix] fixes #1811 - increases the number of event handlers to 6 [#1822](https://github.com/particle-iot/device-os/pull/1822)
- [bugfix] [gen 3] fixes memory usage diagnostics (reported negative values in safe mode) [#1819](https://github.com/particle-iot/device-os/pull/1819)
- [bugfix] System.disableUpdates() operates asynchronously [#1801](https://github.com/particle-iot/device-os/pull/1801)
- [enhancement] allow the bootloader to be flashed over DFU [#1788](https://github.com/particle-iot/device-os/pull/1788)


communication/tests/catch/protocol.cpp
```diff
@@ -66,11 +66,11 @@ void event_handler(const char* event, const char* data)
 {
 }
 
-SCENARIO("5 subscribe messages are registered")
+SCENARIO("6 subscribe messages are registered")
```
zfields: Shouldn't you use MAX_SUBSCRIPTIONS here and in the test as well?

technobly: Yes probably :)  We'll have to fix it going forward.

technobly: Might be hard to include protocol_defs.h though... not sure.  At least if the number changes the test will fail.  That said, this test was/is broken and fixed in PR https://github.com/particle-iot/device-os/pull/1524 so we need to get that fixed up and merged asap.

PR #1833, opened 2019-06-21
Exports part of mbedtls to the user application

avtolstoy:
### Problem

We'd like to have access to mbedTLS functions available in system-firmware from user application in certain cases.

### Solution

This PR exports crypto dynalib from system side and imports it from the user part.

### Steps to Test

TODO

### Example App

TODO

### References

N/A

---

### Completeness

- [ ] Added documentation


m-mcgowan:
Just for the record, this is not planned to be released but is being made available for hacking. Currently, mbedtls is provide within the system so that the application can include it's own version of the library.

PR #1836, opened 2019-06-23
Feature/ch33992/communications callbacks

m-mcgowan:
### Problem

Sending multiple events over an unreliable connection may result in out of date data being received. 

### Solution

Previous sent event messages are no longe retransmitted, preventing (in part) out of date data. 
Ideally events should have a timestamp set in the device so that they can be re-ordered, or the device service orders events based on their message ID (which monotonically increments, providing a total ordering.)

### Steps to Test

WiP - will be described in the IOTA test plan.

### References

- [[CH33992]](https://app.clubhouse.io/particle/story/33992/communications-callbacks-for-managing-coap-messages)

---

### Completeness

- [ ] Added documentation


system/inc/system_cloud.h
```diff
@@ -206,7 +199,28 @@ typedef struct {
  * @note The periodic functionality is not available for the Spark Core.
  */
 int spark_publish_vitals(system_tick_t period_s, void *reserved);
-bool spark_send_event(const char* name, const char* data, int ttl, uint32_t flags, void* reserved);
+
+typedef spark_protocol_send_event_data spark_send_event_data;
```
sergeuz: I'd keep these structures separate to make it clear that `spark_send_event()` and `spark_protocol_send_event()` can evolve independently, and we're not necessarily going to expose all protocol features to the application layer.

system/src/system_update.cpp
```diff
@@ -93,6 +93,46 @@ const char* flag_to_string(uint8_t flag) {
     return flag ? "true" : "false";
 }
 
+using particle::protocol::message_handle_t;
+
+const message_handle_t MESSAGE_HANDLE_INVALID = -1;
```
sergeuz: Can we guarantee that this ID will never be assigned to a valid message (e.g. by using a larger signed type for `message_handle_t`), or reimplement the client code so that it doesn't depend on magic values?

system/inc/system_cloud.h
m-mcgowan: I agree, and that's fine they can diverge later. if they do need to diverge then we can split them out as separate types and add the requisite copying, but for now it's more efficient to use a typedef, since they are the same thing at present.  Let's not take the hit now for something we may not need.

system/src/system_update.cpp
m-mcgowan: Yes totes, this needs fixing, but it's not new - -1 was used rather than a constant. 

I'll add checks in the message_id generator that it excludes 0 and 65535, and message ids will start close to the maximum so we can see the effects of rollover.

communication/src/lightssl_protocol.h
```diff
@@ -77,9 +80,28 @@ class LightSSLProtocol : public Protocol
...
 	}
 
-	virtual int command(ProtocolCommands::Enum command, uint32_t data) override;
+	void wake() {
+	    // nothing to do here. On waking up the cloud connection and session is started anew.
+	};
+
+	bool has_unacknowledged_requests() {
+        return ack_handlers.size() > 0;
+    }
```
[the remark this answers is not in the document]
m-mcgowan: I really wish I knew why. Happy to use spaces. I'll fix up the files changed in this PR and we can work with the DeviceOS guild to find a time when we can apply consistent formatting across the codebase.

system/src/system_update.cpp
sergeuz: > but it's not new - -1 was used rather than a constant

Could you elaborate on where it was used?

communication/inc/protocol_defs.h
```diff
@@ -62,6 +62,12 @@ system_error_t toSystemError(ProtocolError error);
 
 typedef uint16_t chunk_index_t;
 
+typedef int32_t message_handle_t;
+
+const message_handle_t MESSAGE_HANDLE_INVALID = uint16_t(-1);
```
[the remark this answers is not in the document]
m-mcgowan: I think it should be 0 rather than -1.

communication/src/lightssl_protocol.h
```diff
@@ -77,16 +80,35 @@ class LightSSLProtocol : public Protocol
...
+        ack_handlers.clear();
+    }
 
 	virtual int get_status(protocol_status* status) const override
 	{
 		SPARK_ASSERT(status);
 		status->flags = 0;
 		return 0;
-	}
+    }
```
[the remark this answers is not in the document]
m-mcgowan: The strange thing is that I don't see it in my editor.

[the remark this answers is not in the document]
m-mcgowan: I've made a pass through by indenting/unindenting files which seems to make them consistent. Would be good to have an automated check and automated assistance for this.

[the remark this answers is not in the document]
sergeuz: Thanks! I do care about readability, but I don't want us to spend time on this kind of issues during reviews. I'll submit a PR that fixes indentation in the entire comms module after we merge all current PRs targeting that module.

system/src/system_update.cpp
m-mcgowan: I've removed the use of the constant in the latest commits.

m-mcgowan:
Closing this out since it's been a couple of years. Feel free to reopen and merge if it is still useful.

PR #1838, opened 2019-06-26
Rework pinmap for each platforms

YutingYou:
### Problem

We use conditional compilation to reuse the pin definition which makes our pinmap messy. The pin definition should be moved to platform-specific implementation headers.

Pin information `STM32_Pin_Info` and `NRF5X_Pin_Info` are replaced by `Hal_Pin_Info`, we'll abandon the platform-related naming to get rid of some workarounds such as `typedef NRFX_Pin_Info STM32_Pin_Info`.

### Steps to Test

The old pin definition is fragmented, if you want to review the pin definition, you can utilize VSCode highlight function by adding two lines platform definition to compare.
```
#include "platforms.h"
#define PLATFORM_ID PLATFORM_XSOM 
```
image

### References

[CH34916]

---

### Completeness

- [internal] Refactors platform pinmap to be in platform-specific headers


hal/inc/pinmap_hal.h
```diff
@@ -37,24 +26,24 @@
 extern "C" {
 #endif
 
-/* Exported types ------------------------------------------------------------*/
+typedef struct Hal_Pin_Info Hal_Pin_Info;
```
avtolstoy: Let's remove this and rely on `pinmap_impl.h` to define `Hal_Pin_Info` for us.

hal/src/core/pinmap_hal.c
```diff
@@ -1,73 +1,61 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
-/* Includes ------------------------------------------------------------------*/
 #include "pinmap_hal.h"
 #include "pinmap_impl.h"
```
avtolstoy: Will anyway be included by `pinmap_hal.h`

hal/src/gcc/pinmap_impl.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
```
avtolstoy: Copyright header.

hal/src/gcc/pinmap_impl.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "pinmap_hal.h"
```
avtolstoy: Is this necessary? `pinmap_hal.h` is supposed to include implementation instead.

Same:
https://github.com/particle-iot/device-os/pull/1838/files#diff-164b14eb9d911cd3e319975e9c752d06R20
https://github.com/particle-iot/device-os/pull/1838/files#diff-0170c416cc4060c26c20782c5a669a4bR3
https://github.com/particle-iot/device-os/pull/1838/files#diff-6505829b97e07fa5b407eb51407c4d08R24

hal/src/nRF52840/pinmap_hal.c
```diff
@@ -15,16 +15,14 @@
...
-/* Includes ------------------------------------------------------------------*/
 #include "pinmap_hal.h"
 #include "pinmap_impl.h"
 #include "pwm_hal.h"
 #include "module_info.h"
 
-NRF5x_Pin_Info __PIN_MAP[TOTAL_PINS] =
-{
-#if PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON || PLATFORM_ID == PLATFORM_XENON
+#if PLATFORM_ID == PLATFORM_ARGON
```
avtolstoy: Should we move these to platform-specific source files as well?

hal/src/nRF52840/pinmap_impl.h
```diff
@@ -15,44 +15,498 @@
...
+#define PWM_CHANNEL_NONE CHANNEL_NONE
+#define EXTI_CHANNEL_NONE CHANNEL_NONE
 
-#ifdef  __cplusplus
-}
-#endif
+#define DEFAULT_PWM_FREQ 500 // 500Hz
+#define TIM_PWM_FREQ DEFAULT_PWM_FREQ
+
+#if PLATFORM_ID == PLATFORM_ARGON
```
avtolstoy: Should we move these to platform-specific implementation headers?

hal/src/newhal/pinmap_impl.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
```
avtolstoy: Copyright header.

hal/src/stm32f2xx/pinmap_impl.h
```diff
@@ -56,9 +46,330 @@ extern PinMode HAL_GPIO_Recall_Pin_Mode(uint16_t pin);
...
 #define DAC_CHANNEL_NONE CHANNEL_NONE
 
-#ifdef	__cplusplus
-}
-#endif
+#define TIM_PWM_FREQ 500 // 500Hz
+
+#define SERVO_TIM_PWM_FREQ 50 // 50Hz
+
+#if PLATFORM_ID == PLATFORM_PHOTON_PRODUCTION
```
avtolstoy: Should we move these to platform-specific implementation headers?

hal/src/core/pinmap_impl.h
```diff
@@ -1,64 +1,117 @@
...
+    TIM_TypeDef     *timer_peripheral;
+    uint16_t        timer_ch;
+    PinMode         pin_mode;
+    uint16_t        timer_ccr;
+    int32_t         user_property;
+} Hal_Pin_Info;
 
-extern STM32_Pin_Info PIN_MAP[];
-STM32_Pin_Info* HAL_Pin_Map(void);
+extern Hal_Pin_Info PIN_MAP[];
```
avtolstoy: Since we are already doing the refactoring, we should not expose this array directly and should convert any direct usage of it to use `HAL_Pin_Map()`. This applies to all the platforms.

technobly:
Will this break a lot of library usage of these functions? I would argue this is our public API for low level pin usage, so if it will break things this should be deprecated in 1.x.x and removed only in 2.x.x. We would also need to fix the warnings being seen in the web IDE.

avtolstoy:
> Will this break a lot of library usage of these functions?

Which functions? We are not introducing any breaking changes. This is mostly just a reorganization of where the pin names and internal structures come from: from per-platform hal source folder.

hal/src/stm32f2xx/pinmap_impl.h
```diff
@@ -1,53 +1,43 @@
...
-  pin_t gpio_pin;
-  uint8_t gpio_pin_source;
-  uint8_t adc_channel;
-  uint8_t dac_channel;
-  TIM_TypeDef* timer_peripheral;
-  uint16_t timer_ch;
-  PinMode pin_mode;
-  uint16_t timer_ccr;
-  int32_t user_property;
-} STM32_Pin_Info;
```
avtolstoy: Let's leave a compatibility `STM32_Pin_Info` typedef here just in case.

hal/src/nRF52840/pinmap_impl.h
```diff
@@ -15,44 +15,498 @@
...
+    PinMode     pin_mode;  // GPIO pin mode
+    PinFunction pin_func;
     uint8_t     adc_channel;
     uint8_t     pwm_instance;   // 4 instances on nRF52, range: 0~3
     uint8_t     pwm_channel;    // 4 channels in each instance, range: 0~3
-	uint8_t     pwm_resolution; // default 8bit, max 15bit
+    uint8_t     pwm_resolution; // default 8bit, max 15bit
     uint8_t     exti_channel;   // 16 channels
-    uint32_t    user_data;      
-} NRF5x_Pin_Info;
```
avtolstoy: Let's leave a compatibility `NRF5x_Pin_Info` typedef here just in case.

hal/inc/pinmap_hal.h
YutingYou: This a forward declaration, `Hal_Pin_Info` will be used in driver files or wiring files, they only include `pinmap_hal.h`, we leave the detailed structure implementation in the different platform folders.

hal/src/core/pinmap_hal.c
YutingYou: Fixed in all platforms.

hal/src/gcc/pinmap_impl.h
YutingYou: Added.

hal/src/gcc/pinmap_impl.h
YutingYou: Removed.

hal/src/nRF52840/pinmap_hal.c
YutingYou: I'll split `pinmap_hal.c` and `pinmap_impl.h` into each platform folders, should we rename the pin definition file with the name `pinmap_defines.h` which will be included in `pinmap_impl.h`?

hal/src/nRF52840/pinmap_impl.h
YutingYou: same as ^^

hal/src/newhal/pinmap_impl.h
YutingYou: Added.

hal/src/stm32f2xx/pinmap_impl.h
YutingYou: Fixed, same as ^^

hal/src/core/pinmap_impl.h
YutingYou: Removed.

hal/src/stm32f2xx/pinmap_impl.h
YutingYou: Added.

hal/src/nRF52840/pinmap_impl.h
YutingYou: Added.

hal/src/electron/pinmap_hal.c
```diff
@@ -50,12 +40,7 @@
  * GPIO_PinSource15: D5, LVLOE_UC
  */
 
-
-
-/* Private typedef -----------------------------------------------------------*/
-
-STM32_Pin_Info __PIN_MAP[TOTAL_PINS] =
-{
+Hal_Pin_Info __PIN_MAP[TOTAL_PINS] = {
```
avtolstoy: Let's avoid using symbols starting with `_` as those are reserved for internal compiler/libc/libstdc++ stuff.

```suggestion
static Hal_Pin_info s_pin_map[TOTAL_PINS] = {
```

Same thing in other implementation files.

hal/src/xenon/pinmap_defines.h
```diff
@@ -0,0 +1,171 @@
...
+#define RX1 D26
+#define CTS1 D22
+#define RTS1 D23
+
+#define BTN 22
+#define RGBR 23
+#define RGBG 24
+#define RGBB 25
+
+#define WKP D8
```
avtolstoy: `WKP` should be changed to `A7` on SoM platforms due to https://github.com/particle-iot/device-os/pull/1837

hal/src/electron/pinmap_hal.c
YutingYou: Fixed

hal/src/xenon/pinmap_defines.h
YutingYou: Fixed

avtolstoy:
@YutingYou Please rebase on current develop and let's merge this.

PR #1839, opened 2019-06-26
[Gen 3] Fixes mesh pub/sub socket consuming all packet buffers

### LINKED ISSUE #1828: Gen 3 devices run out of network packet buffers
## Bug Report

Originally mentioned on:
https://community.particle.io/t/using-mesh-publish-kills-the-cloud-connection/50129

Gen 3 devices run out of network packet buffers when only doing `Mesh.publish()`es without even a single `Mesh.subscribe()`.

### Steps to Reproduce

See the test app in the community forums thread, or simply have multiple nodes within the network doing only `Mesh.publish()`es without any subscriptions.

### Test App

See community forums thread.

### References

- https://community.particle.io/t/using-mesh-publish-kills-the-cloud-connection/50129

avtolstoy:
### Problem

See #1828

### Solution

This PR mainly simply starts mesh pub/sub thread that reads the data out of the socket even if only `publish()` is called.

Apart from that it:
1. Refactors/cleans up `MeshPublish` class
2. Provides a workaround for wiring `Thread` class where the thread being started was not given processing time due to `os_thread_yield`
3. Enables mutlicast loop LwIP options which enables on-device testing of mesh pub/sub and adds the `MESH_01_PublishWithoutSubscribeStillReadsDataOutOfPubSubSocket` test.

### Steps to Test

- `wiring/no_fixture`: `MESH_01_PublishWithoutSubscribeStillReadsDataOutOfPubSubSocket`

### Example App

N/A

### References

- Closes #1828 
- [CH35133]

---

### Completeness

- [bugfix] [Gen 3] Fixes mesh pub/sub socket consuming all packet buffers

### DIFF
--- a/wiring/src/spark_wiring_mesh.cpp
+++ b/wiring/src/spark_wiring_mesh.cpp
@@ -27,10 +27,6 @@
 
 namespace spark {
 
-int mesh_loop() {
-    return spark::Mesh.poll();
-}
-
 bool MeshPublish::Subscriptions::event_handler_exists(const char *event_name, EventHandler handler,
         void *handler_data, SubscriptionScope::Enum scope, const char* id)
 {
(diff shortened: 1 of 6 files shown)


avtolstoy:
[CH35133]

avtolstoy:
@technobly If we are doing 1.3.0-rc.2, sounds like a good idea to put it in there instead of 1.3.1

wiring/src/spark_wiring_mesh.cpp
```diff
@@ -137,37 +133,71 @@ int MeshPublish::fetchMulticastAddress(IPAddress& mcastAddr) {
...
     return SYSTEM_ERROR_NONE;
 }
 
-int MeshPublish::uninitialize_udp() {
+int MeshPublish::uninitializeUdp() {
+    // NB: this function is not really used at the moment
+
+    // Use atomic exit_ as a simple synchronization primitive to ensure thread-safety
+    // instead of a mutex here.
+    if (exit_) {
```
m-mcgowan: there's a race condition here between reading and setting the value - an atomic exchange will avoid this.

```suggestion
    if (exit_.exchange(true) {
```

PR #1841, opened 2019-07-02
fixes RSSI regression on G350 (2G) devices [ch35194]

technobly:
# This PR also needs to be cherry-picked to `release/v1.2.1` for the 1.2.1 default release.

### Problem

A regression in the Cellular.RSSI() API was introduced in Device OS 1.2.0-rc.1 PR #1779 where the G350 RSSI functionality was broken.

### Solution

Forces ACT_GSM for G350 in functions that query AT+COPS?

### Steps to Test

- Run new test CELLULAR_07_rssi_is_valid in TEST=wiring/no_fixture
- Try pressing the MODE button on G350 one time to see Signal Strength indicated in "Bars" flashed with a green LED.  With 1.2.0-rc.1 / 1.2.1-rc.3 / 1.3.0-rc.1, the G350 will NOT flash 1-5 blips of the green LED.  With this PR it should at least flash 1 green blip, if not more.

### References

https://community.particle.io/t/no-signal-strength-information-for-electron-2g-on-device-os-1-0-0/47460/16

### Known Issues

With the added test CELLULAR_07_rssi_is_valid, it became clear that on Gen 3 devices Boron & B SoM, the Cellular.RSSI() API was not implemented correctly and needs to be fixed.  This will be fixed in a future PR to reduce risk to this fix which will make its way into 1.2.1 default without a 1.2.1-rc.4.

---

### Completeness

- [N/A] Added documentation

---

- [bugfix] fixes RSSI regression on G350 (2G) devices [#1841](https://github.com/particle-iot/device-os/pull/1841)

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1346,9 +1346,9 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
         if (RESP_OK != waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
             goto failure;
         }
-        // +CREG, +CGREG, +COPS does not contain <AcT> for G350 devices.
-        // After we are registered to CSD or PSD, populate _net.act with ACT_GSM to ensure
-        // Device Diagnostics and RSSI() API have a RAT for conversion lookup purposes.
+        // +CREG, +CGREG, +COPS do not contain <AcT> for G350 devices.
+        // Force _net.act to ACT_GSM to ensure Device Diagnostics and
+        // RSSI() API's have a RAT for conversion lookup purposes.
         if (_dev.dev == DEV_SARA_G350) {
             _net.act = ACT_GSM;
         }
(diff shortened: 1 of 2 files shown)


technobly:
Thanks @avtolstoy !  I have tested on G350 / U260 / R410 and Boron (actually it fails on Boron though as noted above).  I've also created and shared some binaries in the community for the original reporter of this issue to test, so I'm hoping to get more feedback there.

avtolstoy:
> actually it fails on Boron though as noted above

Do you have any info about that?

One suggestion I would make is to possibly change the test to not use `rssi` and `qual` directly and instead use the new APIs: `getAccessTechnology()`, `getStrength()`, `getStrengthValue()`, `getQuality()`, `getQualityValue()`.

technobly:
> Do you have any info about that?

The issue with Boron is that CellularSignal .rssi and .qual are not set to proper values (which seem to be available via the newer vitals calculations).

> One suggestion I would make is to possibly change the test to not use rssi and qual directly and instead use the new APIs: getAccessTechnology(), getStrength(), getStrengthValue(), getQuality(), getQualityValue().

@avtolstoy The purpose of the test is to insure the Cellular.RSSI() API is functioning properly for .rssi (and adding .qual wouldn't be a bad idea either), but if we don't have a test yet for the other new vitals functions those could be good to perform a quick smoke test on as well!

PR #1846, opened 2019-07-09
Fixes Boron charging current limitation when powered by VUSB

YutingYou:
### Problem

Customers are limited to 500mA charging current when Boron is powered by VUSB.

Affected Device: Gen3 devices

### Analysis
When Boron is powered by VUSB pin, the power supply is blocked by the diode. For nRF52840,
> The USB peripheral has a dedicated internal voltage regulator for converting the VBUS supply to 3.3 V used by the USB signalling interface (D+ and D- lines, and pull-up on D+). The rest of the USB peripheral (USBD) is supplied through the main supply like any other on-chip feature. As a consequence, both VBUS and either VDDH or VDD supplies are required for USB peripheral operation.

image

Since the USB peripheral is left to an uncontrolled state, PMIC would wrongly assume the board is connected to a USB host.

### Solution

1. Add a workaround in system power manager to manually increase charging current limit and decrease input voltage limit.
2. Add an API to detect USB state

When the PMIC driver detects the USB host connection, it will also check whether USB is really connected. If not, it will increase charging current.

### Steps to Test

1. Charge Boron via VUSB pin, observe whether the charging current is more than 500mA, should be around 800mA.
2. Connect Boron to a USB host, observe whether the charging current is no more than 500mA.

### Test result

image

image

### Example App

N/A

### References

[CH34730]

---

### Completeness

- [enhancement] [Boron/BSoM] power: Increases charging current to 900mA when powered through VIN (VUSB) pin


hal/inc/usb_hal.h
```diff
@@ -47,6 +47,16 @@
 extern "C" {
 #endif
 
+typedef enum {
```
avtolstoy: `typedef enum HAL_USB_State { ... } HAL_USB_State;`

hal/inc/hal_platform.h
```diff
@@ -246,4 +246,8 @@
 #define HAL_PLATFORM_NFC 0
 #endif /* HAL_PLATFORM_NFC */
 
+#ifndef HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VUSB
+#define HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VUSB (0)
```
avtolstoy: This sounds confusing as `VUSB` usually comes from the USB port, let's change this to `WHEN_POWERED_BY_VIN`.

system/src/system_power_manager.cpp
```diff
@@ -148,6 +152,23 @@ void PowerManager::handleUpdate() {
...
+        // and decrease input voltage limit to 3880mV.
+        // More details are in clubhouse [CH34730]
+        if (HAL_USB_Get_State() == HAL_USB_STATE_DETACHED) {
+            if (power.getInputCurrentLimit() != DEFAULT_INPUT_CURRENT_LIMIT) {
+                power.setInputCurrentLimit(DEFAULT_INPUT_CURRENT_LIMIT);
+                power.setInputVoltageLimit(3880);
+            }
+        } else {
+            power.setInputCurrentLimit(USB_HOST_INPUT_CURRENT_LIMIT);
+            power.setInputVoltageLimit(4360);
```
avtolstoy: Any reason not to set the input voltage limit to 3.88V globally if everything functions correctly?

system/src/system_power_manager.cpp
```diff
@@ -148,6 +152,23 @@ void PowerManager::handleUpdate() {
...
     switch (vbus_stat) {
       case 0x01:
+#if HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VUSB
+        // Workaround: 
+        // when Gen3 device is powered via VUSB, USB peripheral gets no power supply.
+        // In this case BQ24195 will wrongly assume the device is connected to a USB host.
+        // This workaround is to manually increase charging current limit from 500mA to 900mA
+        // and decrease input voltage limit to 3880mV.
+        // More details are in clubhouse [CH34730]
+        if (HAL_USB_Get_State() == HAL_USB_STATE_DETACHED) {
```
avtolstoy: Nitpicking: formatting. The file uses 2 spaces as indentation.

system/src/system_power_manager.cpp
```diff
@@ -148,6 +152,23 @@ void PowerManager::handleUpdate() {
     uint8_t vbus_stat = status >> 6;
     switch (vbus_stat) {
       case 0x01:
+#if HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VUSB
```
avtolstoy: This logic doesn't cover the case when the device is powered by both VIN and through USB port and then during the runtime the USB cable is disconnected or reconnected.

We need to subscribe to changes of the USB state and manually trigger DPDM detection when the state changes to `HAL_USB_STATE_DETACHED` or when it changes from `HAL_USB_STATE_DETACHED` to any other. I propose we add a HAL function to register a callback to received USB state changes:

```c++
typedef void (*HAL_USB_State_Callback)(HAL_USB_State state, void* data);
int HAL_USB_Set_State_Change_Callback(HAL_USB_State_Callback cb, void* context, void* reserved);
```

system/src/system_power_manager.cpp
YutingYou: Looks like it is unnecessary, removed.

system/src/system_power_manager.cpp
YutingYou: Now, it supports automatically increasing charging current when powered both by VUSB and USB host. Test several times to plug out and plug in the USB cable, the charging current changes correctly.

hal/src/nRF52840/usb_hal_cdc.c
```diff
@@ -87,6 +83,9 @@ typedef struct {
     volatile bool           rx_done;
 
     void (*bit_rate_changed_handler)(uint32_t bitRate);
+    HAL_USB_State_Callback  state_callback[MAX_USB_STATE_CB_NUM];
+    void*                   state_callback_context[MAX_USB_STATE_CB_NUM];
+    int                     state_callback_count;
```
avtolstoy: This seems unnecessary. The unused slots should be `NULL`.

hal/src/nRF52840/usb_hal_cdc.c
```diff
@@ -485,3 +494,17 @@ bool usb_hal_is_connected(void) {
...
     m_usb_instance.bit_rate_changed_handler = handler;
 }
+
+HAL_USB_State usb_hal_get_state() {
+    return m_usb_instance.state;
+}
+
+int usb_hal_set_state_change_callback(HAL_USB_State_Callback cb, void* context, void* reserved) {
+    if (m_usb_instance.state_callback_count >= MAX_USB_STATE_CB_NUM) {
+        return -1;
```
avtolstoy: Let's use `system_error_t` codes here.

system/src/system_power_manager.cpp
```diff
@@ -477,4 +495,9 @@ void PowerManager::deinit() {
   detachInterrupt(LOW_BAT_UC);
 }
 
+void PowerManager::usbStateChangeHandler(HAL_USB_State state, void* context) {
+  PowerManager* power = (PowerManager*)context;
+  power->update();
```
avtolstoy: I'm not sure that this is enough. When I was testing the previous iteration of this PR, I didn't see DPDM detection automatically running on USB cable disconnection/reconnection. Could you please verify that it does happen and if not, we need to set a flag to force it in `handleUpdate()`.

hal/src/nRF52840/usb_hal_cdc.c
YutingYou: `state_callback_count` is used as the index of the callback slot when registering or calling the callback.

system/src/system_power_manager.cpp
YutingYou: Did you connect Boron to the USB host? If so, it wouldn't change the charging current. `if (HAL_USB_Get_State() == HAL_USB_STATE_DETACHED) `

---
I added the log in:
```
...
  power_source_t src = g_powerSource;
  if (pwr_good) {
    uint8_t vbus_stat = status >> 6;
    LOG_DEBUG(INFO, "vbus_stat: 0x%x, usb state: %d", vbus_stat, HAL_USB_Get_State());
    switch (vbus_stat) {
      case 0x01:
...
```
Keep VIN pin connected and disconnect/reconnect USB cable, got the log:
```
00124238 [sys.power] TRACE: re-enabling charging
0000149072 [sys.power] TRACE: Battery state UNKNOWN -> CHARGING
0000149073 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000149076 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000158376 [sys.power] TRACE: Battery state CHARGING -> DISCHARGING
0000158644 [sys.power] TRACE: Battery state DISCHARGING -> NOT_CHARGING
0000158645 [sys.power] INFO: vbus_stat: 0x0, USB state: 3
0000158854 [sys.power] TRACE: Battery state NOT_CHARGING -> CHARGING
0000158854 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000168480 [sys.power] INFO: vbus_stat: 0x1, USB state: 0
0000172427 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000172430 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000176617 [sys.power] TRACE: Battery state CHARGING -> DISCHARGING
0000176886 [sys.power] TRACE: Battery state DISCHARGING -> NOT_CHARGING
0000176887 [sys.power] INFO: vbus_stat: 0x0, USB state: 3
0000177049 [sys.power] TRACE: Battery state NOT_CHARGING -> CHARGING
0000177049 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000183957 [sys.power] INFO: vbus_stat: 0x1, USB state: 0
0000186186 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000186189 [sys.power] INFO: vbus_stat: 0x1, USB state: 3
0000192023 [sys.power] INFO: vbus_stat: 0x1, USB state: 0
```
Number 3(`HAL_USB_STATE_DEFAULT`) means it connected to the USB host. When disconnecting USB host, the charging current increased to 800mA.

hal/src/nRF52840/usb_hal_cdc.c
avtolstoy: But why? You can simply traverse the array, find a `NULL` entry and set replace it with the callback.

YutingYou: OK, got it.

system/src/system_power_manager.cpp
avtolstoy: Does it get back to 500mA though? And is it possible that these changes are actually caused by the battery state fluctuation we see here in the log? What happens if you connect a USB adapter?

YutingYou: > Does it get back to 500mA though? 
1. Insert USB current measure tool to a socket, use a USB cable to connect the tool and  Boron VIN pin, the current on the USB current measure tool is 800mA.
2. Connect USB Host to the USB port on Boron, the current decreases to less than 500mA.
3. Disconnect USB Host, the current increases to 800mA.

> And is it possible that these changes are actually caused by the battery state fluctuation we see here in the log? 
The log is triggered by the interrupts of VIN pin and USB. In the interrupt handler, `update_` flag is updated, the log is printed in the thread loop.

> What happens if you connect a USB adapter?
Since the VIN pin is connected to a USB adapter, I didn't try to connect the USB port to a USB adapter. I can try it tomorrow to see whether the charging current will remain high.

system/src/system_power_manager.cpp
```diff
@@ -146,10 +150,28 @@ void PowerManager::handleUpdate() {
   power_source_t src = g_powerSource;
   if (pwr_good) {
     uint8_t vbus_stat = status >> 6;
+    // LOG_DEBUG(INFO, "vbus_stat: 0x%x, usb state: %d, current limit: %d", vbus_stat, HAL_USB_Get_State(), power.getInputCurrentLimit());
     switch (vbus_stat) {
-      case 0x01:
+      case 0x01: {
+#if HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VIN
+        // Workaround: 
+        // when Gen3 device is powered via VUSB, USB peripheral gets no power supply.
```
avtolstoy: Let's change the wording to 'powered by VIN'. I know the silkscreen on the PCB says VUSB, but it's confusing as VUSB usually comes from the USB port.

system/src/system_power_manager.h
```diff
@@ -20,9 +20,11 @@
 #include "system_tick_hal.h"
 #include "concurrent_hal.h"
 #include "hal_platform.h"
+#include "usb_hal.h"
 
 namespace particle { namespace power {
 
+static const uint16_t USB_HOST_INPUT_CURRENT_LIMIT = 500;
```
avtolstoy: This can be removed.

system/src/system_power_manager.cpp
```diff
@@ -165,6 +187,8 @@ void PowerManager::handleUpdate() {
           // so just check input current source register whenever we are in this state
           if (power.getInputCurrentLimit() != DEFAULT_INPUT_CURRENT_LIMIT) {
             power.setInputCurrentLimit(DEFAULT_INPUT_CURRENT_LIMIT);
+            power.setInputVoltageLimit(3880);
```
avtolstoy: Let's be cautious and enable this only if `HAL_INCREASE_CHARGING_CURRENT_WHEN_POWERED_BY_VIN` is `1`, so for Gen 3 only.

PR #1852, opened 2019-07-12
Refactors linker files

XuGuohui:
Details to be documented ...


build/arm/linker/linker_nrf52840_common.ld
```diff
@@ -1,22 +1,17 @@
+/* Default stack sizes. These are used by the startup in order to allocate stacks for the
```
XuGuohui: This file is only referenced by bootloader, would it be better to rename it to `linker_nrf52840_bootloader_common.ld`?

XuGuohui:
The `linker.ld` under `modules/<gen3_platform>/system_part1` and `modules/<gen3_platform>/user_part` can be merged into a common `memory_system_part1.ld` and `memory_user_part.ld` respectively.

m-mcgowan:
@XuGuohui could you fix up the conflicts please!  Do we have a plan for which release this will be included in?

m-mcgowan:
Please fix up the conflicts and write up the goals of this PR so that we can review.

XuGuohui:
> @XuGuohui could you fix up the conflicts please! Do we have a plan for which release this will be included in?

I believe that this is not an urgent stuff, unlike feature and bug-fix, so I think it should be included since 1.4.0? Cc @avtolstoy

avtolstoy:
Yes, it's planned to go in 1.4.0. I've updated the milestone.

PR #1862, opened 2019-07-23
[electron, gen3] Temporarily increase IDLE task priority whenever a thread exits (calls vTaskDelete)

avtolstoy:
### Problem

Due to the busy-loop nature of the application thread and higher priority (2) FreeRTOS IDLE thread responsible for thread resources cleanup (stack) is not being given CPU time, causing memory leaks, unless the application thread goes into a blocked state (waiting on a synchronization primitive or delaying).

### Solution

Temporarily increase IDLE thread priority to `configMAX_PRIORITIES - 1` (maximum) whenever `vTaskDelete(NULL)` is called using `portPRE_TASK_DELETE_HOOK` to clean up the stack as soon as possible and restore it back to `tskIDLE_PRIORITY` once IDLE thread runs using idle hook.

### Steps to Test

Set a breakpoint to `vApplicationIdleHook`, make some thread exit, it should get executed.

### Example App

N/A

### References

- [CH36038]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/rtos_hook.cpp
+++ b/hal/src/electron/rtos_hook.cpp
@@ -19,6 +19,13 @@
 #include "task.h"
 #include "service_debug.h"
 #include "concurrent_hal.h"
+#include <atomic>
+
+namespace {
+
+std::atomic_bool sRestoreIdleThreadPriority(false);
+
+} // anonymous
 
 extern "C" {
 
@@ -28,9 +35,13 @@ void vApplicationStackOverflowHook(TaskHandle_t, char*) {
 }
 #endif
 
-void vApplicationTaskDeleteHook(void *pvTaskToDelete, volatile BaseType_t* pxPendYield) {
+void vApplicationTaskDeleteHook(void* pvTaskToDelete, volatile BaseType_t* pxPendYield) {
     (void)pvTaskToDelete;
     (void)pxPendYield;
+
+    // NOTE: this hook is executed within a critical section
+
+    sRestoreIdleThreadPriority = true;
 
     // Temporarily raise IDLE thread priority to (configMAX_PRIORITIES - 1) (maximum)
     // to give it some processing time to clean up the deleted task resources.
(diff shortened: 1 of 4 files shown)


hal/src/electron/rtos_hook.cpp
```diff
@@ -10,4 +28,22 @@ void vApplicationStackOverflowHook(TaskHandle_t, char*) {
...
+    // to give it some processing time to clean up the deleted task resources.
+    vTaskPrioritySet(xTaskGetIdleTaskHandle(), configMAX_PRIORITIES - 1);
+
+    // Immediately request the scheduler to yield to now higher priority IDLE thread
+    *pxPendYield = pdTRUE;
+}
+
+void vApplicationIdleHook(void) {
+    // Restore IDLE thread priority back to the default one
+    vTaskPrioritySet(xTaskGetIdleTaskHandle(), tskIDLE_PRIORITY);
```
sergeuz: How about adding a condition here, so that we only change the priority if it was raised previously?

avtolstoy: If you take a look at `vTaskPrioritySet` implementation, nothing will actually happen if the set priority is the same as the current one.

We'd still need to add a critical section around the check and `vTaskPrioritySet`, so I don't seem much of a performance benefit for example, it's almost 1-1.

technobly: Since the [FreeRTOS docs](https://www.freertos.org/a00129.html) don't spell out what will happen when the same priority is set, it could change in the future for better or worse.  I agree it would be a good idea to guard against this.  Since it should only run the one time that a thread is deleted, there shouldn't be much overhead in an additional safety check.

BTW this works great on Boron!  Clever use of vApplicationTaskDeleteHook :)  This should really be a documented pattern, I see so many SO questions about how this issue.

avtolstoy: I'll add an atomic flag just in case :+1:

PR #1863, opened 2019-07-25
communications public interface

m-mcgowan:
### Problem

The internal and public headers were contained in the `src` folder which was exported as the public interface, exporting internal headers at the same time. 

### Solution

Add an `inc` folder that contains the public headers while keeping the private headers in the `src` folder. 

### Steps to Test

Build all platforms. 

---

### Completeness

- [ ] Added documentation


m-mcgowan:
In chatting with @zfields he asked me to not fix the failing build since he's working on a PR based on this where he will fix up the include path since it's in code that is changing.

PR #1868, opened 2019-07-30
Fix/reload dct functions

XuGuohui:
### Problem

Particle Update fails to update a Photon/P1 from Device OS <1.2.1-rc.3 to 1.2.1 via DFU mode. This is caused by that after updating the system parts, the address of DCT functions that resided in system parts aren't consistent, so when bootloader tries accessing the DCT after that, it then crashes.

### Solution

Dirty a flag after every DFU write operation to the internal flash, so that after updating the system parts, next time when bootloader calls DCT functions, it will reload the DCT functions to make sure that the address of the DCT functions in new system parts are valid.

### Steps to Test

1. Build and flash the new bootloader
2. Flash older DeviceOS e.g. 1.0.0 onto the device
3. Reset and go into DFU
4. Write into DCT
5. Write system-part1 and system-part2 of DeviceOS release v1.2.1 without resetting
6. Write into DCT again, this should succeed and not crash the device

### Example App

N/A

### References

https://github.com/particle-iot/particle-cli/pull/499
https://community.particle.io/t/device-os-1-2-1-and-particle-cli-1-43-fail-in-particle-update/51140

- [bugfix] [Photon/P1] Bootloader correctly re-imports the DCT functions from system firmware after its modification through DFU.

### DIFF
--- a/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
+++ b/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
@@ -1,3 +1,6 @@
 
 TARGET_USB_FS_PATH = $(PLATFORM_MCU_PATH)/STM32_USB_Device_Driver
 INCLUDE_DIRS += $(TARGET_USB_FS_PATH)/inc
+
+# bootloader_dct.h is referenced in usbd_flash_if.c
+INCLUDE_DIRS += $(PLATFORM_MODULE_PATH)/../bootloader/src/photon
--- a/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
+++ b/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
@@ -1,6 +1,3 @@
 
 TARGET_USB_FS_PATH = $(PLATFORM_MCU_PATH)/STM32_USB_Device_Driver
 INCLUDE_DIRS += $(TARGET_USB_FS_PATH)/inc
-
-# bootloader_dct.h is referenced in usbd_flash_if.c
-INCLUDE_DIRS += $(PLATFORM_MODULE_PATH)/../bootloader/src/photon
--- a/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/src/usbd_flash_if.c
+++ b/platform/MCU/STM32F2xx/STM32_USB_Device_Driver/src/usbd_flash_if.c
@@ -30,9 +30,10 @@
 #include "usbd_dfu_mal.h"
 #include "usb_bsp.h"
 #include "hw_config.h"
-
-#if PLATFORM_ID == 6 || PLATFORM_ID == 8
+#include "platforms.h"
+#if PLATFORM_ID == PLATFORM_PHOTON_PRODUCTION || PLATFORM_ID == PLATFORM_P1
 #include "bootloader_dct.h"
+#include "module_info.h"
 #endif
 
 /* Private typedef -----------------------------------------------------------*/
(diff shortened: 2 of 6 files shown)


platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
```diff
@@ -1,3 +1,6 @@
 
 TARGET_USB_FS_PATH = $(PLATFORM_MCU_PATH)/STM32_USB_Device_Driver
 INCLUDE_DIRS += $(TARGET_USB_FS_PATH)/inc
+
+# bootloader_dct.h is referenced in usbd_flash_if.c
+INCLUDE_DIRS += $(PLATFORM_MODULE_PATH)/../bootloader/src/photon
```
avtolstoy: It feels somewhat weird to introduce a dependency on the bootloader, so perhaps we should simply add an `extern` declaration of the required function to `usbd_flash_if.c` or move `bootloader_dct.h` into `hal`.

platform/MCU/STM32F2xx/STM32_USB_Device_Driver/src/usbd_flash_if.c
```diff
@@ -155,6 +160,15 @@ uint16_t FLASH_If_Write(uint32_t Add, uint32_t Len)
   /* Lock the internal flash */
   FLASH_Lock();
 
+  /* This make sure that after the system parts being updated,
+   * the address of DCT functions that resided in system parts are up to date,
+   * otherwise, bootloader cannot access the DCT after updating the system parts without a reset. */
+#if PLATFORM_ID == PLATFORM_PHOTON_PRODUCTION || PLATFORM_ID == PLATFORM_P1
+#if MODULE_FUNCTION == MOD_FUNC_BOOTLOADER
+  dct_reload_functions();
```
avtolstoy: Let's also add a check on the address and length and trigger the call to this function only when modifying system-part1 and system-part2 area.

platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
XuGuohui: This can be fixed by simply adding a forward function declaration.

avtolstoy: Or we can even move the declaration into `dct.h`, which actually resides in `platform` and abolish the `bootloader_dct.h` header.

platform/MCU/STM32F2xx/STM32_USB_Device_Driver/src/usbd_flash_if.c
XuGuohui: I'm afraid that we can only check the address, but not the length. The length being passed into the `FLASH_If_Write()` function is for the current small data block.

avtolstoy: :+1: That's fine.

platform/MCU/STM32F2xx/STM32_USB_Device_Driver/inc/include.mk
XuGuohui: 👍

platform/MCU/STM32F2xx/STM32_USB_Device_Driver/src/usbd_flash_if.c
```diff
@@ -155,6 +164,17 @@ uint16_t FLASH_If_Write(uint32_t Add, uint32_t Len)
   /* Lock the internal flash */
   FLASH_Lock();
 
+  /* This make sure that after the system parts being updated,
+   * the address of DCT functions that resided in system parts are up to date,
+   * otherwise, bootloader cannot access the DCT after updating the system parts without a reset. */
+#if PLATFORM_ID == PLATFORM_PHOTON_PRODUCTION || PLATFORM_ID == PLATFORM_P1
+#if MODULE_FUNCTION == MOD_FUNC_BOOTLOADER
+  if (Add >= module_system_part1.start_address && Add <= module_system_part2.end_address) {
```
avtolstoy: `Add < module_system_part2.end_address`?

PR #1872, opened 2019-08-02
Set path to Boost libraries

sergeuz:
### Problem

It is now required to have Boost libraries installed globally in the system to build CMake-based unit tests locally (this change was introduced in https://github.com/particle-iot/device-os/pull/1869). This PR changes the paths to Boost libraries back to `$BOOST_ROOT` which is set by CI scripts.

Note: This is a temporary solution. Ideally, we'd want all necessary paths to be detected and set using built-in CMake mechanisms.

### Steps to Test

- Run `source ci/install_boost.sh`.
- Run `ci/build_boost.sh`.
- Compile and run CMake-based unit tests.

### DIFF
--- a/ci/run_tests.sh
+++ b/ci/run_tests.sh
@@ -10,7 +10,6 @@ if contains "${BUILD_PLATFORM[*]}" unit-test; then
 	( source ./ci/install_gcovr.sh
 	source ./ci/install_boost.sh
 	./ci/build_boost.sh &&
-	cp -r ${BOOST_ROOT}/boost/ /usr/include/ &&
 	./ci/unit_tests.sh ) || die
 fi
 
--- a/test/unit_tests/CMakeLists.txt
+++ b/test/unit_tests/CMakeLists.txt
@@ -45,6 +45,11 @@ link_libraries(
   gcov
 )
 
+# TODO: These should be coming from the dependency's CMake module or find_package()
+include_directories($ENV{BOOST_ROOT})
+link_directories($ENV{BOOST_ROOT}/stage/lib)
+link_libraries(boost_program_options boost_regex boost_system boost_thread)
+
 # Build and discover unit-tests
 add_subdirectory(cellular)
 add_subdirectory(cloud)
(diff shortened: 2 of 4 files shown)


test/unit_tests/CMakeLists.txt
```diff
@@ -45,6 +45,11 @@ link_libraries(
   gcov
 )
 
+# TODO: These should be coming from the dependency's CMake module or find_package()
```
zfields: Let's get this `#TODO: ` in place before we merge.

This solution will be some form of this:
https://cmake.org/cmake/help/v3.2/module/FindBoost.html

I'll play with it and make a PR against this branch.

sergeuz: Wouldn't `find_package()` set the paths to whatever version of Boost you have installed globally in your system? I thought that we agreed that we're going to stick to the Boost libraries which are downloaded by the CI scripts for now.

zfields: No, my understanding is that it uses `$BOOST_ROOT` as a hint to find Boost.

sergeuz: Well, if we can provide `find_package()` with some hints that should work :+1:

test/unit_tests/CMakeLists.txt
```diff
@@ -14,16 +14,20 @@ get_filename_component(
...
-  REALPATH
+set(TEST_DIR ${DEVICE_OS_DIR}/test)
+set(THIRD_PARTY_DIR ${DEVICE_OS_DIR}/third_party)
+
+# Add Boost from git submodule
+set(Boost_USE_MULTITHREADED OFF)
+find_package(Boost
+  1.59.0 EXACT
+  REQUIRED
+  COMPONENTS program_options random regex system thread
```
zfields: `random` is in this list because of `<particle-iot/device-os>/ci/build_boost.sh`, but based on the list in your first PR, we may not need to check for it.

sergeuz:
@zfields The boost path changes work like a charm, thank you! I enabled multithreading support in Boost – not sure why we were building it in single-threaded mode

ci/run_tests.sh
```diff
@@ -10,7 +10,6 @@ if contains "${BUILD_PLATFORM[*]}" unit-test; then
 	( source ./ci/install_gcovr.sh
 	source ./ci/install_boost.sh
 	./ci/build_boost.sh &&
-	cp -r ${BOOST_ROOT}/boost/ /usr/include/ &&
```
sergeuz: :+1:

PR #1875, opened 2019-08-03
Fix/gen3 i2c timeout

YutingYou:
### Problem

The I2C driver on Gen3 device doesn't support timeout, once the external I2C slave device dies, the I2C bus will be held, Device OS gets stuck in the endless wait, the user application doesn't run anymore.

### Solution

Add timeout to the driver.

This PR also fixes an assertion failure on `rand()` usage from an ISR in `i2c_master_slave/i2c_slave` test.

#### Unit Test
1. Make sure this fix can pass the unit test under `user/tests/wiring/i2c_master_slave`

#### Test App
1. Download below application to Xenon-I2C-Master and Xenon-I2C-Slave devices
2. Connect Xenon-I2C-Master's SCL pin to the GND
3. Connect Xenon-I2C-Master's SCL pin back to the  Xenon-I2C-Slave's SCL pin
4. Observe the log, the user application of I2C master should not be blocked and the data should be correct.

### Example App

```c
#include "application.h"

#define MASTER 1
#define speed 100000
#define SLAVE_ADDRESS 0x18

SYSTEM_MODE(MANUAL);

Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL, {{"app", LOG_LEVEL_ALL}});

#if MASTER
int sendData(String cmd) {
   for (int i = 0; i < 1; ++i) {
        uint8_t count = Wire.requestFrom(SLAVE_ADDRESS, 6);
        // delay(10); // doesn't change anything
        while(Wire.available()) {
            char c = Wire.read();
            Serial.print(c);
        }
        Serial.println();
    }
    return 0;
}

void setup() {
    Serial.begin(115200);
    Wire.begin();
}

void loop() {
    if (Serial.available()) {
        uint8_t cmd = Serial.read();
        switch(cmd) {
            case '1': {
                Serial.println("send hello...");
                sendData("hello");
                break;
            }
            default:
                break;
        }
    }
    delay(1000);
    Serial.println("loop");
}

#else 
void requestEvent() {
    Wire.print("123456789");
}

void setup() {  
    //Wire.setSpeed(speed); // optional but same result
    Wire.begin(SLAVE_ADDRESS);    
    Wire.onRequest(requestEvent);
}
#endif
```

### References

[CH36348]
---

### Completeness

- [enhancement] [Gen3] Adds timeouts to I2C HAL operations

### DIFF
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -27,10 +27,30 @@
 #include "interrupts_hal.h"
 #include "pinmap_impl.h"
 #include "logging.h"
+#include "timer_hal.h"
 
 #define TOTAL_I2C                   2
 #define BUFFER_LENGTH               I2C_BUFFER_LENGTH
 #define I2C_IRQ_PRIORITY            APP_IRQ_PRIORITY_LOWEST
+#define EVENT_TIMEOUT               100000 // 100ms
+
+#define WAIT_TIMED(what) ({ \
+    uint32_t _micros = HAL_Timer_Get_Micro_Seconds();                           \
+    bool res = true;                                                            \
+    while((what))                                                               \
+    {                                                                           \
+        int32_t dt = (HAL_Timer_Get_Micro_Seconds() - _micros);                 \
+        bool nok = ((EVENT_TIMEOUT < dt)                                        \
+                   && (what))                                                   \
+                   || (dt < 0);                                                 \
+        if (nok)                                                                \
+        {                                                                       \
+            res = false;                                                        \
+            break;                                                              \
+        }                                                                       \
+    }                                                                           \
+    res;                                                                        \
+})
 
 /* TWI instance. */
 static nrfx_twim_t m_twim0 = NRFX_TWIM_INSTANCE(0);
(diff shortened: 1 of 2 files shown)


hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -214,7 +237,7 @@ static int twi_init(HAL_I2C_Interface i2c) {
 
 void HAL_I2C_Init(HAL_I2C_Interface i2c, void* reserved) {
     if (m_i2c_map[i2c].enabled) {
-        twi_uinit(i2c);
+        twi_uninit(i2c);
```
avtolstoy: As discussed, let's not call uninit here to avoid issues for the I2C peripheral shared between system and user code (the one that has PMIC and FuelGauge).

Going further down let's change:

```cpp
    os_thread_scheduling(false, NULL);
    if (m_i2c_map[i2c].mutex == NULL) {
        os_mutex_recursive_create(&m_i2c_map[i2c].mutex);
    } else {
        // Already initialized
        return;
    }
    HAL_I2C_Acquire(i2c, NULL);
    os_thread_scheduling(true, NULL);
```

YutingYou: Should we re-enable schedule here?
```
    os_thread_scheduling(false, NULL);
    if (m_i2c_map[i2c].mutex == NULL) {
        os_mutex_recursive_create(&m_i2c_map[i2c].mutex);
    } else {
        // Already initialized
        os_thread_scheduling(true, NULL);
        return;
    }
    HAL_I2C_Acquire(i2c, NULL);
    os_thread_scheduling(true, NULL);
```

YutingYou: Should we re-enable schedule here?
```   
os_thread_scheduling(false, NULL);
    if (m_i2c_map[i2c].mutex == NULL) {
        os_mutex_recursive_create(&m_i2c_map[i2c].mutex);
    } else {
        // Already initialized
        os_thread_scheduling(true, NULL);
        return;
    }
    HAL_I2C_Acquire(i2c, NULL);
    os_thread_scheduling(true, NULL);
```

avtolstoy: Yes, we should :+1:

avtolstoy:
@YutingYou Please rebase on develop as well.

PR #1879, opened 2019-08-06
Fix/gen3 spi concurrent

Cheong2K:
### Problem

The EthernetWing uses the high speed SPI for Gen3 devices, but the hardware design blocks to use the 2nd SPI interface unless to do some modifications to the EthernetWing.

Besides the above, the high speed SPI was not able to be shared with other SPI peripheral devices to use it concurrently with the EthernetWing since the SPI HAL has not implemented any synchronization mechanism.

### Solution

Adds HAL_SPI_Acquire() and HAL_SPI_Release() in all the functions of SPI HAL and exposes these two to Wiring layer.

### Steps to Test

Tested on a Xenon + EthernetWing + ST7735 SPI TFT Display.

Use the following Example App to test, the TFT Display and the Cloud connection should work as expected to allow user apps to use.

### Example App

```c
#include <Adafruit_ST7735.h>

SYSTEM_MODE(MANUAL);

#define CS   A1
#define DC   A0
#define RST  A2  

Adafruit_ST7735 tft = Adafruit_ST7735(CS, DC, RST);

void setup() {
    Serial.begin(115200);
    delay(3000);
    Serial.println("SPI Test Started");

    tft.initG();

    Ethernet.on();
    Ethernet.connect();
    Serial.println("Waiting to connect to Ethernet...");
    while(!Ethernet.ready());
    Serial.println("Connected to Ethernet.");
    
    Particle.connect();
}

void loop() {
    SPI.lock();
    tft.fillScreen(ST7735_BLACK);
    tft.setCursor(0, 0);
    tft.setTextColor(ST7735_WHITE);
    tft.setTextWrap(true);
    tft.print("HELLO!");
    SPI.unlock();
    delay(300);

    SPI.lock();
    tft.fillScreen(ST7735_BLACK);
    tft.setCursor(0, 0);
    tft.setTextColor(ST7735_WHITE);
    tft.setTextWrap(true);
    tft.print("PARTICLE");    
    SPI.unlock(); 
    delay(300);
}
```

### References

[CH35954]

---

### Completeness

- [ ] Added documentation


hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -50,7 +50,11 @@ LOG_SOURCE_CATEGORY("net.en")
 #endif /* WIZNET_SPI_MODE */
 
 #ifndef WIZNET_SPI_CLOCKDIV
+#if PLATFORM_ID == PLATFORM_BSOM || PLATFORM_ID == PLATFORM_ASOM || PLATFORM_ID == PLATFORM_XSOM
+#define WIZNET_SPI_CLOCKDIV 4
```
avtolstoy: Let's introduce a feature flag in `hal_platform.h` for this and set it for SoM platforms. Something along the lines of `HAL_PLATFORM_WIZNET_SPI_CLOCKDIV`.

hal/inc/hal_dynalib_spi.h
```diff
@@ -55,6 +55,8 @@ DYNALIB_FN(12, hal_spi, HAL_SPI_Set_Callback_On_Select, void(HAL_SPI_Interface,
 DYNALIB_FN(13, hal_spi, HAL_SPI_DMA_Transfer_Cancel, void(HAL_SPI_Interface))
 DYNALIB_FN(14, hal_spi, HAL_SPI_DMA_Transfer_Status, int32_t(HAL_SPI_Interface, HAL_SPI_TransferStatus*))
 DYNALIB_FN(15, hal_spi, HAL_SPI_Set_Settings, int32_t(HAL_SPI_Interface, uint8_t, uint8_t, uint8_t, uint8_t, void*))
+DYNALIB_FN(16, hal_spi, HAL_SPI_Acquire, int32_t(HAL_SPI_Interface, void*))
```
avtolstoy: These are only available on Gen 3 platforms, so let's introduce a feature flag to `hal_platform.h` (e.g. `HAL_PLATFORM_SPI_HAL_THREAD_SAFETY`) and conditionally export these functions for Gen 3 only.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -275,13 +275,18 @@ void HAL_SPI_Init(HAL_SPI_Interface spi) {
 }
 
 void HAL_SPI_Begin(HAL_SPI_Interface spi, uint16_t pin) {
+    HAL_SPI_Acquire(spi, nullptr);
```
avtolstoy: We did a similar thing (guarding each function with acquire/release) for I2C hal previously in order not to break the applications built with previous DeviceOS version, because we've started using PMIC and FuelGauge in power manager.

In case of SPI and Ethernet we've always had this issue, we are not introducing a potential problem, so let's just have proper locking in wiring and require the applications to be rebuilt.

wiring/inc/spark_wiring_spi.h
```diff
@@ -228,14 +228,14 @@ class SPIClass {
   void lock()
   {
 #if PLATFORM_THREADING
-    mutex_.lock();
+    HAL_SPI_Acquire(_spi, NULL);
```
avtolstoy: This should also be conditional on `HAL_PLATFORM_SPI_HAL_THREAD_SAFETY`.

hal/network/lwip/wiznet/wiznetif.cpp
zfields: How about `HAL_PLATFORM_HW_FORM_FACTOR_SOM` instead?

avtolstoy: Nope, but if needed we can introduce if it makes certain things easier. This particular settings IMO should be an explicit value coming from `hal_platform.h` though.

zfields: Should WIZNET really be in the name `HAL_PLATFORM_WIZNET_SPI_CLOCKDIV` or would it be better served to be generalized as `HAL_PLATFORM_SPI_CLOCKDIV`.

Then you could use `HAL_PLATFORM_SPI_CLOCKDIV` and `HAL_PLATFORM_WIZNET_SPI` in conjunction.

avtolstoy: Since this is a setting that specifically applies when we talk to the Wiznet chip, I'd say it should be `HAL_PLATFORM_WIZNET_SPI_CLOCKDIV`. We are not limiting the SPI clock speed in general, we are limiting it just for particular platforms for interaction with the Ethernet featherwing.

zfields: I just keep getting the feeling that we are not abstracting around the right thing if we are calling out WIZNET specifically in _OUR_ `hal_platform.h`.

Wouldn't `HAL_PLATFORM_ETHERNET_FEATHERWING_SPI_CLOCKDIV` make better sense?

Then we could simplify the code in `hal/network/lwip/wiznet/wiznetif.cpp` to become...

```c++
#define WIZNET_SPI_CLOCKDIV HAL_PLATFORM_ETHERNET_FEATHERWING_SPI_CLOCKDIV
```

avtolstoy: Sure, it can be more generic. At the same time, there is no reason to be more specific. You can find specific defines that talk about e.g. BQ24195 PMIC.

wiring/inc/spark_wiring_spi.h
zfields: Now that the Spark Core platform has been removed from the code base, does `#if PLATFORM_THREADING` only block GCC, platform id 3?

Can `PLATFORM_THREADING` just be swapped with `HAL_PLATFORM_SPI_HAL_THREAD_SAFETY`?

avtolstoy: Yup.

hal/network/lwip/wiznet/wiznetif.cpp
zfields: In that case, I would like to implement it as described in the comment above.

avtolstoy: No strong preferences. My choices was merely due to the class/source file being named wiznetxxx, so it kind of makes sense to base the setting on that as well so that it's less confusing.

zfields:
I found a bug with `SPI.tryLock()`. It will always return true, because `SPI` now has an unused `Mutex` member.

For `v1.5.0-rc.1`, should I implement a `HAL_SPI_TryLock()`, promote it through the dynalib and call it from the `SPI.tryLock`, or file a bug and always return `true`? Then implement `SPI.tryLock()` before `v1.5.0`?

zfields:
@avtolstoy

avtolstoy:
> should I implement a HAL_SPI_TryLock()
That should probably be preferred. The main method is called `HAL_SPI_Acquire` though, so TryLock seems somewhat surprising.

Alternatively, `HAL_SPI_Acquire` can be extended to take a struct with e.g. a timeout setting to limit the amount of time wait for exclusive access. `tryLock()` then can be implemented in wiring simply by setting the timeout to 0.

zfields:
> Alternatively, HAL_SPI_Acquire can be extended to take a struct with e.g. a timeout setting to limit the amount of time wait for exclusive access. tryLock() then can be implemented in wiring simply by setting the timeout to 0.

@avtolstoy, I really like this idea! At first I thought you were saying that was already implemented, but since it is not I feel like we should push out this work _(next RC)_ so it can be properly designed, unit-tested and documented _(doxygen-style)_.

For now, lets have it return `true` (as before), so it doesn't prevent our RC release. Does that sound good, or do you think it needs to be part of this RC?

avtolstoy:
If you feel that it's going to take too much time to implement, let's for now stub it, add a FIXME and if we have enough time - fix it. Ideally, we should not be breaking existing functionality.

hal/inc/hal_platform.h
```diff
@@ -207,6 +211,10 @@
 #define HAL_PLATFORM_SPI_DMA_SOURCE_RAM_ONLY (0)
 #endif // HAL_PLATFORM_SPI_DMA_SOURCE_RAM_ONLY
 
+#ifndef HAL_PLATFORM_SPI_HAL_THREAD_SAFETY
+#define HAL_PLATFORM_SPI_HAL_THREAD_SAFETY (0)
```
zfields: Defaulting to `(1)`, true, and disabling for `PLATFORM_GCC` specifically.

zfields: After discussing, we decided to default this value to zero, and enable where appropriate.

wiring/src/spark_wiring_spi.cpp
```diff
@@ -28,162 +28,226 @@
...
+ * \warning This method is NOT THREADSAFE and callers will need to utilize
+ *          HAL synchronization primatives.
+ */
 static void querySpiInfo(HAL_SPI_Interface spi, hal_spi_info_t* info)
 {
-  memset(info, 0, sizeof(hal_spi_info_t));
-  info->version = HAL_SPI_INFO_VERSION_1;
-  HAL_SPI_Info(spi, info, nullptr);
+    memset(info, 0, sizeof(hal_spi_info_t));
+    info->version = HAL_SPI_INFO_VERSION_1;
```
zfields: Why are we using v1 instead of v2?

avtolstoy:
@zfields Just a kind reminder that we should add a test based on `spi_master_slave` to validate these changes. Since the Ethernet Featherwing cannot be used for plugging a second device, let's presume that one needs to manually attach a slave device with jumper wires.

wiring/src/spark_wiring_spi.cpp
```diff
@@ -192,59 +256,126 @@ void SPIClass::computeClockDivider(unsigned reference, unsigned targetSpeed, uin
...
+    uint16_t result;
+    if (!lock())
+    {
+        result = HAL_SPI_Send_Receive_Data(_spi, _data);
+        unlock();
+    }
+    else
+    {
+        // zero is an error code `HAL_SPI_DMA_Transfer_Status`
+        result = 0;
```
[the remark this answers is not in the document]
zfields: I'm accustomed to the explicit statement of return values for each branch, this ensures you have considered each path. Also, the compiler can warn you, if you leave a result branch uninitialized (i.e. unconsidered).

wiring/inc/spark_wiring_spi.h
```diff
@@ -217,24 +224,34 @@ class SPIClass {
...
+    return HAL_SPI_Acquire(_spi, nullptr);
+#elif PLATFORM_THREADING
+    _mutex.lock();
+    return 0;
+#else
+    return 0;
 #endif
   }
 
   void unlock()
```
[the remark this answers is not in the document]
zfields: What would an `unlock()` error look like, and how could a user respond to an `unlock()` error?

wiring/src/spark_wiring_spi.cpp
```diff
@@ -192,59 +256,126 @@ void SPIClass::computeClockDivider(unsigned reference, unsigned targetSpeed, uin
...
+        computeClockDivider(info.system_clock, targetSpeed, divisor, clock);
+
+        // Update SPI peripheral
+        HAL_SPI_Set_Clock_Divider(_spi, divisor);
+        unlock();
+    }
+    else
+    {
+        // Unable to acquire lock
+        clock = 0;
```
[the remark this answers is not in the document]
zfields: I'm accustomed to the explicit statement of return values for each branch, this ensures you have considered each path. Also, the compiler can warn you, if you leave a result branch uninitialized (i.e. unconsidered).

wiring/src/spark_wiring_spi.cpp
```diff
@@ -192,59 +256,126 @@ void SPIClass::computeClockDivider(unsigned reference, unsigned targetSpeed, uin
...
+
+    if (!lock())
+    {
+        result = HAL_SPI_Is_Enabled(_spi);
+        unlock();
+    }
+    else
+    {
+        // Unable to acquire lock
+        result = false;
```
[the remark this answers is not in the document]
zfields: I'm accustomed to the explicit statement of return values for each branch, this ensures you have considered each path. Also, the compiler can warn you, if you leave a result branch uninitialized (i.e. unconsidered).

wiring/inc/spark_wiring_spi.h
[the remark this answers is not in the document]
avtolstoy: I agree, it doesn't make much sense, since potentially the only errors that could be returned are when the lock was not held, or when it was unlocked by a wrong thread, so maybe let's leave it as-is, however for consistency I think it's either
```
int lock();
bool tryLock();
int unlock();
```

or 
```
void lock();
bool tryLock();
void unlock();
```

Let's leave as-is.

wiring/src/spark_wiring_spi.cpp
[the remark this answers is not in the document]
avtolstoy: I'm afraid I will have to insist on this one. It is an error-prone approach to not explicitly initialize variables with trivial types in almost all the cases.

> Also, the compiler can warn you, if you leave a result branch uninitialized (i.e. unconsidered).
We do not promote warnings to errors in user applications, so it is especially dangerous in wiring.

We also did have a lengthy discussion about the usage of this waterfall if/else pattern both during the PR reviews and at one of the guild meetings and collectively came to a decision that we should not use it as breaks RAII usage and decreases readability. It's almost always better to return/error/throw early, instead of having to jump between if/else code blocks to figure out the execution flow. Inability to use `CHECK()` macros in certain cases also does not prevent from returning/erroring out early.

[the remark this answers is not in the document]
zfields: > We do not promote warnings to errors in user applications, so it is especially dangerous in wiring.

I had not considered this ^^^. Without being able to rely on the compiler, I agree this technique is more dangerous than helpful.

avtolstoy:
@zfields The problem was with with the `SPIClass` constructor. We cannot attempt to lock before the HAL initializes the SPI peripheral. `HAL_SPI_Init()` has its own constructs to prevent a race condition when called from different threads.

I've also added a separate test `wiring/spi_master_slave/spi_master_ethernet_concurrency` which sends data over Ethernet while running the normal `spi_master` tests.

It currently fails, so there is still some problem with concurrent access to SPI.

A caveat: when testing, depending on the length of the jumper wires, you might need to increase `HAL_PLATFORM_ETHERNET_FEATHERWING_SPI_CLOCKDIV`.

avtolstoy:
Fixed (see the latest commits). The tests pass. @zfields Please verify that they also pass for you.

PR #1882, opened 2019-08-09
BLE: alternately broadcast user's and Particle-specific advertising d…

XuGuohui:
### Problem

The advertising data and scan response data set by user application will not be broadcasting when device in the Listening mode. What that means is, for custom mobile App it would not be able to filter their device by the advertising data set in user application when device in the Listening mode.

### Solution

Alternately broadcast user's and Particle-specific advertising data when device in the Listening mode. By this way, both of the advertising data will be broadcasting and observed by Particle mobile App and custom mobile App. The only defect is that when device in the listening mode, the advertising and connection parameters set in user application are adjusted to appropriate values to coordinate the Particle mobile setup. But once device exits the Listening mode, these parameters are restored for user application.

Effective advertising parameters in the Listening mode:
```
|<--------------->|<--------------->|<--------------->|<--------------->|
User ADV          Particle ADV      User ADV          Particle ADV 

+----------+---------+
| Interval | Timeout |
+----------+---------+
| 20ms     | 1000ms  |
+----------+---------+
```

Effective connection parameters in the Listening mode:
```
+--------------+--------------+---------+---------+
| MIN Interval | MAX Interval | Latency | Timeout |
+--------------+--------------+---------+---------+
| 30ms         | 50ms         | 0       | 5000ms  |
+--------------+--------------+---------+---------+
```

### Steps to Test

Build and flash the monolithic firmware with log enabled:
    `$ make MODULAR=n DEBUG_BUILD=y PLATFORM=xenon TEST=<path_to_the_example>`
    `$ dfu-util -d 2b04:d00e -a 0 -s 0x30000:leave -D <path_to_the_built_binary>`

1. Device is neither broadcasting nor connected before entering the Listening mode:
    - Check the log message when device in the Listening mode, it should be only broadcasting the Particle-specific advertising data.
    - It should be able to be connected by the Particle mobile App.
    - Make the device leave the listening mode and disconnect from mobile App, device shouldn't be broadcasting.

2. Device is broadcasting before entering the Listening mode:
    - Check the log message when device in the Listening mode, it should be alternately broadcasting user's and Particle-specific advertising data in every 500ms interval.
    - It should be able to be connected by the Particle mobile App.
    - Make the device leave the listening mode and disconnect from mobile App, device should be only broadcasting the user's advertising data with the advertising parameters set in user application.

3. Device is connected before entering the Listening mode:
    - Check the log message when device in the Listening mode, it shouldn't be broadcasting.
    - Disconnect device when device in the listening mode, it should be alternately broadcasting user's and Particle-specific advertising data in every 500ms interval.
    - It should be able to be connected by the Particle mobile App.
    - Make the device leave the listening mode and disconnect from mobile App, device should be only broadcasting the user's advertising data with the advertising parameters set in user application.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
    BLE.setAdvertisingTimeout(1000); // 1000 * 10 ms

    BleAdvertisingData data;
    BLE.advertise(&data);
}

void loop() {
}
```

### DIFF
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -381,7 +381,7 @@ int hal_ble_unlock(void* reserved);
  *
  * @returns    0 on success, system_error_t on error.
  */
-int hal_ble_notify_enter_listening_mode(void* reserved);
+int hal_ble_enter_locked_mode(void* reserved);
 
 /**
  * Notify that device has exited the Listening mode, BLE advertising data/parameters and
@@ -391,7 +391,7 @@ int hal_ble_notify_enter_listening_mode(void* reserved);
  *
  * @returns    0 on success, system_error_t on error.
  */
-int hal_ble_notify_exit_listening_mode(void* reserved);
+int hal_ble_exit_locked_mode(void* reserved);
 
 /**
  * Initialize the BLE stack. This function must be called previous to any other BLE APIs.
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -374,6 +374,26 @@ int hal_ble_lock(void* reserved);
 int hal_ble_unlock(void* reserved);
 
 /**
+ * Notify that device has entered the Listening mode, in which case the BLE
+ * advertising data/parameters and connection parameters are not allowed to be modified.
+ *
+ * @param      reserved  The reserved
+ *
+ * @returns    0 on success, system_error_t on error.
+ */
+int hal_ble_notify_enter_listening_mode(void* reserved);
+
+/**
+ * Notify that device has exited the Listening mode, BLE advertising data/parameters and
+ * connection parameters are allowed to be modified.
+ *
+ * @param      reserved  The reserved
+ *
+ * @returns    0 on success, system_error_t on error.
+ */
+int hal_ble_notify_exit_listening_mode(void* reserved);
+
+/**
  * Initialize the BLE stack. This function must be called previous to any other BLE APIs.
  *
  * @param[in]   reserved    Reserved for future use.
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -410,6 +410,15 @@ int hal_ble_select_antenna(hal_ble_ant_type_t antenna, void* reserved);
 int hal_ble_set_callback_on_adv_events(hal_ble_on_adv_evt_cb_t callback, void* context, void* reserved);
 
 /**
+ * Deregister the callback on BLE advertising events.
+ *
+ * @param[in]   callback    The callback function.
+ *
+ * @returns     0 on success, system_error_t on error.
+ */
+int hal_ble_cancel_callback_on_adv_events(hal_ble_on_adv_evt_cb_t callback, void* context, void* reserved);
+
+/**
  * Set the callback on BLE Peripheral link events.
  *
  * @param[in]   callback    The callback function.
--- a/system/src/ble_listening_mode_handler.cpp
+++ b/system/src/ble_listening_mode_handler.cpp
@@ -296,17 +296,16 @@ int BleListeningModeHandler::exit() {
     // FIXME: BLE is still connected to finalize the mobile setup when this function is called. In that case,
     // there is a chance that the mobile setup process is terminated by other thread or user application.
 
-    preAdvData_.clear();
-    preSrData_.clear();
-    ctrlReqAdvData_.clear();
-    ctrlReqSrData_.clear();
-
     // It should not CHECK() in this function, in case that the BLE HAL status cannot be restored correctly.
 
     // Do not allow other thread to modify the BLE configurations until this function exits.
     hal_ble_lock(nullptr);
     SCOPE_GUARD ({
         hal_ble_unlock(nullptr);
+        preAdvData_.clear();
+        preSrData_.clear();
+        ctrlReqAdvData_.clear();
+        ctrlReqSrData_.clear();
     });
 
     // Now the BLE configurations are modifiable.
--- a/system/src/ble_listening_mode_handler.cpp
+++ b/system/src/ble_listening_mode_handler.cpp
@@ -30,6 +30,7 @@ LOG_SOURCE_CATEGORY("system.listen.ble")
 namespace {
 
 using namespace particle::system;
+using namespace particle::ble;
 
 } // unnamed
 
@@ -93,13 +94,11 @@ int BleListeningModeHandler::cacheUserConfigurations() {
     Vector<uint8_t> tempSrData(BLE_MAX_ADV_DATA_LEN);
 
     // Advertising data set by user application
-    size_t len = hal_ble_gap_get_advertising_data(tempAdvData.data(), BLE_MAX_ADV_DATA_LEN, nullptr);
-    CHECK(len);
+    size_t len = CHECK(hal_ble_gap_get_advertising_data(tempAdvData.data(), BLE_MAX_ADV_DATA_LEN, nullptr));
     tempAdvData.resize(len);
 
     // Scan response data set by user application
-    len = hal_ble_gap_get_scan_response_data(tempSrData.data(), BLE_MAX_ADV_DATA_LEN, nullptr);
-    CHECK(len);
+    len = CHECK(hal_ble_gap_get_scan_response_data(tempSrData.data(), BLE_MAX_ADV_DATA_LEN, nullptr));
     tempSrData.resize(len);
 
     // Advertising and connection parameters set by user application
@@ -123,10 +122,7 @@ int BleListeningModeHandler::restoreUserConfigurations() {
     CHECK_FALSE(exited_, SYSTEM_ERROR_INVALID_STATE);
 
     // Do not allow other thread to modify the BLE configurations.
-    CHECK(hal_ble_lock(nullptr));
-    SCOPE_GUARD ({
-        hal_ble_unlock(nullptr);
-    });
+    BleLock lk;
 
     // Restore the advertising data, advertising parameters and connection parameters.
     CHECK(hal_ble_gap_set_advertising_data(preAdvData_.data(), preAdvData_.size(), nullptr));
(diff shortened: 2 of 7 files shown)


system/src/ble_listening_mode_handler.cpp
```diff
@@ -41,110 +42,99 @@ BleListeningModeHandler::BleListeningModeHandler()
...
 }
 
 BleListeningModeHandler::~BleListeningModeHandler() {
 }
 
-int BleListeningModeHandler::enter() {
-    // TODO: Would it be better to call the method of BLE Control Request Channel to isolate the logic?
-    // Cache the application specific advertising data, advertising parameters and connection parameters.
+int BleListeningModeHandler::constructControlRequestAdvData() {
+    ctrlReqAdvData_.reset(new(std::nothrow) uint8_t[BLE_MAX_ADV_DATA_LEN]);
```
avtolstoy: Let's use the following pattern:

```cpp
{
    auto ctrlReqAdvData = std::make_unique<uint8_t[]>(BLE_MAX_ADV_DATA_LEN);
    CHECK_TRUE(ctrlReqAdvData);
    // ...
    auto ctrlReqSrData = std::make_unique<uint8_t[]>(BLE_MAX_ADV_DATA_LEN);
    CHECK_TRUE(ctrlReqSrData);
    // ...
    // A bunch of other stuff that may return from the function prematurely e.g. due to an error
    // ...
    ctrlReqAdvData_.swap(ctrlReqAdvData);
    ctrlReqSrData_.swap(ctrlReqSrData);
    return SYSTEM_ERROR_NONE;
}
```

Same thing in `cacheUserConfigurations()`.

system/src/ble_listening_mode_handler.cpp
```diff
@@ -41,110 +42,99 @@ BleListeningModeHandler::BleListeningModeHandler()
...
+    ctrlReqSrData_[ctrlReqSrDataLen_++] = BLE_SIG_AD_TYPE_128BIT_SERVICE_UUID_COMPLETE;
+    memcpy(&ctrlReqSrData_[ctrlReqSrDataLen_], BLE_CTRL_REQ_SVC_UUID, sizeof(BLE_CTRL_REQ_SVC_UUID));
+    ctrlReqSrDataLen_ += sizeof(BLE_CTRL_REQ_SVC_UUID);
+    return SYSTEM_ERROR_NONE;
+}
+
+int BleListeningModeHandler::cacheUserConfigurations() {
+    LOG_DEBUG(TRACE, "Cache user's BLE configurations.");
+
+    // Advertising data set by user application
```
avtolstoy: We should acquire the BLE HAL lock for the whole duration of this function, in order to get a snapshot of the application BLE configuration without giving it a chance to change it in-between.

Also, how do we guard the application from changing the configuration while its running concurrently with the listening mode?

system/src/ble_listening_mode_handler.h
```diff
@@ -45,6 +63,12 @@ class BleListeningModeHandler {
     bool preAdvertising_;
     bool preConnected_;
     hal_ble_auto_adv_cfg_t preAutoAdv_;
+    bool userAdv_;
+
+    std::unique_ptr<uint8_t[]> ctrlReqAdvData_;
```
avtolstoy: We could potentially used a `Vector` instead, since it's exactly a pointer + size.

system/src/ble_listening_mode_handler.cpp
```diff
@@ -185,4 +175,91 @@ int BleListeningModeHandler::exit() {
...
+    if (!preAdvertising_ && !preConnected_) {
+        // Start advertising if it is neither connected nor advertising.
+        CHECK(hal_ble_gap_start_advertising(nullptr));
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
+int BleListeningModeHandler::exit() {
+    CHECK(restoreUserConfigurations());
+    CHECK(hal_ble_cancel_callback_on_adv_events(onBleAdvEvents, this, nullptr));
```
avtolstoy: `preAdvData_`, `preSrData_`, `ctrlReqAdvData_` and `ctrlReqSrData_` need to be deallocated here.

system/src/ble_listening_mode_handler.cpp
XuGuohui: If we're going to use `Vector` instead, this should be different.

system/src/ble_listening_mode_handler.cpp
XuGuohui: How about introduce HAL API `hal_ble_notify_enter_listening_mode()` and `hal_ble_notify_exit_listening_mode()`? And if so, these two APIs will not be exported as part of dynalib.

system/src/ble_listening_mode_handler.h
XuGuohui: Sounds good! 👍

system/src/ble_listening_mode_handler.cpp
XuGuohui: In case that we were using `Vector`, the vector should be cleared on exiting listening mode as well.

system/src/ble_listening_mode_handler.cpp
avtolstoy: Same pattern applies to the Vector. The following can also be used:
```cpp
ctrlReqAdvData_ = std::move(ctrlReqAdvData);
```

What I'm trying to suggest here is to fill in the member variables only at the end of the function.

system/src/ble_listening_mode_handler.cpp
avtolstoy: What would they do?

system/src/ble_listening_mode_handler.cpp
XuGuohui: I see. In that case memory allocated in this function will be freed automatically if error occurs before copying the constructed data or cached data and the vector will empty then. 👍

system/src/ble_listening_mode_handler.cpp
XuGuohui: They will set/clear a flag in BLE HAL to indicate whether the advertising data, advertising parameters, connection parameters and other configurations that might be changed during the mobile setup. Besides, we can also make use of this flag to handle some conflicts between the Listening mode and user application, such like `BLE.off()` being called in user application will then remain the peripheral connection if device is in the Listening mode.

XuGuohui: And I think with that HAL APIs implemented, it's unnecessary to acquire/release the BLE HAL lock in this function, since the configurations are guarded these APIs as well. No?

avtolstoy: As long as there is some mechanism in place that makes the current application configuration non-modifiable, yes.

Let's also make sure that we don't forget to call `hal_ble_notify_exit_listening_mode()` in case something goes wrong (e.g. memory allocation fails etc).

XuGuohui:
Merge #1890

system/src/ble_listening_mode_handler.h
```diff
@@ -52,23 +53,24 @@ class BleListeningModeHandler {
...
+    Vector<uint8_t> preSrData_;
     hal_ble_adv_params_t preAdvParams_;
+    int8_t preTxPower_;
     hal_ble_conn_params_t prePpcp_;
     bool preAdvertising_;
     bool preConnected_;
     hal_ble_auto_adv_cfg_t preAutoAdv_;
     bool userAdv_;
+    bool restoreUserConfig_;
+    static bool exited_;
```
avtolstoy: Does this need to be `static`? I don't think we are destroying this object ever.

hal/inc/ble_hal.h
```diff
@@ -373,6 +373,26 @@ int hal_ble_lock(void* reserved);
...
 
+/**
+ * Notify that device has entered the Listening mode, in which case the BLE
+ * advertising data/parameters and connection parameters are not allowed to be modified.
+ *
+ * @param      reserved  The reserved
+ *
+ * @returns    0 on success, system_error_t on error.
+ */
+int hal_ble_notify_enter_listening_mode(void* reserved);
```
avtolstoy: Let's perhaps rename this to something more generic. HAL ideally shouldn't be aware of high-level concepts like listening mode. Something generic like `hal_ble_enter_locked_mode()` (not ideal, but naming things is difficult) sounds better IMO.

system/src/ble_listening_mode_handler.cpp
```diff
@@ -34,112 +34,103 @@ using namespace particle::system;
...
 
     // Advertising data set by user application
-    preAdvDataLen_ = hal_ble_gap_get_advertising_data(nullptr, 0, nullptr);
-    CHECK(preAdvDataLen_);
-    if (preAdvDataLen_ > 0) {
-        preAdvData_.reset(new(std::nothrow) uint8_t[preAdvDataLen_]);
-        CHECK_TRUE(preAdvData_, SYSTEM_ERROR_NO_MEMORY);
-        hal_ble_gap_get_advertising_data(preAdvData_.get(), preAdvDataLen_, nullptr);
-    }
+    size_t len = hal_ble_gap_get_advertising_data(tempAdvData.data(), BLE_MAX_ADV_DATA_LEN, nullptr);
```
avtolstoy: `ssize_t` or even better:

```cpp
size_t len = CHECK(hal_ble_gap_get_advertising_data(tempAdvData.data(), BLE_MAX_ADV_DATA_LEN, nullptr));
```

avtolstoy: Same in https://github.com/particle-iot/device-os/pull/1882/commits/fc3555e8e938d92fd5f954ae3665d11d5e968e25#diff-79e0c2ab765d993495c5c34f5f338f74R101

system/src/ble_listening_mode_handler.cpp
```diff
@@ -124,6 +124,9 @@ int BleListeningModeHandler::restoreUserConfigurations() {
 
     // Do not allow other thread to modify the BLE configurations.
     CHECK(hal_ble_lock(nullptr));
+    SCOPE_GUARD ({
```
avtolstoy: Let's instead move `WiringBleLock` out of cpp file into the header and utilize RAII.

system/src/ble_listening_mode_handler.cpp
```diff
@@ -198,49 +203,125 @@ int BleListeningModeHandler::applyControlRequestConfigurations() {
...
+    if (restoreUserConfig_) {
+        restoreUserConfig_ = false;
+        if (restoreUserConfigurations() != SYSTEM_ERROR_NONE) {
+            LOG(ERROR, "Failed to restore user configuration.");
+        }
+    }
+    hal_ble_cancel_callback_on_adv_events(onBleAdvEvents, this, nullptr);
+    hal_ble_unlock(nullptr);
+
+    exited_ = true;
```
avtolstoy: Let's set if before unlocking the BLE lock.

system/src/ble_listening_mode_handler.h
XuGuohui: Just in case, unless it has side-effects.

system/src/ble_listening_mode_handler.cpp
XuGuohui: If we used `BleLock`, it is always set before unlocking the BLE lock.

PR #1890, opened 2019-08-22
BLE: introduces BLE.begin() and BLE.end() wiring APIs.

XuGuohui:
### Problem

In Device OS 1.3.1-rc.1 and below, to terminate all the BLE activities initiated by user application, we can call `BLE.off()` to just shut off the BLE radio. While this API will also prevent the mobile setup from going further when device is in the Listening mode, since BLE will stop advertising and Peripheral connection is dropped if connected when that API is called.

### Solution

By introducing the `BLE.end()` API, it makes sure that only the BLE activities initiated by user application is terminated, leaving the mobile setup works as expected.

### Examples App

1. BLE scanner
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void bleOnScanResultCallback(const BleScanResult* result, void* context) {
    LOG(TRACE, "---------- END ----------");
    BLE.end()
}

void setup() {
}

void loop() {
    BLE.scan(bleOnScanResultCallback, nullptr);
    delay(3000);
}
```
2. BLE Peripheral
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup() {
    BleAdvertisingData data;
    BLE.advertise(&data);
}

void loop() {
    delay(20000);
    LOG(TRACE, "---------- END ----------");
    BLE.end()
}
```
3. BLE Central
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

#define SCAN_RESULT_COUNT       20

BlePeerDevice peer;

void setup() {
}

void loop() {
    if (peer.connected()) {
        delay(10000);
        LOG(TRACE, "---------- END ----------");
        BLE.end()
    }
    else {
        delay(3000);
        if (count > 0) {
                BleUuid foundServiceUUID;
                if (svcCount > 0 && foundServiceUUID == "6E400001-B5A3-F393-E0A9-E50E24DCCA9E") {
                    peer = BLE.connect(results[i].address);
                    break;
                }
            }
        }
    }
}
```

### Steps to Test

1. Build and flash the 1st application, see if BLE stops scanning on device scanned.
2. Build and flash the 2nd application:
    - See if BLE stops advertising after 20 seconds.
    - On device started, make it enter the Listening mode, see if BLE is still advertising after 20 seconds.
    - On device started, make it enter the Listening mode and connect to the device using a mobile App, see if BLE connection is retained after 20 seconds.
3. Build and flash the 3rd application (need another device that is running the 1st application but without the `BLE.end()` is called), see if BLE connection is dropped after device being connected for 10 seconds.

### References

Known issue: https://github.com/particle-iot/device-os/pull/1847

### DIFF
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -1609,6 +1609,7 @@ int BleLocalDevice::end() const {
 
 int BleLocalDevice::on() const {
     WiringBleLock lk;
+    CHECK(hal_ble_stack_init(nullptr));
     return SYSTEM_ERROR_NONE;
 }
 
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -1583,12 +1583,36 @@ void BleLocalDevice::onDisconnected(BleOnDisconnectedCallback callback, void* co
     impl()->onDisconnectedCallback(callback, context);
 }
 
-int BleLocalDevice::on() {
+int BleLocalDevice::begin() const {
     WiringBleLock lk;
     return SYSTEM_ERROR_NONE;
 }
 
-int BleLocalDevice::off() {
+int BleLocalDevice::end() const {
+    /*
+     * 1. Disconnects all the connections initiated by user application.
+     * 2. Disconnects Peripheral connection if it is not in the Listening mode.
+     * 3. Stops advertising if it is not in the Listening mode.
+     * 4. Stops scanning if initiated.
+     *
+     * FIXME: If device is broadcasting before entering the Listening mode and
+     * then this API is called during device in the Listening mode, device will
+     * restart broadcasting automatically when device exits the Listening mode.
+     */
+    WiringBleLock lk;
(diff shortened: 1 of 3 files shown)


wiring/src/spark_wiring_ble.cpp
```diff
@@ -1583,12 +1583,36 @@ void BleLocalDevice::onDisconnected(BleOnDisconnectedCallback callback, void* co
...
+    WiringBleLock lk;
+    disconnectAll(); // BLE HAL will guard that the Peripheral connection is remained if device is in the Listening mode.
+    impl()->peers().clear();
+    stopAdvertising(); // BLE HAL will guard that device keeps broadcasting if device is in the Listening mode.
+    stopScanning();
+    return SYSTEM_ERROR_NONE;
+}
+
+int BleLocalDevice::on() const {
+    WiringBleLock lk;
```
avtolstoy: `hal_ble_stack_init()`?

avtolstoy: Shouldn't we also call it when entering listening mode?

Issue #1888, opened 2019-08-16
onDataReceived callback not invoked on central


sviraaj:
<details>
  <summary><i>submission notes</i></summary>

```
**Important:** Please sanitize/remove any confidential info like usernames, passwords, org names, product names/ids, access tokens, client ids/secrets, or anything else you don't wish to share.

Please choose one of the following templates **Bug Report** or **Feature/Enhancement Request** and delete the other.

You may also delete this submission notes header if you'd like. Thank you for contributing!
```
</details>

## Bug Report
Xenon working as BLE central and it connects to the peripheral without issues after the service is detected. however, when the indication is received, the callback is not being invoked and it is also not acknowledging the indication to BLE peripheral. Working on "v1.3.0" branch.

### Expected Behavior
When indication received, data received callback should be called.

### Observed Behavior

Callback is not invoked. No indication confirmed received on peripheral.

### Steps to Reproduce

Run the below code on xenon, and connect to a peripheral which has a temperature service with the temperature measurement characteristic with UUID as used in code. Check if the temp  measurement characteristic indication from peripheral invokes the callback on xenon central.

### Test App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);

SerialLogHandler logHandler(LOG_LEVEL_ALL);

const size_t SCAN_RESULT_MAX = 30;

BleCharacteristic temperatureMeasurementCharacteristic;

BleScanResult scanResults[SCAN_RESULT_MAX];
BlePeerDevice peer;

void onDataReceivedTemp(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context);

void setup()
{
    (void)logHandler; // Does nothing, just to eliminate the unused variable warning

    temperatureMeasurementCharacteristic.onDataReceived(onDataReceivedTemp, NULL);
}

void loop()
{
    if (BLE.connected())
    {
    }
    else
    {
        Log.info("BLE scan!");
        int count = BLE.scan(scanResults, SCAN_RESULT_MAX);

        if (count)
        {
            Log.info("%d devices found", count);
        }

        for (int ii = 0; ii < count; ii++)
        {
            uint8_t buf[BLE_MAX_ADV_DATA_LEN];
            size_t len;

            // We're looking for devices that have a temperature service (0x1809)
            len = scanResults[ii].advertisingData.get(
                BleAdvertisingDataType::SERVICE_UUID_16BIT_COMPLETE, buf, BLE_MAX_ADV_DATA_LEN);
[...]

PR #1891, opened 2019-08-22
BLE: minor fixes

XuGuohui:
### Bugfixes

- Restored default BLE device address is incorrect.
- Read BLE device name might be contracted.
- `operator&` of the `BleCharacteristicProperty` enum class doesn't work as expected.
- The length of got advertising and scan response data is not updated.

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -974,9 +974,12 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
     CHECK(suspend());
     if (connHandle_ != BLE_INVALID_CONN_HANDLE) {
         tempParams.type = BLE_ADV_SCANABLE_UNDIRECTED_EVT;
-        if (configure(&tempParams) != SYSTEM_ERROR_NONE) {
-            LOG(ERROR, "Failed to configure advertising parameters.");
-            return resume();
+        int ret = configure(&tempParams);
+        if (ret != SYSTEM_ERROR_NONE) {
+            if (resume() != SYSTEM_ERROR_NONE) {
+                LOG(ERROR, "Failed to resume BLE advertising.");
+            }
+            return ret;
         }
         if (params == nullptr) {
             tempParams.type = BLE_ADV_CONNECTABLE_SCANNABLE_UNDIRECRED_EVT;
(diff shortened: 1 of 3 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -979,7 +981,9 @@ int BleObject::Broadcaster::setAdvertisingData(const uint8_t* buf, size_t len) {
         len = 0;
     }
     advDataLen_ = len;
-    configure(nullptr);
+    if (configure(nullptr) != SYSTEM_ERROR_NONE) {
```
avtolstoy: Should this error be propagated?

avtolstoy: Same in https://github.com/particle-iot/device-os/pull/1891/files#diff-fba7ee4853a5cf25cf4e9837d8c5f803R1009

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1125,7 +1131,8 @@ int BleObject::Broadcaster::configure(const hal_ble_adv_params_t* params) {
                   bleGapAdvParams.interval*0.625, bleGapAdvParams.duration*10);
         ret = sd_ble_gap_adv_set_configure(&advHandle_, &bleGapAdvData, &bleGapAdvParams);
     }
-    return nrf_system_error(ret);
+    CHECK(nrf_system_error(ret));
```
avtolstoy: Any reasoning for the change? :)

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1728,15 +1733,23 @@ ssize_t BleLocalDevice::getAdvertisingData(BleAdvertisingData* advertisingData)
     if (advertisingData == nullptr) {
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
-    return hal_ble_gap_get_advertising_data(advertisingData->data(), BLE_MAX_ADV_DATA_LEN, nullptr);
+    advertisingData->clear();
+    ssize_t len = hal_ble_gap_get_advertising_data(advertisingData->data(), BLE_MAX_ADV_DATA_LEN, nullptr);
```
avtolstoy: Prefer
```cpp
size_t len = CHECK(hal_ble_gap_get_advertising_data(advertisingData->data(), BLE_MAX_ADV_DATA_LEN, nullptr));
```

Same in https://github.com/particle-iot/device-os/pull/1891/files#diff-93f9dd9012bb15fde13e60440775ab25R1749

avtolstoy:
Please also indicate any GH issues this PR closes.

hal/src/nRF52840/ble_hal.cpp
XuGuohui: Because it won't log any error if failed to configure advertising set without this `CHECK()`.

XuGuohui: Also it's fine to log the error where the `configure()` is called.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -954,15 +957,19 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
...
+                LOG(ERROR, "Failed to resume BLE advertising.");
+            }
+            return ret;
         }
         memcpy(&advParams_, &tempParams, std::min(advParams_.size, tempParams.size));
         advParams_.size = sizeof(hal_ble_adv_params_t);
         advParams_.version = BLE_API_VERSION;
     }
-    return resume();
+    CHECK(resume());
```
avtolstoy: Let's perhaps use `return CHECK(resume())`. Seems more concise.

XuGuohui:
Fixes #1888

PR #1895, opened 2019-08-27
Expose select() and poll() to apps

larseggert:
### Problem

`select` and `poll` system calls are not exposed to applications

### Solution

Patch exposes `select` and `poll` via dynalib

---

### Completeness

- [x] Added documentation


hal/inc/socket_hal_posix.h
```diff
@@ -40,6 +40,8 @@ extern "C" {
  *
  */
 
+/** Compatibility nfds_t */
+typedef unsigned int nfds_t;
```
avtolstoy: We shouldn't be declaring this here. The compatibility types in this section are only for compatibility with non-POSIX socket hal.

`nfds_t` should be coming from the implementation header `socket_hal_posix_impl.h` and in turn from `lwip/sockets.h`.

hal/inc/socket_hal_posix.h
```diff
@@ -258,6 +260,11 @@ int sock_socket(int domain, int type, int protocol);
  */
 int sock_fcntl(int s, int cmd, ...);
 
+int sock_poll(struct pollfd* fds, nfds_t nfds, int timeout);
```
avtolstoy: Please also add doxygen documentation, since the rest of the functions in this header file are documented.

hal/inc/socket_hal_posix.h
```diff
@@ -258,6 +260,11 @@ int sock_socket(int domain, int type, int protocol);
...
+
+int sock_select(int maxfdp1, fd_set* readset, fd_set* writeset,
```
avtolstoy: Let's use more standard definition:
```cpp
int lwip_select(int nfds, fd_set* readfds, fd_set* writefds,
                fd_set* exceptfds, struct timeval* timeout);
```

hal/network/lwip/socket_hal.cpp
```diff
@@ -94,3 +94,14 @@ int sock_fcntl(int s, int cmd, ...) {
   va_end(vl);
   return lwip_fcntl(s, cmd, val);
 }
+
+int sock_poll(struct pollfd* fds, nfds_t nfds, int timeout) {
+  int ret = lwip_poll(fds, nfds, timeout);
```
avtolstoy: Should be fine to simply `return lwip_poll(fds, nfds, timeout);`. Same in https://github.com/particle-iot/device-os/pull/1895/files#diff-a0762610abcc5f13d78b64aeb1f49facR105

hal/inc/socket_hal_posix.h
larseggert: OK, so should I include that header instead?

larseggert: Done

hal/inc/socket_hal_posix.h
larseggert: Done

hal/inc/socket_hal_posix.h
larseggert: Done

hal/inc/socket_hal_posix.h
avtolstoy: It is already included in `socket_hal_posix_impl.h`, so in theory you shouldn't need to add anything at all unless for some reason `POLLIN` or `POLLOUT` are defined by newlib (I don't think they are): https://github.com/particle-iot/lwip/blob/particle/src/include/lwip/sockets.h#L506

larseggert:
@avtolstoy I think I addressed your review comments

avtolstoy:
@larseggert I've approved and scheduled this for 1.4.0. One last suggestion I'd make is to perhaps remove empty merge commits (https://github.com/particle-iot/device-os/pull/1895/commits/ed6d3643480a86459d5602f9e1b597522efe2c73 and https://github.com/particle-iot/device-os/pull/1895/commits/fde938f082ddd2532e66bfd76c0cf99db23ba029) if possible in order to keep the history clean.

PR #1897, opened 2019-08-29
Fixes SoftDevice OTA updates

avtolstoy:
### Problem

1. The bootloader doesn't correctly handle `MODULE_INFO_FLAG_DROP_MODULE_INFO` flag. The returned module info for the SoftDevice module is a copy into a static variable, so direct pointer comparisons can't work.
2. The state checksum doesn't change after an update of either SoftDevice or NCP module, which causes the device to get stuck in safe mode

### Solution

1. Use `memcmp()` instead
2. Add a dummy suffix structure to module info and fill SHA for NCP and SoftDevice modules

**This should probably be backported into 1.3.0 default**

### Steps to Test

1. Flash the bootloader and then flash the SoftDevice binary OTA. It should apply correctly.
2. TBD on staging.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/argon/platform_ncp_argon.cpp
+++ b/hal/src/argon/platform_ncp_argon.cpp
@@ -117,12 +117,33 @@ int platform_ncp_fetch_module_info(hal_system_info_t* sys_info, bool create) {
                 info->platform_id = PLATFORM_ID;
                 info->module_function = MODULE_FUNCTION_NCP_FIRMWARE;
 
-                module->info = info;
                 // assume all checks pass since it was validated when being flashed to the NCP
                 module->validity_result = module->validity_checked;
+
+                // IMPORTANT: a valid suffix with SHA is required for the communication layer to detect a change
+                // in the SYSTEM DESCRIBE state and send a HELLO after the NCP update to
+                // cause the DS to request new DESCRIBE info
+                auto suffix = new module_info_suffix_t();
+                if (!suffix) {
+                    delete info;
+                    return SYSTEM_ERROR_NO_MEMORY;
+                }
+                memset(suffix, 0, sizeof(module_info_suffix_t));
+
+                // FIXME: NCP firmware should return some kind of a unique string/hash
+                // For now we simply fill the SHA field with version
+                for (uint16_t* sha = (uint16_t*)suffix->sha;
+                        sha < (uint16_t*)(suffix->sha + sizeof(suffix->sha));
+                        ++sha) {
+                    *sha = version;
+                }
+
+                module->info = info;
+                module->suffix = suffix;
             }
             else {
                 delete module->info;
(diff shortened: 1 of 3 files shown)


hal/src/argon/platform_ncp_argon.cpp
```diff
@@ -117,12 +117,33 @@ int platform_ncp_fetch_module_info(hal_system_info_t* sys_info, bool create) {
...
+
+                // FIXME: NCP firmware should return some kind of a unique string/hash
+                // For now we simply fill the SHA field with version
+                for (uint16_t* sha = (uint16_t*)suffix->sha;
+                        sha < (uint16_t*)(suffix->sha + sizeof(suffix->sha));
+                        ++sha) {
+                    *sha = version;
+                }
+
+                module->info = info;
```
XuGuohui: Are the `info` and `suffix` supposed to be released somewhere?

avtolstoy: Yes, in this function in the `else` condition.

XuGuohui: So this function is supposed to be called again later on to free the memory by passing `false` as the `create` argument. If so, why don't we implement a separated API to explicitly delete the module info. Or at least add a note that this memory will be freed wherever the module info is no longer used.

avtolstoy: While I agree that this could be refactored and documented to make things clearer, this is not new functionality and follows the same pattern as https://github.com/particle-iot/device-os/blob/develop/hal/src/nRF52840/ota_flash_hal.cpp#L155 (which is somewhat documented https://github.com/particle-iot/device-os/blob/develop/hal/inc/ota_flash_hal.h#L96) and is not the main purpose of this PR.

hal/src/argon/platform_ncp_argon.cpp
```diff
@@ -117,12 +117,33 @@ int platform_ncp_fetch_module_info(hal_system_info_t* sys_info, bool create) {
...
+                auto suffix = new module_info_suffix_t();
+                if (!suffix) {
+                    delete info;
+                    return SYSTEM_ERROR_NO_MEMORY;
+                }
+                memset(suffix, 0, sizeof(module_info_suffix_t));
+
+                // FIXME: NCP firmware should return some kind of a unique string/hash
+                // For now we simply fill the SHA field with version
+                for (uint16_t* sha = (uint16_t*)suffix->sha;
```
m-mcgowan: any reason why we don't append the module suffix to the NCP firmware as part of the build?

avtolstoy: The NCP firmware resides in the NCP flash memory which we don't have access directly to. The FIXME here still stands: we need some kind of an AT command that would return it for us.

PR #1898, opened 2019-08-30
Fixes issues encountered during 1.4.0 rc.1 testing

avtolstoy:
### Overview

This PR:

1. **CRITICAL:** Fixes session resume broken after an upgrade of MbedTLS. One of the internal fields in the `mbedtls_ssl_context` changed its meaning.
2. MbedTLS logging was not re-enabled on all the platforms. `#define MBEDTLS_DEBUG_C` was added to platform configs where it was commented out.
3. Fixes `CellularSignal` `rssi` and `qual` compatibility with Gen 2 on Gen 3 platforms. Fixes `CELLULAR_07_rssi_is_valid` test
4. Removes a rogue `$(error)` call in tinker makefile when building with `LOG_SERIAL=y`
5. **CRITICAL:** fixes heap and application static RAM overlap. This was the cause of `SYSTEM_07_fragmented_heap` stress test failure and who knows how many of the other weird Gen 3 issues. **This needs be backported into 1.3.1 default** as we broke things in 1.3.0-rc.1 in https://github.com/particle-iot/device-os/pull/1819
6. `MDM_01_socket_writes_with_length_more_than_1023_work_correctly`, `MDM_02_at_commands_with_long_response_are_correctly_parsed_and_flow_controlled` enabled for non-LTE Gen 3 devices. `MDM_03_restore_cloud_connections` added
7. `MESH_01_PublishWithoutSubscribeStillReadsDataOutOfPubSubSocket` is automatically skipped if there are no Mesh credentials on the device
8. Relaxes `APPLICATION_WATCHDOG_02_doesnt_fire_when_app_checks_in` and `APPLICATION_WATCHDOG_03_doesnt_leak_memory` slightly
9. Strengthens `SYSTEM_07_fragmented_heap` to also fill the allocated blocks with random data to more easily indicate potential issues with heap/static RAM overlap and heap corruptions.
10. Adds `SYSTEM_09_out_of_memory_restore_state` test.
11. Disconnects cloud and network during PWM tests
12. Removes DNS and DHCP server checks in `WIFI_04_config` for Gen 3 devices

### Steps to Test

Run `wiring/no_fixture` on Gen 2 and Gen 3 platforms.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/boron/cellular_hal.cpp
+++ b/hal/src/boron/cellular_hal.cpp
@@ -284,8 +284,54 @@ int cellular_signal(CellularSignalHal* signal, cellular_signal_t* signalExt) {
     const auto strn = s.strength();
     const auto qual = s.quality();
     if (signal) {
-        signal->rssi = strn;
-        signal->qual = qual;
+        // Compatibility with Gen 2
+        if (strn != 99 && strn != 255) {
+            int compatStrn = strn;
+            switch (s.strengthUnits()) {
+            case CellularStrengthUnits::RXLEV: {
+                // Leave as-is
+                break;
+            }
+            case CellularStrengthUnits::RSCP: {
+                // Simply re-map from [0-96] to [0-63]
+                compatStrn = (compatStrn * 63) / 96;
+                break;
+            }
+            case CellularStrengthUnits::RSRP: {
+                // Simply remap from [0-97] to [0-63]
+                compatStrn = (compatStrn * 63) / 97;
+                break;
+            }
+            }
+            // -113 to -50dBm
+            signal->rssi = -113 + compatStrn;
+        }
+        // see 3GPP TS 45.008 [20] subclause 8.2.4
+        static const char compatQualMap[] = { 49, 43, 37, 25, 19, 13, 7, 0 };
+        if (qual != 99 && qual != 255) {
+            int compatQual = qual;
+            switch (s.qualityUnits()) {
+            case CellularQualityUnits::RXQUAL:
+            case CellularQualityUnits::MEAN_BEP: {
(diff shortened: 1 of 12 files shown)


hal/src/boron/cellular_hal.cpp
```diff
@@ -284,8 +284,54 @@ int cellular_signal(CellularSignalHal* signal, cellular_signal_t* signalExt) {
...
+                // Re-map from [0-34] to [0-7]. Table in UBX-13002752 - R62 (7.2.4)
+                compatQual = (compatQual < 10) ? (compatQual / 5) : ((std::min(compatQual, 30) - 10) / 4) + 2;
+                break;
+            }
+            }
+            // Just in case validate that we are not going to go out of bounds
+            if (compatQual >= 0 && compatQual <= 7) {
+                signal->qual = compatQualMap[compatQual];
+            }
+        }
```
technobly: Since we already calculate the signal/qual in signalExt, could we fix this with less code and more accuracy by adding `signal->rssi = signalExt->rsrp / 100;` and `signal->qual = signalExt->rsrq / 100;` in `cellular_hal.cpp`?  I would like to make sure we are not calculating two different things for the Cellular.RSSI() API and Vitals.  Also there is a ch35974 for this.

avtolstoy: We can't. This is exactly what is needed to provide the compatibility with what's available on Electron. Going forward we simply need to encourage the usage of https://docs.particle.io/reference/device-os/firmware/boron/#cellularsignal-class, which is the same thing that the vitals are using.

PR #1901, opened 2019-09-04
BLE: user can optionally send characteristic value with/without ACK.

XuGuohui:
### Problem

For local characteristic with `NOTIFY` and `INDICATE` property, it will only send notification even if peer device subscribed indication.

For peer characteristic with `NOTIFY` and `INDICATE` property, it will only subscribe notification.

For peer characteristic with `WRITE` and `WRITE_WO_RSP` property, it will only send data without response.

### Solution

- Introduces `BleCharacteristic::setValue(..., BleTxRxType)` API for user to optionally send data with/without acknowledgement.
- Subscribes both notification and indication if both properties are presented with peer characteristic.

### Steps to Test

1. Build and flash the below example app to two Gen3 devices respectively.
2. Connect two devices to serial terminal respectively.
3. Simply type '1', '2', and '3' on one side terminal to see if another terminal can successfully receive these characters, and vice versa.

### Example App
BLE Peripheral:
```c
#include "Particle.h"

#define UART_TX_BUF_SIZE        20

SYSTEM_MODE(MANUAL);

void onDataReceived(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context);

const char* rxUuid = "6E400002-B5A3-F393-E0A9-E50E24DCCA9E";
const char* txUuid = "6E400003-B5A3-F393-E0A9-E50E24DCCA9E";

BleCharacteristic txCharacteristic("tx",
                                   BleCharacteristicProperty::NOTIFY | BleCharacteristicProperty::INDICATE,
                                   txUuid,
                                   serviceUuid);

BleCharacteristic rxCharacteristic("rx",
                                   BleCharacteristicProperty::WRITE_WO_RSP | BleCharacteristicProperty::WRITE,
                                   rxUuid,
                                   serviceUuid,
                                   onDataReceived, &rxCharacteristic);

uint8_t txBuf[UART_TX_BUF_SIZE];
size_t txLen = 0;

    for (uint8_t i = 0; i < len; i++) {
        Serial.write(data[i]);
    }
}

void setup() {
    Serial.begin(115200);

    BLE.addCharacteristic(rxCharacteristic);

    BleAdvertisingData data;
    BLE.advertise(&data);
}

void loop() {
    if (BLE.connected()) {
        while (Serial.available() && txLen < UART_TX_BUF_SIZE) {
            txBuf[txLen++] = Serial.read();
            Serial.write(txBuf[txLen - 1]);
        }

        if (txLen > 0) {
            if (txBuf[0] == '1') {
                txCharacteristic.setValue(txBuf, txLen);
            } else if (txBuf[0] == '2') {
                txCharacteristic.setValue(txBuf, txLen, BleTxRxType::ACK);
            } else {
                txCharacteristic.setValue(txBuf, txLen, BleTxRxType::NACK);
            }
            txLen = 0;
        }
    }
}
```
BLE Central:
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

#define UART_TX_BUF_SIZE        20
#define SCAN_RESULT_COUNT       20

BleCharacteristic peerTxCharacteristic;
BleCharacteristic peerRxCharacteristic;
BlePeerDevice peer;

uint8_t txBuf[UART_TX_BUF_SIZE];
size_t txLen = 0;

    for (uint8_t i = 0; i < len; i++) {
        Serial.write(data[i]);
    }
}

void setup() {
    Serial.begin(115200);
    peerTxCharacteristic.onDataReceived(onDataReceived, &peerTxCharacteristic);
}

void loop() {
    if (BLE.connected()) {
        while (Serial.available() && txLen < UART_TX_BUF_SIZE) {
            txBuf[txLen++] = Serial.read();
            Serial.write(txBuf[txLen - 1]);
        }
        if (txLen > 0) {
            if (txBuf[0] == '1') {
                peerRxCharacteristic.setValue(txBuf, txLen);
            } else if (txBuf[0] == '2') {
                peerRxCharacteristic.setValue(txBuf, txLen, BleTxRxType::ACK);
            } else {
                peerRxCharacteristic.setValue(txBuf, txLen, BleTxRxType::NACK);
            }
            txLen = 0;
        }
    }
    else {
        delay(3000);
        if (count > 0) {
                BleUuid foundServiceUUID;
                if (svcCount > 0 && foundServiceUUID == "6E400001-B5A3-F393-E0A9-E50E24DCCA9E") {
                    peer = BLE.connect(results[i].address);
                    if (peer.connected()) {
                        peer.getCharacteristicByDescription(peerTxCharacteristic, "tx");
                        peer.getCharacteristicByUUID(peerRxCharacteristic, "6E400002-B5A3-F393-E0A9-E50E24DCCA9E");
                    }
                    break;
                }
            }
        }
    }
}
```

### References

N/A


hal/inc/ble_hal_defines.h
```diff
@@ -270,9 +270,10 @@ typedef enum ble_sig_char_prop_t {
 /**@defgroup BLE Characteristic properties
  * @{ */
 typedef enum ble_sig_cccd_value_t {
-    BLE_SIG_CCCD_VAL_DISABLED     = 0x00, /**< Neither notification nor indication is enabled. */
-    BLE_SIG_CCCD_VAL_NOTIFICATION = 0x01, /**< Notification is enabled. */
-    BLE_SIG_CCCD_VAL_INDICATION   = 0x02 /**< Indication is enabled. */
+    BLE_SIG_CCCD_VAL_DISABLED     = 0x0000, /**< Neither notification nor indication is enabled. */
+    BLE_SIG_CCCD_VAL_NOTIFICATION = 0x0001, /**< Notification is enabled. */
+    BLE_SIG_CCCD_VAL_INDICATION   = 0x0002, /**< Indication is enabled. */
+    BLE_SIG_CCCD_VAL_NOTI_IND     = 0x0003  /**< Both Notification and indication are enabled. */
```
avtolstoy: If the intent is to make `ble_sig_cccd_value_t` at least 16-bits, we should add `BLE_SIG_CCCD_VAL_MAX = 0x7fff`

wiring/inc/spark_wiring_ble.h
```diff
@@ -406,6 +417,17 @@ class BleCharacteristic {
...
 
+    template<typename T>
+    typename std::enable_if<std::is_integral<T>::value, ssize_t>::type
+    setValue(T val, bool ack) {
+        uint8_t buf[BLE_MAX_ATTR_VALUE_PACKET_SIZE];
+        size_t len = std::min(sizeof(T), (unsigned)BLE_MAX_ATTR_VALUE_PACKET_SIZE);
+        for (size_t i = 0, j = len - 1; i < len; i++, j--) {
+            buf[i] = reinterpret_cast<const uint8_t*>(&val)[j];
+        }
+        return setValue(buf, len, ack);
```
avtolstoy: What's the problem with simple `setValue(reinterpret_cast<const uint8_t*>(&val), std::min(sizeof(T), BLE_MAX_ATTR_VALUE_PACKET_SIZE), ack)` ?

hal/inc/ble_hal_defines.h
avtolstoy: This will also affect `hal_ble_cccd_config_evt_t` and `hal_ble_cccd_config_t`.

wiring/inc/spark_wiring_ble.h
```diff
@@ -395,9 +402,13 @@ class BleCharacteristic {
...
+    ssize_t setValue(const uint8_t* buf, size_t len, bool ack);
+    ssize_t setValue(const String& str, bool ack);
+    ssize_t setValue(const char* str, bool ack);
 
     template<typename T>
-    ssize_t setValue(T val) {
+    typename std::enable_if<std::is_integral<T>::value, ssize_t>::type
+    setValue(T val) {
         uint8_t buf[BLE_MAX_ATTR_VALUE_PACKET_SIZE];
         size_t len = std::min(sizeof(T), (unsigned)BLE_MAX_ATTR_VALUE_PACKET_SIZE);
```
avtolstoy: This also doesn't seem necessary for _integral_ types as they are definitely less than ~250 bytes. And if even if they were larger, it makes more sense not to send anything at all, rather that send up partial data. Same in other places like https://github.com/particle-iot/device-os/pull/1901/files#diff-75b7ed256204b5221da93d699fee0429R422

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1202,14 +1255,16 @@ int BleCharacteristic::subscribe(bool enable) const {
...
     config.cccd_handle = impl()->attrHandles().cccd_handle;
     config.value_handle = impl()->attrHandles().value_handle;
+    config.cccd_value = BLE_SIG_CCCD_VAL_DISABLED;
     if (enable) {
-        if ((impl()->properties() & BleCharacteristicProperty::NOTIFY) == BleCharacteristicProperty::NOTIFY) {
-            config.cccd_value = BLE_SIG_CCCD_VAL_NOTIFICATION;
-        } else {
+        if ((impl()->properties() & BleCharacteristicProperty::NOTIFY) == BleCharacteristicProperty::NOTIFY &&
+                (impl()->properties() & BleCharacteristicProperty::INDICATE) == BleCharacteristicProperty::INDICATE) {
+            config.cccd_value = BLE_SIG_CCCD_VAL_NOTI_IND;
```
avtolstoy: Nitpicking: could be done with `|`.

wiring/inc/spark_wiring_ble.h
```diff
@@ -395,9 +402,13 @@ class BleCharacteristic {
     ssize_t setValue(const uint8_t* buf, size_t len);
     ssize_t setValue(const String& str);
     ssize_t setValue(const char* str);
+    ssize_t setValue(const uint8_t* buf, size_t len, bool ack);
+    ssize_t setValue(const String& str, bool ack);
+    ssize_t setValue(const char* str, bool ack);
 
     template<typename T>
-    ssize_t setValue(T val) {
+    typename std::enable_if<std::is_integral<T>::value, ssize_t>::type
```
avtolstoy: This also prevents things like:
```cpp
struct TestStruct {
    int data;
    bool data1;
};
TestStruct test;
XXX.setValue(test);
```

Is this something we want to do? Perhaps this should be `is_standard_layout` instead.

XuGuohui: I don't think we support sending this kind of struct for now, besides, it may send random data due to struct alignment.

avtolstoy: All I'm saying is that right now it's possible and whether we should introduce a restriction.

> it may send random data due to struct alignment.

Or it may not if both sides follow the format.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1135,20 +1135,33 @@ size_t BleCharacteristic::description(char* buf, size_t len) const {
 }
 
 ssize_t BleCharacteristic::setValue(const uint8_t* buf, size_t len) {
-    if (buf == nullptr || len == 0) {
+    if (buf == nullptr || len == 0 || len > BLE_MAX_ATTR_VALUE_PACKET_SIZE) {
```
avtolstoy: I think I suggested doing this only for `is_integral` case. This is still useful for buffers and strings to have streaming (e.g. UART over BLE).

wiring/inc/spark_wiring_ble.h
XuGuohui: If using `is_standard_layout`, it causes ambiguous calling to `setValue(const uint8_t* buf, size_t len)`. BTW, `setValue(const T& val)` would be better in comparison to `setValue(T val)`. Also type restriction should also apply to the `getValue(T* val)` template.

avtolstoy: > it causes ambiguous calling to

That's weird. It does handle things like `uint8_t[123]`, but there is no `size_t` argument. It might introduce ambiguity 
to `const char*` variant I would expect if passed string is `const char[]`.

I've commented on `getValue` in another PR. I think it's fine to leave things as is and require the users to cast to `uint8_t*`, however we might as well instead:

1. `std::is_standard_layout && !std::is_array` to perhaps allow standard layout structs and scalars and disallow arrays
2. Or potentially "simply" exclude ambiguous types (I would expect `uint8_t[]` and `char[]`) to also allow arrays of standard layout types. It's somewhat convoluted in C++14, but should be easier to do in C++17 with `if constexpr()`.

> setValue(const T& val) would be better

Perhaps, practically there shouldn't be any difference in optimizations.

I propose we simply go with the first approach for now in order to keep the existent behavior. Might as well make to `const T&`.

XuGuohui: The ambiguity I mentioned is introduced due to `setValue(const T& val, bool ack)`.

wiring/inc/spark_wiring_ble.h
```diff
@@ -393,29 +399,21 @@ class BleCharacteristic {
...
-    }
+    ssize_t setValue(const uint8_t* buf, size_t len, BleTxRxType type = BleTxRxType::AUTO);
+    ssize_t setValue(const String& str, BleTxRxType type = BleTxRxType::AUTO);
+    ssize_t setValue(const char* str, BleTxRxType type = BleTxRxType::AUTO);
 
     template<typename T>
-    typename std::enable_if<std::is_integral<T>::value, ssize_t>::type
-    setValue(T val, bool ack) {
-        return setValue(reinterpret_cast<const uint8_t*>(&val), sizeof(T), ack);
+    typename std::enable_if<std::is_standard_layout<T>::value, ssize_t>::type
```
avtolstoy: Can be simplified with `std::enable_if_t` by the way.

PR #1903, opened 2019-09-04
[ch34058] update release script to include Electron without DEBUG_BUILD=y and tinker-serial-debugging apps for Gen 3

technobly:
### Problem

- Some customers would like the Electron released with DEBUG_BUILD=n
- When debugging Gen 3 devices, it is very convenient to use `tinker-serial-debugging` if you do not have a Serial USB adapter for use with `tinker-serial1-debugging` that is already being generated.

### Solution

- Add Electron release binaries without DEBUG_BUILD=y
- Add Gen 3 tinker-serial-debugging apps (USB output)

### Steps to Test

```
build $ ./make_release.sh --platform=electron --debug --publish=1.4.0-rc.1
build $ ./make_release.sh --platform=boron --debug --publish=1.4.0-rc.1
// repeat last command for other Gen 3 devices if desired
```

### References

ch34058

---

### Completeness

- [x] N/A Added documentation

---

-- [internal] [ch34058] update release script to include Electron without DEBUG_BUILD=y and `tinker-serial-debugging` apps for Gen 3 [#1903](https://github.com/particle-iot/device-os/pull/1903)

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -390,19 +390,23 @@ elif [ $PLATFORM_ID -eq 6 ] || [ $PLATFORM_ID -eq 8 ]; then
 elif [ $PLATFORM_ID -eq 10 ]; then
     # Configure
     cd ../modules
-    DEBUG_BUILD="y"
     MODULAR="y"
+    declare -a debugBuildOptions=("y" "n")
 
-    # Compose, echo and execute the `make` command
-    MAKE_COMMAND="make -s clean all PLATFORM_ID=$PLATFORM_ID COMPILE_LTO=n DEBUG_BUILD=$DEBUG_BUILD MODULAR=$MODULAR USE_SWD_JTAG=$USE_SWD_JTAG USE_SWD=n APP=tinker"
-    echo $MAKE_COMMAND
-    eval $MAKE_COMMAND
+    for debugBuildOption in ${debugBuildOptions[@]}; do
+        DEBUG_BUILD=$debugBuildOption
 
-    # Migrate file(s) into output interface
-    release_binary "system-part1" "system-part1" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
-    release_binary "system-part2" "system-part2" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
-    release_binary "system-part3" "system-part3" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
-    release_binary "user-part" "tinker" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
+        # Compose, echo and execute the `make` command
+        MAKE_COMMAND="make -s clean all PLATFORM_ID=$PLATFORM_ID COMPILE_LTO=n DEBUG_BUILD=$DEBUG_BUILD MODULAR=$MODULAR USE_SWD_JTAG=$USE_SWD_JTAG USE_SWD=n APP=tinker"
+        echo $MAKE_COMMAND
+        eval $MAKE_COMMAND
+
+        # Migrate file(s) into output interface
+        release_binary "system-part1" "system-part1" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
+        release_binary "system-part2" "system-part2" "-m" "$DEBUG_BUILD" "$USE_SWD_JTAG"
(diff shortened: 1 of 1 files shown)


build/release.sh
```diff
@@ -417,24 +421,32 @@ elif [ $PLATFORM_ID -eq 12 ] || [ $PLATFORM_ID -eq 13 ] || [ $PLATFORM_ID -eq 14
...
+
+        # Migrate file(s) into output interface
+        if [ "$MODULAR" = "n" ]; then
+            release_binary "$debugApp" "$debugApp" "$SUFFIX" "$DEBUG_BUILD" "$USE_SWD_JTAG"
+        else
+            release_binary "system-part1" "system-part1" "$SUFFIX" "$DEBUG_BUILD" "$USE_SWD_JTAG"
+            release_binary "user-part" "tinker" "$SUFFIX" "$DEBUG_BUILD" "$USE_SWD_JTAG"
+        fi
+
+        # Keep as explicit last step
```
zfields: You should mention why this needs to be the last step (i.e. Only non-modular builds should loop).

build/release.sh
```diff
@@ -390,19 +390,23 @@ elif [ $PLATFORM_ID -eq 6 ] || [ $PLATFORM_ID -eq 8 ]; then
...
     cd ../modules
-    DEBUG_BUILD="y"
     MODULAR="y"
+    declare -a debugBuildOptions=("y" "n")
 
-    # Compose, echo and execute the `make` command
-    MAKE_COMMAND="make -s clean all PLATFORM_ID=$PLATFORM_ID COMPILE_LTO=n DEBUG_BUILD=$DEBUG_BUILD MODULAR=$MODULAR USE_SWD_JTAG=$USE_SWD_JTAG USE_SWD=n APP=tinker"
-    echo $MAKE_COMMAND
-    eval $MAKE_COMMAND
+    for debugBuildOption in ${debugBuildOptions[@]}; do
```
zfields: Is this really Electron only logic?

zfields:
Firstly, the build logic looks like it will work as expected.

My only concern is regarding the introduction of "special logic" for certain platforms. My preference would be for us to make a decision about what we generate during the build, and those artifacts should be generated for all applicable platforms.

build/release.sh
technobly: I had that comment originally when I had the break just after the `release_binary "user-part" ...` step, but then thought it should really be made the explicit last step in case more code needs to be added.  Because the code is so simple `if MODULAR=y, break` I thought it should be self documenting.

build/release.sh
technobly: Yes, because it's the only platform built by default with DEBUG_BUILD=y for the release build.

technobly:
>My only concern is regarding the introduction of "special logic" for certain platforms. My preference would be for us to make a decision about what we generate during the build, and those artifacts should be generated for all applicable platforms.

@zfields The motivation behind this is just to keep the build process as short as possible.  We could use more debugging apps for Gen 3, but those same debugging apps are not useful for Core/Photon/P1 as those platforms don't have many (or any) logs available by default. So no reason to build and distribute more than just the one debugging app (--debug) for those.

build/release.sh
zfields: To that point, it is obvious what's happening only because there is so little code.

My preference would be to break immediately. No reason to write code for things that don't exist, and when the code becomes more complex it may still require a more complex solution.

avtolstoy:
@technobly Do we want to backport this to 1.3.1?

technobly:
@avtolstoy it’s not absolutely required but it would be a nice thing to do. Let me know if you’d like me to make a PR for it.

build/release.sh
technobly: I would say your suggestion is craftier, although it's not easier to understand that `declare -a apps=("tinker")` doesn't do anything besides cause the for loop to be executed once, and act as a break.  Oddly enough, this might be more clear `declare -a apps=("modular-y-this-is-not-used")`, or maybe changing `release_binary "user-part" "tinker"` to `release_binary "user-part" "$app"`.

zfields: I would disagree, "declare apps" followed by "for app in apps" practically reads like English.

Again, I'm not blocking the PR, I'm just offering my opinion at this point.

technobly: I merged our suggestions together :)

PR #1905, opened 2019-09-04
Various bugfixes

sergeuz:
This PR fixes a number of issues which I encountered while testing Device OS with a local device service instance:

- Due to a regression introduced in 1.1.0, the system layer always sends its handshake messages, even if the session was resumed. This causes increased data usage.

- `rand()` isn't properly seeded in multiple threads, including the system thread. On Gen 3 specifically this breaks allocation of ephemeral port numbers in LwIP.

- Control requests that reset the device (e.g. `particle usb dfu`) can cause unnecessary reconnection to the cloud.

- User module isn't initialized in monolithic builds.


avtolstoy:
I'm going to backport these to 1.3.1 as well.

PR #1908, opened 2019-09-06
Notify the cloud about keepalive interval changes

sergeuz:
### Problem

User applications can change the keepalive interval of the cloud connection via the `Particle.keepAlive()` method. The interval can also be changed by the system depending on the cellular provider and/or the type of the external connectivity used by the gateway device in a mesh network.

At present, the cloud has no means to determine the keepalive interval set on a particular device, and this affects the ability of the cloud to track the online status of the device accurately.

### Solution

Publish a special private event (`particle/device/keepalive`) containing the actual value of the keepalive interval every time the interval changes.

### Steps to Test

Flash the example application to a device. Subscribe to the device's events and verify that keepalive events get published when the keepalive interval is changed.

**Note:** The cloud-side handler for the keepalive events is not yet implemented, so this test requires running a local instance of the device service, which has `particle/device/keepalive` added to the list of known internal events.

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(MANUAL);

namespace {

int cloudStatus = cloud_status_disconnected;
bool done = false;

void cloudStatusChanged(system_event_t event, int param) {
    cloudStatus = param;
}

void waitCloudStatus(int status) {
    while (cloudStatus != status) {
        Particle.process();
    }
}

} // unnamed

void setup() {
    System.on(cloud_status, cloudStatusChanged);
}

void loop() {
    if (done) {
        return;
    }

    // The default keepalive interval should be published after the protocol handshake
    Particle.connect();
    waitCloudStatus(cloud_status_connected);
    delay(1000);

    // This should generate a new keepalive event
    Particle.keepAlive(10);

    // This should not generate a new event, since the interval has not changed
    Particle.keepAlive(10);

    // Disconnect from the cloud
    delay(1000);
    Particle.disconnect();
    waitCloudStatus(cloud_status_disconnected);

    // Set the interval before establishing a connection to the cloud. The interval should be
    // published after the protocol handshake
    Particle.keepAlive(20);
    Particle.connect();
    waitCloudStatus(cloud_status_connected);

    // Disconnect from the cloud
    delay(1000);
    Particle.disconnect();
    waitCloudStatus(cloud_status_disconnected);

    // Set the interval while the handshake is in progress. Two events should be published during
    // the handshake. Note: At present, the events may be delivered in the wrong order
    Particle.connect();
    waitCloudStatus(cloud_status_connecting);
    // Run one more iteration of the system loop to let the system start the handshake
    Particle.process();
    Particle.keepAlive(30);

    // Disconnect from the cloud
    delay(1000);
    Particle.disconnect();
    waitCloudStatus(cloud_status_disconnected);

    done = true;
}
```

### DIFF
--- a/communication/src/ping.h
+++ b/communication/src/ping.h
@@ -45,9 +45,11 @@ public:
 		// TODO: It feels that this logic should have been implemented in the system layer
 		if ( !(this->keepalive_source == KeepAliveSource::USER && source == KeepAliveSource::SYSTEM) )
 		{
-			this->ping_interval = interval;
 			this->keepalive_source = source;
-			return true;
+			if (this->ping_interval != interval) {
+				this->ping_interval = interval;
+				return true;
+			}
 		}
 		return false;
 	}
--- a/communication/src/ping.h
+++ b/communication/src/ping.h
@@ -24,7 +24,15 @@ public:
 		this->keepalive_source = KeepAliveSource::SYSTEM;
 	}
 
-	void set_interval(system_tick_t interval, keepalive_source_t source)
+	/**
+	 * Sets the ping interval.
+	 *
+	 * @param interval New interval in milliseconds.
+	 * @param source Source of the interval change. The interval set by the user application takes
+	 *               precedence over the interval set by the system.
+	 * @return `true` if the current interval has been changed or `false` otherwise.
+	 */
+	bool set_interval(system_tick_t interval, keepalive_source_t source)
 	{
 		/**
 		 * LAST  CURRENT  UPDATE?
@@ -34,11 +42,28 @@ public:
 		 * USER  SYS      NO
 		 * USER  USER     YES
 		 */
+		// TODO: It feels that this logic should have been implemented in the system layer
 		if ( !(this->keepalive_source == KeepAliveSource::USER && source == KeepAliveSource::SYSTEM) )
 		{
 			this->ping_interval = interval;
 			this->keepalive_source = source;
+			return true;
 		}
+		return false;
+	}
+
+	/**
+	 * Returns the current ping interval.
+	 *
+	 * @param source[out] Source of the last interval change.
+	 * @return Interval in milliseconds.
+	 */
+	system_tick_t get_interval(keepalive_source_t* source = nullptr) const
+	{
+		if (source) {
+			*source = keepalive_source;
+		}
+		return ping_interval;
 	}
 
 	void reset()
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -54,6 +54,7 @@
 
 using particle::CloudDiagnostics;
 using particle::publishEvent;
+using particle::publishKeepaliveInterval;
 
 #ifndef SPARK_NO_CLOUD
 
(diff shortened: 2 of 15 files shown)


system/src/system_cloud_internal.cpp
```diff
@@ -358,6 +359,7 @@ constexpr const char KEY_RESTORE_EVENT[] = "spark/device/key/restore";
 constexpr const char DEVICE_UPDATES_EVENT[] = "particle/device/updates/";
 constexpr const char FORCED_EVENT[] = "forced";
 constexpr const char UPDATES_PENDING_EVENT[] = "pending";
+constexpr const char KEEPALIVE_INTERVAL_EVENT[] = "particle/device/keepalive";
```
m-mcgowan: The device service presently already has the `[spark/particle]/session/timeout` event, which was made before we had UDP, so the name is now somewhat confusing, but it serves to feed the IdleChecker about how long to consider the device offline.

sergeuz: Good to know, thanks! I also noticed that the device service expects the interval to be in seconds, while the current code sends it in milliseconds. Are we sure it's safe if Device OS will start publishing `session/timeout` events? cc @JamesHagerman

sergeuz: @m-mcgowan @JamesHagerman Pinging here again. Should I change the event name to `particle/session/timeout`?

JamesHagerman: We still have at least two listeners for that `[spark/particle]/session/timeout` event in the Device Service. Here are the details:

In `reservedEventHandler.js`, we take the data from that event and attempt to store it in MongoDB in the `cores` record under `hardware.session_timeout`.

In `IdleChecker.js` (which is only used by TCP Devices currently) we attempt to pull that record out of mongo when the device connects. If it exists, we use it to control timing around when we call `this.device.disconnect()`.

I queried the production DB using `db.getCollection('cores').find({'hardware.session_timeout': {$exists: true}})` and there are exactly 35 records that contain that key. 

`db.getCollection('cores').find({'hardware.session_timeout': {$gt: 0}})` returns 0 results.

It is safe to say that we are not using that event any more. All of that is legacy

KeepAlive is not related to the session, but rather to the connection itself. At least for UDP devices, the session lives on after a device goes offline.

I think `particle/device/keepalive` is a lot closer to what we're looking for in terms of a namespace for that event. Let's use the version you have in this PR.

sergeuz: For visibility: we discussed this internally and decided to use the new event name for now.

sergeuz: @m-mcgowan Does the rest of the PR look good to you?

communication/src/ping.h
```diff
@@ -34,11 +42,30 @@ class Pinger
 		 * USER  SYS      NO
 		 * USER  USER     YES
 		 */
+		// TODO: It feels that this logic should have been implemented in the system layer
```
m-mcgowan: I agree with this.  The comms layer just provides the timeout and the system layer manages the multiple stakeholders that may choose to change it.  Similarly to how we manage the LED.

PR #1915, opened 2019-09-13
BLE: fixes that setValue() doesn't work with BleTxRxType::AUTO if characteristic has WRITE property.

### LINKED ISSUE #1913: Xenon: Central BLE setValue does not update characteristic value
### Expected Behavior

Using a Xenon and a Nordic Thingy 52, the purpose of my exercise was to prove the Xenon in central mode can both read and write to a particular characteristic. 

When connected, `ledCharacteristic.getValue(buf, sizeof(buf));` should retrieve the contents  of the register.

It should also be able to set the value of the characteristic using `ledCharacteristic.setValue(buf, sizeof(buf));`

(Where buf is `uint8_t buf[4];` in this case)

### Observed Behavior

`ledCharacteristic.getValue` works great. Returns the data as expected. 

When the data is modified, say from 0x01ffffff to 0x0100ffff. And then passed to ledCharacteristic.setValue, it does not update accordingly.

To be specific I'm writing to the LED control characteristic: ` BleUuid("EF680301-9B35-4933-9B10-52FFA9740042");` When `.properties()` is run on the characteristic it returns 0xa (read and write, as expected)

### Steps to Reproduce

Device OS 1.3.1
I was able to set the value using both nRF Connect and Light Blue explorer. 

### Test App

```c
/*
 * Project bluetooth-scan-and-connect
 * Description:
 * Author:
 * Date:
 */

#include "Particle.h"

// For debugging only
SYSTEM_MODE(MANUAL);

// For logging
SerialLogHandler logHandler(115200, LOG_LEVEL_ERROR, {
    { "app", LOG_LEVEL_TRACE }, // enable all app messages
});

// Used in centra mode to track the thingy and the LED characteristic
const BleUuid thingyLedUUID = BleUuid("EF680301-9B35-4933-9B10-52FFA9740042");
BleAddress thingyAddress = BleAddress("ff:ff:ff:ff:ff:ff");
BlePeerDevice thingy;
BleCharacteristic ledCharacteristic;

// Callback when a new device is found advertising
void scanResultCallback(const BleScanResult *scanResult, void *context) {

  String name = scanResult->advertisingData.deviceName();

  // Return if the name does not equal what we're looking for
  if( name != "Thingy" ) return;

  // Set the address
  thingyAddress = scanResult->address;

  // Stop scanning
  BLE.stopScanning();

}

// setup() runs once, when the device is first turned on.
void setup() {
  (void)logHandler; // Does nothing, just to eliminate warning for unused variable
}

// loop() runs over and over again, as quickly as it can execute.
void loop() {
  // Check for scan start

  if( !thingy.connected() ) {
    if( !(thingyAddress == BleAddress("ff:ff:ff:ff:ff:ff")) ) {
      // Connect
      Log.trace("Connecting to Thingy..");
      thingy = BLE.connect(thingyAddress);
    } else {
      Log.trace("Scanning for Thingy..");
      BLE.scan(scanResultCallback, NULL);
    }

  } else {
    if( !ledCharacteristic.valid() ) {
      Log.trace("Characteristic not valid!");
      thingy.getCharacteristicByUUID(ledCharacteristic, thingyLedUUID);
    } else {
      // Get the properties
      // BleCharacteristicProperty prop = ledCharacteristic.properties();
      // Log.trace("Prop %x",(uint8_t)prop);

      // Get the value of the characteristic
      uint8_t buf[4];

      // Get the current value
      ledCharacteristic.getValue(buf, sizeof(buf));
      Log.trace("Buffer value %x %x %x %x", buf[0], buf[1], buf[2], buf[3] );

      // Iterate over all colors
      if( buf[1] == 0xff) {
        buf[1] = 0x00;
        buf[2] = 0xff;
      }else if( buf[2] == 0xff) {
        buf[2] = 0x00;
        buf[3] = 0xff;
      } else if( buf[3] == 0xff) {
        buf[3] = 0x00;
        buf[1] = 0xff;
      }

      // Output buffer value
      Log.trace("Buffer write value %x %x %x %x", buf[0], buf[1], buf[2], buf[3] );

      // Writing the value
      ledCharacteristic.setValue(buf, sizeof(buf));

    }
  }

  delay(1000);
}
```

### References

None at this time.

### LINKED ISSUE #1924: Resetting the BLE notify callback in central mode upon new peripheral connection
<details>

</details>

## Bug Report

Setup: Multiple peripheral to single central (xenon)

Whenever there is a new peripheral connection, the callback for the same characteristic of previous peripheral (peer) is set to null which results in the app callback not being called for those peripherals. Is there a reason why inheritedCallback resets the callbackptr (https://github.com/particle-iot/device-os/blob/2896ead2f717e5cdea2f733e502593064ee8c125/wiring/src/spark_wiring_ble.cpp#L805)?

### Expected Behavior

The app callback should be called for all peripherals upon receiving data notify for characteristic.

### Observed Behavior

The callback is invoked only for the last peripheral connected.

### Steps to Reproduce
Connect multiple peripherals, upon data notify, the application callback will not be invoked in case of the peripherals which connected earlier than the last peripheral. This is due to the inheritCallback (https://github.com/particle-iot/device-os/blob/2896ead2f717e5cdea2f733e502593064ee8c125/wiring/src/spark_wiring_ble.cpp#L805) setting the callback to null.

Using the following possible workarounds currently:
1. Execute the below code upon every connection before getCharacteristic
`temperatureMeasurementCharacteristic.onDataReceived(onDataReceivedTemp, NULL);`

2. Do not reset the callback ptr in inheritCallback (https://github.com/particle-iot/device-os/blob/2896ead2f717e5cdea2f733e502593064ee8c125/wiring/src/spark_wiring_ble.cpp#L805).

### Test App
```
#include "application.h"

SYSTEM_MODE(MANUAL);

const size_t SCAN_RESULT_MAX = 30;

BleCharacteristic temperatureMeasurementCharacteristic;

BleScanResult scanResults[SCAN_RESULT_MAX];
BlePeerDevice peer;

void onDataReceivedTemp(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context);

void setup()
{
    (void)logHandler; // Does nothing, just to eliminate the unused variable warning

    temperatureMeasurementCharacteristic.onDataReceived(onDataReceivedTemp, NULL);
}

void loop()
{
    if (BLE.connected())
    {
    }
    //else
    {
        Log.info("BLE scan!");
        int count = BLE.scan(scanResults, SCAN_RESULT_MAX);

        if (count)
        {
            Log.info("%d devices found", count);
        }

        for (int ii = 0; ii < count; ii++)
        {
            uint8_t buf[BLE_MAX_ADV_DATA_LEN];
            size_t len;

            // We're looking for devices that have a temperature service (0x1809)
            len = scanResults[ii].advertisingData.get(
                BleAdvertisingDataType::SERVICE_UUID_16BIT_COMPLETE, buf, BLE_MAX_ADV_DATA_LEN);
            if (len > 0)
            {
                for (size_t jj = 0; jj < len; jj += 2)
                {
                    if (*(uint16_t*)&buf[jj] == BLE_SIG_UUID_HEALTH_THERMONETER_SVC)
                    { // 0x1809
                        // Found a device with a temperature service

                        Log.info("Connect to rssi=%d address=%02X:%02X:%02X:%02X:%02X:%02X ",
                                 scanResults[ii].rssi, scanResults[ii].address[0],
                                 scanResults[ii].address[1], scanResults[ii].address[2],
                                 scanResults[ii].address[3], scanResults[ii].address[4],
                                 scanResults[ii].address[5]);

                        peer = BLE.connect(scanResults[ii].address);
                        if (peer.connected())
                        {
                            Log.info("successfully connected!");

                            // Get the temperature measurement characteristic
                            temperatureMeasurementCharacteristic = peer.getCharacteristicByUUID(BleUuid(0x2A1C));
                            //peer.getCharacteristicByUUID(temperatureMeasurementCharacteristic, BleUuid(0x2A1C));
                            Log.info("prop %d", temperatureMeasurementCharacteristic.properties());                   
                        }
                        else
                        {
                            Log.info("connection failed");
                        }
                    }
                }
            }
        }
    }
}

void onDataReceivedTemp(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context)
{
    for (int i = 0; i < len; i++)
    {
        Log.info("temperature[%d]=%x", i, data[i]);
    }
}
```
### References

https://github.com/particle-iot/device-os/blob/2896ead2f717e5cdea2f733e502593064ee8c125/wiring/src/spark_wiring_ble.cpp#L805

---

XuGuohui:
### Problem

https://github.com/particle-iot/device-os/issues/1913

### Solution

Send characteristic value, which property is `WRITE`, when `BleTxRxType::AUTO` is used by default.

### Steps to Test

1. Build and flash the test applications `user/tests/wiring/ble_central_peripheral` to two Gen3 devices respectively.
2. Simply connect them to serial terminals and type `t` on the terminal to start the test.

### Example App

N/A

### References

- Closes #1913
- Closes #1924

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
+++ b/user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
@@ -7,14 +7,24 @@
 
 BleScanResult results[SCAN_RESULT_COUNT];
 
-BleCharacteristic peerTxCharacteristic;
-BleCharacteristic peerRxCharacteristic;
+BleCharacteristic peerCharRead;
+BleCharacteristic peerCharWrite;
+BleCharacteristic peerCharWriteWoRsp;
+BleCharacteristic peerCharWriteAndWriteWoRsp;
+BleCharacteristic peerCharNotify;
+BleCharacteristic peerCharIndicate;
+BleCharacteristic peerCharNotifyAndIndicate;
 BlePeerDevice peer;
 
-const String str1("This is string 1.");
-const String str2("Hello from Particle.");
-const String str3("Have a good day!");
-bool str1Rec = false, str2Rec = false, str3Rec = false;
+const String str1("05a9ae9588dd");
+const String str2("4f62bd40046a");
+const String str3("6619918032a2");
+const String str4("359b4deb3f37");
+const String str5("a2ef18ec6eaa");
+const String str6("7223b5dd3342");
+const String str7("d4b4249bbbe3");
+
+bool str1Rec = false, str2Rec = false, str3Rec = false, str4Rec = false, str5Rec = false, str6Rec = false, str7Rec = false;
 
 static void onDataReceived(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context) {
     String str((const char*)data, len);
@@ -27,10 +37,24 @@ static void onDataReceived(const uint8_t* data, size_t len, const BlePeerDevice&
     if (str == str3) {
         str3Rec = true;
     }
+    if (str == str4) {
+        str4Rec = true;
+    }
+    if (str == str5) {
+        str5Rec = true;
+    }
+    if (str == str6) {
+        str6Rec = true;
+    }
+    if (str == str7) {
+        str7Rec = true;
+    }
 }
 
 test(BLE_01_Central_Scan_And_Connect) {
-    peerTxCharacteristic.onDataReceived(onDataReceived, &peerTxCharacteristic);
+    peerCharNotify.onDataReceived(onDataReceived, &peerCharNotify);
+    peerCharIndicate.onDataReceived(onDataReceived, &peerCharIndicate);
+    peerCharNotifyAndIndicate.onDataReceived(onDataReceived, &peerCharNotifyAndIndicate);
 
     int ret = BLE.setScanTimeout(100); // Scan timeout: 1s
     assertEqual(ret, 0);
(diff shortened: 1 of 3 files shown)


XuGuohui:
Closes #1924

PR #1917, opened 2019-09-16
[electron, gen2] fixes RSSI failing due to Power Saving mode active [ch38428]

### LINKED ISSUE #1892: Electron 3G U270 does not report RSSI since 1.3.1-rc.1
### Expected Behavior
CellularSignal sig = Cellular.RSSI();
sig.getStrength()
sig.getQuality()
sig.getAccessTechnology()

Should report meaningful values > 0

### Observed Behavior
In some cases (possibly when connected to 2G network) 
sig.getStrength=-1
sig.getQuality=-1
sig.getAccessTechnology=0

### Steps to Reproduce
Was working fine with FW 1.2.1
Broken with 1.3.1-rc.1

Sample Code:
```
loop() {
//timer5s is increased every 5 seconds
		if (timer5s % 720 == 0) {
            snprintf(INFOchar, 620, "V: %s FW: %s FreeMem: %d Up: %ds", FWVERSION, (const char*)System.version(), System.freeMemory(), millis() / 1000);
			if (Particle.connected()) {
				CellularSignal sig = Cellular.RSSI();
                snprintf(INFOchar, 620, "V: %s FW: %s FreeMem: %d Up: %ds %.0f%%/%.0f%%/%d", FWVERSION, (const char*)System.version(), System.freeMemory(), millis() / 1000, sig.getStrength(),sig.getQuality(),sig.getAccessTechnology());
				result = Particle.publish("Info", INFOchar, 300, PRIVATE);
			}
		}
}
```
With FW 1.2.1 the results look like this:

V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 155137s 56%/88%/4 | 10:14 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 151529s 56%/88%/4 | 09:14 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 147924s 56%/88%/4 | 08:14 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 144315s 56%/88%/4 | 07:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 140709s 56%/88%/4 | 06:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 137103s 56%/88%/4 | 05:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 133496s 56%/88%/4 | 04:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 129889s 56%/88%/4 | 03:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 126282s 56%/88%/4 | 02:13 22.08.2019
V: e2.0.11 20190805 FW: 1.2.1 FreeMem: 70472 Up: 122674s 56%/88%/4 | 01:13 22.08.2019

With 1.3.1-rc.1:

V: e2.0.11 20190822 FW: 1.3.1-rc.1 FreeMem: 73072 Up: 18053s -1%/-1%/0 | 15:22 22.08.2019
V: e2.0.11 20190822 FW: 1.3.1-rc.1 FreeMem: 73088 Up: 14438s -1%/-1%/0 | 14:22 22.08.2019
V: e2.0.11 20190822 FW: 1.3.1-rc.1 FreeMem: 73088 Up: 10830s -1%/-1%/0 | 13:22 22.08.2019
V: e2.0.11 20190822 FW: 1.3.1-rc.1 FreeMem: 73088 Up: 7225s 58%/88%/4 | 12:22 22.08.2019
V: e2.0.11 20190822 FW: 1.3.1-rc.1 FreeMem: 73088 Up: 3618s 58%/88%/4 | 11:22 22.08.2019

technobly:
### Problem

PR #1886 added some AT / OK checks before commands with long delays, to force a quicker timeout when the AT interface was not responsive.  Unfortunately this negatively affected Gen 2 2G/3G modems due to the Power Saving mode being set to UPSV=1 by default which enables low power mode after about 9.2 seconds in most cases.  The issue is that when power saving mode is active, the AT / OK check can take longer than the 1 second timeout is currently has set, while the modem is transitioning from the IDLE to ACTIVE state.  In practice, I see this take anywhere from 0.5 seconds to 3 seconds.  The U2/G3 integration manual does not state how long this should take.

Gen 2 & Gen 3 LTE devices are not affected due to power saving mode being disabled by default.

Gen 3 2G/3G devices are also not affected due to power saving mode being disabled by default.

### Solution

If AT / OK check does not respond in 1 second, and the device is not an LTE device and does have power savings mode enabled (Gen 2 2G/3G devices), wait up to 5 seconds for the OK response and CTS line to indicate the modem is ready.

### Steps to Test

Test the example app with U260, U201, R410, and G350.  Every time getSignalStrength() is called, initial AT / OK check should succeed and not timeout before successive commands are sent.  This requires inspecting the logs.
✅ R410
✅ U201
✅ U260
✅ G350

### Device OS for this PR (use debug version)

⬇️ [particle-device-os@1.4.0+pr1917.zip](https://www.dropbox.com/s/5cxcpwapmguirc1/particle-device-os%401.4.0%2Bpr1917.zip?dl=1)

### Example App

```c
SYSTEM_MODE(SEMI_AUTOMATIC)
char myData[1024];

void setup() {    
    Particle.connect();
}

void loop() {
    if (Particle.connected()) {
        CellularSignal sig = Cellular.RSSI();
        snprintf(myData, 620, "FW: %s FreeMem: %d Up: %ds %.0f%%/%.0f%%/%d", (const char*)System.version(), System.freeMemory(), millis() / 1000, sig.getStrength(),sig.getQuality(),sig.getAccessTechnology());
        Particle.publish("Info", myData, 300, PRIVATE);
    }
    delay(20000);
}
```

### References

Fixes #1892 

---

### Completeness

- [x] N/A Added documentation

---

- [bugfix] [electron, gen2] fixes RSSI failing due to Power Saving mode active [#1917](https://github.com/particle-iot/device-os/pull/1917)

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -335,7 +335,21 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
 bool MDMParser::_atOk(void)
 {
     sendFormated("AT\r\n");
-    return (RESP_OK == waitFinalResp(nullptr, nullptr, AT_TIMEOUT));
+    int resp = waitFinalResp(nullptr, nullptr, AT_TIMEOUT);
+
+    // R410 Power Savings Mode currently disabled, therefore this only affects all other modems
+    if (resp == WAIT && _dev.dev != DEV_SARA_R410 && _dev.lpm == LPM_ACTIVE) {
+        // if low power mode active, may have to wait up to 5s for OK response
+        int cts = HAL_GPIO_Read(CTS_UC); // 1: not ready, 0: ready
+        system_tick_t t0 = HAL_Timer_Get_Milli_Seconds();
+        // wait for any AT response and CTS ready, or timeout after 5s total (1s + 4s)
+        while ((resp == WAIT || cts == 1) && !TIMEOUT(t0, 4000)) {
+            resp = waitFinalResp(nullptr, nullptr, 200);
+            cts = HAL_GPIO_Read(CTS_UC);
+        }
+    }
+
+    return (resp == RESP_OK);
 }
 
 int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
@@ -944,7 +958,9 @@ bool MDMParser::init(DevStatus* status)
             waitFinalResp(); // Not checking for error since we will reset either way
             goto reset_failure;
         }
-        // Force Power Saving mode to be disabled for good measure
+        // Force Power Saving mode to be disabled
+        //
+        // TODO: if we enable this feature in the future update the logic in MDMParser::_atOk
         sendFormated("AT+CPSMS=0\r\n");
         if (RESP_OK != waitFinalResp()) {
             goto failure;
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -870,10 +870,16 @@ int SaraNcpClient::initReady() {
             }
         }
         CHECK_PARSER_OK(lastError);
-        // Force Power Saving mode to be disabled for good measure
+        // Force Power Saving mode to be disabled
+        //
+        // TODO: if we enable this feature in the future add logic to CHECK_PARSER macro(s)
+        // to wait longer for device to become active (see MDMParser::_atOk)
         CHECK_PARSER_OK(parser_.execCommand("AT+CPSMS=0"));
     } else {
-        // Power saving
+        // Force Power Saving mode to be disabled
+        //
+        // TODO: if we enable this feature in the future add logic to CHECK_PARSER macro(s)
+        // to wait longer for device to become active (see MDMParser::_atOk)
         CHECK_PARSER_OK(parser_.execCommand("AT+UPSV=0"));
     }
 


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -335,7 +335,21 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
 bool MDMParser::_atOk(void)
 {
     sendFormated("AT\r\n");
-    return (RESP_OK == waitFinalResp(nullptr, nullptr, AT_TIMEOUT));
+    bool atok = (RESP_OK == waitFinalResp(nullptr, nullptr, AT_TIMEOUT));
```
zfields: I think we need to be testing for the TIMEOUT case. There are a lot of reasons it could be not `RESP_OK`, and all the remaining logic is designed to handle the case when there is specifically a timeout.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -335,7 +335,21 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
...
+
+    // R410 Power Savings Mode currently disabled, therefore this only affects all other modems
+    if (!atok && _dev.dev != DEV_SARA_R410 && _dev.lpm == LPM_ACTIVE) {
```
zfields: Instead of testing against a specific modem, perhaps we could test against a capability.

> Gen 2 & Gen 3 LTE devices are not affected due to power saving mode being disabled by default.

What if power saving mode has been enabled on these platforms, does this logic account for that situation?

technobly: No it does not, but we shouldn't need to add that logic until we enable power saving mode on those devices (currently it's hard coded to be disabled)

hal/src/electron/modem/mdm_hal.cpp
technobly: It's very unlikely that it would return RESP_ERROR, RESP_ABORTED or RESP_PROMPT, but I don't mind adding an additional check to see if it is in fact the WAIT (timeout) condition.

hal/src/electron/modem/mdm_hal.cpp
zfields: Ahh. Okay, I didn't realize it was hardcoded.

However, I do think if we don't account for it now, then we will forget it once we enable power saving mode, and have a hard to find bug at that time.

hal/src/electron/modem/mdm_hal.cpp
zfields: That's great! When possible, I believe we should be coding for our exact intent and nothing else.

technobly: Added!  Decided to rebase instead of adding extra commits since there's not much code.

hal/src/electron/modem/mdm_hal.cpp
technobly: I added TODO comments near the code that hard codes the disabling of the power modes.  Since the behavior of Power Saving mode for R410 is different than 2G/3G, it is not possible to code this logic in now without fully developing this feature for R410 (which we won't do at this time).

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -335,7 +335,22 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
 bool MDMParser::_atOk(void)
 {
     sendFormated("AT\r\n");
-    return (RESP_OK == waitFinalResp(nullptr, nullptr, AT_TIMEOUT));
+    int resp = waitFinalResp(nullptr, nullptr, AT_TIMEOUT);
+    bool atok = (resp == RESP_OK);
+
+    // R410 Power Savings Mode currently disabled, therefore this only affects all other modems
+    if (!atok && resp == WAIT && _dev.dev != DEV_SARA_R410 && _dev.lpm == LPM_ACTIVE) {
```
zfields: Four conditions is intimidating. This compound `if` could benefit from removing the `!atok` check. It is redundant, because it is always true when `resp == WAIT`.

hal/src/electron/modem/mdm_hal.cpp
zfields: I really love the simplified logic!

However, I still have an issue with checking the `_dev.dev != DEV_SARA_R410`. It looks like this logic would be covered and more importantly _ABSTRACTED_ by `_dev.lpm == LPM_ACTIVE`.

hal/src/electron/modem/mdm_hal.cpp
zfields: I really love the simplified logic!

However, I still have an issue with checking the _dev.dev != DEV_SARA_R410. It looks like this logic would be covered, and more importantly ABSTRACTED, by _dev.lpm == LPM_ACTIVE.

PR #1920, opened 2019-09-19
Increase VitalsPublisher coverage

zfields:
### Problem

Despite VitalsPublisher being a new class, it is not fully covered.

### Solution

Added test to improve coverage. Mark all untestable code and create a CH ticket to identify the work required to make it testable.

### Steps to Test

Run the DeviceOS CMake Catch2 tests

```bash
cd <particle-iot/device-os>/test/unit-tests/
mkdir .build && cd .build/
cmake ..
make all test coverage
```

### References

CH38583: Improve VitalsPublisher Coverage

---

### Completeness

### DIFF
--- a/system/src/system_publish_vitals.cpp
+++ b/system/src/system_publish_vitals.cpp
@@ -7,16 +7,33 @@
 #include "system_cloud.h"
 #include "system_threading.h"
 
-namespace {
+namespace
+{
 
 using namespace particle::system;
 
-int postDescription() {
-    const auto r = spark_protocol_post_description(spark_protocol_instance(), particle::protocol::DESCRIBE_METRICS, nullptr);
-    return spark_protocol_to_system_error(r);
+inline int postDescription()
+{
+    int error;
+
+    if (spark_cloud_flag_connected())
+    {
+        // Transmit CoAP message via communication layer
+        error = spark_protocol_post_description(
+            spark_protocol_instance(), particle::protocol::DESCRIBE_METRICS, nullptr);
+
+        // Convert `protocol` error to `system` error
+        error = spark_protocol_to_system_error(error);
+    }
+    else
+    {
+        error = SYSTEM_ERROR_INVALID_STATE;
+    }
+
+    return error;
 }
 
-} // unnamed
+} // namespace
 
 template <class Timer>
 VitalsPublisher<Timer>::VitalsPublisher(Timer* timer_)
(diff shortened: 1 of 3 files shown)


system/src/system_publish_vitals.cpp
```diff
@@ -7,16 +7,33 @@
...
-namespace {
+namespace
+{
 
 using namespace particle::system;
 
-int postDescription() {
-    const auto r = spark_protocol_post_description(spark_protocol_instance(), particle::protocol::DESCRIBE_METRICS, nullptr);
-    return spark_protocol_to_system_error(r);
+inline int postDescription()
```
m-mcgowan: I would add a comment that this function is called from an ISR.

system/src/system_publish_vitals.cpp
```diff
@@ -84,12 +101,15 @@ void VitalsPublisher<Timer>::period(system_tick_t period_s_)
...
-    if (!spark_cloud_flag_connected()) {
-        return SYSTEM_ERROR_INVALID_STATE;
-    }
     return postDescription();
 }
 
+// Functionality covered by E2E tests
+#define LCOV_EXCL_START
+/*
+ * TODO: CH38588 - Abstract ISRTaskQueue boilerplate
```
m-mcgowan: <3

system/src/system_publish_vitals.cpp
zfields: Will do.

For posterity, it is called from more than just the ISR, but specifically it is `static` so it can be called from the ISR.

I will make a comment to reflect this.

sergeuz: It's not called from an ISR though :)

system/src/system_publish_vitals.cpp
```diff
@@ -100,12 +129,11 @@ void VitalsPublisher<Timer>::publishFromTimer(void)
     }
     task->func = [](ISRTaskQueue::Task* task) {
         delete task;
-        if (spark_cloud_flag_connected()) {
-            postDescription();
-        }
```
sergeuz: Why did you remove this?

system/src/system_publish_vitals.cpp
m-mcgowan: oh is it not? I thought it was called from an ISRTask

test/unit_tests/cloud/publish_vitals.cpp
```diff
@@ -27,6 +27,9 @@
 #include "mock/mock_types.h"
 #include "system_publish_vitals.h"
 
+bool spark_cloud_flag_connected_called;
+int spark_cloud_flag_connected_result;
```
m-mcgowan: There are test frameworks like hippomocks and fakeit that can help avoid the need to manually set return values and track calls. Not a call for change, just a heads up!

system/src/system_publish_vitals.cpp
sergeuz: Yeah, the name is somewhat confusing nowadays. Originally, `ISRTaskQueue` was used to process events posted from ISRs in the system thread, but now other threads use it too (mostly because there's no event queue available when threading is disabled)

test/unit_tests/cloud/publish_vitals.cpp
zfields: This is handling open function calls, and not methods. This may be provided for by hippomocks, but I don't think this is/was available in fakeit at time of writing; I may be wrong.

system/src/system_publish_vitals.cpp
zfields: I've updated the comment to reflect the fact it is enqueued on the ISRTaskQueue, so it can be called on the system thread. :+1:

Issue #1926, opened 2019-09-24
BLE Allocate memory failed after running for long time


sviraaj:
<details>
  <summary><i>submission notes</i></summary>
</details>

## Bug Report
Setup: Xenon central with Multiple peripherals. Since currently device OS supports max 3 peripherals to be connected to Xenon central, can test with 3. More the number of peripherals earlier the repro.

Seeing the following error after running for long times. If number of peripherals) (i.e. number of central links) increased to 8, repro will can be seen within 20-30 mins. 
For a quick test with 8 peripherals use the following commit for supporting 8 peripherals:
https://github.com/sviraaj/device-os/commit/00304d54f4be3870daad0023b1a138ec32fa1683

```
ERROR: Allocate memory for received data failed.
ERROR: Allocate memory for BLE event failed.
ERROR: Allocate memory for adv report data failed.
ERROR: Allocate memory for adv report data failed.
```

### Expected Behavior

Should work without error

### Observed Behavior
Observing the above ERROR prints

### Steps to Reproduce
Run any Xenon central example with multiple peripherals (having data notify property for any characteristics) and keep running for some time. With 8 peripherals connected, error should repro in less than 30 mins (hopefully).

### References

https://github.com/sviraaj/device-os/commit/00304d54f4be3870daad0023b1a138ec32fa1683

---

XuGuohui:
@sviraaj Could you tell me more parameters about the BLE connectivity, like how often those peripherals sending notification? Besides, according to the log, the central kept scanning BLE peripherals, right?

sviraaj:
@XuGuohui The peripherals send a notification every 30 seconds. The scanning does continue, but fails to retrieve any allocate memory for advertising data.

XuGuohui:
This is either encountered by the limited preallocated RAM for the BLE event pool or memory leak of the pool. Since the central are connecting to 8 peripherals sending notification and keep scanning, the pool preallocated is not large enough for caching incoming BLE events. Would you mind help running you project again after increasing the size of the preallocated pool: https://github.com/particle-iot/device-os/blob/develop/hal/src/nRF52840/ble_hal.cpp#L85? In the other hand, I'll try to figure out whether there is memory leak regarding the BLE event pool.

sviraaj:
@XuGuohui Sure. I will check it with the above change. Just FYI, I was able to see an error even in case of 3 peripherals, after running for a longer time. Thanks.

XuGuohui:
[...]

PR #1929, opened 2019-09-27
Fix: BLE scanning memory leak

XuGuohui:
### Problem

BLE memory leak during scanning.
Issue: https://github.com/particle-iot/device-os/issues/1926

### Solution

- Release the memory allocated for scanned advertising data if that event is not processed
- Do not acquire BLE wiring lock in characteristic data received callback.

### Steps to Test

1. Build the test applications below with `DEBUG_BUILD=y MODULAR=n` under the main folder
2. Flash the Central test app into a Xenon and flash the Peripheral test app into 3 Xenons respectively.
3. Run it for a long period and observe the log on the Central side.

### Example App

Central:
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

#define SCAN_RESULT_COUNT       50

BleCharacteristic peerTxCharacteristic[3];
BlePeerDevice peers[3];

    for (uint8_t i = 0; i < 3; i++) {
        if (peer == peers[i]) {
            LOG(TRACE, "Received data from peer %d, %s", i, peer.address().toString().c_str());
            return;
        }
    }
    LOG(TRACE, "Received data from unknown peer device, %s", peer.address().toString().c_str());
}

void setup() {
    for (uint8_t i = 0; i < 3; i++) {
        peerTxCharacteristic[i].onDataReceived(onDataReceived, &peerTxCharacteristic[i]);
    }
}

void loop() {
    delay(3000);

    if (count > 0) {
            BleUuid foundServiceUUID;
            if (svcCount > 0 && foundServiceUUID == "12340001-B5A3-F393-E0A9-E50E24DCCA9E") {
                for (uint8_t j = 0; j < 3; j++) {
                    if (peers[j].connected()) {
                        continue;
                    }
                    LOG(TRACE, "Peer %d is connecting to %s", j, results[i].address.toString().c_str());
                    peers[j] = BLE.connect(results[i].address);
                    if (peers[j].connected()) {
                        peers[j].getCharacteristicByDescription(peerTxCharacteristic[j], "tx");
                    }
                    break;
                }
            }
        }
    }
}
```

Peripheral:
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

const char* serviceUuid = "12340001-B5A3-F393-E0A9-E50E24DCCA9E";
const char* txUuid = "12340002-B5A3-F393-E0A9-E50E24DCCA9E";

BleCharacteristic txCharacteristic("tx",
                                   BleCharacteristicProperty::NOTIFY,
                                   txUuid,
                                   serviceUuid);
void setup() {

    BleAdvertisingData data;
    BLE.advertise(&data);
}

void loop() {
    if (BLE.connected()) {
        txCharacteristic.setValue("hello from particle");
        delay(1000);
    }

```

### References

Closes: https://github.com/particle-iot/device-os/issues/1926

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -1456,7 +1456,12 @@ void BleObject::Observer::notifyScanResultEvent(const hal_ble_scan_result_evt_t&
 }
 
 int BleObject::Observer::processAdvReportEventFromThread(const ble_evt_t* event) {
-    CHECK_TRUE(isScanning_, SYSTEM_ERROR_INVALID_STATE);
+    if (!isScanning_) {
+        if (event->evt.gap_evt.params.adv_report.data.p_data) {
+            BleObject::getInstance().dispatcher()->freeEventData(event->evt.gap_evt.params.adv_report.data.p_data);
+        }
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
     const ble_gap_evt_adv_report_t& advReport = event->evt.gap_evt.params.adv_report;
     hal_ble_addr_t newAddr = toHalAddress(advReport.peer_addr);
     if (isCachedDevice(newAddr)) {
@@ -1485,7 +1490,7 @@ int BleObject::Observer::processAdvReportEventFromThread(const ble_evt_t* event)
         // Scan response data packet
         hal_ble_scan_result_evt_t* result = getPendingResult(newAddr);
         if (!result) {
-            goto continue_scanning;
+            goto free;
         }
         constructObserverEvent(*result, advReport);
         notifyScanResultEvent(*result);
@@ -3240,7 +3245,7 @@ void BleObject::GattClient::processGattClientEvents(const ble_evt_t* event, void
             break;
         }
         case BLE_GATTC_EVT_HVX: {
-            LOG_DEBUG(TRACE, "BLE GATT Server event: data notified.");
+            LOG_DEBUG(TRACE, "BLE GATT Client event: data notified.");
             ble_evt_t* dataNotifiedEvent = (ble_evt_t*)BleObject::getInstance().dispatcher()->allocEventData(sizeof(ble_evt_t) +
                     SUB1(event->evt.gattc_evt.params.hvx.len) * sizeof(uint8_t));
             if (!dataNotifiedEvent) {
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -1038,7 +1038,8 @@ private:
  */
 void BleCharacteristicImpl::onBleCharEvents(const hal_ble_char_evt_t *event, void* context) {
     auto impl = static_cast<BleCharacteristicImpl*>(context);
-    WiringBleLock lk;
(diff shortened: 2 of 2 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1456,7 +1456,12 @@ void BleObject::Observer::notifyScanResultEvent(const hal_ble_scan_result_evt_t&
 }
 
 int BleObject::Observer::processAdvReportEventFromThread(const ble_evt_t* event) {
-    CHECK_TRUE(isScanning_, SYSTEM_ERROR_INVALID_STATE);
+    if (!isScanning_) {
+        if (event->evt.gap_evt.params.adv_report.data.p_data) {
+            BleObject::getInstance().dispatcher()->freeEventData(event->evt.gap_evt.params.adv_report.data.p_data);
```
avtolstoy: Can we simply jump to `free` label instead to avoid call duplication?

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1485,7 +1490,7 @@ int BleObject::Observer::processAdvReportEventFromThread(const ble_evt_t* event)
         // Scan response data packet
         hal_ble_scan_result_evt_t* result = getPendingResult(newAddr);
         if (!result) {
-            goto continue_scanning;
+            goto free;
```
avtolstoy: Can `continue_scanning` and `free` be simply coalesced, since there is anyway a check for `p_data`?

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1040,7 +1040,8 @@ class BleLocalDeviceImpl {
  */
 void BleCharacteristicImpl::onBleCharEvents(const hal_ble_char_evt_t *event, void* context) {
     auto impl = static_cast<BleCharacteristicImpl*>(context);
-    WiringBleLock lk;
+    // This callback won't modified any data in wiring.
```
avtolstoy: Can something else (e.g. application thread) modify anything accessed in this function? I don't see `peers_` being modified by non-ble event thread, but just in case asking if I've missed something.

hal/src/nRF52840/ble_hal.cpp
XuGuohui: No, we can't. For the case that an advertising packet is received and it expects a scan response packet, though the `p_data` is not `nullptr`, we cannot free this piece of memory, but continue scanning. This piece of memory will be freed when the scan response packet is received and notified to user application or the scan procedure finished.

wiring/src/spark_wiring_ble.cpp
XuGuohui: For sure, this callback won't modified anything in wiring **so far**. If it does, then we'll absolutely be aware of that in this callback, since we've left the comment on entering this callback.

hal/src/nRF52840/ble_hal.cpp
XuGuohui: Oh, I just realized that in this check we need to return an error, while the `free` will lead to `continue_scanning` and return `SYSTEM_ERROR_NONE`. So we cannot simply jump to the `free` label.

PR #1930, opened 2019-09-30
Improves cellular test coverage [ch35587]

technobly:
### Problem

Some tests missing for currently covered cellular code.

### Solution

Adds tests and fixes some comments.

### Steps to Test

Coveralls percentages for these files should have increased above these baseline amounts
```
63.79% cellular_internal.cpp
44.68% spark_wiring_cellular_printable.cpp
```
to
```
100% cellular_internal.cpp
100% spark_wiring_cellular_printable.cpp
```

---

### Completeness

- [x] Added documentation

---

- [internal] Increases cellular test coverage [ch35587] [#1930](https://github.com/particle-iot/device-os/pull/1930)

### DIFF
--- a/test/unit_tests/cellular/cellular.cpp
+++ b/test/unit_tests/cellular/cellular.cpp
@@ -22,11 +22,32 @@
 #include "net_hal.h"
 #include <limits>
 #include "spark_wiring_cellular_printable.h"
+#include "appender.h"
 
 #undef WARN
 #undef INFO
 #define CATCH_CONFIG_MAIN
 #include "catch2/catch.hpp"
+
+namespace {
+
+    class FakeStream: public Print {
+    private:
+        BufferAppender append_;
+
+    protected:
+        virtual size_t write(uint8_t c) override {
+            append_.append(&c, 1);
+            return 1;
+        }
+
+    public:
+        FakeStream(char* buf, size_t size) :
+                append_((uint8_t*)buf, size) {
+        }
+    };
+
+} // namespace
 
 using namespace detail;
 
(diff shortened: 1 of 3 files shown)


test/unit_tests/cellular/cellular.cpp
```diff
@@ -22,12 +22,33 @@
...
 
+namespace {
+
+    class FakeStream: public Print {
+    private:
+        BufferAppender append_;
+
+    protected:
+        virtual size_t write(uint8_t c) override {
+            append_.append(&c, 1);
```
zfields: I don't know enough about `BufferAppender::append`, why does this require a pointer?

Is it expecting a temporary buffer to copy the contents out of, or is it going to try and hold onto the pointer?

test/unit_tests/cellular/cellular.cpp
```diff
@@ -78,6 +99,64 @@ TEST_CASE("cellular_signal()") {
         }
     }
 
+    SECTION("CellularSignalHal RSSI and QUAL") {
+        NetStatus status;
+        memset(&status, 0, sizeof(status));
+        status.rssi = -50; // arbitrary non-zero values
+        status.qual = 37;  //  "
+
+        SECTION("CellularSignalHal::rssi and CellularSignalHal::qual are set to expected values") {
```
zfields: I was told to use the BDD (Behavior Driven Development) style. SECTION (API subset) GIVEN (environment) WHEN (input) THEN (output).

I would have expected the structure to look more like this...

```
SECTION("CellularSignalHal RSSI and QUAL") {
  GIVEN("A known NetStatus structure") {
    WHEN("cellular_signal_impl is invoked with an uninitialized CellularSignalHal structure and known NetStatus structure") {
      THEN("the value of NetStatus::rssi is loaded into CellularSignalHal::rssi") {}
      THEN("the value of NetStatus::qual is loaded into CellularSignalHal::qual") {}
    }
  }
}
```

@m-mcgowan may have more input on this topic

test/unit_tests/cellular/cellular.cpp
```diff
@@ -78,6 +99,64 @@ TEST_CASE("cellular_signal()") {
...
+                CellularSignal cs;
+                REQUIRE(cs.fromHalCellularSignal(sig) == true);
+                REQUIRE(cs.getAccessTechnology() == NET_ACCESS_TECHNOLOGY_NONE);
+                REQUIRE(cs.getStrength() < 0.0f);
+                REQUIRE(cs.getStrengthValue() == 0.0f);
+                REQUIRE(cs.getQuality() < 0.0f);
+                REQUIRE(cs.getQualityValue() == 0.0f);
+            }
+
+            SECTION("CellularSignal") {
```
zfields: See previous comment. This feels like `SECTION` abuse.

test/unit_tests/cellular/cellular.cpp
```diff
@@ -435,4 +523,158 @@ TEST_CASE("cellular_signal()") {
...
+            cellular_signal_t sig = {0};
+            sig.size = sizeof(sig);
+            REQUIRE(cellular_signal_impl(nullptr, &sig, true, status) == SYSTEM_ERROR_NONE);
+            REQUIRE(sig.rat == data.expected_act);
+            REQUIRE(sig.strength < 0);
+            REQUIRE(sig.quality < 0);
+            REQUIRE(sig.rsrp == std::numeric_limits<int32_t>::min());
+            REQUIRE(sig.rsrq == std::numeric_limits<int32_t>::min());
+
+            SECTION("CellularSignal") {
```
zfields: `SECTION` abuse

test/unit_tests/cellular/cellular.cpp
```diff
@@ -78,6 +99,64 @@ TEST_CASE("cellular_signal()") {
...
+
+        SECTION("values reported by ACT_UNKNOWN case") {
+            status.rxlev = 10; // seed all values as 10 just to ensure default case is hit
+            status.rxqual = 10;
+            status.rscp = 10;
+            status.ecno = 10;
+            status.rsrp = 10;
+            status.rsrq = 10;
+
+            cellular_signal_t sig = {0};
```
zfields: I wasn't going to mention it, but since you're doing it here, you really should zero initialize all your structs the same way (preferably this way).

test/unit_tests/cellular/cellular.cpp
```diff
@@ -78,6 +99,64 @@ TEST_CASE("cellular_signal()") {
...
+            sig.size = sizeof(sig);
+            REQUIRE(cellular_signal_impl(nullptr, &sig, true, status) == SYSTEM_ERROR_UNKNOWN);
```
zfields: This test seems naked (i.e. it should have it's own `WHEN` `THEN`).

test/unit_tests/cellular/cellular.cpp
```diff
@@ -78,6 +99,64 @@ TEST_CASE("cellular_signal()") {
...
+            REQUIRE(sig.rat == NET_ACCESS_TECHNOLOGY_NONE);
+            REQUIRE(sig.strength == 0);
+            REQUIRE(sig.quality == 0);
+            REQUIRE(sig.rssi == std::numeric_limits<int32_t>::min());
+            REQUIRE(sig.qual == std::numeric_limits<int32_t>::min());
+
+            SECTION("CellularSignal") {
```
zfields: Need more descriptive test names, I have no idea what this is testing.

test/unit_tests/cellular/cellular.cpp
technobly: I followed the convention used in the existing file, they both mean the same thing.  I can change them if we feel it's necessary: https://github.com/catchorg/Catch2/blob/master/docs/test-cases-and-sections.md

test/unit_tests/cellular/cellular.cpp
technobly: When using TEST_CASE and SECTION style, there is no section abuse... that's just how you use SECTION s.

test/unit_tests/cellular/cellular.cpp
technobly: 👍

test/unit_tests/cellular/cellular.cpp
technobly: It's just compartmentalizing the asserts done on the CellularSignal class.  It could be unrolled, but then if someone wanted to add another section later they would have to know to wrap up CellularSignal assertions ahead of time.  The test description comes from hierarchy above: `cellular_signal() - UNKNOWN - values reported by ACT_UNKNOWN case`

test/unit_tests/cellular/cellular.cpp
technobly: It's just one of two defined overloads for the `BufferAppender::append` and I picked the first to be explicit.  The second will do the same thing but in the background.

test/unit_tests/cellular/cellular.cpp
technobly: This is really more of a THEN that is always done before CellularSignal outputs are tested.  Can't wrap it in another nested SECTION or else it won't work correctly.

PR #1938, opened 2019-10-10
[LTE, u-blox] adds a mitigation to keep DNS Client responsive [ch38990]

### LINKED ISSUE #1936: [LTE, u-blox] DNS client fails to return results after a variable period of time via 8.8.8.8:53 over UDP on SARA-R410M-02B
### Expected Behavior

When there is a valid IP connection, DNS lookup on LTE based u-blox SARA-R410M-02B modules should return a valid response.  This can be done via `Cellular.resolve(host)` API or when used internally for resolving the Particle Cloud IP address.

### Observed Behavior

The DNS client fails to return results and continuously times out after a variable period of time via 8.8.8.8:53 over UDP on SARA-R410M-02B.  This can be immediately, 1 hour or potentially 24 hours or more later.  This appears to be a bug in the u-blox modem, potentially with older revisions of firmware.

**Workaround:** Currently, the known recovery mechanism is to detect the DNS lookup failure and power cycle the modem with Cellular.off() / Cellular.on() or call a System.sleep() command that results in the modem powering off temporarily.

### Steps to Reproduce

This can be reproduced with the following test app.

Eventually this test app will start to publish `0.0.0.0` as the resolved IP address, due to DNS lookup timing out with no response from Google DNS.

### Test App

```c
#include "Particle.h"
system_tick_t lastDNSLookup = 0;
void setup() {
}
void loop()
{
    if (millis() - lastDNSLookup >= 120000UL) {
        Serial.println("\r\n\r\nDNS lookup =====================================");
        IPAddress ip = Cellular.resolve("google.com");
        Particle.publish("Google HOST IP", String(ip), PRIVATE, NO_ACK);
        lastDNSLookup = millis();
    }
}
```

✏️ Particle is currently investigating this issue with the intention to implement a fix in the next Device OS release 1.4.1 default, scheduled for October 14th, 2019.

technobly:
### Problem

See #1936 

### Solution

- Open the socket for DNS Client and keep it open, do not close it.
- If the socket send fails, retry once after closing and reopening the socket.
- If the socket send fails a second time, close the socket and return the error.
- Changes polling of socket for pending data to using URCs to indicate pending data with `socketReadable()`
- Automatic socket/handle cleanup on creation of sockets
- Detects state mismatch between system and modem and invalidates DNS Client sock_ handle and also reinitializing mdm_hal sockets[] state
- Removes log noise by closing only untracked socket handles that are open [ch37610]
- Increases USOCO timeout to 130s based on observation that 120s timeout is not being respected by u-blox modem when the command results in CME ERROR.

### Steps to Test

See #1936 and additional tests in app below

### Firmware from this PR (built against 1.4.1-rc.1)
⬇️ [particle-device-os@1.4.1-rc.1+debug.pr1938.zip](https://www.dropbox.com/s/ri3od0yht8fnxe3/particle-device-os%401.4.1-rc.1%2Bdebug.pr1938.zip?dl=0)

### Test app

```c++
#include "Particle.h"
#include "socket_hal.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
system_tick_t lastDNSLookup = 0;

void setup() {
    pinMode(B1, INPUT_PULLUP);
    pinMode(B2, INPUT_PULLUP);
    pinMode(B3, INPUT_PULLUP);
    Particle.connect();

    // Open a TCP socket just to test
    Serial.println("\r\n\r\nOPEN TCP SOCKET =====================================");
    int port = 80;
    socket_create(AF_INET, SOCK_STREAM, IPPROTO_TCP, port, NIF_DEFAULT);
}

void loop()
{
    if (millis() - lastDNSLookup >= 10000UL) {
        Serial.println("\r\n\r\nDNS lookup =====================================");
        Cellular.command("AT+CCLK?\r\n"); // timestamp
        IPAddress ip = Cellular.resolve("google.com");

        while (digitalRead(B3) == LOW); // Pull antenna and wait for AT command failure before continuing
        // Checks DNS Client & Modem state mismatch

        Particle.publish("Google HOST IP", String(ip), PRIVATE, NO_ACK);
        lastDNSLookup = millis();

        // Check return path after DNS failure
        if (digitalRead(B1) == LOW) {
            Particle.publish("Google HOST IP", String(ip), PRIVATE, WITH_ACK);
            while (digitalRead(B1) == LOW) Particle.process();
            delay(100);
        }

        // Checks DNS Client & Modem state mismatch
        if (digitalRead(B2) == LOW) {
            Cellular.off();
            Cellular.on();
            delay(100);
            while (digitalRead(B2) == LOW) Particle.process();
            delay(100);
        }

        // press RESET to check MCU / DNS Client & Modem state mismatch
    }
}
```

### References

Closes #1936 
Closes [ch38990]()
Closes [ch37610]()

---

### Completeness

---

- [bugfix] [LTE, u-blox] adds a mitigation to keep DNS Client responsive [ch38990]() [#1938](https://github.com/particle-iot/device-os/pull/1938)
- [internal] [LTE] Removes log noise by closing only untracked socket handles that are open [ch37610]() [#1938](https://github.com/particle-iot/device-os/pull/1938)

### DIFF
--- a/hal/src/electron/modem/dns_client.cpp
+++ b/hal/src/electron/modem/dns_client.cpp
@@ -130,7 +130,9 @@ struct DnsQuery {
 class UdpSocket {
 public:
     UdpSocket() :
-            sock_(-1) {
+            sock_(MDM_SOCKET_ERROR),
+            oldModemStateChangeCount_(0),
+            newModemStateChangeCount_(0) {
     }
 
     ~UdpSocket() {
@@ -138,6 +140,7 @@ public:
     }
 
     int open() {
+        _isSocketValid(); // invalidates handle if necessary so we can re-open
         if (sock_ >= 0) {
             return 0; // already open
         }
@@ -150,21 +153,30 @@ public:
             return MDM_SOCKET_ERROR;
         }
         sock_ = sock;
+        oldModemStateChangeCount_ = newModemStateChangeCount_;
         return 0;
     }
 
     void close() {
-        if (sock_ >= 0) {
+        if (_isSocketValid() && sock_ >= 0) {
             electronMDM.socketFree(sock_);
-            sock_ = -1;
+            sock_ = MDM_SOCKET_ERROR;
         }
     }
 
     int sendTo(MDM_IP addr, int port, const char* data, size_t size) {
+        if (!_isSocketValid()) {
+            // do not try to free socket since it is invalidated
+            return MDM_SOCKET_ERROR;
+        }
         return electronMDM.socketSendTo(sock_, addr, port, data, size);
     }
 
     int recvFrom(MDM_IP* addr, int* port, char* data, size_t size) {
+        if (!_isSocketValid()) {
+            // do not try to free socket since it is invalidated
+            return MDM_SOCKET_ERROR;
+        }
         // See if any data is waiting for us
         int result = electronMDM.socketReadable(sock_);
         if (result <= 0) {
@@ -175,7 +187,20 @@ public:
     }
 
 private:
+    bool _isSocketValid() {
+        // Used to detect when the modem has been power cycled off/on/reset since the last time
+        // sock_ has been used, and if it has changed the sock_ handle is invalidated.
+        newModemStateChangeCount_ = electronMDM.getModemStateChangeCount();
+        if (oldModemStateChangeCount_ != newModemStateChangeCount_) {
+            sock_ = MDM_SOCKET_ERROR;
+            return 0;
+        }
+        return 1;
+    }
+
     int sock_;
+    int oldModemStateChangeCount_;
+    int newModemStateChangeCount_;
 };
 
 UdpSocket sock;
--- a/hal/src/electron/modem/dns_client.cpp
+++ b/hal/src/electron/modem/dns_client.cpp
@@ -140,7 +140,7 @@ public:
     }
 
     int open() {
-        _isSocketValid(); // invalidates handle if necessary so we can re-open
+        _invalidateSocketIfNeeded();
         if (sock_ >= 0) {
             return 0; // already open
         }
@@ -158,14 +158,14 @@ public:
     }
 
     void close() {
-        if (_isSocketValid() && sock_ >= 0) {
+        if (!_invalidateSocketIfNeeded() && sock_ >= 0) {
             electronMDM.socketFree(sock_);
             sock_ = MDM_SOCKET_ERROR;
         }
     }
 
     int sendTo(MDM_IP addr, int port, const char* data, size_t size) {
-        if (!_isSocketValid()) {
+        if (_invalidateSocketIfNeeded()) {
             // do not try to free socket since it is invalidated
             return MDM_SOCKET_ERROR;
         }
@@ -173,7 +173,7 @@ public:
     }
 
     int recvFrom(MDM_IP* addr, int* port, char* data, size_t size) {
-        if (!_isSocketValid()) {
+        if (_invalidateSocketIfNeeded()) {
             // do not try to free socket since it is invalidated
             return MDM_SOCKET_ERROR;
         }
@@ -187,15 +187,15 @@ public:
     }
 
 private:
-    bool _isSocketValid() {
+    bool _invalidateSocketIfNeeded() {
         // Used to detect when the modem has been power cycled off/on/reset since the last time
         // sock_ has been used, and if it has changed the sock_ handle is invalidated.
         newModemStateChangeCount_ = electronMDM.getModemStateChangeCount();
         if (oldModemStateChangeCount_ != newModemStateChangeCount_) {
             sock_ = MDM_SOCKET_ERROR;
-            return 0;
+            return 1;
         }
-        return 1;
+        return 0;
     }
 
     int sock_;
@@ -203,7 +203,7 @@ private:
     int newModemStateChangeCount_;
 };
 
-UdpSocket sock;
+UdpSocket g_sock;
 
 inline uint16_t htons(uint16_t val) {
 #if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
@@ -451,7 +451,7 @@ int getHostByName(const char* name, MDM_IP* addr) {
     if (!q.buf) {
         return SYSTEM_ERROR_NO_MEMORY;
     }
-    int ret = sock.open();
+    int ret = g_sock.open();
     if (ret < 0) {
         LOG(ERROR, "Unable to create socket");
         return ret;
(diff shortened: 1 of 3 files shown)


hal/src/electron/modem/dns_client.cpp
```diff
@@ -162,13 +165,21 @@ class UdpSocket {
     }
 
     int recvFrom(MDM_IP* addr, int* port, char* data, size_t size) {
+        // See if any data is waiting for us
+        int result = electronMDM.socketReadable(sock_);
+        if (result <= 0) {
```
avtolstoy: Shouldn't the negative errors be propagated as-is instead?

hal/src/electron/modem/dns_client.cpp
```diff
@@ -137,7 +138,9 @@ class UdpSocket {
     }
 
     int open() {
-        close();
+        if (sock_ >= 0) {
```
avtolstoy: What happens if the modem is power-cycled? The handle will no longer be valid or what's even worse we might be writing into the wrong socket handle.

hal/src/electron/modem/dns_client.cpp
technobly: This mimics the same operation as `socket_receivefrom()` in `socket_hal.cpp`.  Even though `socketReadable()` returns -1 most of the time when there is no data, we interpret that as no data, same as 0.  We wouldn't want to detect -1 and close the socket either.  We need to return 0 specifically to keep this code from bailing out:
```c++
ret = recvDnsResponse(&sock, &q);
if (ret < 0) {
    return ret;
}
```

hal/src/electron/modem/dns_client.cpp
technobly: I'll make a change here based on our conversation this morning, adding a function the DNS Client can call in `mdm_hal.cpp` to know if the modem was power cycled.  If the modem is power cycled the DNS Client's `sock_` handle will be invalidated so that a new socket is opened.

hal/src/electron/modem/dns_client.cpp
avtolstoy: > Even though `socketReadable()` returns -1 most of the time when there is no data

I thought it returned `0` if there is no data and `-1` in case of errors last time I looked at its code, so that's why I suggested short-circuiting on negative values.

hal/src/electron/modem/dns_client.cpp
technobly: Ok, I pushed a commit with a bunch of improvements: automatic socket/handle cleanup on creation of sockets, detects state mismatch between system and modem, and reduces socket log noise [ch37610].

hal/src/electron/modem/dns_client.cpp
technobly: I'm going to mark this one resolved, please let me know if you think there is more to do here.

sergeuz: I still don't understand why we're hiding a potential error here :man_shrugging:

hal/src/electron/modem/dns_client.cpp
```diff
@@ -147,28 +153,58 @@ class UdpSocket {
...
+        }
+        return 1;
+    }
+
     int sock_;
+    int oldModemStateChangeCount_;
+    int newModemStateChangeCount_;
 };
 
+UdpSocket sock;
```
sergeuz: Let's either use a longer name for this variable or add a `g_` prefix to it. I wouldn't expect `sock` to be a global variable looking at the `getHostByName()` code.

hal/src/electron/modem/dns_client.cpp
```diff
@@ -433,7 +478,7 @@ int getHostByName(const char* name, MDM_IP* addr) {
             }
         } while (HAL_Timer_Get_Milli_Seconds() - timeStart < timeout);
         ++retries;
-    } while (retries <= MAX_RETRIES);
+    } while (retries <= MAX_DNS_QRY_RETRIES);
     return SYSTEM_ERROR_TIMEOUT;
```
sergeuz: Given that the issue addressed in this PR is specific to LTE modems and we plan to use this DNS client on 2G/3G Electrons as well, I'd probably close the socket on unaffected platforms here.

technobly: Since this is such an integral part of the system (DNS lookup for the Cloud and use of UDP/TCP Clients, how about we reserve this socket just like we do for the Cloud).   Sure that means one less socket for users, but there are still 5 additional sockets on Electrons (7 total) and that's quite a lot.

sergeuz: Sure, that makes sense too :+1: Also, implementing this logic is outside the scope of this PR.

hal/src/electron/modem/dns_client.cpp
```diff
@@ -147,28 +153,58 @@ class UdpSocket {
...
+        int result = electronMDM.socketReadable(sock_);
+        if (result <= 0) {
+            return 0; // no data
+        }
+
         return electronMDM.socketRecvFrom(sock_, addr, port, data, size);
     }
 
 private:
+    bool _isSocketValid() {
```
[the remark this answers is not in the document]
technobly: yeah I struggled with that, because then if we use your name suggestion it doesn't make as much sense in other parts of the code:
```c++
    void close() {
        if (_isSocketValid() && sock_ >= 0) {
            electronMDM.socketFree(sock_);
            sock_ = MDM_SOCKET_ERROR;
        }
    }
```

I'll think about how to resolve this..

hal/src/electron/modem/dns_client.cpp
technobly: Sounds good to me 😎

hal/src/electron/modem/dns_client.cpp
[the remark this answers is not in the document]
technobly: I'm going to use your name, and invert the `bool` return values.  I think that makes it much more tidy, thanks!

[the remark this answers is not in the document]
sergeuz: I typically make such methods return void, as the validity of an object in question usually can be checked by other means as well:
```
void close() {
    invalidateSocketIfNeeded();
    if (sock_ >= 0) {
        electronMDM.socketFree(sock_);
        sock_ = MDM_SOCKET_ERROR;
    }
}
```
Anyway, I'm just nitpicking here :)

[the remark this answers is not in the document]
technobly: Yeah I was trying not to duplicate `sock_ = MDM_SOCKET_ERROR;` and tests for `sock_ >= 0` in multiple places.

PR #1941, opened 2019-10-11
[electron] enables modem logs in non-debug builds [ch35402]

avtolstoy:
### Problem

Electron modem logs are only enabled in debug builds (DEBUG_BUILD=y). They are useful, but we don't want to keep using debug binaries as default release binaries.

### Solution

This PR modifies `mdm_hal.cpp` to use non-debug `LOG_PRINTF` calls instead of `DEBUG_D`.

### Steps to Test

Build example app, modem logs should be output on `Serial`.

### Example App

```c++

void setup() {
}

void loop() {

}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/electronserialpipe_hal.cpp
+++ b/hal/src/electron/modem/electronserialpipe_hal.cpp
@@ -265,13 +265,13 @@ int ElectronSerialPipe::rxSize(void)
 
 void ElectronSerialPipe::txDump(void)
 {
-    DEBUG_D("TX ");
+    MDM_PRINTF("TX ");
     _pipeTx.dump();
 }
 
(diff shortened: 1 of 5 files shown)


technobly:
@sergeuz is there a way we can add a logging category here with the LOG_PRINTF direct output?  It doesn't have to happen for this PR, but it might be nice in the future to be able to disable the AT output and keep the upper levels.

sergeuz:
@technobly Yeah, there's a macro called `LOG_PRINTF_C()`. I won't mind updating this PR though, it's a trivial fix

avtolstoy:
[CH35402]

PR #1943, opened 2019-10-14
Implement a command to reset all network interfaces of a device remotely

sergeuz:
### Problem

We'd like to have a way to reset the modem of a cellular device remotely. This PR extends the existing system handler for `device/reset` events to support a new reset type (`network`), which toggles all active network interfaces off an on.

### Steps to Test

Testing requires manipulating the Device Service to send the above event to a device. As of now, there's no easy way to send a reserved event to a device.

### References

- [ch39772]

### DIFF
--- a/system/src/system_network_internal.cpp
+++ b/system/src/system_network_internal.cpp
@@ -23,6 +23,7 @@ namespace particle {
 
 namespace {
 
+[[gnu::unused]] // Suppress a warning on the newhal platform
 bool turnOffNetworkIfNeeded(network_interface_index iface) {
     if (network_ready(iface, NETWORK_READY_TYPE_ANY, nullptr) || network_connecting(iface, 0, nullptr)) {
         network_off(iface, 0, 0, nullptr);
--- a/system/src/system_network_internal.cpp
+++ b/system/src/system_network_internal.cpp
@@ -36,6 +36,7 @@ void networkOn(network_interface_index iface) {
 } // namespace
 
 void resetNetworkInterfaces() {
+    // TODO: There's no cross-platform API to enumerate available network interfaces
 #if HAL_PLATFORM_MESH
     const bool resetMesh = networkOff(NETWORK_INTERFACE_MESH);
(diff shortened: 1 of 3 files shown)


system/src/system_network_internal.cpp
```diff
@@ -17,6 +17,62 @@
...
+void networkOn(network_interface_index iface) {
+    network_on(iface, 0, 0, nullptr);
+}
+
+} // namespace
+
+void resetNetworkInterfaces() {
+    // TODO: There's no cross-platform API to enumerate available network interfaces
+#if HAL_PLATFORM_MESH
+    const bool resetMesh = networkOff(NETWORK_INTERFACE_MESH);
```
avtolstoy: Why not simply call `network_off()` with `NETWORK_INTERFACE_ALL` ?

system/src/system_network_internal.cpp
```diff
@@ -17,6 +17,62 @@
 
 #include "system_network_internal.h"
 
+namespace particle {
+
+namespace {
+
+bool networkOff(network_interface_index iface) {
```
avtolstoy: `newhal` platforms seems to fail building, these should probably get wrapped in `SPARK_NO_CLOUD` or something like that.

system/src/system_network_internal.cpp
sergeuz: I didn't want to change the state of interfaces which are turned on but have no link. The system network API doesn't provide much visibility into an interface's state, unfortunately. On second thought, I think we should also invoke `network_connect()` for affected interfaces in this function.

system/src/system_network_internal.cpp
sergeuz: :+1:

system/src/system_network_internal.cpp
avtolstoy: Ah, okay, makes sense. :+1: I hope we'll be able to resolve that in Q4. I have a WIP branch that should allow to manage/query individual states.

PR #1946, opened 2019-10-15
Extract version from release scripts

zfields:
### Problem

Currently, version information is captured in the `./build/release.sh` script. This makes sharing the version information with other consumers unnecessarily difficult.

### Solution

Extract version information into its own file, then refactor the release scripts, `./build/release.sh` and `./build/release-tests.sh` to consume the information from that file.

### Steps to Test

Invoke the `./build/make_release.sh` with the `--tests` flag.

### Example App

```bash
./make_release.sh --platform photon --tests
```

### References

- [ch39831](https://app.clubhouse.io/particle/story/39831/refactor-release-sh-and-release-publish-sh-to-consume-the-new-version-file) - Refactor `release.sh` and `release-publish.sh` to consume the new version file
- Preparation for build system unification.

---

### Completeness

- [x] Added documentation

### DIFF
--- a/version.semver.json
+++ b/version.semver.json
@@ -1,0 +1,15 @@
+{
+    "semver": {
+        "spec": "2.0",
+        "version": {
+            "major": 1,
+            "minor": 4,
+            "patch": 1,
+            "pre-release": [
+                "rc",
+                "1"
+            ],
+            "build-metadata": []
+        }
+    }
+}
--- a/build/print-version.sh
+++ b/build/print-version.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+VERSION_FILE="../version.semver.json"
+
+# Confirm version file exists and is readable
+if [ ! -r "$VERSION_FILE" ]; then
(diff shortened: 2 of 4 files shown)


version.semver.json
```diff
@@ -0,0 +1,15 @@
+{
+    "semver": {
```
avtolstoy: I may very well be asking a dumb question, but why so much complexity? Why not simply `1.4.1.-rc.1`?

zfields: My goal is not to make it complex, but explicit. The idea is to provide a machine readable way to reason about the version information.

As you know, a tag is sufficiently unique to work with our external tools as they are today, and indeed the `print-version.sh` script is intended to generate the such a tag, `1.4.1-rc.1`.

I elected to go this route, so our tools can evolve to have more complex reasoning around version information _(i.e. testing implicit API compatibility guaranteed by the SemVer notation)_. Since this information is versioned, it is guaranteed to be available and easily parsable by future tools from this point forward.

zfields:
The version information is already available to Concourse, so this work is unnecessary at this time.

PR #1949, opened 2019-10-18
[photon] wlan: add wiced_wlan_connectivity_initialized() check in wlan_fetch_ipconfig()

### LINKED ISSUE #1858: System thread causes macAddress function call in setup routine to fault with an SOS
## Bug Report

Calling Wifi.macAddress in the setup routine with the system thread enabled causes an instant SOS type 1 fault. If I move the function call below a ```waitUntil(Particle.connected);``` the fault goes away but I would like to obtain the MAC before the device is connected. I suspect there is a threading issue here. Maybe this is related to #1805?

### Expected Behavior

Wifi.macAddress() should return the device MAC address when called in the setup routine.

### Observed Behavior

Calling Wifi.macAddress() in the setup routine with the system thread enabled causes an SOS type 1.

### Steps to Reproduce

Flash the test app below. Note: disabling the system_thread fixes the issue.

### Test App

```c
SYSTEM_THREAD(ENABLED);

byte mac[6];

void setup() {
    WiFi.macAddress(mac);
}

void loop() {

}
```

### References

Maybe related to #1805

avtolstoy:
### Problem

https://github.com/particle-iot/device-os/issues/1858

### Solution

Calling `wiced_wifi_get_mac_address())` or potentially other functions before `wiced_wlan_connectivity_initialized()` returns true seems to lead to a hardfault. Add `wiced_wlan_connectivity_initialized()` check.

### Steps to Test

Test app shouldn't cause a hardfault.

### Example App

```c++
SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup() {
    WiFi.on();
    uint8_t mac[6] = {};
    WiFi.macAddress(mac);
}

void loop() {
    Particle.process();
}
```

### References

- Closes #1858 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/photon/wlan_hal.cpp
+++ b/hal/src/photon/wlan_hal.cpp
@@ -1395,25 +1395,28 @@ int wlan_fetch_ipconfig(WLanConfig* config)
 #endif // LWIP_DHCP
     }
 
-    wiced_mac_t my_mac_address;
-    if (wiced_wifi_get_mac_address( &my_mac_address)==WICED_SUCCESS)
-        memcpy(config->nw.uaMacAddr, &my_mac_address, 6);
+    // XXX: calling wiced_wifi_get_mac_address or potentially other functions
+    // before wiced_wlan_connectivity_initialized() returns true will lead to a hardfaul
+    if (wiced_wlan_connectivity_initialized()) {
+        wiced_mac_t my_mac_address;
+        if (wiced_wifi_get_mac_address( &my_mac_address)==WICED_SUCCESS)
+            memcpy(config->nw.uaMacAddr, &my_mac_address, 6);
 
-    wl_bss_info_t ap_info;
-    wiced_security_t sec;
+        wl_bss_info_t ap_info;
+        wiced_security_t sec;
 
-    if ( wwd_wifi_get_ap_info( &ap_info, &sec ) == WWD_SUCCESS )
-    {
-        uint8_t len = std::min(ap_info.SSID_len, uint8_t(32));
-        memcpy(config->uaSSID, ap_info.SSID, len);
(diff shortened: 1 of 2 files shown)


YutingYou:
I ran some other API tests and got Hardfault, it's easy to reproduce under poor signal(removing antenna).
```
SYSTEM_THREAD(ENABLED);

byte mac[6];
byte bssid[6];

void setup() {
    (void)WiFi.macAddress(mac);
    (void)WiFi.wifi_config();
    (void)WiFi.localIP();
    (void)WiFi.subnetMask();
    (void)WiFi.gatewayIP();
    (void)WiFi.dnsServerIP();
    (void)WiFi.dhcpServerIP();
    (void)WiFi.RSSI();
    (void)WiFi.getAntenna();

    (void)WiFi.SSID();           // ERROR: wiced_join_ap_specific(), result: 1025
    (void)WiFi.BSSID(bssid);     // ERROR: wiced_join_ap_specific(), result: 1024
                                 // ERROR: wiced_join_ap_specific(), result: 1006
}
```

avtolstoy:
:disappointed: Thanks for looking into this. There are probably some other places that also require a similar guard check.

YutingYou:
Since it's not open source, sometimes it's hard to predict the behavior. 🤷‍♀️

avtolstoy:
@YutingYou I've added a test to `wiring/no_fixture` and also tried running the app you posted under the same conditions and couldn't reproduce the problem. Can I ask you to collect the stacktrace if it's reproducible for you?

YutingYou:
Approve this PR since I cannot reproduce the Hardfault issue. And those three error codes don't affect the network connection.

PR #1954, opened 2019-10-30
Fix spelling in comments

matthewbeckler:
### Problem

Fix a common spelling mistake in some comments, basically:
* `s/recieving/receiving/g`
* `s/recieve/receive/g`

### Solution

Puts the `e` before `i`, due to being after `c`

### Steps to Test

No testing needed. Changes are only to comments, and should produce binary-identical builds.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/communication/inc/dsakeygen.h
+++ b/communication/inc/dsakeygen.h
@@ -30,7 +30,7 @@ extern "C" {
 
 /**
  * Generates a DER formatted RSA PCKS#1 private key.
- * @param buffer            The buffer to recieve the DER file.
+ * @param buffer            The buffer to receive the DER file.
  * @param max_length        The length of the buffer
  * @param f_rng             A random number generator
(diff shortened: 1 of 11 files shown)


matthewbeckler:
Hi all, is there something else I need to do to trigger the `coveralls` check? Thanks much.

avtolstoy:
@matthewbeckler Thanks for the PR!

It's a security feature, so coveralls won't run for external contributions.

I'm sure others will chime in as well, but I'd be against modifying anything in the third-party headers/sources (in this case anything under `hal/src/photon/wiced/`), despite the fact that they are pulled in our tree.

matthewbeckler:
@avtolstoy Thanks much, that all makes sense. I'll remove the changes under `hal/src/photon/wiced/` and look into suggesting those changes to the third-party repos.

PR #1956, opened 2019-11-05
Implement HandshakeComplete message

sergeuz:
### Problem

At present, there is no explicit indication that the application handshake has completed. The cloud does not know with any certainty that the device has completed the application handshake and considers itself online. The possibility of out of order delivery of events compounds this problem even further.

### Solution

This PR changes the handshake procedure so that it looks as follows:

1. the application handshake is performed and various messages are sent to the cloud
2. the system waits for these messages to be acknowledged
3. once all messages are acknowledged,  a new `HandshakeComplete` CoAP message is sent to the cloud.
4. on acknowledgement of that message, `Particle.connected()` returns true.

### Steps to Test

_TBD_

### References

- [ch40066]

### DIFF
--- a/communication/src/protocol.cpp
+++ b/communication/src/protocol.cpp
@@ -292,8 +292,8 @@ int Protocol::begin()
 	ack_handlers.clear();
 	last_ack_handlers_update = callbacks.millis();
 
-	uint32_t channel_flags = 0;
-	ProtocolError error = channel.establish(channel_flags, application_state_checksum());
+	flags &= ~SKIP_SESSION_RESUME_HELLO;
+	ProtocolError error = channel.establish(flags, application_state_checksum());
 	bool session_resumed = (error==SESSION_RESUMED);
 	if (error && !session_resumed) {
 		LOG(ERROR,"handshake failed with code %d", error);
@@ -304,9 +304,6 @@ int Protocol::begin()
 	{
 		// for now, unconditionally move the session on resumption
 		channel.command(MessageChannel::MOVE_SESSION, nullptr);
-		if (channel_flags & SKIP_SESSION_RESUME_HELLO) {
-			flags |= SKIP_SESSION_RESUME_HELLO;
-		}
 	}
 
 	// hello not needed because it's already been sent and the server maintains device state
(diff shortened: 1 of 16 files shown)


system/src/system_task.cpp
```diff
@@ -347,79 +387,100 @@ int cloud_handshake()
...
+                    g_cloudHandshakeState = CloudHandshakeState::DONE;
+                }
+            } else if (err == protocol::ProtocolError::NOT_IMPLEMENTED) { // Core?
+                g_cloudHandshakeState = CloudHandshakeState::DONE;
+                err = 0;
+            }
+        }
+        if (g_cloudHandshakeState == CloudHandshakeState::DONE) {
+            // TODO: There's no protocol API to get the current session state, so we're running
+            // one more iteration of the communication loop to make sure all handshake messages
```
m-mcgowan: I am kind of puzzled by this one too!

communication/src/protocol.cpp
```diff
@@ -292,8 +292,8 @@ int Protocol::begin()
 	ack_handlers.clear();
 	last_ack_handlers_update = callbacks.millis();
 
-	uint32_t channel_flags = 0;
-	ProtocolError error = channel.establish(channel_flags, application_state_checksum());
+	flags &= ~SKIP_SESSION_RESUME_HELLO;
```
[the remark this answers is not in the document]
sergeuz: This fixes a bug where the `SKIP_SESSION_RESUME_HELLO` flag was persisting through multiple connection attempts.

system/src/system_task.cpp
sergeuz: This is not a new code, but, yeah, the comment could be more clear. The problem is that when a CoAP message expires, the event loop function doesn't return an error (instead, the protocol implementation [closes the channel](https://github.com/particle-iot/device-os/blob/2a8f892646022f50f9e5572a8132811961ed6111/communication/src/coap_channel.cpp#L57-L63) silently), and we have to invoke the loop function one more time to determine if the current protocol session is still valid.

sergeuz:
I'm closing this PR as it's unlikely for it to land in any Device OS release in its current form in foreseeable future.

Issue #1150, opened 2016-10-29
Low Power Modes API


technobly:
## **Goal:** 
>propose a Low Power Modes API that is intuitive, hardware/device agnostic as possible, camelCase and easy to remember.

**This issue is related to these issues pertaining to different low power modes:**
#xxxx - Wi-Fi Eco mode
#xxxx - Cellular Eco Mode
#xxxx - System Eco mode
#922  - Wi-Fi powersave mode on Photon/P1
#935  - Low Power Run mode

| API | What it does |
|:--|:--|
| WiFi.enableEco() | Turns Eco mode on with a 500ms[1] throughput delay by default |
| WiFi.enableEco(10) | Turns Eco mode on with a throughput delay of 10 milliseconds (drops to low power mode 10 milliseconds after TX/RX), pass a 0 for no throughput delay. |
| WiFi.disableEco() | Turns Eco mode off, resets any currently set throughput delay |
| WiFi.ecoEnabled() | Return if Eco mode is on, off or on with throughput delay |
| Cellular.enableEco() | Turns Eco mode on |
| Cellular.disableEco() | Turns Eco mode off |
| Cellular.ecoEnabled() | Return if Eco mode is enabled |
| System.enableEco() | Put system in Eco mode, until System.run() is called. |
| System.enableEco(60) | Put system in Eco mode for 60 seconds or until System.disableEco() is called, code can still execute, limited peripherals |
| System.disableEco() | Put system back in full power operation mode |
| System.ecoEnabled() | Return if Eco mode is enabled |

| Future API | What it does |
|:--|:--|
| System.setSpeed(clockSpeed) | Sets the system clockSpeed, e.g. CLOCK_SPEED_120MHZ |
| System.enablePeripheral(systemPeripheral) | Enable the systemPeripheral, e.g. PERIPHERAL_SPI. Only takes effect if System.ecoEnabled() is true |
| System.disablePeripheral(systemPeripheral) | Disable the systemPeripheral, e.g. PERIPHERAL_SPI. Only takes effect if System.ecoEnabled() is true |
| System.setConfig(SystemConfig& profile) | Change the default configuration used in System.enableEco() to enable/disable other | clock/peripheral settings. This API that combines the use of other future APIs: System.setSpeed() and System.enable/disablePeripheral() as a single command |
| System.getConfig(SystemConfig& profile) | Retreive the currently set system configuration |

[...]

Issue #1149, opened 2016-10-27
Sleep API Refactoring


technobly:
## **Goal:** 
>refactor Sleep API's to remove problems, make intuitive and easy to remember.

**This issue is related to these previous issues:**

#501 - Additional sleep modes and sleep API review
#502 - Wiring API supports multiple Network types 
#251 - Comment on Network.on/off vs WiFi.on/off

**As a recap, the current Sleep API looks like this:**

| API | What it does | Network | Application |
|:-:|:-:|:-:|:-:|
| System.sleep(60) | Put the NETWORK module to sleep for 60s. Application code continues running. | OFF | RUNNING |
| System.sleep(D2, RISING, 60) | Pause execution, using less power. Resume execution with the next statement after 60s, or when D2 transitions from LOW to HIGH, whichever is sooner. | OFF | PAUSED |
| System.sleep(D2, RISING, 60, SLEEP_NETWORK_STANDBY) | Same as above with network connection still active.  Only available on Electron. | ON | PAUSED |
| System.sleep(SLEEP_MODE_DEEP, 60) | Stop execution, using much less power. Restart the system after 60 seconds or on a rising edge on the WKP pin (photon/P1/electron only) or falling edge on RESET pin.  The application restarts - memory is cleared, setup() is called.) | OFF | TERMINATED |
| System.sleep(SLEEP_MODE_DEEP, 60, SLEEP_NETWORK_STANDBY) | Same as above with network connection still active.  Only available on Electron. | ON | TERMINATED |
| System.sleep( SLEEP_MODE_SOFTPOWERDOWN, 60) | Same as above with network connection off and also sleeps the Fuel Gauge further reducing power.  Only available on Electron. | OFF | TERMINATED |

**The current Sleep API has these issues:**

| API | Issue |
|:-:|:-:|
| System.sleep(60) | Doesn’t stop application code.  Not intuitive. |
| System.sleep(D2… | Doesn’t intuitively convey STOP mode which is a STM naming convention. This corresponds to application level naming convention PAUSE. |
| System.sleep(SLEEP_MODE_DEEP... | Doesn’t intuitively convey STANDBY mode which is a STM naming convention. Hard to remember. This corresponds to application level naming convention DEEPSLEEP. |
[...]

PR #1965, opened 2019-11-13
Refactors system sleep implementation

XuGuohui:
### Problem

Currently, both HAL and Wiring sleep APIs are not very extensible and support only a subset of possible modes and wake-up options. It has been a long time goal for us to improve the sleep APIs and we’ve already taken some steps in that direction e.g.:
- Wake up reason
- Multiple wake-up pins support for STOP mode sleep
- Type-safe sleep flags in wiring

#### Sleep Modes

- Stop mode:
  - What's disabled: The resources occupied by system, e.g. CPU, RGB, external flash etc., and network interfaces that is not featured as wakeup source.
  - Wakeup sources: Any source that the platform supported.
  - On-exit: It resumes the disabled resources and continue running.

- Ultra-Low power mode:
   - What's disabled: The resources occupied by system and network interface and all other resources those are not featured as wakeup source.
   - Wakeup sources: Any source that the platform supported
   - On-exit: It resumes the disabled resources (by restoring peripherals' configuration) and network connection if necessary and continue running.

- Hibernate mode:
  - What's disabled: Most of resources except particular wakeup source and retention RAM
  - Wakeup sources: Particular pins and NFC if supported
  - On-exit: Reset

In comparison to the stop mode, the ultra-low power mode will additionally disable internal peripherals that is not featured as wakeup source. before disabling them, Device OS will cache those peripherals' configurations and restore them on exiting sleep mode.

### References

- Sleep API refactoring: https://github.com/particle-iot/device-os/issues/1149
- Power levels: https://github.com/particle-iot/device-os/issues/1150

---

### Completeness

- [ ] Added documentation


hal/inc/core_hal.h
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
...
+    HAL_CORE_WAKEUP_SOURCE_TYPE_USB = 11,
+    HAL_CORE_WAKEUP_SOURCE_TYPE_BLE = 12,
+    HAL_CORE_WAKEUP_SOURCE_TYPE_NFC = 13,
+    HAL_CORE_WAKEUP_SOURCE_TYPE_NETWORK = 14,
+    HAL_CORE_WAKEUP_SOURCE_TYPE_MAX = 0x7FFFF
+} hal_core_wakeup_source_type_t;
+static_assert(sizeof(hal_core_wakeup_source_type_t) == 2, "length of hal_core_wakeup_source_type_t should be 2-bytes aligned.");
+
+/**
+ * Stop mode:
```
XuGuohui: This mode aims to make user be aware of what peripherals keep running and consuming power in sleep mode and system will fully manage the reset of peripherals to save more power as possible.

hal/inc/core_hal.h
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
...
+ *     Only the wakyup sources specified by user are enabled during in stop mode.
+ *     Other peripherals will be temporarily disabled, but their configurations are cached
```
XuGuohui: Caching the peripheral's configurations, for example the UART baud-rate, SPI mode etc., so that once the device is woke up user application can call UART/SPI transmission APIs directly.

wiring/src/spark_wiring_system.cpp
```diff
@@ -41,6 +41,52 @@ void SystemClass::reset(uint32_t data)
...
+    bool cloudAutoConnect;
+#endif
+    bool wifiEnabled;
+    bool wifiConnect;
+
+    if (config.sleepMode() == SystemSleepMode::NONE || !config.halCoreSleepConfig()->wakeup_sources) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+
+    if ((config.sleepMode() == SystemSleepMode::STOP && !config.wakeupSourceFeatured(HAL_CORE_WAKEUP_SOURCE_TYPE_NETWORK)) ||
```
XuGuohui: This piece of code is just a copy from the `system_sleep.cpp`. In sleep 2.0, the `system_sleep.cpp` and `system_sleep.h` will be deleted. `spark_wiring_system.cpp` will call the core HAL APIs directly.

hal/inc/core_hal.h
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
 
 void HAL_Set_Event_Callback(HAL_Event_Callback callback, void* reserved);
 
+
+#define HAL_CORE_SLEEP_VERSION 2
+
+typedef enum hal_core_wakeup_source_type_t {
```
avtolstoy: Do you think it makes sense to move all these definitions into a separate header (as well as implementation into a separate source file) to keep `core_hal` clean?

hal/inc/core_hal.h
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
...
+ *     Only certain GPIO can be the wakeup source.
+ *     CPU, every peripherals, system clock and all flash and RAM memory are completely shut off.
+ *     Once device is wakeup, it performs a reset.
+ */
+typedef enum hal_core_sleep_mode_t {
+    HAL_CORE_SLEEP_MODE_NONE = 0,
+    HAL_CORE_SLEEP_MODE_STOP = 1,
+    HAL_CORE_SLEEP_MODE_HIBERNATE = 2,
+    HAL_CORE_SLEEP_MODE_SHUTDOWN = 3,
+    HAL_CORE_SLEEP_MODE_MAX = 0x7FFFF
```
avtolstoy: Probably should be fine to leave it at 0x7f max, I don't expect us to have that many sleep modes :)

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+    }
+
+    // Setters
+    SystemSleepConfiguration& mode(SystemSleepMode mode) {
+        config_.mode = static_cast<hal_core_sleep_mode_t>(mode);
+        return *this;
+    }
+
+    SystemSleepConfiguration& gpio(pin_t pin, InterruptMode mode) {
+        clearGpioWakeupSources();
```
XuGuohui: The reason why it need to clear the previous configured wakeup source is in case that the `SystemSleepConfiguration` object is reused to configured a different pin. The same mechanism is applied to the RTC wakeup source.

hal/inc/core_hal.h
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
...
+} hal_core_sleep_mode_t;
+static_assert(sizeof(hal_core_sleep_mode_t) == 2, "length of hal_core_sleep_mode_t should be 2-bytes aligned.");
+
+/**
+ * HAL sleep wakeup source: GPIO
+ */
+typedef struct hal_core_wakeup_source_gpio_t {
+    uint16_t pin;
```
avtolstoy: All these wakeup sources need to use C-style inheritance and need to have `hal_core_wakeup_source_t` at the top:
```c++
typedef struct hal_core_wakeup_source_gpio_t {
    struct hal_core_wakeup_source_t; // anonymous
    uint16_t pin;
    InterruptMode mode; // Caution: This might not be 1-byte length, depending on linker options.
    uint8_t reserved;
} hal_core_wakeup_source_gpio_t;

File: hal/inc/core_hal.h [hunk 8]
```diff
@@ -284,6 +285,100 @@ typedef void(*HAL_Event_Callback)(int event, int flags, void* data);
...
+
+/**
+ * HAL sleep wakeup source configurations
+ */
+typedef struct hal_core_wakeup_source_t {
+    uint16_t size;
+    uint16_t version;
+    hal_core_wakeup_source_type_t type;
+    uint16_t reserved;
+    union {
```
avtolstoy: Let's better use C-style inheritance. See comment above. The size of this `source` union will be a maximum size of any wakeup source, and I don't think this is something we want to do.

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+    SystemSleepConfiguration& duration(std::chrono::milliseconds ms) {
+        return duration(ms.count());
+    }
+
+    SystemSleepConfiguration& duration(std::chrono::seconds s) {
+        std::chrono::milliseconds ms = std::chrono::duration_cast<std::chrono::milliseconds>(s);
+        return duration(ms.count());
+    }
+
+    SystemSleepConfiguration& duration(std::chrono::minutes min) {
```
avtolstoy: There is no need to do this. It's fine to just have one method that takes minimum resolution (in our case milliseconds) and it'll all work out of the box: https://cppinsights.io/s/7dd33ff3

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+    }
+    ~SystemSleepConfiguration() {
+        // Free memory
+        auto wakeupSource = config_.wakeup_sources;
+        while (!wakeupSource) {
+            auto next = wakeupSource->next;
+            delete wakeupSource;
+            wakeupSource = next;
+        }
+    }
```
avtolstoy: Copy/move constructors/operators. We'll have a double free:
```c++
{
    {
        auto c = SystemSleepConfiguration().gpio(D0, HIGH);
        auto copy = c;
    }
    // Double free   
}
```

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+    // Getters
+    hal_core_sleep_config_t* halCoreSleepConfig() const {
+        return &config_;
+    }
+
+    SystemSleepMode sleepMode() const {
+        return static_cast<SystemSleepMode>(config_.mode);
+    }
+
+    bool wakeupSourceFeatured(hal_core_wakeup_source_type_t type) const {
```
avtolstoy: I would also suggest adding `start` argument, so that we can continue the search from a certain position. E.g. if we want to check for duplicate gpio entries and change the return type to a pointer to an actual wakeup source instead of just `bool`, so we can inspect its contents.

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+    NONE        = SYSTEM_SLEEP_MODE_NONE,
+    STOP        = SYSTEM_SLEEP_MODE_STOP,
+    HIBERNATE   = SYSTEM_SLEEP_MODE_HIBERNATE,
+    SHUTDOWN    = SYSTEM_SLEEP_MODE_SHUTDOWN
+};
+
+class SystemSleepConfiguration {
+public:
+    SystemSleepConfiguration() {
+        config_.size = sizeof(hal_core_sleep_config_t);
```
avtolstoy: It's probably better to zero-out `config_` at first.

wiring/inc/spark_wiring_system.h
```diff
@@ -105,6 +105,164 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+        config_.mode = static_cast<hal_core_sleep_mode_t>(mode);
+        return *this;
+    }
+
+    SystemSleepConfiguration& gpio(pin_t pin, InterruptMode mode) {
+        clearGpioWakeupSources();
+        return addGpioWakeupSource(pin, mode);
+    }
+
+    SystemSleepConfiguration& gpios(std::initializer_list<hal_core_wakeup_source_gpio_t> gpios) {
```
avtolstoy: I am not sure if this is the right way. I think `.gpio(XX, XX).gpio(YY, YY).gpio(ZZ, ZZ)` makes more sense.

wiring/src/spark_wiring_system.cpp
avtolstoy: Let's have a discussion about this tomorrow. I think we might have to still have a proxy system level sleep method (still delegating to HAL) that might do a couple of things:
1. Deal with system level stuff
2. Deal with system level sleep modes (if we are to introduce any)
3. Potentially generate system events after wakeup
etc

avtolstoy:
@XuGuohui I'd also propose we create main `feature/sleep20` branch or something along those lines, to where we'll initially be merging these smaller bite-sized PRs and then eventually merge it into `develop`.

hal/inc/core_hal.h
avtolstoy: I would vote against _disabling_ peripherals in this mode. Keep them running, mask their interrupts etc just like we do right now.

If we really want to reduce current while in STOP mode and do want to power down certain peripherals, I think we need to introduce a new sleep mode.

hal/inc/core_hal.h
XuGuohui: It makes sense I think. I was going to introduce a class to cache and restore peripherals' configurations, that's not allowed in c file though.

hal/inc/core_hal.h
XuGuohui: Sounds reasonable. 👍

hal/inc/core_hal.h
avtolstoy: > to cache and restore peripherals' configurations

See https://github.com/particle-iot/device-os/pull/1965#discussion_r345664545 about that. I'm not sure that should be part of our initial Sleep 2.0 offering.

hal/inc/core_hal.h
XuGuohui: Could you help me understand why we need this?

hal/inc/core_hal.h
XuGuohui: Yep, I agree. It saves heap.

hal/inc/core_hal.h
avtolstoy: It's C-style inheritance. Each wakeup source will be derived from a base struct `hal_core_wakeup_source_t` and will have all its fields. We can now easily cast all of them to `hal_core_wakeup_source_t`, inspect each of them as `hal_core_wakeup_source_t`, and easily cast back to a specific wakeup type struct.
e.g.
```c++
void someCallingCode() {
    hal_core_sleep_config_t conf = {};
    // Fills hal_core_sleep_config_t
    hal_core_wakeup_source_gpio_t gpio = {};
    hal_core_wakeup_source_rtc_t rtc = {};
    conf.wakeup_sources = &gpio;
    gpio.next = &rtc;
    hal_core_sleep(&conf, nullptr);
}
int hal_core_sleep(const hal_core_sleep_config_t* config, void* reserved) {
    for (auto src = config->wakeup_sources; src != nullptr; src = src->next) {
        switch (src->type) {
            case HAL_CORE_WAKEUP_SOURCE_TYPE_GPIO: {
                auto gpio = static_cast<hal_core_wakeup_source_gpio_t*>(src);
                // Inspect hal_core_wakeup_source_gpio_t-specific fields
                (void)gpio->pin;
                break;
            }
            // ...
        }
    }
}
```

wiring/inc/spark_wiring_system.h
XuGuohui: Good catch!

wiring/inc/spark_wiring_system.h
```diff
@@ -52,6 +52,8 @@
 #include "hw_ticks.h"
 #endif
 
+using spark::Vector;
```
XuGuohui: This will be deleted.

system/inc/system_sleep.h
```diff
@@ -41,12 +45,147 @@ typedef enum System_Sleep_Flag
     SYSTEM_SLEEP_FLAG_NO_WAIT = 0x04
 } System_Sleep_Flag;
 
+
+enum class SystemSleepMode: uint16_t {
```
XuGuohui: This should inherit from `uint8_t`

wiring/src/spark_wiring_system.cpp
```diff
@@ -1,5 +1,6 @@
 
 #include "core_hal.h"
+#include "sleep_hal.h"
```
XuGuohui: System wiring layer should not access the sleep HAL APIs directly.

system/inc/system_sleep.h
```diff
@@ -41,12 +45,147 @@ typedef enum System_Sleep_Flag
...
+};
+
+
 /**
  * @param param A SystemSleepNetwork enum cast as an integer.
  */
 int system_sleep(Spark_Sleep_TypeDef mode, long seconds, uint32_t param, void* reserved);
 int system_sleep_pin(uint16_t pin, uint16_t mode, long seconds, uint32_t param, void* reserved);
 int system_sleep_pins(const uint16_t* pins, size_t pins_count, const InterruptMode* modes, size_t modes_count, long seconds, uint32_t param, void* reserved);
+int system_sleep_ext(const SystemSleepConfiguration& config, void* reserved);
```
avtolstoy: `const SystemSleepConfiguration& config` -> `const SystemSleepConfiguration* config`

XuGuohui: To avoid invoking the copy constructor?

avtolstoy: Nope, you simply shouldn't use references in dynalib APIs.

XuGuohui: Ah 👍  ! I was thinking the system layer is compiled along with the wiring layer.

XuGuohui:
@avtolstoy I'd like you to go though the sleep modes defined in `sleep_hal.h` first before I going further to implement the `hal_sleep()` API.

PR #1970, opened 2019-11-21
[Gen 2] Fixes thread-safety issues in libc functions

YutingYou:
### Problem

Float formatting (log, sprintf, etc) can cause Hardfault on Gen2 platform.

The problem has been identified to be disabled FreeRTOS feature `configUSE_NEWLIB_REENTRANT`, which makes FreeRTOS allocate and manage thread-local libc reentrant structure.

### Solution

Enable that option for Gen 2 platforms (Electron, Photon).

### Steps to Test

Run the example app, it shouldn't crash.

### Example App

```c
#include "Particle.h"
SerialLogHandler *logHandler = new  SerialLogHandler(115200, LOG_LEVEL_ALL);


SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);
void setup()
{
    Particle.connect();
}
void loop()
{
  static uint32_t loop_sec = System.uptime();
  char buf[100];
  for(unsigned int i=0; i < 100; i++)
  {
    snprintf(buf, sizeof(buf), "%f", (double) 1234567890.12345667890);
  }
  if(loop_sec != System.uptime())
  {
    Log.info("loop heartbeat: %s", buf);
  }
}
```

### References

- [CH41912]
- 

---

### Completeness

- [ ] Added documentation


MrZANE42:
Do you know which previous releases also have this bug?
What was the first release affected by this, or has this been there since the beginning?

avtolstoy:
@MrZANE42 The issue is present in all DeviceOS releases for Gen 2 platforms < 1.4.3. Gen 3 devices are not affected.

Issue #1968, opened 2019-11-15
Red SOS on P1 when it doesn't have valid wifi creds


markterrill:
<details>
  <summary><i>P1 Red SOS flash if it doesn't have valid wifi creds. May be hardware related, may be a very weird sequence of provisioning but it seems the wifi creds are corrupted and cause a SOS</i></summary>

</details>

### Expected Behavior

What should it do? Keep connecting / go to wifi listen mode as appropriate. Not a red SOS

### Observed Behavior

What did it actually do?  Red SOS after about 10 seconds.

### Steps to Reproduce

How did you make it do that, and which system/user firmware versions were tested?
1.4.2 / P1 chip

- China factory: Board was updated from 1.2.x (from a tape and reel) to 1.4.2. Wifi creds were added before update due to issues we were seeing with the chinese internet connection, seemed to be faster later on. 
- Board was given our firmware, claimed via particle device add, renamed, etc etc. Put on a plane to us in Australia
- Turn on device, red flash after ~15 seconds. Presumably because it didn't have a known wifi network to attach to. 
- Problems persisted even after flashing the below stripped down code

The fix?
- Hardware button wifi reset. See below additional commentary.

### Test App

```c
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);

char strVersion[32];

void setup() {

    delay (4000);
    // Start serial at 9600 baud
    Serial.begin(9600);

    Serial.println("start of setup 0 functions");

    // Spark read/write variables
    //Particle.variable("version", strVersion, STRING);
    //Particle.function("reset", callReset);

    Serial.println("start of setup 1 wifi on");

    WiFi.on();
    Serial.println("start of setup 1 wifi connect");

    WiFi.connect();

    Serial.println("start of setup 1 particle connect");
    Particle.connect();

}


int callReset(String nonsenseArg){

    Serial.println("Going to reset device now due to cloud function call!");
    Particle.publish("info", "Going to reset device now due to cloud function call!", 60, PRIVATE);

    delay(2000); // 2 seconds

    System.reset();

    return 1;

}

void loop() {

}

```

### References

None. There isn't published best practice code for semi automatic & system thread and how you reliably get it online.

---

markterrill:
[...]

PR #1976, opened 2019-11-28
[Photon/P1] Fixes handling of invalid WiFi access point config entries in DCT

avtolstoy:
### Problem

1. See #1968, WiFi AP list gets initialized to 0xff instead of expected 0x00, which causes invalid (filled with 0xff) entries to be treated as valid credentials resulting in crashes
2. There aren't any sanity checks on the WiFi AP entries retrieved from DCT

### Solution

1. Fill the stored AP list with 0x00 in factory DCT image file: `dct_prep.bin`. Previously this list should have been initialized with `NVMEM_SPARK_Reset_SysFlag` system flag set, however (unconfirmed, most likely) due to restructuring of the factory image in #1887 and `NVMEM_SPARK_Reset_SysFlag` being used to perform a factory reset in the bootloader (after which it's reset), the system no longer runs `wlan_reset_credentials_store()`.
2. Add sanity checks to `wlan_hal`.

#### Sanity checks
1. Flash https://github.com/particle-iot/device-os/files/3877387/dct_dump_full.bin.zip @ 0x08004000 using the debugger
2. Flash stock DeviceOS
3. The device should crash after failing to connect to the WiFi AP
4. Flash DeviceOS out of this branch
5. The device should not crash

#### Factory image

Follow the steps in #1887, get the device to boot into DeviceOS, put into DFU mode, get a full two-page DCT dump (`dfu-util -d 2b04:c006 -s 0x08004000:0x8000 -a 0 -U dct_dump_full.bin`), make sure that `0x4a30` - `0x4a90` and `0x8a30` - `0x8a90` are zero-initialized.

### Example App

N/A

### References

- #1968
- [CH42251]

---

### Completeness

- [ ] Added documentation


sergeuz:
There's a check in wlan_hal which seems to be there to prevent this kind of issues, however, it doesn't work correctly after [this change](https://github.com/particle-iot/device-os/commit/6882e8b36652a61d251ee559f97af8d7a04bccf2#diff-4a5db8e337fd9b5fdba27421bc9d6742R69) (`device_configured` evaluates to `true` if DCT is not initialized). Should we fix it in this PR as well?

avtolstoy:
Nice catch, @sergeuz. Yeah, we definitely need to fix it.

avtolstoy:
Fixed.

PR #1978, opened 2019-12-02
Implements bitwise operants for enum class

XuGuohui:
<details>

### Problem

Enum class doesn't provide bitwise operants for type-safety purpose. While for bit mask flags, each bit of it may be a member of an enum class, in which case we need to enable bitwise operants to operate the flag basing on the given enum class.

### Solution

Introduce a class template that uses enum class as the only type argument and has a member, which type is the underlying type of the enum class. The corresponding object can be constructed using one of the enum value and it also provides bitwise operants for operating the object using the enum values.

### Steps to Test

N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


services/inc/enumclass.h
```diff
@@ -22,12 +22,225 @@
...
+};
+
+#define ENABLE_ENUM_CLASS_BITWISE(T) \
+template<>                           \
+struct EnableBitwise<T> {            \
+    static const bool enable = true; \
+}
+
+template<typename T>
+class BitMaskFlags<T, typename std::enable_if_t<std::is_enum<T>::value>> {
```
avtolstoy: Let's move this class into its own header named after the class, insead of `enumclass.h`. Let's also perhaps rename to `EnumFlags`, as that tells the intended usage better.

services/inc/enumclass.h
```diff
@@ -22,12 +22,225 @@
...
+
+    BitMaskFlags<T> operator<<(size_t n) const;
+    BitMaskFlags<T>& operator<<=(size_t n);
+    BitMaskFlags<T> operator>>(size_t n) const;
+    BitMaskFlags<T>& operator>>=(size_t n);
+
+    typename std::underlying_type<T>::type value() const;
+
+    operator bool() const;
+    operator ValueType() const;
```
avtolstoy: We should remove both these operators, as otherwise we lose the type-safety by enabling implicit casting to bool and `ValueType`. Separate methods like `bool isSet()` and `value()` are a better choice even if require a bit more typing when using this class.

services/inc/enumclass.h
```diff
@@ -22,12 +22,225 @@
...
+    BitMaskFlags<T>& operator^=(const BitMaskFlags<T>& flags);
+
+    BitMaskFlags<T> operator~() const;
+
+    BitMaskFlags<T> operator<<(size_t n) const;
+    BitMaskFlags<T>& operator<<=(size_t n);
+    BitMaskFlags<T> operator>>(size_t n) const;
+    BitMaskFlags<T>& operator>>=(size_t n);
+
+    typename std::underlying_type<T>::type value() const;
```
avtolstoy: `ValueType`? We already have the typedef.

services/inc/enumclass.h
```diff
@@ -22,12 +22,225 @@
...
+
+    bool operator==(const BitMaskFlags<T>& flags) const;
+    bool operator!=(const BitMaskFlags<T>& flags) const;
+    bool operator>(const BitMaskFlags<T>& flags) const;
+    bool operator>=(const BitMaskFlags<T>& flags) const;
+    bool operator<(const BitMaskFlags<T>& flags) const;
+    bool operator<=(const BitMaskFlags<T>& flags) const;
+
+private:
+    typename std::underlying_type<T>::type value_;
```
avtolstoy: `ValueType`?

services/inc/enumclass.h
```diff
@@ -22,12 +22,225 @@
...
+#define ENABLE_ENUM_CLASS_BITWISE(T) \
+template<>                           \
+struct EnableBitwise<T> {            \
+    static const bool enable = true; \
+}
+
+template<typename T>
+class BitMaskFlags<T, typename std::enable_if_t<std::is_enum<T>::value>> {
+public:
+    typedef typename std::underlying_type<T>::type ValueType;
```
avtolstoy: There is also `std::underlying_type_t<T>` since C++14 :) A bit less of typing.

services/inc/enumclass.h
XuGuohui: I agree that we remove the `ValueType` operator. While the `bool` operator is very useful when we are doing something like:
```
if (enumFlags & ENUM_VALUE) {
}
```

avtolstoy: That's won't call the `bool` operator though. It uses `ValueType` operator.

XuGuohui: What if I remove the `ValueType` operator? Can we still keep the `bool` operator?

avtolstoy: I think we will still lose type-safety, because int <-> bool implicit conversions will still happen.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1143,22 +1143,22 @@ ssize_t BleCharacteristic::setValue(const uint8_t* buf, size_t len, BleTxRxType
     if (impl()->local()) {
         int ret = SYSTEM_ERROR_NOT_SUPPORTED;
         // Updates the local characteristic value for peer to read.
-        if (impl()->properties() & BleCharacteristicProperty::READ) {
+        if (impl()->properties().isSet(BleCharacteristicProperty::READ)) {
```
avtolstoy: What's wrong with `operator&`? :)

XuGuohui: It returns `EnumFlags`, while we've removed the `bool()` operator, so it will result compilation error.

PR #1979, opened 2019-12-02
Implement Particle.cancel()

sergeuz:
### Problem

This PR implements a method called `Particle.cancel()` which closes the cloud connection synchronously, without waiting for any unacknowledged messages.

### Steps to Test

_TBD_

### References

- [ch42232]

### DIFF
--- a/system/inc/system_cloud.h
+++ b/system/inc/system_cloud.h
@@ -240,6 +240,28 @@ void spark_cloud_flag_disconnect(void);    // should be set connected since it m
  */
 bool spark_cloud_flag_auto_connect(void);
 
+/**
+ * Option flags changing the behavior of `spark_cloud_disconnect()`.
+ */
+typedef enum cloud_disconnect_flag {
+    CLOUD_DISCONNECT_GRACEFULLY = 0x01, ///< Disconnect gracefully.
+    CLOUD_DISCONNECT_DONT_CLOSE = 0x02, ///< Do not close the socket.
+    CLOUD_DISCONNECT_SYNC = 0x04 ///< Disconnect synchronously.
+} cloud_disconnect_flag;
+
+/**
+ * Parameters for `spark_cloud_disconnect()`.
+ */
+typedef struct spark_cloud_disconnect_param {
+    uint16_t size; ///< Size of this structure.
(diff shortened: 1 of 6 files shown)


wiring/src/spark_wiring_cloud.cpp
```diff
@@ -76,3 +76,9 @@ Future<bool> CloudClass::publish_event(const char *eventName, const char *eventD
 int CloudClass::publishVitals(system_tick_t period_s_) {
     return spark_publish_vitals(period_s_, nullptr);
 }
+
+void CloudClass::cancel() {
```
m-mcgowan: I would hold off adding a new public API for a bugfix release and instead give the customer this code block as an example of how to synchronously cancel the cloud connection. Let's also document the expected behavior, specifically that the ongoing connection isn't actively interrupted, but will continue to the next logical step. E.g. if the handshake is ongoing then the handshake will continue to completion or failure, but will not be prematurely cancelled.

system/inc/system_cloud.h
```diff
@@ -240,6 +240,23 @@ void spark_cloud_flag_disconnect(void);    // should be set connected since it m
  */
 bool spark_cloud_flag_auto_connect(void);
 
+/**
+ * Option flags changing the behavior of `spark_cloud_disconnect()`.
+ */
+typedef enum cloud_disconnect_flag {
+    CLOUD_DISCONNECT_GRACEFULLY = 0x01, ///< Disconnect gracefully.
+    CLOUD_DISCONNECT_DONT_CLOSE = 0x02 ///< Do not close the socket.
```
m-mcgowan: What does it mean to disconnect without closing the socket?

wiring/src/spark_wiring_cloud.cpp
sergeuz: I renamed this method to `Particle._cancel()` as we discussed.

system/inc/system_cloud.h
sergeuz: It doesn't make much sense from the user's perspective, but we use it internally as an optimization. These flags are also defined in [#1899](https://github.com/particle-iot/device-os/pull/1899/files#diff-b43faa0ec0ffb23fa014851243cadf77R87), so having the same set of flags in this PR will simplify merging.

PR #1984, opened 2019-12-06
Update manifest.json to support GCC8

zfields:
This will enable you to use/test the workbench for local development.

### DIFF
--- a/.workbench/manifest.json
+++ b/.workbench/manifest.json
@@ -3,7 +3,7 @@
     "toolchains": [
         {
             "firmware": "deviceOS@source",
-            "compilers": "gcc-arm@5.3.1",
+            "compilers": "gcc-arm@8.3.1",
             "debuggers": "openocd@0.11.2-adhoc6ea4372.0",
             "platforms": [6, 8, 10, 12, 13, 14, 22, 23, 24],
             "scripts": "buildscripts@1.8.0",
@@ -15,10 +15,10 @@
             "x64": [
                 {
                     "name": "gcc-arm",
-                    "version": "5.3.1",
+                    "version": "8.3.1",
                     "main": "./bin",
-                    "url": "https://binaries.particle.io/gcc-arm/windows/x64/gcc-arm-v5.3.1.tar.gz",
-                    "sha256": "9d5106811e6ad8f3fd63ac266fb3e6e31798e14333b50f849130f604e331ba3d"
+                    "url": "https://binaries.particle.io/gcc-arm/windows/x64/gcc-arm-v8.3.1.tar.gz",
+                    "sha256": "f677b13807d76271f57c160c095225710c3d14f70af70736a5bcf0537dd35298"
                 }
             ],
             "x86": []
@@ -27,10 +27,10 @@
             "x64": [
                 {
                     "name": "gcc-arm",
-                    "version": "5.3.1",
+                    "version": "8.3.1",
                     "main": "./bin",
-                    "url": "https://binaries.particle.io/gcc-arm/darwin/x64/gcc-arm-v5.3.1.tar.gz",
-                    "sha256": "f41f32436188c3017422635b187378c8bf0938a3ad396506b2d3b8ce1fb402c9"
+                    "url": "https://binaries.particle.io/gcc-arm/darwin/x64/gcc-arm-v8.3.1.tar.gz",
+                    "sha256": "20297b21b9a6af340799458e0adbe3798ce89ef5c82c7cef91bd44aea2aa7914"
                 }
             ],
             "x86": []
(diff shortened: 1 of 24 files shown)


YutingYou:
@zfields @busticated  After changing `manifest.json` configuration, will the default GCC toolchain be changed to GCC8?  For now, the compile-time of GCC8 is twice as long as GCC5, so it might be not friendly to our user.

YutingYou:
For the compile-time, it should be fine for modular firmware, it takes a long time to build monolithic firmware. After discussing with Andrey(@avtolstoy ), we have more concerns about directly changing the GCC toolchain to GCC8. We hope to keep the reliable GCC5 toolchain and give our users an option to use GCC8, e.g. we'll release `device-os` (default: GCC5) and `device-os-gcc8`.

avtolstoy:
We can probably get away with having two release branches/tags, one with this change, one without. If there are other ways we can do this, let's discuss them as well.

busticated:
> For now, the compile-time of GCC8 is twice as long as GCC5, so it might be not friendly to our user

the aim here is to ensure the bundle we ultimately provide to users is well-tested and works as expected. we shouldn't release it to users until we're confident it works well. my understanding is that you (@YutingYou) are doing this work now in the branch this PR targets and once things are knocked into shape, it'll be merged into `develop` and / or released in an official version of the Device OS. only at that point would GCC8 become the default and only if you all say it's good to go.

> We hope to keep the reliable GCC5 toolchain and give our users an option to use GCC8

that's not a possibility today or for the foreseeable future. if / when we're happy with our GCC8 support, we can adopt it in the toolchain until then, it can live in a WIP branch or similar.

what we _absolutely_ want to avoid is having the Device OS docs recommend using one compiler while Workbench, etc is using another.

avtolstoy:
@busticated 

> that's not a possibility today or for the foreseeable future. if / when we're happy with our GCC8 support, we can adopt it in the toolchain until then, it can live in a WIP branch or similar.

> what we absolutely want to avoid is having the Device OS docs recommend using one compiler while Workbench, etc is using another.

This is not the intent. We are exploring an option of running two separate release lines e.g. `1.5.0` and `1.5.0+gcc8` in parallel, with the main one built still with gcc 5.3.1 for some time. I would not expect us having problems with having two different releases using two different toolchains, is that right?

busticated:
> We are exploring an option of running two separate release lines e.g. 1.5.0 and 1.5.0+gcc8 in parallel, with the main one built still with gcc 5.3.1 for some time.

let's make sure the product team is up-to-speed and weighs in.

>  I would not expect us having problems with having two different releases using two different toolchains

correct. any given Device OS release can use any given toolchain. the same Device OS release cannot use two different toolchains or otherwise support switching toolchain dependencies.

avtolstoy:
> let's make sure the product team is up-to-speed and weighs in.

This has already been discussed a couple of times. We don't want for now to suddenly make a switchover, despite our testing showing that we've resolve all the issues, while at the same time want to make gcc8 available as an opt-in feature. This seems like the best way of doing so.

> correct. any given Device OS release can use any given toolchain. the same Device OS release cannot use two different toolchains or otherwise support switching toolchain dependencies.

I think we are aligned then. These are going to be separate releases, out of separate branches/tags.

busticated:
> We ...want to make gcc8 available as an opt-in feature. This seems like the best way of doing so.

one option to consider: prior to issuing official releases, use Workbench to facilitate that testing and feedback cycle. you can point people to a long-running branch - `develop/gcc8-support` (or similar) - and then use the `.workbench/manifest.json` override file to facilitate testing via the `deviceOS@source` option. you could even tweak the build system output to show a message guiding users on how to provide feedback.

avtolstoy:
> one option to consider: prior to issuing official releases, use Workbench to facilitate that testing

I think we are slightly already past the experimental stage and could potentially make the switchover, we are just being cautious. If we use just Workbench, this removes the whole layer of other things outside in that case: 1) buildpacks/cloud compilation; 2) OTA/SMH updates (although it's still not clear how to manage SMH updates in this case, we'll need to do something with module versions between these two lines); 3) system/application interoperability when built with different compiler versions etc

zfields:
So are we good to merge this in to Eugene's GCC8 branch since it will enable local compilation via the Workbench?

avtolstoy:
> So are we good to merge this in to Eugene's GCC8 branch since it will enable local compilation via the Workbench?

I'd say no. We are going to merge GCC8 support branch in `develop` as soon as we can, because it does not break anything and merely fixes a few things to allow the DeviceOS to be buildable with GCC8, and will be pulling the manifest changes into appropriate release branches. If one wants to try using workbench with gcc with develop, we'd for now instead need to keep rebasing this branch.

busticated:
@avtolstoy 

if the aim in merging the GCC8 support branch into `develop` is to facilitate testing, for obvious reasons that testing should be done using our official toolchain.

what's the downside in merging this PR and having the `deviceOS@source` toolchain use GCC8 while official releases continue to use GCC5? how else do you propose we vet the new GCC8 bundle and associated toolchain?

avtolstoy:
> what's the downside in merging this PR and having the deviceOS@source toolchain use GCC8 while official releases continue to use GCC5?

Because the official releases are going to be branched off of `develop` and given the fact that we are transitioning to using the WB dependency installer for buildpacks we'd automatically be switching over to gcc8 in this case and would have to 'fix' things for every release to downgrade to gcc5. IMO we'd rather do the opposite.

zfields:
@YutingYou @avtolstoy @busticated 

If I'm understanding correctly, we are not ready to advance the compiler for DeviceOS to GCC 8.3.1.

The aim of this commit is preliminary exploration to refactor the code base in such a way as to enable the compilation using GCC 8.3.1, thus making it _POSSIBLE_ to switch compilers at a later date. At that time, there will be a great deal more testing, and only then would we want to apply this commit to develop.

**I will close this PR, but not delete the branch, because this commit can be cherry-picked to enable local development via the workbench.**

It would be of great benefit to the company if future _(similar)_ work to enable GCC 8.3.1 would
1. Apply this commit
1. Do local development via the Workbench
1. Revert this commit
1. Merge changes

PR #1986, opened 2019-12-07
Feature/sleep20

XuGuohui:
### Problem

Currently, both HAL and Wiring sleep APIs are not very extensible and support only a subset of possible modes and wake-up options. It has been a long time goal for us to improve the sleep APIs and we’ve already taken some steps in that direction e.g.:

- Wake up reason
- Multiple wake-up pins support for Standby mode on Gen 3
- Type-safe sleep flags in wiring
- Woken up by BLE on Gen 3

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup() {
    config.mode(SystemSleepMode::STOP)
          .gpio(D0, RISING)
          .duration(30min)
          .ble();
    System.sleep(config);
}

void loop() {
}
```

### References

- https://github.com/particle-iot/device-os/pull/1965
- https://github.com/particle-iot/device-os/pull/1972
- https://github.com/particle-iot/device-os/pull/1975
- https://github.com/particle-iot/device-os/pull/1977
- https://github.com/particle-iot/device-os/pull/1982

---

### Completeness

- [ ] Added documentation


system/src/system_sleep.cpp
```diff
@@ -251,3 +252,147 @@ int system_sleep_pins(const uint16_t* pins, size_t pins_count, const InterruptMo
...
+            Spark_Sleep();
+        }
+        cloudResume = spark_cloud_flag_auto_connect();
+        // Clear the auto connect status
+        spark_cloud_flag_disconnect();
+    }
+#endif
+
+    // Network disconnect
+#if HAL_PLATFORM_CELLULAR
```
avtolstoy: `if_get_list()` could potentially be used on Gen3 to traverse the list of interfaces, but let's leave for now as-is, as it allows to share to code between Gen 2 and Gen 3. Maybe add a FIXME.

wiring/inc/spark_wiring_network.h
```diff
@@ -60,7 +60,7 @@ class NetworkClass
     virtual uint16_t getListenTimeout();
     virtual bool listening();
 
-    operator network_interface_t() {
+    operator network_interface_t() const {
```
avtolstoy: :+1:

wiring/inc/spark_wiring_system.h
```diff
@@ -435,6 +437,7 @@ class SystemClass {
 
 private:
     SleepResult sleepResult_;
+    SystemSleepResult systemSleepResult_;
```
avtolstoy: Ideally we should also fill this structure when waking up from deep sleep (same as with `SleepResult`). Does it make sense to simply use `SystemSleepResult` everywhere instead and provide `operator SleepResult()` not to lose compatibility?

wiring/src/spark_wiring_system.cpp
```diff
@@ -41,16 +47,101 @@ void SystemClass::reset(uint32_t data)
...
+    System.toSleepResult();
+    return System.systemSleepResult_;
+}
+
 SleepResult SystemClass::sleep(Spark_Sleep_TypeDef sleepMode, long seconds, SleepOptionFlags flags)
 {
-    int ret = system_sleep(sleepMode, seconds, flags.value(), NULL);
-    System.sleepResult_ = SleepResult(WAKEUP_REASON_NONE, static_cast<system_error_t>(ret));
-    return System.sleepResult_;
+    if (sleepMode == SLEEP_MODE_WLAN) {
```
avtolstoy: Let's keep the compatibility layer in the system instead. We anyway need to keep both system and HAL sleep 1.0 functions around for ABI compatibility, so it does not make sense to have this in wiring.

avtolstoy:
@XuGuohui Main things:
1. https://github.com/particle-iot/device-os/pull/1986#discussion_r355388666
2. https://github.com/particle-iot/device-os/pull/1986#discussion_r355298849

hal/inc/hal_dynalib_core.h
```diff
@@ -79,6 +80,9 @@ DYNALIB_FN(32, hal_core, HAL_Set_Event_Callback, void(HAL_Event_Callback, void*)
 DYNALIB_FN(33, hal_core, HAL_Core_Enter_Stop_Mode_Ext, int(const uint16_t*, size_t, const InterruptMode*, size_t, long, void*))
 DYNALIB_FN(34, hal_core, HAL_Core_Execute_Standby_Mode_Ext, int(uint32_t, void*))
 
+DYNALIB_FN(35, hal_core, hal_sleep_validate_config, int(const hal_sleep_config_t*, void*))
```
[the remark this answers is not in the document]
XuGuohui: In that case, we need to allocate a `hal_sleep_config_t` structure from stack or heap to allow changing the dry-run flag, since the sleep configuration can be const from user aspect.

hal/inc/sleep_hal.h
```diff
@@ -0,0 +1,162 @@
...
+    HAL_WAKEUP_SOURCE_TYPE_TIMER = 9,
+    HAL_WAKEUP_SOURCE_TYPE_CAN = 10,
+    HAL_WAKEUP_SOURCE_TYPE_USB = 11,
+    HAL_WAKEUP_SOURCE_TYPE_BLE = 12,
+    HAL_WAKEUP_SOURCE_TYPE_NFC = 13,
+    HAL_WAKEUP_SOURCE_TYPE_NETWORK = 14,
+    HAL_WAKEUP_SOURCE_TYPE_MAX = 0x7FFF
+} hal_wakeup_source_type_t;
+
+typedef enum hal_sleep_wait_t {
```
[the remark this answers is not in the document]
XuGuohui: It can be in the HAL in case that we're going to wait some data transaction to be finished in the HAL layer before entering sleep mode in the future.

hal/inc/sleep_hal.h
```diff
@@ -0,0 +1,162 @@
...
+    HAL_SLEEP_WAIT_NO_WAIT = 0,
+    HAL_SLEEP_WAIT_CLOUD = 1,
+    HAL_SLEEP_WAIT_MAX = 0x7F,
+} hal_sleep_wait_t;
+
+#if PLATFORM_ID > 3
```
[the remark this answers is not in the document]
XuGuohui: Fixed.

hal/src/core/sleep_hal.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "check.h"
+#include "core_hal.h"
+#include "sleep_hal.h"
+
+int hal_sleep_validate_config(const hal_sleep_config_t* config, void* reserved) {
+    return SYSTEM_ERROR_NONE;
+}
+
+int hal_sleep_enter(const hal_sleep_config_t* config, hal_wakeup_source_base_t** wakeup_source, void* reserved) {
```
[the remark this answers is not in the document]
XuGuohui: Done.

hal/src/nRF52840/core_hal.c
```diff
@@ -106,6 +106,12 @@ extern void* malloc_heap_end();
 void* module_user_pre_init();
 #endif
 
+// C file cannot include header file including C++ grammar.
+extern int hal_ble_stack_init(void* reserved);
```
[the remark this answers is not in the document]
XuGuohui: Already fixed that header file instead.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+#include "check.h"
+#include "core_hal.h"
+#include "interrupts_hal.h"
+#include "sleep_hal.h"
+#include "spark_wiring_vector.h"
+
+using spark::Vector;
+
+static bool isWakeupSourceSupportedGpio(hal_sleep_mode_t mode, const hal_wakeup_source_gpio_t* gpio) {
```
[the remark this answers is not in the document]
XuGuohui: How about `validateGpioWakeupSource()` ?

system/inc/system_sleep.h
```diff
@@ -41,12 +51,406 @@ typedef enum System_Sleep_Flag
...
+};
+
+enum class SystemSleepWait: uint8_t {
+    NO_WAIT = HAL_SLEEP_WAIT_NO_WAIT,
+    CLOUD   = HAL_SLEEP_WAIT_CLOUD
+};
+
+enum class SystemSleepWakeupReason: uint16_t {
+    UNKNOWN = HAL_WAKEUP_SOURCE_TYPE_UNKNOWN,
+    BY_GPIO = HAL_WAKEUP_SOURCE_TYPE_GPIO,
```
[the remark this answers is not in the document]
XuGuohui: For Gen2 platforms, `ADC`, ` GPIO` etc. are also macros in chip level header file like stm32f2xx.h. That would result compilation errors. Any better naming scheme? Neither I like this style honestly.

system/inc/system_sleep.h
```diff
@@ -41,12 +51,406 @@ typedef enum System_Sleep_Flag
...
+            start = start->next;
+        }
+        return nullptr;
+    }
+
+private:
+    const hal_sleep_config_t* config_;
+};
+
+class SystemSleepConfiguration: protected SystemSleepConfigurationHelper {
```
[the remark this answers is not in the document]
XuGuohui: It can be used by system layer potentially though. For example, the system may enter sleep mode automatically if `SYSTEM_MODE(AUTOMATIC)` is specified when system is in idle state. Besides, it would mess up the spark_wiring_system.h, as I'd like all sleep relevant stuff are resides in a dedicated header file, i.e. system_sleep.h.

system/inc/system_sleep.h
```diff
@@ -41,12 +51,406 @@ typedef enum System_Sleep_Flag
...
+              valid_(config.valid_) {
+        memcpy(&config_, &config.config_, sizeof(hal_sleep_config_t));
+        config.config_.wakeup_sources = nullptr;
+    }
+
+    // Copy constructor
+    SystemSleepConfiguration(const SystemSleepConfiguration& config) = delete;
+
+    // Copy operator
+    SystemSleepConfiguration& operator=(const SystemSleepConfiguration& config) = delete;
```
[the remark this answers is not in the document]
XuGuohui: Added.

wiring/src/spark_wiring_system.cpp
XuGuohui: Done.

system/inc/system_sleep.h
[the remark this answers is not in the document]
avtolstoy: It may very well then use `wiring` `System.sleep()` or even directly call HAL. My main point is that this class is used only by wiring APIs, and hence should ideally be defined immediately where it's used.

system/inc/system_sleep.h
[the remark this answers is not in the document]
avtolstoy: Ugh. Does it help if we change the order of the includes?

hal/inc/sleep_hal.h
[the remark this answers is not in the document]
avtolstoy: What I'm trying to say is that HAL ideally has no idea about the cloud. It's fine to keep as-is.

hal/inc/hal_dynalib_core.h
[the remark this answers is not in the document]
avtolstoy: That's a good point :+1:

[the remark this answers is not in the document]
XuGuohui: How about adding one more argument in `bool` to indicate dry-run or not?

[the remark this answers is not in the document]
avtolstoy: In that case it'd have to be instead probably an argument with bitmask flags, which might be confusing (with flags in `hal_sleep_config_t` and in function call). Let's keep a separate function.

system/inc/system_sleep.h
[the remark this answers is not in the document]
XuGuohui: It won't help as it anyway includes stm32f2xx.h eventually previous to the code where `ADC` is presented in user application.

system/inc/system_sleep.h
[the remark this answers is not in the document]
XuGuohui: To maintain the compatibility, we've reworked the existing system sleep APIs, which now need to construct a `SystemSleepConfiguration` object and then invoke the new system sleep API. So this class is also used by the system layer.

wiring/inc/spark_wiring_system.h
XuGuohui: I'd like to introduce `SystemSleepResult` as the private member of `SleepResult` and keep the methods of it acting as wrappers of the `SystemSleepResult` methods. Sounds good?

avtolstoy: It still needs to be castable to `SleepResult` if we are to change the return type of Sleep 1.0 functions, because the user code may have:
```c++
SleepResult result = System.sleep(XXX);
```

hal/inc/hal_platform.h
```diff
@@ -271,4 +271,8 @@
 #define HAL_PLATFORM_SETUP_BUTTON_UX (0)
 #endif /* HAL_PLATFORM_SETUP_BUTTON_UX */
 
+#ifndef HAL_PLATFORM_SLEEP_2_0
+#define HAL_PLATFORM_SLEEP_2_0 (0)
```
avtolstoy: `HAL_PLATFORM_SLEEP20` for simplicity? Less awkward typing :slightly_smiling_face:

system/src/system_sleep.cpp
```diff
@@ -406,3 +383,119 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
...
+ */
+int system_sleep_pin(uint16_t wakeUpPin, uint16_t edgeTriggerMode, long seconds, uint32_t param, void* reserved)
+{
+    // Cancel current connection attempt to unblock the system thread
+    network_connect_cancel(0, 1, 0, 0);
+    InterruptMode m = (InterruptMode)edgeTriggerMode;
+    return system_sleep_pin_impl(&wakeUpPin, 1, &m, 1, seconds, param, reserved);
+}
+
+int system_sleep(Spark_Sleep_TypeDef sleepMode, long seconds, uint32_t param, void* reserved)
```
avtolstoy: Let's think of a test to validate that we haven't broken anything by changing the internal implementation.

hal/inc/sleep_hal.h
```diff
@@ -56,23 +61,24 @@ typedef enum hal_sleep_mode_t {
...
+    HAL_WAKEUP_SOURCE_TYPE_UART = 0x20,
+    HAL_WAKEUP_SOURCE_TYPE_I2C = 0x40,
+    HAL_WAKEUP_SOURCE_TYPE_SPI = 0x80,
+    HAL_WAKEUP_SOURCE_TYPE_TIMER = 0x100,
+    HAL_WAKEUP_SOURCE_TYPE_CAN = 0x200,
+    HAL_WAKEUP_SOURCE_TYPE_USB = 0x400,
+    HAL_WAKEUP_SOURCE_TYPE_BLE = 0x800,
+    HAL_WAKEUP_SOURCE_TYPE_NFC = 0x1000,
+    HAL_WAKEUP_SOURCE_TYPE_NETWORK = 0x2000,
+    HAL_WAKEUP_SOURCE_TYPE_MAX = 0x7FFFFFFF
```
avtolstoy: Any reason to make it 32-bit?

hal/inc/sleep_hal.h
```diff
@@ -83,18 +89,18 @@ typedef enum hal_sleep_wait_t {
...
 static_assert(sizeof(hal_sleep_mode_t) == 1, "length of hal_sleep_mode_t should be 1-bytes aligned.");
-static_assert(sizeof(hal_wakeup_source_type_t) == 2, "length of hal_wakeup_source_type_t should be 2-bytes aligned.");
+static_assert(sizeof(hal_wakeup_source_type_t) == 4, "length of hal_wakeup_source_type_t should be 4-bytes aligned.");
 static_assert(sizeof(hal_sleep_wait_t) == 1, "length of hal_sleep_wait_t should be 1-bytes aligned.");
 #endif
 
 /**
  * HAL sleep wakeup source base
  */
+typedef struct hal_wakeup_source_base_t hal_wakeup_source_base_t;
```
avtolstoy: Why double typedefs?

wiring/inc/spark_wiring_system.h
XuGuohui: Can't we just `return SleepResult(System.systemSleepResult_);`?

hal/inc/sleep_hal.h
XuGuohui: See: https://github.com/particle-iot/device-os/blob/feature/sleep20/hal/src/nRF52840/sleep_hal.cpp#L98 and https://github.com/particle-iot/device-os/blob/feature/sleep20/hal/src/nRF52840/sleep_hal.cpp#L280 to identify the enabled wakeup sources.

hal/inc/sleep_hal.h
XuGuohui: C compiler will throw an error that the `hal_wakeup_source_base_t*` in this struct is an incomplete type.

system/src/system_sleep.cpp
XuGuohui: Is the test to validate if the API is buildable with different arguments, i.e. API test, or to validate if this function behaves as expected?

system/inc/system_sleep.h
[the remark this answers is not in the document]
XuGuohui: So, I'll keep them as-is?

system/src/system_sleep.cpp
avtolstoy: I think we already have API tests, if not we need to add/adjust them. However, I'm talking more about an on-device test that can verify that the device behaves as expected and exactly the same as before this change.

system/inc/system_sleep.h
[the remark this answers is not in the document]
avtolstoy: It still somewhat annoys me that we can't name things how we want. Let's leave things as-is for now, but I'm anyway going to take a look today to see if we can fix it somehow :smile:

wiring/inc/spark_wiring_system.h
avtolstoy: Yes, as long as `System.systemSleepResult_` is convertible to `SleepResult`.

hal/inc/sleep_hal.h
avtolstoy: That can still be done using 16-bit enums, right..?

hal/inc/sleep_hal.h
avtolstoy: You can avoid this by using adding `struct` when defining `next` as well FYI.

hal/inc/sleep_hal.h
XuGuohui: Yes, but I'd like to reserve more bits for future use case.

avtolstoy: Oh, I missed that this is a bitmask now. I don't think that's the best solution. What exactly is the problem in handling normal non-bitmasked ids?

XuGuohui: Making them as bit mask values is to replace the `exit_conditions` that is used to identify the wakeup sources. I don't want to new a bit-mask enum type to introduce complexity. So I just make use of the wakeup source type and I think 32-bits is sufficient for iterating all possible wakeup source type.

XuGuohui: Ideally we should use the enum class bitwise, while that branch hasn't been merged. If making the wakeup source types as bit-masks makes sense to you, I'll update it in the enum class bitwise PR: https://github.com/particle-iot/device-os/pull/1978/files, the same as what I've done to `BleCharacteristicProperty`.

avtolstoy: Honestly, I still don't really understand why we need to use bitmasks here. The HAL API itself does not require in any way for the wakeup source types to be combined (ORed), so the only reason I see is we are doing this for convenience of the internal implementation. And even then, I'm not sure why we'd want that, as we always have access to the configuration object and can traverse the list of wakeup options anyway.

> So I just make use of the wakeup source type and I think 32-bits is sufficient for iterating all possible wakeup source type.

I wouldn't be so sure that 31 wakeup options will be enough to cover all platforms in some time.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    SleepResult result = System.sleep(nullptr, 0, nullptr, 0, 3s);
+
+    while (!Serial.isConnected());
+
+    assertEqual(result.error(), SYSTEM_ERROR_NONE);
+    assertEqual((int)result.reason(), (int)WAKEUP_REASON_RTC);
+}
+
+#endif // HAL_PLATFORM_SLEEP20
```
avtolstoy: Just a kind reminder that we should also add API tests to `wiring/api`.

hal/inc/sleep_hal.h
XuGuohui: > as we always have access to the configuration object and can traverse the list of wakeup options anyway

That's exactly what I did before using the bit-mask. The reason why I switched to use bit-mask is that it can reduce the time to re-enter sleep again if the wakeup source is not as expected after exiting sleep mode. Searching `wakeupSourceTypes` in sleep_hal.cpp you'll find that it used very often. If we don't go with that way, we can of course define a bunch of functions like `isWakeupSourceGpioEnabled()` to iterate the wakeup source chain to check if certain wakeup source is enabled.

avtolstoy: Reduce by how much? :smile: Traversing a singly-linked list is just `O(n)`. You don't really need to traverse it multiple times. My main argument here still is that we are influencing the API just for the sake of simplifying the internal implementation while also limiting ourselves to just 31 wakeup options. Wasn't one of our goals to have a flexible and easily portable HAL API?

XuGuohui: > You don't really need to traverse it multiple times
So simply define several bool variables instead to store the traversed result?

XuGuohui:
> Can we instead let `system_power_management_sleep()` handle this?

So I want to have more discussion about the power management on entering/exiting sleep modes. I'd like to call `FuelGauge::sleep()` in `PowerManager::sleep(true)`. But I was worry about that this would break the functionality when device in stop mode, leaving some resources working still. And should we call `system_power_management_sleep(false)` explicitly after exiting `hal_sleep_enter()`, regardless of the sleep mode it exited from?

hal/inc/sleep_hal.h
avtolstoy: Or better yet, can we simply configure things while we traverse the list? So:
1. Traverse the list and configure necessary peripherals for wakeup
2. Go into sleep
3. Wake up, figure out what woke us up
4. Traverse the list once again and deconfigure the peripherals that were used as wakeup sources

avtolstoy:
>  And should we call system_power_management_sleep(false) explicitly after exiting hal_sleep_enter(), regardless of the sleep mode it exited from?

:+1: 

> But I was worry about that this would break the functionality when device in stop mode, leaving some resources working still.

I think things should work fine even for stop mode. We don't really need to keep the fuelgauge running while we are in stop mode _for now_ until we add a new wake up source based on it e.g. `LOW_BATTERY` or something like that. Once we add that, we can modify `system_power_management_sleep()` to take a new sleep type argument to take specific actions depending on which sleep mode we are entering/exiting.

system/inc/system_sleep.h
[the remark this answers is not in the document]
XuGuohui: As a compromise, we can potentially use Camel-Case for enum values to avoid the conflict with macro definitions, as we might still suffer from this conflict in the future.

hal/inc/hal_platform.h
```diff
@@ -275,4 +279,8 @@
 #define HAL_PLATFORM_BACKUP_RAM (0)
 #endif /* HAL_PLATFORM_BACKUP_RAM */
 
+#ifndef HAL_PLATFORM_SLEEP20
+#define HAL_PLATFORM_SLEEP20 (0)
+#endif
+
```
XuGuohui: This can be removed as we no more have platform don't support this feature.

hal/inc/hal_platform_compat.h
```diff
@@ -54,8 +54,10 @@
 #endif
 
 #if PLATFORM_ID == 6 || PLATFORM_ID == 8 || PLATFORM_ID == 10
+#define HAL_PLATFORM_STM32F2XX 1
 #define HAL_PLATFORM_DCT 1
 #define HAL_PLATFORM_RNG 1
+#define HAL_PLATFORM_SLEEP20 1
```
XuGuohui: This can be removed as we no more have platform don't support this feature.

hal/src/nRF52840/hal_platform_nrf52840_config.h
```diff
@@ -92,3 +92,4 @@
 
 #define HAL_PLATFORM_BACKUP_RAM (1)
 
+#define HAL_PLATFORM_SLEEP20 (1)
```
XuGuohui: This can be removed as we no more have platform don't support this feature.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -0,0 +1,334 @@
...
+                }
+            }
+            HAL_Pin_Mode(gpioWakeup->pin, wakeUpPinMode);
+            HAL_InterruptExtraConfiguration irqConf = {0};
+            irqConf.version = HAL_INTERRUPT_EXTRA_CONFIGURATION_VERSION_2;
+            irqConf.IRQChannelPreemptionPriority = 0;
+            irqConf.IRQChannelSubPriority = 0;
+            irqConf.keepHandler = 1;
+            irqConf.keepPriority = 1;
+            HAL_Interrupts_Attach(gpioWakeup->pin, NULL, NULL, gpioWakeup->mode, &irqConf);
```
XuGuohui: nullptring all the NULLs in this file.

wiring/inc/spark_wiring_system.h
```diff
@@ -106,6 +108,167 @@ inline SleepResult::SleepResult(WakeupReason r, system_error_t e, pin_t p)
...
+};
+
+class SystemSleepResult {
+public:
+    SystemSleepResult()
+            : wakeupSource_(nullptr),
+              error_(SYSTEM_ERROR_NONE) {
+    }
+
+    SystemSleepResult(SleepResult r)
```
XuGuohui: If construct result by this way, calling methods of `SystemSleepResult` may return incorrect result information, because the `wakeupSource_` is nullptred.

PR #1992, opened 2019-12-27
[Boron/Electron] Reverts defaults PMIC settings changes, adds higher level power config API

avtolstoy:
### Problem

1. We've recently increased maximum current and decreased voltage limit when powered by VIN by default in order to allow Electrons and Borons to function without a battery, as well as fixed a problem with Borons incorrectly detecting when they are powered by VIN vs USB Host See https://github.com/particle-iot/device-os/pull/1846 https://github.com/particle-iot/device-os/pull/1921

2. Applications requiring special PMIC settings cannot reliably set them, because of System Power Manager within DeviceOS. There currently exists a workaround for this, but it's not ideal https://community.particle.io/t/1-3-1-rc-1-change-to-boron-solar-charging-via-vusb-behavior/51987/5?u=avtolstoy

These two problems together prevent certain applications e.g. solar.

### Solution

This PR does a number of things:

We've reverted the changes to the default PMIC settings we use when powered by VIN. These are now again 3.88V min, 900mA max for the PMIC to enter DPM mode and reduce current consumption.

When powered by USB host or USB-compliant adapter, the current limit is automatically negotiated with the other side and voltage input limit is still set to 3.88V.

Another addition is a higher level power configuration API allowing to notify the system about the expected PSU parameters. These settings are persisted and work correctly even without the application running (e.g. while the device is in safe mode). See 'Example App' section for the example usage.

This PR also adds a couple of methods to query current power source, battery state and state of charge more easily. See 'Example App' section as well.

### Steps to Test

- `wiring/api`
- `wiring/power_management`
- See also example apps

### Example App

```c++
void setup() {
    // Apply custom configuration
    // We are specifically using the values that don't exactly match what
    // the PMIC can use as settings and validating that they are being
    // mapped correctly later on.
    conf.powerSourceMaxCurrent(550) // Set maximum current the power source can provide (applies only when powered through VIN)
        .powerSourceMinVoltage(4300) // Set minimum voltage the power source can provide (applies only when powered through VIN)
        .batteryChargeCurrent(850) // Set battery charge current
        .batteryChargeVoltage(4210) // Set battery termination voltage
        .feature(SystemPowerFeature::PMIC_DETECTION) // For platforms with external PMIC/FuelGauge (e.g. BSoM), enables
                                                     // runtime PMIC presence detection and enables power management
                                                     // if found.
                                                     // NOTE: This supersedes the System.enable(SYSTEM_FLAG_PM_DETECTION) call
        .feature(SystemPowerFeature::USE_VIN_SETTINGS_WITH_USB_HOST) // For the cases where the device is powered through VIN
                                                                     // but the USB cable is connected to a USB host, this feature flag
                                                                     // enforces the voltage/current limits specified in the configuration
                                                                     // (where by default the device would be thinking that it's powered by the USB Host)
    int res = System.setPowerConfiguration(conf); // returns SYSTEM_ERROR_NONE (0) in case of success

    // To restore the default configuration
    System.setPowerConfiguration(SystemPowerConfiguration());

    // Settings are persisted.
}

void loop() {
    {
        PMIC power(true);
        Log.trace("Current PMIC settings:");
        Log.trace("VIN Vmin: %u", power.getInputVoltageLimit());
        Log.trace("VIN Imax: %u", power.getInputCurrentLimit());
        Log.trace("Ichg: %u", power.getChargeCurrentValue());
        Log.trace("Iterm: %u", power.getChargeVoltageValue());

        int powerSource = System.powerSource();
        int batteryState = System.batteryState();
        float batterySoc = System.batteryCharge();
        
        };
            "usb otg", "battery"
        };

        Log.trace("Power source: %s", powerSources[std::max(0, powerSource)]);
        Log.trace("Battery state: %s", batteryStates[std::max(0, batteryState)]);
        Log.trace("Battery charge: %f", batterySoc);
    }
    delay(1000);
}
```

```c++
void setup() {
    // See system_power.h for these definitions
    // typedef enum {
    //     BATTERY_STATE_UNKNOWN = 0,
    //     BATTERY_STATE_NOT_CHARGING = 1,
    //     BATTERY_STATE_CHARGING = 2,
    //     BATTERY_STATE_CHARGED = 3,
    //     BATTERY_STATE_DISCHARGING = 4,
    //     BATTERY_STATE_FAULT = 5,
    //     BATTERY_STATE_DISCONNECTED = 6
    // } battery_state_t;

    // typedef enum {
    //     POWER_SOURCE_UNKNOWN = 0,
    //     POWER_SOURCE_VIN = 1,
    //     POWER_SOURCE_USB_HOST = 2,
    //     POWER_SOURCE_USB_ADAPTER = 3,
    //     POWER_SOURCE_USB_OTG = 4,
    //     POWER_SOURCE_BATTERY = 5
    // } power_source_t;

    int powerSource = System.powerSource();
    int batteryState = System.batteryState();
    float batterySoc = System.batteryCharge();
    
    };
        "usb otg", "battery"
    };

    Log.trace("Power source: %s", powerSources[std::max(0, powerSource)]);
    Log.trace("Battery state: %s", batteryStates[std::max(0, batteryState)]);
    Log.trace("Battery charge: %f", batterySoc);
}

void loop() {

}
```

### References

- [CH42646]
- https://community.particle.io/t/1-3-1-rc-1-change-to-boron-solar-charging-via-vusb-behavior/51987/5?u=avtolstoy

---

### Completeness

- [ ] **Added documentation**


avtolstoy:
@YutingYou @sergeuz Can I get your review on this?

YutingYou:
Hi @avtolstoy , sorry for the late review, I‘ll do it this week once we solve the flow control issue.

platform/MCU/nRF52840/inc/dct.h
```diff
@@ -107,9 +108,9 @@ typedef struct __attribute__((packed)) application_dct {
     uint8_t device_secret[15];           // Device secret data (aka "mobile secret")
     uint8_t setup_done;                  // Setup done flag
     uint32_t ncp_id;                     // NCP identifier
-    uint8_t pm_detect;                   // Detect power management peripherals flag
+    hal_power_config power_config;       // Power management configuration
     uint8_t radio_antenna;               // Mesh/BLE antenna: 0x01 - internal, 0x02 - external, 0xff - default
```
YutingYou: The power_config broke the radio_antenna setting. `selectAntenna(radio_antenna_type antenna)`  doesn't handle the default flash value 0xFF, so the GPIO of antenna switch will be left in the default mode.

YutingYou:
Power detection worked correctly when powered by VIN or USB host individually, but when it is powered by VIN and USB host together, the Input Voltage Limit was set to 3880mV. If the user wants to read some data via USB on a device that is powered by solar, after removing the USB, the Input Voltage Limit is set to the low value.

Should we detect the change of USB state in the `PowerManager::loop()`?

avtolstoy:
@YutingYou Please take a look at the example:
```cpp
        .feature(SystemPowerFeature::USE_VIN_SETTINGS_WITH_USB_HOST) // For the cases where the device is powered through VIN
                                                                     // but the USB cable is connected to a USB host, this feature flag
                                                                     // enforces the voltage/current limits specified in the configuration
                                                                     // (where by default the device would be thinking that it's powered by the USB Host)
```

>  after removing the USB, the Input Voltage Limit is set to the low value.

This does sound like an oversight :+1: I'll fix this.

platform/MCU/nRF52840/inc/dct.h
avtolstoy: This is deliberate. This feature (antenna setting) has not been released in any public DeviceOS release, so we can still make such changes.

YutingYou: I see, we have the default setting `RADIO_ANT_DEFAULT` in `initRadioAntenna()`, mark it resolved.

hal/shared/power_hal.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include "power_hal.h"
+#include "dct.h"
+#include "check.h"
+#include "system_error.h"
+#include <algorithm>
+
+int hal_power_load_config(hal_power_config* conf, void* reserved) {
+    CHECK_TRUE(conf, SYSTEM_ERROR_INVALID_ARGUMENT);
+    int err = dct_read_app_data_copy(DCT_POWER_CONFIG_OFFSET, conf, conf->size);
```
YutingYou: Nitpick: why not determine the conf size in `hal_power_load_config `, it's convenient that the user doesn't need to set the size. And it's safe for memory copy.

avtolstoy: For backwards compatibility. The stored configuration may potentially be larger (newer version) than what the caller can handle.

YutingYou: That make sense, 👍

PR #1994, opened 2020-01-10
[Gen 3] Logging modem AT commands and data TX/RX by default on modular Boron/B SoM [ch33624]

technobly:
### Problem

As a user of a Gen 3 device, I would like the logging of modem AT commands and data TX/RX to be included in firmware by default on modular Boron/B SoM.

Unfortunately there is not enough space in modular firmware to enable all logging for Boron/B SoM, which is why it's only enabled for monolithic firmware.

### Solution

Enable a minimal subset of logs for Boron/B SoM by default in modular firmware:
- AT commands and URCs
- PPP Client TX and RX bytes

### Steps to Test

Local Build outputting logs on TX (Serial1):
`device-os/modular $ make clean all -s PLATFORM=boron COMPILE_LTO=n APP=tinker-serial1-debugging program-dfu`

### Example App

```c
// Build as modular app
// Outputting logs on USB (Serial)

void setup() {
}
void loop() {
}
```

### References

ch33624

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/ppp_client.cpp
+++ b/hal/network/lwip/ppp_client.cpp
@@ -31,6 +31,10 @@ extern "C" {
 #include "system_error.h"
 #include "lwiplock.h"
 
+#undef LOG_COMPILE_TIME_LEVEL
+#define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
+LOG_SOURCE_CATEGORY("net.ppp.client");
+
 using namespace particle::net::ppp;
 
 std::once_flag Client::once_;
@@ -168,6 +172,7 @@ int Client::input(const uint8_t* data, size_t size) {
       case STATE_CONNECTING:
       case STATE_DISCONNECTING:
       case STATE_CONNECTED: {
+        LOG(TRACE, "RX: %lu", size);
         err_t err = pppos_input_tcpip(pcb_, (u8_t*)data, size);
         if (err) {
           return SYSTEM_ERROR_INTERNAL;
@@ -355,7 +360,7 @@ uint32_t Client::outputCb(ppp_pcb* pcb, uint8_t* data, uint32_t len, void* ctx) 
 }
 
 uint32_t Client::output(const uint8_t* data, size_t len) {
-  LOG_DEBUG(TRACE, "Outputing %lu bytes", len);
+  LOG(TRACE, "TX: %lu", len);
 
   if (oCb_) {
     auto r = oCb_(data, len, oCbCtx_);
--- a/hal/network/ncp/at_parser/at_parser_impl.cpp
+++ b/hal/network/ncp/at_parser/at_parser_impl.cpp
@@ -30,7 +30,9 @@
 #include <cctype>
 #include <cassert>
 
-LOG_SOURCE_CATEGORY("ncp.at")
+#undef LOG_COMPILE_TIME_LEVEL
+#define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
+LOG_SOURCE_CATEGORY("ncp.at");
 
 namespace particle {
 
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -39,6 +39,9 @@
 
 #include <algorithm>
 
+#undef LOG_COMPILE_TIME_LEVEL
+#define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
+
 #define CHECK_PARSER(_expr) \
         ({ \
             const auto _r = _expr; \


hal/network/lwip/ppp_client.cpp
```diff
@@ -31,6 +31,10 @@ extern "C" {
 #include "system_error.h"
 #include "lwiplock.h"
 
+#undef LOG_COMPILE_TIME_LEVEL
+#define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL // include errors
```
avtolstoy: Why only errors? :) It will include trace, info, warning, error and critical levels.

hal/network/lwip/ppp_client.cpp
```diff
@@ -31,6 +31,10 @@ extern "C" {
...
+LOG_SOURCE_CATEGORY("net.ppp.client")
```
avtolstoy: Better to add `;` so that some editors/viewers e.g. Github don't get confused.

hal/network/lwip/ppp_client.cpp
```diff
@@ -168,6 +172,7 @@ int Client::input(const uint8_t* data, size_t size) {
       case STATE_CONNECTING:
       case STATE_DISCONNECTING:
       case STATE_CONNECTED: {
+        LOG(TRACE, "RX: %lu", size);
```
avtolstoy: Do we need this in non-debug builds? I would argue that this seems too verbose for every packet. Same in https://github.com/particle-iot/device-os/pull/1994/files#diff-a2527e86865833a37b99004691cf38a7R363

hal/src/nRF52840/lwip/lwip_logging.h
```diff
@@ -28,10 +28,6 @@ void lwip_log_message(const char *fmt, ...);
 }
 #endif /* __cplusplus */
 
-#ifdef DEBUG_BUILD
```
avtolstoy: This also has a potential to pollute things too much as it enables logging for every LwIP subsystem. What's more this duplicates the TX/RX logs mentioned above. May affect the performance as well (e.g. PPP logs are very verbose in case of integrity issues and will produce a torrent of lines).

avtolstoy:
Just an idea: perhaps we should add a default filter list where by default certain categories will be disabled (or have a lower logging level) until explicitly enabled by the application.

hal/network/lwip/ppp_client.cpp
technobly: Some might have thought to just enable TRACE for AT commands, so it was a comment to that effect. Perhaps I’ll just remove the comments on these.

technobly: Some might have thought to just enable TRACE for AT commands, so it was a comment to that effect. Perhaps I’ll just remove the comments on these.

hal/network/lwip/ppp_client.cpp
technobly: Pretty sure I just copied this from one of the other files but I’m happy to add ;

hal/network/lwip/ppp_client.cpp
technobly: These logs are already present in the form of the LWIP layer, but harder to enable by this method. I’d say there’s much less logs in the modular build, just the right amount. There was already a ton in monolithic so I don’t think one extra RX to match the TX that was already there will hurt. Plus I reduced the text from “Outputing XX bytes” to “TX: XX”

technobly: These logs are already present in the form of the LWIP layer, but harder to enable by this method. I’d say there’s much less logs in the modular build, just the right amount. There was already a ton in monolithic so I don’t think one extra RX to match the TX that was already there will hurt. Plus I reduced the text from “Outputing XX bytes” to “TX: XX”

hal/src/nRF52840/lwip/lwip_logging.h
technobly: Hmm, I didn’t see that in testing. What build option would cause this? I tried modular without debug and mono with debug build.

technobly: Hmm, I didn’t see that in testing. What build option would cause this? I tried modular without debug and mono with debug build.

hal/network/lwip/ppp_client.cpp
avtolstoy: Yeah, we have a lot of places where we don't have it, but you can see in the diff that it messes things up.

technobly:
I would prefer not to reduce the log output received for LOG_LEVEL_ALL but agree that some of the PPP/LWIP logging could be cleaned up to be more succinct. I think that is beyond the scope of this PR though.

hal/src/nRF52840/lwip/lwip_logging.h
avtolstoy: Ah, you are right. My bad. For this to work `LWIP_DEBUG` and other `XXX_DEBUG` options need to be enabled in `lwipopts.h`. So, what does this enable us then..?

hal/network/lwip/ppp_client.cpp
technobly: I typically do use that convention with macros, you can never have enough semicolons in C/C++ I’ll fix it in the morning.

avtolstoy:
@technobly I agree it is beyond the scope of the PR, just something we might consider doing going forward.

hal/network/lwip/ppp_client.cpp
technobly: Also I would say we do need this for minimal debugging purposes and it can be easily filtered away.

hal/src/nRF52840/lwip/lwip_logging.h
technobly: Maybe nothing :-) originally I was after the “lwip” logs, so I should restore this. Will fix in the morning. Thanks for calling it out.

hal/network/lwip/ppp_client.cpp
technobly: Updated and removed the comments.

hal/network/lwip/ppp_client.cpp
technobly: Added ;

hal/src/nRF52840/lwip/lwip_logging.h
technobly: Turns out it wasn't needed, thanks!  Fixed.

Issue #1993, opened 2019-12-29
I think `BleUuid::shorted()` is implemented wrong


ScruffR:
Assuming `BleUuid::shorted()` (which may be better called `::shortened()`, `::shorten()` or `::short()`) is meant to provide the 16bit UUID from a 128bit UUID it would need to *not* report the least siginificant two bytes of the 128 bits but actually the top 3rd and 4th (1st to 4th would be the 32bit) most siginificant.

According to the BT Core Specs a 128bit UUID is calculated as follows
```
128_bit_value = 16_bit_value * 2^96 + Bluetooth_Base_UUID
// or
128_bit_value = 32_bit_value * 2^96 + Bluetooth_Base_UUID
```
where `00000000-0000-1000-8000-00805F9B34FB` is the `Bluetooth_Base_UUID`.

Hence the 16 bit (short) UUID should be this part of any given 128bit UUID
```
0000xxxx-0000-1000-8000-00805F9B34FB
```
*BUT* `BleUuid::shorted()` always returns `0x34FB` which is *NOT* the 16bit short UUID but the least siginificant 16bit of the Base UUID - which is of no use for anything whatsoever - that's a fix value that could be hard coded just as well.

This may also impact `BlePeerDevice::getCharacteristicByUUID()` when the UUID is down-converted to 16bit the UUID will not be found as it won't be unique anymore.

A test code running on two (peripheral and central) Xenons 1.4.4 shows that the characteristics won't be found correctly 
_(a somewhat shorter test code can be found in my second post)_
```
#define CENTRAL             // comment out to create a mockup sensor with ModeChange and LiveSensorData characteristics
#define CORRECT_BEHAVIOUR   // comment out if not working correctly

#include "Particle.h"

SYSTEM_MODE(MANUAL)
SYSTEM_THREAD(ENABLED)

SerialLogHandler log(LOG_LEVEL_ALL);

const uint16_t vendor = 0xFE95;
const BleUuid serviceUuidRoot   ("0000FE95-0000-1000-8000-00805F9B34FB");

const BleUuid serviceUuidGeneric("00001800-0000-1000-8000-00805F9B34FB");
const BleUuid charUuid_r0x0003  ("00002800-0000-1000-8000-00805F9B34FB");

const BleUuid serviceUuidData   ("00001204-0000-1000-8000-00805F9B34FB");
const BleUuid charUuid_w0x0033  ("00001A00-0000-1000-8000-00805F9B34FB");
[...]

PR #1997, opened 2020-01-14
Fixes BLE UUID conversion issue.

XuGuohui:
### Problem

As detailed here: https://github.com/particle-iot/device-os/issues/1993

### Solution

Retrieves the 12th and 13th byte of a long UUID as its short UUID.

### Steps to Test

Running unit test: wiring/no_fixture_ble:

>Running tests
Test BLE_01_Set_BLE_Device_Address passed.
Test BLE_02_Set_BLE_Device_Name passed.
Test BLE_03_Set_BLE_Tx_Power passed.
Test BLE_04_Select_BLE_Antenna passed.
Test BLE_05_Set_BLE_Advertising_Parameters passed.
Test BLE_06_Set_BLE_Advertising_Data passed.
Test BLE_07_Set_BLE_Scan_Response_Data passed.
Test BLE_08_BLE_Advertising_Control passed.
Test BLE_09_Set_BLE_Scanning_Parameters passed.
Test BLE_10_Add_BLE_Local_Characteristics passed.
Test BLE_11_BLE_UUID_Conversion passed.
Test summary: 11 passed, 0 failed, and 0 skipped, out of 11 test(s).

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup()
{
    Serial.begin(115200);
    while(!Serial.isConnected());
    Serial.println("Application started.");

    BleUuid uuid("6E401234-B5A3-F393-E0A9-E50E24DCCA9E");
    Serial.printlnf("Short UUID: 0x%04X", uuid.shorted());

    BleUuid shortUuid(0x5678);
    Serial.printlnf("Short UUID: 0x%04X", shortUuid.shorted());
}

void loop()
{
}
```

### References

https://github.com/particle-iot/device-os/issues/1993

---

### Completeness

- [ ] Added documentation


wiring/src/spark_wiring_ble.cpp
```diff
@@ -328,15 +328,28 @@ hal_ble_uuid_t BleUuid::halUUID() {
...
 uint16_t BleUuid::shorted() const {
-    return uuid_.uuid16;
+    if (type() == BleUuidType::SHORT) {
+        return uuid_.uuid16;
+    } else {
+        return ((uint16_t)uuid_.uuid128[12] | ((uint16_t)uuid_.uuid128[13] << 8));
+    }
 }
 
 void BleUuid::rawBytes(uint8_t uuid128[BLE_SIG_UUID_128BIT_LEN]) const {
```
avtolstoy: Should this return the number of bytes copied?

XuGuohui: I was just aware of this. But should it be fixed in this PR? Will do it as long as it makes sense to you.

avtolstoy: I'm not sure that's why I'm asking :) It just felt weird that we require a `BLE_SIG_UUID_128BIT_LEN` buffer but copy only 2 bytes.

XuGuohui: I agree. As a workaround, we can add a private member for the `BleUuid` class, which is an 16-bytes length of array to store the 128-bits UUID that is derived from short UUID.

avtolstoy: I was just going to suggest that.

1. `BleUuid::rawBytes(uint8_t uuid128[BLE_SIG_UUID_128BIT_LEN])` should generate a 128-bit UUID derived from short UUID and copy into the supplied buffer
2. `BleUuid::rawBytes()` should return a pointer to 128-bit UUID derived from short UUID. Ideally we should re-use the same `hal_ble_uuid_t` for storage though, otherwise it seems rather wasteful. This should probably be doable as long as we stick to a convention that `BleUuid` class stores everything as 128-bit UUIDs and perfroms any necessary conversions to/from short form on the boundaries (API and HAL).

wiring/inc/spark_wiring_ble.h
```diff
@@ -253,7 +253,12 @@ class BleUuid {
     void construct(const char* uuid);
     void toBigEndian(uint8_t buf[BLE_SIG_UUID_128BIT_LEN]) const;
 
-    hal_ble_uuid_t uuid_;
+    const uint8_t BASE_UUID[BLE_SIG_UUID_128BIT_LEN] = {0xFB,0x34,0x9B,0x5F,0x80,0x00, 0x00,0x80, 0x00,0x10, 0x00,0x00, 0x00,0x00,0x00,0x00};
```
avtolstoy: Prefer `constexpr` for better optimizations.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -308,35 +316,49 @@ BleUuid::BleUuid(const String& uuid)
...
 
 uint16_t BleUuid::shorted() const {
-    return uuid_.uuid16;
+    return ((uint16_t)uuid128_[UUID16_LO] | ((uint16_t)uuid128_[UUID16_HI] << 8));
 }
 
-void BleUuid::rawBytes(uint8_t uuid128[BLE_SIG_UUID_128BIT_LEN]) const {
-    memcpy(uuid128, uuid_.uuid128, BLE_SIG_UUID_128BIT_LEN);
+size_t BleUuid::rawBytes(uint8_t uuid128[BLE_SIG_UUID_128BIT_LEN]) const {
+    if (type() == BleUuidType::SHORT) {
```
avtolstoy: I thought we agreed that this should always return full 128-bit UUID and apply the base uuid to the shortened ones?

wiring/inc/spark_wiring_ble.h
```diff
@@ -253,7 +253,12 @@ class BleUuid {
...
+    const uint8_t UUID16_LO = 12;
+    const uint8_t UUID16_HI = 13;
+
+    uint8_t uuid128_[BLE_SIG_UUID_128BIT_LEN];
```
avtolstoy: Any reason not to keep using `hal_ble_uuid_t` and using `BLE_UUID_TYPE_128BIT_SHORTED` to indicate that it's a 16-bit UUID expanded to 128 bit to avoid unnecessary copying?

XuGuohui: Since the 16-bits UUID and 128-bits UUID is organized in a union struct. And in BLE HAL where a `hal_ble_uuid_t` being referenced it just uses `.uuid16` to get shorted UUID, which will actually return the lowest two bytes of the expanded 128-bits UUID. If we just pass the private `hal_ble_uuid_t` member to BLE HAL, it would result incorrect 16-bits UUID in BLE HAL. And FYI, the `BLE_UUID_TYPE_128BIT_SHORTED` is actually a temporary hack to resolve unknown 128-bits vendor UUID, so that BLE HAL can retrieve complete 128-bits UUID afterwards before exposing it to user.

XuGuohui: We can also use `hal_ble_uuid_t` as a member of wiring BLE UUID class. But we need to refactor `hal_ble_uuid_t` to be that 16-bits UUID and 128-bits UUID are organized as separated member of `hal_ble_uuid_t`. Does it make sense?

avtolstoy: I think it's fine to leave things as-is, but I'm concerned with too much copying going on.

Is there any problem with using `BLE_UUID_TYPE_128BIT_SHORTED` as a convention in both wiring and HAL that it's a 16-bit UUID expanded to 128 bits? For simplicity in wiring we'd always be expanding 16-bit UUIDs into this type.

PR #2006, opened 2020-01-29
[b5som] hal: quectel modems don't support CHAP:<APN> in CGDCONT

avtolstoy:
### Problem

Quectel modems don't support `CHAP:` option for `AT+CGDCONT` in the form: `AT+CGDCONT=<id>,<type>,CHAP:<apn>`.

This is something that has been copied over from ublox NCP client and should be removed.

This will cause problems (i.e. inability to get cellular connectivity due to `ERROR` response from the modem) for custom credentials with username and password as well as with empty credentials and Kore SIM cards due to our IMSI-based credentials store (see `network_config_db.cpp`: `{ "204", "04", "vfd1.korem2m.com", "kore", "kore" } // Kore/Vodafone`).

```
0000078901 [ncp.at] TRACE: > AT+CGDCONT=1,"IP","CHAP:vfd1.korem2m.com"
0000078949 [ncp.at] TRACE: < ERROR
```

### Solution

Remove `CHAP:` generation for `AT+CGDCONT`.

### Steps to Test

Use Kore SIM (e.g. internal), clear the credentials, the device should not connect on `develop`, `AT+CGDCONT`` should case an error. It should connect fine with this branch.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/b5som/network/quectel_ncp_client.cpp
+++ b/hal/src/b5som/network/quectel_ncp_client.cpp
@@ -727,8 +727,7 @@ int QuectelNcpClient::configureApn(const CellularNetworkConfig& conf) {
         netConf_ = networkConfigForImsi(buf, strlen(buf));
     }
     // FIXME: for now IPv4 context only
-    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s%s\"",
-            (netConf_.hasUser() && netConf_.hasPassword()) ? "CHAP:" : "",
+    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s\"",
             netConf_.hasApn() ? netConf_.apn() : "");
     const int r = CHECK_PARSER(resp.readResult());
     CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);


hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -721,8 +721,7 @@ int QuectelNcpClient::configureApn(const CellularNetworkConfig& conf) {
         netConf_ = networkConfigForImsi(buf, strlen(buf));
     }
     // FIXME: for now IPv4 context only
-    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s%s\"",
-            (netConf_.hasUser() && netConf_.hasPassword()) ? "CHAP:" : "",
+    auto resp = parser_.sendCommand("AT+CGDCONT=1,\"IP\",\"%s\"",
             netConf_.hasApn() ? netConf_.apn() : "");
     const int r = CHECK_PARSER(resp.readResult());
```
keeramis: @avtolstoy While this is a good solution, why not disable hasUser and hasPassword? Because hasUser and hasPassword are being returned in cellular_credentials_get(), and if we are not going to use only user and password, we might convey that information through the API call also?

avtolstoy: I did not get any of that :sweat_smile: What do you mean by 'disable hasUser and hasPassword'?

keeramis: In this line - 
> (netConf_.hasUser() && netConf_.hasPassword()) ? "CHAP:" : "",

So if user and password are true for the module, we use a "CHAP:" in the APN name. So if we are completely not going to user name and password for quectel modules or b5soms, then instead of removing that line, we can just set hasUser and hasPasswrod to false, no?

Additionally for example, we are also using hasUser here in cellular_credentials_get(). If we are not going to use username and password for quectel modems on b5soms altogether, we can set them false, and that way the users can access that info from the API call also.
https://github.com/particle-iot/device-os/blob/77f129979ad51544690da04976ad6992c5261cfa/hal/network/ncp/cellular/cellular_hal.cpp#L239

keeramis: We can get rid of them here for quectel / b5soms?? 

https://github.com/particle-iot/device-os/blob/77f129979ad51544690da04976ad6992c5261cfa/hal/network/ncp/cellular/network_config_db.cpp#L39

keeramis: like add a #define or whichever is the best way.

avtolstoy: Keerthy and I discussed why this is not needed in a short meeting. Marking as resolved.

PR #2007, opened 2020-01-29
Adds internal function to modify cellular registration timeout

eberseth:
### Problem

Cellular registration timeout is hardcoded to 5 minutes.  There may be instances where this needs to be specified by the user for their particular carrier/application.

### Solution

Add cellular_registration_timout_set() API to override the 5 minute default timeout.

### Example App

```c
#include "Particle.h"
SYSTEM_THREAD(ENABLED);
void setup() {
  cellular_registration_timeout_set(60 * 60 * 1000, nullptr); // 60 minutes
}
```

### References

- [CH45850]

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -143,7 +143,7 @@ bool cellular_sim_ready(void* reserved);
  * 
  * @param timeout: registration timeout in milliseconds
  */
-cellular_result_t cellular_registration_timeout_set(system_tick_t timeout, void* reserved = NULL);
+cellular_result_t cellular_registration_timeout_set(system_tick_t timeout, void* reserved);
 
 /**
  * Attempts to stop/resume the cellular modem from performing AT operations.
--- a/hal/inc/cellular_hal.h
+++ b/hal/inc/cellular_hal.h
@@ -139,6 +139,13 @@ cellular_result_t cellular_global_identity(CellularGlobalIdentity* cgi, void* re
 bool cellular_sim_ready(void* reserved);
 
 /**
+ * Change the timeout a SIM has to register on the cellular network before the modem is reset
+ * 
+ * @param timeout: registration timeout in milliseconds
+ */
+cellular_result_t cellular_registration_timeout_set(system_tick_t timeout, void* reserved = NULL);
+
+/**
  * Attempts to stop/resume the cellular modem from performing AT operations.
  * Called from another thread or ISR context.
  *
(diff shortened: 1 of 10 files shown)


hal/inc/cellular_hal.h
```diff
@@ -138,6 +138,13 @@ cellular_result_t cellular_global_identity(CellularGlobalIdentity* cgi, void* re
 
 bool cellular_sim_ready(void* reserved);
 
+/**
+ * Change the timeout a SIM has to register on the cellular network before the modem is reset
+ * 
+ * @param timeout: registration timeout in milliseconds
+ */
+cellular_result_t cellular_registration_timeout_set(system_tick_t timeout, void* reserved = NULL);
```
avtolstoy: This is a C header and a C function, there cannot be a default argument, which is a C++-specific feature.

eberseth: Good catch

eberseth: Modular and monolithic builds will need separate calls.  One with a nullptr and the other without.

avtolstoy: Why? It's a standard practice in DeviceOS to leave a `void* reserved` for ABI compatibility and future extensibility and to require the caller to fill that with `nullptr` by default. What is the difference between monolithic and modular builds in this regard?

eberseth: Fixed

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -504,6 +505,11 @@ int QuectelNcpClient::getSignalQuality(CellularSignalQuality* qual) {
     return SYSTEM_ERROR_NONE;
 }
 
+int QuectelNcpClient::setRegistrationTimeout(unsigned timeout) {
+    registrationTimeout_ = std::max(timeout, REGISTRATION_CHECK_INTERVAL);
```
technobly: @monkbroc @eberseth @avtolstoy This `registrationTimeout_` value should be limited to >= 5 minutes.  As written here, the registration timeout could be reduced to 15 seconds.  Simply need to change `REGISTRATION_CHECK_INTERVAL` to `REGISTRATION_TIMEOUT`.  This needs to be changed in both the `quectel_ncp_client.cpp` and `sara_ncp_client.ncp`

monkbroc: OK. There was no minimum discussed so we decided to use the 15 second check interval, but I understand that using 5 minutes makes sense if we only allow extending the registration timeout.

monkbroc: https://github.com/particle-iot/device-os/pull/2014

PR #2011, opened 2020-01-31
Removes asom and xsom platforms

avtolstoy:
### References

- [CH45106]

---

### Completeness

- [ ] Added documentation


wiring/src/spark_wiring_wifitester.cpp
```diff
@@ -392,7 +392,7 @@ void WiFiTester::checkWifiSerial(char c) {
             if (ok) {
                 if (!strcmp("ALL", parts[1])) {
 #if PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON || \
-    PLATFORM_ID == PLATFORM_XSOM || PLATFORM_ID == PLATFORM_ASOM || PLATFORM_ID == PLATFORM_BSOM
+    PLATFORM_ID == PLATFORM_BSOM
```
XuGuohui: Can we use `#if HAL_PLATFORM_NRF52840` instead? I think it's also applicable for B5SoM.

avtolstoy: I think we need to review this part honestly for each platform, because it's not clear whether A6-A7 and D8 should be supported for each particular platform.

user/applications/tinker/application.cpp
```diff
@@ -52,19 +52,11 @@ const PinMapping g_pinmap[] = {
...
-#  elif PLATFORM_ID == PLATFORM_XSOM
-    ,
-    PIN(D24), PIN(D25), PIN(D26), PIN(D27), PIN(D28), PIN(D29), PIN(D30), PIN(D31)
-#  endif // PLATFORM_ID == PLATFORM_XSOM
+# if PLATFORM_ID == PLATFORM_BSOM || PLATFORM_ID == PLATFORM_B5SOM
     ,
+    PIN(D20), PIN(D21), PIN(D22), PIN(D23),
     PIN(A6), PIN(A7)
-# endif // PLATFORM_ID == PLATFORM_XSOM || PLATFORM_ID == PLATFORM_ASOM || PLATFORM_ID == PLATFORM_BSOM
+# endif // PLATFORM_ID == PLATFORM_BSOM
```
XuGuohui: Nitpicking: ` || PLATFORM_ID == PLATFORM_B5SOM`.

platform/MCU/nRF52840/inc/platform_config.h
```diff
@@ -76,8 +74,7 @@
 #define PREPSTRING(x) PREPSTRING2(x)
 
 #if PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON || \
-    PLATFORM_ID == PLATFORM_XSOM || PLATFORM_ID == PLATFORM_ASOM || PLATFORM_ID == PLATFORM_BSOM || \
-    PLATFORM_ID == PLATFORM_B5SOM
+    PLATFORM_ID == PLATFORM_BSOM || PLATFORM_ID == PLATFORM_B5SOM
```
XuGuohui: How about including `hal_platform_nrf52840_config.h` and use `#if HAL_PLATFORM_NRF52840`? Or even we can move the `hal_platform_nrf52840_config.h` into this header file, as it is a platform-specific header file.

platform/MCU/nRF52840/inc/platform_config.h
```diff
@@ -35,9 +35,7 @@
 #if PLATFORM_ID == PLATFORM_ARGON || \
```
XuGuohui: See: https://github.com/particle-iot/device-os/pull/2011/files#r373967791

hal/src/xenon/pinmap_hal.c
```diff
@@ -64,60 +64,6 @@ const uint8_t NRF_PIN_LOOKUP_TABLE[48] = {
 
 #endif // PLATFORM_ID == PLATFORM_XENON
```
XuGuohui: We can probably remove this macro condition. Same source file with regard the Argon platform.

hal/src/xenon/pinmap_defines.h
```diff
@@ -87,85 +87,3 @@
 #define ANTSW2 30
 
 #endif // PLATFORM_ID == PLATFORM_XENON
```
XuGuohui: We can probably remove this macro condition. Same header file with regard to the Argon platform.

avtolstoy:
@XuGuohui Thanks, I'll clean those up and merge.

platform/MCU/nRF52840/inc/platform_config.h
avtolstoy: This is located already in `nRF52840`-specific folder, so I'm going to leave as-is. For an unknown nRF52840-based platform this will cause a compiler error, which sounds like a good reminder to add a new platform here explicitly.

platform/MCU/nRF52840/inc/platform_config.h
avtolstoy: Same. I'm going to leave this as-is because it's a header in nRF52840 platform specific folder.

PR #2022, opened 2020-02-11
Introduce configurable I2C buffer

zfields:
<details>

```

```
</details>

### Problem

Increase or Make Custom I2C Buffer Length

### Solution

The buffer will now be allocated at runtime, and passed into the
`HAL_I2C_Init` function via the existing `reserved` parameter.
`HAL_I2C_Init` is currently invoked during the initialization of
the `TwoWire` object, which is instantiated at program startup.

A weakly linked function `acquireI2cBuffer` has been introduced
in the Wiring layer, and can be overridden by the user application.
This function is used to allocate and supply buffers to the I2C
interface via the `HAL_I2C_Init` function.

As of this commit, access to the buffer is limited to 255, due to
the data type `uint8_t` used in the `quantity` parameter of the
`HAL_I2C_Request_Data` API. As a result, the buffer is effectively
limited to 255 bytes.

### Steps to Test

Run `wiring/i2c_master_slave` test suite on Gen2 and Gen3 devices in all combinations (slave/master/buffer size)

The following `BUFFER_SIZE` combinations need to be tested:

1. Default buffer allocation: `make PLATFORM=XXX TEST=wiring/i2c_master_slave/i2c_master clean all`
2. Compatibility allocation: `make PLATFORM=XXX TEST=wiring/i2c_master_slave/i2c_master BUFFER_SIZE=0 clean all`
3. Custom allocation: `make PLATFORM=XXX TEST=wiring/i2c_master_slave/i2c_master BUFFER_SIZE=254 clean all`

**NOTE:** 255 is maximum buffer size for this PR.

### Example App

N/A

### References

Mini PRD in related CH story [ch42311](https://app.clubhouse.io/particle/story/42311/increase-or-make-custom-i2c-buffer-length).

---

### Completeness

- [ ] Added documentation


hal/inc/i2c_hal.h
```diff
@@ -44,19 +46,36 @@ typedef enum HAL_I2C_Interface {
...
+typedef enum HAL_I2C_Config_Version {
+    HAL_I2C_CONFIG_VERSION_1 = 0,
+    HAL_I2C_CONFIG_VERSION_LATEST = HAL_I2C_CONFIG_VERSION_1,
+} HAL_I2C_Config_Version;
+
+typedef struct HAL_I2C_Config {
+    uint16_t size;
+    uint16_t version;
+    uint8_t* rx_buffer;
+    uint32_t rx_buffer_size;
```
XuGuohui: Prefer using `size_t`. Same to `tx_buffer_size`.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -27,15 +27,17 @@
...
 
 #define TOTAL_I2C                   2
-#define BUFFER_LENGTH               I2C_BUFFER_LENGTH
 #define I2C_IRQ_PRIORITY            APP_IRQ_PRIORITY_LOWEST
-#define EVENT_TIMEOUT               100000 // 100ms
+#define EVENT_TIMEOUT               (HAL_I2C_DEFAULT_TIMEOUT_MS * 1000)
 
 #define WAIT_TIMED(what) ({ \
-    uint32_t _micros = HAL_Timer_Get_Micro_Seconds();                           \
+    system_tick_t _micros = HAL_Timer_Get_Micro_Seconds();                           \
```
XuGuohui: Nitpicking: format.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -235,31 +239,61 @@ static int twi_init(HAL_I2C_Interface i2c) {
...
+        return SYSTEM_ERROR_NONE;
     }
 
+    // Capture the mutex and re-enable threading
     HAL_I2C_Acquire(i2c, NULL);
     os_thread_scheduling(true, NULL);
 
+    // Initialize internal data structure
+    if (HAL_I2C_Config_Is_Valid(config)) {
+        m_i2c_map[i2c].rx_buf = config->rx_buffer;
```
XuGuohui: This buffer can potentially be freed by user application. Same to `config->tx_buffer`.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -235,31 +239,61 @@ static int twi_init(HAL_I2C_Interface i2c) {
...
+        m_i2c_map[i2c].rx_buf_size = config->rx_buffer_size;
+        m_i2c_map[i2c].tx_buf = config->tx_buffer;
+        m_i2c_map[i2c].tx_buf_size = config->tx_buffer_size;
+    } else {
+        // Allocate default buffers
+        m_i2c_map[i2c].rx_buf = new (std::nothrow) uint8_t[I2C_BUFFER_LENGTH];
```
XuGuohui: It might be `nullptr`. Same to `m_i2c_map[i2c].tx_buf`.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -299,6 +299,9 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
             // m_spi_map[spi].ss_pin = PIN_INVALID;
         }
     } else {
+        if (pin >= TOTAL_PINS) {
```
XuGuohui: This seams to be resided in another PR. Should it be included in this PR?

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -181,54 +188,99 @@ static void HAL_I2C_SoftwareReset(HAL_I2C_Interface i2c)
...
 }
 
-void HAL_I2C_Init(HAL_I2C_Interface i2c, void* reserved)
+static bool HAL_I2C_Config_Is_Valid(const HAL_I2C_Config* config) {
+    if ((config == NULL) ||
+            (config->rx_buffer == NULL ||
+             config->rx_buffer_size == 0 ||
+             config->tx_buffer == NULL ||
+             config->tx_buffer_size == 0))
+    {
```
XuGuohui: Nitpicking: move it to following the `if` expression.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -181,54 +188,99 @@ static void HAL_I2C_SoftwareReset(HAL_I2C_Interface i2c)
...
+        return false;
+    }
+
+    return true;
+}
+
+static bool HAL_I2C_Buffers_Initialized(HAL_I2C_Interface i2c) {
+    if (i2cMap[i2c]->rxBuffer == NULL ||
```
XuGuohui: Nitpicking: nullptring.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -181,54 +188,99 @@ static void HAL_I2C_SoftwareReset(HAL_I2C_Interface i2c)
...
         HAL_I2C_Acquire(i2c, NULL);
         os_thread_scheduling(true, NULL);
     }
-#endif // PLATFORM_ID == 10
+#endif // PLATFORM_ID == PLATFORM_ELECTRON_PRODUCTION
 
+    // Initialize internal data structure
+    if (HAL_I2C_Config_Is_Valid(config))
+    {
+        i2cMap[i2c]->rxBuffer = config->rx_buffer;
```
XuGuohui: This buffer can potentially be freed by user application. Same to `config->tx_buffer`.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -181,54 +188,99 @@ static void HAL_I2C_SoftwareReset(HAL_I2C_Interface i2c)
...
+        i2cMap[i2c]->rxBufferSize = config->rx_buffer_size;
+        i2cMap[i2c]->txBuffer = config->tx_buffer;
+        i2cMap[i2c]->txBufferSize = config->tx_buffer_size;
+    } else if (!HAL_I2C_Buffers_Initialized(i2c)) {
+        i2cMap[i2c]->rxBuffer = (uint8_t*)malloc(I2C_BUFFER_LENGTH);
```
XuGuohui: This might return `nullptr`. Same to `i2cMap[i2c]->txBuffer`.

user/tests/wiring/i2c_master_slave/common/common.inc
```diff
@@ -0,0 +1,53 @@
...
+#ifndef TEST_I2C_BUFFER_SIZE
+#define TEST_I2C_BUFFER_SIZE (I2C_BUFFER_LENGTH)
+#else
+
+HAL_I2C_Config USE_ACQUIRE_WIRE_BUFFER() {
+#if TEST_I2C_BUFFER_SIZE != 0
+    HAL_I2C_Config config = {
+		.size = sizeof(HAL_I2C_Config),
+		.version = HAL_I2C_CONFIG_VERSION_1,
+		.rx_buffer = new (std::nothrow) uint8_t[TEST_I2C_BUFFER_SIZE],
```
XuGuohui: This might be `nullptr`.

wiring_globals/src/wiring_globals_i2c.cpp
```diff
@@ -8,16 +8,52 @@
 
 #ifndef SPARK_WIRING_NO_I2C
 
+namespace {
+
+HAL_I2C_Config defaultWireConfig() {
+	HAL_I2C_Config config = {
+		.size = sizeof(HAL_I2C_Config),
+		.version = HAL_I2C_CONFIG_VERSION_1,
+		.rx_buffer = new (std::nothrow) uint8_t[I2C_BUFFER_LENGTH],
```
XuGuohui: This might be `nullptr`.

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: What do you propose we do? It's not supposed to be freed.

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: I agree, though the only sane thing we can probably do is panic in this case.

hal/src/nRF52840/spi_hal.cpp
avtolstoy: :+1: I'll remove it.

hal/src/stm32f2xx/i2c_hal.c
avtolstoy: This is a C source file.

hal/src/stm32f2xx/i2c_hal.c
avtolstoy: See https://github.com/particle-iot/device-os/pull/2022#discussion_r381789506

hal/src/nRF52840/i2c_hal.cpp
XuGuohui: We should allocate the buffer in HAL and return error if failed to do that, instead of inheriting the pointer to the buffer that is allocated in user application. We cannot guarantee that user don't free the buffer.

hal/inc/i2c_hal.h
avtolstoy: There is no real difference and it is fine to affix the 'size' variables to `uint32_t` or any other type if needed.

user/tests/wiring/i2c_master_slave/common/common.inc
avtolstoy: It should be ok, because HAL will handle this case.

wiring_globals/src/wiring_globals_i2c.cpp
avtolstoy: https://github.com/particle-iot/device-os/pull/2022#discussion_r381793729

hal/src/nRF52840/i2c_hal.cpp
XuGuohui: Besides, the `config->rx_buffer` can be a temporary buffer allocated in stack in user application. In that case, the buffer is freed automatically. Unless we document it with bold font.

avtolstoy: > We should allocate the buffer in HAL and return error if failed to do that, instead of inheriting the pointer to the buffer that is allocated in user application. 

I disagree. We follow the same API style we already have for increasing the buffer size for Serial: https://docs.particle.io/reference/device-os/firmware/electron/#acquireserialbuffer-

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: Fixed.

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: Added assertion.

hal/src/nRF52840/spi_hal.cpp
avtolstoy: Rebased and removed.

hal/src/stm32f2xx/i2c_hal.c
avtolstoy: Fixed.

hal/src/stm32f2xx/i2c_hal.c
avtolstoy: Added assertion.

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: This also has the additional benefit of not allocating these buffers if the application does not use I2C at all.

avtolstoy: Btw, USART buffers are also provided by the application: https://github.com/particle-iot/device-os/blob/develop/wiring/src/spark_wiring_usartserial.cpp#L134

XuGuohui: We can allocate the buffer in `HAL_I2C_Init()` and free it in `HAL_I2C_Reset()`. That wouldn't consume RAM if I2C is not used. It's fine to leave it as-is, and just keep in mind that there is still risk to mess up the RAM data if the buffer is free unexpectedly.

avtolstoy: Let's leave it as-is and just make sure that the documentation clearly states that these buffers need to be preserved.

The difference between allocating completely in HAL vs in application, is that the application may provide statically allocated buffers, which somewhat controls the potential heap allocation errors. Additionally custom schemes may be used where a buffer is shared between multiple peripherals as long as they are not used concurrently.

PR #2023, opened 2020-02-13
[wiring] Fixes a regression in SPI due to thread-safety additions from #1879

### LINKED ISSUE #1668: Arduino library compatibility needs #define SPI_HAS_TRANSACTION 1
Many recent Arduino libraries rely on the existence of `#define SPI_HAS_TRANSACTION 1` when the platform supports SPI transactions.
Since Particle devices do, we should also have that macro defined (in `Particle.h` not only `Arduino.h`)

See this discussion
https://community.particle.io/t/electron-ssd1351-oled-sdcard-issues-with-adafruit-ssd1351-sdfat-libs/46798/2?u=scruffr

e.g. these libraries
https://github.com/technologiescollege/arduino/blob/master/hardware/arduino/avr/libraries/SPI/src/SPI.h
https://github.com/adafruit/Adafruit-GFX-Library/blob/master/Adafruit_SPITFT_Macros.h
https://github.com/adafruit/Adafruit_VS1053_Library/blob/master/Adafruit_VS1053.cpp
https://github.com/adafruit/Adafruit_SSD1305_Library/blob/master/Adafruit_SSD1305.cpp
https://github.com/esp8266/Basic/blob/master/libraries/Adafruit_STMPE610/Adafruit_STMPE610.cpp

avtolstoy:
### Problem

1. Using SPI object in global object constructors leads to a crash due to undefined initialization order of global objects
2. `pinAvailable()` calls `SPI.isEnabled()`, potentially from interrupt context.
3. #1668

### Solution

1. Use singleton-ish type of initialization for SPI objects with a function returning a reference to static object
2. Remove locking from `SPI.isEnabled()`, since in HAL it is anyway implemented with a volatile variable.
3. Adds `SPI_HAS_TRANSACTION` define

### Steps to Test

N/A

### Example App

N/A

### References

- https://community.particle.io/t/updating-from-deviceos-1-4-4-to-1-5-0-rc-1-sos-panic-1/54566
- #1879
- Closes #1668 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_arduino.h
+++ b/wiring/inc/spark_wiring_arduino.h
@@ -72,4 +72,8 @@
 #define pgm_read_word_near(x) ((uint16_t)(*(x)))
 #endif
 
+#ifndef SPI_HAS_TRANSACTION
+#define SPI_HAS_TRANSACTION (1)
+#endif // SPI_HAS_TRANSACTION
+
 #endif	/* SPARK_WIRING_ARDUINO_H */
--- a/wiring/inc/spark_wiring_spi.h
+++ b/wiring/inc/spark_wiring_spi.h
@@ -262,25 +262,33 @@ public:
 
 #ifndef SPARK_WIRING_NO_SPI
 
-extern SPIClass SPI;
+namespace particle {
+namespace globals {
+
+SPIClass& instanceSpi();
+#define SPI ::particle::globals::instanceSpi()
 
 #if Wiring_SPI1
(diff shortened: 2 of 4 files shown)


elcojacobs:
My application would not boot and hang or hardfault on 1.5.0-rc1. Just compiled against develop and issue seems to be resolved! :+1:

avtolstoy:
Thanks for confirming!

PR #2024, opened 2020-02-13
Describe message fixes

sergeuz:
### Problem

- Wrap all checksums that describe a cached application state (firmware modules, functions, variables, subscriptions) into a class (`AppStateDescriptor`). Implement a method to compare cached and current application states so that each field is compared individually (previously, that code was relying on a single CRC32 checksum derived from all other checksums).

- For each of the application state fields, introduce a persistently stored flag that explicitly determines whether the field is set or not (see `SessionPersistData::app_state_flags`).

- Store protocol flags in the persistent session data (`SessionPersistData::protocol_flags`) so that a change in the protocol configuration would trigger a new Hello message to be sent to the server.

- **Automatically send Describe messages when changes in the application/system state are detected (see notes below).**

- **Send Describe messages requested by the server as confirmable CoAP responses.** A failure to deliver a Describe message to the server is a critical connection error.

- **Update cached application state only after corresponding Describe messages have been acknowledged by the server.** See [this](https://s.slack.com/archives/CKRRAGTSB/p1573489983281100) thread in Slack for details.

- Fix the issue with devices always performing full handshake in the `AUTOMATIC` non-threaded mode.

**Note:** In the `AUTOMATIC` mode, the `setup()` function is not executed until the connection with cloud is fully established. In this mode, the application's Describe message will be sent immediately after the application returns from `setup()`. Depending on the threading mode, this may happen either during the handshake with the cloud or immediately after it.

### Steps to Test

_TBD_

### References

- [ch45275]


elijthomas:
💥 @sergeuz You are a super star ! I cant wait to see this one pushed across the line 👍 One epic PR 💯

system/src/system_cloud_internal.cpp
```diff
@@ -1064,23 +1090,21 @@ int Spark_Handshake(bool presence_announce)
         if (presence_announce) {
             Multicast_Presence_Announcement();
         }
-        LOG(INFO,"Send subscriptions");
-        spark_protocol_send_subscriptions(sp);
-        // important this comes at the end since it requires a response from the cloud.
         spark_protocol_send_time_request(sp);
-        Spark_Process_Events();
```
sergeuz: @avtolstoy I couldn't find out why it might be important to send this at the end of the application handshake and why we run a loop iteration afterwards, so any thoughts are welcomed. So far it looks as a safe change.

avtolstoy: Isn't this call here just to make sure that the comms event loop is immediately run? I don't see any special reason why we should call it immediately though.

sergeuz: ah, found it: https://github.com/particle-iot/device-os/commit/2cde8bc02d08d5f4d74dcb5cc1ec10ef8c6da6f4

communication/inc/dtls_session_persist.h
```diff
@@ -187,8 +194,8 @@ class __attribute__((packed)) SessionPersist : public SessionPersistOpaque
 
 	void clear(save_fn_t saver)
 	{
+		memset(this, 0, sizeof(*this));
```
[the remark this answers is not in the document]
sergeuz: Fixed

communication/src/protocol.cpp
```diff
@@ -62,11 +79,49 @@ ProtocolError Protocol::handle_received_message(Message& message,
 		LOG(TRACE, "Reply recieved: type=%d, code=%d", type, code);
 		// todo - this is a little too simple in the case of an empty ACK for a separate response
 		// the message should then be bound to the token. see CH19037
-		if (type==CoAPType::RESET) {		// RST is sent with an empty code. It's like an unspecified error
+		if (type == CoAPType::RESET) { // RST is sent with an empty code. It's like an unspecified error
 			LOG(TRACE, "Reset received, setting error code to internal server error.");
 			code = CoAPCode::INTERNAL_SERVER_ERROR;
 		}
 		notify_message_complete(msg_id, code);
+		if (descriptor.app_state_selector_info) {
```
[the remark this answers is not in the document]
sergeuz: Fixed

PR #2025, opened 2020-02-13
Dynalib documentation

monkbroc:
The application binary interface between system firmware binary and user firmware binary called Dynalib is not currently described. This document is a start to describe the goals, implementation and rules of Dynalib.


avtolstoy:
@monkbroc Thanks for starting this. I'll take over the PR and edit a few things to clarify/detail some exceptions where the changes are valid to make.

(I was planning on making a presentation and turning that into documentation :sweat_smile:)

PR #2028, opened 2020-02-17
[wiring] SPI: fixes issue that the default SS pin is incorrect callin…

XuGuohui:
### Problem

`SPI.begin(SPI_MODE_MASTER);` doesn't configure the default SS pin to be output high initially.

### Solution

Assign default parameter for the second argument of the `SPIClass::begin(SPI_Mode mode, uint16_t ss_pin = SPI_DEFAULT_SS)`.

### Steps to Test

1. Build and download the following example App.
2. Check if the default SS pin according to https://docs.particle.io/reference/device-os/firmware/electron/#begin--4 can be toggled periodically.

### Example App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);

#if HAL_PLATFORM_STM32F2XX
#define SPI_DEFAULT_SS_PIN  A2
#elif PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON
#define SPI_DEFAULT_SS_PIN  A5
#endif

void setup()
{
    SPI.begin(SPI_MODE_MASTER);
}

void loop()
{
    digitalWrite(SPI_DEFAULT_SS_PIN, LOW);
    delay(3000);
    digitalWrite(SPI_DEFAULT_SS_PIN, HIGH);
    delay(3000);
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/spi_hal.cpp
+++ b/hal/src/nRF52840/spi_hal.cpp
@@ -299,6 +299,9 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
             // m_spi_map[spi].ss_pin = PIN_INVALID;
         }
     } else {
+        if (pin >= TOTAL_PINS) {
+            return;
+        }
         m_spi_map[spi].ss_pin = pin;
     }
 
--- a/hal/src/stm32f2xx/spi_hal.c
+++ b/hal/src/stm32f2xx/spi_hal.c
@@ -289,8 +289,13 @@ void HAL_SPI_Begin(HAL_SPI_Interface spi, uint16_t pin)
 
 void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void* reserved)
 {
-    if (pin == SPI_DEFAULT_SS)
+    if (pin == SPI_DEFAULT_SS) {
         pin = spiMap[spi].SPI_SS_Pin;
+    }
+
+    if (pin >= TOTAL_PINS) {
+        return;
+    }
 
     spiState[spi].SPI_SS_Pin = pin;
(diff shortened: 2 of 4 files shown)


hal/src/nRF52840/spi_hal.cpp
```diff
@@ -299,6 +299,9 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
             // m_spi_map[spi].ss_pin = PIN_INVALID;
         }
     } else {
+        if (pin >= TOTAL_PINS) {
```
avtolstoy: Why are we removing all the checks for `TOTAL_PINS`? Especially in HAL, which will break backwards compatibility.

It's entirely possible to pass `PIN_INVALID` as the pin or some other invalid number.

XuGuohui: I'm don't how it breaks the backwards compatibility. By moving the check routine from wiring to HAL just gaining the time that the wiring API acquiring/releasing the lock. Even if user pass invalid pin number, the HAL can still handle it correctly. So from user's perspective, the result calling the wiring API makes no difference. The root cause of the issue is that the `begin()` APIs are not type safe, so when user calling `SPI.begin(SPI_MODE_MASTER)`, it will call the `SPIClass::begin(uint16_t ss_pin)` instead. To fix it I just assign a default value `SPI_DEFAULT_SS` for the second argument of the `SPIClass::begin(SPI_Mode mode, uint16_t ss_pin)` API, so that it can invoke the corresponding overloaded function. If we don't move the check routine to HAL, the wiring API will return immediately if calling `SPI.begin(SPI_MODE_MASTER)`.

avtolstoy: You are right. I mistakenly thought we also removed this check here, not added it :smile:

:+1:

PR #2029, opened 2020-02-17
[Gen 3] Fixes renegotiation of options and adds IPCP packet parsing

avtolstoy:
### Problem

1. We sometimes see that the devices establish the PPP link, however remain stuck in blinking green. This is caused by the remote side changing offered DNS servers during the negotiation (i.e. it offers one set first and then after we accept if, for some reason offers another).
2. IPCP packets are not easy to debug and are dumped raw
3. [B5 SoM] For some reason we are not receiving CREG/CGREG/CEREG URCs on muxed AT channel and they only appear in the muxed data channel, which causes us to not notice loss of carrier.

### Solution

1. All default IPCP options (IP addresses, Primary/Secondary DNS) are now overridable by the remote side. Once the PPP link is established, we now also check whether we've agreed on a set of DNS servers, and if not, teardown the link.
2. IPCP packet parsing is added.
3. [B5 SoM] Adds `AT+QCFG="cmux/urcport",1` to receive URCs only on muxed AT channel.

This PR also disables IPV6CP.

### Steps to Test

1. Force the device (b5som) into a state where it'll throw `Peer NAK'd an option unknown to us`, it should connect succesfully
2. Build monolithic DeviceOS with `DEBUG_BUILD=y`, IPCP packet dumps should be visible in the logs

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/ppp_ipcp.cpp
+++ b/hal/network/lwip/ppp_ipcp.cpp
@@ -50,7 +50,7 @@ Ipcp::~Ipcp() {
 }
 
 void Ipcp::enable() {
-  LOG(TRACE, "IPCP: enable %d %d %d", admState_, lowerState_, state_);
+  LOG(TRACE, "enable %d %d %d", admState_, lowerState_, state_);
   if (!admState_) {
     admState_ = true;
     if (lowerState_ && !state_) {
@@ -60,7 +60,7 @@ void Ipcp::enable() {
 }
 
 void Ipcp::disable() {
-  LOG(TRACE, "IPCP: disable %d %d %d", admState_, lowerState_, state_);
+  LOG(TRACE, "disable %d %d %d", admState_, lowerState_, state_);
   if (admState_) {
     admState_ = false;
     if (lowerState_ && state_) {
@@ -74,18 +74,18 @@ void Ipcp::setDnsEntryIndex(int idx) {
 }
 
 void Ipcp::init() {
-  LOG(TRACE, "IPCP: init");
+  LOG(TRACE, "init");
 
   fsm_init(&fsm_);
 }
 
 void Ipcp::input(uint8_t* pkt, int len) {
-  LOG(TRACE, "IPCP: input %d", len);
+  LOG(TRACE, "input %d", len);
   fsm_input(&fsm_, pkt, len);
 }
 
 void Ipcp::protocolReject() {
-  LOG(TRACE, "IPCP: Protocol-Reject");
+  LOG(TRACE, "Protocol-Reject");
   /* RFC 1661:
    * Upon reception of a Protocol-Reject, the implementation MUST stop
    * sending packets of the indicated protocol at the earliest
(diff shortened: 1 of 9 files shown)


avtolstoy:
@technobly @keeramis ping.

hal/network/lwip/ppp_ipcp.cpp
```diff
@@ -116,14 +116,14 @@ void Ipcp::open() {
...
-  LOG(TRACE, "ipcp:close");
+  LOG(TRACE, "close");
   fsm_close(&fsm_, reason);
 }
 
 /* State machine callbacks */
 /* Reset our Configuration Information */
 void Ipcp::resetConfigurationInformation() {
-  LOG(TRACE, "IPCP: reset ci");
+  LOG(TRACE, "Resetting CI");
```
keeramis: Nitpicking: I would use lower case for consistency.

hal/network/lwip/ppp_ipcp.cpp
```diff
@@ -173,7 +172,7 @@ int Ipcp::getConfigurationInformationLength() {
     }
   });
 
-  LOG(TRACE, "IPCP: Our CI length: %lu", len);
+  LOG(TRACE, "Our CI length: %lu", len);
```
keeramis: Nitpicking: I would use lower case for consistency.

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -695,6 +695,10 @@ int QuectelNcpClient::initReady() {
 
     muxerSg.dismiss();
 
+    // Make sure that we receive URCs only on AT channel, ignore response code
+    // just in case
+    CHECK_PARSER(parser_.execCommand("AT+QCFG=\"cmux/urcport\",1"));
```
keeramis: I remember that we were discussing about this command officially only supported on BG96. I do not see any reference in the EG91. Is it worth making sure this is officially supported on the EG91 just in case a future radio-version does not support it?

hal/src/nRF52840/lwip/lwippppopts.h
```diff
@@ -127,7 +131,7 @@
 /**
  * PPP_IPV6_SUPPORT==1: Enable PPP IPv6 support
  */
-#define PPP_IPV6_SUPPORT                (LWIP_IPV6)
+#define PPP_IPV6_SUPPORT                0
```
keeramis: General question - Looks like we need to turn on IPV6 packet support manually if it is needed. In general, how do we figure out that we need this logging?

hal/src/b5som/network/quectel_ncp_client.cpp
avtolstoy: For that case specifically, you can see that there is no check for the return OK status, so it's 'best effort': if it works, it works :)

hal/src/nRF52840/lwip/lwippppopts.h
avtolstoy: We don't support IPv6 anyway at the moment on any interface other than Mesh. This completely disables PPP IPv6 support, not just the logging. We can always flip the switch back any time needed.

PR #2030, opened 2020-02-18
Change the prefix of BLE broadcasting name for B5SoM.

XuGuohui:
### Problem

The prefix of BLE broadcasting name is `B5som-`, while we want it to be `Boron-`

### Solution

Refactor the `get_device_name()` API in HAL.

### Steps to Test

1. Build and flash the test application: `make PLATFORM=b5som TEST=app/ble/uart_peripheral`
2. Use mobile App to scan BLE peripherals and see if your B5SoM is broadcasting as `Boron-xxxxxx`.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/shared/platforms.h
```diff
@@ -39,5 +39,40 @@
...
+#endif
+
+#if PLATFORM_ID == PLATFORM_XENON
+#define PRODUCT_SERIES                      "xenon"
+#endif
+
+#if PLATFORM_ID == PLATFORM_NEWHAL
+#define PRODUCT_SERIES                      "newhal"
+#endif
+
```
avtolstoy: `#error` if not set.

hal/shared/platforms.h
```diff
@@ -39,5 +39,40 @@
 #define PLATFORM_B5SOM                      25
 #define PLATFORM_NEWHAL                     60000
 
+#if PLATFORM_ID == PLATFORM_GCC
+#define PRODUCT_SERIES                      "gcc"
+#endif
+
+#if PLATFORM_ID == PLATFORM_PHOTON_DEV || \
```
avtolstoy: We should probably not bother with anything other than `P1` and `PHOTON_PRODUCTION`.

hal/shared/platforms.h
```diff
@@ -39,5 +39,40 @@
 #define PLATFORM_B5SOM                      25
 #define PLATFORM_NEWHAL                     60000
 
+#if PLATFORM_ID == PLATFORM_GCC
```
avtolstoy: Would per-platform `hal_platform_config.h` potentially be a better choice for this?

PR #2031, opened 2020-02-18
[Gen 2] Fixes socket leak in UDP.begin()

avtolstoy:
### Problem

`UDP::begin()` can be called when already active with an open socket. It will unconditionally create open a new socket and discard its reference to the old socket while leaving it open.

If this sequence is repeated the system can run out of socket handles preventing new UDP sockets from being opened and even preventing cloud connection.

While users should call `UDP::end()` to close any open socket before a new `UDP::begin()` the implementation should defensively check for and close an open socket on `UDP::begin()`.

**NOTE: this bug only applies to Gen 2 devices**

### Solution

Call `stop()` in `begin()`.

This PR also frees up some flash space in `wiring/no_fixture` test suite for cellular platforms, as we started to overflow.

### Steps to Test

Run `wiring/no_fixture` tests on Gen 2 and Gen 3 platforms. All tests can be excluded (`E` command) except for `MDM_01_socket_writes_with_length_more_than_1023_work_correctly` on cellular platforms and `UDP_begin_does_not_leak_sockets_without_calling_stop` on all the platforms (`i` command to enable the tests selectively).

### Example App

N/A

### References

- [CH27231]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/cellular.cpp
+++ b/user/tests/wiring/no_fixture/cellular.cpp
@@ -20,6 +20,7 @@
 #include "application.h"
 #include "unit-test/unit-test.h"
 #include "socket_hal.h"
+#include "random.h"
 
 #if Wiring_Cellular == 1
 bool skip_r410 = false;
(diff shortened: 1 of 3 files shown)


user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -190,16 +205,17 @@ test(MDM_01_socket_writes_with_length_more_than_1023_work_correctly) {
...
 
-    char* responseBuf = new char[2048];
-    memset(responseBuf, 0, 2048);
+    auto responseBuf = std::make_unique<char[]>(bufferSize);
+    assertTrue((bool)responseBuf);
+    memset(responseBuf.get(), 0, bufferSize);
     int responseSize = 0;
     uint32_t mil = millis();
     while(1) {
         while (c.available()) {
```
eberseth: What to do if responseSize >= bufferSize?

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -169,7 +168,16 @@ test(MDM_01_socket_writes_with_length_more_than_1023_work_correctly) {
 #endif // HAL_PLATFORM_NCP
 
     // https://github.com/spark/firmware/issues/1104
-    const char request[] =
+
+    const int dataSize = 1024;
+    const int bufferSize = 2048;
+    auto randData = std::make_unique<char[]>(1024);
```
eberseth: Use dataSize to initialize instead of magic number.

eberseth:
Can you fix or mute the Codacy/PR review issues?

avtolstoy:
> Can you fix or mute the Codacy/PR review issues?

Just ignore them.

user/tests/wiring/no_fixture/cellular.cpp
avtolstoy: Fixed.

user/tests/wiring/no_fixture/cellular.cpp
avtolstoy: Fixed.

PR #2033, opened 2020-02-18
Replace CSQ on Gen2 with UCGED (mode 5)

keeramis:
<details>

```

```
</details>

### Problem

Signal strength on Gen2 is collected using CSQ command. 
1. This command does not provide quality value on Gen2
2. This command is quite unreliable, meaning that is sometimes gives invalid signal strength value in an expectedly solid cell coverage location. 
u-blox recommends to use UCGED command, which is expected to work more reliably than the CSQ command. 

### Solution

Replace CSQ command with UCGED for R410M radio. UCGED gives RSRP/RSRQ in dBm/dB and CSQ gives in index values. RSRP/RSRQ from UCGED are converted from dBm/dB to CESQ-type index values and further processed. 

**Implementation:**
- Hard-bounding values for RSRP in [-200,0] and assign error value otherwise(255)
- Hard-bounding values for RSRQ in [?,0] and assign error value otherwise(255)
- Check that UCGED contains valid values for both RSRP and RSRQ string, else call it an error
- No retry algorithm on the UCGED command
- In a multiple-response UCGED command, uses the first set of values as the correct readings and ignore the rest

### Steps to Test

Manual testing required here
1. Check that correct signal strength and quality are reported by comparing the values reported by the previous Device OS and the values reported by your code for the same device in the same location.
2. Check an Electron 3G
3. Check the null case (no coverage)
4. Drive test 

### Example App

```
#include "application.h"

Serial1LogHandler logHandler (115200, LOG_LEVEL_ALL);

void setup() {
   Particle.publishVitals(30);
}

void loop() {
}
```

### References

https://app.clubhouse.io/particle/story/47681/replace-csq-on-gen2-with-ucged-mode-5

---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -96,6 +96,7 @@ std::recursive_mutex mdm_mutex;
 #define UMNOPROF_TIMEOUT  ( 1 * 1000)
 #define CEDRXS_TIMEOUT    (  1 * 1000)
 #define CFUN_TIMEOUT      (180 * 1000)
+#define UCGED_TIMEOUT     (  1 * 1000)
 
 // num sockets
 #define NUMSOCKETS      ((int)(sizeof(_sockets)/sizeof(*_sockets)))
@@ -965,6 +966,12 @@ bool MDMParser::init(DevStatus* status)
             waitFinalResp(); // Not checking for error since we will reset either way
             goto reset_failure;
         }
+
+        // For signal strength RSRP/RSRQ values on R410M
+        sendFormated("AT+UCGED=5\r\n");
+        if (RESP_OK != waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT)) {
+            goto failure;
+        }
         // Force Power Saving mode to be disabled
         //
         // TODO: if we enable this feature in the future update the logic in MDMParser::_atOk
@@ -1395,11 +1402,29 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
         if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
             goto failure;
         }
-        // get the signal strength indication
-        sendFormated("AT+CSQ\r\n");
-        if (RESP_OK != waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
-            goto failure;
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED=5\r\n");
+            if (RESP_OK != waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT)) {
+                goto failure;
+            }
+
+            // Default to 255 because UCGED response is multi-line
+            _net.rsrp = 255;
+            _net.rsrq = 255;
+            sendFormated("AT+UCGED?\r\n");
+            if (RESP_OK != waitFinalResp(_cbUCGED, &_net, UCGED_TIMEOUT)) {
+                goto failure;
+            }
         }
+        else
+        {
+            sendFormated("AT+CSQ\r\n");
+            if (RESP_OK != waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
+                goto failure;
+            }
+        }
+
         // +CREG, +CGREG, +COPS do not contain <AcT> for G350 devices.
         // Force _net.act to ACT_GSM to ensure Device Diagnostics and
         // RSSI() API's have a RAT for conversion lookup purposes.
@@ -1458,10 +1483,29 @@ bool MDMParser::getSignalStrength(NetStatus &status)
             _net.act = ACT_GSM;
         }
 
-        sendFormated("AT+CSQ\r\n");
-        if (RESP_OK == waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
-            ok = true;
-            status = _net;
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED=5\r\n");
+            if (RESP_OK != waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT)) {
+                goto cleanup;
+            }
+
+            // Default to 255 because UCGED response is multi-line
+            _net.rsrp = 255;
+            _net.rsrq = 255;
+            sendFormated("AT+UCGED?\r\n");
+            if (RESP_OK == waitFinalResp(_cbUCGED, &_net, UCGED_TIMEOUT)) {
+                ok = true;
+                status = _net;
+            }
+        }
+        else
+        {
+            sendFormated("AT+CSQ\r\n");
+            if (RESP_OK == waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
+                ok = true;
+                status = _net;
+            }
         }
     }
 
@@ -1727,6 +1771,52 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
         int a;
         if ((sscanf(buf, "\r\n+CNUM: \"My Number\",\"%31[^\"]\",%d", num, &a) == 2) &&
             ((a == 129) || (a == 145))) {
+        }
+    }
+    return WAIT;
+}
+
+int MDMParser::_cbUCGED(int type, const char* buf, int len, NetStatus* status)
+{
+    const int min_rsrq_mul_by_100 = -1950;
+    const int max_rsrq_mul_by_100 = -300;
+
+    if ((type == TYPE_PLUS || type == TYPE_UNKNOWN) && status) {
+        int rsrp;
+        int rsrq_n;
+        unsigned long rsrq_f;
+        // AT+UCGED?
+        // +RSRP: 314,5110,"-102.60",163,5110,"-097.90",173,5110,"-100.40",
+        // +RSRQ: 314,5110,"-19.60",163,5110,"-16.60",173,5110,"-18.60",
+
+        // RSRP maps from dBm [-141,-44] to [0,97]
+        // We are defining hard boundaries for RSRP to be between [-200,0],
+        // and consider other values as error and call it 255
+        if (sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"", &rsrp) == 1) {
+            if (rsrp < -141 && rsrp >= -200) {
+                status->rsrp = 0;
+            } else if (rsrp >= -44 && rsrp <=0) {
+                status->rsrp = 97;
+            } else if (rsrp >= -141 && rsrp < -44) {
+                status->rsrp = rsrp + 141;
+            } else {
+                status->rsrp = 255;
+            }
+        }
+        // RSRQ maps from dBm [-20,-3] to [0,34]
+        // We are defining hard boundaries for RSRP to be between [],
+        // and consider other values as error and call it 255
+        if (sscanf(buf, "\r\n+RSRQ: %*d,%*d,\"%d.%lu\"", &rsrq_n, &rsrq_f) == 2) {
+            int rsrq_mul_100 = rsrq_n * 100 - rsrq_f;
+            if (rsrq_mul_100 < min_rsrq_mul_by_100 && rsrq_mul_100 >= -2000) {
+                status->rsrq = 0;
+            } else if (rsrq_mul_100 >= max_rsrq_mul_by_100 && rsrq_mul_100 <= 0) {
+                status->rsrq = 34;
+            } else if (rsrq_mul_100 >= min_rsrq_mul_by_100 && rsrq_mul_100 < max_rsrq_mul_by_100) {
+                status->rsrq = (rsrq_mul_100 + 2000)/50;
+            } else {
+                status->rsrq = 255;
+            }
         }
     }
     return WAIT;
--- a/hal/src/electron/modem/mdm_hal.h
+++ b/hal/src/electron/modem/mdm_hal.h
@@ -557,6 +557,7 @@ protected:
     static int _cbUACTIND(int type, const char* buf, int len, int* i);
     static int _cbUDOPN(int type, const char* buf, int len, char* mccmnc);
     static int _cbCGPADDR(int type, const char* buf, int len, MDM_IP* ip);
+    static int _cbUCGED(int type, const char* buf, int len, NetStatus* status);
     struct CGDCONTparam { char type[8]; char apn[32]; };
     static int _cbCGDCONT(int type, const char* buf, int len, CGDCONTparam* param);
     static int _cbURAT(int type, const char *buf, int len, bool *matched_default);


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -964,6 +965,11 @@ bool MDMParser::init(DevStatus* status)
             sendFormated("AT+UMNOPROF=%d\r\n", UBLOX_SARA_UMNOPROF_SIM_SELECT);
             waitFinalResp(); // Not checking for error since we will reset either way
             goto reset_failure;
+        }
+		// For signal strength RSRP/RSRQ values on R410M
+        sendFormated("AT+UCGED=5\r\n");
+        if (RESP_OK != waitFinalResp()) {
```
keeramis: For `AT+UCGED?` query to work, we need to set `AT+UCGED=5` every time the radio powers up.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1395,11 +1401,21 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
         if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
             goto failure;
         }
-        // get the signal strength indication
-        sendFormated("AT+CSQ\r\n");
-        if (RESP_OK != waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED?\r\n");
+            if (RESP_OK != waitFinalResp(_cbUCGED, &_net, UCGED_TIMEOUT)) {
```
keeramis: Currently, I am only checking this once and reading the RSRP/RSRQ values, and calling them final. 

There is a bit of information about this command. The command as per their AT manual, should give a response like this: 
```
AT+UCGED?
+RSRP: 473,1000,"-105.10",
+RSRQ: 473,1000,"-20.00",
OK
```

But in practice, based on if it's at the edge of towers or moving, is like the following, where it gives multiple responses on the RSRP/RSRQ string: 
```
AT+UCGED?
+RSRP: 473,1000,"-105.10",193,1000,"-100.80",
+RSRQ: 473,1000,"-20.00",193,1000,"-15.80",
OK
```
The goal is to be able to pick the value of the serving LTE cell. However, which cell (among 473 and 193 in the example above) is the serving cell is not obvious. We checked with u-blox, and they mentioned that the first set of values (473 in this example) is always _likely_ the serving cell. 
Currently, I am only collecting the RSRP/RSRQ values from the first cell id in the string. Another approach ( as Brett suggested ) is to run this command a few times every second, and make sure we are always on the same serving cell before finalizing the value? But I hear that, that could cause other processes to wait until this is complete?

keeramis:
<details>

```

```
</details>

### Problem

Signal strength on Gen2 is collected using CSQ command. This command is quite unreliable, meaning that is sometimes gives invalid signal strength value in an expectedly solid cell coverage location. u-blox recommends to use UCGED command, which is expected to work more reliably than the CSQ command. 

### Solution

Replace CSQ command with UCGED for R410M radio. UCGED gives RSRP/RSRQ in dBm/dB and CSQ gives in index values. RSRP/RSRQ from UCGED are converted from dBm/dB to CESQ-type index values and further processed. 

### Steps to Test

Manual testing required here

### Example App

### References

https://app.clubhouse.io/particle/story/47681/replace-csq-on-gen2-with-ucged-mode-5


---

### Completeness

- [x] Added documentation

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1760,45 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+        if (sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d\"",&rsrp) == 1){
+            if (rsrp < -140) {
+                status->rsrp = 0;
+            }
+            else if (rsrp >= -44) {
+                status->rsrp = 97;
+            }
+            else {
+                status->rsrp = rsrp + 140;
+            }
```
avtolstoy: According to the documentation value of `255` indicates an error, which should be forwarded as-is, as that's also true for CSQ/CESQ. Same for RSRQ.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1760,45 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+        }
+        // RSRQ maps from dBm [-20,-3] to [0,34]
+        if (sscanf(buf, "\r\n+RSRQ: %*d,%*d,\"%d.%lu\"",&rsrq_n, &rsrq_f) == 2){
+            int rsrq_mul_100 = rsrq_n * 100 - rsrq_f;
+            if (rsrq_mul_100 < -19.5 * 100) {
```
avtolstoy: Nitpicking: this will be optimized by the compiler anyway, but probably better to have a non-floating point constant.

keeramis:
@avtolstoy Andrey - I made this ticket only for Gen2. Wanted to split the stories as it made a little simple for me. Gen3 is a separate ticket and coming. Thoughts on this?

hal/src/electron/modem/mdm_hal.cpp
keeramis: As per this particular AT command, there is no such thing as `255` for `UCGED Mode5`. I tested out removing the antenna and there is only `CME ERROR: no network service` or generic `ERROR`. 
And this is from the manual - 
```
<RSRP> - Reference Signal Received Power (RSRP) as defined in 3GPP TS 36.133 [105] expressed in dBm
<RSRQ> - Reference Signal Received Quality (RSRQ) as defined in 3GPP TS 36.133 [105] expressed in dB
```

But I can default the value to 255, and let the AT command change it when the value is collected from AT command. Please let me know what you think?

keeramis: In any case, when there is no coverage, it will error out and will not give an `OK`...

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1760,45 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+    if ((type == TYPE_PLUS || type == TYPE_UNKNOWN) && status) {
+        int rsrp;
+        int rsrq_n;
+        unsigned long rsrq_f;
+        // AT+UCGED?
+        // +RSRP: 314,5110,"-102.60",163,5110,"-097.90",173,5110,"-100.40",
+        // +RSRQ: 314,5110,"-19.60",163,5110,"-16.60",173,5110,"-18.60",
+
+        // RSRP maps from dBm [-140,-44] to [0,97]
+        if (sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d\"",&rsrp) == 1){
```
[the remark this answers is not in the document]
keeramis: I am guessing you are referring to the space after the `)` and before `{`. I added that in but please comment if it's not that.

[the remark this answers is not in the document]
keeramis: When I added `%*lu` for the fractional part, the compiler is complaining, so I am doing `%u`? :)

`use of assignment suppression and length modifier together in gnu_scanf format`

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: image

I was basing my comment on the SARA R4 AT commands reference ^^

avtolstoy: Let's default to 255 just in case and also handle UCGED potentially returning 255 for us.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
avtolstoy: The compiler is right :) Since we are suppressing assignment, `scanf` doesn't really care about the range/size of the argument variable and only about what characters may represent it.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Would it potentially make sense to run it every time before querying `AT+UCGED` like what we do in `getCellularGlobalIdentity()` with `AT+COPS=3,2`?

keeramis: We can also do that, but it _needs_ to be run only when the modem is ON.

hal/src/electron/modem/mdm_hal.cpp
keeramis: I defaulted the status->rsrp to 255 before we start calculating the value. And if there are no parsing errors, and if the value reasonably falls in 0 to 97 range which it will for about all int values :D , it will get updated in the `if` loop. 

Please see that: 
-(min_any_value) to -140 dBm : 0
-140 to -45 dBm : 1 to 96
-44 to -(max_any_value) : 97

hal/src/electron/modem/mdm_hal.cpp
keeramis: Me being OCD, so adding `UCGED=5` every time before calling `UCGED?`

keeramis:
@technobly ping!

hal/src/electron/modem/mdm_hal.cpp
technobly: My suggestion here is if we see multiple values, to leave the RSRP and RSRQ as their last known value.  If we read multiple values for a second time, then perhaps we should set the values to one that indicates an error since we don't trust the readings any longer.  One problem with this approach is we don't currently control the retry interval.  A user could be calling Particle.publishVitals() every 1 second, or maybe every 30s.  So that persisted value could hang around for much longer than it should.

Another approach would be to force one retry in Device OS after a delay of 300ms if the first read contained multiple values.  This might not be long enough for the tower hand off to have occurred though, so some characterization may need to be done here with mobile testing.  One possible negative affect to this approach is occasionally the Particle.publishVitals() API will block for a bit longer while it tries to resolve this issue.

The reason why I like this approach is because in practice when mobile testing, when I would see multiple values reported for RSRP and RSRQ, I would still be able to publish data and it was received by the server.  So if we ultimately will be using RSRP and RSRQ to help qualify network operations (publishes, etc..), I would like to avoid being overly cautious and optimize for being overly optimistic (due to previous testing).  Reason being, if we can't send certain messages due to being overly cautious, it can cause an even longer delay while we tear down the network connection and rebuild it.

Without having an explicit interpretation of what multiple responses means, we must characterize it and implement a strategy for interpreting it that best suits our customers.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1458,11 +1479,27 @@ bool MDMParser::getSignalStrength(NetStatus &status)
...
         }
 
-        sendFormated("AT+CSQ\r\n");
-        if (RESP_OK == waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
-            ok = true;
-            status = _net;
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED=5\r\n");
+            if (RESP_OK != waitFinalResp()) {
```
technobly: Needs `UCGED_TIMEOUT` or else this will default to 10s.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1395,11 +1402,25 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
...
             goto failure;
         }
-        // get the signal strength indication
-        sendFormated("AT+CSQ\r\n");
-        if (RESP_OK != waitFinalResp(_cbCSQ, &_net, CSQ_TIMEOUT)) {
-            goto failure;
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED=5\r\n");
+            if (RESP_OK != waitFinalResp()) {
```
technobly: Needs `UCGED_TIMEOUT` or else this will default to 10s.

hal/src/electron/modem/mdm_hal.cpp
technobly: If `255` is returned instead of `-NNN.NN`, I don't think `sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp) == 1` will be matched and 255 will never be set.  I think you'd potentially have to explicitly scan for it additionally with `sscanf(buf, "\r\n+RSRP: %*d,%*d,\"255\"",&rsrp) == 1` and set it explicitly `status->rsrp = 255;`

I don't think you'd want to implement it as a default case, e.g. `else { status->rsrp = 255; }` because there could be cases where you get back nothing, or garbage and that doesn't exactly equate to 255.

hal/src/electron/modem/mdm_hal.cpp
keeramis: As per the u-blox's AT command manual, the response time is not given, so I let it to default.
Any other suggestions than asking Flint or using 10? :) 
image

keeramis:
@technobly I am replying to this coz I could not see a reply option there - https://github.com/particle-iot/device-os/pull/2033#discussion_r386639823

I do not expect to see a 255 in the command for RSRP/RSRQ because the UCGED command only supposed to give values in dBm, and 255 is an indexed value (not dBm) from CSQ/CESQ command. Do you think the UCGED command will give a 255? 
Either way, it wouldn't hurt to do what you mentioned. I will do this (and handle the case justttt in case 255 doesn't come in quotations) - Does that look about correct? 
```
if ((sscanf(buf, "\r\n+RSRP: %*d,%*d,\"255\"",&rsrp) == 1) || 
                  (sscanf(buf, "\r\n+RSRP: %*d,%*d,255",&rsrp) == 1)) {
       status->rsrp = 255;
}
```

hal/src/electron/modem/mdm_hal.cpp
keeramis: @technobly
I do not expect to see a 255 in the command for RSRP/RSRQ because the UCGED command only supposed to give values in dBm, and 255 is an indexed value (not dBm) from CSQ/CESQ command. Do you think the UCGED command will give a 255? 
Either way, it wouldn't hurt to do what you mentioned. I will do this (and handle the case justttt in case 255 doesn't come in quotations) - Does that look about correct? 
```
if ((sscanf(buf, "\r\n+RSRP: %*d,%*d,\"255\"",&rsrp) == 1) || 
                  (sscanf(buf, "\r\n+RSRP: %*d,%*d,255",&rsrp) == 1)) {
       status->rsrp = 255;
}
```

hal/src/electron/modem/mdm_hal.cpp
keeramis: Had a chat about this with Julien when you were in EU, and Julien also mentioned about Particle.publishvitals going to have to be blocked until the signal strength approach is resolved. I do not fully understand the potential solution here - I do not know if Particle.publishvitals is ok to be blocked for some time. 

On the other hand - FAE from ublox mentioned that the first value is likely from the serving LTE cell (which is the cell we should be collecting values from).
Further questions on this: 
- Is it ok to let Particle.publishvitals be blocked? 
- If yes, how many times should be retry at 300ms / how much time can that call be blocked?

technobly: > On the other hand - FAE from ublox mentioned that the first value is likely from the serving LTE cell (which is the cell we should be collecting values from).
If we do some testing to corroborate that info, maybe that's the better method.  I could have sworn that I saw results that indicated the first value wasn't always the previous tower, or the next tower that the modem ended up selecting.

We should try our best not to block the call, so it might be something we create a 300ms+ timer for and process it on the system thread asynchronously.  I think one properly timed retry would be good.  Say we set it to 300ms, and keep getting multiple-"multiple responses" from UCGED.  Then the timing is not long enough for tower hand off.  300ms is the minimum retry time for UCGED per u-blox.  But perhaps it takes 500ms to reliably acquire the next tower... perhaps longer.  I know when I would poll every 2 seconds, I never saw multiple-multiples... so it's likely less than 2 seconds.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: There is no need for an additional sscanf for 255, it should be handled correctly by the current one as well. I would also argue that the calling code expects to get 255 if something went wrong (nothing or garbage).

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1769,52 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+            }
+            else if (rsrp >= -44) {
+                status->rsrp = 97;
+            }
+            else {
+                status->rsrp = rsrp + 140;
+            }
+        }
+        // RSRQ maps from dBm [-20,-3] to [0,34]
+        if (sscanf(buf, "\r\n+RSRQ: %*d,%*d,\"%d.%lu\"",&rsrq_n, &rsrq_f) == 2) {
```
avtolstoy: Nit: formatting :)

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1769,52 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+    if ((type == TYPE_PLUS || type == TYPE_UNKNOWN) && status) {
+        int rsrp;
+        int rsrq_n;
+        unsigned long rsrq_f;
+        // AT+UCGED?
+        // +RSRP: 314,5110,"-102.60",163,5110,"-097.90",173,5110,"-100.40",
+        // +RSRQ: 314,5110,"-19.60",163,5110,"-16.60",173,5110,"-18.60",
+
+        // RSRP maps from dBm [-140,-44] to [0,97]
+        if (sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp) == 1) {
```
avtolstoy: Nit: formatting.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1783,60 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+
+    if ((type == TYPE_PLUS || type == TYPE_UNKNOWN) && status) {
+        int rsrp;
+        int rsrq_n;
+        unsigned long rsrq_f;
+        // AT+UCGED?
+        // +RSRP: 314,5110,"-102.60",163,5110,"-097.90",173,5110,"-100.40",
+        // +RSRQ: 314,5110,"-19.60",163,5110,"-16.60",173,5110,"-18.60",
+
+        // RSRP maps from dBm [-140,-44] to [0,97]
```
technobly: Nit: -141,-44

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -115,6 +116,10 @@ std::recursive_mutex mdm_mutex;
 static volatile uint32_t gprs_timeout_start;
 static volatile uint32_t gprs_timeout_duration;
 
+// flags to check ucged command gives both rsrp and rsrq strings in its response
+static bool ucged_parse_rsrp_ok = false;
+static bool ucged_parse_rsrq_ok = false;
+
```
technobly: These probably do not need to be in global space.  They can be calculated fresh each time _cbUCGED is called with local variables (stack memory), and the rsrp and rsrq can be invalidated.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1458,11 +1487,33 @@ bool MDMParser::getSignalStrength(NetStatus &status)
...
+            ucged_parse_rsrp_ok = false;
+            ucged_parse_rsrq_ok = false;
+            sendFormated("AT+UCGED?\r\n");
+            if (RESP_OK == waitFinalResp(_cbUCGED, &_net, UCGED_TIMEOUT)) {
+                if (ucged_parse_rsrp_ok && ucged_parse_rsrq_ok)
+                {
+                    ok = true;
+                    status = _net;
+                }
+            }
```
technobly: `if !(ucged_parse_rsrp_ok && ucged_parse_rsrq_ok)` we should be ensuring rsrp AND rsrq are set to 255.  Again, this can all be handled in _cbUCGED.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1395,11 +1406,29 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
...
+            if (RESP_OK != waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT)) {
+                goto failure;
+            }
+            // Reset the ucged parsing flags before checking for UCGED command
+            ucged_parse_rsrp_ok = false;
+            ucged_parse_rsrq_ok = false;
+            sendFormated("AT+UCGED?\r\n");
+            if ((RESP_OK != waitFinalResp(_cbUCGED, &_net, UCGED_TIMEOUT))
+                            || (!ucged_parse_rsrp_ok || !ucged_parse_rsrq_ok)) { // rsrp or rsrq not parsed correctly in ucged
+                goto failure;
```
technobly: I think it's ok if ucged_parse_rsrp_ok and ucged_parse_rsrq_ok are false... we don't have to `goto failure;` in that case.  These should only be used to invalidate the rsrp and rsrq values.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1395,11 +1402,29 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
...
-            goto failure;
-        }
+        // AT command used to collect signal stregnth is different for R410M radio
+        if (_dev.dev == DEV_SARA_R410) {
+            sendFormated("AT+UCGED=5\r\n");
+            if (RESP_OK != waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT)) {
+                goto failure;
+            }
+
+			// Default to 255 because UCGED response is multi-line
```
avtolstoy: Formatting: tabs/spaces.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1732,6 +1776,58 @@ int MDMParser::_cbCNUM(int type, const char* buf, int len, char* num)
...
+        // +RSRQ: 314,5110,"-19.60",163,5110,"-16.60",173,5110,"-18.60",
+
+        // RSRP maps from dBm [-141,-44] to [0,97]
+        // We are defining hard boundaries for RSRP to be between [-200,0],
+        // and consider other values as error and call it 255
+        if (sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp) == 1) {
+            if (rsrp < -141 && rsrp >= -200) {
+                status->rsrp = 0;
+            }
+            else if (rsrp >= -44 && rsrp <=0) {
```
avtolstoy: Nit: formatting, same line.

hal/src/electron/modem/mdm_hal.cpp
technobly: So we ended up just trusting the first value in a multi-response.

hal/src/electron/modem/mdm_hal.cpp.orig
```diff
@@ -0,0 +1,3554 @@
+/*
```
avtolstoy: This file should not have been committed.

Issue #2032, opened 2020-02-18
USBSerial.read returns 0 when there are no characters


ZakCodes:
### Expected Behavior
When calling `USBSerial.read()` from the Particle Boron, it should return `-1` when there are no characters available.

### Observed Behavior
Instead of returning `-1` it returns `0`.

### Steps to Reproduce
System info:
* Firmware versions: 1.4.4, 1.5.0-rc.1
* Device: Particle Boron

Steps:
* Flash the test app to the Boron
* Connect to the Boron's USB serial port
* Open a Serial Monitor for reading and writing to the USB serial port
* Write a single character on the Serial Monitor (other than \<Enter\>)
* You should see the ASCII code for that character followed by a lot of zeros
* Type \<Enter\>
* You should see that the zeros have stopped

### Test App

```cpp
#include <Particle.h>

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(MANUAL);

void setup() {
    Serial.begin(9600);
}

void loop() {
    Serial.print(Serial.read());
    Serial.print(", ");
    delay(100);
}
```
This reads the characters from the serial port and writes them as integers back to the same serial port.

PR #2034, opened 2020-02-19
[hal] usb serial: return -1 instead of 0 when there's no data to read

ZakCodes:
### Problem
The `USBSerial::read` method returns 0 instead of returning an invalid character to signal that there are no characters to read. It makes it impossible to make the difference between the 0 character and there being no characters to read from the serial port. 

This causes problems with functions that expect this method to return -1 whenever there are no characters left to read. This is the case of `Serial::readStringUntil`.

### Solution
To solve the problem, I've added a check before reading a byte from the serial port to see whether there is one to read. If there isn't, it returns -1, otherwise the function works just like it did before.

### Steps to Test
System info:
* Firmware: built from this pull request
* Device: Particle Boron

Steps:
* Upload the following code to the device:
```c
#include <Particle.h>

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(MANUAL);

void setup() {
    Serial.begin(9600);
}

void loop() {
    Serial.print(Serial.read());
    Serial.print(", ");
    delay(100);
}
```
* Connect to the Boron's serial port using a serial port monitor
* Send characters to the Boron through its' Serial port
* When you aren't pressing any key, you should see -1 being printed
* When you are pressing a key, you should see the ASCII code of the character sent being printed

### Example App

```c
#include <Particle.h>

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(MANUAL);

void setup() {
    Serial.begin(9600);
}

void loop() {
    if (Serial.available() != 0) {
        String s = Serial.readStringUntil('\n');
        Serial.println(s);
    }
}
```

This app shows that `Serial.readStringUntil` can now return strings with more than a single character, because it isn't inserting a bunch of `0` everywhere in the string when there are no characters to read.

### References
Here's the issue that this PR resolves: #2032

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/usb_hal.cpp
+++ b/hal/src/nRF52840/usb_hal.cpp
@@ -81,6 +81,10 @@ int32_t HAL_USB_USART_Available_Data_For_Write(HAL_USB_USART_Serial serial) {
 }
 
 int32_t HAL_USB_USART_Receive_Data(HAL_USB_USART_Serial serial, uint8_t peek) {
+    if (usb_uart_available_rx_data() == 0) {
+        return -1;
+    }
+
     if (peek) {
         return usb_uart_peek_rx_data(0);
     } else {


avtolstoy:
@ZakCodes Thanks for the PR! We'll include this fix in the upcoming 1.5.0-rc.2.

ZakCodes:
You're welcome!

PR #2037, opened 2020-02-21
Workaround for failing Device OS builds on Windows

sergeuz:
### Problem

Workbench fails to build Device OS on Windows due to the path length limit on that platform.

### Solution

Move the OpenThread's radio driver sources closer to the submodule's root directory :joy:


avtolstoy:
@sergeuz Could you please commit this OpenThread changes into `release/20190130-particle` branch.

PR #2038, opened 2020-02-26
Adding changes for R410M to use UCGED instead of CESQ

keeramis:
<details>

```

```
</details>

### Problem

R410M uses CESQ to collect data, which is an unstable AT command to collect signal strength values.

### Solution

Replace CESQ with UCGED (mode 5) only on R410M. UCGED gives RSRP/RSRQ in dBm/dB and CESQ gives in index values. RSRP/RSRQ from UCGED are converted from dBm/dB to CESQ-type index values and further processed.

### Steps to Test

Manual testing required here

1. Check that correct signal strength and quality are reported by comparing the values reported by the previous Device OS and the values reported by your code for the same device in the same location.
2. Check an Boron 2/3G
3. Check the null case (no coverage)
4. Drive test

### Example App

```c
#include "Particle.h"

void setup() {
    Particle.publishVitals(30);
}

void loop() {
}
```

### References

---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -506,52 +506,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
     CHECK(checkParser());
     CHECK(queryAndParseAtCops(qual));
 
+    // Min and max RSRQ index values multiplied by 100
+    // Min: -19.5 and max: -3
+    const int min_rsrq_mul_by_100 = -1950;
+    const int max_rsrq_mul_by_100 = -300;
+
     if (ncpId() == PLATFORM_NCP_SARA_R410) {
-        int rxlev, rxqual, rscp, ecn0, rsrq, rsrp;
-        auto resp = parser_.sendCommand("AT+CESQ");
-        int r = CHECK_PARSER(resp.scanf("+CESQ: %d,%d,%d,%d,%d,%d", &rxlev, &rxqual,
-                &rscp, &ecn0, &rsrq, &rsrp));
-        CHECK_TRUE(r == 6, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
-        r = CHECK_PARSER(resp.readResult());
+        int rsrp;
+        int rsrq_n;
+        unsigned long rsrq_f;
+
+        // Default to 255 in case RSRP/Q are not found
+        qual->strength(255);
+        qual->quality(255);
+
+        // Set UCGED to mode 5 for RSRP/RSRQ values on R410M
+        CHECK_PARSER_OK(parser_.execCommand("AT+UCGED=5"));
+        auto resp = parser_.sendCommand("AT+UCGED?");
+
+        int val;
+        unsigned long val2;
+        while (resp.hasNextLine()) {
+            char type = 0;
+            const int r = CHECK_PARSER(resp.scanf("+RSR%c: %*d,%*d,\"%d.%lu\"", &type, &val, &val2));
+            if (r >= 2) {
+                if (type == 'P') {
+                    rsrp = val;
+                    if (rsrp < -141 && rsrp >= -200) {
+                        qual->strength(0);
+                    } else if (rsrp >= -44 && rsrp <=0) {
+                        qual->strength(97);
+                    } else if (rsrp >= -141 && rsrp < -44) {
+                        qual->strength(rsrp + 140);
+                    } else {
+                        // If RSRP is not in the expected range
+                        qual->strength(255);
+                    }
+                } else if (type == 'Q' && r == 3) {
+                    rsrq_n = val;
+                    rsrq_f = val2;
+                    int rsrq_mul_100 = rsrq_n * 100 - rsrq_f;
+                    if (rsrq_mul_100 < min_rsrq_mul_by_100 && rsrq_mul_100 >= -2000) {
+                        qual->quality(0);
+                    } else if (rsrq_mul_100 >= max_rsrq_mul_by_100 && rsrq_mul_100 <=0) {
+                        qual->quality(34);
+                    } else if (rsrq_mul_100 >= min_rsrq_mul_by_100 && rsrq_mul_100 < max_rsrq_mul_by_100) {
+                        qual->quality((rsrq_mul_100 + 2000)/50);
+                    } else {
+                        // If RSRQ is not in the expected range
+                        qual->quality(255);
+                    }
+                }
+            }
+        }
+
+        const int r = CHECK_PARSER(resp.readResult());
         CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
 
-        switch (qual->strengthUnits()) {
-            case CellularStrengthUnits::RXLEV: {
-                qual->strength(rxlev);
-                break;
-            }
-            case CellularStrengthUnits::RSCP: {
-                qual->strength(rscp);
-                break;
-            }
-            case CellularStrengthUnits::RSRP: {
-                qual->strength(rsrp);
-                break;
-            }
-            default: {
-                // Do nothing
-                break;
-            }
-        }
-
-        switch (qual->qualityUnits()) {
-            case CellularQualityUnits::RXQUAL: {
-                qual->quality(rxqual);
-                break;
-            }
-            case CellularQualityUnits::ECN0: {
-                qual->quality(ecn0);
-                break;
-            }
-            case CellularQualityUnits::RSRQ: {
-                qual->quality(rsrq);
-                break;
-            }
-            default: {
-                // Do nothing
-                break;
-            }
-        }
     } else {
         int rxlev, rxqual;
         auto resp = parser_.sendCommand("AT+CSQ");
--- a/user/tests/wiring/no_fixture/cellular.cpp
+++ b/user/tests/wiring/no_fixture/cellular.cpp
@@ -157,6 +157,65 @@ test(CELLULAR_07_rssi_is_valid) {
     assertMoreOrEqual(s.rssi, -150);
 }
 
+// Collects cellular signal strength and quality and validates Accesstechnology (RAT)
+test(CELLULAR_08_sigstr_is_valid) {
+    connect_to_cloud(6*60*1000);
+    assertTrue(Particle.connected());
+    CellularSignal s;
+    bool values_in_range = false;
+    const int num_retries = 10;
+    // Checks 10 times with 500ms gap to get valid signal strength
+    for (int x = 0; x < num_retries; x++) {
+        s = Cellular.RSSI();
+        // Verify that strength and quality values are in range for the given AccessTechnology
+        switch (s.getAccessTechnology()) {
+            case NET_ACCESS_TECHNOLOGY_GSM:     // GSM strength [-111, -48] and quality [-3.70, -0.60]
+                if ((s.getStrengthValue() <= -48.0f && s.getStrengthValue() >= -111.0f)
+                    && (s.getQualityValue() <= -0.6f && s.getQualityValue() >= -3.7f)) {
+                        values_in_range = true;
+                        x = num_retries;
+                        break;
+                }
+                Serial.printlnf("StrV: %.2f, QualV: %.2f, RAT: %d", s.getStrengthValue(), s.getQualityValue(), s.getAccessTechnology());
+                break;
+            case NET_ACCESS_TECHNOLOGY_EDGE:     // EDGE strength [-111, -48] and quality [-3.70, -0.60]
+                if ((s.getStrengthValue() <= -48.0f && s.getStrengthValue() >= -111.0f)
+                    && (s.getQualityValue() <= -0.6f && s.getQualityValue() >= -3.7f)) {
+                        values_in_range = true;
+                        x = num_retries;
+                        break;
+                }
+                Serial.printlnf("StrV: %.2f, QualV: %.2f, RAT: %d", s.getStrengthValue(), s.getQualityValue(), s.getAccessTechnology());
+                break;
+            case NET_ACCESS_TECHNOLOGY_UTRAN:     // UTRAN strength [-121, -25] and quality [-24.5, 0]
+                if ((s.getStrengthValue() <= -25.0f && s.getStrengthValue() >= -121.0f)
+                    && (s.getQualityValue() <= -0.0f && s.getQualityValue() >= -25.4f)) {
+                        values_in_range = true;
+                        x = num_retries;
+                        break;
+                }
+                Serial.printlnf("StrV: %.2f, QualV: %.2f, RAT: %d", s.getStrengthValue(), s.getQualityValue(), s.getAccessTechnology());
+                break;
+            case NET_ACCESS_TECHNOLOGY_LTE:
+            case NET_ACCESS_TECHNOLOGY_LTE_CAT_M1:  // LTE CAT-M1 strength [-141, -44] and quality [-20, -3]
+            case NET_ACCESS_TECHNOLOGY_LTE_CAT_NB1:
+                if ((s.getStrengthValue() <= -44.0f && s.getStrengthValue() >= -141.0f)
+                    && (s.getQualityValue() <= -3.0f && s.getQualityValue() >= -20.0f)) {
+                        values_in_range = true;
+                        x = num_retries;
+                        break;
+                }
+                Serial.printlnf("StrV: %.2f, QualV: %.2f, RAT: %d", s.getStrengthValue(), s.getQualityValue(), s.getAccessTechnology());
+                break;
+            default:
+                break;
+        }
+        delay(500);
+    }
+    assertFalse((s.getAccessTechnology() == NET_ACCESS_TECHNOLOGY_UNKNOWN) || (s.getAccessTechnology() == NET_ACCESS_TECHNOLOGY_NONE));
+    assertTrue(values_in_range);
+}
+
 test(MDM_01_socket_writes_with_length_more_than_1023_work_correctly) {
 
 #if HAL_PLATFORM_NCP


hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
             }
-            default: {
-                // Do nothing
-                break;
+            if (strstr(buf, "+RSRQ:")) {
+                r = sscanf(buf, "\r\n+RSRQ: %*d,%*d,\"%d.%lu\"",&rsrq_n, &rsrq_f);
+                CHECK_TRUE(r == 2, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
             }
+            buf[0] = '\0';
         }
```
technobly: So one thing that differs logically from the new code and old code, is that there is no check to make sure both RSRP and RSRQ are found.  Only IF they are found with `strstr` is there a CHECK_TRUE for a result.  So in theory RSRP OR RSRQ OR NEITHER could be found and the rest of the code will try to process the "found" values.  Previously if either was not found, the code would error out.  So we need some code here to keep this operation atomic and sane.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+            qual->strength(rsrp + 140);
+        }
+
+        int rsrq_mul_100 = rsrq_n * 100 - rsrq_f;
+        if (rsrq_mul_100 < min_rsrq_mul_by_100) {
+            qual->quality(0);
+        }
+        else if (rsrq_mul_100 >= max_rsrq_mul_by_100) {
+            qual->quality(34);
+        }
```
technobly: Something about bounding the RSRQ value to 0 and 34 is bugging me, because we're supposed to be trusting UCGED more than CSQ and CESQ (which we know can often return 0, 1 and 34 when they shouldn't).  If UCGED happens to return a value outside of that which we expect, how about we return an error... and not a valid value that might be wrong?  If it returns a value outside of this range, it might be due to some other reason like noise or a parsing issue.

The same logic applies to RSRP above with 0 and 97.

We should take a look at Gen 2 with this in mind as well.

keeramis: @technobly 
The problem with UCGED command is that there is no good way to find the value that we **don't expect**. As per the manual and 3GPP specs, everything -inf to +inf can be mapped from 0-34. So we just don't know which values to not expect. Does that sound about right? Please see this pic. Shall we define a range of values to _expect_?

image

keeramis: According to my understanding - 
UCGED command just gives the right values when it can read the signal strength. 
When it cannot read the signal strength, it just ERRORs out using an `ERROR` or `CME ERROR`. If it didn't error, then we trust the value that it gives us.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+        else {
+            qual->quality((rsrq_mul_100 + 2000)/50);
+        }
+        const int r = CHECK_PARSER(resp.readResult());
```
keeramis: If one of the RSRP/RSRQ is not valid, let's return an ERROR.

Or check up above in scanf. 
**If !(r1 && r2) is one idea to invalidate this.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+        qual->strength(255);
+        qual->quality(255);
+
+        auto resp = parser_.sendCommand("AT+UCGED?");
+        char buf[1024] = {};     // keep this string long in case there are multiple tower infos in the string
+        while (resp.hasNextLine()) {
+            CHECK_PARSER(resp.readLine(buf, sizeof(buf)));
+            int r = 0;
+            if (strstr(buf, "+RSRP:")) {    // check for RSRP response
+                r = sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp);
```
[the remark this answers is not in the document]
avtolstoy: `sscanf` will anyway ignore `\r\n`: https://godbolt.org/z/2dmJNP, but I agree, `\r\n` should be removed here.

It's probably also not really necessary to allocate 1024 byte buffer here and let the at parser handle memory allocations etc automagically, but that's just a preference:
```c++
while (resp.hasNextLine()) {
    char type = 0;
    const int r = CHECK_PARSER(resp.scanf("+RSR%c: %*d,%*d,\"%d.%lu\"", &type, &val, &val2));
    if (r >= 2) {
        if (type == 'P') {
            rsrp = val;
        } else if (type == 'Q' && r == 3) {
            rsrq_n = val;
            rsrq_f = val2;
    }
}
```

Another potential way is to:

```
const CString str = CHECK_PARSER(resp.readLine());
...
```

Also Nit: formatting.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -993,6 +1003,7 @@ int SaraNcpClient::configureApn(const CellularNetworkConfig& conf) {
 
 int SaraNcpClient::setRegistrationTimeout(unsigned timeout) {
     registrationTimeout_ = std::max(timeout, REGISTRATION_TIMEOUT);
+    LOG(INFO," Regtimeset: %u",registrationTimeout_ );
```
avtolstoy: Nit: formatting.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+            int r = 0;
+            if (strstr(buf, "+RSRP:")) {    // check for RSRP response
+                r = sscanf(buf, "\r\n+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp);
+                CHECK_TRUE(r == 1, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
             }
-            default: {
-                // Do nothing
-                break;
+            if (strstr(buf, "+RSRQ:")) {
+                r = sscanf(buf, "\r\n+RSRQ: %*d,%*d,\"%d.%lu\"",&rsrq_n, &rsrq_f);
```
avtolstoy: Nit: formatting.

hal/src/boron/network/sara_ncp_client.cpp
keeramis: Added code to check RSRP/Q are found before collecting results. Please see : 
https://github.com/particle-iot/device-os/blob/daf4878627829e849bf9c70e5ac46b25060f6a7a/hal/src/boron/network/sara_ncp_client.cpp#L525

I am parsing RSRP/Q separately so that even if one of the values is found, we will collect that. For the value that is not found it is defaulted to 255 in the beginning of UCGED command parsing. Please see here: https://github.com/particle-iot/device-os/blob/daf4878627829e849bf9c70e5ac46b25060f6a7a/hal/src/boron/network/sara_ncp_client.cpp#L512

hal/src/boron/network/sara_ncp_client.cpp
keeramis: ^ Not doing this because if one of the RSRP/Q values is not valid, I am still collecting the other one. This coud be a useful data point to have to check why one is valid and other is not.

hal/src/boron/network/sara_ncp_client.cpp
keeramis: This line should not even be a part of code review. I will try to get rid of this, but will change formatting on others.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,66 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+        CHECK_PARSER_OK(parser_.execCommand("AT+UCGED=5"));
+        auto resp = parser_.sendCommand("AT+UCGED?");
+        int r1 = 0;
+        int r2 = 0;
+        while (resp.hasNextLine()) {
+            r1 = CHECK_PARSER(resp.scanf("+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp));
+            r2 = CHECK_PARSER(resp.scanf("+RSRQ: %*d,%*d,\"%d.%lu\"",&rsrq_n, &rsrq_f));
+        }
 
-        switch (qual->strengthUnits()) {
```
avtolstoy: I'd propose we leave something similar here to check that RAT acquired in `queryAndParseAtCops()` is in fact LTE.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,66 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+        qual->strength(255);
+        qual->quality(255);
+
+        // Set UCGED to mode 5 for RSRP/RSRQ values on R410M
+        CHECK_PARSER_OK(parser_.execCommand("AT+UCGED=5"));
+        auto resp = parser_.sendCommand("AT+UCGED?");
+        int r1 = 0;
+        int r2 = 0;
+        while (resp.hasNextLine()) {
+            r1 = CHECK_PARSER(resp.scanf("+RSRP: %*d,%*d,\"%d.%*u\"",&rsrp));
```
avtolstoy: Nit: formatting.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,66 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+            else {
+                // If RSRQ is not in the expected range
+                qual->quality(255);
             }
         }
-    } else {
+        const int r = CHECK_PARSER(resp.readResult());
+        CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
+    }
+    else {
```
avtolstoy: Nit: formatting.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,53 +498,66 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
-            case CellularStrengthUnits::RSCP: {
-                qual->strength(rscp);
-                break;
+            else if (rsrp >= -44 && rsrp <=0) {
+                qual->strength(97);
             }
-            case CellularStrengthUnits::RSRP: {
-                qual->strength(rsrp);
-                break;
+            else if (rsrp >= -141 && rsrp < -44) {
```
avtolstoy: Nit: formatting. Follow the style of the rest of the file.

hal/src/boron/network/sara_ncp_client.cpp
keeramis: I removed that because R410M supports only LTE. TMK either it acquires LTE or it doesn't. Am I missing something?

keeramis: I can leave the switch case as-in and put in this additional stuff.

avtolstoy: The upper layer code expects RAT to be available. So maybe it's fine to leave as-is. Let's leave as-is.

keeramis: OK.. 
You mean I need to add the switch case or should we not add the switch case and leave it as-it-is?
The one thing I am OCD about is if ublox supports multiple RATs on this R410M, or if we start supporting NB-IoT on this that will be a different RAT. If the upper layer is still taking care of this, it's probably OK to not add the switch case?

avtolstoy: I think it is fine to leave things as-is. Even if RAT is not set properly by COPS, the whole thing will be ignore by upper layer code.

keeramis: For future reference - Even if it's NB-IoT, it will still be RSRP/Q params for signal strength

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,6 +157,56 @@ test(CELLULAR_07_rssi_is_valid) {
...
+
+    assertNotEqual(rat, NET_ACCESS_TECHNOLOGY_UNKNOWN || NET_ACCESS_TECHNOLOGY_NONE);
+
+    // Verify strength and quality values
+    switch (rat) {
+        case NET_ACCESS_TECHNOLOGY_GSM:     // GSM strength [-111, -48] and quality [?, ?]
+            assertLessOrEqual(s.getStrengthValue(), -48);
+            assertMoreOrEqual(s.getStrengthValue(), -111);
+            assertLessOrEqual(s.getQualityValue(), 0);
+            assertMoreOrEqual(s.getQualityValue(), -20);
```
keeramis: Wanting to confirm the limits on GSM and EDGE might not be correct. Is the correct qualityvalue range between [0,7]?

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,6 +157,56 @@ test(CELLULAR_07_rssi_is_valid) {
     assertMoreOrEqual(s.rssi, -150);
 }
 
+test(CELLULAR_08_sigstr_is_valid) {     // collects signal strength and quality
```
keeramis: Tested this on Boron LTE, and seems reflective of the changes. I used certain values and debug statements to verify we are seeing some values. Also, interestingly, the RAT is `6 (NET_ACCESS_TECHNOLOGY_LTE)`

```
Test CELLULAR_07_rssi_is_valid passed.
StrV: -95.00
QualV: -15.50
RAT: 6
Test CELLULAR_08_sigstr_is_valid passed.
Test CLOUD_01_Particle_Connect_Does_Not_Block_In_SemiAutomatic_Mode passed.
```

keeramis: Probably a separate topic - but I think the RAT just reads 6 for M1? I printed `sigext rat` below in the log.
```
0000060224 [ncp.at] TRACE: > AT+COPS?
0000060272 [ncp.at] TRACE: < +COPS: 0,2,"310410",7
0000060275 [ncp.at] TRACE: < OK
0000060276 [ncp.at] TRACE: > AT+UCGED=5
0000060322 [ncp.at] TRACE: < OK
0000060323 [ncp.at] TRACE: > AT+UCGED?
0000060372 [ncp.at] TRACE: < +RSRP: 314,5110,"-094.00",
0000060375 [ncp.at] TRACE: < +RSRQ: 314,5110,"-13.20",
0000060377 [ncp.at] TRACE: < OK
0000060378 [wiring] INFO: sigext rat: 6
```

avtolstoy: This is weird, because the AT command reference states:
```
SARA-R410M-02B•   <AcT>=8 is used for LTE Cat M1 on SARA-R410M-02B-00.
```

Here we can see AcT=7 for some reason.

avtolstoy: Ah, but it also says:
```
SARA-R410M-02B / SARA-R410M-52B - 7, 9
```

avtolstoy: For SARA R4 we should set `NET_ACCESS_TECHNOLOGY_LTE_CAT_M1` if AcT is 7 or 8, I think that's the solution and `NET_ACCESS_TECHNOLOGY_LTE_CAT_NB1` if it's 9.

keeramis: Sounds about right.

keeramis: This is probably where we need to make this change - https://github.com/particle-iot/device-os/blob/f6a0a5970c90284c07e078b9894ca0b8566f0fad/hal/network/ncp/cellular/cellular_hal.cpp#L85

avtolstoy: I would actually patch it here depending on the modem type: https://github.com/particle-iot/device-os/blob/develop/hal/src/boron/network/sara_ncp_client.cpp#L471 

And for Gen 2:
https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L1718
https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L496

Again depending on the modem type.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,52 +498,59 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
+        qual->strength(255);
+        qual->quality(255);
+
+        // Set UCGED to mode 5 for RSRP/RSRQ values on R410M
+        CHECK_PARSER_OK(parser_.execCommand("AT+UCGED=5"));
+        auto resp = parser_.sendCommand("AT+UCGED?");
+        int r1 = 0;
+        int r2 = 0;
+        while (resp.hasNextLine()) {
+            r1 = CHECK_PARSER(resp.scanf("+RSRP: %*d,%*d,\"%d.%*u\"", &rsrp));
```
avtolstoy: This will not work correctly, because underneath `resp.scanf()` will call `readLine()`. So, these two calls will read two lines, whereas the intended behavior is supposed to be to try to parse either RSRP or RSRQ for the line. `readLine()` + `sscanf()` should be used or a single `resp.scanf()` call that can parse both RSRP/RSRQ (I gave an example earlier).

user/tests/wiring/no_fixture/cellular.cpp
avtolstoy: https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp/cellular/cellular_hal.cpp#L395

GSM: 3GPP TS 45.008 8.2.4 [0.14%, 18.10%]

EDGE: 3GPP TS 45.008 10.2.3.3 [-0.60, -3.70]

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,6 +157,56 @@ test(CELLULAR_07_rssi_is_valid) {
...
+    for (int x = 0; x < 10; x++) {
+        s = Cellular.RSSI();
+        if (s.getStrengthValue() <= 0 && s.getQualityValue() <= 0) {
+            break;
+        }
+        delay(5000);
+    }
+
+    // Verify RAT
+    uint8_t rat = s.getAccessTechnology();
```
avtolstoy: `auto` or `hal_net_access_tech_t`.

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,6 +157,56 @@ test(CELLULAR_07_rssi_is_valid) {
...
 }
 
+test(CELLULAR_08_sigstr_is_valid) {     // collects signal strength and quality
+    connect_to_cloud(6*60*1000);
+    assertTrue(Particle.connected());
+    CellularSignal s;
+    for (int x = 0; x < 10; x++) {
+        s = Cellular.RSSI();
+        if (s.getStrengthValue() <= 0 && s.getQualityValue() <= 0) {
+            break;
```
avtolstoy: I would actually perform the check here multiple times.

user/tests/wiring/no_fixture/cellular.cpp
avtolstoy: This can be done as a separate PR.

user/tests/wiring/no_fixture/cellular.cpp
keeramis: Changed! (I saw in one of the declarations in the deeper layers, it was a `uint8_t` so thought that might not hurt.)

user/tests/wiring/no_fixture/cellular.cpp
keeramis: Added!

user/tests/wiring/no_fixture/cellular.cpp
keeramis: Changed!

user/tests/wiring/no_fixture/cellular.cpp
keeramis: Coolio!

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,54 +157,48 @@ test(CELLULAR_07_rssi_is_valid) {
...
+            case NET_ACCESS_TECHNOLOGY_LTE:
+            case NET_ACCESS_TECHNOLOGY_LTE_CAT_M1:  // LTE CAT-M1 strength [-141, -44] and quality [-20, -3]
+            case NET_ACCESS_TECHNOLOGY_LTE_CAT_NB1:
+                assertLessOrEqual(s.getStrengthValue(), -44.0f);
+                assertMoreOrEqual(s.getStrengthValue(), -141.0f);
+                assertLessOrEqual(s.getQualityValue(), -3.0f);
+                assertMoreOrEqual(s.getQualityValue(), -20.0f);
+                break;
+            default:
+                assertFalse((s.getAccessTechnology() == NET_ACCESS_TECHNOLOGY_UNKNOWN) \
```
avtolstoy: `\` is unnecessary.

avtolstoy:
@keeramis Please rebase and feel free to merge.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -498,52 +498,63 @@ int SaraNcpClient::getSignalQuality(CellularSignalQuality* qual) {
...
-                break;
+        // Set UCGED to mode 5 for RSRP/RSRQ values on R410M
+        CHECK_PARSER_OK(parser_.execCommand("AT+UCGED=5"));
+        auto resp = parser_.sendCommand("AT+UCGED?");
+
+        int val;
+        unsigned long val2;
+        while (resp.hasNextLine()) {
+            char type = 0;
+            const int r = CHECK_PARSER(resp.scanf("+RSR%c: %*d,%*d,\"%d.%lu\"", &type, &val, &val2));
```
technobly: Nice change here!

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -157,6 +157,50 @@ test(CELLULAR_07_rssi_is_valid) {
...
+                        || (s.getAccessTechnology() == NET_ACCESS_TECHNOLOGY_NONE));
+                break;
+        }
+        delay(5000);
```
technobly: Can we speed this up?  Maybe 500ms delay?  Also this one seems susceptible to start up issues.  Hopefully the previous test 07 delays enough that we have valid RSSI values.  The other RSSI test delays 5s between tries because it exits the test on the first valid response.  We could do a similar thing here (x = 10; break;).  Ideally since there are 130 - 150 individual tests for all platforms, we want to make each one as fast as possible to keep the testing process quick.

PR #2039, opened 2020-02-26
Redo: Fixes the issue that SPI default SS pin is not configured

XuGuohui:
### Problem

`SPI.begin(SPI_MODE_MASTER);` doesn't configure the default SS pin to be output high initially.

### Solution

Assign default parameter for the second argument of the `SPIClass::begin(SPI_Mode mode, uint16_t ss_pin = SPI_DEFAULT_SS)`.

### Steps to Test

1. Build and download the following example App.
2. Check if the default SS pin according to https://docs.particle.io/reference/device-os/firmware/electron/#begin--4 can be toggled periodically.

### Example App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);

#if HAL_PLATFORM_STM32F2XX
#define SPI_DEFAULT_SS_PIN  A2
#elif PLATFORM_ID == PLATFORM_XENON || PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON
#define SPI_DEFAULT_SS_PIN  A5
#endif

void setup()
{
    SPI.begin(SPI_MODE_MASTER);
}

void loop()
{
    digitalWrite(SPI_DEFAULT_SS_PIN, LOW);
    delay(3000);
    digitalWrite(SPI_DEFAULT_SS_PIN, HIGH);
    delay(3000);
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/pinmap_hal.h
+++ b/hal/inc/pinmap_hal.h
@@ -63,6 +63,8 @@ void HAL_Set_Pin_Function(pin_t pin, PinFunction pin_func);
 
 #include "pinmap_impl.h"
 
+#define HAL_Pin_Is_Valid(pin) (pin < TOTAL_PINS)
+
 #ifdef __cplusplus
 }
 #endif
--- a/hal/inc/pinmap_hal.h
+++ b/hal/inc/pinmap_hal.h
@@ -63,7 +63,7 @@ void HAL_Set_Pin_Function(pin_t pin, PinFunction pin_func);
 
 #include "pinmap_impl.h"
 
-#define HAL_Pin_Is_Valid(pin) (pin < TOTAL_PINS)
+#define HAL_Pin_Is_Valid(pin) ((pin) < TOTAL_PINS)
 
 #ifdef __cplusplus
 }
--- a/hal/src/stm32f2xx/spi_hal.c
+++ b/hal/src/stm32f2xx/spi_hal.c
@@ -293,7 +293,7 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
         pin = spiMap[spi].SPI_SS_Pin;
     }
 
-    if (pin >= TOTAL_PINS) {
+    if (mode == SPI_MODE_SLAVE && !HAL_Pin_Is_Valid(pin)) {
         return;
     }
 
@@ -315,7 +315,7 @@ void HAL_SPI_Begin_Ext(HAL_SPI_Interface spi, SPI_Mode mode, uint16_t pin, void*
     HAL_Pin_Mode(spiMap[spi].SPI_MISO_Pin, AF_OUTPUT_PUSHPULL);
     HAL_Pin_Mode(spiMap[spi].SPI_MOSI_Pin, AF_OUTPUT_PUSHPULL);
 
-    if (mode == SPI_MODE_MASTER)
+    if (mode == SPI_MODE_MASTER && HAL_Pin_Is_Valid(pin))
     {
         // Ensure that there is no glitch on SS pin
         PIN_MAP[pin].gpio_peripheral->BSRRL = PIN_MAP[pin].gpio_pin;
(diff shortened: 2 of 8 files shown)


hal/inc/pinmap_hal.h
```diff
@@ -63,6 +63,8 @@ void HAL_Set_Pin_Function(pin_t pin, PinFunction pin_func);
 
 #include "pinmap_impl.h"
 
+#define HAL_Pin_Is_Valid(pin) (pin < TOTAL_PINS)
```
avtolstoy: `((pin) < TOTAL_PINS)` just in case.

user/tests/wiring/no_fixture_spi/spi.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+    hal_spi_info_t info = {};
+
+    SPI.begin(SPI_DEFAULT_SS);
+    querySpiInfo(HAL_SPI_INTERFACE1, &info);
+    assertTrue(info.enabled == true);
+    assertTrue(info.mode == SPI_MODE_MASTER);
+#if PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON || PLATFORM_ID == PLATFORM_XENON
+    assertTrue(info.ss_pin == D14);
+#elif PLATFORM_ID == PLATFORM_BSOM
```
avtolstoy: B5 SoM?

user/tests/wiring/no_fixture_spi/spi.cpp
```diff
@@ -0,0 +1,399 @@
...
+test(SPI_2_SPI_Begin_With_Ss_Pin)
+{
+    // Just in case
+    SPI.end();
+
+    hal_spi_info_t info = {};
+
+    SPI.begin(SPI_DEFAULT_SS);
+    querySpiInfo(HAL_SPI_INTERFACE1, &info);
+    assertTrue(info.enabled == true);
```
avtolstoy: Prefer `assertEqual`

hal/src/stm32f2xx/spi_hal.c
```diff
@@ -644,6 +649,9 @@ void HAL_SPI_Info(HAL_SPI_Interface spi, hal_spi_info_t* info, void* reserved)
         info->mode = spiState[spi].mode;
         info->bit_order = spiState[spi].SPI_InitStructure.SPI_FirstBit == SPI_FirstBit_MSB ? MSBFIRST : LSBFIRST;
         info->data_mode = spiState[spi].SPI_InitStructure.SPI_CPOL | spiState[spi].SPI_InitStructure.SPI_CPHA;
+        if (info->version >= HAL_SPI_INFO_VERSION_2) {
+            info->ss_pin = spiState[spi].SPI_SS_Pin;
```
avtolstoy: Any implications on `spi_ensure_configured()` from wiznetif.cpp?

XuGuohui: Did you mean that the SS pin might be temporarily modified by wiznetif.cpp, so that we get the wrong pin?

avtolstoy: I think we are going to now always be calling `SPI.begin()` due to this change:
https://github.com/particle-iot/device-os/blob/develop/hal/network/lwip/wiznet/wiznetif.cpp#L120

XuGuohui: As we discussed, we should rework the wiznetif.cpp to store the SPI configuration before begin() and restore before end() to make sure the system firmware doesn't affect user's SPI configuration. But does this change here has any potential defect? This function gets the current SPI info and it does what we want.

avtolstoy: We are already doing that, if you take a look at `reg_wizchip_cris_cbfunc()` call. It should work fine, as I was looking at the stm32f2xx change. Resolving this.

PR #2042, opened 2020-03-03
Enable hwfc for Quectel modem during runtime and add support for EG91-EX

YutingYou:
### Problem

The pre-production B5SoM boards don't support hardware flow control. To distinguish them from the new b5som boards, we're going to automatically enable hwfc by detecting the hardware version. 

This PR adds two features:
1. Enable hwfc for Quectel modem during runtime
2. Add support for EG91-EX

### Steps to Test

We don't have the new B5SoM boards on hand, provide a rough test app.

### Example App

```
#include "application.h"
#include "deviceid_hal.h"
#include "exflash_hal.h"

SYSTEM_MODE(MANUAL);

Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL, {
    {"app", LOG_LEVEL_ALL},
    {"ncp", LOG_LEVEL_ALL},
    {"hal.ble", LOG_LEVEL_WARN},
    {"lwip", LOG_LEVEL_WARN},
    {"ot", LOG_LEVEL_WARN},
    {"sys", LOG_LEVEL_WARN},
    {"system", LOG_LEVEL_WARN}}
);

void setup(void) {
}

void loop() {
    uint8_t hw_data[4] = {};
    int r = hal_exflash_read_special(HAL_EXFLASH_SPECIAL_SECTOR_OTP, 32, hw_data, sizeof(hw_data));
    if (!r) {
        Log.info("OTP area dump: NCP_ID: %x HW_VER: %x Feature_Flags: %x %x", hw_data[0], hw_data[1], hw_data[2], hw_data[3]);
    } else {
        Log.error("r: %d", r);
    }

    uint32_t hw_version;
    if (hal_get_device_hw_version(&hw_version, nullptr) == SYSTEM_ERROR_NONE) {
        Log.info("hw version: %d", hw_version);
        if (hw_version== hw_data[1]) {
            Log.info("TEST PASSED!");
        } else {
            Log.info("TEST FAILED!");
        }
    } else {
        Log.info("TEST FAILED!");
    }

    delay(10000);
}
```

### References

[CH47749]

---

### Completeness

- [ ] Added documentation


hal/inc/deviceid_hal.h
```diff
@@ -83,6 +83,11 @@ int hal_get_device_serial_number(char* str, size_t size, void* reserved);
 // TODO: Move this function to an appropriate module
 int hal_get_device_secret(char* data, size_t size, void* reserved);
 
+/**
+ * Get the device's hardware version.
+ */
+int hal_get_device_hw_version();
```
avtolstoy: Let's instead use `int hal_get_device_hw_revision(uint32_t* revision, void* reserved)` and return `system_error_t` instead.

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -84,7 +85,7 @@ inline system_tick_t millis() {
 }
 
 const auto QUECTEL_NCP_DEFAULT_SERIAL_BAUDRATE = 115200;
-const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 115200;
+const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 460800;
```
avtolstoy: Let's not make this change for now.

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -114,7 +115,15 @@ int QuectelNcpClient::init(const NcpClientConfig& conf) {
     conf_ = static_cast<const CellularNcpClientConfig&>(conf);
 
     // Initialize serial stream
-    auto sconf = SERIAL_8N1 | SERIAL_FLOW_CONTROL_NONE; // TODO: replace with SERIAL_FLOW_CONTROL_RTS_CTS when HW is fixed.
+    auto sconf = SERIAL_8N1;
+
+    // Hardware version
+    // V003 - 0x00 (disable hwfc)
+    // V004 - 0x01 (enable hwfc)
+    if (hal_get_device_hw_version() > 0x00) {
```
avtolstoy: Let's do the opposite: if the revision is specifically V003 - disable hwfc, in all the other cases - enable.

hal/src/nRF52840/deviceid_hal.cpp
```diff
@@ -78,6 +79,19 @@ int hal_get_device_serial_number(char* str, size_t size, void* reserved)
...
 
+int hal_get_device_hw_version()
+{
+    // HW Data format: | NCP_ID | HW_VERSION | HW Feature Flags |
+    //                 | byte 0 |   byte 1   |    byte 2/3      |
+    uint8_t hw_data[4] = {};
+    int r = hal_exflash_read_special(HAL_EXFLASH_SPECIAL_SECTOR_OTP, HW_DATA_OTP_ADDRESS, hw_data, 4);
+
+    if (r != 0 || hw_data[1] == 0xFF) {
+        return -1;
```
avtolstoy: `system_error_t` codes should be used instead.

hal/src/b5som/network/quectel_ncp_client.cpp
YutingYou: Reverted

hal/src/b5som/network/quectel_ncp_client.cpp
YutingYou: Fixed

hal/src/nRF52840/deviceid_hal.cpp
YutingYou: Fixed

hal/inc/deviceid_hal.h
YutingYou: Fixed

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -685,21 +696,30 @@ int QuectelNcpClient::initReady() {
...
+    if (ncpId() == PLATFORM_NCP_QUECTEL_EG91_E || \
+        ncpId() == PLATFORM_NCP_QUECTEL_EG91_NA || \
+        ncpId() == PLATFORM_NCP_QUECTEL_EG91_EX) {
         CHECK_PARSER(parser_.execCommand("AT+QDSIM=0"));
     }
 
     // Enable flow control and change to runtime baudrate
-    // TODO: Uncomment when hardware is rev'd to fix CTS/RTS swap.
-    // CHECK_PARSER(parser_.execCommand("AT+IFC=2,2"));
-    CHECK(changeBaudRate(QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE));
```
avtolstoy: Do we need to remove this?

YutingYou: It should be 
```
    // Enable flow control and change to runtime baudrate
    auto runtimeBaudrate = QUECTEL_NCP_DEFAULT_SERIAL_BAUDRATE;
    uint32_t hwVersion = 0xFF;
    hal_get_device_hw_version(&hwVersion, nullptr);
    if (hwVersion == 0x00) {
        CHECK_PARSER(parser_.execCommand("AT+IFC=0,0"));
    } else {
        runtimeBaudrate = QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE;
        CHECK_PARSER(parser_.execCommand("AT+IFC=2,2"));
    }
    CHECK(changeBaudRate(runtimeBaudrate));
```
If we want to change baudrate to 460800, we can simply change `QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE`

avtolstoy: Ah, sorry, I've missed it. :+1:

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -114,7 +115,17 @@ int QuectelNcpClient::init(const NcpClientConfig& conf) {
...
     // Initialize serial stream
-    auto sconf = SERIAL_8N1 | SERIAL_FLOW_CONTROL_NONE; // TODO: replace with SERIAL_FLOW_CONTROL_RTS_CTS when HW is fixed.
+    auto sconf = SERIAL_8N1 | SERIAL_FLOW_CONTROL_RTS_CTS;
+
+    // Hardware version
+    // V003 - 0x00 (disable hwfc)
+    // V004 - 0x01 (enable hwfc)
+    uint32_t hwVersion = 0xFF;
+    hal_get_device_hw_version(&hwVersion, nullptr);
+    if (hwVersion == 0x00) {
```
avtolstoy: Let's make sure we check the return code from `hal_get_device_hw_version` as well. In `initReady()` as well.

avtolstoy: Also, the revisions ideally should be declared somewhere, instead of using magic codes.

YutingYou: What about adding this to `deviceid_hal.h`?
```
// Undefine hardware version
#define HAL_VERSION_UNDEFINED 0xFF
```

avtolstoy: That's fine, but I'd probably add B5 SoM hw revision v003 somewhere instead.

YutingYou: Since it only belongs to B5SoM, I'll add `#define HAL_VERSION_B5SOM_V003 0x00` to `quectel_ncp_client.cpp`, do you think so?

avtolstoy: That's ok. Prefer constants vs defines though.

YutingYou: Fixed

YutingYou: Fixed

hal/src/b5som/network/quectel_ncp_client.cpp
YutingYou: Fixed

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -117,12 +124,9 @@ int QuectelNcpClient::init(const NcpClientConfig& conf) {
...
     auto sconf = SERIAL_8N1 | SERIAL_FLOW_CONTROL_RTS_CTS;
 
-    // Hardware version
-    // V003 - 0x00 (disable hwfc)
-    // V004 - 0x01 (enable hwfc)
-    uint32_t hwVersion = 0xFF;
-    hal_get_device_hw_version(&hwVersion, nullptr);
-    if (hwVersion == 0x00) {
+    uint32_t hwVersion = HW_VERSION_UNDEFINED;
+    CHECK(hal_get_device_hw_version(&hwVersion, nullptr));
```
avtolstoy: `CHECK()` will return from function in case of failure, we should probably just resort to the default behavior.

YutingYou: That's why I didn't add `CHECK` but using a default value. But I agree with your idea that adding `CHECK` here, if we're using B5SoM_V003 and reading flash data failed for some reason, we shouldn't enable HW flow control.

YutingYou: Or we can combine these two conditions to not using HWFC

avtolstoy: We should, that is the point. The default behavior should be with flow control. Only if the revision is 100% the affected one we should enable it.

YutingYou: We can simply remove `CHECK()` here?

avtolstoy: Yes, but we do need to check the return code from `hal_get_device_hw_version()`. So the condition for disabling for flow control becomes:
`if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_B5SOM_V003)`

YutingYou: Fixed

PR #2045, opened 2020-03-09
OTA adjustments for Argon

sergeuz:
### Problem

1. Reduce the maximum number of firmware chunks that can be requested by an Argon during OTA from 40 to 20. Argon doesn't seem to be able to receive more than 25 chunks from a given flight of chunks, which causes it to drop `UpdateDone` messages repeatedly.

2. Do not wait for an ACK for a `ChunkMissed` message synchronously. This causes the comms library to drop packets and blocks the system loop during OTA.

3. Rework logging in the OTA code so that it's easier to debug.

### References

- [ch48671]

### DIFF
--- a/communication/inc/protocol.h
+++ b/communication/inc/protocol.h
@@ -232,11 +232,6 @@ protected:
 	}
 
 	/**
-	 * The number of missed chunks to send in a single flight.
-	 */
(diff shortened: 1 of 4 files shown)


avtolstoy:
@sergeuz Do we want to adjust number of chunks for other platforms as well?

PR #2049, opened 2020-03-15
Ensure NCP connection state is processed when changed by the muxer [ch48825]

technobly:
### Problem

The loss of signal can cause the modem to draw higher currents and if the power supply is not sufficient, the modem power supply will brown out and the NCP client will become unresponsive to keep alive pings.  `[gsm0710muxer] ERROR: The other end has not replied to keep alives (TESTs) 5 times, considering muxed connection dead` can be observed in the logs when this happens.  This could be immediately after signal is lost, or approx. 5.2 minutes afterwards.  This causes the AT command & Data muxer to go through a teardown, which in turn causes the NCP client to turn off the modem, but the PPP client doesn't know about this and keeps trying to retransmit a configure-request for IPCP.  The modem will attempt to be turned on, the retransmitted configure-request will fail, and the modem initialization will immediately fail again before it turns on.

### Solution

The root cause is that the NCP connection state is changed directly in the muxer callback function, but that does not get propagated to the NCP function that handles state changes, which would send an event to the PPP client to inform it that NCP is disconnected.  This would prevent the failure loop discussed above from occurring. 

The solution is to call `connectionState()` directly from the muxer callback to ensure we process events in the NCP client when NCP state is changed from the muxer callback.  Note: we should not do this with the CONNECTED state, since that will cause the connectionState() function to call back into the `muxer_` object.  Fortunately right now this is not a requirement in the muxer callback.

Another problem area was found as well that does not seem to attribute to the fix here, but it should be noted as changed.  `disable()` was being called before connectionState(NcpConnectionState::DISCONNECTED); below. I've swapped them and added a comment.  They were also swapped in the muxer callback.

Screen Shot 2020-03-12 at 9 40 37 PM

### Steps to Test

There are multiple ways to reproduce this issue, but given that it requires a power failure it can be tricky to reproduce.  

One way, is to run the test application, wait for the device to connect initially to the tower and Particle Cloud. Disconnect the battery and run on USB power only. Ensure publishes are occurring every 5 seconds, then disconnect the antenna and wait for the device to show a similar muxer teardown error as below.  After the error, the battery should be re-attached. The device should start into the failure loop as described, and reconnecting the antenna will not recover it.

Screen Shot 2020-03-12 at 2 46 29 PM

Another more complicated but more reliable way of reproducing the issue is to run only battery input only from a lab power supply (no USB) set to 3.6 ~ 3.8V and limit the supply current to 500mA.

### Example App

See example app referenced in ch48825, but a threaded app that publishes data every 5 seconds should be similar enough to reproduce this issue.

### References

ch48825

---

### Completeness

- [x] N/A Added documentation

### DIFF
--- a/hal/src/b5som/network/quectel_ncp_client.cpp
+++ b/hal/src/b5som/network/quectel_ncp_client.cpp
@@ -360,7 +360,20 @@ int QuectelNcpClient::updateFirmware(InputStream* file, size_t size) {
 }
 
 int QuectelNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
-    return muxer_.writeChannel(QUECTEL_NCP_PPP_CHANNEL, data, size);
+    int err = muxer_.writeChannel(QUECTEL_NCP_PPP_CHANNEL, data, size);
+
+    if (err) {
+        // Make sure we are going into an error state if muxer for some reason fails
+        // to write into the data channel.
+        connectionState(NcpConnectionState::DISCONNECTED);
+        // disable() sets the ncpState_ to NcpState::DISABLED which prevents connectionState()
+        // from even trying to send events to the PPP client to go to a LOWER_DOWN state,
+        // and it might still not get there if connState_ is already DISCONNECTED, but order matters.
+        // Call disable() after connectionState()
+        disable();
+    }
+
+    return err;
 }
 
 void QuectelNcpClient::processEvents() {
@@ -922,22 +935,25 @@ int QuectelNcpClient::muxChannelStateCb(uint8_t channel, decltype(muxer_)::Chann
     // This callback is executed from the multiplexer thread, not safe to use the lock here
     // because it might get called while blocked inside some muxer function
 
+    // Also please note that connectionState() should never be called with the CONNECTED state
+    // from this callback.
+
     // We are only interested in Closed state
     if (newState == decltype(muxer_)::ChannelState::Closed) {
         switch (channel) {
-        case 0: {
-            // Muxer stopped
-            self->disable();
-            self->connState_ = NcpConnectionState::DISCONNECTED;
-            break;
-        }
-        case QUECTEL_NCP_PPP_CHANNEL: {
-            // PPP channel closed
-            if (self->connState_ != NcpConnectionState::DISCONNECTED) {
-                self->connState_ = NcpConnectionState::CONNECTING;
+            case 0: {
+                // Muxer stopped
+                self->connectionState(NcpConnectionState::DISCONNECTED);
+                self->disable();
+                break;
             }
-            break;
-        }
+            case QUECTEL_NCP_PPP_CHANNEL: {
+                // PPP channel closed
+                if (self->connState_ != NcpConnectionState::DISCONNECTED) {
+                    self->connectionState(NcpConnectionState::CONNECTING);
+                }
+                break;
+            }
         }
     }
 
--- a/hal/src/b5som/network/quectel_ncp_client.cpp
+++ b/hal/src/b5som/network/quectel_ncp_client.cpp
@@ -365,11 +365,6 @@ int QuectelNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size)
     if (err) {
         // Make sure we are going into an error state if muxer for some reason fails
         // to write into the data channel.
-        connectionState(NcpConnectionState::DISCONNECTED);
-        // disable() sets the ncpState_ to NcpState::DISABLED which prevents connectionState()
-        // from even trying to send events to the PPP client to go to a LOWER_DOWN state,
-        // and it might still not get there if connState_ is already DISCONNECTED, but order matters.
-        // Call disable() after connectionState()
         disable();
     }
 
@@ -943,13 +938,15 @@ int QuectelNcpClient::muxChannelStateCb(uint8_t channel, decltype(muxer_)::Chann
         switch (channel) {
             case 0: {
                 // Muxer stopped
-                self->connectionState(NcpConnectionState::DISCONNECTED);
                 self->disable();
                 break;
             }
             case QUECTEL_NCP_PPP_CHANNEL: {
                 // PPP channel closed
                 if (self->connState_ != NcpConnectionState::DISCONNECTED) {
+                    // It should be safe to notify the PPP netif/client about a change of state
+                    // here exactly because the muxer channel is closed and there is no
+                    // chance for a deadlock.
                     self->connectionState(NcpConnectionState::CONNECTING);
                 }
                 break;
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -360,8 +360,12 @@ int SaraNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
     if (err) {
         // Make sure we are going into an error state if muxer for some reason fails
         // to write into the data channel.
-        disable();
         connectionState(NcpConnectionState::DISCONNECTED);
+        // disable() sets the ncpState_ to NcpState::DISABLED which prevents connectionState()
+        // from even trying to send events to the PPP client to go to a LOWER_DOWN state,
+        // and it might still not get there if connState_ is already DISCONNECTED, but order matters.
+        // Call disable() after connectionState()
+        disable();
     }
 
     return err;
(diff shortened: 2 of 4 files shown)


hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -360,8 +361,12 @@ int SaraNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
     if (err) {
         // Make sure we are going into an error state if muxer for some reason fails
         // to write into the data channel.
-        disable();
         connectionState(NcpConnectionState::DISCONNECTED);
+        // disable() sets the ncpState_ to NcpState::DISABLED which prevents connectionState()
+        // from even trying to send events to the PPP client to go to a LOWER_DOWN state,
+        // and it might still not get there if connState_ is already DISCONNECTED, but order matters.
+        // Call disable() after connectionState()
+        disable();
```
technobly: It should be noted that B5 SoM does not contain this same error handling block.  Should we add it?

avtolstoy: We should. B5 SoM code was branched off of this one before the change was made.

hal/src/boron/network/sara_ncp_client.h
```diff
@@ -73,6 +73,7 @@ class SaraNcpClient: public CellularNcpClient {
     volatile NcpState ncpState_ = NcpState::OFF;
     volatile NcpState prevNcpState_;
     volatile NcpConnectionState connState_ = NcpConnectionState::DISCONNECTED;
+    bool muxerChangedState_ = false;
```
avtolstoy: Let's not introduce flags to track specific conditions where we already have states and a kind of a state machine to handle things. This approach has the tendency to get out of control (see our main system event loop as an example of that).

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -1130,12 +1137,14 @@ int SaraNcpClient::muxChannelStateCb(uint8_t channel, decltype(muxer_)::ChannelS
                 // Muxer stopped
                 self->disable();
                 self->connState_ = NcpConnectionState::DISCONNECTED;
```
avtolstoy: I think this is actually the main problem. We've changed the state manually, without generating the event and also preventing the NCP client to do so properly. Calling just `disable()` should be enough to trigger a call to `enable()` in the next loop invocation (in the `PppNcpNetif`), which in turn will call `off()`, which will call `ncpState()`, which will call `connectionState()`.

We may potentially do the same thing similarly in https://github.com/particle-iot/device-os/pull/2049/files#diff-b3f74fd1e22ae53aae264584d668ea1cL363

technobly: What about the other UBLOX_NCP_PPP_CHANNEL case in the muxer callback:
```
if (self->connState_ != NcpConnectionState::DISCONNECTED) {
    self->connState_ = NcpConnectionState::CONNECTING;
}
```
How would you propose we change/update this?

Initially I did simply remove `self->connState_ = NcpConnectionState::DISCONNECTED;` from the callback, and it seemed to resolve the issue... but when I checked what state was being seen by `connectionState()` on the next call that produced a PPP state change event, it was `NcpConnectionState::CONNECTING`, not `DISCONNECTED`.  This didn't seem like the right way to resolve the issue.  We want the shut down of the muxer to trigger a disable/shut down of the NCP and also the PPP client.  Seeing a `CONNECTING` state didn't seem like the right way to do that.

Also, the ESP32 code calls the `connectionState()` function directly from the muxer callback.  That would be a nice way to fix it as well, but it didn't seem like it was thread safe from the comments in the SARA muxer callback.

hal/src/boron/network/sara_ncp_client.h
technobly: I had thought about adding a `NcpConnectionState::DISCONNECTING` state, but it was not obvious what else that might impact.  The idea of this flag is just to know when the NcpConnectionState has been changed outside of the place where it should only be changed.  If we can resolve that and make sure state is changed only in connectionState() directly, we shouldn't need any extra flags.  I would argue we don't need any additional state machines to track this either (the fact that NcpConnectionState is updated in the muxer callback).

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: > Also, the ESP32 code calls the connectionState() function directly from the muxer callback. That would be a nice way to fix it as well, but it didn't seem like it was thread safe from the comments in the SARA muxer callback.

We can actually do that (why don't we do that)? It's safe to call with `NcpConnectionState::DISCONNECTED`, because that won't call into the muxer. Let's do that.

> How would you propose we change/update this?

```c++
connectionState(NcpConnectionState::DISCONNECTED);
```

>  but when I checked what state was being seen by connectionState() on the next call that produced a PPP state change event, it was NcpConnectionState::CONNECTING, not DISCONNECTED

I'm not really sure where `CONNECTING` would be coming from, as that state is only generated by processEvents/checkRegistrationState, which would be bypassed in disabled state or `registerNet()`. And anyway, that state will trigger the teardown of the PPP session.

I'd also vote for calling just `disable()` in `dataChannelWrite()` on error conditions.

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -878,6 +891,7 @@ void QuectelNcpClient::connectionState(NcpConnectionState state) {
     if (ncpState_ == NcpState::DISABLED) {
         return;
     }
+    // Allow entry if state changed, or if the muxer directly changed the state
```
avtolstoy: Old comment?

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -922,22 +936,25 @@ int QuectelNcpClient::muxChannelStateCb(uint8_t channel, decltype(muxer_)::Chann
...
-            self->connState_ = NcpConnectionState::DISCONNECTED;
-            break;
-        }
-        case QUECTEL_NCP_PPP_CHANNEL: {
-            // PPP channel closed
-            if (self->connState_ != NcpConnectionState::DISCONNECTED) {
-                self->connState_ = NcpConnectionState::CONNECTING;
+            case 0: {
+                // Muxer stopped
+                self->disable();
```
avtolstoy: Will not work in this order. I would argue that `connectionState()` call is not needed here.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -1123,19 +1128,22 @@ int SaraNcpClient::muxChannelStateCb(uint8_t channel, decltype(muxer_)::ChannelS
...
+    // from this callback.
+
     // We are only interested in Closed state
     if (newState == decltype(muxer_)::ChannelState::Closed) {
         switch (channel) {
             case 0: {
                 // Muxer stopped
                 self->disable();
-                self->connState_ = NcpConnectionState::DISCONNECTED;
+                self->connectionState(NcpConnectionState::DISCONNECTED);
```
avtolstoy: Will not work in this order. I would argue that `connectionState()` call is not needed here.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -1080,6 +1084,7 @@ void SaraNcpClient::connectionState(NcpConnectionState state) {
     if (ncpState_ == NcpState::DISABLED) {
         return;
     }
+    // Allow entry if state changed, or if the muxer directly changed the state
```
avtolstoy: Old comment?

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -360,7 +360,20 @@ int QuectelNcpClient::updateFirmware(InputStream* file, size_t size) {
 }
 
 int QuectelNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
-    return muxer_.writeChannel(QUECTEL_NCP_PPP_CHANNEL, data, size);
+    int err = muxer_.writeChannel(QUECTEL_NCP_PPP_CHANNEL, data, size);
+
+    if (err) {
+        // Make sure we are going into an error state if muxer for some reason fails
+        // to write into the data channel.
+        connectionState(NcpConnectionState::DISCONNECTED);
```
avtolstoy: It's also probably fine to just call disable here and in all the other `dataChannelWrite()` functions.

avtolstoy:
PR updated. Remove `connectionState()` from a couple places to not hit potential race conditions/deadlocks. Test `NETWORK_02_network_connection_recovers_after_ncp_failure` in `wiring/no_fixture_long_running` test suite also has been added to simulate the failure that triggers this problem.

The issue this PR resolves is only reproducible on 2G/3G Borons or BSoMs due to the fact that the GSM0710 muxer keeplive timeout is lower than the PPP keepalive timeout: https://github.com/particle-iot/device-os/blob/develop/hal/src/boron/network/sara_ncp_client.cpp#L933

TODO: just in case in the future we could increase the timeouts to be the same as on R4-based Borons.

PR #2051, opened 2020-03-17
Fix YModem transfer

sergeuz:
### Problem

This PR fixes a regression introduced in https://github.com/particle-iot/device-os/pull/2041, which relies on a wrong assumption that the `FileTransfer::Descriptor` class correctly initializes itself. This is a good example of why all class member and stack variables should be initialized explicitly at the time of their declaration, even if their initial values are overwritten later in the code.

### Steps to Test

Firmware update via YModem should succeed.

### DIFF
--- a/communication/inc/file_transfer.h
+++ b/communication/inc/file_transfer.h
@@ -61,7 +61,12 @@ namespace FileTransfer {
 
     struct Descriptor : public Chunk
     {
-        Descriptor() { size = sizeof(*this); }
+        Descriptor() :
+                Chunk(),
+                file_length(0),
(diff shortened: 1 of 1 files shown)


elcojacobs:
Will I have to wait until April 6 before I can use an official working system layer from Particle with the latest bugfixes? Or will you have a release candidate earlier than that?
I seem to have unreliable connections with earlier versions, 1.5.0-rc.2 works well except for the Ymodem issue.

Current develop seems to work without any issues. I can use Ymodem to send new firmware over Serial or WiFi.

PR #2052, opened 2020-03-18
[hal] concurrent: expose full concurrent APIs.

XuGuohui:
Expose the rest of concurrent HAL APIs.

### Completeness

- [ ] Added documentation


eberseth:
Please add conditional variable support as well

avtolstoy:
@eberseth Conditional variable implementation is broken. We've previously agreed that we are not going to expose them. If we do require this feature for other work, we should prioritize changing the implementation.

We can always add the support in a separate PR.

eberseth:
Okay, we'll use an alternate implementation.

avtolstoy:
@eberseth Alternate implementation within concurrent HAL or outside of it?

eberseth:
Perhaps copying out the cv implementation into user space versus using a standard library.

PR #2056, opened 2020-03-20
[system] close cloud socket on errors

avtolstoy:
### Problem

Cloud socket is not closed on communication errors. With heavy traffic unread data may be left in it, which may prevent Gen 3 devices from establishing PPP link or performing network communications due to lack of TX/RX buffers.

### Solution

Close the cloud socket on errors.

Caveat: SARA R4 internal TCP/IP stack (used on Electrons) leaks sockets (opening/closing them too often will eventually prevent us from creating any more sockets). We leave the existing behavior on Electrons.

### Steps to Test

Run the following app on cellular Gen 3 device. Eventually after successfully publishing for a while, the device may get into a blinking green state. It should recover successfully.

```c++
#include "application.h"
void setup() {
}

const int num_runs = 10;

void loop() {
    static system_tick_t last = 0;
    for (int i=0; i < num_runs; i++) {
        delay(250);
        Particle.publish("dummy", "123");
        if (millis() - last >= 1000) {
            last = millis();
            int32_t value = -1;
            AbstractIntegerDiagnosticData::get(DIAG_ID_CLOUD_RATE_LIMITED_EVENTS, value);
            Log.trace("rate limited: %d", value);
        }
    }
}
```

### References

- [CH50189]

---

### Completeness

- [ ] Added documentation


system/src/system_cloud_internal.cpp
```diff
@@ -844,7 +844,7 @@ bool system_cloud_active()
         if (SPARK_CLOUD_CONNECTED && ((now-lastCloudEvent))>SYSTEM_CLOUD_TIMEOUT)
         {
             WARN("Disconnecting cloud due to inactivity! %d, %d", now, lastCloudEvent);
-            cloud_disconnect(false); // TODO: Do we need to specify a reason of the disconnection here?
+            cloud_disconnect(HAL_PLATFORM_MAY_LEAK_SOCKETS ? false : true, false, CLOUD_DISCONNECT_REASON_ERROR);
```
keeramis: Are the true and false supposed to be other way round? Meaning, 
`HAL_PLATFORM_MAY_LEAK_SOCKETS ? true : false` because we now want to close socket on errors.

avtolstoy: The opposite. `HAL_PLATFORM_MAY_LEAK_SOCKETS` -> we cannot close sockets easily -> `cloud_disconnect(false)` -> do not close socket. This is for Electrons with SARA R4:

> Caveat: SARA R4 internal TCP/IP stack (used on Electrons) leaks sockets (opening/closing them too often will eventually prevent us from creating any more sockets). We leave the existing behavior on Electrons.

avtolstoy: If you have a better name for this feature flag in mind, I'm open to suggestions :smile:

keeramis: OK I understand. Yes, when I first read the `HAL_PLATFORM_MAY_LEAK_SOCKETS` it sounded like it may leak sockets, so fix it. 
I would rename it to something like `HAL_PLATFORM_ALLOWED_TO_LEAK_SOCKETS` but it's too long, so `HAL_PLATFORM_ALLOW_SOCKET_LEAK_IF_ANY` or how about just `HAL_PLATFORM_ALLOW_SOCKET_LEAK`

keeramis:
Neat change! Would we want to notify docs or elsewhere about Gen2 could still blink green (or become weird) if the device is abused with lots of Particle.publishes or such?

system/src/system_cloud_internal.cpp
avtolstoy: I was trying to phrase the opposite that the platform (Electron with R4) itself leaks sockets. Not that we are leaking them, but the underlying network stack is buggy and we have to workaround it.

avtolstoy:
@keeramis We won't see such an issue on Gen 2 cellular devices, simply because the packet buffers are not managed by us.

keeramis:
Would we want to notify docs or elsewhere about Gen2 not

PR #2062, opened 2020-03-27
[Boron/B SoM/B5 SoM] Fix Cellular Global Identity generation

avtolstoy:
### Problem

There is a race condition between CREG/CGREG/CEREG URC handlers that may leave LAC and Cell ID parts of CGI invalid.

### Solution

Set conditionally based on RAT in CGREG and CEREG handlers, fallback to CREG.

### Steps to Test

N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/b5som/network/quectel_ncp_client.cpp
+++ b/hal/src/b5som/network/quectel_ncp_client.cpp
@@ -203,7 +203,7 @@ int QuectelNcpClient::initParser(Stream* stream) {
         self->checkRegistrationState();
         // Cellular Global Identity (partial)
         // Only update if unset
-        if (r >= 2 && r >= 3) {
+        if (r >= 3) {
             if (self->cgi_.location_area_code != std::numeric_limits<LacType>::max()) {
                 self->cgi_.location_area_code = static_cast<LacType>(val[1]);
             }
@@ -237,8 +237,9 @@ int QuectelNcpClient::initParser(Stream* stream) {
         }
         self->checkRegistrationState();
         // Cellular Global Identity (partial)
-        if (r >= 2 && r >= 3) {
-            switch (self->act_) {
+        if (r >= 3) {
+            auto rat = r >= 4 ? static_cast<CellularAccessTechnology>(val[3]) : self->act_;
+            switch (rat) {
                 case CellularAccessTechnology::GSM:
                 case CellularAccessTechnology::GSM_COMPACT:
                 case CellularAccessTechnology::UTRAN:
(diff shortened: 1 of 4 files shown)


hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -193,8 +193,15 @@ int SaraNcpClient::initParser(Stream* stream) {
         }
         self->checkRegistrationState();
         // Cellular Global Identity (partial)
-        self->cgi_.location_area_code = r >= 2 ? static_cast<LacType>(val[1]) : std::numeric_limits<LacType>::max();
-        self->cgi_.cell_id = r >= 3 ? static_cast<CidType>(val[2]) : std::numeric_limits<CidType>::max();
+        // Only update if unset
+        if (r >= 2 && r >= 3) {
```
keeramis: Redundant? just `r>=3` is enough?

avtolstoy: Already fixed)

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -492,7 +524,21 @@ int QuectelNcpClient::getCellularGlobalIdentity(CellularGlobalIdentity* cgi) {
...
+    // FIXME: this is a workaround for CH28408
+    CellularSignalQuality qual;
+    CHECK(queryAndParseAtCops(&qual));
+    CHECK_TRUE(qual.accessTechnology() != CellularAccessTechnology::NONE, SYSTEM_ERROR_INVALID_STATE);
+    // Update current RAT
+    act_ = qual.accessTechnology();
+    // Invalidate LAC and Cell ID
+    cgi_.location_area_code = std::numeric_limits<LacType>::max();
+    cgi_.cell_id = std::numeric_limits<CidType>::max();
+    // Fill in LAC and Cell ID based on current RAT, prefer PSD and EPS
```
keeramis: In case we did not get the LAC and CID from the At commands (for any reason), we will be sending the max values to console. Is that something we need to aware the device-service with......?

avtolstoy: That's an error value, which is handled in `cellular_global_identity()` resulting in `SYSTEM_ERROR_BAD_DATA` sent in vitals, which we've seen as `-280` errors.

keeramis: Max will look like `0xFFFFFFF`? 
But that is also a valid cell id as per 
https://en.wikipedia.org/wiki/GSM_Cell_ID

avtolstoy: Max is 0xffffffff, which is not a valid cell id, as it's 2^32 -1.

> A valid CID ranges from 0 to 65535 (2^16 − 1) on GSM and CDMA networks and from 0 to 268,435,455 (2^28 − 1) on UMTS and LTE networks.

keeramis: Perfect, thanks

PR #2070, opened 2020-04-01
[electron] cellular: compat rssi/qual for LTE devices using UCGED

avtolstoy:
### Problem

Old API rssi and qual values are not available on LTE Electrons/E Series because we've switched over to UCGED in #2033.

### Solution

Add a compatibiltiy layer, same as on Gen 3.

### Steps to Test

Use the app. On Electron LTE it should provide sane values instead of `2,0` and should be comparable to pre-1.5.0.

### Example App

```c++

void setup() {
}

void loop() {
    auto sig = Cellular.RSSI();
    Log.trace("%d, %d", sig.rssi, sig.qual);
    delay(5s);
}
```

### References

- [CH51264]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -156,6 +156,9 @@ AcT toCellularAccessTechnology(int rat) {
     }
 }
 
+// see 3GPP TS 45.008 [20] subclause 8.2.4
+const char compatQualMap[] = { 49, 43, 37, 25, 19, 13, 7, 0 };
+
 } // anonymous
 
 #ifdef MDM_DEBUG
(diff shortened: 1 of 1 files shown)


keeramis:
Result from an Electron LTE after this fix: 

image

keeramis:
Logs from Electron LTE:

> 0000011291 [app] TRACE: -86, 19
0000016340 [app] TRACE: -87, 25
0000036650 [app] TRACE: -86, 25
0000041711 [app] TRACE: -87, 37
0000046779 [app] TRACE: -85, 37

keeramis:
Also tested a case where signal is lost and there is a `CME ERROR` on `UCGED`, and it gives 
`0001991722 [app] TRACE: 1, 0`

Issue #2055, opened 2020-03-19
ASSERTION_ERROR with modular build on p1


elcojacobs:
##  Bug Report
When compiling the default app for a p1 with MODULAR=y, I get an assertion error SOS.

### Observed Behavior

Panic code of 10 red flashes.

### Steps to Reproduce

cd device-os/main
make clean all program-dfu PLATFORM=p1 MODULAR=n

### Test App

Not needed, juist build the default app
I'm building with 8-2019-q3-update 8.3.1 20190703 and Boost 1.72

Please let me know if this reproduces. I can't get openocd hardware debugging to work anymore, possibly this is related.
I can start the debugger and connect to the target, but it seems to get an exception before I can break anywhere meaningful.

avtolstoy:
> I'm building with 8-2019-q3-update 8.3.1 20190703 

We currently do not support any gcc version other than 5.3.1, even if DeviceOS may currently be buildable with it. Is the issue reproducible with 5.3.1?

elcojacobs:
No hardfault with 5.3.

Warning, what follows is a bit of a rant. I am getting a bit frustrated. To the point where I think my next hardware generation won't have a chipset from Particle.

So, my take on this:
This is a bug. If it doesn't express itself with 5.3, you are just lucky. Perhaps it is an uninitialized variable that just happens to be zero most of the time. Maybe next release it isn't.

The recent bug with Ymodem not working because of uninitialized member variables, it is just silly and unnecessary. 
What we have seen in our app for months is that a Ymodem firmware sometimes failed with a hardfault. The device would reboot and it would work on the second or third try. Only with 1.5.0-rc.2 it would fail every time. Uninitialized memory is more likely to be zero just after boot, which explains why a reboot often helped. Until 1.5.0-rc.2, in which memory use elsewhere int he code probably increased the change of it not being zero by luck. But this bug was totally unnecessary. Use `-Wmissing-field-initializers` and the compiler will just tell you to fix it.


Using an outdated compiler is annoying because new features cannot be used, it includes on old version of the STL and because the warnings are a lot better in more recent versions.

Try building a minimal app with some extra compiler warnings on GCC 8. May I suggest some:

```
CFLAGS += -ffunction-sections -Wall
CFLAGS += -Wno-deprecated
CFLAGS += -Wnull-dereference 
CPPFLAGS += -fdiagnostics-show-template-tree -fno-elide-type # GCC8 required
CFLAGS += -Wextra
[...]

PR #2073, opened 2020-04-02
fix uninizialized member in led signal struct

elcojacobs:
Such a simple fix. Disappointed to see it was not fixed in the latest release.
You might want to initialize to something other than zero, but don't leave it uninitialized.

Please take a serious look at the other error messages in #2055.
The mismatched prototypes for system event handlers that are reinterpret_casted anyway is a bit harder for me to fix.


elcojacobs:
Another thing I noticed:
The order of the arguments switches around between the overloads. Are you sure that's not causing issues?
```
LEDStatus::LEDStatus(uint32_t color = RGB_COLOR_WHITE, LEDPattern pattern = LED_PATTERN_SOLID, LEDPriority priority = LED_PRIORITY_NORMAL); // 1
```
From the docs here:
https://docs.particle.io/reference/device-os/firmware/photon/#ledstatus-class

Which overload will it call? I don't see an overload with priority as the last argument!


```

// particle::LEDStatus
inline particle::LEDStatus::LEDStatus(LEDPriority priority, LEDSource source) :
        LEDStatus(LED_PATTERN_SOLID, priority, source) { // Use solid pattern by default
}

inline particle::LEDStatus::LEDStatus(LEDPattern pattern, LEDPriority priority, LEDSource source) :
        LEDStatus(pattern, LED_SPEED_NORMAL, priority, source) { // Use normal speed by default
}

inline particle::LEDStatus::LEDStatus(LEDPattern pattern, LEDSpeed speed, LEDPriority priority, LEDSource source) :
        LEDStatus(pattern, led_pattern_period(pattern, speed, nullptr), priority, source) { // Get pattern period for a predefined speed value
}

inline particle::LEDStatus::LEDStatus(LEDPattern pattern, uint16_t period, LEDPriority priority, LEDSource source) :
        LEDStatus(RGB_COLOR_WHITE, pattern, period, priority, source) { // Use white color by default
}

inline particle::LEDStatus::LEDStatus(uint32_t color, LEDPriority priority, LEDSource source) :
        LEDStatus(color, LED_PATTERN_SOLID, priority, source) { // Use solid pattern by default
}

inline particle::LEDStatus::LEDStatus(uint32_t color, LEDPattern pattern, LEDPriority priority, LEDSource source) :
        LEDStatus(color, pattern, LED_SPEED_NORMAL, priority, source) { // Use normal speed by default
}

inline particle::LEDStatus::LEDStatus(uint32_t color, LEDPattern pattern, LEDSpeed speed, LEDPriority priority, LEDSource source) :
        LEDStatus(color, pattern, led_pattern_period(pattern, speed, nullptr), priority, source) { // Get pattern period for a predefined speed value
}



```

sergeuz:
That initializer list in the constructor of the `LEDSystemTheme` class already initializes all members of `d_`, except for `version`, to 0. Explicitly initializing `signals` via `{0}` is redundant:

> If the number of initializer clauses is less than the number of members or initializer list is completely empty, the remaining members are value-initialized ([link](https://en.cppreference.com/w/cpp/language/aggregate_initialization)).

As for overloaded constructors of `LEDSignal`, please open an issue if they cause ambiguity errors in your application, or contradict the documentation.

PR #2079, opened 2020-04-16
[Gen 3] Network stack enhancements

avtolstoy:
### Problem

There are some issues in Gen 3 network stack that affect its performance:

1. Polling nature of `SerialStream`
2. Polling nature of GSM0710 muxer
3. Half-baked implementation of per-channel flow control in gsm0710muxer / `MuxerChannelStream`

### Solution

1. Gen 3 USART serial now has READABLE/WRITEABLE events available through the FreeRTOS event group. `SerialStream::waitEvent()` uses that to implement proper timed blocking waiting for these events
2. https://github.com/particle-iot/gsm0710muxer/pull/5
3. https://github.com/particle-iot/gsm0710muxer/pull/5

Apart from that this PR
1. Implements pbuf flow control using the facilities now provided by GSM07.10 muxer plus LwIP hooks implemented in https://github.com/particle-iot/lwip/pull/6
2. Adds a log line to chunked transfer to show how long the transfer took, as well as transfer rate
3. Increases the baudrate for all the cellular platforms whenever possible
4. Fixes a bug with firmware_update_complete / firmware_update_failed event generation
5. Makes sure that ENOMEM socket errors are not propagated to the communication layer, as those are most of the times temporary
6. Fixes a crash with event handling in `netif_ext_callback_handler()`
7. Adds flow control facilities to NCP-based interface drivers

### Steps to Test

Mostly just run OTA tests on all the affected platforms and compare.

### Example App

N/A

### References

- https://github.com/particle-iot/lwip/pull/6
- https://github.com/particle-iot/gsm0710muxer/pull/5
- [CH52051]
- [CH51306]
- [CH51305]
- [CH51304]

---

### Completeness

- [ ] Added documentation


hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -890,32 +924,37 @@ int SaraNcpClient::initReady() {
...
-        int r = CHECK_PARSER(resp.scanf("%32[^\n]", verExtended));
-        CHECK_TRUE(r == 1, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
-        r = CHECK_PARSER(resp.readResult());
-        CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
-        if (!strcmp(verExtended, "L0.0.00.00.05.06,A.02.00")) {
-            memoryIssuePresent_ = true;
+                if (r == SYSTEM_ERROR_NONE) {
+                    // Make sure flow control is enabled as well
+                    CHECK_PARSER_OK(parser_.execCommand("AT+IFC=2,2"));
+                }
```
sergeuz, line 946: This won't be executed unless the above call to `changeBaudRate()` fails with `SYSTEM_ERROR_AT_NOT_OK`

hal/src/nRF52840/gsm0710muxer/platform.h
```diff
@@ -41,8 +43,16 @@ const auto taskPriority = OS_THREAD_PRIORITY_NETWORK;
 #define GSM0710_ASSERT(x) SPARK_ASSERT(x)
 
 // #define GSM0710_MODEM_STATUS_SEND_EMPTY_BREAK
-#define GSM0710_PUMP_INPUT_DATA 50
+// #define GSM0710_PUMP_INPUT_DATA 50
 // #define GSM0710_ENABLE_DEBUG_LOGGING
+#define GSM0710_SHARED_STREAM_EVENT_GROUP
+#define GSM0710_SHARED_STREAM_EVENT_BASE (HAL_USART_PVT_EVENT_END - 1)
```
sergeuz, line 49: What does the value of `HAL_USART_PVT_EVENT_END` mean?

system/src/system_cloud_connection_posix.cpp
```diff
@@ -287,17 +287,29 @@ int system_cloud_disconnect(int flags)
 int system_cloud_send(const uint8_t* buf, size_t buflen, int flags)
 {
     (void)flags;
-    return sock_send(s_state.socket, buf, buflen, 0);
+    int r = sock_send(s_state.socket, buf, buflen, 0);
+    if (r < 0) {
+        if (errno == ENOMEM) {
+            /* Not an error */
+            r = 0;
```
sergeuz, line 294: Let's add a debug-only warning here

system/src/system_cloud_connection_posix.cpp
```diff
@@ -287,17 +287,29 @@ int system_cloud_disconnect(int flags)
...
 }
 
 int system_cloud_recv(uint8_t* buf, size_t buflen, int flags)
 {
     (void)flags;
     int recvd = sock_recv(s_state.socket, buf, buflen, MSG_DONTWAIT);
     if (recvd < 0) {
-        if (errno == EWOULDBLOCK || errno == EAGAIN) {
+        if (errno == EWOULDBLOCK || errno == EAGAIN || errno == ENOMEM) {
             /* Not an error */
```
sergeuz, line 309: Same here for `ENOMEM`

system/src/system_update.cpp
```diff
@@ -401,7 +401,7 @@ int Spark_Finish_Firmware_Update(FileTransfer::Descriptor& file, uint32_t flags,
         if (file.store==FileTransfer::Store::FIRMWARE)
         {
             hal_update_complete_t result = HAL_FLASH_End(module ? (hal_module_t*)module : &mod);
-            system_notify_event(firmware_update, result<=HAL_UPDATE_ERROR ? firmware_update_complete : firmware_update_failed, &file);
+            system_notify_event(firmware_update, result<=HAL_UPDATE_ERROR ? firmware_update_failed : firmware_update_complete, &file);
```
sergeuz, line 404: :joy:

hal/src/nRF52840/gsm0710muxer/platform.h
avtolstoy: The next flag bit we can use for our own purposes.

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: The condition is `if (r == SYSTEM_ERROR_NONE) {` though, so it only gets executed if the above call to change the baud rate succeeds, right?

sergeuz: The above call returns from the function on any value including SYSTEM_ERROR_NONE

avtolstoy: Oh, right. :+1: Let's fix that.

hal/src/nRF52840/gsm0710muxer/platform.h
sergeuz: Then it should use something like `__builtin_ffs`? It's currently set to 0x03 which is a combination of flag bits, not a single flag bit

avtolstoy: Yeah, let's make it a bit clearer. It does work _currently_ (https://github.com/particle-iot/gsm0710muxer/pull/5/files#diff-8d017d4d053da8f60b8c24e8c0f581a3R217) but agreed that it's easy to break.

sergeuz: It does, it's just the formula that didn't make any sense to me

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -890,32 +924,37 @@ int SaraNcpClient::initReady() {
...
+            // L0.0.00.00.05.06,A.02.00 has a memory issue
+            memoryIssuePresent_ = (fwVersion == UBLOX_NCP_R4_APP_FW_VERSION_MEMORY_LEAK_ISSUE);
+            // There is a set of other revisions which do not have hardware flow control
+            if (!(fwVersion >= UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MIN &&
+                    fwVersion <= UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MAX)) {
+                // Change the baudrate to 460800
+                // NOTE: ignoring AT errors just in case to accommodate for some revisions
+                // potentially not supporting anything other than 115200
+                int r = changeBaudRate(UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_R4);
+                if (r != SYSTEM_ERROR_AT_NOT_OK) {
```
keeramis, line 939: This should be `r==` and not `r!=`?

keeramis, line 939: May need to check for AT here, before we tend to set `AT+IFC` in the next lines
```
    // Check that the modem is responsive at the new baudrate
    skipAll(serial_.get(), 1000);
    CHECK(waitAtResponse(10000));
```

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -84,7 +85,11 @@ inline system_tick_t millis() {
 }
 
 const auto UBLOX_NCP_DEFAULT_SERIAL_BAUDRATE = 115200;
-const auto UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_U2 = 115200;
+const auto UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_U2 = 921600;
+const auto UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_R4 = 460800;
```
keeramis, line 89: Currently, the baud rate setting is stored in NVM on the current latest cell modem version `L0.0.00.00.05.08,A.02.04`, so it persists across device reboots. This only happens on the R410, and doesn't apply to U210.
Our code assumes that the baud rate setting is not persistent, we always start by setting the USART2 modem's serial baud rate to `115200`, which in this case will cause the modem to not respond if the modem was previously set to a different baud rate like `460800`. We can potentially use a flash variable to check this, but this is not ideal because it will not be backwards compatible - for example, if a user goes back to 1.4.4, they may no longer be able to correctly talk to the cell radio. Need to check with ublox if we can make this change non-persistent similar to U201.

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: Yeah, you are right, let's do that.

hal/src/nRF52840/gsm0710muxer/platform.h
avtolstoy: Fixed.

system/src/system_cloud_connection_posix.cpp
avtolstoy: Added.

system/src/system_cloud_connection_posix.cpp
avtolstoy: Added.

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: Added a check for `!= SYSTEM_ERROR_NONE && != SYSTEM_ERROR_AT_NOT_OK`

avtolstoy: Simply moved `AT+IFC=2,2` below that check.

hal/src/boron/network/sara_ncp_client.cpp
avtolstoy: Disabled 460800 baudrate for now until we get a response from ublox. Any additional changes after receiving reply from ublox can be made in a separate PR.

hal/inc/usart_hal_private.h
```diff
@@ -0,0 +1,45 @@
...
+#include "system_tick_hal.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+typedef enum HAL_USART_Pvt_Events {
+    HAL_USART_PVT_EVENT_READABLE = 0x01,
+    HAL_USART_PVT_EVENT_WRITABLE = 0x02,
+    HAL_USART_PVT_EVENT_READABLE_WRITABLE = 0x03,
```
sergeuz, line 34: I'd remove this enum value just to make it obvious that `HAL_USART_PVT_EVENT_MAX` is the maximum flag value, not the maximum of all enum values. This value is also not used anywhere and is redundant given that `HAL_USART_Pvt_Wait_Event()` takes a combination of flags anyway.

avtolstoy: Sounds fine to me. Done.

PR #2080, opened 2020-04-17
[hal] i2c: bus is driven low if building with JTAG/SWD enabled.

XuGuohui:
### Problem

I2C bus is driven LOW after `Wire.begin()` when building with JTAG/SWD enabled on Electron.

### Solution

Initialize I2C pins to be float input.
### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {
    Wire.begin();
}

void loop() {
}
```

### Steps to test

1. Build modular firmware with JTAG/SWD enabled:
`make PLATFORM=electron DEBUG_BUILD=y USE_SWD_JTAG=y`

2. Flash system firmware and the above application.
3. Use logic analyser to monitor the SDA(D0) and SCL(D1) and see if both pins go to high level or not.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/stm32f2xx/core_hal_stm32f2xx.c
+++ b/hal/src/stm32f2xx/core_hal_stm32f2xx.c
@@ -345,25 +345,24 @@ void HAL_Core_Config(void)
     PWR_WakeUpPinCmd(DISABLE);
 
     //Wiring pins default to inputs
-    for (pin_t pin = 0; pin <= 19; pin++) {
+    for (pin_t pin = 0; pin < TOTAL_ESSENTIAL_PINS; pin++) {
 #if defined(USE_SWD_JTAG) || defined(USE_SWD)
-        if (pin >= 3 && pin <= 7) {
+        if (pin >= D3 && pin <= D7) { // JTAG pins
             continue;
         }
 #endif
         HAL_Pin_Mode(pin, INPUT);
     }
-#if PLATFORM_ID==8 // Additional pins for P1
-    for (pin_t pin = 24; pin <= 29; pin++) {
+
+#if HAS_EXTRA_PINS
+    for (pin_t pin = FIRST_EXTRA_PIN; pin <= LAST_EXTRA_PIN; pin++) {
         HAL_Pin_Mode(pin, INPUT);
-    }
-    if (isWiFiPowersaveClockDisabled()) {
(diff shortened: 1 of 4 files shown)


hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -345,20 +345,26 @@ void HAL_Core_Config(void)
     PWR_WakeUpPinCmd(DISABLE);
 
     //Wiring pins default to inputs
-#if !defined(USE_SWD_JTAG) && !defined(USE_SWD)
-    for (pin_t pin=0; pin<=19; pin++)
+    for (pin_t pin = 0; pin <= 19; pin++) {
```
avtolstoy, line 348: I understand this is very old code, but since we are modifying it, can we perhaps refactor this somehow, so that we use `TOTAL_PINS`, pin names etc here?

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -237,6 +237,8 @@ int HAL_I2C_Init(HAL_I2C_Interface i2c, const HAL_I2C_Config* config)
     if(i2c == HAL_I2C_INTERFACE1)
     {
         i2cMap[i2c] = &I2C_MAP[I2C1_D0_D1];
+        HAL_Pin_Mode(i2cMap[i2c]->I2C_SDA_Pin, INPUT);
```
avtolstoy, line 240: Looks weird that we are only initializing D0 and D1 and not not any pins from the other interfaces. Do we need to do this in init. or can we rely on `HAL_I2C_Begin()` to simply initialize pins as AF?

hal/src/stm32f2xx/core_hal_stm32f2xx.c
```diff
@@ -345,20 +345,25 @@ void HAL_Core_Config(void)
...
-    for (pin_t pin=24; pin<=29; pin++)
+    }
+
+#if HAS_EXTRA_PINS
+    for (pin_t pin = FIRST_EXTRA_PIN; pin <= LAST_EXTRA_PIN; pin++) {
         HAL_Pin_Mode(pin, INPUT);
+    }
+#endif
+
+#if PLATFORM_ID==8
```
avtolstoy, line 363: Use definitions from `platforms.h`

PR #2083, opened 2020-04-20
Map AccessTechnology correctly on SARA R410M

keeramis:
- Access technology on R410M modems on Boron LTE is sometimes registered as `7` and sometimes registered as `8`. Hence, mapping them correctly for the R410M platform, by considering both `7` and `8` as Cat-m1 RAT.
- Rename access technology related enums to be more understandable.

Story details: https://app.clubhouse.io/particle/story/49272

### DIFF
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -40,6 +40,7 @@
 #include <algorithm>
 #include <limits>
 #include <lwip/memp.h>
+#include "enumclass.h"
 
 #undef LOG_COMPILE_TIME_LEVEL
 #define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
@@ -270,8 +271,8 @@ int SaraNcpClient::initParser(Stream* stream) {
             auto rat = r >= 4 ? static_cast<CellularAccessTechnology>(val[3]) : self->act_;
             switch (rat) {
                 case CellularAccessTechnology::LTE:
-                case CellularAccessTechnology::EC_GSM_IOT:
-                case CellularAccessTechnology::E_UTRAN: {
+                case CellularAccessTechnology::LTE_CAT_M1:
+                case CellularAccessTechnology::LTE_NB_IOT: {
                     self->cgi_.location_area_code = static_cast<LacType>(val[1]);
                     self->cgi_.cell_id = static_cast<CidType>(val[2]);
                     break;
@@ -494,6 +495,12 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
     cgi_.mobile_country_code = static_cast<uint16_t>(::atoi(mobileCountryCode));
     cgi_.mobile_network_code = static_cast<uint16_t>(::atoi(mobileNetworkCode));
 
+    if (ncpId() == PLATFORM_NCP_SARA_R410) {
+        if (act == particle::to_underlying(CellularAccessTechnology::LTE)) {
+            act = particle::to_underlying(CellularAccessTechnology::LTE_CAT_M1);
+        }
+    }
+
     switch (static_cast<CellularAccessTechnology>(act)) {
         case CellularAccessTechnology::NONE:
         case CellularAccessTechnology::GSM:
@@ -504,8 +511,8 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
         case CellularAccessTechnology::UTRAN_HSUPA:
         case CellularAccessTechnology::UTRAN_HSDPA_HSUPA:
         case CellularAccessTechnology::LTE:
-        case CellularAccessTechnology::EC_GSM_IOT:
-        case CellularAccessTechnology::E_UTRAN: {
+        case CellularAccessTechnology::LTE_CAT_M1:
+        case CellularAccessTechnology::LTE_NB_IOT: {
             break;
         }
         default: {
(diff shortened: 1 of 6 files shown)


keeramis:
@avtolstoy I checked that this is not cell radio firmware specific. I have tested this on `05.08-02.04` and `05.06-02.00`, and they both atm report `8` on Cat-M1. I do believe this is either network conditions or SIM specific difference.

avtolstoy:
@keeramis The CI fails to build.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -268,10 +268,16 @@ int SaraNcpClient::initParser(Stream* stream) {
         // Cellular Global Identity (partial)
         if (r >= 3) {
             auto rat = r >= 4 ? static_cast<CellularAccessTechnology>(val[3]) : self->act_;
+            if (self->ncpId() == PLATFORM_NCP_SARA_R410) {
+                if (rat == CellularAccessTechnology::LTE) {
+                    rat = CellularAccessTechnology::LTE_CAT_M1;
+                    //LOG(INFO, "RAT changed to report Cat-M1");
```
avtolstoy, line 274: Do we need to keep these commented out comments here and in all the other places as well?

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -268,10 +268,16 @@ int SaraNcpClient::initParser(Stream* stream) {
         // Cellular Global Identity (partial)
         if (r >= 3) {
             auto rat = r >= 4 ? static_cast<CellularAccessTechnology>(val[3]) : self->act_;
+            if (self->ncpId() == PLATFORM_NCP_SARA_R410) {
+                if (rat == CellularAccessTechnology::LTE) {
```
avtolstoy, line 272: This is only use locally and does not matter whether it's LTE, LTE Cat M1 or any other LTE type. We can probably omit the fixup here specifically.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -494,6 +500,13 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
     cgi_.mobile_country_code = static_cast<uint16_t>(::atoi(mobileCountryCode));
     cgi_.mobile_network_code = static_cast<uint16_t>(::atoi(mobileNetworkCode));
 
+    if (ncpId() == PLATFORM_NCP_SARA_R410) {
+        if (act == static_cast<int>(CellularAccessTechnology::LTE)) {
```
avtolstoy, line 504: (Just FYI) There is also `particle::to_underlying` in `enumclass.h` to make the enum class conversions a bit more pretty.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -497,6 +497,12 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                             if ((r >= 4) && (c != (int)0xFFFFFFFF))  _net.cgi.cell_id  = c;
                             // access technology
                             if (r >= 5) {
+                                if (_dev.dev == DEV_SARA_R410) {
+                                    if (d == static_cast<int>(UbloxSaraCellularAccessTechnology::UBLOX_SARA_RAT_LTE)) {
+                                        d = static_cast<int>(UbloxSaraCellularAccessTechnology::UBLOX_SARA_RAT_LTE_CAT_M1);
```
avtolstoy, line 502: Can we instead move the conversions out of the callbacks and into the actual `getSignalStrength()` where it really matters? We already have the fixup there for G350.

hal/src/boron/network/sara_ncp_client.cpp
keeramis: For now yes. But this will be used if in the future, we use a SARA with a Cat-1 modem for example, and then differentiate RATs for whatever we plan to do here? 🤔

avtolstoy: The main thing we are checking here is still whether we are on LTE, no matter its type. So, I don't see why we need to add a fixup here. 7 is still LTE and this is just a local variable.

keeramis: For example a case where, we add additional stuff into `case CellularAccessTechnology::LTE:`, then we want to differentiate the cat-m1 modem and the cat-1/lteonly modem.

avtolstoy: I doubt we'd ever need to do that here specifically. This value is only used to know whether we are on LTE locally within this function (any type of it) to generate the CGI.

keeramis: If we don't ever do that then that's good. The point I am mentioning here is that if we do something different for LTE vs LTE-Catm1 for a different radio in the future for sara, then the new code will look like this, and we might need a differentiation between R410 and other LTE radios in that case...........? 
```
case CellularAccessTechnology::LTE:
       // CGI data + something specific for an LTE modem only
case CellularAccessTechnology::LTE_CAT_M1:
      // CGI data
case CellularAccessTechnology::LTE_NB_IOT: 
       // CGI data
       break;
```

avtolstoy: I understand that possibility, however for all intents and purposes the data here specifically returned by `AT+CEREG` is exactly the same no matter which LTE type we are on.

If you are really keen on fixing this everywhere, let's make it a bit more cleaner and introduce some kind of a function which would take RAT number and modem type and would correctly return `CellularAccessTechnology` for us.

avtolstoy:
@keeramis Please rebase as well.

PR #2084, opened 2020-04-21
Gen3: fixes I2C hang up issue

XuGuohui:
### Problem

I2C might get hang up by slave device and never lease the bus.

### Solution

Reset I2C bus by simulating 9 clocks on the SCL line and reset the I2C state machine.

This PR also fixes the issue that Fuel Gauge returns 256 if it fails to read the SoC.

### Steps to Test

1. Build and flash the given example below for B5SoM or BSoM (SoM evaluation board is required).

2. Connect the SoM evaluation board to your computer and open a serial monitor to watch the log.

3. Disconnect and reconnect the `PM_SCL` jumper on the SoM evaluation board randomly, quickly and repeatedly.

4. See if the SoC is get updated as long as the `PM_SCL` jumper is connected. And if the `PM_SCL` is disconnected, the SoC should be `-1`.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

FuelGauge fuel;

void setup() {
}

void loop() {
    static system_tick_t startTime = millis();
    if (millis() - startTime >= 500) {
        startTime = millis();
        LOG(INFO, "SoC: %4.2f %%", fuel.getSoC());
    }
}
```

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -182,7 +182,6 @@ static void twim_handler(nrfx_twim_evt_t const * p_event, void * p_context) {
             m_i2c_map[inst_num].transfer_state = TRANSFER_STATE_ERROR_ADDRESS;
             break;
         }
-            break;
         case NRFX_TWIM_EVT_DATA_NACK: {
             LOG_DEBUG(TRACE, "NRFX_TWIM_EVT_DATA_NACK");
             m_i2c_map[inst_num].transfer_state = TRANSFER_STATE_ERROR_DATA;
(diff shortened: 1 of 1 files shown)


hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -391,19 +390,20 @@ int32_t HAL_I2C_Request_Data_Ex(HAL_I2C_Interface i2c, const HAL_I2C_Transmissio
         // FIXME: There is a bug in nrfx_twim driver, if we call nrfx_twim_rx repeatedly and quickly,
         // p_cb->busy will be set and never cleared, in this case nrfx_twim_rx always returns busy error
         LOG_DEBUG(TRACE, "BUSY ERROR, restore twi.");
-        twi_uninit(i2c);
```
avtolstoy, line 394: Do we still need to reinitialize I2C here?

wiring/src/spark_wiring_fuel.cpp
```diff
@@ -119,6 +119,9 @@ float FuelGauge::getSoC() {
 	byte LSB = 0;
 
 	readRegister(SOC_REGISTER, MSB, LSB);
+    if (MSB > 99 || LSB == 255) {
```
avtolstoy, line 122: Nit: formatting (spaces/tabs).

Also, this does not seem like a very reliable approach to track errors. If we are trying to fix this, `readRegister()` should be checking for errors and propagating them as well.

XuGuohui: That also makes sense. As `readRegister()` is private, we can change its prototype at our wish.

hal/src/nRF52840/i2c_hal.cpp
XuGuohui: `HAL_I2C_Reset(i2c, 0, nullptr);` will simulates 9 clocks on SCL followed by calling `HAL_I2C_Begin()`, which will reinitialize I2C.

avtolstoy: :+1:

wiring/src/spark_wiring_fuel.cpp
avtolstoy: We can also do this as a separate PR as the main thing we are trying to resolve here is an unrecoverable (without manual intervention) bus lockup after an error.

PR #2085, opened 2020-04-21
Particle.process() is a no-op when called from a custom thread.

keeramis:
…a custom thread

<details>

```

```
</details>

### Problem

If `Particle.process()` is called from custom thread, it runs `Spark_idle_events()`
(It's not safe is that Particle.process (and the underlying spark function) also handle things like serial event handlers, subscription handlers, function handlers, and (in 1.5.0 and later) variable queries. These are all expected to only be called from the loop context, otherwise you'd need thread safe constructs around them and in the loop code.)

### Solution

Particle.process() should be a no-op when called from a custom thread.

#### Test App that attempts to run Particle.process() from custom thread

```c
#include "application.h"

SYSTEM_THREAD(ENABLED);
void fn() {
    while (true) {
        Particle.process();
    }
}

void setup() {
    Thread* th = new Thread("name", fn, OS_THREAD_PRIORITY_DEFAULT);
}
void loop() {
}
```

### References

[ch38445](https://app.clubhouse.io/particle/story/38445/particle-process-should-be-a-no-op-when-called-from-a-custom-thread)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/thread.cpp
+++ b/user/tests/wiring/no_fixture/thread.cpp
@@ -1,8 +1,10 @@
 
 #include "application.h"
 #include "unit-test/unit-test.h"
+#include "scope_guard.h"
 
 #if PLATFORM_THREADING
+#include "system_threading.h"
 
 static uint32_t s_ram_free_before = 0;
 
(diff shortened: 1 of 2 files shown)


user/tests/wiring/no_fixture/thread.cpp
```diff
@@ -53,13 +54,142 @@ test(THREAD_04_with_lock)
 
 }
 
-test(THREAD_04_try_lock)
+test(THREAD_05_try_lock)
```
keeramis, line 57: the previous one was also `4`, so i renamed this to `5`

PR #2087, opened 2020-04-27
[electron] Fix "Sim not inserted" debug log when modem is not responsive

keeramis:
### Problem

Gen2 reports `SIM not inserted` or `Sim not inserted? Detecting...` when modem is not responsive. It is misleading because modem AT was not responsive and is confusing users to think that Sim is not inserted.

### Solution

Changed log message from `Sim not inserted` to `SIM/modem not responsive or SIM not inserted/requires a PIN.`

### Example App
May spoof AT not working by changing baud rate, or by connecting cell USB at the same time for testing purposes.
```c
#include "application.h"
void setup() {
}

void loop() {
}
```

### References

[ch50557](https://app.clubhouse.io/particle/story/50557/remove-sim-not-inserted-for-unresponsive-modem)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_network_cellular.h
+++ b/system/src/system_network_cellular.h
@@ -133,11 +133,11 @@ public:
     bool has_credentials() override
     {
         bool rv = cellular_sim_ready(NULL);
-        LOG(INFO,"%s", (rv)?"Sim Ready":"Sim not inserted? Detecting...");
+        LOG(INFO,"%s", (rv)?"Sim Ready":"SIM/modem not responsive or SIM not inserted/requires a PIN.");
         if (!rv) {
             cellular_on(NULL);
             rv = cellular_sim_ready(NULL);
-            LOG(INFO,"%s", (rv)?"Sim Ready":"Sim not inserted.");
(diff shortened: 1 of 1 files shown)


system/src/system_network_cellular.h
```diff
@@ -133,11 +133,15 @@ class CellularNetworkInterface : public ManagedIPNetworkInterface<CellularConfig
     bool has_credentials() override
     {
         bool rv = cellular_sim_ready(NULL);
-        LOG(INFO,"%s", (rv)?"Sim Ready":"Sim not inserted? Detecting...");
+        if (rv) {
+            LOG(INFO,"Sim Ready");
+        }
```
keeramis, line 138: Reporting only `Sim ready`. 
When sim is not ready and AT is working, it will be anyway reported here - https://github.com/particle-iot/device-os/blob/47bac6ae62977435096b33341aad0ee20d036ed8/hal/src/electron/modem/mdm_hal.cpp#L848

PR #2091, opened 2020-04-30
[hal] SPI: fixes potential deadlock on Gen3 platforms.

XuGuohui:
### Problem

Gen3 Device may get blocked when priority of the thread doing SPI transmission higher than the thread doing USB `Serial` logging.

### Solution

Check and handle SPI `NRF_SPIM_EVENT_END` while waiting SPI transaction to be completed.

### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

Thread *spi_thread;

__SPISettings spi_settings(5*MHZ, MSBFIRST, SPI_MODE0);

uint32_t spi_count = 0;

void spi_thread_f(void) {
    while (true) {
        uint8_t rx_buf[64];
        SPI.beginTransaction(spi_settings);
        SPI.transfer(NULL, rx_buf, sizeof(rx_buf), NULL);
        SPI.endTransaction();
        spi_count++;
        delay(1);
    }
}

void setup() {
    SPI.setDataMode(SPI_MODE0);
    SPI.setClockSpeed(5, MHZ);
    SPI.begin();

    Serial.begin(115200);
    while(!Serial.isConnected());

    spi_thread = new Thread("spiTest", spi_thread_f, OS_THREAD_PRIORITY_CRITICAL);

    pinMode(D7, OUTPUT);
}

void loop() {
    Log.info("heartbeat: %lu", spi_count);
    delay(1);

    static system_tick_t start = millis();
    static bool toggle = false;
    if (millis() - start >= 1000) {
        start = millis();
        if (toggle) {
            digitalWrite(D7, HIGH);
        } else {
            digitalWrite(D7, LOW);
        }
        toggle = !toggle;
    }
}
```

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/spi_hal.cpp
+++ b/hal/src/nRF52840/spi_hal.cpp
@@ -369,11 +369,17 @@ uint16_t HAL_SPI_Send_Receive_Data(HAL_SPI_Interface spi, uint16_t data) {
     tx_buffer = data;
 
     m_spi_map[spi].spi_dma_user_callback = NULL;
-    spi_tx_rx(spi, &tx_buffer, &rx_buffer, 1);
+    SPARK_ASSERT(spi_tx_rx(spi, &tx_buffer, &rx_buffer, 1) == 1);
 
     // Wait for SPI transfer finished
     while(m_spi_map[spi].transmitting) {
-        ;
+        // FIXME: This is a dirty hack to fix SPI race condition which is caused by invoking `CRITICAL_REGION_EXIT()` somewhere, e.g. USB stack.
+        if (nrf_spim_event_check(m_spi_map[spi].master->p_reg, NRF_SPIM_EVENT_END)) {
+            nrf_spim_event_clear(m_spi_map[spi].master->p_reg, NRF_SPIM_EVENT_END);
+            m_spi_map[spi].transmitting = false;
+            // Reset the transmission statue in nrfx_spim.c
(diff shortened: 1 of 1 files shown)


avtolstoy:
@XuGuohui Let's fix this properly as we've discussed previously by overriding `CRITICAL_REGION_ENTER()`/`EXIT()` to use `PRIMASK` instead of altering NVIC registers.

YutingYou:
> @XuGuohui Let's fix this properly as we've discussed previously by overriding `CRITICAL_REGION_ENTER()`/`EXIT()` to use `PRIMASK` instead of altering NVIC registers.

It might potentially affect BLE protocol which is based on the timer0 and other interrupts.

avtolstoy:
> It might potentially affect BLE protocol which is based on the timer0 and other interrupts.

Which will not get blocked. We are already using this approach throughout DeviceOS. See how `HAL_disable_irq()` is implemented for Gen 3.

XuGuohui:
>We may need to use priority 4 vs 2 here specifically though.

I still got a SOS when use priority 2 building against the latest develop branch. I tested on Argon.

XuGuohui:
We have another workaround that makes more sense to fix the issue. This PR will be closed and superseded by another PR when it is ready.

PR #2094, opened 2020-05-06
FuelGauge: return error if it failed to read/write registers.

XuGuohui:
### Problem

Returned value of FuelGauge's wiring APIs cannot indicate error on failed.

### Steps to Test

1. Build and flash the following example application for B5SoM that is attached on SoM evaluation board.
2. Connect the SoM evaluation board to serial terminal.
3. Detach the jumper for PM_SCL or PM_SDA and see if a negative value is printed on serial terminal.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup()
{
    Serial.begin(115200);
    while(!Serial.isConnected());
}

void loop() {
    FuelGauge gauge;
    LOG(TRACE, "SoC: %4.2f %%", gauge.getSoC());
    delay(1000);
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/src/spark_wiring_fuel.cpp
+++ b/wiring/src/spark_wiring_fuel.cpp
@@ -5,7 +5,7 @@
  * @version V1.0.0
  * @date    11-August-2015
  * @brief   driver for the fuel gauge IC  MAX17043
- 			MAX17043 datasheet: http://datasheets.maximintegrated.com/en/ds/MAX17043-MAX17044.pdf
+            MAX17043 datasheet: http://datasheets.maximintegrated.com/en/ds/MAX17043-MAX17044.pdf
  ******************************************************************************
   Copyright (c) 2013-2015 Particle Industries, Inc.  All rights reserved.
 
@@ -28,6 +28,7 @@
 
 #include <mutex>
 #include "spark_wiring_power.h"
+#include "check.h"
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


wiring/src/spark_wiring_fuel.cpp
```diff
@@ -108,7 +108,7 @@ float FuelGauge::getVCell() {
 	byte MSB = 0;
 	byte LSB = 0;
 
-	readRegister(VCELL_REGISTER, MSB, LSB);
+	CHECK(readRegister(VCELL_REGISTER, MSB, LSB));
```
avtolstoy, line 111: Returning numeric errors codes through a method that normally returns a float seems strange. IMO we should return some fixed documented value not overlapping with normal range on errors.

XuGuohui: Sounds good. 👍

wiring/inc/spark_wiring_i2c.h
```diff
@@ -105,9 +105,9 @@ class TwoWire : public Stream
   void end();
   uint8_t endTransmission(void);
   uint8_t endTransmission(uint8_t);
-  size_t requestFrom(uint8_t, size_t);
-  size_t requestFrom(uint8_t, size_t, uint8_t);
-  size_t requestFrom(const WireTransmission& transfer);
+  ssize_t requestFrom(uint8_t, size_t);
```
avtolstoy, line 108: I understand the idea, but this is a breaking change. The documentation states:
> Returns: byte : the number of bytes returned from the slave device. If a timeout occurs, will return 0.

Do we need to change the API here? It should still be possible to capture errors by checking for `== 0`.

hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -673,7 +689,7 @@ int32_t HAL_I2C_Request_Data_Ex(HAL_I2C_Interface i2c, const HAL_I2C_Transmissio
             /* SW Reset the I2C Peripheral */
             HAL_I2C_SoftwareReset(i2c);
             HAL_I2C_Release(i2c, NULL);
-            return 0;
+            return SYSTEM_ERROR_TIMEOUT;
```
avtolstoy, line 692: These changes will also break backwards compatibility. Old application and new DeviceOS will start returning absurdly large sizes.

XuGuohui: Given that the return type is `int32_t`, returning a negative value indicating an error wouldn't break the size range. Does it make sense? Or could you help me understand how it break the backwards compatibility.

avtolstoy: Sure, but what about `TwoWire::requestFrom()` ?

avtolstoy: Its _current_ implementation is just this:
```
size_t TwoWire::requestFrom(const WireTransmission& transfer) {
    auto conf = transfer.halConfig();
    return HAL_I2C_Request_Data_Ex(_i2c, &conf, nullptr);
}
```

And before the latest changes it was even worse:
```
  uint8_t result = HAL_I2C_Request_Data(_i2c, address, quantity, sendStop, NULL);
  return result;
```

A negative value would be cast into a `uint8_t` or `size_t` resulting in unexpected values for the application.

Again, this will happen for cases where the application was built for older DeviceOS and you have a system part running with these changes where we now return negative error codes.

XuGuohui: I just committed to revert the changes in spark_wiring_i2c.h and made proper change in spark_wiring_i2c.cpp. Kindly refer to the Files changed.

avtolstoy: This will not help existing applications built with old DeviceOS versions. We cannot change how things work in those.

Issue #1562, opened 2018-07-28
Incompatibility between WLanSecurityType and SecurityType


ScruffR:
As discussed in this thread 
https://community.particle.io/t/setsecurity-doesnt-seem-to-be-working-compiling/43226

There is an incompatiblity between `WLanSecurityType` and `SecurityType` which causes the documented syntax to fail
https://docs.particle.io/reference/firmware/photon/#setsecurity-
```
// SYNTAX
WiFiCredentials& WiFiCredentials::setSecurity(SecurityType security);
// EXAMPLE - setting security type
WiFiCredentials credentials;
credentials.setSecurity(WPA2);
```
*(tested with 0.7.0 and 0.8.0-rc.8)*

But this works
```
  WiFiCredentials credentials;
  credentials.setSecurity(WLAN_SEC_WPA2);
```

avtolstoy:
Resolved in #2098

PR #2098, opened 2020-05-11
Add a new member to differentiate internal and wiring setSecurity

keeramis:
</details>

### Problem

As per documentation,` setSecurity()` should be given a value belonging to `SecurityType` enum, but currently it expects a `WLANSecurityType` in the code, and hence results in compile error for `WLAN.setSecurity(WPA2)`

### Solution

Currently, the argument for `setSecurity()` is `WLANSecurityType` which throws the compile error. Changing the argument to `SecurityType` fixes it but breaks other internal code that passes `WLANSecurityType` to the `setSecurity()`. Hence, changed the `setSecurity()`'s argument to `SecurityType` for wiring and added another member `setWlanSecurity()` for internal use. 

### The following should compile successfully on Wifi platforms

```c
#include "application.h"

void setup() {
  WiFiCredentials credentials;
  credentials.setSecurity(WPA2);
}

void loop() {
}
```

### References

- [ch50408](https://app.clubhouse.io/particle/story/50408/incompatibility-between-wlansecuritytype-and-securitytype)
- https://github.com/particle-iot/device-os/issues/1562

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_wifi_credentials.h
+++ b/wiring/inc/spark_wiring_wifi_credentials.h
@@ -59,17 +59,17 @@ enum SecurityType {
 class WiFiCredentials {
 public:
     WiFiCredentials(SecurityType security = UNSEC) {
-        setSecurity((WLanSecurityType)security);
+        setSecurity(security);
     }
 
     WiFiCredentials(const char* ssid, SecurityType security = UNSEC) {
-        setSecurity((WLanSecurityType)security);
+        setSecurity(security);
         setSsid(ssid);
     }
 
     virtual ~WiFiCredentials() {}
 
-    virtual WiFiCredentials& setSecurity(WLanSecurityType security) {
+    virtual WiFiCredentials& setSecurity(SecurityType security) {
         creds_.security = (WLanSecurityType)security;
         return *this;
     }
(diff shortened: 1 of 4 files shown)


keeramis:
@avtolstoy May be we should add unit tests for this, do you think?

wiring/inc/spark_wiring_wifi_credentials.h
```diff
@@ -59,17 +59,17 @@ enum SecurityType {
 class WiFiCredentials {
 public:
     WiFiCredentials(SecurityType security = UNSEC) {
-        setSecurity((WLanSecurityType)security);
+        setSecurity((SecurityType)security);
```
avtolstoy, line 62: Why do we need a cast here and the other places? It's the same type.

keeramis: That's right - the places in this file can be gotten rid of. 
The other couple places fails compiling because 
1. `toSecurity` returns a `WLANSecurityType`. (https://github.com/particle-iot/device-os/blob/27a0177608ef33014bf9a02489714fb8f5adcc57/hal/src/photon/softap.cpp#L638)
2. This as well `security_` is of `WLANSecurityType`

avtolstoy: That one is correct :+1:

avtolstoy:
@keeramis Please rebase and merge.

PR #2099, opened 2020-05-11
Update the Particle server hostname wildcard to include the Device OS major version [ch53190]

technobly:
### Problem

Currently the default particle server key for UDP devices is `$id.udp.particle.io` and `$id.udp-mesh.particle.io`. The wildcard `$id` is expanded to the device ID. In order to support LTS and to be able to route all devices on a particular major version of Device OS to a specific Device Service, expand the wildcard `$id` to the `<device_id>.v<device_os_major_version>` so for a device `1234` at `Device OS v2.1.0` it would connect to `1234.v2.udp.particle.io`

### Solution

If appropriate, the public server key is updated after loaded with the major version from v0 ~ v255.

### Steps to Test

- Unit tests have been added and can be run
- UDP devices can be tested and should continue to connect appropriately now, and can also be routed in the future as required.
- Tested these devices, all connect:
```
Electron
0000007381 [system] INFO: Resolved 123412341234123412341234.v1.udp.particle.io to 12.34.567.890

Argon
0000019412 [system] TRACE: Resolving 123412341234123412341234.v1.udp-mesh.particle.io#5684

B5 SoM
0000021277 [system] TRACE: Resolving 123412341234123412341234.v1.udp-mesh.particle.io#5684
```

### References

Closes ch53190

---

### Completeness

### DIFF
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -55,6 +55,7 @@
 #include "network/ncp/cellular/ncp.h"
 #include "network/ncp/cellular/cellular_ncp_client.h"
 #endif // HAL_PLATFORM_MUXER_MAY_NEED_DELAY_IN_TX
+#include "system_version.h"
 
 #include <stdio.h>
 #include <stdint.h>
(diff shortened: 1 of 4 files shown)


technobly:
I'll take a look at those functions you mentioned to see if they are more ideal, and as easily testable.  Most of the complexity here is just safety checks to prevent buffer overflow and unintended results, and tests :D

avtolstoy:
There are already unit tests for that functionality: `user/tests/unit/string_interpolate.cpp`

monkbroc:
I'll let you take another stab at it before I review Brett.

technobly:
@avtolstoy @monkbroc please review with https://github.com/particle-iot/device-os/pull/2099/files?w=1

system/src/system_cloud_internal.cpp
```diff
@@ -627,15 +628,24 @@ void Spark_Signal(bool on, unsigned, void*)
...
-	return 0;
+    if (var_len==2 && memcmp("id", var, 2)==0)
+    {
+        String deviceID = spark_deviceID();
+        size_t id_len = deviceID.length();
+
+        SystemVersionInfo sys_ver;
+        system_version_info(&sys_ver, nullptr);
+        uint8_t mv = BYTE_N(sys_ver.versionNumber, 3);
+        String majorVer = String::format(".v%d", mv);
```
avtolstoy, line 639: Nit: ideally we shouldn't use `wiring` within DeviceOS whenever possible.

user/tests/unit/string_interpolate.cpp
```diff
@@ -1,78 +1,408 @@
-#include "catch.hpp"
-#include <string.h>
+/**
+ ******************************************************************************
```
avtolstoy, line 2: Nit: old style header.

system/src/system_cloud_internal.cpp
technobly: I considered replacing what was there already, but then thought... hey this makes sure we don't break String :D and probably 5 less complaints from Codacy about strlen() and strncpy().

avtolstoy: > hey this makes sure we don't break String

We have tests for that :smirk: 

>  5 less complaints from Codacy about strlen() and strncpy().

Just ignore them, those specifically are annoying and don't bring any value.

PR #2101, opened 2020-05-13
[Gen3] USB bugfixes

avtolstoy:
### Problem

1. USB serial port state is not being reliably detected on Gen 3 devices. This easily causes a deadlock when writing data into USB serial port when there is no other side to read out the data
2. USB control requests fail sometimes partially due to ^^, there is also another issue whether extra control requests are forwarded

### Solution

1. Refactors USB implementation for Gen 3. Utilize SOF for detecting that the other side is not reading data out and consider the connection to the host broken. This is similar (but not quite) to how we do things on Gen 2. The difference is that we are not marking the serial port is closed and simply abort TX transfer and flush 
2. Filter out non-vendor requests

This PR also changes the implementation of the critical sections for Nordic SDK to use BASEPRI. This change supersedes and fixes the issue mentioned in #2091.

### Steps to Test

1. Make the device generate some data on Serial e.g. by using SerialLogHandler. Open the serial port, close it, wait for a while for the device to generate some more data. Attempt to run a USB control request - it should not timeout.
2. Validate the test from https://github.com/particle-iot/device-os/pull/2091

### Example App

N/A

### References

- [CH42073]
- [CH50902]

---

### Completeness

- [ ] Added documentation


busticated:
this story also seem relevant 👉 https://app.clubhouse.io/particle/story/53189/error-libusb-transfer-timed-out-after-waking-host

@avtolstoy can you confirm your fixes will address this story?

avtolstoy:
@busticated Suspend/Resume events seem to still be incorrectly handled. I'll take a look into that as well. Thanks for linking to that story.

YutingYou:
LGTM. 

For USB timeout issue, I tried with `particle usb dfu`, sometimes it failed probably due to an unexpected reset to break the USB  transaction, but tried with `particle USB listen`(no reset occurs), it worked fine for me.

I also ran a stress test to keep sending data via USB serial port to Mac serial software `screen`. No byte is missing. But tried with Windows system (installed in the virtual Machine Parallels), it has byte missing, I'm not sure the issue is with Virtual Machine or Windows USB driver or Serial Software.

image

avtolstoy:
> For USB timeout issue, I tried with particle usb dfu, sometimes it failed probably due to an unexpected reset to break the USB transaction,

As long as the device enters DFU mode, it's fine for the purposes of this PR.

> it has byte missing, I'm not sure the issue is with Virtual Machine or Windows USB driver or Serial Software.

Probably not reading data fast enough out of the serial port (over 1s delay). We can't do anything sane about it unfortunately.

I'll add a few more fixes to the PR to make it more aligned with Gen 3 implementation. OUT (host->device) transfer should not be cancelled in `usb_cdc_change_port_state()` and the rx buffer should not be flushed.

avtolstoy:
Suspend/resume should also be fixed.

PR #2103, opened 2020-05-15
GCC 9 support

avtolstoy:
### Description

Resolves various issues/warnings encountered when building with GCC9, switches over the CI to using GCC9.

Updates Workbench manifest with GCC 9 support as well.

### Steps to Test

N/A

### Example App

N/A

### References

- [CH42647]

---

### Completeness

- [ ] Added documentation


bootloader/src/main.c
```diff
@@ -466,9 +464,10 @@ int main(void)
 
             Reset_System();
 
-            // Initialize user application's Stack Pointer
-            __set_MSP(*(__IO uint32_t*) ApplicationAddress);
-            Jump_To_Application();
+            // Jump to system firmware
+            uint32_t addr = *(volatile uint32_t*)(ApplicationAddress + 4);
+            uint32_t stack = *(volatile uint32_t*)ApplicationAddress;
+            jump_to_system(addr, stack);
```
avtolstoy, line 470: For reference: this is required because GCC9 now throws a deprecated error when calling CMSIS `__set_MSP()`, which clobbers `sp`, which is in fact not what is supposed to happen.

hal/src/electron/core_hal.c
```diff
@@ -42,9 +42,9 @@
 /**
  * Start of interrupt vector table.
  */
-extern char link_interrupt_vectors_location;
-extern char link_ram_interrupt_vectors_location;
-extern char link_ram_interrupt_vectors_location_end;
+extern uintptr_t link_interrupt_vectors_location[];
```
avtolstoy, line 45: For reference: this is now required because of array bounds checks warnings in GCC9.

wiring/src/spark_wiring_string.cpp
```diff
@@ -830,9 +830,3 @@ String String::format(const char* fmt, ...)
     }
     return result;
 }
-
-std::ostream& operator << ( std::ostream& os, const String& value ) {
```
avtolstoy, line 834: For reference: this was a source of a major binary size bloat due to the usage of iostreams, which was bringing std strings, locales, exceptions and system_error stuff into the binary no matter what.

wiring/inc/spark_wiring_print.h
```diff
@@ -127,12 +127,20 @@ size_t Print::print(T n, int base)
         size_t t = 0;
         using PrintNumberType = typename PrintNumberTypeSelector<T>::type;
         PrintNumberType val;
+// FIXME: avoids 'comparison of constant '0' with boolean expression is always false'
+#if __GNUC__ >= 9
```
avtolstoy, line 131: Adding a separate `bool` specialization brings its own set of problems, so for now we are disabling this warning here specifically as a workaround.

bootloader/src/main.c
avtolstoy: A quote from recent gcc docs:

> Another restriction is that the clobber list should not contain the stack pointer register. This is because the compiler requires the value of the stack pointer to be the same after an asm statement as it was on entry to the statement. However, previous versions of GCC did not enforce this rule and allowed the stack pointer to appear in the list, with unclear semantics. This behavior is deprecated and listing the stack pointer may become an error in future versions of GCC.

wiring/src/spark_wiring_string.cpp
monkbroc: Is this a breaking change that we ought to announce? What would be typical usage? Something like `Serial << String("foo")` ? Once that's clear I can check with Studios if they use this pattern in any application.

avtolstoy: This is not a breaking change. DeviceOS does not support iostreams in any form. I have no explanation why that code was there if only for unit tests. For the example you listed to work `Serial` would have to be a subclass of `std::ostream`, which it isn't and none of the other wiring classes are.

elcojacobs:
I'm very happy that the compiler has finally been upgraded!
Just compiled against 2.0.0-rc.1 and apart from the simple api deprecations changes, there seem to be no issues for my huge application.

Now that we run on gcc 9, can the makefiles for a local build be upgraded to use C++17 too? The still use std=gnu++14.

avtolstoy:
We aren't changing the C++ standard version for 2.x. We'll migrate most likely to C++20 in 3.x.

PR #2105, opened 2020-05-15
[Boron] Radio unresponsive when sending large data or application desc msgs

keeramis:
### Problem

Boron LTE-M1 cell radio becomes unresponsive (by getting stuck at an AT command) after "large" data or application describe message is sent continuously over muxer data channel

### Solution

This is a callback that writes data into muxer channel 2 (data PPP channel). Whenever we encounter a large packet, we enforce a certain number of ms to pass before transmitting anything else on this channel. After we send large packet of size 250, we drop messages(bytes) for a certain amount of time defined by UBLOX_NCP_R4_WINDOW_SIZE_MS (50ms)

### Steps to Test

### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

void event_handler(const char *event, const char *data)
{
}

int func_handler(String data)
{
    return 0;
}

void setup()
{
    static int dummy_int = 0;
    static String dummy_string("abc");

    waitFor(Serial.isConnected, 10000);

    Particle.function("CMD", func_handler);

    Particle.variable("HEARTBEAT", &dummy_int, INT);
    Particle.variable("HW_VER", &dummy_string,   STRING);
    Particle.variable("SELFTEST", &dummy_string, STRING);
    Particle.variable("CONFIG", &dummy_string,   STRING);
    Particle.variable("DEBUG", &dummy_string,    STRING);
    Particle.variable("ERROR", &dummy_string,    STRING);
    Particle.variable("ALARM", &dummy_string,    STRING);;

    Particle.subscribe(System.deviceID() + "/hook-response/", event_handler, MY_DEVICES);
    Particle.subscribe(System.deviceID() + "/api_get_fw/", event_handler, MY_DEVICES);
    Particle.subscribe("particle/device/name", event_handler, MY_DEVICES);
    Particle.connect();
}

void loop()
{
    static uint32_t last_sec = System.uptime();

    if(System.uptime() != last_sec)
    {
        last_sec = System.uptime();
        Log.info("heartbeat");
    }
}
```

### References

[ch53040](https://app.clubhouse.io/particle/story/53040/boron-lte-stuck-rapidly-breathing-cyan)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/boron/network/sara_ncp_client.cpp
+++ b/hal/src/boron/network/sara_ncp_client.cpp
@@ -110,8 +110,9 @@ const unsigned REGISTRATION_TIMEOUT = 10 * 60 * 1000;
 using LacType = decltype(CellularGlobalIdentity::location_area_code);
 using CidType = decltype(CellularGlobalIdentity::cell_id);
 
-const system_tick_t UBLOX_NCP_R4_LARGE_PACKET_TIMEOUT_MS = 250;
-const size_t UBLOX_NCP_R4_LARGE_PACKET_THRESHOLD_SIZE = 512;
+const size_t UBLOX_NCP_R4_BYTES_PER_WINDOW_THRESHOLD = 512;
+const system_tick_t UBLOX_NCP_R4_WINDOW_SIZE_MS = 50;
+
 
 } // anonymous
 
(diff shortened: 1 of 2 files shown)


hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -110,8 +110,9 @@ const unsigned REGISTRATION_TIMEOUT = 10 * 60 * 1000;
 using LacType = decltype(CellularGlobalIdentity::location_area_code);
 using CidType = decltype(CellularGlobalIdentity::cell_id);
 
-const system_tick_t UBLOX_NCP_R4_LARGE_PACKET_TIMEOUT_MS = 250;
-const size_t UBLOX_NCP_R4_LARGE_PACKET_THRESHOLD_SIZE = 512;
+const size_t UBLOX_NCP_R4_BYTES_PER_WINDOW_THRESHOLD = 512;
+const system_tick_t UBLOX_NCP_R4_WINDOW_SIZE_MS = 50;
```
keeramis, line 114: Tested that it could go down to `50ms` but cannot go higher than `512bytes`

keeramis: I have not tested lesser than `50ms` thinking that `50ms` is probably safer than something like `10ms`?

PR #2102, opened 2020-05-14
[Gen 3] Modem warm bootup

avtolstoy:
### Problem

Gen 3 cellular devices currently always power off and power the modem back on on boot. This affects the boot/cloud connection times significantly especially on platforms where poweron/poweroff times are noticeable.

### Solution

Implement support for warm bootup!

### Steps to Test

N/A

### Example App

N/A

### References

- [CH50404]

---

### Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [x] Problem and Solution clearly stated
- [ ] Run unit/integration/application tests on device
- [ ] Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

PR #2107, opened 2020-05-19
[boron] fixes SARA R4 power on sequence in light of recent baudrate c…

avtolstoy:
### Problem

When changing the default baudrate to 460800 on LTE Borons/BSoMs we didn't account for some power on conditions in https://github.com/particle-iot/device-os/pull/2102

### Solution

Fixes that.

### Steps to Test

- Verify initialization from modem power on works on 02.00 and 02.04

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


keeramis:
Do we want to make this change on Quectel to ensure consistency? (Although quectel as of now does not have baud rate persistent setting, is it?)

avtolstoy:
It should be fine to have different platform-specific behavior.

PR #2110, opened 2020-05-24
Fix the power leakage issue when putting Cellular device …

XuGuohui:
### Problem

Cellular modem is not completely turned off when device is going into sleep mode.

### Solution

Wait until the Cellular modem is turned off reliably before device entering sleep mode.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

SYSTEM_THREAD(ENABLED);

void setup() {
    pinMode(A1, INPUT_PULLUP);

    Cellular.on();
}

void loop() {
    if (digitalRead(A1) == 0) {
        LOG(TRACE, "Device is going to enter sleep mode.");
        config.mode(SystemSleepMode::HIBERNATE).gpio(A0, RISING);
        System.sleep(config);

        Cellular.on();
    }
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/network/api/ifapi.h
```diff
@@ -254,6 +254,7 @@ if_event_handler_cookie_t if_event_handler_self(if_t iface, if_event_handler_t h
 int if_event_handler_del(if_event_handler_cookie_t cookie);
 
 int if_request(if_t iface, int type, void* req, size_t reqsize, void* reserved);
+void if_notify_events(if_t iface, const struct if_event* evt, void* reserved);
```
avtolstoy, line 257: Nit: `notify_event`?

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -254,6 +254,21 @@ void PppNcpNetif::ncpEventHandlerCb(const NcpEvent& ev, void* ctx) {
...
+        const auto& cev = static_cast<const NcpStateChangedEvent&>(ev);
+        if_event evt = {};
+        struct if_event_power_state ev_if_power_state = {};
+        evt.ev_len = sizeof(if_event);
+        evt.ev_type = IF_EVENT_POWER_STATE;
+        evt.ev_power_state = &ev_if_power_state;
+        if (cev.state == NcpState::ON) {
+            evt.ev_power_state->state = IF_POWER_STATE_UP;
+            LOG(TRACE, "NCP power state changed: IF_POWER_STATE_UP");
+        } else {
```
avtolstoy, line 267: If we take a look at `off()` implementation, we can see that it's still possible that the NCP client will go into an off state even if the modem fails to power down.

system/inc/system_network.h
```diff
@@ -102,6 +102,8 @@ void network_setup(network_handle_t network, uint32_t flags, void* reserved);
 int network_set_hostname(network_handle_t network, uint32_t flags, const char* hostname, void* reserved);
 int network_get_hostname(network_handle_t network, uint32_t flags, char* buffer, size_t buffer_len, void* reserved);
 
+void network_wait_modem_off(network_handle_t network, system_tick_t timeout, void*);
```
avtolstoy, line 105: Not implemented for Gen 2, CI fails.

system/src/system_network_manager.h
```diff
@@ -112,6 +114,7 @@ class NetworkManager {
         if_t iface = nullptr;
         std::atomic<ProtocolState> ip4State;
         std::atomic<ProtocolState> ip6State;
+        bool on = false;
```
avtolstoy, line 117: Needs to be volatile/atomic.

system/src/system_sleep.cpp
```diff
@@ -89,6 +89,8 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
         if (system_sleep_network_suspend(NETWORK_INTERFACE_CELLULAR)) {
             cellularResume = true;
         }
+        // There might be up to 30s delay to turn off the modem for particular platforms.
+        network_wait_modem_off(NETWORK_INTERFACE_CELLULAR, 60000/*ms*/, nullptr);
```
avtolstoy, line 93: This will affect all sleep timings when using STOP mode with default settings.

1. Let's discuss more if we can potentially do this in a manner which does not "break" things
2. We should look into potentially improving power off timings e.g. by using `AT+CPWROFF` or others

system/src/system_network_manager_api.cpp
```diff
@@ -308,6 +308,21 @@ void network_off(network_handle_t network, uint32_t flags, uint32_t param, void*
     }());
 }
 
+void network_wait_modem_off(network_handle_t network, system_tick_t timeout, void*) {
+    SYSTEM_THREAD_CONTEXT_ASYNC_CALL([&]() {
```
avtolstoy, line 312: This does not make sense. A wait cannot be async. Since this is an internal function, we can skip this whole system thread marshalling.

system/src/system_sleep.cpp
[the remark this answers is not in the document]
avtolstoy: system sleep functions are already executed from the system thread. See https://github.com/particle-iot/device-os/pull/2110#discussion_r429725385 as well.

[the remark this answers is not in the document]
YutingYou: I saw it here 👇
```
system_sleep_ext() {
    ...
    network_wait_modem_off();
    ...
    hal_sleep_enter();
    ...
}
```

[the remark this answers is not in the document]
avtolstoy: Yeah, but:

```
int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t** reason, void* reserved) {
    SYSTEM_THREAD_CONTEXT_SYNC(system_sleep_ext(config, reason, reserved));
```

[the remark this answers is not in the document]
YutingYou: Ah, I see 👍

XuGuohui: I didn't understand that why it breaks the sleep timing, especially for the STOP mode. I think if the device is requested to enter STOP mode without specifying Cellular interface as wake-up source, we should also wait util the modem is completely turned off.

avtolstoy: Which might be up to tens of seconds on certain platforms. So, the actual sleep will now suddenly be postponed by that amount of time after we introduce this change. I would consider this a change in behavior.

hal/network/lwip/cellular/pppncpnetif.cpp
XuGuohui: Can `modemHardReset(true)` turn off the modem reliably? I do see it is invoked if modem fails to be powered off:
```
        int rv = modemPowerOff();
        if (rv != 0) {
            modemHardReset(true);
        }
        ncpState(NcpState::OFF);
```

avtolstoy: Supposedly, however ideally we should still check the current modem state (`modemPowerState()`).

avtolstoy: On boot with `SYSTEM_MODE(AUTOMATIC)` or `MANUAL` the NCP client will also default to `OFF` state without reflecting the actual power state of the modem.

XuGuohui: But in `system_power_manager`, the `on` of `InterfaceRuntimeState` reflect the initial state of modem power and only when `SaraNcpClient::ncpState(NcpState state)` is called the `on` variable will be updated, despite the current value of `ncpState_` in `sara_ncp_client.cpp`.

avtolstoy: Sure, and by default it would be assumed that the modem is off, where in fact it might be on.

XuGuohui: Then we should initialize `ncpState_` accordingly in certain init function by checking the modem state, either through AT command or reading `UBVINT`.

system/src/system_network_manager.h
```diff
@@ -105,13 +107,15 @@ class NetworkManager {
...
-                  ip6State(ProtocolState::UNCONFIGURED) {
+                  ip6State(ProtocolState::UNCONFIGURED),
+                  on(false) {
         }
         InterfaceRuntimeState* next = nullptr;
         bool enabled = false;
         if_t iface = nullptr;
         std::atomic<ProtocolState> ip4State;
         std::atomic<ProtocolState> ip6State;
+        volatile std::atomic_bool on;
```
avtolstoy, line 118: either `volatile bool` or `std::atomic_bool`. It can't be both.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -254,6 +254,24 @@ void PppNcpNetif::ncpEventHandlerCb(const NcpEvent& ev, void* ctx) {
...
+        evt.ev_type = IF_EVENT_POWER_STATE;
+        evt.ev_power_state = &ev_if_power_state;
+        if (cev.state == NcpPowerState::ON) {
+            evt.ev_power_state->state = IF_POWER_STATE_UP;
+            LOG(TRACE, "NCP power state changed: IF_POWER_STATE_UP");
+        } else if (cev.state == NcpPowerState::OFF) {
+            evt.ev_power_state->state = IF_POWER_STATE_DOWN;
+            LOG(TRACE, "NCP power state changed: IF_POWER_STATE_DOWN");
+        } else {
+            evt.ev_power_state->state = IF_POWER_STATE_NONE;
```
avtolstoy, line 271: Should we be generating such an event? What does it mean?

hal/src/argon/network/esp32_ncp_client.cpp
```diff
@@ -602,6 +609,17 @@ void Esp32NcpClient::ncpState(NcpState state) {
     }
 }
 
+void SaraNcpClient::ncpPowerState(NcpPowerState state) {
+    pwrState_ = state;
```
avtolstoy, line 613: Skip event generation if `pwrState_ == state`?

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -1383,6 +1404,22 @@ int QuectelNcpClient::modemPowerOff() const {
     return SYSTEM_ERROR_NONE;
 }
 
+int QuectelNcpClient::modemSoftPowerOff() {
+    if (modemPowerState()) {
+        LOG(TRACE, "Powering modem off using AT command");
```
avtolstoy, line 1409: We should probably check for `ready_` here before attempting to execute AT commands. `parser_.execCommand()` will fail anyway, but it might be cleaner to even not attempt to do that.

avtolstoy: Same for SaraNcpClient.

hal/src/argon/network/esp32_ncp_client.cpp
avtolstoy: Same for QuectelNcpClient.

system/inc/system_sleep_configuration.h
```diff
@@ -71,16 +79,28 @@ class SystemSleepConfigurationHelper {
...
     }
 
-    particle::EnumFlags<SystemSleepFlag> sleepFlags() const {
-        return particle::EnumFlags<SystemSleepFlag>::fromUnderlying(config_->flags);
+    EnumFlags<SystemSleepNetworkFlag> networkFlags(network_interface_index index) const {
+        auto wakeup = wakeupSourceFeatured(HAL_WAKEUP_SOURCE_TYPE_NETWORK);
+        while (wakeup) {
+            auto networkWakeup = reinterpret_cast<const hal_wakeup_source_network_t*>(wakeup);
+            if (networkWakeup->index == index) {
+                return EnumFlags<SystemSleepNetworkFlag>::fromUnderlying(networkWakeup->flags);
```
avtolstoy, line 95: Struct size needs to be checked. This new field may not be present for applications built with older DeviceOS versions.

hal/src/b5som/network/quectel_ncp_client.cpp
```diff
@@ -1383,6 +1404,22 @@ int QuectelNcpClient::modemPowerOff() const {
...
+        int r = CHECK_PARSER(parser_.execCommand("AT+CPWROFF"));
```
avtolstoy, line 1410: There is no `AT+CPWROFF` on Quectel modems. `AT+QPOWD`.

hal/network/lwip/cellular/pppncpnetif.cpp
XuGuohui: We should notify the system power manager on startup to sync the modem state. On ESP32, we don't have a pin to indicate the modem state though.

avtolstoy: We do not use this for non-cellular interfaces though, right? ESP32 will also always be turned off on boot (see `esp32ncpnetif.cpp`).

system/inc/system_sleep_configuration.h
```diff
@@ -88,6 +88,9 @@ class SystemSleepConfigurationHelper {
     }
 
     EnumFlags<SystemSleepNetworkFlag> networkFlags(network_interface_index index) const {
+        if (config_->version < HAL_SLEEP_VERSION) {
```
avtolstoy, line 91: Needs to be a fixed version when this feature appeared in, otherwise once we bump it again it will fail again.

system/src/system_network_manager.cpp
```diff
@@ -812,6 +830,25 @@ void NetworkManager::resetInterfaceProtocolState(if_t iface) {
     }
 }
 
+int NetworkManager::waitModemOff(if_t iface, system_tick_t timeout) const {
```
avtolstoy, line 833: Let's generalize this. This is waiting for interface to turn off, not modem specifically.

system/src/system_network_manager_api.cpp
```diff
@@ -308,6 +308,20 @@ void network_off(network_handle_t network, uint32_t flags, uint32_t param, void*
     }());
 }
 
+int network_wait_modem_off(network_handle_t network, system_tick_t timeout, void*) {
```
avtolstoy, line 311: Same here: needs to be generalized. This is not about the modem specifically.

system/src/system_sleep.cpp
```diff
@@ -86,8 +88,15 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
...
-            cellularResume = true;
+        if (configHelper.networkFlags(NETWORK_INTERFACE_CELLULAR).isSet(SystemSleepNetworkFlag::INACTIVE_STANDBY)) {
+            // Pause the modem Serial, while leaving the modem keeps running.
+            cellular_pause(nullptr);
+        } else {
+            if (system_sleep_network_suspend(NETWORK_INTERFACE_CELLULAR)) {
+                cellularResume = true;
+            }
+            // There might be up to 30s delay to turn off the modem for particular platforms.
+            CHECK(network_wait_modem_off(NETWORK_INTERFACE_CELLULAR, 60000/*ms*/, nullptr));
```
avtolstoy, line 99: It should also probably be fine to add this for WiFi on Gen 3, right?

system/inc/system_sleep_configuration.h
XuGuohui: Good catch! 👍

system/src/system_sleep.cpp
XuGuohui: Yes. But we should re-test the power consumption just in case.

hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -420,6 +420,11 @@ int WizNetif::powerDown() {
     return down();
 }
 
+int WizNetif::getPowerState(if_power_state_t* state) const {
+    // TODO: implement it
+    return 0;
```
avtolstoy, line 425: Should remove an error anyway.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -318,10 +319,13 @@ int SaraNcpClient::off() {
...
     }
     muxer_.stop();
-    // Disable voltage translator
-    modemSetUartState(false);
-    // Power down
-    modemPowerOff();
+    // Try using AT command to turn off the modem first.
+    if (modemSoftPowerOff() != 0) {
+        // Disable voltage translator
+        modemSetUartState(false);
```
avtolstoy, line 325: Even with soft power off you need to disable it, otherwise you'll be leaking current through it.

user/src/application.cpp
```diff
@@ -1,30 +1,26 @@
...
+
+SYSTEM_THREAD(ENABLED);
+
+Serial1LogHandler log(LOG_LEVEL_ALL);
 
-/* This function loops forever -----------------------------------------------*/
-void loop()
-{
-    // This will run in a loop
+void setup() {
```
avtolstoy, line 9: Let's also please not modify the default blank app.

system/inc/system_sleep_configuration.h
```diff
@@ -71,16 +79,31 @@ class SystemSleepConfigurationHelper {
...
             }
             wakeup = wakeupSourceFeatured(HAL_WAKEUP_SOURCE_TYPE_NETWORK, wakeup->next);
         }
         return false;
     }
 
-    particle::EnumFlags<SystemSleepFlag> sleepFlags() const {
-        return particle::EnumFlags<SystemSleepFlag>::fromUnderlying(config_->flags);
+    EnumFlags<SystemSleepNetworkFlag> networkFlags(network_interface_index index) const {
+        if (config_->version < 3) {
```
avtolstoy, line 91: Magic number. This needs to be a define somewhere.

hal/src/boron/network/sara_ncp_client.cpp
```diff
@@ -1697,6 +1721,31 @@ int SaraNcpClient::modemPowerOff() {
...
+        if (ready_) {
+            int r = CHECK_PARSER(parser_.execCommand("AT+CPWROFF"));
+            if (r == AtResponse::OK) {
+                LOG(TRACE, "Modem is powered off");
+                // Disable voltage translator
+                modemSetUartState(false);
+                // WARN: We assume that the modem can turn off itself reliably.
+                ncpPowerState(NcpPowerState::OFF);
+            } else {
+                LOG(TRACE, "AT+CPWROFF command is not echoed");
```
avtolstoy, line 1736: Echo in the context of AT commands is something different. Let's make this clearer. No response?

hal/src/boron/network/sara_ncp_client.cpp
XuGuohui: It does if sending the AT command successfully.

user/src/application.cpp
XuGuohui: This was going to be an intermediate commit for testing purpose. I'll create a separated application then.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -89,7 +89,7 @@ inline system_tick_t millis() {
 }
 
 const auto QUECTEL_NCP_DEFAULT_SERIAL_BAUDRATE = 115200;
-const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 460800;
+const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 230400;
```
XuGuohui, line 92: Need to be reverted.

system/src/system_network_manager.h
```diff
@@ -105,13 +109,15 @@ class NetworkManager {
     struct InterfaceRuntimeState {
         InterfaceRuntimeState()
                 : ip4State(ProtocolState::UNCONFIGURED),
-                  ip6State(ProtocolState::UNCONFIGURED) {
+                  ip6State(ProtocolState::UNCONFIGURED),
+                  pwrState(IF_POWER_STATE_DOWN) {
```
XuGuohui, line 113: `IF_POWER_STATE_NONE`?

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -322,9 +323,30 @@ int QuectelNcpClient::off() {
...
+        // Power down using hardware
+        if (!modemPowerOff()) {
+            ncpPowerState(NcpPowerState::OFF);
+        }
+        // FIXME: else there is power leakage still.
+    }
+
+    // Disable the UART interface.
+    LOG(TRACE, "Deinit modem serial.");
+    serial_.reset();
```
avtolstoy, line 348: See comments about Stream in esp32 code.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1301,47 +1341,98 @@ int QuectelNcpClient::processEventsImpl() {
 }
 
 int QuectelNcpClient::modemInit() const {
-    hal_gpio_config_t conf = {.size = sizeof(conf), .version = 0, .mode = OUTPUT, .set_value = false, .value = 0};
+    hal_gpio_config_t conf = {
+        .size = sizeof(conf),
+        .version = 0,
+        .mode = INPUT_PULLUP,
+        .set_value = false,
+        .value = 0};
```
avtolstoy, line 1349: Nit: formatting.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1301,47 +1341,98 @@ int QuectelNcpClient::processEventsImpl() {
 }
 
 int QuectelNcpClient::modemInit() const {
```
avtolstoy, line 1343: @YutingYou Please review these.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1301,47 +1341,98 @@ int QuectelNcpClient::processEventsImpl() {
...
-    conf.mode = INPUT_PULLUP;
-    CHECK(HAL_Pin_Configure(BGVINT, &conf));
-
     LOG(TRACE, "Modem low level initialization OK");
 
     return SYSTEM_ERROR_NONE;
 }
 
-int QuectelNcpClient::modemPowerOn() const {
+bool QuectelNcpClient::waitModemPowerOff(system_tick_t timeout) const {
```
avtolstoy, line 1379: Combine these two?

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1301,47 +1341,98 @@ int QuectelNcpClient::processEventsImpl() {
...
+           break;
+       }
+       HAL_Delay_Milliseconds(5);
+   }
+   return powerGood;
+}
+
+int QuectelNcpClient::modemPowerOn() {
+    // The serial_ is always released if powering down the modem is requested.
+    if (!serial_.get()) {
```
avtolstoy, line 1407: Duplicate code. See comments in esp32 implementation.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -301,42 +302,65 @@ int SaraNcpClient::on() {
...
+        ncpPowerState(NcpPowerState::OFF);
+    } else {
+        // Power down using hardware
+        if (!modemPowerOff()) {
+            ncpPowerState(NcpPowerState::OFF);
+        }
+        // FIXME: else there is power leakage still.
+    }
+
+    parser_.destroy();
```
avtolstoy, line 344: See serial stream/parser comments in esp32 code.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1584,11 +1626,46 @@ int SaraNcpClient::modemInit() const {
 
     LOG(TRACE, "Modem low level initialization OK");
 
-    return 0;
+    return SYSTEM_ERROR_NONE;
 }
 
-int SaraNcpClient::modemPowerOn() const {
+bool SaraNcpClient::waitModemPowerOff(system_tick_t timeout) const {
```
avtolstoy, line 1632: Combine these two?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1584,11 +1626,46 @@ int SaraNcpClient::modemInit() const {
...
+           break;
+       }
+       HAL_Delay_Milliseconds(5);
+   }
+   return powerGood;
+}
+
+int SaraNcpClient::modemPowerOn() {
+    // The serial_ is always released if powering down the modem is requested.
+    if (!serial_.get()) {
```
avtolstoy, line 1660: Duplicate initialization code. See comments in esp32 code.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1694,7 +1758,38 @@ int SaraNcpClient::modemPowerOff() {
...
 
     CHECK_TRUE(!modemPowerState(), SYSTEM_ERROR_INVALID_STATE);
-    return 0;
+    return SYSTEM_ERROR_NONE;
+}
+
+int SaraNcpClient::modemSoftPowerOff() {
+    if (modemPowerState()) {
+        LOG(TRACE, "Try powering modem off using AT command");
+        if (ready_) {
```
avtolstoy, line 1767: Nit: refactor to reduce indentation levels? :sweat_smile:

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1734,9 +1829,15 @@ int SaraNcpClient::modemHardReset(bool powerOff) {
         if (!powerOff) {
             LOG(TRACE, "Powering on the modem after the hard reset");
             return modemPowerOn();
+        } else {
+            ncpPowerState(NcpPowerState::OFF);
+            parser_.destroy();
```
avtolstoy, line 1834: See comments about serial stream/parser in esp32 code.

system/src/system_network_manager.cpp
```diff
@@ -812,6 +843,55 @@ void NetworkManager::resetInterfaceProtocolState(if_t iface) {
...
+        ifState->pwrState = IF_POWER_STATE_POWERING_UP;
+        LOG(TRACE, "Request to power on the interface");
+    } else {
+        req.state = IF_POWER_STATE_DOWN;
+        // Update the interface power here to avoid race condition
+        // The power state will be updated on NCP power events
+        ifState->pwrState = IF_POWER_STATE_POWERING_DOWN;
+        LOG(TRACE, "Request to power off the interface");
+    }
+    if (iface == nullptr) {
```
avtolstoy, line 866: This seems to be not reachable.

system/src/system_sleep.cpp
```diff
@@ -81,13 +93,20 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
         spark_cloud_flag_disconnect();
     }
 
+    // TODO: restore network state if network is disconnected but it failed to enter sleep mode.
```
avtolstoy, line 96: We are still restoring it now, right? `system_sleep_network_suspend()` still returns `resume = true` even if something goes wrong.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -102,6 +102,7 @@ void PppNcpNetif::loop(void* arg) {
         self->celMan_->ncpClient()->enable(); // Make sure the client is enabled
         if (!r) {
             // Event
+            LOG(TRACE, "PPP netif event from queue: %d", ev);
```
avtolstoy, line 105: TODO: peek events off of the queue and retry in case of failure.

hal/network/ncp/at_parser/at_parser_impl.cpp
```diff
@@ -600,6 +600,7 @@ int AtParserImpl::nextLine(unsigned* timeout) {
 int AtParserImpl::readMore(unsigned* timeout) {
     assert(bufPos_ < INPUT_BUF_SIZE);
     const auto strm = conf_.stream();
+    CHECK_TRUE(strm, SYSTEM_ERROR_INVALID_STATE);
```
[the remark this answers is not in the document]
XuGuohui: I did hit it, because we de-initialized the modem serial if modem is to be powered off, in which case if user called `Cellular.command()` it will cause sos. Though I added `parser_.destroy()` when the modem is powered off, it won't hurt anything after adding this sanity check.

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
```diff
@@ -169,9 +178,15 @@ int Esp32NcpClient::off() {
         return SYSTEM_ERROR_INVALID_STATE;
     }
     muxer_.stop();
+    ncpPowerState(NcpPowerState::TRANSIENT_OFF);
     espOff();
     ready_ = false;
     ncpState(NcpState::OFF);
+    parser_.destroy();
```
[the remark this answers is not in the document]
XuGuohui: To reset the `std::unique_ptr<detail::AtParserImpl> p_` to nullptr in `AtParser` so that the `Cellular.command()` won't cause any SOS if modem is powered off.

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
```diff
@@ -169,9 +178,15 @@ int Esp32NcpClient::off() {
...
+    // Disable the UART interface.
+    LOG(TRACE, "Deinit modem serial.");
+    serial_.reset();
```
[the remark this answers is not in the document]
XuGuohui: Sounds good. I'll add `.on()` and `.off()` particularly for the `SerialStream` class. Make sense?

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
```diff
@@ -265,6 +284,7 @@ int Esp32NcpClient::updateFirmware(InputStream* file, size_t size) {
     LOG(TRACE, "Initiating XMODEM transfer");
     parser_.reset();
     const auto strm = parser_.config().stream();
+    CHECK_TRUE(strm, SYSTEM_ERROR_INVALID_STATE);
```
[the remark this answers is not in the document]
XuGuohui: See: https://github.com/particle-iot/device-os/pull/2110#discussion_r444194925

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
XuGuohui: Add a space before `0` and `}`?

system/src/system_sleep.cpp
XuGuohui: Just a reminder and in case that we use `CHECK()` in this function.

hal/network/lwip/cellular/pppncpnetif.cpp
XuGuohui: I'd like to achieve this feature in a separated PR, sounds good?

hal/network/ncp/at_parser/at_parser_impl.cpp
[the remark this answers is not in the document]
avtolstoy: I'm afraid I'll disagree. We are masking a bug in the `cellular_command()` implementation. It's not consulting with the NCP client state and that's the reason why it's able to access the parser with an invalid (no stream) state.

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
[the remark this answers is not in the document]
avtolstoy: https://github.com/particle-iot/device-os/pull/2110#discussion_r445373236

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
avtolstoy: newline before `}` :smile:

hal/network/ncp/at_parser/at_parser_impl.cpp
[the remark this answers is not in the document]
XuGuohui: So let's check the power state of modem in `cellular_command()` like this: `CHECK_TRUE(client->ncpPowerState() == NcpPowerState::ON, SYSTEM_ERROR_INVALID_STATE);` and not destroy the AT parser?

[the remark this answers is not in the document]
avtolstoy: It's not enough to check the power state. An AT command can be issued only if the client is an ON and ready state. See how other NCP client methods deal with this (e.g. `SaraNcpClient::getIccid`).

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1376,47 +1376,21 @@ int QuectelNcpClient::modemInit() const {
...
-   system_tick_t now = HAL_Timer_Get_Milli_Seconds();
-   while (HAL_Timer_Get_Milli_Seconds() - now < timeout) {
-       powerGood = modemPowerState();
-       if (powerGood) {
-           break;
-       }
-       HAL_Delay_Milliseconds(5);
-   }
-   return powerGood;
+bool QuectelNcpClient::waitModemPowerState(bool onOff, system_tick_t timeout) const {
```
avtolstoy, line 1379: Nit: consider taking `NcpPowerState` just to be a bit more verbose perhaps?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1629,40 +1627,21 @@ int SaraNcpClient::modemInit() const {
...
-   system_tick_t now = HAL_Timer_Get_Milli_Seconds();
-   while (HAL_Timer_Get_Milli_Seconds() - now < timeout) {
-       powerGood = modemPowerState();
-       if (powerGood) {
-           break;
-       }
-       HAL_Delay_Milliseconds(5);
-   }
-   return powerGood;
+bool SaraNcpClient::waitModemPowerState(bool onOff, system_tick_t timeout) const {
```
avtolstoy, line 1630: Nit: consider taking `NcpPowerState` just to be a bit more verbose perhaps?

PR #2113, opened 2020-05-26
[Gen 3] System thread wakeup on cloud data

avtolstoy:
### Problem

With threading enabled system thread is only scheduled to run every 100ms, or when a task is posted into its queue. This negatively affects OTA times.

### Solution

This PR attempts to add support for system thread wakeup whenever data arrives on a cloud socket. This is only support at the moment for Gen 3 platforms by adding `SIOCSPGRP` ioctl support to LwIP and using threads instead of normal `pid_t` (should we change that and still use `pid_t` and cast from/to abstract thread handles?).

### Steps to Test

0. Use cellular Gen 3 platforms
1. Run a simple blank app with threading disabled, perform OTA of a system part, capture OTA time/transfer rate (available in logs)
2. Enable threading, perform OTA of a system part, capture time/transfer rate (available in logs)
3. Compare the numbers, they should be similar

### Example App

N/A

### References

- https://github.com/particle-iot/lwip/pull/7

---

### Completeness

- [ ] Added documentation


hal/network/lwip/socket_hal.cpp
```diff
@@ -111,3 +111,7 @@ ssize_t sock_recvmsg(int s, struct msghdr *message, int flags) {
 ssize_t sock_sendmsg(int s, const struct msghdr *message, int flags) {
   return lwip_sendmsg(s, message, flags);
 }
+
+int sock_ioctl(int s, long cmd, void* argp) {
```
sergeuz, line 115: Let's expose it via dynalib as well,

Issue #2109, opened 2020-05-21
Buffer overrun on device version


ashesman:
### Expected Behavior

Not overrun the DevStatus.ver string

### Observed Behavior

DevStatus.ver is 16 characters long.  The R410 modem can return a version that is sixteen characters long e.g. L0.0.00.00.05.08, 17 characters with NUL terminator.  So overruns by one character.

### Steps to Reproduce

By observation,  _cbSting callback does not check string length before writing to the buffer it is sent as an argument.

https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp Line 565 is called from line 897.

In fact this callback does not check the length of any string it writes to...

PR #2115, opened 2020-05-27
[Gen2] Fix buffer overrun in modem hal 

keeramis:
### Problem

Modem software version of `DevStatus` is allocated 16 bytes overall, but the modem version itself is 16 bytes, so it is missing an additional byte for the null termination. Similarly, other functions in modem hal are to be checked for buffer size before writing into them.

### Solution

Add an extra byte in the `ver` member of the `DevStatus` struct to accommodate 16 bytes data and one null. 
Fix `_cbString` and other relevant `_cb` functions by adding a `CStringHelper` that allows passing string and string size into the callback, which can be used for size checking when writing into buffers. 

### Example App
Check that version number is logging correctly. This may _not_ be indicative of failure/success but a  basic test app that might help uncover the problem.
```c
#include "application.h"
SerialLogHandler logHandler (LOG_LEVEL_ALL);
void setup() {
}

void loop() {
}
```

### References

[ch55492](https://app.clubhouse.io/particle/story/55492/buffer-overrun-on-device-version)
Closes https://github.com/particle-iot/device-os/issues/2109

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/shared/cellular_enums_hal.h
+++ b/hal/shared/cellular_enums_hal.h
@@ -67,7 +67,7 @@ typedef struct {
     char meid[18+1];    //!< Mobile Equipment IDentifier
     char manu[16];      //!< Manufacturer (u-blox)
     char model[16];     //!< Model Name (LISA-U200, LISA-C200 or SARA-G350)
-    char ver[16];       //!< Software Version
+    char ver[16+1];       //!< Software Version
 } DevStatus;
 //! Registration Status
 typedef enum {
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -359,7 +359,7 @@ bool MDMParser::_atOk(void)
 
 int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                              void* param /* = NULL*/,
-                             system_tick_t timeout_ms /*= 5000*/)
+                             system_tick_t timeout_ms /*= 10000*/)
 {
     if (_cancel_all_operations) return WAIT;
 
@@ -478,9 +478,9 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG
                     // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC
                     b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1;
-                    r = sscanf(cmd, "%s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
+                    r = sscanf(cmd, "%s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);    // XXX: Fix for buffer overrun if applicable
                     if (r <= 1)
-                        r = sscanf(cmd, "%s %d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
+                        r = sscanf(cmd, "%s %d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);    // XXX: Fix for buffer overrun if applicable
                     if (r >= 2) {
                         Reg *reg = !strcmp(s, "CREG:")  ? &_net.csd :
                                    !strcmp(s, "CGREG:") ? &_net.psd :
@@ -562,10 +562,12 @@ int MDMParser::_cbCEDRXS(int type, const char* buf, int len, EdrxActs* edrxActs)
     return WAIT;
 }
 
-int MDMParser::_cbString(int type, const char* buf, int len, char* str)
+int MDMParser::_cbString(int type, const char* buf, int len, CStringHelper* str)
 {
-    if (str && (type == TYPE_UNKNOWN)) {
-        if (sscanf(buf, "\r\n%s\r\n", str) == 1)
+    if (str && str->str && (str->size > 1) && (type == TYPE_UNKNOWN)) {
+        char format[32] = {};
+        snprintf(format, sizeof(format), "\r\n%%%us\r\n", str->size - 1);
+        if (sscanf(buf, format, str->str) == 1)
             /*nothing*/;
     }
     return WAIT;
@@ -872,6 +874,15 @@ bool MDMParser::init(DevStatus* status)
 {
     LOCK();
     MDM_INFO("\r\n[ Modem::init ] = = = = = = = = = = = = = = =");
+
+    // Define all cstringhelpers up here, because "goto"s do not like bypassing inits
+    CStringHelper str_imei(_dev.imei, sizeof(_dev.imei));
+    CStringHelper str_manu(_dev.manu, sizeof(_dev.manu));
+    CStringHelper str_ver(_dev.ver, sizeof(_dev.ver));
+    CStringHelper str_imsi(_dev.imsi, sizeof(_dev.imsi));
+    CStringHelper str_verExt(_verExtended, sizeof(_verExtended));
+    CStringHelper str_ccid(_dev.ccid, sizeof(_dev.ccid));
+
     if (_dev.dev == DEV_SARA_R410) {
         // TODO: Without this delay, some commands, such as +CIMI, may return a SIM failure error.
         // This probably has something to do with the SIM initialization. Should we check the SIM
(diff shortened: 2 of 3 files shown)


keeramis:
@technobly 
The issue also references that we are not actually limiting to the length of the string `str` in `cbString` here - https://github.com/particle-iot/device-os/blob/30a8f2e409d3351bfcdff85b4f0efb53fe8eddae/hal/src/electron/modem/mdm_hal.cpp#L565

However, it does not look like the length allocated to members in `DevStatus` are actually accessible by the callback (the length here - https://github.com/particle-iot/device-os/blob/30a8f2e409d3351bfcdff85b4f0efb53fe8eddae/hal/shared/cellular_enums_hal.h#L70)?

hal/shared/cellular_enums_hal.h
```diff
@@ -67,7 +67,7 @@ typedef struct {
     char meid[18+1];    //!< Mobile Equipment IDentifier
     char manu[16];      //!< Manufacturer (u-blox)
     char model[16];     //!< Model Name (LISA-U200, LISA-C200 or SARA-G350)
-    char ver[16];       //!< Software Version
+    char ver[16+1];       //!< Software Version
```
avtolstoy, line 70: This also does not resolve the issue for a case where we are running an old application (without this change yet) and system part with this fix. This is why normally for ABI stability we have `size` and/or `version` fields in structs that go through dynalib interface.

We are also not really fixing the root cause of the problem as you've already pointed out: `_cbString` has no idea of the size of the buffer it was passed. This should be trivially fixable by passing a pointer to a stack-allocated struct with `char*` and size, instead of just `char*`.

```c++
// Somewhere
struct CStringHelper {
    CStringHelper(char* s, size_t sz)
        : str{s},
          size{sz} {
    }
    CStringHelper() = delete;

    char* str;
    size_t size;
};

// ...
CStringHelper str(_dev.ver, sizeof(_dev.ver));
sendFormated("AT+CGMR\r\n");
if (RESP_OK != waitFinalResp(_cbString, &str)) {
    // ...
// ...

int MDMParser::_cbString(int type, const char* buf, int len, CStringHelper* str)
{
    if (str && (type == TYPE_UNKNOWN)) {
        char format[32] = {};
        snprintf(format, sizeof(format), "\r\n%%%us\r\n", str->size);
        if (sscanf(buf, format, str->str) == 1)
            /*nothing*/;
    }
    return WAIT;
}
```

technobly: I like Andrey's suggestion here, and it looks good to me.  @keeramis would you be willing to test this approach?  We would update the PR to patch all _cbString callback calls as well.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -478,9 +478,9 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG
                     // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC
                     b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1;
-                    r = sscanf(cmd, "%s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
+                    r = sscanf(cmd, "%s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);    // XXX: Fix for buffer overrun if applicable
```
keeramis, line 481: Unclear that this needs to be fixed or looked at. Please advice. Does this look like another place to fix this and probably limit the `%s` to `%5s` or so?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -3186,6 +3201,7 @@ bool MDMParser::smsRead(int ix, char* num, char* buf, int len)
     param.num = num;
     param.buf = buf;
     sendFormated("AT+CMGR=%d\r\n",ix);
+    // FIXME: check for SMS buffer size when smsRead() is being used
```
keeramis, line 3204: This function `smsRead()` is currently not being used. And also unclear that this needs a fix?

hal/shared/cellular_enums_hal.h
keeramis: Also made fixes in other places where strings sizes are probably not checked. Please check..

And tested that values are getting stored into the `_dev.variable` correctly as well.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: `%31s` as `char s[32]` above.

avtolstoy: Same below.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1202,7 +1213,7 @@ int MDMParser::_cbCPIN(int type, const char* buf, int len, Sim* sim)
     if (sim) {
         if (type == TYPE_PLUS){
             char s[16];
-            if (sscanf(buf, "\r\n+CPIN: %[^\r]\r\n", s) >= 1)
+            if (sscanf(buf, "\r\n+CPIN: %16[^\r]\r\n", s) >= 1)
```
avtolstoy, line 1216: 15

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1212,10 +1223,12 @@ int MDMParser::_cbCPIN(int type, const char* buf, int len, Sim* sim)
...
 
-int MDMParser::_cbCCID(int type, const char* buf, int len, char* ccid)
+int MDMParser::_cbCCID(int type, const char* buf, int len, CStringHelper* str)
 {
-    if ((type == TYPE_PLUS) && ccid) {
-        if (sscanf(buf, "\r\n+CCID: %[^\r]\r\n", ccid) == 1) {
+    if (str && str->str && (str->size > 1) && (type == TYPE_PLUS)) {
+        char format[32] = {};
+        snprintf(format, sizeof(format), "\r\n+CCID: %%%u[^\r]\r\n", str->size-1);
+        if (sscanf(buf, "\r\n+CCID: %[^\r]\r\n", str->str) == 1) {
```
avtolstoy, line 1231: `format` should be used here

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -359,7 +359,7 @@ bool MDMParser::_atOk(void)
 
 int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                              void* param /* = NULL*/,
-                             system_tick_t timeout_ms /*= 5000*/)
+                             system_tick_t timeout_ms /*= 10000*/)
```
avtolstoy, line 362: Any reasoning behind the change?

keeramis: Fixed the comment because we are using `10000 msec` 
please see this? https://github.com/particle-iot/device-os/blob/30a8f2e409d3351bfcdff85b4f0efb53fe8eddae/hal/src/electron/modem/mdm_hal.h#L454

hal/src/electron/modem/mdm_hal.cpp
keeramis: My bad, that was the whole point of using format. 🤦

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -478,9 +478,9 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG
                     // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC
                     b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1;
-                    r = sscanf(cmd, "%s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
+                    r = sscanf(cmd, "%31s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
```
avtolstoy, line 481: There is another UUPSD scanf which uses `%s` above that also needs to be `%31s`.

Another case is `_cbURDFILE()`, but it's not used so not really important.

hal/src/electron/modem/mdm_hal.cpp
technobly: Probably ok to leave it for now since it's not an official API

avtolstoy:
Please rebase and merge.

PR #2116, opened 2020-05-28
Fixes the issue that FuelGauge doesn't work as expected after being woken up

XuGuohui:
### Problem

SoC (State of Charging) is not updating after device being woken up.

### Solution

From the FuelGauge's datasheet, we should add a delay about 500ms after the FuelGauge exiting sleep mode.

We simply add the delay in system power manager, instead in the FuelGauge library. This provides an opportunity that users can omit such a delay if they use the FuelGauge library directly and don't need to update the SoC right after the FuelGauge being woken up. @rickkas7 We should probably document this for visibility.

Reference: 
https://community.particle.io/t/sleepy-boron-soc-not-updating/55086/37

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_power_manager.cpp
+++ b/system/src/system_power_manager.cpp
@@ -143,6 +143,8 @@ void PowerManager::sleep(bool s) {
       initDefault();
       update();
       gauge.wakeup();
+      // Delay for at least 500ms to make sure the fuelgauge is ready to function after woken up.
+      HAL_Delay_Milliseconds(500);
     }
   }
 }


system/src/system_power_manager.cpp
```diff
@@ -143,6 +143,8 @@ void PowerManager::sleep(bool s) {
       initDefault();
       update();
       gauge.wakeup();
+      // Delay for at least 500ms to make sure the fuelgauge is ready to function after woken up.
+      delay(500);
```
avtolstoy, line 147: Let's use `HAL_Delay_Milliseconds()` here.

avtolstoy:
Please rebase and merge.

PR #2119, opened 2020-06-01
Process 'CEREG: 0' URCs on R410M

keeramis:
### Problem

A registration URC like `CEREG: 0` is not processed by Electron LTEs, and hence even though a `CEREG: 0` is received by the device, the device still blinks cyan assuming it's connected to cloud, and does not attempt a disconnection process.

### Solution

Read and process reg URCs for `CxREG :0` to notify a disconnection

### Example App

After a successful cloud connection, remove antenna or trigger (somehow) a loss of registration, by which you will see a `CEREG: 0` on the module. When this happens, notice that the device should go back to blinking green, signaling a new connection attempt. 

```c
void setup() {
}

void loop() {
}
```

### References

[ch36648](https://app.clubhouse.io/particle/story/36648/electron-lte-stuck-breathing-cyan-but-offline-when-de-registered-urc-received-by-not-acted-on-to-force-network-teardown)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -418,8 +418,6 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                     MDM_PRINTF("New SMS at index %d\r\n", a);
                     if (sms_cb) SMSreceived(a);
                 }
-                // TODO: This should include other LTE devices,
-                // perhaps adding _net.act < 7
                 else if ((_dev.dev != DEV_SARA_R410) && (sscanf(cmd, "CIEV: 9,%d", &a) == 1)) {
                     MDM_PRINTF("CIEV matched: 9,%d\r\n", a);
                     // Wait until the system is attached before attempting to act on GPRS detach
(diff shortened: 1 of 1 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1606,7 +1613,7 @@ bool MDMParser::getCellularGlobalIdentity(CellularGlobalIdentity& cgi_) {
 }
 
 void MDMParser::_setBandSelectString(MDM_BandSelect &data, char* bands, int index /*= 0*/) {
-    char band[5];
+    char band[6];
```
technobly, line 1616: I'm guessing this band change shouldn't be in this PR?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -493,6 +491,15 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
...
                             else if (a == 5) *reg = REG_ROAMING;  // 5: registered, roaming
+                            if (_dev.dev == DEV_SARA_R410 && _attached && !strcmp(s, "CEREG:") && *reg == REG_NONE) {
+                                MDM_PRINTF("Cell reg disconnected\r\n");
+                                _attached_urc = (*reg == REG_NONE) ? 0:1;
+                                if (!_attached_urc) { // Regitration detected as 0 or 2
+                                    _ip = NOIP;
+                                    _attached = false;
+                                    HAL_NET_notify_disconnected();
+                                }
+                            }
```
technobly, line 502: Nit: spelling on "Registration"

Should `&& *reg == REG_NONE` be `&& (*reg != REG_HOME || *reg != REG_ROAMING))` since that's what we define as `REG_OK()`

`_attached_urc = (*reg == REG_NONE) ? 0:1;` would also not need to be tested, just set to 0, and also then no need to test for `if (!_attached_urc)`.

Perhaps also put the `&& !strcmp(s, "CEREG:")` last, as it will chew up more cycles.  

We can also test `&& reg == &_net.eps` instead of strcmp() again?

PR #2120, opened 2020-06-01
[Gen3] Hide wifi wiring API that are not supported

keeramis:
### Problem

Wifi wiring calls that are either unimplemented or unsupported might cause confusion to customers who are trying to use them (may be because they are likely supported on different platforms)

### Solution

Throw compile time errors for unimplemented or unsupported wifi wiring calls on Gen3

### Example App

The following code should throw a compile error when built on Gen3 - Argon, for example
such as 
```
filename.cpp:6:28: error: 'class spark::WiFiClass' has no member named 'hostname'; did you mean 'setHostname'?
```

```c
#include "application.h"

void setup() {
    String hostname = WiFi.hostname();
    Serial.printlnf("Wifi hostname is %s" ,hostname);
}

void loop() {
}
```

### References

[ch51761](https://app.clubhouse.io/particle/story/51761/hide-wifi-hostname-on-gen-3)
[ch51357](https://app.clubhouse.io/particle/story/51357/wifi-hostname-is-unimplemented-in-gen-3)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_wifi.h
+++ b/wiring/inc/spark_wiring_wifi.h
@@ -238,6 +238,7 @@ public:
     }
 #endif // !HAL_USE_INET_HAL_POSIX
 
+#if !HAL_PLATFORM_NCP
     void setStaticIP(const IPAddress& host, const IPAddress& netmask,
         const IPAddress& gateway, const IPAddress& dns)
     {
@@ -248,6 +249,7 @@ public:
     {
         setIPAddressSource(STATIC_IP);
     }
+#endif //!HAL_PLATFORM_NCP
 
     void useDynamicIP()
     {
@@ -267,6 +269,7 @@ public:
 
     int getCredentials(WiFiAccessPoint* results, size_t result_count);
 
+#if !HAL_PLATFORM_NCP
     String hostname()
     {
         const size_t maxHostname = 64;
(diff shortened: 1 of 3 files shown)


keeramis:
For reference - the flag covers implementations on Boron, B5SoM, and Argon.

wiring/inc/spark_wiring_wifi.h
```diff
@@ -267,13 +269,15 @@ class WiFiClass : public NetworkClass
...
     String hostname()
     {
         const size_t maxHostname = 64;
         char buf[maxHostname] = {0};
         network_get_hostname(*this, 0, buf, maxHostname, nullptr);
         return String(buf);
     }
+#endif
 
     int setHostname(const String& hostname)
```
avtolstoy, line 282: `setHostname()` is also not supported.

wiring/inc/spark_wiring_wifi.h
```diff
@@ -253,6 +254,7 @@ class WiFiClass : public NetworkClass
     {
```
avtolstoy, line 254: This method can probably be left enabled.

wiring/inc/spark_wiring_wifi.h
keeramis: 👍 This returns a `SYSTEM_ERROR_NOT_SUPPORTED` though, once compiled.

https://github.com/particle-iot/device-os/blob/4105ae862d01a57df14a306c842136645b3fc2b2/hal/src/argon/wlan_hal.cpp#L389

avtolstoy:
@keeramis API tests (`wiring/api`) also need to account for this changes. See CI error logs.

PR #2121, opened 2020-06-01
[Gen2] Replace sprintf with snprintf in modem hal

keeramis:
### Problem

Compile error when building Electron after updating to gcc compiler 9.2.1
```
error: 'sprintf' may write a terminating nul past the end of the destination
```

### Solution

Use `snprintf` instead of `sprintf` to give number of bytes.

### Steps to Test

Compiling Electron on gcc 9.2.1 likely triggers this error.

### Example App

```c
void setup() {
}

void loop() {
}
```

### References

[ch55930](https://app.clubhouse.io/particle/story/55930/electron-fix-an-sprintf-error-surfacing-in-gcc-9-2-1)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1608,7 +1608,7 @@ failure:
 void MDMParser::_setBandSelectString(MDM_BandSelect &data, char* bands, int index /*= 0*/) {
     char band[5];
     for (int x=index; x<data.count; x++) {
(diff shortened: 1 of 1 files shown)


keeramis:
Interestingly enough, after some other random work on my computer, I no longer see this error on Electron. Similar to other folks. However, since this a valid error to be checked, added in the potential fix mentioned. There are no other `sprintf`s in mdm_hal.cpp

PR #2125, opened 2020-06-05
Implement sleep function for UART peripheral.

XuGuohui:
### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
    Serial1.begin(115200);

    Serial1.println("Application started.");
}

void loop() {
    static system_tick_t now = millis();
    if (millis() - now >= 1000) {
        now = millis();
        Serial1.println("Suspend UART");
        hal_usart_sleep(HAL_USART_SERIAL1, true, nullptr);
        delay(1000);
        hal_usart_sleep(HAL_USART_SERIAL1, false, nullptr);
        Serial1.println("UART state is restored");
        Serial1.println("fjkgsajfgkajsgfkyusdgh");
    }
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/usart_hal.cpp
+++ b/hal/src/nRF52840/usart_hal.cpp
@@ -88,7 +88,7 @@ inline uint32_t pinToNrf(pin_t pin) {
 }
 
 class Usart;
-Usart* getInstance(HAL_USART_Serial serial);
+Usart* getInstance(hal_usart_interface_t serial);
 void uarte0InterruptHandler(void);
 void uarte1InterruptHandler(void);
 
@@ -123,12 +123,12 @@ public:
         unsigned int config;
     };
 
-    int init(const HAL_USART_Buffer_Config& conf) {
-        if (enabled_) {
+    int init(const hal_usart_buffer_config_t& conf) {
+        if (isEnabled()) {
             CHECK(end());
         }
 
-        if (configured_) {
+        if (isConfigured()) {
             CHECK(deInit());
         }
 
@@ -152,7 +152,7 @@ public:
     }
 
     int begin(const Config& conf) {
-        CHECK_TRUE(configured_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isConfigured(), SYSTEM_ERROR_INVALID_STATE);
         CHECK_TRUE(validateConfig(conf.config), SYSTEM_ERROR_INVALID_ARGUMENT);
         auto nrfBaudRate = CHECK_RETURN(getNrfBaudrate(conf.baudRate), SYSTEM_ERROR_INVALID_ARGUMENT);
 
@@ -160,7 +160,7 @@ public:
 
         suspend_ = false;
 
-        if (enabled_) {
+        if (isEnabled()) {
             end();
         }
 
@@ -280,7 +280,7 @@ public:
     }
 
     int suspend() {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_NONE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_NONE);
         CHECK_FALSE(suspend_, SYSTEM_ERROR_NONE);
         suspend_ = true;
         flush();
@@ -288,13 +288,13 @@ public:
     }
 
     int restore() {
-        CHECK_TRUE(configured_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isConfigured(), SYSTEM_ERROR_INVALID_STATE);
         CHECK_TRUE(suspend_, SYSTEM_ERROR_NONE);
         return begin(config_);
     }
 
     ssize_t data() {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         RxLock lk(uarte_);
         ssize_t d = rxBuffer_.data();
         if (receiving_) {
@@ -309,13 +309,13 @@ public:
     }
 
     ssize_t space() {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         TxLock lk(uarte_);
         return txBuffer_.space();
     }
 
     ssize_t read(uint8_t* buffer, size_t size) {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         const ssize_t maxRead = CHECK(data());
         const size_t readSize = std::min((size_t)maxRead, size);
         CHECK_TRUE(readSize > 0, SYSTEM_ERROR_NO_MEMORY);
@@ -331,7 +331,7 @@ public:
     }
 
     ssize_t peek(uint8_t* buffer, size_t size) {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         const ssize_t maxRead = CHECK(data());
         const size_t peekSize = std::min((size_t)maxRead, size);
         CHECK_TRUE(peekSize > 0, SYSTEM_ERROR_NO_MEMORY);
@@ -340,7 +340,7 @@ public:
     }
 
     ssize_t write(const uint8_t* buffer, size_t size) {
-        CHECK_TRUE(enabled_, SYSTEM_ERROR_INVALID_STATE);
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         const ssize_t canWrite = CHECK(space());
         const size_t writeSize = std::min((size_t)canWrite, size);
         CHECK_TRUE(writeSize > 0, SYSTEM_ERROR_NO_MEMORY);
@@ -378,7 +378,7 @@ public:
     }
 
     void pump() {
-        if (!willPreempt() && enabled_) {
+        if (!willPreempt() && isEnabled()) {
             interruptHandler();
         }
     }
@@ -730,7 +730,7 @@ const auto UARTE0_INTERRUPT_PRIORITY = APP_IRQ_PRIORITY_LOW;
 // TODO: move this to hal_platform_config.h ?
 const auto UARTE1_INTERRUPT_PRIORITY = (app_irq_priority_t)_PRIO_SD_LOWEST;
 
-Usart* getInstance(HAL_USART_Serial serial) {
+Usart* getInstance(hal_usart_interface_t serial) {
     static Usart usartMap[] = {
         {NRF_UARTE0, uarte0InterruptHandler, UARTE0_INTERRUPT_PRIORITY, NRF_TIMER2, NRF_PPI_CHANNEL4, TX, RX, CTS, RTS},
         {NRF_UARTE1, uarte1InterruptHandler, UARTE1_INTERRUPT_PRIORITY, NRF_TIMER3, NRF_PPI_CHANNEL5, TX1, RX1, CTS1, RTS1}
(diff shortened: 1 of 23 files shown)


hal/src/nRF52840/usart_hal.cpp
```diff
@@ -262,17 +268,33 @@ class Usart {
...
-        return 0;
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int suspend() {
+        CHECK_TRUE(enabled_, SYSTEM_ERROR_NONE);
+        CHECK_FALSE(suspend_, SYSTEM_ERROR_NONE);
+        suspend_ = true;
+        flush();
+        return end(true);
```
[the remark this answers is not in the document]
XuGuohui: I will configure the TX/RTS pins to `INPUT_PULLUP`, and configure the RX/CTS pins to `PIN_MODE_NONE`(gen3)/`INPUT`(gen2) for USART. By this way, it won't mess up the other side.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -226,49 +228,76 @@ class Usart {
...
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int suspend() {
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_NONE);
+        CHECK_FALSE(suspend_, SYSTEM_ERROR_NONE);
+
+        AtomicSection lk;
+
+        flush();
```
YutingYou, line 273: `AtomicSection` disables the interrupts, `flush()` will not work anymore.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -226,49 +228,76 @@ class Usart {
...
         receiving_ = 0;
         rxBuffer_.reset();
         txBuffer_.reset();
 
-        return 0;
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int restore() {
+        CHECK_TRUE(suspend_, SYSTEM_ERROR_NONE);
```
YutingYou, line 299: Race condition when accessing `suspend_`, needs a lock or `AtomicSection` here.

hal/src/stm32f2xx/usart_hal.c
```diff
@@ -262,8 +262,8 @@ void usartEndImpl(hal_usart_interface_t serial, bool cache) {
     USART_Cmd(usartMap[serial]->peripheral, DISABLE);
 
     // Switch pins to INPUT
-    HAL_Pin_Mode(usartMap[serial]->rx_pin, INPUT);
-    HAL_Pin_Mode(usartMap[serial]->tx_pin, INPUT);
+    HAL_Pin_Mode(usartMap[serial]->usart_rx_pin, INPUT);
+    HAL_Pin_Mode(usartMap[serial]->usart_tx_pin, INPUT_PULLUP);
```
avtolstoy: Only when going into a sleep state?

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -231,49 +231,34 @@ class Usart {
...
-
-        nrf_uarte_disable(uarte_);
+        suspend_ = false;
 
-        nrf_uarte_txrx_pins_disconnect(uarte_);
-        nrf_uarte_hwfc_pins_disconnect(uarte_);
+        CHECK(disable());
 
-        HAL_Pin_Mode(txPin_, INPUT);
+        // Configuring the input pins' mode to PIN_MODE_NONE will disconnect input buffer to enable power savings
```
avtolstoy, line 242: Only when going into sleep state? This also looks almost 1-1 to `suspend()`.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -289,6 +289,7 @@ class Usart {
         enabled_ = false;
         transmitting_ = false;
         receiving_ = 0;
+        rxBuffer_.reset();
```
avtolstoy, line 292: Why? There might be some leftover data that may be needed.

hal/src/stm32f2xx/usart_hal.c
```diff
@@ -286,14 +286,11 @@ void usartEndImpl(hal_usart_interface_t serial, bool cache) {
...
 
-    if (!cache) {
-        // clear any received data
-        usartMap[serial]->rx_buffer->head = usartMap[serial]->rx_buffer->tail;
-        // Undo any pin re-mapping done for this USART
-        // ...
-        memset(usartMap[serial]->rx_buffer, 0, sizeof(hal_usart_ring_buffer_t));
-    }
-
+    // clear any received data
```
avtolstoy, line 289: Not needed for sleep case.

user/tests/wiring/serial_loopback2/loopback.cpp
```diff
@@ -19,36 +19,59 @@
...
+        assertEqual(available, Serial1.available());
+        if (available) {
+            rxBuf[pos] = (char)Serial1.read();
+        }
+    } while (++pos <= bufferSize);
+
+    assertTrue(!strncmp(txBuf, rxBuf, bufferSize));
+}
+
+test(SERIAL_00_LoopbackNoDataLossAndAvailableIsCorrect) {
```
avtolstoy, line 46: Another test case should be added that any leftover data in the receive buffer is left untouched between the sleep/wake.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -226,49 +228,76 @@ class Usart {
...
+        receiving_ = 0;
+        config_ = {};
+        rxBuffer_.reset();
+        txBuffer_.reset();
 
-        disableTimer();
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int suspend() {
```
avtolstoy, line 267: In order for things to be somewhat more robust, if using hardware flow control we should probably deassert RTS first to make sure that the other side does not send anything while we deinitialize things. Same applies to STM32 implementation.

hal/src/stm32f2xx/usart_hal.c
XuGuohui: Wouldn't be better to also optimize power consumption if USART is ended?

hal/src/nRF52840/usart_hal.cpp
XuGuohui: Without reseting the `rxBuffer_`, the receiver cannot work properly. It will fail the unit test. If the RX buffer needs to be retained, I'll try to figure out the root cause.

XuGuohui: I think the unit test failed due to not flushing the TX buffer as pointed out: https://github.com/particle-iot/device-os/pull/2125#discussion_r441307149. I'll add a test case to verify if the RX buffer is retained.

hal/src/stm32f2xx/usart_hal.c
avtolstoy: In the case where we are just stopping the UART we supposedly don't have anything powered on the other side. I would assume that `INPUT_PULLUP` would consume more power than `INPUT`.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -221,54 +221,81 @@ class Usart {
...
+        txBuffer_.reset();
 
-        disableTimer();
+        return SYSTEM_ERROR_NONE;
+    }
 
-        enabled_ = false;
-        config_ = {};
+    int suspend() {
+        CHECK_TRUE(isEnabled(), SYSTEM_ERROR_NONE);
```
YutingYou, line 263: I'd prefer to return `SYSTEM_ERROR_INVALID_STATE` here, after calling suspend() API, the user would think the UART is in the suspended state, but actually it is in the disabled state which is confusing. The suspend operation should only occur when the device is in the enabled state.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -221,54 +221,81 @@ class Usart {
...
+        rxBuffer_.prune();
         txBuffer_.reset();
 
-        return 0;
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int restore() {
+        AtomicSection lk;
+        CHECK_TRUE(state_ == HAL_USART_STATE_SUSPENDED, SYSTEM_ERROR_NONE);
```
YutingYou, line 297: Same here and Gen2 platform

hal/src/stm32f2xx/usart_hal.c
```diff
@@ -547,8 +528,39 @@ void hal_usart_begin_config(hal_usart_interface_t serial, uint32_t baud, uint32_
 }
 
 void hal_usart_end(hal_usart_interface_t serial) {
-    usartMap[serial]->suspended = false;
     usartEndImpl(serial);
+
+    // Switch pins to INPUT
+    HAL_Pin_Mode(usartMap[serial]->rx_pin, INPUT);
```
avtolstoy, line 534: Nit: this seems almost identical to the code block in `hal_usart_sleep()`. Consider refactoring.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -265,38 +260,41 @@ class Usart {
...
+            HAL_Pin_Mode(ctsPin_, PIN_MODE_NONE);
+        }
 
-        suspend_ = true;
-        enabled_ = false;
+        state_ = HAL_USART_STATE_SUSPENDED;
         transmitting_ = false;
         receiving_ = 0;
-        rxBuffer_.reset();
+        rxBuffer_.prune();
```
avtolstoy, line 289: This should potentially not be needed since we've ran `data()` previously.

hal/src/nRF52840/usart_hal.cpp
avtolstoy: I don't see this implemented. This will probably work fine on Gen 3 (again, probably). Both on Gen 2 and Gen 3 we should be able to disconnect HWFC pins from the peripheral and manually assert the necessary state.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -221,54 +221,81 @@ class Usart {
...
+        data();
+        
+        CHECK(disable());
+
+        // Configuring the input pins' mode to PIN_MODE_NONE will disconnect input buffer to enable power savings
+        // Configuring the output pins' mode to INPUT_PULLUP to not polluting the state on other side.
+        // NOTE: Do not reset pin function!
+        HAL_Pin_Mode(txPin_, INPUT_PULLUP);
+        HAL_Pin_Mode(rxPin_, PIN_MODE_NONE);
+        if (config_.config & SERIAL_FLOW_CONTROL_RTS) {
```
XuGuohui, line 279: Answer to https://github.com/particle-iot/device-os/pull/2125/files?file-filters%5B%5D=.c&file-filters%5B%5D=.cpp#r451252956

hal/src/nRF52840/usart_hal.cpp
XuGuohui: Without clearing the pending head and tail, the available data is incorrect after restoring the UART. The `user/tests/wiring/serial_loopback2` will fail.

hal/src/stm32f2xx/usart_hal.c
XuGuohui: Is there any defect configuring these pins to `INPUT` if UART is going to be ended? Or is it just violate the existing behavior?

hal/src/nRF52840/usart_hal.cpp
avtolstoy: I'm saying that they should not be cleared but commited (as in `data()`) function.

hal/src/stm32f2xx/usart_hal.c
avtolstoy: I'm just saying that you have code duplication :) We can leave things as-is or to make it a bit more concise to refactor into a common function with argument.

hal/src/nRF52840/usart_hal.cpp
avtolstoy: This is still somewhat of a race condition. Ideally the steps should be to de-assert RTS before performing any deinitialization so that the other side again stops sending data before start doing anything.

avtolstoy: Probably fine as-is for now.

XuGuohui: Will refactor to change the sequence.

hal/src/nRF52840/usart_hal.cpp
XuGuohui: We should at least clear the `headPending_`, otherwise, when restoring the UART the receiver cannot be started, as the acquirable size will be 0. The value of `headPending_` after calling `data()` is actually the aborted acquired space of the RX ring buffer.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -497,9 +470,16 @@ class Usart {
...
 
 private:
-    int disable() {
+    int disable(bool end = true) {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
 
+        if (config_.config & SERIAL_FLOW_CONTROL_RTS) {
+            if (receiving_) {
+                // Issue the STOPRX task to deactive the RTS
+                nrf_uarte_task_trigger(uarte_, NRF_UARTE_TASK_STOPRX);
```
avtolstoy, line 479: :+1: Like it.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -511,6 +491,27 @@ class Usart {
 
         disableTimer();
```
avtolstoy, line 493: One more thing: you might need to clear pending UART IRQ here.

PR #2126, opened 2020-06-10
Use newlib-nano headers, FreeRTOS _impure_ptr hooks

avtolstoy:
### Problem

When building DeviceOS with ARM GCC toolchain, we link with a variant of newlib (an implementation of C standard library) called nano, which has certain trade-offs to reduce the code/RAM size.

In order to support thread-safety for a number of libc functions, we are allocating a so-called reentrant structure for each thread. This is a FreeRTOS feature which we have enabled on all the platforms because we've seen various crashes due to absence of that.

There is currently a number of problems:

1. We are only linking with nano version of newlib, however when building we are not specifying that in CFLAGS, which causes us to use incorrect include paths, which increases size of some of the libc structures. One of them is as you would expect the reentrant structure mentioned above. With nano version it's ~100 bytes, with full libc variant it's over a 1000. We are erroneously allocating more than needed and on top of that the nano version of newlib would allocate even more things dynamically as needed
2. WICED, see particle-iot/photon-wiced#36
3. Incompatibility between newlibs in WICED and DeviceOS (see particle-iot/photon-wiced#35)
4. Each DeviceOS module (system-part1,2,3, user module) has its own reentrant structure and any module except for the system part where FreeRTOS resides does not actually have thread safety if some libc function is not being used through dynalib, as the module-local `_impure_ptr` pointer residing in RAM will not get changed on thread context switches. We need to add hooks into other modules, so that we can do that from the system part with FreeRTOS.

### Solution

1. We are using `--specs=nano.specs` in CFLAGS now, to ensure that not only we are linking with nano variant of the newlib, but also that we are using the correct include paths when building DeviceOS
2. See particle-iot/photon-wiced#36
3. Necessary changes to support particle-iot/photon-wiced#35
4. We have added hooks into FreeRTOS and necessary dynalib functions to ensure that on FreeRTOS context switches we are modifying `_impure_ptr` accordingly in all the modules

This PR also adds appropriate tests to `wiring/no_fixture` and moves SPI tests out of that test suite into a separate `wiring/no_fixture_spi` in order to fit 128K on Electron.

### Steps to Test

- `wiring/no_fixture`
- Test Gen 2 (Photon/P1/Electron) and at least on Gen 3 platform
- Step-by-step OTA/ymodem update from 1.5.x

### Example App

N/A

### References

- [CH56479]
- particle-iot/photon-wiced#35
- particle-iot/photon-wiced#36

---

### Completeness

- [ ] Added documentation


modules/shared/stm32f2xx/inc/system_part3_loader.c
```diff
@@ -56,6 +57,10 @@ void module_system_part3_init()
 
 }
 
+void module_system_part3_newlib_impure_set(struct _reent* r, size_t size, uint32_t version, void* ctx) {
+    _impure_ptr = r;
```
sergeuz, line 61: I'm wondering if we should add a run-time assert that would check that the size of the reent structure matches the size of the structure this module was built with – here and in the user part's callback, or, alternatively, have a static assert somewhere that would indicate that the size of the structure has changed between GCC releases.

hal/src/electron/FreeRTOSConfig.h
```diff
@@ -139,6 +139,12 @@ extern void vApplicationTaskDeleteHook(void *pvTaskToDelete, volatile long* pxPe
 // Allocate newlib's reent structure for each task
 #define configUSE_NEWLIB_REENTRANT 1
 
+// This macro allows us to understand when we are included from FreeRTOS sources
+#ifdef MPU_WRAPPERS_INCLUDED_FROM_API_FILE
+extern void newlib_impure_ptr_change(struct _reent* r);
+#define traceTASK_SWITCHED_IN() newlib_impure_ptr_change(&(pxCurrentTCB->xNewLib_reent))
+#endif // MPU_WRAPPERS_INCLUDED_FROM_API_FILE
```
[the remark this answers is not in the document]
avtolstoy: Removed.

modules/shared/stm32f2xx/inc/system_part3_loader.c
avtolstoy: As per discussion, we'll leave this as is, as it doesn't guarantee anything.

PR #2130, opened 2020-06-17
Implement sleep function for SPI peripheral

XuGuohui:
This PR implements the `hal_spi_sleep()` API to assist in implementing the ultra-low power mode.

---

### Completeness

- [ ] Added documentation


hal/src/stm32f2xx/spi_hal.c
```diff
@@ -618,7 +624,22 @@ static void spiOnSelectedHandler(void *data) {
...
+    if (spiState[spi].select_user_callback) {
+        spiState[spi].select_user_callback(state);
+    }
+}
+
+int hal_spi_sleep(hal_spi_interface_t spi, bool sleep, void* reserved) {
+    if (sleep) {
+        CHECK_TRUE(hal_spi_is_enabled(spi), SYSTEM_ERROR_NONE);
+        CHECK_FALSE(spiState[spi].suspended, SYSTEM_ERROR_NONE);
+        hal_spi_transfer_dma_cancel(spi);
```
avtolstoy, line 636: This seems different from Gen 3 implementation where we wait for the transfer to complete. Let's set on a single behavior between platforms.

user/tests/wiring/spi_master_slave/spi_master/spi_master.cpp
```diff
@@ -23,10 +23,12 @@
 
 #if (USE_SPI == 0 || USE_SPI == 255) // default to SPI
 #define MY_SPI SPI
+#define MY_SPI_IF HAL_SPI_INTERFACE1
```
avtolstoy, line 26: Nit: or alternatively add a method to look it up from `SPIClass`.

user/tests/wiring/spi_master_slave/spi_master/test.mk
```diff
@@ -34,4 +34,9 @@ USE_CS_VAL=255
 endif
 
 CFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
-CXXFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
\ No newline at end of file
+CXXFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
+
+ifeq ("${SPI_SLEEP}","y")
```
avtolstoy, line 39: I would argue that testing this functionality should be enabled by default and there shouldn't be a conditional at all.

user/tests/wiring/spi_master_slave/spi_slave/test.mk
```diff
@@ -34,4 +34,9 @@ USE_CS_VAL=255
 endif
 
 CFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
-CXXFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
\ No newline at end of file
+CXXFLAGS += -DUSE_SPI=${USE_SPI_VAL} -DUSE_CS=${USE_CS_VAL}
+
+ifeq ("${SPI_SLEEP}","y")
```
avtolstoy, line 39: Same here.

hal/src/stm32f2xx/spi_hal.c
avtolstoy: I'd argue that when going to sleep we should wait instead in order not to leave the remote peripheral with an unfinished transfer and inconsistent data.

Also endless loop. `dma_aborted` is normally `0` and `hal_spi_transfer_dma_cancel()` will not do anything if there is not an ongoing transfer.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -412,108 +418,129 @@ void HAL_SPI_Info(HAL_SPI_Interface spi, hal_spi_info_t* info, void* reserved) {
...
 
     return 0;
 }
 
-int32_t HAL_SPI_Acquire(HAL_SPI_Interface spi, const HAL_SPI_AcquireConfig* conf) {
+int hal_spi_sleep(hal_spi_interface_t spi, bool sleep, void* reserved) {
+    if (sleep) {
+        CHECK_TRUE(hal_spi_is_enabled(spi), SYSTEM_ERROR_NONE);
+        hal_spi_transfer_dma_cancel(spi);
+        while (spiMap[spi].transmitting);
```
avtolstoy, line 527: Same here. See comments from STM32 implementation.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -522,19 +523,17 @@ int32_t hal_spi_set_settings(hal_spi_interface_t spi, uint8_t set_default, uint8
...
+        spiUninit(spi);
+        // spiUninit() clears pin function, retrieve it.
         HAL_Set_Pin_Function(spiMap[spi].sck_pin, PF_SPI);
         HAL_Set_Pin_Function(spiMap[spi].mosi_pin, PF_SPI);
         HAL_Set_Pin_Function(spiMap[spi].miso_pin, PF_SPI);
-        spiMap[spi].suspended = true;
+        spiMap[spi].state = HAL_SPI_STATE_SUSPENDED;
     } else {
-        CHECK_TRUE(spiMap[spi].suspended, SYSTEM_ERROR_NONE);
+        CHECK_TRUE(spiMap[spi].state == HAL_SPI_STATE_SUSPENDED, SYSTEM_ERROR_NONE);
```
avtolstoy, line 535: Let's make sure we have the same behavior across all APIs. Sometimes we return INVALID_STATE, sometimes it's ERROR_NONE.

PR #2134, opened 2020-06-19
Fixes the issue that gen3 devices cannot be woken up by RTC.

XuGuohui:
### Problem

`System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(3s));` doesn't work on Gen3 platforms.

### Solution

This issue is caused by not clearing RTC event in its interrupt handler. After device being woken up in `hal_sleep_enter()`, it will repeat entering the RTC interrupt handler.

### Steps to Test

- Build and flash the example below
- Build and flash the test application in `user/tests/wiring/sleep20`

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(3s));
}

void loop() {
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/timer_hal.cpp
+++ b/hal/src/nRF52840/timer_hal.cpp
@@ -270,6 +270,7 @@ int hal_timer_deinit(void* reserved) {
 extern "C" void RTC_IRQ_HANDLER(void) {
     // Handle overflow.
     if (nrf_rtc_event_pending(RTC_INSTANCE, NRF_RTC_EVENT_OVERFLOW)) {
+        nrf_rtc_event_clear(RTC_INSTANCE, NRF_RTC_EVENT_OVERFLOW);
         // Disable OVERFLOW interrupt to prevent lock-up in interrupt context while mutex is locked from lower priority
         // context and OVERFLOW event flag is stil up. OVERFLOW interrupt will be re-enabled when mutex is released -
         // either from this handler, or from lower priority context, that locked the mutex.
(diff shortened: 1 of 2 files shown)


hal/src/nRF52840/timer_hal.cpp
```diff
@@ -284,6 +285,16 @@ extern "C" void RTC_IRQ_HANDLER(void) {
         sTimerMicrosAtLastOverflow = getCurrentTime();
         sTickCountAtLastOverflow = DWT->CYCCNT;
         __set_PRIMASK(pri);
+    } else if (nrf_rtc_event_pending(RTC_INSTANCE, NRF_RTC_EVENT_TICK)) {
```
avtolstoy, line 288: Do we know which event in particular causes this? While it is fine to just clear everything in the handler, we should have been able to do that in `hal_timer_init()` after waking up from sleep and enable only what is required.

XuGuohui: We make use of the `NRF_RTC_EVENT_COMPARE_0` event. Clearing events in `hal_timer_init()` doesn't work after waking up from sleep, as the interrupt handler is executed immediately following `__enable_irq();`, which is before `hal_timer_init()` being executed.

avtolstoy: That should not be happening though, right? We unbump the priority level of both the GPIOTE and RTC interrupts right before calling `__enable_irq()`. If they do get triggered that's a bug in the sleep code. This has also been working previously just fine for quite a while. This is not new code.

avtolstoy: Just to clarify: I meant that calling `__enable_irq()` at that point in sleep code does not unblock interrupts with priorities `>= 2` because we are still under effect of `BASEPRI` at that point. We do re-enable interrupts with priority `>= 5` while waiting for SoftDevice to do certain things, but that should also not apply to RTC as it's configured normally at 6.

PR #2136, opened 2020-06-22
Implement sleep function for external flash

YutingYou:
### Feature

Implement sleep function for external flash

### Steps to Test

Test the sleep API with the example application, make sure the sleep API won't affect the behavior of the file system.

### Example App

```c
void setup(void) {
}

void loop() {
    static int index = 0;
    EEPROM.begin();
    EEPROM.write(10, '0' + index++ % 10);
    Log.info("Read: %c", EEPROM.read(10));
    EEPROM.end();
    hal_exflash_sleep(true, nullptr);
    delay(1000);
    hal_exflash_sleep(false, nullptr);
}
```

### References

[CH55886]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/exflash_hal.c
+++ b/hal/src/nRF52840/exflash_hal.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Particle Industries, Inc.  All rights reserved.
+ * Copyright (c) 2020 Particle Industries, Inc.  All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -155,16 +155,14 @@ static int configure_memory() {
 
     // Send reset enable
     err_code = exflash_qspi_cinstr_xfer(&cinstr_cfg, NULL, NULL);
-    if (err_code)
-    {
+    if (err_code) {
         return -1;
     }
 
     // Send reset command
     cinstr_cfg.opcode = QSPI_STD_CMD_RST;
     err_code = exflash_qspi_cinstr_xfer(&cinstr_cfg, NULL, NULL);
-    if (err_code)
-    {
+    if (err_code) {
         return -2;
     }
 
@@ -183,8 +181,7 @@ static int configure_memory() {
 #error "Unsupported platform for external flash"
 #endif
 
-    if (err_code)
-    {
+    if (err_code) {
         return -3;
     }
 
@@ -203,8 +200,7 @@ static int exit_secure_otp() {
     return exflash_qspi_cinstr_quick_send(QSPI_MX25_CMD_EXSO, 1, NULL);
 }
 
-int hal_exflash_init(void)
-{
+int hal_exflash_init(void) {
     int ret = 0;
 
     // PATCH: Initialize CS pin, external memory discharge
@@ -240,16 +236,13 @@ int hal_exflash_init(void)
     };
 
     ret = nrfx_qspi_init(&config, NULL, NULL);
-    if (ret)
-    {
-        if (ret == NRFX_ERROR_TIMEOUT)
-        {
+    if (ret) {
+        if (ret == NRFX_ERROR_TIMEOUT) {
             // Could have timed out because the flash chip might be in an ongoing program/erase operation.
             // Suspend it.
             LOG_DEBUG(TRACE, "QSPI NRFX timeout error. Suspend pgm/ers op.");
             ret = hal_exflash_special_command(HAL_EXFLASH_COMMAND_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, NULL, NULL, 0);
-            if (ret)
-            {
+            if (ret) {
                 goto hal_exflash_init_done;
             }
         }
@@ -259,14 +252,12 @@ int hal_exflash_init(void)
 
     // Wake up external flash from deep power down mode
     ret = hal_exflash_special_command(HAL_EXFLASH_COMMAND_NONE, HAL_EXFLASH_COMMAND_WAKEUP, NULL, NULL, 0);
-    if (ret)
-    {
+    if (ret) {
         goto hal_exflash_init_done;
     }
 
     ret = configure_memory();
-    if (ret)
-    {
+    if (ret) {
         goto hal_exflash_init_done;
     }
 
@@ -275,8 +266,7 @@ hal_exflash_init_done:
     return ret;
 }
 
-int hal_exflash_uninit(void)
-{
+int hal_exflash_uninit(void) {
     hal_exflash_lock();
 
     nrfx_qspi_uninit();
@@ -294,8 +284,7 @@ int hal_exflash_uninit(void)
     return 0;
 }
 
-int hal_exflash_write(uintptr_t addr, const uint8_t* data_buf, size_t data_size)
-{
+int hal_exflash_write(uintptr_t addr, const uint8_t* data_buf, size_t data_size) {
     hal_exflash_lock();
     int ret = hal_flash_common_write(addr, data_buf, data_size,
                                      &perform_write, &hal_flash_common_dummy_read);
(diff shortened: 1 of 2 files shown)


hal/src/nRF52840/exflash_hal.c
```diff
@@ -49,6 +50,9 @@ enum qspi_cmds_t {
 
 static const size_t MX25_OTP_SECTOR_SIZE = 4096 / 8; /* 4Kb or 512B */
 
+static bool qspi_initialized = false;;
```
avtolstoy, line 53: Nit: perhaps use a `state` instead of two almost exclusive boolean variables?

hal/src/nRF52840/exflash_hal.c
```diff
@@ -574,3 +584,32 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
+        hal_exflash_uninit();
+        qspi_suspended = true;
+    } else {
+        // Restore I2C
+        if (!qspi_suspended) {
+            hal_exflash_unlock();
+            return SYSTEM_ERROR_NONE;
+        }
+        hal_exflash_init();
+        qspi_suspended = false;
```
avtolstoy, line 611: This seems to be already done in `hal_exflash_init()`. The errors should also be propagated from `hal_exflash_init()`. It may very well potentially fail, which can as usual lead to filesystem corruption.

hal/src/nRF52840/exflash_hal.c
YutingYou: Fixed

hal/src/nRF52840/exflash_hal.c
YutingYou: Fixed, also check other qspi API inside the sleep API

hal/src/nRF52840/exflash_hal.c
```diff
@@ -574,3 +581,33 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
     hal_exflash_unlock();
     return ret;
 }
+
+int hal_exflash_sleep(bool sleep, void* reserved) {
+    hal_exflash_lock();
```
XuGuohui, line 586: We should acquire the lock if entering sleep mode is requested, and release it after exiting sleep mode.

YutingYou: Not very clear why we should lock the external flash so long time. And in stop mode, we also stop the scheduler.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -574,3 +581,33 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
+    if (sleep) {
+        // Suspend I2C
+        if (qspi_state != HAL_EXFLASH_STATE_ENABLED) {
+            hal_exflash_unlock();
+            return SYSTEM_ERROR_INVALID_STATE;
+        }
+        // Put external flash into sleep and disable QSPI peripheral
+        if (hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SLEEP, NULL, NULL, 0) != SYSTEM_ERROR_NONE ||
```
avtolstoy, line 594: 1. This is probably going to fail if there is an ongoing write/erase operation. See the `HAL_EXFLASH_COMMAND_SUSPEND_PGMERS` logic in `hal_exflash_init`. Probably same thing should be done here.
2. It's not clear whether both `hal_exflash_special_command` and `hal_exflash_uninit` calls should happen. Please refactor to state the intent clearer, as right now it seems we are calling `hal_exflash_uninit` only if sleep command fails. If the sleep command does not fail, do we need to reinitialize pins? As far as I remember there is a leak through one of them.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -574,3 +581,33 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
+        qspi_state = HAL_EXFLASH_STATE_SUSPENDED;
+    } else {
+        // Restore I2C
+        if (qspi_state != HAL_EXFLASH_STATE_SUSPENDED) {
+            hal_exflash_unlock();
+            return SYSTEM_ERROR_INVALID_STATE;
+        }
+        if (hal_exflash_init() != SYSTEM_ERROR_NONE) {
+            hal_exflash_unlock();
+            return SYSTEM_ERROR_INTERNAL;
```
avtolstoy, line 608: The error from `hal_exflash_init()` should be propagated instead.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -585,27 +554,33 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
         }
         // Put external flash into sleep and disable QSPI peripheral
-        if (hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SLEEP, NULL, NULL, 0) != SYSTEM_ERROR_NONE ||
-            hal_exflash_uninit() != SYSTEM_ERROR_NONE) {
-            hal_exflash_unlock();
-            return SYSTEM_ERROR_INTERNAL;
+        if (hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SLEEP, NULL, NULL, 0)) {
+            // It may fail as the flash chip might be in an ongoing program/erase operation.
+            // Suspend it.
+            int ret = hal_exflash_special_command(HAL_EXFLASH_COMMAND_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, NULL, NULL, 0);
```
avtolstoy, line 566: If we get here we are not putting the QSPI flash back into sleep state though.

XuGuohui: Just looked through the datasheet and found that the sleep command is not acceptable when program/erase is suspended. So if the first attempt executing the sleep command failed, we should return, leaving the external flash running.

avtolstoy: We can always abort the pending write/erase as well. It shouldn't cause any problems.

XuGuohui: So simply re-initialize the external flash if executing the sleep command failed and then re-execute the sleep command to put it into sleep state?

avtolstoy: I'd probably suggest:

1. Sleep command
2. If fails suspend write/erase
3. QSPI_STD_CMD_RSTEN (which will abort suspended operation)
4. Sleep
5. If fails, give up probably

XuGuohui: Good suggestion! Will do.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -571,6 +542,48 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
+        if (qspi_state != HAL_EXFLASH_STATE_ENABLED) {
+            hal_exflash_unlock();
+            return SYSTEM_ERROR_INVALID_STATE;
+        }
+        // Put external flash into Deep Power-down mode
+        if (hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SLEEP, NULL, NULL, 0)) {
+            // It may fail as the flash chip might be in an ongoing program/erase operation. Suspend it.
+            hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, NULL, NULL, 0);
+            // Abort the suspended write/erase operation.
+            hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_RESET, NULL, NULL, 0);
```
avtolstoy, line 563: Just one minor thing: after this command you might need to send `QSPI_STD_CMD_WRSR`. Double check the datasheet.

PR #2138, opened 2020-06-25
Add SPISettings while keeping __SPISettings as an alias for wiring layer

keeramis:
### Problem

__SPISettings naming is not compatible with Arduino layer

### Solution

Rename `__SPISettings` to `SPISettings` and still keep `__SPISettings` as an alias

### Steps to Test

Existing unit tests, but also something like this app should compile without errors 

```
#include "application.h"

void setup() {
    int32_t r = SPI.beginTransaction(__SPISettings(SPI_CLK_SYSTEM, SPI_MODE0, MSBFIRST));
    (void)r;
    int32_t r2 = SPI.beginTransaction(SPISettings(SPI_CLK_SYSTEM, SPI_MODE0, MSBFIRST));
    (void)r2;
}

void loop() {
}
```

### References

[ch55527](https://app.clubhouse.io/particle/story/55527/rename-spisettings-to-spisettings-while-keeping-spisettings-as-an-alias)
[ch54115](https://app.clubhouse.io/particle/story/54115/free-up-flash-space-removing-default-in-spisettings)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_spi.h
+++ b/wiring/inc/spark_wiring_spi.h
@@ -47,9 +47,9 @@ enum FrequencyScale
 };
 
 namespace particle {
-class __SPISettings : public Printable {
+class SPISettings : public Printable {
 public:
-  __SPISettings(uint32_t clock, uint8_t bitOrder, uint8_t dataMode)
+  SPISettings(uint32_t clock, uint8_t bitOrder, uint8_t dataMode)
     : default_{false},
       clock_{clock},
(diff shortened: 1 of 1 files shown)


keeramis:
There are also multiple locations `__SPISettings` can be changed, and potentially add unit tests for `SPISettings` instead of / in addition to `__SPISettings`, but don't think it is needed to change/add this every where.

PR #2139, opened 2020-06-25
[ublox] Do not run AT+COPS=0 when performing warm bootup if already registering/registered

avtolstoy:
### Problem

image

Despite being registered, an `AT+COPS=0` call under some conditions may take substantial amount of time, which will prolong the bootup times.

### Solution

1. Only perform `AT+COPS=0` conditionally. This is only implemented for uBlox-based devices, as Quectel seems to behave correctly and also we cannot reliably tell whether it's attempting to perform registration or not (it may report `+COPS: 0` even if it's not attempting to find an operator).
2. Additionally we've enabled packet domain errors (`AT+CGEREP`) which are useful in debugging connectivity issues
3. This PR also makes sure that failure to open muxer data channel used for PPP connections is treated as a non-recoverable failure, causing modem reset.

### Steps to Test

1. Make sure a cellular device is connected to the cloud
2. Reset
3. `AT+COPS=0` should not be executed
4. It should reconnect back to the cloud almost instantaneously without blocking for up to minutes in `AT+COPS=0` despite being already registered

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -1459,9 +1459,10 @@ void SaraNcpClient::connectionState(NcpConnectionState state) {
             return SYSTEM_ERROR_NONE;
         }, this);
         if (r) {
+            LOG(ERROR, "Failed to open data channel");
+            ready_ = false;
             connState_ = NcpConnectionState::DISCONNECTED;
         }
-        muxer_.resumeChannel(UBLOX_NCP_PPP_CHANNEL);
     }
 
     const auto handler = conf_.eventHandler();
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -120,6 +120,9 @@ using CidType = decltype(CellularGlobalIdentity::cell_id);
 const size_t UBLOX_NCP_R4_BYTES_PER_WINDOW_THRESHOLD = 512;
 const system_tick_t UBLOX_NCP_R4_WINDOW_SIZE_MS = 50;
 
+const int UBLOX_DEFAULT_CID = 1;
+const char UBLOX_DEFAULT_PDP_TYPE[] = "IP";
+
 
 } // anonymous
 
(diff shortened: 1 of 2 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1377,11 +1383,20 @@ int SaraNcpClient::registerNet() {
...
+    CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
 
     // NOTE: up to 3 mins (FIXME: there seems to be a bug where this timeout of 3 minutes
     //       is not being respected by u-blox modems.  Setting to 5 for now.)
-    r = CHECK_PARSER(parser_.execCommand(5 * 60 * 1000, "AT+COPS=0,2"));
+    if (copsState != 0) {
+        // If the set command with <mode>=0 is issued, a further set
+        // command with <mode>=0 is managed as a user reselection
+        r = CHECK_PARSER(parser_.execCommand(5 * 60 * 1000, "AT+COPS=0,2"));
+    }
```
keeramis, line 1399: One quick thought on this: As written in the attached AT manual, this will trigger a re-selection which we want to avoid for faster bootup times (which is all good). If we have to deal with roaming locations though, is it worth having an `AT+COPS=0` setting to in fact trigger a re-selection because otherwise what if it doesn't attach to network as expected. Am I making sense? Should we have a check that tells when to set `AT+COPS=0` and when not to set it? I know we have 10 min timeout, but even after 10min when we hit this line, looks like we won't do another `AT+COPS=0` in case which we may not register on a given network with edge case roaming scenarios and such?

avtolstoy: But after 10 mins and a modem reset the modem on its own will start a new registration process (auto-COPS), right?

keeramis: Hmm.. good point. That's also right. It should. (Reset is supposed to be more powerful than anything else.)

keeramis:
Please note the caveats that were discussed related to errors like `118` for ` Host is unreachable`. 
Log snippet:

```
0000098020 [lwip] TRACE: sent [IPCP ConfReq id=0x4 <addr 10.97.60.89> <msdns1 174.47.40.106> <msdns2 174.47.20.106>]
0000098023 [net.ppp.client] TRACE: TX: 29
0000098024 [lwip] TRACE: pppos_write[3]: len=26
0000099282 [net.ppp.client] TRACE: RX: 10
0000099284 [lwip] TRACE: pppos_input[3]: got 10 bytes
0000099286 [net.ppp.client] TRACE: RX: 28
0000099287 [lwip] TRACE: rcvd [IPCP ConfReq id=0x5]
0000099289 [lwip] TRACE: pppos_input[3]: got 28 bytes
0000099291 [net.ppp.ipcp] TRACE: input 4
0000099293 [lwip] TRACE: sent [IPCP ConfAck id=0x5]
0000099295 [net.ppp.client] TRACE: TX: 12
0000099296 [lwip] TRACE: pppos_write[3]: len=8
0000099298 [lwip] TRACE: rcvd [IPCP ConfAck id=0x4 <addr 10.97.60.89> <msdns1 174.47.40.106> <msdns2 174.47.20.106>]
.
.
```

PR #2142, opened 2020-06-26
Allow resetting the device via a low-level vendor-specific USB request

sergeuz:
### Problem

When testing Device OS in automated scenarios, we frequently need to recover devices under test from some blocked state caused by, for example, the network registration process.

### Solution

Allow resetting the device and cancelling the cloud connection directly in the USB HAL's ISR.

### Steps to Test

Using [this](https://github.com/particle-iot/particle-usb/tree/reset_req/ch57751) branch of particle-usb:
```js
const usb = require('particle-usb');

async function run() {
    let dev = null;
    try {
        const devs = await usb.getDevices();
        if (!devs.length) {
            throw new Error('No devices found');
        }
        dev = devs[0];
        await dev.open();
        // This should reset the device immediately
        // await dev.reset({ force: true });
        // This should cause the device to disconnect from the cloud immediately
        // await dev.disconnectFromCloud({ force: true });
    } catch (err) {
        console.error(err.stack);
        process.exit(1);
    } finally {
        if (dev) {
            await dev.close();
        }
    }
}

run();
```

### References

- [ch57751]

### DIFF
--- a/system/src/usb_control_request_channel.cpp
+++ b/system/src/usb_control_request_channel.cpp
@@ -24,6 +24,7 @@
 #include "system_task.h"
 #include "system_threading.h"
 
+#include "core_hal.h"
 #include "deviceid_hal.h"
 
 #include "spark_wiring_interrupts.h"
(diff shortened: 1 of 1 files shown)


system/src/usb_control_request_channel.cpp
```diff
@@ -521,6 +535,15 @@ bool particle::UsbControlRequestChannel::processVendorRequest(HAL_USB_SetupReque
     } else {
         // Host-to-device
         switch (req->wIndex) {
+        case CTRL_REQUEST_RESET: {
+            // TODO: This will cause an error at the host side
+            HAL_Core_System_Reset_Ex(RESET_REASON_USER, 0, nullptr);
```
avtolstoy, line 540: :fearful:

sergeuz: :shrug:

PR #2149, opened 2020-07-10
Implement the Ultra-low Power mode

XuGuohui:
The Ultra-Low Power mode aims to achieve extreme low and reliable power consumption for sleepy devices. The power consumption in this mode is similar to the Hibernate mode, besides, it enables that device can be woken up by many sources and can continue running after waking up.

### Example App

```c
void setup() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).duration(10s));
}

void loop() {

}
```
---

### Completeness

- [ ] Added documentation


XuGuohui:
Tests need to be added in `user/tests/wiring/sleep20`.

PR #2155, opened 2020-07-20
BLE cannot restart advertising when it is asleep.

XuGuohui:
### Problem

BLE cannot restart advertising when disconnected by peer device if sleep mode is enabled

### Steps to Test

1. Build and flash the below example application.
2. Connect to device using mobile App, e.g. LightBlue.
3. Disconnect from device using mobile App.
4. Device should restart advertising automatically.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
    BLE.advertise();
}

void loop() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).gpio(D0, RISING).ble());
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/sleep_hal.cpp
+++ b/hal/src/nRF52840/sleep_hal.cpp
@@ -53,6 +53,7 @@ using namespace particle;
 typedef struct WakeupSourcePriorityCache {
     uint32_t gpiotePriority;
     uint32_t rtc2Priority;
+    uint32_t blePriority;
 } WakeupSourcePriorityCache;
 
 static void bumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources, WakeupSourcePriorityCache* priority, uint32_t newPriority) {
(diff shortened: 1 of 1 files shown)


hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -379,6 +387,10 @@ static void fpu_sleep_prepare(void) {
     SPARK_ASSERT((fpscr & 0x07) == 0);
 }
 
+#ifdef DEBUG_BUILD
```
avtolstoy, line 390: I assume this is here only for going through this function more easily with a debugger. Let's remove it.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -590,6 +602,22 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
         // we are still under the effect of HAL_disable_irq that masked all but SoftDevice interrupts using BASEPRI
         __enable_irq();
 
+        /*
+         * SD_EVT_IRQn is set pending after __enable_irq() and SoftDevice interrupts being executed.
```
avtolstoy, line 606: Let's reword this, I had to read it multiple times and ask additional questions to understand how this works:

Wake-up on BLE events relies mainly on `SD_EVT_IRQn` interrupt, which will be disabled through NVIC (not masked) when entering `sd_nvic_critical_region_enter()`. Normally on BLE activity we would be woken up by other high-priority SoftDevice interrupts (e.g. timers, radio events etc), however `SD_EVT_IRQn` would be set as pending only after we unmask high-priority SoftDevice interrupts (`__enable_irq()`) and after the appropriate SoftDevice interrupt handlers finish executing. We would then immediately go back to sleep and wouldn't notice this change in `SD_EVT_IRQn` state. As a workaround, for the duration of the sleep, we'll be bumping the priority of `SD_EVT_IRQn`, so that it can wake us up from stop mode (`WFI`) as well.

avtolstoy: Alternatively, I think it might be viable to simply move pending IRQ checks under `__enable_irq()`.

PR #2159, opened 2020-07-21
FuelGauge waking up enhancement

XuGuohui:
By applying the changes, device can resume from sleep faster if system power manager is enabled.
### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void startup() {
    // conf.feature(SystemPowerFeature::DISABLE);
    System.setPowerConfiguration(conf);
}

STARTUP( startup() );

void setup() {
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(3s));
    LOG(TRACE, "Now: %d", millis());
}

void loop() {
}
```

Log if system power manager is not enabled:
>0000004027 [app] application.cpp:17, setup(): TRACE: Now: 4027

Log if system power manager is enabled:
>0000003537 [app] application.cpp:17, setup(): TRACE: Now: 3537
---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_power_manager.cpp
+++ b/system/src/system_power_manager.cpp
@@ -123,29 +123,25 @@ void PowerManager::update() {
 }
 
 void PowerManager::sleep(bool s) {
-  if (isRunning()) {
 #if HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL
-   if (detect_) {
+  if (detect_ && isRunning()) {
 #else
-    {
+  if (isRunning()) {
 #endif
-      FuelGauge gauge;
-      // When going into sleep we do not want to exceed the default charging parameters set
-      // by initDefault(), which will be reset in case we are in a DISCONNECTED state with
-      // PMIC watchdog enabled. Reset to the defaults and disable watchdog before going into sleep.
-      if (s) {
-        // Going into sleep
-        if (g_batteryState == BATTERY_STATE_DISCONNECTED) {
-          initDefault();
-        }
-        gauge.sleep();
-      } else {
-        // Wake up
+    // When going into sleep we do not want to exceed the default charging parameters set
+    // by initDefault(), which will be reset in case we are in a DISCONNECTED state with
+    // PMIC watchdog enabled. Reset to the defaults and disable watchdog before going into sleep.
+    if (s) {
+      // Going into sleep
+      if (g_batteryState == BATTERY_STATE_DISCONNECTED) {
         initDefault();
-        Event ev = Event::Wakeup;
-        os_queue_put(queue_, (const void*)&ev, CONCURRENT_WAIT_FOREVER, nullptr);
-        update();
       }
+      FuelGauge gauge;
+      gauge.sleep();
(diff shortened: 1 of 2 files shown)


system/src/system_power_manager.cpp
```diff
@@ -142,9 +142,14 @@ void PowerManager::sleep(bool s) {
       // Wake up
       initDefault();
       update();
-      gauge.wakeup();
-      // Delay for at least 500ms to make sure the fuelgauge is ready to function after woken up.
-      HAL_Delay_Milliseconds(500);
+      if (isRunning()) {
+        Event ev = Event::Wakeup;
+        os_queue_put(queue_, (const void*)&ev, CONCURRENT_WAIT_FOREVER, nullptr);
```
avtolstoy, line 147: This should be done before or combined with `update()` call. As otherwise the fuelgauge will anyway be accessed in `handleUpdate()`.

system/src/system_power_manager.cpp
```diff
@@ -123,28 +123,29 @@ void PowerManager::update() {
 }
 
 void PowerManager::sleep(bool s) {
+  if (isRunning()) {
 #if HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL
```
avtolstoy, line 127: Nit: formatting.

This can also probably be changed to:
```c++
#if HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL
if (detect_ && isRunning()) {
#else
if (isRunning()) {
#endif //HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL
```

system/src/system_power_manager.cpp
```diff
@@ -123,28 +123,29 @@ void PowerManager::update() {
...
+      // PMIC watchdog enabled. Reset to the defaults and disable watchdog before going into sleep.
+      if (s) {
+        // Going into sleep
+        if (g_batteryState == BATTERY_STATE_DISCONNECTED) {
+          initDefault();
+        }
+        gauge.sleep();
+      } else {
+        // Wake up
         initDefault();
```
avtolstoy, line 144: This should be moved to be processed in the PM thread as we've discussed.

system/src/system_power_manager.cpp
```diff
@@ -330,6 +331,10 @@ void PowerManager::loop(void* arg) {
         // Do not re-run DPDM
         self->initDefault(false);
         self->update_ = true;
+      } else if (ev == Event::Wakeup) {
```
avtolstoy, line 334: This event should trigger: `initDefault()` + `handleUpdate()` call here in this thread.

system/src/system_power_manager.cpp
```diff
@@ -330,6 +327,12 @@ void PowerManager::loop(void* arg) {
...
+        initDefault();
```
avtolstoy, line 331: Nit: might be a good idea to immediately acquire fuelgauge here rather than after `initDefault()`.

PR #2161, opened 2020-07-21
ensure client member vars are initialized before use

### LINKED ISSUE #2160: nif is used uninitialized in spark_wiring_client / TCP client
Running cross compiled firmware under valgrind gave a lot of warnings about the use of uninitialized memory, in TCPClient::available.

Line that causes it is:
```
    if(Network.from(nif).ready() && isOpen(d_->sock))
```

Turns out that `nif` is used without being initialized.

Fixed by changing in spark_wiring_client:
```
network_interface_t nif;
```
to
```
network_interface_t nif = 0;
```

A similar bug is found in spark_wiring_udp.h, where these two initializations should be added:
```

    /**
     * The network interface this UDP socket should bind to.
     */
    network_interface_t _nif = 0;

    /**
     * Set to non-zero if the buffer was dynamically allocated by this class.
     */
    uint8_t _buffer_allocated = false;
```

elcojacobs:
fixes #2160

### DIFF
--- a/wiring/inc/spark_wiring_client.h
+++ b/wiring/inc/spark_wiring_client.h
@@ -27,7 +27,7 @@
 
 class Client : public Stream {
 protected:
-    network_interface_t nif;
+    network_interface_t nif = 0;
 
 public:
   virtual int connect(IPAddress ip, uint16_t port, network_interface_t=0) =0;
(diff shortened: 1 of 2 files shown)


avtolstoy:
Thanks for a PR. I've just merged #2167 which fixed a couple of additional things on top.

PR #2162, opened 2020-07-22
Supports waking up from STOP/ULP mode by USART/Wi-FI/Cellular on Gen3 platforms

XuGuohui:
As the title describes.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(AUTOMATIC);

void setup() {
    Serial1.begin(115200);
}

void loop() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).usart(Serial1));
    // System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).network(Cellular));
    // System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).network(WiFi));
}
```

---

### Completeness

- [ ] Added documentation


wiring/inc/spark_wiring_usartserial.h
```diff
@@ -72,6 +72,10 @@ class USARTSerial : public Stream
   hal_usart_interface_t interface() const {
     return _serial;
   }
+
+  operator hal_usart_interface_t() const {
```
avtolstoy, line 76: Let's not add implicit conversion operators. There is already `interface()` function above as well.

hal/inc/hal_platform_compat.h
```diff
@@ -22,10 +22,13 @@
 
 #if PLATFORM_ID <= PLATFORM_P1 || PLATFORM_ID == PLATFORM_NEWHAL
     #define HAL_PLATFORM_WIFI 1
+    // Just define it
+    #define HAL_PLATFORM_WIFI_SERIAL
```
avtolstoy, line 26: Seems confusing. Photon/P1 use SDIO.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -84,13 +95,19 @@ static void unbumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSo
         } else if (source->type == HAL_WAKEUP_SOURCE_TYPE_BLE) {
             NVIC_SetPriority(SD_EVT_IRQn, priority->blePriority);
             NVIC_DisableIRQ(SD_EVT_IRQn);
+        } else if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            NVIC_SetPriority(UARTE0_UART0_IRQn, priority->usart0Priority);
+            nrf_uarte_int_disable(NRF_UARTE0, NRF_UARTE_INT_RXDRDY_MASK);
+        } else if (source->type == HAL_WAKEUP_SOURCE_TYPE_NETWORK) {
+            NVIC_SetPriority(UARTE1_IRQn, priority->usart1Priority);
+            nrf_uarte_int_disable(NRF_UARTE0, NRF_UARTE_INT_RXDRDY_MASK);
```
avtolstoy, line 103: Typo. UARTE0->UARTE1

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -133,6 +165,21 @@ static int constructBleWakeupReason(hal_wakeup_source_base_t** wakeupReason) {
     return SYSTEM_ERROR_NONE;
 }
 
+static int constructNetworkWakeupReason(hal_wakeup_source_base_t** wakeupReason, network_interface_index index) {
```
avtolstoy, line 168: Alternatively, we could hide this in system sleep layer and pass HAL just USART wakeupsource to make it more generalized and then on wakeup convert usart -> network.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -265,7 +312,33 @@ static void configRtcWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
...
+        source = source->next;
+    }
+}
+
+static void configNetworkWakeupSource(const hal_wakeup_source_base_t* wakeupSources) {
+    auto source = wakeupSources;
+    while (source) {
+        if (source->type == HAL_WAKEUP_SOURCE_TYPE_NETWORK) {
+            nrf_uarte_int_disable(NRF_UARTE1, NRF_UARTE_INT_RXDRDY_MASK);
+            nrf_uarte_event_clear(NRF_UARTE1, NRF_UARTE_EVENT_RXDRDY);
```
avtolstoy, line 333: This may cause problems for code that is blocked on USART waiting for incoming data. We've cleared an event despite potentially having received data already leaving the thread waiting on the data still in a blocked state.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -844,14 +981,6 @@ static int enterHibernateMode(const hal_sleep_config_t* config, hal_wakeup_sourc
         wakeupSource = wakeupSource->next;
     }
 
-    // Disable PWM
-    nrf_pwm_disable(NRF_PWM0);
```
avtolstoy, line 848: Do we not need this?

wiring/inc/spark_wiring_usartserial.h
XuGuohui: I was thinking that we have the same operators for network class, to be consistent and convenient for user application to just pass `Serial1` as the `.usart()` parameter.

hal/inc/hal_platform_compat.h
XuGuohui: How about defining `HAL_PLATFORM_WIFI_SDIO` and asserting if `!defined HAL_PLATFORM_WIFI_SERIAL && !defined HAL_PLATFORM_WIFI_SDIO` in `hal_platform.h`

hal/src/nRF52840/sleep_hal.cpp
XuGuohui: Good catch! 👍

hal/src/nRF52840/sleep_hal.cpp
XuGuohui: System sleep layer is not HAL specific. If new introduced platform didn't use USART as the interface to Cellular or Wi-Fi modem, we need to refactor it.

hal/src/nRF52840/sleep_hal.cpp
XuGuohui: We've called `hal_pwm_sleep(true, nullptr);` before getting here.

hal/inc/hal_platform_compat.h
avtolstoy: There is already `HAL_PLATFORM_NCP_AT` as well.

wiring/inc/spark_wiring_usartserial.h
avtolstoy: In that case it's better for the sleep configuration object to take wiring class type instead and perform necessary conversions.

hal/src/nRF52840/sleep_hal.cpp
avtolstoy: Sounds reasonable.

hal/src/nRF52840/sleep_hal.cpp
avtolstoy: :+1:

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -84,13 +95,19 @@ static void unbumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSo
         } else if (source->type == HAL_WAKEUP_SOURCE_TYPE_BLE) {
             NVIC_SetPriority(SD_EVT_IRQn, priority->blePriority);
             NVIC_DisableIRQ(SD_EVT_IRQn);
+        } else if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            NVIC_SetPriority(UARTE0_UART0_IRQn, priority->usart0Priority);
+            nrf_uarte_int_disable(NRF_UARTE0, NRF_UARTE_INT_RXDRDY_MASK);
```
avtolstoy, line 100: We cannot just disable it. The state of RXDRDY should be checked, cached and then restored otherwise I think there might still be a case where the threads blocked waiting a READABLE event will not get it.

Same https://github.com/particle-iot/device-os/pull/2162/files#diff-624879f8eaa372ccb534536b578173daR103

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -220,8 +223,22 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
     // We need to do this before disabling systick/interrupts, otherwise
     // there is a high chance of a deadlock
     if (config->mode == HAL_SLEEP_MODE_ULTRA_LOW_POWER) {
+        int skipUsart = -1;
```
avtolstoy, line 226: This should not be needed on Gen 2, right? We are entering a low power mode on top of regular WFI/WFE where only EXTI and RTC work anyway.

XuGuohui: We've introduced a network flag `SystemSleepNetworkFlag::INACTIVE_STANDBY`, which allow the network modem to be on while the device is asleep. When the ULP mode is specified, along with this flag, we should skip suspending the USART that is used by the modem.

avtolstoy: I understand that, however due to the fact that we are now turning off 1.2V power domain both in STOP and ULP modes, this does not make sense anyway, because the UART is not running. HSE/LSI are also off.

Moreover, we are already correctly putting the UART into a proper state in `cellular_pause()` on Gen 2.

XuGuohui: That makes perfect sense. 👍

avtolstoy: So, I propose we rollback this change (at least for now), also since this PR targets Gen 3.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -893,6 +908,13 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
         __set_BASEPRI(basePri);
     }
 
+    if (usartRxdRdy) {
```
avtolstoy, line 911: Simply `|| networkRxRdy`?

XuGuohui: We need to distinguish `NRF_UARTE0` and `NRF_UARTE1` anyway.

PR #2164, opened 2020-07-25
Only run AT+COPS=0 if currently de-registered, to avoid PLMN reselection

technobly:
### Problem

COPS=0 running when not de-registered.

### Solution

Only check for COPS: 2 mode and switch to automatic registration when deregistered.

---

### Completeness

- [x] N/A Added documentation

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -1192,15 +1192,14 @@ int QuectelNcpClient::registerNet() {
     connectionState(NcpConnectionState::CONNECTING);
 
     auto resp = parser_.sendCommand("AT+COPS?");
-    int copsState = -1;
+    int copsState = 2;
     r = CHECK_PARSER(resp.scanf("+COPS: %d", &copsState));
     CHECK_TRUE(r == 1, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
     r = CHECK_PARSER(resp.readResult());
     CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
 
-    if (copsState != 0) {
-        // Only run AT+COPS=0 if not currently registering, otherwise we will
-        // perform PLMN reselection
+    if (copsState != 0 && copsState != 1) {
+        // Only run AT+COPS=0 if currently de-registered, to avoid PLMN reselection
         // NOTE: up to 3 mins
         r = CHECK_PARSER(parser_.execCommand(QUECTEL_COPS_TIMEOUT, "AT+COPS=0,2"));
         // Ignore response code here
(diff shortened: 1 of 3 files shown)


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1192,15 +1192,14 @@ int QuectelNcpClient::registerNet() {
...
+    int copsState = 2;
     r = CHECK_PARSER(resp.scanf("+COPS: %d", &copsState));
     CHECK_TRUE(r == 1, SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED);
     r = CHECK_PARSER(resp.readResult());
     CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
 
-    if (copsState != 0) {
-        // Only run AT+COPS=0 if not currently registering, otherwise we will
-        // perform PLMN reselection
+    if (copsState == 2) {
```
avtolstoy, line 1201: Should we instead use `copsState !=0 && copsState != 1` to also handle potentially cases where `AT+COPS?` may return something other than 0, 1 or 2?

keeramis: image

I think it should be fine? At least for mode `4`, I think it's ok, but idk what mode `3` does.

technobly: Mode 3 just sets the format (numeric/alphanumeric).  We actually set this to numeric for MCCMNC reporting.  It believe it should never return +COPS: 3.

avtolstoy: Well, there is 4 as well.

keeramis: Actually may be good to use `copsState !=0 && copsState != 1` justtt in case we get "bad" values from COPS?

technobly: `copsState != 0 && copsState != 1 && copsState != 3 && copsState != 4` then? :D

avtolstoy: 0 and 1 should probably be enough, right? Because we want to avoid PLMN reselection only in those cases. Anything else - fair game.

technobly: I removed 3... because if for some odd reason it throws back a 3, well then I don't trust it... and we can set COPS=0.

avtolstoy: I'm mainly offering this because the appnote that we've kind of based this on from ublox clearly specifies that the action needs to be taken on COPS != 0, so if we are adding something to support 1, IMO we should specifically target that.

technobly: Updated to `copsState != 0 && copsState != 1` as discussed.

PR #2165, opened 2020-07-27
[Gen3] Workaround when unable to obtain DNS servers from remote

keeramis:
### Problem

Sometimes the device may not the DNS servers from the remote in which case we don't progress further.

### Solution

Start off by requesting public primary and secondary DNS servers (in this case, google's) which will help when we are unable to obtain DNS servers from remote (Separately, if the remote changes them, they are overwritten.)

### Example App

```c
void setup() {
}

void loop() {
}
```

### References

[ch59747](https://app.clubhouse.io/particle/story/59747/gen3-hard-code-dns-servers-if-unable-to-obtain-automatically)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/ppp_client.cpp
+++ b/hal/network/lwip/ppp_client.cpp
@@ -40,6 +40,10 @@ extern "C" {
 
 LOG_SOURCE_CATEGORY("net.ppp.client");
 
+/* Public Google DNS Servers */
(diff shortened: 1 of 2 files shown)


keeramis:
An example looks like this:
```
0000008109 [lwip] TRACE: sent [IPCP ConfReq id=0x1 <addr 0.0.0.0> <msdns1 8.8.8.8> <msdns2 8.8.4.4>]
```

PR #2166, opened 2020-07-29
if_power_state_t is an enum

larseggert:
Fixes this error:
```
../hal/network/api/ifapi.h:261:36: error: unknown type name 'if_power_state_t'
  261 | int if_get_power_state(if_t iface, if_power_state_t* state);
      |                                    ^~~~~~~~~~~~~~~~
```

### DIFF
--- a/wiring/inc/spark_wiring_client.h
+++ b/wiring/inc/spark_wiring_client.h
@@ -27,7 +27,7 @@
 
 class Client : public Stream {
 protected:
-    network_interface_t nif;
+  network_interface_t nif_ = NETWORK_INTERFACE_ALL;
 
 public:
   virtual int connect(IPAddress ip, uint16_t port, network_interface_t=0) =0;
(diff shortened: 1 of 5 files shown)


avtolstoy:
I assume this issue comes up when using ifapi from C code. The correct way to resolve this is by making all these enums into typedefs:

```c++
typedef enum if_power_state_t {
    //
} if_power_state_t;
```

larseggert:
Yes, this is from C.

larseggert:
@avtolstoy I updated the PR

hal/network/api/ifapi.h
```diff
@@ -29,7 +29,7 @@ extern "C" {
 typedef void* if_t;
 #endif /* IF_T_DEFINED */
 
-enum if_flags_t {
+typedef enum {
```
avtolstoy, line 32: It should be preferrable to use named enums:
```c++
typedef enum if_flags_t {
    //
} if_flags_t;
```

PR #2172, opened 2020-08-12
Support woken up by analog pin on Gen2 platforms.

XuGuohui:
As the title describes. But unlike the same wakeup source on Gen3, this wakeup source is only valid when it is used accompany with the STOP mode only.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {
    static system_tick_t now = millis();
    if (millis() - now > 5000) {
        now = millis();
        System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).analog(A0, 1500, AnalogInterruptMode::ABOVE));
    } else {
        LOG(INFO, "A0: %d", analogRead(A0));
        delay(1000);
    }
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/stm32f2xx/sleep_hal.cpp
+++ b/hal/src/stm32f2xx/sleep_hal.cpp
@@ -64,6 +64,23 @@ static int constructRtcWakeupReason(hal_wakeup_source_base_t** wakeupReason) {
     return SYSTEM_ERROR_NONE;
 }
 
+static int constructAnalogWakeupReason(hal_wakeup_source_base_t** wakeupReason, pin_t pin) {
+    if (wakeupReason) {
+        auto lpcomp = (hal_wakeup_source_lpcomp_t*)malloc(sizeof(hal_wakeup_source_lpcomp_t));
+        if (lpcomp) {
+            lpcomp->base.size = sizeof(hal_wakeup_source_lpcomp_t);
+            lpcomp->base.version = HAL_SLEEP_VERSION;
+            lpcomp->base.type = HAL_WAKEUP_SOURCE_TYPE_LPCOMP;
+            lpcomp->base.next = nullptr;
+            lpcomp->pin = pin;
+            *wakeupReason = reinterpret_cast<hal_wakeup_source_base_t*>(lpcomp);
+        } else {
+            return SYSTEM_ERROR_NO_MEMORY;
+        }
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
 static int configGpioWakeupSource(const hal_wakeup_source_base_t* wakeupSources) {
     auto source = wakeupSources;
     while (source) {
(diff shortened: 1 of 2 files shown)


platform/MCU/shared/STM32/src/hw_ticks.c
```diff
@@ -66,7 +66,9 @@ uint64_t GetSystem1MsTick64()
     const int is = __get_PRIMASK();
     __disable_irq();
 
-    millis = system_millis + (DWT->CYCCNT - system_millis_clock) / SYSTEM_US_TICKS / 1000;
+    system_tick_t elapsedTicks = (DWT->CYCCNT >= system_millis_clock) ? (DWT->CYCCNT - system_millis_clock) : (sizeof(uint32_t) - system_millis_clock + DWT->CYCCNT);
```
avtolstoy, line 69: This is not going to work as intended. CYCCNT is a 32 bit counter that ticks at 120MHz and thus wraps every ~30 seconds. The check for `DWT->CYCNT >= system_millis_clock` doesn't really tell us anything.

Resetting `system_millis_clock` to current `DWT->CYCCNT` on wakeup should be sufficient.

platform/MCU/shared/STM32/src/hw_ticks.c
```diff
@@ -89,8 +91,9 @@ system_tick_t GetSystem1UsTick()
 
     base_millis = system_millis;
     base_clock = system_millis_clock;
+    system_tick_t elapsedTicks = (DWT->CYCCNT >= base_clock) ? (DWT->CYCCNT - base_clock) : (sizeof(uint32_t) - base_clock + DWT->CYCCNT);
```
avtolstoy, line 94: Same here.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -100,6 +117,90 @@ static int configGpioWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
...
+                    ADC_AnalogWatchdogThresholdsConfig(ADC1, th, 0);
+                    break;
+                }
+                case HAL_SLEEP_LPCOMP_BELOW: {
+                    ADC_AnalogWatchdogThresholdsConfig(ADC1, 4095, th);
+                    break;
+                }
+                case HAL_SLEEP_LPCOMP_CROSS: {
+                    if (currVol >= th) {
+                        ADC_AnalogWatchdogThresholdsConfig(ADC1, 4095, th);
```
avtolstoy, line 174: This can probably instead changed to `[th - X mV, th + X mV]` similar to how Gen 3 MCU does it.

avtolstoy (changes requested):

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -186,6 +302,42 @@ static int validateWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_source_b
     return SYSTEM_ERROR_NOT_SUPPORTED;
 }
 
+static void enterSleepSleep() {
```
avtolstoy, line 305: This should really have a better name :sweat_smile: Doesn't really tell you what it does.

Do we want a separate function? As far as I can see the only difference is that we may not need to call `PWR_EnterSTOPMode()` and keep HSE/HSI running. 

Let's also add comments on why we prefer HSI to HSE here. I wouldn't expect a lot of difference in power consumption.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -100,6 +117,90 @@ static int configGpioWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
...
+            }
+
+            // Interrupt configuration
+            ADC_ClearFlag(ADC1, ADC_FLAG_AWD);
+            ADC_ITConfig(ADC1, ADC_IT_AWD, ENABLE);
+
+            //select NVIC channel to configure
+            NVIC_InitTypeDef NVIC_InitStructure = {};
+            NVIC_InitStructure.NVIC_IRQChannel = ADC_IRQn;
+            NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 7;
```
avtolstoy, line 190: This will probably also have the same problem as https://github.com/particle-iot/device-os/pull/2173#discussion_r475970347

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: This should probably apply to `HAL_SLEEP_LPCOMP_ABOVE` and `HAL_SLEEP_LPCOMP_BELOW` as well.

avtolstoy: Those should be fine as-is, unless you want to introduce some margin to them, though it shouldn't be necessary.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: From the comment: https://github.com/particle-iot/device-os/pull/2173/files?file-filters%5B%5D=#r476527209, which approach would you prefer?

avtolstoy: Honestly, we could probably go either way. If I had to choose, I'd probably keep HSE.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -100,6 +117,97 @@ static int configGpioWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
...
+                case HAL_SLEEP_LPCOMP_ABOVE: {
+                    ADC_AnalogWatchdogThresholdsConfig(ADC1, th, 0);
+                    break;
+                }
+                case HAL_SLEEP_LPCOMP_BELOW: {
+                    ADC_AnalogWatchdogThresholdsConfig(ADC1, 4095, th);
+                    break;
+                }
+                case HAL_SLEEP_LPCOMP_CROSS: {
+                    if (currVol >= th) {
```
avtolstoy, line 174: I'd probably remove this altogether. As long as we guard [th - XmV, th + XmV] area it should be fine just to setup the analog watchdog and rely on it to trigger once we are out of the protected range.

So, we'd be setting up the thresholds as:
```c++
ADC_AnalogWatchdogThresholdsConfig(ADC1, th + hysteresis, th - hysteresis);
```

avtolstoy (changes requested):

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -186,7 +309,36 @@ static int validateWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_source_b
     return SYSTEM_ERROR_NOT_SUPPORTED;
 }
 
-static void enterStopSleep() {
+static void enterPlatformSleepMode() {
+    int basePri = __get_BASEPRI();
+    __set_BASEPRI(2 << (8 - __NVIC_PRIO_BITS));
```
avtolstoy, line 314: On second though, this will potentially break other wakeup sources, right? Unless we bump their priorities. Same with USART PR. Also, have we verified that we don't have any interrupts with priority 1 enabled by default?

avtolstoy (changes requested):

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: If doing as you proposed, the device may exit sleep mode immediately, as the `hysteresis` is 15mV usually and the current voltage is most likely out of the protected range.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: >this will potentially break other wakeup sources, right?

Good catch! I'll bump all wakeup sources' priority for the generic `__WFI()` mode.

>Also, have we verified that we don't have any interrupts with priority 1 enabled by default?

Good point! I dumped all the `NVIC->IP` and `NVIC->ISER` registers after entering the `hal_sleep_enter()` function and found that only the `RTC_WKUP_IRQn` is set with priority 1, but it is not enabled. So we can safely set `BASEPRI` to 2 and bump the priority to 1 for necessary wakeup sources.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -30,6 +30,28 @@
 #include "interrupts_hal.h"
 #include "check.h"
 
+static constexpr uint8_t GPIO_IRQn[] = {
```
avtolstoy, line 33: Nit: prefer anonymous namespace in C++ code.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -111,6 +133,11 @@ static int configGpioWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
             irqConf.keepHandler = 1;
             irqConf.keepPriority = 1;
             HAL_Interrupts_Attach(gpioWakeup->pin, nullptr, nullptr, gpioWakeup->mode, &irqConf);
+            
+            Hal_Pin_Info* pinMap = HAL_Pin_Map();
+            uint8_t pinSource = pinMap[gpioWakeup->pin].gpio_pin_source;
+            extiPriority[pinSource] = NVIC_GetPriority(static_cast<IRQn_Type>(GPIO_IRQn[pinSource]));
+            NVIC_SetPriority(static_cast<IRQn_Type>(GPIO_IRQn[pinSource]), 1);
```
avtolstoy, line 140: Nit: This can be done when calling `HAL_Interrupts_Attach()` instead in `IRQChannelPreemptionPriority`.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -30,6 +30,28 @@
...
+    EXTI9_5_IRQn,   //9
+    EXTI15_10_IRQn, //10
+    EXTI15_10_IRQn, //11
+    EXTI15_10_IRQn, //12
+    EXTI15_10_IRQn, //13
+    EXTI15_10_IRQn, //14
+    EXTI15_10_IRQn  //15
+};
+
+static uint8_t extiPriority[16];
```
avtolstoy, line 52: Nit: can be stored on stack perhaps and passed to the appropriate function.

avtolstoy (approved):

avtolstoy:
@XuGuohui I'd still potentially change https://github.com/particle-iot/device-os/pull/2172#discussion_r478854059 behavior, as relying on current ADC value seems to be fragile as well and moreover will prolong the time before we are sleeping.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: That would violate the behavior we have implemented for Gen3.

avtolstoy: Can you elaborate why? I'm pretty sure it works the same way on Gen 3.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: It can be, while by doing so it also introduces complexity on restoring the priority after waking up. For instance, if a pin is not attached an interrupt in user application while this pin is specified as wakeup source, we cannot call `HAL_Interrupts_Attach()` to restore the priority for this pin.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: On Gen3, we just set a threshold the LPCOM and the hysteresis is fixed and achieved by hardware. If the current analog value if either above or below the threshold and we specify the interrupt type to be `CROSS`, then the Gen3 device can enter sleep, because the current analog value is not crossing the threshold. But for Gen2, the current value is out of the guarded range, the analog watchdog to set NVIC pending, thus it cannot enter sleep mode actually.

avtolstoy: Okay, now I see. You are right that if we are immediately out of the protected range we'll get an interrupt and what we want instead is to only get it when we cross the threshold :+1:

PR #2173, opened 2020-08-12
Support woken up by USART on Gen2 platforms.

XuGuohui:
As the title describes. But unlike the same wakeup source on Gen3, this wakeup source is only valid when it is used accompany with the STOP mode only.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(AUTOMATIC);

void setup() {
}

void loop() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).usart(Serial1));
    LOG(TRACE, "Device wakes up.");
}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/stm32f2xx/sleep_hal.cpp
+++ b/hal/src/stm32f2xx/sleep_hal.cpp
@@ -104,6 +104,20 @@ static int constructAnalogWakeupReason(hal_wakeup_source_base_t** wakeupReason, 
         } else {
             return SYSTEM_ERROR_NO_MEMORY;
         }
+    return SYSTEM_ERROR_NONE;
+}
+
+static int constructUsartWakeupReason(hal_wakeup_source_base_t** wakeupReason, hal_usart_interface_t serial) {
+    auto usart = (hal_wakeup_source_usart_t*)malloc(sizeof(hal_wakeup_source_usart_t));
+    if (usart) {
+        usart->base.size = sizeof(hal_wakeup_source_usart_t);
+        usart->base.version = HAL_SLEEP_VERSION;
+        usart->base.type = HAL_WAKEUP_SOURCE_TYPE_USART;
+        usart->base.next = nullptr;
+        usart->serial = serial;
+        *wakeupReason = reinterpret_cast<hal_wakeup_source_base_t*>(usart);
+    } else {
+        return SYSTEM_ERROR_NO_MEMORY;
     }
     return SYSTEM_ERROR_NONE;
 }
@@ -281,6 +295,18 @@ static int configRtcWakeupSource(const hal_wakeup_source_base_t* wakeupSources) 
     return SYSTEM_ERROR_NONE;
 }
 
+static int configUsartWakeupSource(const hal_wakeup_source_base_t* wakeupSources, uint8_t* configured) {
+    auto source = wakeupSources;
+    while (source) {
+        if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            *configured = 1;
+            // We don't need to configure anything, as the RX interrupt is enabled when the USART is enabled.
+        }
+        source = source->next;
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
 static int validateGpioWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_source_gpio_t* gpio) {
     switch(gpio->mode) {
         case RISING:
@@ -335,6 +361,16 @@ static int validateNetworkWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_s
     return SYSTEM_ERROR_NONE;
 }
 
+static int validateUsartWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_source_usart_t* usart) {
+    if (!hal_usart_is_enabled(usart->serial)) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    if (mode != HAL_SLEEP_MODE_STOP) {
+        return SYSTEM_ERROR_NOT_SUPPORTED;
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
 static int validateWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_source_base_t* base) {
     if (base->type == HAL_WAKEUP_SOURCE_TYPE_GPIO) {
         return validateGpioWakeupSource(mode, reinterpret_cast<const hal_wakeup_source_gpio_t*>(base));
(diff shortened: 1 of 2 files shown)


platform/MCU/shared/STM32/src/hw_ticks.c
```diff
@@ -66,7 +66,9 @@ uint64_t GetSystem1MsTick64()
     const int is = __get_PRIMASK();
     __disable_irq();
 
-    millis = system_millis + (DWT->CYCCNT - system_millis_clock) / SYSTEM_US_TICKS / 1000;
+    system_tick_t elapsedTicks = (DWT->CYCCNT >= system_millis_clock) ? (DWT->CYCCNT - system_millis_clock) : (sizeof(uint32_t) - system_millis_clock + DWT->CYCCNT);
```
avtolstoy, line 69: See https://github.com/particle-iot/device-os/pull/2172/files#r475946225

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -175,17 +202,69 @@ static int validateNetworkWakeupSource(hal_sleep_mode_t mode, const hal_wakeup_s
...
         return validateRtcWakeupSource(mode, reinterpret_cast<const hal_wakeup_source_rtc_t*>(base));
     } else if (base->type == HAL_WAKEUP_SOURCE_TYPE_NETWORK) {
         return validateNetworkWakeupSource(mode, reinterpret_cast<const hal_wakeup_source_network_t*>(base));
+    } else if (base->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+        return validateUsartWakeupSource(mode, reinterpret_cast<const hal_wakeup_source_usart_t*>(base));
     }
     return SYSTEM_ERROR_NOT_SUPPORTED;
 }
 
+static void enterSleepSleep(uint8_t highAccuracy) {
```
avtolstoy, line 228: See https://github.com/particle-iot/device-os/pull/2172/files#r475966911

Also, this implementation is different and uses HSE instead. Which one are we using?

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -134,6 +149,18 @@ static int configRtcWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
     return SYSTEM_ERROR_NONE;
 }
 
+static int configUsartWakeupSource(const hal_wakeup_source_base_t* wakeupSources, uint8_t* configured) {
+    auto source = wakeupSources;
+    while (source) {
+        if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            *configured = 1;
+            // We don't need to configure anything, as the RX interrupt is enabled when the USART is enabled.
```
avtolstoy, line 157: I'm pretty sure we need to do at least something here. Everything still gets clocked and other interrupts than USART are enabled. This is not a problem with `PWR_Regulator_LowPower`, because only RTC and EXTI are clocked, but with the normal WFI mode this is not going to work. We resolve this problem on Gen 3 with BASEPRI. Something similar probably needs to be done here.

Moreover, any other USART event may wake us up.

avtolstoy (changes requested):

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: I will re-estimate the power consumption using HSI against using HSE. If the diversity is minimum, then let's use HSE instead, otherwise, we're going to using HSI for analog wakeup source and HSE for USART wakeup source. The reason we have to use HSE  for USART is that USART need accurate clock to guarantee the baudrate.

avtolstoy: Sounds good. I'd prefer that we used HSE as well if the difference is negligible. This will also allow to enter/exit sleep mode more quickly.

XuGuohui: For the USART wakeup source, we need to enable PLL so that the baudrate would not be affected, in which case it results in ~17mA average current. While using HSI only results in ~6mA average current. So we should use HSI for analog wakeup source conditionally, while use HSE along with PLL for USART wakeup source.

XuGuohui: Or use HSE for analog wakeup source as well, but need to disable PLL to optimize the power consumption.

avtolstoy: How much of a difference is HSE without PLL vs HSI?

XuGuohui: It's ~1mA higher than HSI if using HSE only.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -306,6 +445,44 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
         } else if (wakeupSource->type == HAL_WAKEUP_SOURCE_TYPE_GPIO) {
             auto gpioWakeup = reinterpret_cast<hal_wakeup_source_gpio_t*>(wakeupSource);
             HAL_Interrupts_Detach_Ext(gpioWakeup->pin, 1, nullptr);
+        } else if (wakeupSource->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            auto usartWakeup = reinterpret_cast<hal_wakeup_source_usart_t*>(wakeupSource);
+            // Enabled unwanted interrupts and unbump the interrupt priority
+            NVIC_InitTypeDef NVIC_InitStructure;
+            switch (usartWakeup->serial) {
```
avtolstoy, line 452: Should this partially be handled in `hal_usart_sleep()` instead perhaps?

avtolstoy (approved):

avtolstoy (changes requested):
See https://github.com/particle-iot/device-os/pull/2172#discussion_r478862227

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -306,6 +445,44 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
...
+                }
+                case HAL_USART_SERIAL5: {
+                    NVIC_InitStructure.NVIC_IRQChannel = UART5_IRQn;
+                    USART_ITConfig(UART5, USART_IT_TXE, ENABLE);
+                    break;
+                }
+#endif
+                default: return SYSTEM_ERROR_INVALID_ARGUMENT; // It should not reach here.
+            }
+            NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 7;
```
avtolstoy, line 482: Nit: magic number.

avtolstoy (changes requested):
Looks ok, however this PR does not have the set of changes to handle the BASEPRI in the context of RTC/EXTI unlike #2172. Let's make sure to pull-in necessary changeset.

hal/src/stm32f2xx/sleep_hal.cpp
XuGuohui: Ideally, this value should be the same as what we set in `usart_hal.c`, But there isn't a variable or a macro for this value in usart_hal.c. So I'll just leave a comment here for better understanding.

avtolstoy (approved):

PR #2175, opened 2020-08-18
Set QCFG commands after every COPS=0,2 

keeramis:
### Problem

BG96 falls back to search on 2G (if LTE-M is not available) even if the respective commands `QCFG` are set in initialization to work on LTE-M only

### Solution

`COPS=0,2` resets the `QCFG - nwscanmode` command to search all technologies. Therefore, setting `QCFG="nwscanmode"` to do LTE-M only every time after COPS is run. 

### Steps to Test

Run any tinker like app firmware with Quectel BG96 where there is no LTE-M coverage. The device should not fallback to 2G

### Example App

```c
void setup() {
}

void loop() {
}
```

### References

[CH60704](https://app.clubhouse.io/particle/story/60704/2g-fallback-is-not-disabled-on-bg96)
https://github.com/particle-iot/device-os/pull/2112
---

### Completeness

- [ ] Added documentation


YutingYou:
@keeramis We still probability run `AT+COPS=0` in some other conditions. `nwscanmode` and `iotopmode` should also be added.

YutingYou (approved):

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -965,7 +965,7 @@ int QuectelNcpClient::initReady(ModemState state) {
         CHECK(selectSimCard());
 
         // Just in case disconnect
-        int r = CHECK_PARSER(parser_.execCommand("AT+COPS=2"));
+        // int r = CHECK_PARSER(parser_.execCommand("AT+COPS=2"));
```
avtolstoy, line 968: Does everything still work correctly? e.g. changing EDRX/PSM settings while registered or QDSIM on EG91?

avtolstoy (approved):

PR #2176, opened 2020-08-19
Clear OTA slots after updating firmware modules

sergeuz:
### Problem

The bootloader prematurely clears OTA flags in the DCT before ensuring that the firmware modules have been successfully copied from the OTA section to the internal flash.

### Solution

Clear OTA flags **after** the firmware modules have been updated.

### Steps to Test

1. Compile and flash the bootloader from this branch to a Gen 2 or Gen 3 device.
2. Flash a system part to the device OTA.
3. Wait until the device resets to apply the update and reset it via the reset button.
4. The device should reapply the update after the reset.

### References

- [ch61683]

### DIFF
--- a/bootloader/src/main.c
+++ b/bootloader/src/main.c
@@ -33,11 +33,6 @@
 #include "dct.h"
 #include "feature_flags.h"
 
-#if PLATFORM_ID == 6 || PLATFORM_ID == 8
-#define LOAD_DCT_FUNCTIONS
-#include "bootloader_dct.h"
-#endif
-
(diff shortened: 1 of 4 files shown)


avtolstoy (approved):
Looks good. Tested both Gen 2 (Electron, Photon) and Gen 3 (BSoM) interrupting while bootloader is updating modules:
- Bootloader (Gen 3)
- System parts, application (Gen 2/3)
- SoftDevice (Gen 3)
- Compressed/combined (2.0.0-alpha.4 -> 2.0.0-rc.1) (Gen 3)

PR #2178, opened 2020-08-20
Allowing the ALT server public key to be set in DCT programatically

JamesHagerman:
### Problem

The documentation for `System Config [ set ]` suggests that the public server key for a device (including the host and key) can be updated from user applications as follows:

```
System.set(SYSTEM_CONFIG_SERVER_KEY, newKeyBytes, newKeyString.length());
```

This works as expected on Photon but was never updated for UDP devices.

### Solution

This PR updates the `stm32f2xx` and `nRF52840` HAL layers to update the server public key stored in DCT at `DCT_ALT_SERVER_PUBLIC_KEY_OFFSET` on UDP devices instead of updating the seemingly unused `DCT_SERVER_PUBLIC_KEY_OFFSET` location.

### Steps to Test

1. Generate a hex string for the production key file: `xxd -p -c 768 production.pub.der`
2. Insert that hex string into the test app, compile, and flash it to the test device.
3. Point test device at non-production environment using `particle server key <non-production-keyfile>`
4. Once the device has connected to the non-production environment, call the `change-server` function.
5. Device should update the key locally, reset itself, and connect to the production environment.

### Example App

```c
bool keyLoadRequired = false;
bool resetRequired = false;

int setServerKey(String unused) {
  keyLoadRequired = true;
  return 1;
}

void loadNewKey() {
  String newKeyString = "<key hex string>"; // Use: xxd -p -c 768 production.pub.der
  char newKeyBytes[768] = {0};
  for (unsigned int i = 0, j = 0; i < newKeyString.length(); i += 2, j += 1) {
    String byteString = newKeyString.substring(i, i + 2);
    newKeyBytes[j] = (char) strtol(byteString.c_str(), NULL, 16);
  }
  System.set(SYSTEM_CONFIG_SERVER_KEY, newKeyBytes, newKeyString.length());

  Particle.publish("dut-moving", "production", 60);
  resetRequired = true;
}

void setup() {
  Particle.function("change-server", setServerKey);
  delay(500);
  Particle.publish("dut-boot", "complete", 60);
}

void loop() {
  if (keyLoadRequired) {
    keyLoadRequired = false;
    loadNewKey();
  }

  if (resetRequired) {
    resetRequired = false;
    System.reset();
  }
}
```

### References

* [ch61883]


sergeuz (approved):
How about implementing the same for the device's private key for consistency?

JamesHagerman:
@sergeuz I haven't got a test for it but I pushed a change that should work.

sergeuz:
Looks good, thanks. You want this change to be included in 2.0.0, right?

JamesHagerman:
Yeah, it would be nice to have in 2.0.0. However, I'm glad I tested this because I found a bug.

TCP devices hold their Server Public Key and Server Address in the same DCT field: `server_public_key[768]`. That means it's trivial to update both the key and the address at the same time because `DCT_SERVER_PUBLIC_KEY_SIZE` is 768 bytes which includes both the key and the address.

However, UDP devices hold their Server Public Key and Server Address in separate DCT fields. The logic in `HAL_Set_System_Config()` caps the length of the data to be written. My code, as written, means that only 192 bytes will be written to DCT at `DCT_ALT_SERVER_PUBLIC_KEY_OFFSET`.

That works great for the key, but my intention was to change the server address as well since that's possible on TCP devices.

I'm reworking the code now.

hal/src/stm32f2xx/ota_flash_hal_stm32f2xx.cpp
```diff
@@ -613,16 +613,39 @@ int HAL_Set_System_Config(hal_system_config_t config_item, const void* data, uns
...
+            offset = DCT_DEVICE_PRIVATE_KEY_OFFSET;
+            length = DCT_DEVICE_PRIVATE_KEY_SIZE;
+        }
         break;
     case SYSTEM_CONFIG_SERVER_KEY:
-        offset = DCT_SERVER_PUBLIC_KEY_OFFSET;
-        length = DCT_SERVER_PUBLIC_KEY_SIZE;
+        if (udp) {
+            offset = DCT_ALT_SERVER_PUBLIC_KEY_OFFSET;
+            length = DCT_ALT_SERVER_PUBLIC_KEY_SIZE;
```
sergeuz, line 632: This line can be removed

sergeuz (approved):

PR #2180, opened 2020-08-25
Gen3: fixes USART wakeup source for the ULP mode

XuGuohui:
### Problem

Gen3: device might exit from the ULP mode immediately when USART is specified as the wakeup source

### Solution

Disable TX related interrupts before entering the ULP mode.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {
    delay(1000);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).usart(Serial1));
}
```

Device running the above application may wakes up from the the ULP mode immediately before applying the fix.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/sleep_hal.cpp
+++ b/hal/src/nRF52840/sleep_hal.cpp
@@ -419,25 +419,40 @@ static void configLpcompWakeupSource(const hal_wakeup_source_base_t* wakeupSourc
     }
 }
 
-static bool configUsartWakeupSource(const hal_wakeup_source_base_t* wakeupSources) {
-    bool ret = false;
+static uint32_t configUsartWakeupSource(const hal_wakeup_source_base_t* wakeupSources) {
+    uint32_t intFlags = NRF_UARTE0->INTENSET;
     auto source = wakeupSources;
     while (source) {
         if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            nrf_uarte_int_disable(NRF_UARTE0, NRF_UARTE_INT_ERROR_MASK |
+                      NRF_UARTE_INT_RXSTARTED_MASK |
+                      NRF_UARTE_INT_RXTO_MASK |
+                      NRF_UARTE_INT_ENDRX_MASK |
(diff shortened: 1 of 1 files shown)


avtolstoy (approved):

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -419,25 +419,40 @@ static void configLpcompWakeupSource(const hal_wakeup_source_base_t* wakeupSourc
...
         if (source->type == HAL_WAKEUP_SOURCE_TYPE_USART) {
+            nrf_uarte_int_disable(NRF_UARTE0, NRF_UARTE_INT_ERROR_MASK |
+                      NRF_UARTE_INT_RXSTARTED_MASK |
+                      NRF_UARTE_INT_RXTO_MASK |
+                      NRF_UARTE_INT_ENDRX_MASK |
+                      NRF_UARTE_INT_TXSTARTED_MASK |
+                      NRF_UARTE_INT_TXDRDY_MASK |
+                      NRF_UARTE_INT_TXSTOPPED_MASK |
+                      NRF_UARTE_INT_ENDTX_MASK);
+            nrf_uarte_event_clear(NRF_UARTE0, NRF_UARTE_EVENT_ERROR);
```
avtolstoy, line 435: I think clearing all the events is going to cause problems returning the UART into a working state because we'll miss some of them. Can we just:
1. Disable unnecessary interrupts
2. Clear pending IRQ state
3. Keep the events intact

Would that work?

avtolstoy (changes requested):

hal/src/nRF52840/sleep_hal.cpp
XuGuohui: That works. But the data need to be sent before entering ULP mode is sent after device waking up. We need to flush the TX buffer for ULP mode as well.

avtolstoy (approved):

PR #2181, opened 2020-08-28
[wiring] print: exclude printable objects from an overload taking integral and uint convertible types

### LINKED ISSUE #2124: Bug Report: Wifi returns 1 as localIP
## Bug Report
Bug Report: Wifi returns 1 as localIP

### Expected Behavior
localIP returns something like 0.0.0.0 or some other IP

What should it do?

### Observed Behavior
Returns 1

What did it actually do?

### Steps to Reproduce
```
#include "Particle.h"

void setup() {
  Serial.begin(9600);
  // Wait for a USB serial connection for up to 30 seconds
  waitFor(Serial.isConnected, 30000);
}
void loop(){
  Serial.println(WiFi.localIP());
}
```

How did you make it do that, and which system/user firmware versions were tested?
version deviceOS 1.5.2 on Particle Argon.

### References

Is it related to this problem? Do I need to test with SYSTEM_MODE(SEMI_AUTOMATIC)?
https://github.com/particle-iot/device-os/issues/1631

---

avtolstoy:
### Problem

For printable types (e.g. `IPAddress`) that also happen to be implicitly convertible to an integral or unsigned integer type (e.g. `bool`) `Print` overload printing integers will take precedence over the overload taking `const Printable&`.

For example:
```c++
Serial.println(WiFi.localIP());
```

This will print only `0` or `1`, instead of an actual address.

### Solution

As title says.

### Steps to Test

- unit tests

### Example App

N/A

### References

- Closes #2124 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/test/unit_tests/wiring/print.cpp
+++ b/test/unit_tests/wiring/print.cpp
@@ -47,6 +47,17 @@ private:
     size_t  size_;
 };
 
+class TestConvertiblePrintable: public Printable {
+public:
+    virtual size_t printTo(Print& p) const override {
+        return p.write("TestConvertiblePrintable");
+    }
+
+    operator bool() const {
+        return true;
+    }
+};
+
 } // namespace
 
 TEST_CASE("Print tests for all types") {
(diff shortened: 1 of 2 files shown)


XuGuohui (approved):

divrajBevie:
Hey @avtolstoy How do you exclude printable objects? I could not find this anywhere in the documentation.

divrajBevie:
My device is running on 2.0.1 and when I am flashign firmware via Serial, the IP is returning fine. However when I do an OTA with same firmware, The device returns 1 onto the LCD.

avtolstoy:
@divrajBevie This fix should be available as long as your app is compiled with Device OS >= 2.0.0. There should not be any behavior between how you flash the binary, most likely you are running an application built for earlier version of Device OS.

divrajBevie:
Ahhh Thanks @avtolstoy, thats explains it. Sorry an oversight on my behalf.

Issue #2093, opened 2020-05-05
I2C State transition from slave receiver to slave transmitter not supported. 


studiohsoftware:
<details>
  <summary><i>In slave receiver mode, a transition directly to slave transmitter without STOP condition from master will leave bytes in the rxBuffer that cannot be read, and Wire.available() returns a negative number. This is not an uncommon scenario. I2C memory chips expect a memory address write followed immediately by a series of reads without a STOP condition between. Emulating this behavior is currently not possible.</i></summary>
</details>

### Expected Behavior

The rxBuffer should be left in a known state when slave receiver transitions to slave transmitter.

### Observed Behavior

During slave receive, the event handler in i2c_hal.c modifies rxBuffer in the I2C_EVENT_SLAVE_BYTE_RECEIVED case.

Normally the master issues a STOP condition, and the I2C_EVENT_SLAVE_STOP_DETECTED happens. In this case the rxBuffer is updated so that it can then be read by the user sketch:

i2cMap[i2c]->rxIndexTail = i2cMap[i2c]->rxIndexHead;
i2cMap[i2c]->rxIndexHead = 0;

If however the master issues a READ without first issuing a STOP condition, event I2C_EVENT_SLAVE_TRANSMITTER_ADDRESS_MATCHED is fired instead of I2C_EVENT_SLAVE_STOP_DETECTED. This event does not fix up the rxBuffer, and if there is anything in the buffer, Wire.available() will return negative and the bytes cannot be read. 

To fix, simply add the rxBuffer fixup to the I2C_EVENT_SLAVE_TRANSMITTER_ADDRESS_MATCHED case as well.

### References

https://github.com/particle-iot/device-os/blob/develop/hal/src/stm32f2xx/i2c_hal.c

PR #2184, opened 2020-08-29
Gen2 I2C: fix the issue #2093.

XuGuohui:
### Problem

See: https://github.com/particle-iot/device-os/issues/2093.

### Solution

Notify user application on slave receiver worked for a cycle.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

#define I2C_SLAVE_ADDR  0x32

const auto slaveMsg = "This is slave.";
const auto masterMsg = "This is master.";

// Use Photon or Argon the slave device to test
#if PLATFORM_ID == PLATFORM_PHOTON || PLATFORM_ID == PLATFORM_ARGON
void receiveEvent(int howMany) {
    while (1 < Wire.available()) {
        char c = Wire.read();
        Serial1.print(c);
    }
    char x = Wire.read();
    Serial1.println(x);
}

void requestEvent() {
    Wire.write(slaveMsg);
}
#endif

void setup() {

#if PLATFORM_ID == PLATFORM_PHOTON || PLATFORM_ID == PLATFORM_ARGON
    Wire.begin(I2C_SLAVE_ADDR);
    Wire.onReceive(receiveEvent);
    Wire.onRequest(requestEvent);
#else
    int ret;
    Wire.begin();
    Wire.beginTransmission(I2C_SLAVE_ADDR);
    Wire.write("Hello!");
    ret = Wire.endTransmission(false); // Do not send the STOP bit.
    if (ret) {
        Log.error("endTransmission: %d", ret);
    }
    Wire.beginTransmission(I2C_SLAVE_ADDR);
    Wire.write(masterMsg);
    ret = Wire.endTransmission(false); // Do not send the STOP bit.
    if (ret) {
        Log.error("endTransmission: %d", ret);
    }
    Wire.requestFrom(I2C_SLAVE_ADDR, strlen(slaveMsg), true); // Send the stop bit.
    while (1 < Wire.available()) {
        char c = Wire.read();
        Serial1.print(c);
    }
    char x = Wire.read();
    Serial1.println(x);
#endif
}

void loop() {
}
```

### References

https://github.com/particle-iot/device-os/issues/2093

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/stm32f2xx/i2c_hal.c
+++ b/hal/src/stm32f2xx/i2c_hal.c
@@ -117,6 +117,7 @@ typedef struct stm32_i2c_info_t {
     size_t txIndexTail;
 
     uint8_t transmitting;
+    uint8_t receiving;
 
     void (*callback_onRequest)(void);
     void (*callback_onReceive)(int);
@@ -260,6 +261,7 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     i2cMap[i2c]->txIndexTail = 0;
 
     i2cMap[i2c]->transmitting = 0;
+    i2cMap[i2c]->receiving = 0;
 
     i2cMap[i2c]->ackFailure = false;
     i2cMap[i2c]->prevEnding = I2C_ENDING_UNKNOWN;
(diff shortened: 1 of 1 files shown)


hal/src/stm32f2xx/i2c_hal.c
```diff
@@ -962,12 +966,18 @@ static void HAL_I2C_EV_InterruptHandler(hal_i2c_interface_t i2c) {
...
-        i2cMap[i2c]->rxIndexHead = 0;
+        if (i2cMap[i2c]->transmitting == 1) { // Slave transmitter
+            i2cMap[i2c]->transmitting = 0;
+        }
 
-        if (NULL != i2cMap[i2c]->callback_onReceive) {
-            // alert user program
-            i2cMap[i2c]->callback_onReceive(i2cMap[i2c]->rxIndexTail);
+        if (i2cMap[i2c]->receiving) { // Slave receiver
+            i2cMap[i2c]->receiving = 0;
```
avtolstoy, line 974: Nit: move into a common function. We have 4 cases where we do exactly the same thing.

avtolstoy (changes requested):
We should add tests to `i2c_master_slave` suite to verify this behavior both on Gen 2 and Gen 3.

avtolstoy (approved):
LGTM.

@XuGuohui We are still missing tests for this though.

PR #2185, opened 2020-09-04
Twilio Fixes

technobly:
### Problem

A number of issues found while bringing up the new Twilio Super SIM.

### Solution

- [gen2] [gen3] adds fixes for UMNOPROF when using Twilio Super SIM [ch60727]
- [gen2] move CFUN=1,0 code within if (_init .. block
- ensure CIMI timeout is at least 10s on u-blox modems [ch61658]
- fixes running unit tests via ci/unit_tests.sh locally
- [gen3] delay registration intervening for twilio super sim for first 5 mins [ch60883]

### Steps to Test

- run both sets of unit tests

```
device-os $ source ci/install_boost.sh
device-os $ source ci/build_boost.sh
device-os $ ci/unit_tests.sh
```

- lots of manual testing with the app below to help facilitate resetting the UMNOPROF back to 0 with cold and warm boots

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

bool reset_umnoprof = false;

{
        if (clicks == 3) {
            reset_umnoprof = true;
        }
    }
}

void setup() {
    // Cellular.setActiveSim(EXTERNAL_SIM);
    Particle.connect();
}

void loop() {
    if (reset_umnoprof) {
        reset_umnoprof = false;
        Cellular.command(60000, "AT+CFUN=0");
        Cellular.command(1000, "AT+UMNOPROF=0");
        Cellular.command(1000, "AT+UMNOPROF?");
        Cellular.command(1000, "AT+CFUN=15");
    }
}
```

### References

- ch60727
- ch61658
- ch60883

---

### Completeness

### DIFF
--- a/ci/unit_tests.sh
+++ b/ci/unit_tests.sh
@@ -52,7 +52,8 @@ set -x -e
 # Run CMake-based unit tests
 cd $unit_test_dir
 rm -rf .build/*
-mkdir .build -p && cd .build/
+mkdir -p .build/
+cd .build/
 cmake $cmake_args ..
 # FIXME: coveralls disabled for now as the CI builds for some reason are failing
 make all test # coveralls
--- a/hal/inc/cellular_hal_utilities.h
+++ b/hal/inc/cellular_hal_utilities.h
@@ -1,0 +1,52 @@
+/*
+ * MDM_HAL_UTILITIES is reserved for small pure C/C++ utilities that can be used from cellular_hal or mdm_hal
+ */
+
+#ifndef MDM_HAL_UTILITIES_H
+#define MDM_HAL_UTILITIES_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define DEFINE_NET_PROVIDER_DATA \
+    DEFINE_NET_PROVIDER( CELLULAR_NETPROV_TELEFONICA, "spark.telefonica.com", (23*60), (5684) ),  \
+    DEFINE_NET_PROVIDER( CELLULAR_NETPROV_KORE_VODAFONE, "vfd1.korem2m.com", (23*60), (5684) ),  \
+    DEFINE_NET_PROVIDER( CELLULAR_NETPROV_KORE_ATT, "10569.mcs", (23*60), (5684) ),  \
+    DEFINE_NET_PROVIDER( CELLULAR_NETPROV_TWILIO, "super", (23*60), (5684) ),  \
+    DEFINE_NET_PROVIDER( CELLULAR_NETPROV_MAX, "", (0), (0) )
+
+#define DEFINE_NET_PROVIDER( idx, apn, keepalive, port )  idx
+#ifdef __cplusplus
+enum CellularNetProv { DEFINE_NET_PROVIDER_DATA };
+#else
+typedef enum CellularNetProv CellularNetProv;
+#endif
+
+#undef DEFINE_NET_PROVIDER
+#define DEFINE_NET_PROVIDER( idx, apn, keepalive, port )  { apn, keepalive, port }
+
+#ifdef __cplusplus
+struct CellularNetProvData {
+    const char* apn;
+    int keepalive;
+    uint16_t port;
+};
+#else
+typedef struct CellularNetProvData CellularNetProvData;
+#endif
+
+/* detail functions defined for unit tests */
+namespace particle { namespace detail {
+/**
+ * Function for setting the cellular network provider based on the ICCID and/or IMSI of the SIM card inserted, broken out for unit tests
+ */
+CellularNetProv _cellular_sim_to_network_provider(const char* imsi, const char* iccid);
+
+}} // namespace particle detail
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif  /* MDM_HAL_UTILITIES_H */
--- a/hal/inc/cellular_hal_utilities.h
+++ b/hal/inc/cellular_hal_utilities.h
@@ -5,6 +5,8 @@
 #ifndef MDM_HAL_UTILITIES_H
 #define MDM_HAL_UTILITIES_H
 
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/hal/inc/cellular_hal_utilities.h
+++ b/hal/inc/cellular_hal_utilities.h
@@ -43,7 +43,7 @@ namespace particle { namespace detail {
 /**
  * Function for setting the cellular network provider based on the ICCID and/or IMSI of the SIM card inserted, broken out for unit tests
  */
-CellularNetProv _cellular_sim_to_network_provider(const char* imsi, const char* iccid);
+CellularNetProv cellular_sim_to_network_provider_impl(const char* imsi, const char* iccid);
 
 }} // namespace particle detail
 
--- a/hal/network/ncp/cellular/cellular_network_manager.h
+++ b/hal/network/ncp/cellular/cellular_network_manager.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include "c_string.h"
+#include "network_config_db.h"
 
 namespace particle {
 
@@ -45,12 +46,17 @@ public:
     const char* password() const;
     bool hasPassword() const;
 
+    CellularNetworkConfig& netProv(CellularNetworkProv net_prov);
+    CellularNetworkProv netProv() const;
+    bool hasNetProv() const;
+
     bool isValid() const;
 
 private:
     CString apn_;
     CString user_;
     CString pwd_;
+    CellularNetworkProv net_prov_;
 };
 
 class CellularNetworkManager {
@@ -75,47 +81,58 @@ private:
 };
 
 inline CellularNetworkConfig::CellularNetworkConfig() {
+    net_prov_ = CellularNetworkProv::NONE;
 }
 
+// APN
 inline CellularNetworkConfig& CellularNetworkConfig::apn(const char* apn) {
     apn_ = apn;
     return *this;
 }
-
 inline const char* CellularNetworkConfig::apn() const {
     return apn_;
 }
-
 inline bool CellularNetworkConfig::hasApn() const {
     return apn_ && *apn_;
 }
 
+// USERNAME
 inline CellularNetworkConfig& CellularNetworkConfig::user(const char* user) {
     user_ = user;
     return *this;
 }
-
 inline const char* CellularNetworkConfig::user() const {
     return user_;
 }
-
 inline bool CellularNetworkConfig::hasUser() const {
     return user_ && *user_;
 }
 
+// PASSWORD
 inline CellularNetworkConfig& CellularNetworkConfig::password(const char* pwd) {
     pwd_ = pwd;
     return *this;
 }
-
 inline const char* CellularNetworkConfig::password() const {
     return pwd_;
 }
-
 inline bool CellularNetworkConfig::hasPassword() const {
     return pwd_ && *pwd_;
 }
 
+// NETWORK PROVIDER
+inline CellularNetworkConfig& CellularNetworkConfig::netProv(CellularNetworkProv net_prov) {
+    net_prov_ = net_prov;
+    return *this;
+}
+inline CellularNetworkProv CellularNetworkConfig::netProv() const {
+    return net_prov_;
+}
+inline bool CellularNetworkConfig::hasNetProv() const {
+    return net_prov_ > CellularNetworkProv::NONE && net_prov_ < CellularNetworkProv::MAX;
+}
+
+// IS_VALID
 inline bool CellularNetworkConfig::isValid() const {
     return (apn_ && user_ && pwd_);
 }
--- a/hal/network/ncp/cellular/cellular_network_manager.h
+++ b/hal/network/ncp/cellular/cellular_network_manager.h
@@ -46,9 +46,8 @@ public:
     const char* password() const;
     bool hasPassword() const;
 
-    CellularNetworkConfig& netProv(CellularNetworkProv net_prov);
-    CellularNetworkProv netProv() const;
-    bool hasNetProv() const;
+    CellularNetworkConfig& netProv(CellularNetworkProvider np);
+    CellularNetworkProvider netProv() const;
 
     bool isValid() const;
 
@@ -56,7 +55,7 @@ private:
     CString apn_;
     CString user_;
     CString pwd_;
-    CellularNetworkProv net_prov_;
+    CellularNetworkProvider np_;
 };
 
 class CellularNetworkManager {
@@ -81,7 +80,7 @@ private:
 };
 
 inline CellularNetworkConfig::CellularNetworkConfig() {
-    net_prov_ = CellularNetworkProv::NONE;
+    np_ = CellularNetworkProvider::NONE;
 }
 
 // APN
@@ -121,15 +120,12 @@ inline bool CellularNetworkConfig::hasPassword() const {
 }
 
 // NETWORK PROVIDER
-inline CellularNetworkConfig& CellularNetworkConfig::netProv(CellularNetworkProv net_prov) {
-    net_prov_ = net_prov;
+inline CellularNetworkConfig& CellularNetworkConfig::netProv(CellularNetworkProvider np) {
+    np_ = np;
     return *this;
 }
-inline CellularNetworkProv CellularNetworkConfig::netProv() const {
-    return net_prov_;
-}
-inline bool CellularNetworkConfig::hasNetProv() const {
-    return net_prov_ > CellularNetworkProv::NONE && net_prov_ < CellularNetworkProv::MAX;
+inline CellularNetworkProvider CellularNetworkConfig::netProv() const {
+    return np_;
 }
 
 // IS_VALID
--- a/hal/network/ncp/cellular/network_config_db.cpp
+++ b/hal/network/ncp/cellular/network_config_db.cpp
@@ -27,7 +27,7 @@ namespace particle {
 namespace {
 
 struct NetworkConfig {
-    CellularNetworkProv net_prov;
+    CellularNetworkProvider np;
     const char* mcc;
     const char* mnc;
     const char *iccid_prefix;
@@ -37,11 +37,11 @@ struct NetworkConfig {
 };
 
 const NetworkConfig NETWORK_CONFIG[] = {
-    { CellularNetworkProv::TELEFONICA, "214", "07", "", "spark.telefonica.com", "", "" }, // Telefonica
-    { CellularNetworkProv::KORE_VODAFONE, "204", "04", "", "vfd1.korem2m.com", "kore", "kore" }, // Kore/Vodafone
-    { CellularNetworkProv::KORE_ATT, "310", "410", "", "10569.mcs", "", "" }, // Kore/AT&T
-    { CellularNetworkProv::TWILIO, "", "", "8988323", "super", "", "" }, // Twilio Super SIM
-    { CellularNetworkProv::TWILIO, "", "", "8988307", "super", "", "" }, // Twilio Super SIM
+    { CellularNetworkProvider::TELEFONICA, "214", "07", "", "spark.telefonica.com", "", "" }, // Telefonica
+    { CellularNetworkProvider::KORE_VODAFONE, "204", "04", "", "vfd1.korem2m.com", "kore", "kore" }, // Kore/Vodafone
+    { CellularNetworkProvider::KORE_ATT, "310", "410", "", "10569.mcs", "", "" }, // Kore/AT&T
+    { CellularNetworkProvider::TWILIO, "", "", "8988323", "super", "", "" }, // Twilio Super SIM
+    { CellularNetworkProvider::TWILIO, "", "", "8988307", "super", "", "" }, // Twilio Super SIM
 };
 
 const size_t NETWORK_CONFIG_SIZE = sizeof(NETWORK_CONFIG) / sizeof(NETWORK_CONFIG[0]);
@@ -60,7 +60,7 @@ CellularNetworkConfig networkConfigForIccid(const char* iccid, size_t size) {
     for (size_t i = 0; i < NETWORK_CONFIG_SIZE; ++i) {
         const NetworkConfig& conf = NETWORK_CONFIG[i];
         if (strncmp(iccid, conf.iccid_prefix, ICCID_PREFIX_LEN) == 0) {
-            return CellularNetworkConfig().netProv(conf.net_prov).apn(conf.apn).user(conf.user).password(conf.password);
+            return CellularNetworkConfig().netProv(conf.np).apn(conf.apn).user(conf.user).password(conf.password);
         }
     }
     return CellularNetworkConfig();
@@ -78,7 +78,7 @@ CellularNetworkConfig networkConfigForImsi(const char* imsi, size_t size) {
                 return CellularNetworkConfig();
             }
             if (strncmp(imsi + MCC_SIZE, conf.mnc, mncSize) == 0) {
-                return CellularNetworkConfig().netProv(conf.net_prov).apn(conf.apn).user(conf.user).password(conf.password);
+                return CellularNetworkConfig().netProv(conf.np).apn(conf.apn).user(conf.user).password(conf.password);
             }
         }
     }
--- a/hal/network/ncp/cellular/network_config_db.cpp
+++ b/hal/network/ncp/cellular/network_config_db.cpp
@@ -27,6 +27,7 @@ namespace particle {
 namespace {
 
 struct NetworkConfig {
+    CellularNetworkProv net_prov;
     const char* mcc;
     const char* mnc;
     const char *iccid_prefix;
@@ -36,11 +37,11 @@ struct NetworkConfig {
 };
 
 const NetworkConfig NETWORK_CONFIG[] = {
-    { "214", "07", "", "spark.telefonica.com", "", "" }, // Telefonica
-    { "310", "410", "", "10569.mcs", "", "" }, // Kore/AT&T
-    { "204", "04", "", "vfd1.korem2m.com", "kore", "kore" }, // Kore/Vodafone
-    { "", "", "8988307", "super", "", "" }, // Twilio
-    { "", "", "8988323", "super", "", "" }  // Twilio
+    { CellularNetworkProv::TELEFONICA, "214", "07", "", "spark.telefonica.com", "", "" }, // Telefonica
+    { CellularNetworkProv::KORE_VODAFONE, "204", "04", "", "vfd1.korem2m.com", "kore", "kore" }, // Kore/Vodafone
+    { CellularNetworkProv::KORE_ATT, "310", "410", "", "10569.mcs", "", "" }, // Kore/AT&T
+    { CellularNetworkProv::TWILIO, "", "", "8988323", "super", "", "" }, // Twilio Super SIM
+    { CellularNetworkProv::TWILIO, "", "", "8988307", "super", "", "" }, // Twilio Super SIM
 };
 
 const size_t NETWORK_CONFIG_SIZE = sizeof(NETWORK_CONFIG) / sizeof(NETWORK_CONFIG[0]);
@@ -59,7 +60,7 @@ CellularNetworkConfig networkConfigForIccid(const char* iccid, size_t size) {
     for (size_t i = 0; i < NETWORK_CONFIG_SIZE; ++i) {
         const NetworkConfig& conf = NETWORK_CONFIG[i];
         if (strncmp(iccid, conf.iccid_prefix, ICCID_PREFIX_LEN) == 0) {
-            return CellularNetworkConfig().apn(conf.apn).user(conf.user).password(conf.password);
+            return CellularNetworkConfig().netProv(conf.net_prov).apn(conf.apn).user(conf.user).password(conf.password);
         }
     }
     return CellularNetworkConfig();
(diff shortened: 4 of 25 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1738,6 +1782,23 @@ void SaraNcpClient::checkRegistrationState() {
...
+        if (!strcmp(netConfig.apn(), "super")) { // if Twilio Super SIM
+            isTwilioSuperSIM_ = 1;
+        }
+        iccidChecked_ = 1; // cache
+    }
+
+    if (isTwilioSuperSIM_ && millis() - regStartTime_ <= REGISTRATION_TWILIO_HOLDOFF_TIMEOUT) {
+        return 0;
+    }
+
```
keeramis, line 1801: With this, we will not _start_ running intervene commands before `REGISTRATION_TWILIO_HOLDOFF_TIMEOUT`, but once the time out is expired, we will be running them every 15 sec. Do we want to run the intervention commands every 15 sec or `REGISTRATION_INTERVENTION_TIMEOUT` sec?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -950,7 +951,7 @@ bool MDMParser::init(DevStatus* status)
         goto failure;
     // Request IMSI (International Mobile Subscriber Identification)
     sendFormated("AT+CIMI\r\n");
-    if (RESP_OK != waitFinalResp(_cbString, &str_imsi))
+    if (RESP_OK != waitFinalResp(_cbString, &str_imsi, CIMI_TIMEOUT))
```
keeramis, line 954: For consistency, might need to add `CIMI_TIMEOUT` here?

https://github.com/particle-iot/device-os/blob/3b3984ffe000130ba7c64b7811d76a1be688d84e/hal/src/electron/modem/mdm_hal.cpp#L1529

https://github.com/particle-iot/device-os/blob/3b3984ffe000130ba7c64b7811d76a1be688d84e/hal/src/electron/modem/mdm_hal.cpp#L1474

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1500,6 +1506,22 @@ void QuectelNcpClient::checkRegistrationState() {
 int QuectelNcpClient::interveneRegistration() {
     CHECK_TRUE(connState_ == NcpConnectionState::CONNECTING, SYSTEM_ERROR_NONE);
 
+    // Check and store if we are using a Twilio SIM based on ICCID
+    if (!iccidChecked_) {
```
keeramis, line 1510: Nitpicking: Should we move this `if` condition to a separate function for reusability? This can be a general utility to set `iccidChecked_` instead of being only in `interveneRegistration()`

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1772,6 +1778,23 @@ void SaraNcpClient::checkRegistrationState() {
 int SaraNcpClient::interveneRegistration() {
     CHECK_TRUE(connState_ == NcpConnectionState::CONNECTING, SYSTEM_ERROR_NONE);
 
+    // Check and store if we are using a Twilio SIM based on ICCID
+    if (!iccidChecked_) {
```
keeramis, line 1786: (same comment as above)
Nitpicking: Should we move this if condition to a separate function for reusability? This can be a general utility to set iccidChecked_ instead of being only in interveneRegistration()

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1044,28 +1053,64 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+        int isTwilio = 0;
+        // Check if we are using a Twilio Super SIM based on ICCID
+        char buf[32] = {};
+        auto lenCcid = getIccidImpl(buf, sizeof(buf));
+        CellularNetworkConfig netConfig;
+        CHECK_TRUE(lenCcid > 5, SYSTEM_ERROR_AT_NOT_OK);
+        netConfig = networkConfigForIccid(buf, lenCcid);
+        if (!strcmp(netConfig.apn(), "super")) { // if Twilio Super SIM
+            isTwilio = 1;
+        }
```
keeramis, line 1066: Nit: Why not use `isTwilioSuperSIM_` instead of evaluating for twilio super SIM here as well as here? https://github.com/particle-iot/device-os/pull/2185/files#diff-ece868c5f58dddb291c6b739de0da632R1792

Probably can update `isTwilioSuperSIM_` for once here and then not calculate in the second location? Just a nit, though....

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1173,17 +1179,13 @@ int QuectelNcpClient::checkSimCard() {
...
+        // First look for network settings based on ICCID
         char buf[32] = {};
-        const int ret = CHECK_PARSER(resp.scanf("+CCID: %31s", buf));
-        const int r = CHECK_PARSER(resp.readResult());
-        CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_AT_NOT_OK);
-        if (ret) {
-            netConf_ = networkConfigForIccid(buf, strlen(buf));
-        }
+        auto lenCcid = getIccidImpl(buf, sizeof(buf));
+        CHECK_TRUE(lenCcid > 5, SYSTEM_ERROR_AT_NOT_OK);
```
keeramis, line 1187: Should this error be called something else? Like `SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED`?

`SYSTEM_ERROR_AT_NOT_OK` is already processed here - https://github.com/particle-iot/device-os/pull/2185/files#diff-ece868c5f58dddb291c6b739de0da632R546

keeramis (commented):
Added comments, and testing on my module.

hal/network/ncp_client/sara/sara_ncp_client.cpp
keeramis: Talked with Andrey. We stick to this. thanks

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1044,28 +1053,64 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+                if (isTwilio) { // if Twilio Super SIM
+                    umnoprof = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                }
+                parser_.execCommand(1000, "AT+UMNOPROF=%d", umnoprof);
                 // Not checking for error since we will reset either way
                 reset = true;
+            } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
+                auto respBand = parser_.sendCommand(UBLOX_UBANDMASK_TIMEOUT, "AT+UBANDMASK?");
+                uint64_t ubandCatm1 = 0;
+                auto retBand = CHECK_PARSER(respBand.scanf("+UBANDMASK: 0,%llu", &ubandCatm1));
```
avtolstoy, line 1095: This will not work. We use `newlib-nano` variant of libc, which does not support 64-bit integers in `scanf` and `printf` type of functions. `strtoull` should work though.

ci/unit_tests.sh
```diff
@@ -52,7 +52,8 @@ set -x -e
 # Run CMake-based unit tests
 cd $unit_test_dir
 rm -rf .build/*
-mkdir .build -p && cd .build/
+mkdir -p .build/
```
avtolstoy, line 55: :+1:

avtolstoy (changes requested):

hal/src/electron/modem/mdm_hal.cpp
technobly: Sounds good.  I found another place to add it as well.  Should we have the callback added here?
```c
                // Query IMSI every 60 sec if not regsitered, to detect IMSI switches
                if (HAL_Timer_Get_Milli_Seconds() - start_imsi_check > 60000UL) {
                    start_imsi_check = HAL_Timer_Get_Milli_Seconds();
                    sendFormated("AT+CIMI\r\n");
                    if (RESP_OK != waitFinalResp()) {
                        goto failure;
                    }
                }
```

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: I was just trying to keep the error code the same as it was, but I will change to SYSTEM_ERROR_BAD_DATA.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1044,28 +1053,64 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+        do {
+            // Set UMNOPROF = SIM_SELECT
+            auto resp = parser_.sendCommand("AT+UMNOPROF?");
+            reset = false;
+            int umnoprof = static_cast<int>(UbloxSaraUmnoprof::NONE);
+            auto r = CHECK_PARSER(resp.scanf("+UMNOPROF: %d", &umnoprof));
+            CHECK_PARSER_OK(resp.readResult());
+            if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::SW_DEFAULT) {
+                // Disconnect before making changes to the UMNOPROF
+                auto respCfun = parser_.sendCommand(UBLOX_CFUN_TIMEOUT, "AT+CFUN?");
```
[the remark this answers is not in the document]
technobly: Since we have the ability to check, and repeatedly setting CFUN=0,0 is somewhat of an unknown operation behind the scenes, I think checking first is a bit easier on the system.  And it takes very little time to do.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1044,28 +1053,64 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+                int cfunVal = -1;
+                auto retCfun = CHECK_PARSER(respCfun.scanf("+CFUN: %d", &cfunVal));
+                CHECK_PARSER_OK(respCfun.readResult());
+                if (retCfun == 1 && cfunVal != 0) {
+                    CHECK_PARSER_OK(parser_.execCommand(UBLOX_CFUN_TIMEOUT, "AT+CFUN=0"));
```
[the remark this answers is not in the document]
technobly: Sounds good, yeah I missed that one.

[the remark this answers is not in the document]
technobly: I'll update the other existing ones in the file as well.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1206,6 +1240,19 @@ int MDMParser::_cbCFUN(int type, const char *buf, int len, int* i) {
     return WAIT;
 }
 
+int MDMParser::_cbUBANDMASK(int type, const char *buf, int len, uint32_t* i)
+{
+    if ((type == TYPE_PLUS) && i){
```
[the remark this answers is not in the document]
technobly: Fixed _cbUMNOPROF as well.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1287,20 +1334,19 @@ bool MDMParser::registerNet(const char* apn, NetStatus* status, system_tick_t ti
...
+        // Set to full functionality mode
+        int cfun_val = -1;
+        sendFormated("AT+CFUN?\r\n");
+        if (RESP_OK != waitFinalResp(_cbCFUN, &cfun_val)) {
+            goto failure;
+        }
+        if (cfun_val != 1) {
+            sendFormated("AT+CFUN=1,0\r\n");
+            if (RESP_OK != waitFinalResp(nullptr, nullptr, CFUN_TIMEOUT))
+                goto failure;
```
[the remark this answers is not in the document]
technobly: Oof, copypasta. This is more than a NIT I'd say.   This should be our, and is my personal, coding standard.  I fixed the original as well.

system/src/system_network_internal.h
```diff
@@ -57,15 +57,15 @@ extern uint32_t wlan_watchdog_base;
 inline void ARM_WLAN_WD(uint32_t x) {
     wlan_watchdog_base = HAL_Timer_Get_Milli_Seconds();
     wlan_watchdog_duration = x;
-    WAN_WD_DEBUG("WD Set %d",(x));
+    WAN_WD_DEBUG("WD Set %d\r\n",(x));
```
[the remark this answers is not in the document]
technobly: Oh, well I had changed it to use the new logging.h API and made this change: `LOG_PRINTF_C(INFO, "system.wdt", x, ##__VA_ARGS__)` to `DEBUG(x,__VA_ARGS__)`.  I noticed the lines weren't correct so I added the newlines.  But then I backed out those changes and thought it might have always been that way since we never enable it.  Should I make the new logging API change, since we don't typically compile with DEBUG_WAN_WD it should be removed anyway.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: Resolving this one as it's duplicated :D

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1044,28 +1053,64 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
-        reset = false;
-        int umnoprof = static_cast<int>(UbloxSaraUmnoprof::NONE);
-        auto r = CHECK_PARSER(resp.scanf("+UMNOPROF: %d", &umnoprof));
-        CHECK_PARSER_OK(resp.readResult());
-        if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::SW_DEFAULT) {
-            // Disconnect before making changes to the UMNOPROF
-            r = CHECK_PARSER(parser_.execCommand("AT+COPS=2,2"));
-            if (r == AtResponse::OK) {
+        int resetCount = 0;
+        int isTwilio = 0;
```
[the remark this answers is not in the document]
technobly: Sure why not :D

system/src/system_network_internal.h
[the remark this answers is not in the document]
avtolstoy: I'd remove them for now and make the logging changes in a separate PR. We have a relevant CH story.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1206,6 +1240,19 @@ int MDMParser::_cbCFUN(int type, const char *buf, int len, int* i) {
     return WAIT;
 }
 
+int MDMParser::_cbUBANDMASK(int type, const char *buf, int len, uint32_t* i)
+{
+    if ((type == TYPE_PLUS) && i){
+        uint64_t a;
+        *i = 0;
+        // \r\n+UBANDMASK: 0,6170,1,524420\r\n
+        if (sscanf(buf, "\r\n+UBANDMASK: 0,%llu", &a) == 1) {
```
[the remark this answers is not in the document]
technobly: Good catch, I'll refactor this one.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Good catch, I'll refactor this one.

hal/src/electron/modem/mdm_hal.cpp
keeramis: I don't think that it is needed to add a callback here because it's only for debugging, and we will not use this IMSI for other purposes. But once the device registers on a particular IMSI, we would want that IMSI stored because that's the IMSI that has worked for registration, and in that part of the code, there is the callback. (https://github.com/particle-iot/device-os/blob/23c2920ac3263363a00ab0595db6a17eaf0c2964/hal/src/electron/modem/mdm_hal.cpp#L1529)

technobly: Ok sounds good, I'll just add the explicit timeout `waitFinalResp(nullptr, nullptr, CIMI_TIMEOUT);` which is the same as the default timeout, but better to be explicit with the issue with CIMI.

keeramis: 👍  perfecto

system/src/system_network_internal.h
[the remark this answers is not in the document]
technobly: Will do.

hal/inc/cellular_hal_utilities.h
```diff
@@ -0,0 +1,54 @@
...
+#else
+typedef struct CellularNetProvData CellularNetProvData;
+#endif
+
+/* detail functions defined for unit tests */
+namespace particle { namespace detail {
+/**
+ * Function for setting the cellular network provider based on the ICCID and/or IMSI of the SIM card inserted, broken out for unit tests
+ */
+CellularNetProv _cellular_sim_to_network_provider(const char* imsi, const char* iccid);
```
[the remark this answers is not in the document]
technobly: Fixed with `_impl` and changed a few more in these files.

hal/network/ncp/cellular/network_config_db.h
```diff
@@ -21,6 +21,15 @@
 
 namespace particle {
 
+enum class CellularNetworkProv {
+    NONE = 0,
+    TELEFONICA = 1,
+    KORE_VODAFONE = 2,
+    KORE_ATT = 3,
+    TWILIO = 4,
+    MAX = 100,
```
keeramis, line 30: Is there a reason to add `100` for MAX? Should this be left to enumerate by itself, or set it to 5?

For example, if for some reason the `net_prov_` is '8' or some garbage number, `CellularNetworkConfig::hasNetProv()` will still return `true` while it's supposed to be false.

hal/network/ncp/cellular/network_config_db.cpp
```diff
@@ -27,6 +27,7 @@ namespace particle {
 namespace {
 
 struct NetworkConfig {
+    CellularNetworkProv net_prov;
```
avtolstoy, line 30: `CellularNetworkProvider` ? :) Being verbose does somewhat increase readability.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1197,6 +1199,9 @@ int QuectelNcpClient::configureApn(const CellularNetworkConfig& conf) {
             const int r = CHECK_PARSER(resp.readResult());
             CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);
             netConf_ = networkConfigForImsi(buf, strlen(buf));
+            if (netConf_.hasNetProv()) {
+                netProv_ = netConf_.netProv();
+            }
```
avtolstoy, line 1204: Even if `netConf_` is valid (which means `Cellular.setCredentials()` was called), we still need to make a call potentially to `networkConfigForImsi()` and copy the provider value out of it.

Same with SARA code.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1186,6 +1185,9 @@ int QuectelNcpClient::configureApn(const CellularNetworkConfig& conf) {
         auto lenCcid = getIccidImpl(buf, sizeof(buf));
         CHECK_TRUE(lenCcid > 5, SYSTEM_ERROR_AT_NOT_OK);
         netConf_ = networkConfigForIccid(buf, lenCcid);
+        if (netConf_.hasNetProv()) {
+            netProv_ = netConf_.netProv();
```
avtolstoy, line 1189: Nit: It should probably be fine not to have a separate member variable and just access it through `netConf_.netProv()`, assuming we properly patch it in case of stored credentials.

Same with SARA code.

For convenience one could add a helper method to access provider directly somewhat similar to `ncpId()` for example.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,18 +1090,23 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
             } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
                 auto respBand = parser_.sendCommand(UBLOX_UBANDMASK_TIMEOUT, "AT+UBANDMASK?");
                 uint64_t ubandCatm1 = 0;
-                auto retBand = CHECK_PARSER(respBand.scanf("+UBANDMASK: 0,%llu", &ubandCatm1));
+                char ubandCatm1_str[24] = {};
+                auto retBand = CHECK_PARSER(respBand.scanf("+UBANDMASK: 0,%23[^,\n]", ubandCatm1_str));
```
avtolstoy, line 1094: Nit: `\n` shouldn't be needed, Gen 3 parser will anyway strip it.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,18 +1090,23 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
             } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
                 auto respBand = parser_.sendCommand(UBLOX_UBANDMASK_TIMEOUT, "AT+UBANDMASK?");
                 uint64_t ubandCatm1 = 0;
-                auto retBand = CHECK_PARSER(respBand.scanf("+UBANDMASK: 0,%llu", &ubandCatm1));
+                char ubandCatm1_str[24] = {};
```
avtolstoy, line 1093: Nit: `camelCase`.

hal/network/ncp_client/sara/sara_ncp_client.cpp
avtolstoy: I'm reading the R4 AT reference and it seems that `AT+UBANDMASK?` may give multiple strings. Should we guard for that?

image

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1240,14 +1240,19 @@ int MDMParser::_cbCFUN(int type, const char *buf, int len, int* i) {
...
-    if ((type == TYPE_PLUS) && i){
+    if ((type == TYPE_PLUS) && i) {
         uint64_t a;
+        char s[24] = {};
+        char* pEnd = &s[0];
         *i = 0;
         // \r\n+UBANDMASK: 0,6170,1,524420\r\n
-        if (sscanf(buf, "\r\n+UBANDMASK: 0,%llu", &a) == 1) {
-            *i = (uint32_t)(a & 0xffffffff); // limit to 32-bit (Cat-M R410-02B supports up to band 28)
+        if (sscanf(buf, "\r\n+UBANDMASK: 0,%23[^,\n]", s) == 1) {
```
avtolstoy, line 1251: I'm reading the R4 AT reference and it seems that `AT+UBANDMASK?` may give multiple strings. Should we guard for that?

image

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1240,14 +1240,19 @@ int MDMParser::_cbCFUN(int type, const char *buf, int len, int* i) {
...
+            a = strtoull(s, &pEnd, 10);
+            if (pEnd - s > 0) {
+                *i = a; // limit to 32-bit (Cat-M R410-02B supports up to band 28)
```
avtolstoy, line 1254: Nit: fix the comment. No 32-bit adjustment now.

avtolstoy (changes requested):
Looks good overall, requesting changes mainly for https://github.com/particle-iot/device-os/pull/2185/commits/120d87cc8601ab410b5b946872e77df817e234b9#diff-7cf4c0920642e9913b8848733dacef6fR1204

hal/network/ncp/cellular/network_config_db.h
technobly: I just set it high for when we add more, we don't have to remember to bump the MAX up.  The check for < MAX is kind of just a loose sanity check that we don't have some random garbage.  Even if there was a garbage value of 8, it wouldn't match anything in the enum list, so it shouldn't harm the code that uses this.  That said, I'll let it enumerate itself... that slipped my mind as an option when typing in the list.

hal/network/ncp/cellular/cellular_network_manager.h
```diff
@@ -45,12 +46,17 @@ class CellularNetworkConfig {
     const char* password() const;
     bool hasPassword() const;
 
+    CellularNetworkConfig& netProv(CellularNetworkProv net_prov);
```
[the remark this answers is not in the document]
technobly: Snake case was used here, and the private variables are already snaking with abc_ so it seeeemed appropriate :D If you don't mind too much, I'd like to leave it snake_case since it's internal.
```
struct NetworkConfig {
    CellularNetworkProv net_prov;
    const char* mcc;
    const char* mnc;
    const char *iccid_prefix;
    const char* apn;
    const char* user;
    const char* password;
};
```

hal/src/electron/modem/mdm_hal.cpp
technobly: It does currently spit out 0,bandsm1,1,bandnb1.  But the scanf only matches what it matches and the rest is ignored.  We don't want to match NB1 or GSM, so it should be ok as-is.

hal/network/ncp/cellular/cellular_network_manager.h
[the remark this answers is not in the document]
avtolstoy: No strong preferences, but these are two different things though. I'm not commenting on `NetworkConfig`, but the rest of `CellularNetworkConfig` uses camelCase.

hal/network/ncp_client/sara/sara_ncp_client.cpp
avtolstoy: Based on https://github.com/particle-iot/device-os/pull/2185/files#r485852743 this will need to get modified to read all the intermediate responses similar to https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp_client/sara/sara_ncp_client.cpp#L702

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: `ubandCatm1Str` is not as readable, so I would prefer `ubandCatm1_str`.  I could rename them to `ubandUint` and `ubandStr` I guess.

hal/network/ncp/cellular/network_config_db.h
keeramis: cool 👍

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: It defaults to CellularNetworkProv::NONE, and we don't support 3rd Party SIMs.  So if someone is using our SIM externally, they just need to ensure that credentials are cleared such that Device OS does the right thing.  If someone is using a SIM externally that matches one we use, they might not like settings we apply.  Configuring things properly will be up to them in the user app.

avtolstoy: I think this is arguable. If we can correctly detect that we are on a particular operator SIM even if we are using an external SIM with potentially custom apn/username/password set, we should still perform the lookup based on the ICCID and understand which one it is. If it's an unknown 3rdparty SIM, that's fine, we'll default to NONE, but otherwise if we have some workarounds for a particular operator in place they should work no matter what.

Moreover, we don't have an API to set the operator from the application.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: I'm not sure what you mean here by "assuming we properly patch it in case of stored credentials".  AFAIK, this extra piece of `netProv` info is not going to or need to be saved to flash memory.

For now, I don't think we need more convenience since it's just one check for `if Twilio`.

avtolstoy: > assuming we properly patch it in case of stored credentials

https://github.com/particle-iot/device-os/pull/2185#discussion_r485960534

I on the other hand am not sure what you mean by saving to flash memory :) `netConf_` is completely local to the NCP Client class, even if it gets reloaded in `connect()` by the cellular network manager.

hal/network/ncp/cellular/cellular_network_manager.h
```diff
@@ -75,47 +81,58 @@ class CellularNetworkManager {
...
+// NETWORK PROVIDER
+inline CellularNetworkConfig& CellularNetworkConfig::netProv(CellularNetworkProv net_prov) {
+    net_prov_ = net_prov;
+    return *this;
+}
+inline CellularNetworkProv CellularNetworkConfig::netProv() const {
+    return net_prov_;
+}
+inline bool CellularNetworkConfig::hasNetProv() const {
+    return net_prov_ > CellularNetworkProv::NONE && net_prov_ < CellularNetworkProv::MAX;
```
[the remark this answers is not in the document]
technobly: That is true... I started with `enum` and created this `hasNetProv`, then moved to `enum class`.  I'll remove this.

hal/network/ncp/cellular/network_config_db.cpp
technobly: Sure, ok ;-)

hal/network/ncp/cellular/network_config_db.h
technobly: I just removed MAX because it's an enum class vs. enum which I had started with.

hal/network/ncp/cellular/cellular_network_manager.h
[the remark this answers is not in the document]
technobly: Will change these to something else to be unique and avoid case issues.

avtolstoy (approved):

PR #2187, opened 2020-09-04
Define C++-compatible pin map for Argon

rgiese:
### Problem

The Particle `pinmap_defines.h` headers establish pin IDs like `SCA` with a `#define`, meaning that any stand-alone use of the token `SCA` gets replaced with a number by the preprocessor. Other Arduino-based platforms don't do this. Because of Particle's headers, Particle apps cannot directly use [some Arduino libraries](https://github.com/pololu/tic-arduino) that use a token like `SCA` internally, such as:

```
enum class TicPin
{
  SCL = 0,
  SDA = 1,
  TX  = 2,
  RX  = 3,
  RC  = 4,
};
```

### Solution

We should use C++ to establish constants when we're building in C++. This allows library code to safely use tokens like `SCA` without preprocessor interference.

Since `pinmap_defines.h` is also used to compile some nRF platform code in C, we unfortunately need to make the pin mappings available through both C and C++ means, whichever we're compiling with. That said, these defines are unlikely to change, to put it mildly, so I think this poses a negligible maintenance burden.

This PR offers this change for the Argon platform only (since that's what I'm using for my current project). If we believe that this is overall a good idea, I'm happy to make the change for all headers and platforms.

### Steps to Test

Build `device-os/main` with `make PLATFORM=argon`.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/argon/pinmap_defines.h
+++ b/hal/src/argon/pinmap_defines.h
@@ -19,6 +19,79 @@
 
 #if PLATFORM_ID == PLATFORM_ARGON
 
+#ifdef __cplusplus
+
+constexpr pin_t TOTAL_PINS = 36;
+
+// Digital pins
+constexpr pin_t D0 = 0;
+constexpr pin_t D1 = 1;
+constexpr pin_t D2 = 2;
+constexpr pin_t D3 = 3;
+constexpr pin_t D4 = 4;
+constexpr pin_t D5 = 5;
+constexpr pin_t D6 = 6;
+constexpr pin_t D7 = 7;
+constexpr pin_t D8 = 8;
+constexpr pin_t D9 = 9;
+constexpr pin_t D10 = 10;
+constexpr pin_t D11 = 11;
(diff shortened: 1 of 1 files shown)


rgiese:
cc: @DavidEGrayson

avtolstoy:
Thanks for the PR. That's always the problem with defines with 'generic' names. We have a couple of other cases (e.g. `LED_XXX`) where we should move away from such names that could potentially cause problems for applications.

We will consider applying this change into the next major release (3.x). A few points:
1. The PRs should target `develop` branch, `master` always points to the current stable release
2. Maybe we should come up with something for C as well (e.g. use `static const` instead; enums won't do unfortunately as they are signed and `pin_t` is unsigned and we cannot change its type easily) and ideally minimize duplication as much as possible when defining pins. While the pinmap does not change, it still leaves a possibility for a mistake. Perhaps this can be overcome with tests.
3. This is a breaking change and may break some applications/libraries which rely on `#ifdef PIN_NAME`

PR #2189, opened 2020-09-05
Stream definition should be Arduino-compliant

rgiese:
### Problem

The definition of `Stream::readBytes` is not fully Arduino-compliant: it [should accept both a `char*` and `uint8_t*` for its buffer pointer](https://www.arduino.cc/reference/en/language/functions/communication/serial/readbytes/). Particle's default compiler settings won't allow the cast to happen automagically, which means some dependent libraries don't compile. Sadness.

### Solution

Create a function that accepts the missing type and forwards it with an appropriate cast.

### Steps to Test

Build `device-os/main` with `make PLATFORM=argon`.

### References

Thanks to @DavidEGrayson for recommending this fix on https://github.com/pololu/tic-arduino/pull/1.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_stream.h
+++ b/wiring/inc/spark_wiring_stream.h
@@ -83,6 +83,8 @@ class Stream : public Print
   // terminates if length characters have been read or timeout (see setTimeout)
   // returns the number of characters placed in the buffer (0 means no valid data found)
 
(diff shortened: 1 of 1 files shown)


avtolstoy:
Thanks for the PR. While this does address `Stream::readBytes()`, if we are making this change, other methods should also be considered e.g. `Stream::readBytesUntil()`, `Stream::find()` etc.

PR #2196, opened 2020-09-11
Gen3: SPI.transfer() supports accepting const void* parameter.

XuGuohui:
SPI.transfer takes a `void *` for the TX buffer. On Gen2 (Electron, etc) `SPI.transfer` works when casting a const TX buffer (ex. const array, const string, etc) to a non-const TX buffer. On Gen3 this will result in an immediate hard fault and program crash because the nRF SPI DMA function requires the tx buffer to be in RAM.

Change the interface of `SPI.transfer` to take const void * and to copy the buffer to RAM if it was not in RAM.

### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

static runtime_info_t heapInfo;

void setup() {
    if (heapInfo.size == 0) {
        heapInfo.size = sizeof(heapInfo);
    }

    HAL_Core_Runtime_Info(&heapInfo, nullptr);
    Log.info("Heapinfo: size :%lu, largest_free_block_heap :%lu", heapInfo.freeheap, heapInfo.largest_free_block_heap);

    SPI.setDataMode(SPI_MODE0);
    SPI.setClockSpeed(100 * 1000);
    SPI.begin();
}

void loop() {
    delay(1000);

    SPI.transfer((uint8_t*) "ABC", NULL, sizeof("ABC"), NULL);
    HAL_Core_Runtime_Info(&heapInfo, nullptr);
    Log.info("Heapinfo: size :%lu, largest_free_block_heap :%lu", heapInfo.freeheap, heapInfo.largest_free_block_heap);
}
```

---

### Completeness

- [ ] Added documentation


hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,36 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
         return;
     }
 
+    uint8_t* txPtr = nullptr;
+
+    SCOPE_GUARD ({
+        if (!nrfx_is_in_ram(tx_buffer)) {
+            free(txPtr);
```
avtolstoy, line 456: `txPtr` may very well be null.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,36 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
...
+        }
+    });
+
+    if (!nrfx_is_in_ram(tx_buffer)) {
+        txPtr = (uint8_t*)malloc(length);
```
avtolstoy, line 461: I think we can do better. The whole idea with sending constant data out of flash is not to waste RAM. This approach pretty much negates it. What if there is a rather larger blob resource that needs to be transmitted e.g. > 50K?

We should be able to get away with a rather small stack/heap buffer (e.g. 256 bytes) and just send the data in multiple chunks. This shouldn't cause any problems, as SPI data is anyway clocked by the master and as long as the CS is held.

With SPI Slave mode, it doesn't seem that this would work, unless there was support for double-buffering. So for slave transfers we can probably keep the current approach or not allow out of RAM buffers.

avtolstoy (changes requested):
`transfer()` signature should also reflect these changes. The tx pointer should be const.

hal/src/nRF52840/spi_hal.cpp
XuGuohui: Good catch!

hal/src/nRF52840/spi_hal.cpp
XuGuohui: Will refactor it a little bit.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,53 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
...
 
     spiMap[spi].dma_user_callback = userCallback;
     if (spiMap[spi].spi_mode == SPI_MODE_MASTER) {
-        SPARK_ASSERT(spiTransfer(spi, (uint8_t *)tx_buffer, (uint8_t *)rx_buffer, length) == length);
+        if (!nrfx_is_in_ram(tx_buffer)) {
+            // Truncate the data to reuse an allocated pool.
+            constexpr uint32_t maxPoolSize = 256;
+            uint32_t poolSize = std::min(length, maxPoolSize);
+            txPtr = (uint8_t*)malloc(poolSize);
+            SPARK_ASSERT(txPtr != nullptr);
```
avtolstoy, line 471: This can potentially be a non-critical error.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,53 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
...
+
+            uint32_t index = 0;
+            while (index < length) {
+                uint16_t chunkSize = std::min(length - index, poolSize);
+                memcpy(txPtr, &((uint8_t *)tx_buffer)[index], chunkSize);
+                SPARK_ASSERT(spiTransfer(spi, txPtr, (uint8_t *)rx_buffer + index, chunkSize) == chunkSize);
+                index += chunkSize;
+            }
+        } else {
+            txPtr = (uint8_t*)tx_buffer;
```
avtolstoy, line 481: The scope guard above would free a user-passed buffer.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,53 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
...
+            SPARK_ASSERT(spiTransfer(spi, txPtr, (uint8_t *)rx_buffer, length) == length);
+        }
     } else {
+        if (!nrfx_is_in_ram(tx_buffer)) {
+            txPtr = (uint8_t*)malloc(length);
+            SPARK_ASSERT(txPtr != nullptr);
+            memcpy(txPtr, tx_buffer, length);
+        } else {
+            txPtr = (uint8_t*)tx_buffer;
```
avtolstoy, line 490: The scope guard above would free a user-passed buffer.

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -448,20 +449,53 @@ void hal_spi_transfer_dma(hal_spi_interface_t spi, void* tx_buffer, void* rx_buf
...
+        }
+
         // Use for synchronization
         spiMap[spi].transmitting = true;
         // reset transfer length
         spiMap[spi].transfer_length = 0;
         spiMap[spi].slave_buf_length = length;
-        spiMap[spi].slave_tx_buf = tx_buffer;
+        spiMap[spi].slave_tx_buf = txPtr;
```
avtolstoy, line 498: By the time we exit this function `txPtr` will be freed. `nrfx_spis_buffers_set` just takes care of setting the buffers in the appropriate peripheral registers, no transfer happens at this point. By the time it starts, we'd be writing into invalid location. Any cleanup should probably happen in `spiSlaveEventHandler`.

user/tests/wiring/no_fixture_spi/spix.cpp
```diff
@@ -671,3 +671,25 @@ test(SPIX_21_SPI_Sleep) {
...
+    heapInfo1.size = sizeof(heapInfo1);
+
+    SPI.setClockSpeed(SPI_CLOCK_SPEED);
+    SPI.begin();
+    assertTrue(SPI.isEnabled());
+
+    HAL_Core_Runtime_Info(&heapInfo, nullptr);
+
+    SPI.beginTransaction();
+    SPI.transfer((uint8_t*) "Hello", nullptr, sizeof("Hello"), nullptr);
```
avtolstoy, line 687: Singnature should be changed to accept `const void*`.

avtolstoy (changes requested):

hal/src/nRF52840/spi_hal.cpp
XuGuohui: We only free the internally allocated buffer, as the `txPtr` is freed only when `!nrfx_is_in_ram(tx_buffer)`.

hal/src/nRF52840/spi_hal.cpp
XuGuohui: Good catch! 👍

hal/src/nRF52840/spi_hal.cpp
XuGuohui: Okay. Simply return if `nullptr` then.

hal/src/nRF52840/spi_hal.cpp
XuGuohui: I just realized that if we truncated the tx buffer, the following chunks except the first chunk cannot be sent, since the `spiTransfer` is not a blocking API and it just sets the tx/rx buffer and starts transmitting and then returns.

XuGuohui: We need to introduce some static variables to handle event properly for truncated data transmission.

avtolstoy: Oh, yeah, good catch. We need to add some state management for cases where we are performing a chunked transfer.

avtolstoy (changes requested):

hal/src/nRF52840/spi_hal.cpp
```diff
@@ -286,6 +360,13 @@ void hal_spi_init(hal_spi_interface_t spi) {
     spiMap[spi].dma_user_callback = nullptr;
     spiMap[spi].select_user_callback = nullptr;
     spiMap[spi].transfer_length = 0;
+    spiMap[spi].master_tx_data = nullptr;
```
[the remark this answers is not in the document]
XuGuohui: As we've allocated a big chunk of RAM for slave transmitting const data, we don't need to cache the pointer to const tx data.

user/tests/wiring/no_fixture_spi/spix.cpp
```diff
@@ -671,3 +671,25 @@ test(SPIX_21_SPI_Sleep) {
...
+    SPI.endTransaction();
+
+    HAL_Core_Runtime_Info(&heapInfo1, nullptr);
+
+    assertEqual(heapInfo.freeheap, heapInfo1.freeheap);
```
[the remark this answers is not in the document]
XuGuohui: You're right. Another thread may be allocating heap.

avtolstoy (approved):

PR #2197, opened 2020-09-14
Gen2: fix the electron power leakage issue when device is asleep.

XuGuohui:
**Note: this PR is targeting to the fix/electron-recovery-non-responsive branch**

### Problem

If the Electron is powered on but the modem is not initialized (by calling Cellular.on() or Particle.connect()) before entering sleep mode in the MANUAL mode, the modem persist on and it will consume additional ~15mA after device entering sleep mode.

### Solution

1. We initialize the pins interacting with the modem.
2. Check the modem power state before entering sleep mode. If the modem is on, then perform the power-off sequence to turn off the modem.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup() {
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::HIBERNATE));
}

void loop() {
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_network_compat.cpp
+++ b/system/src/system_network_compat.cpp
@@ -18,6 +18,7 @@
  */
 
 #include "hal_platform.h"
+#include "gpio_hal.h"
 
 /* FIXME: there should be a define that tells whether there is NetworkManager available
  * or not */
@@ -147,7 +148,24 @@ void network_off(network_handle_t network, uint32_t flags, uint32_t param, void*
 }
 
 int network_wait_off(network_handle_t network, system_tick_t timeout, void*) {
-    // Temporarily not necessary for Gen2 platforms.
+    /* This is a workaround for the power leakage issue that happens when the
+     * device is powered on and it enters sleep before the modem is initialized.
+     */
+#if PLATFORM_ID == PLATFORM_ELECTRON
+    cellular_pause(nullptr);
+    if (HAL_GPIO_Read(RXD_UC)) {
+        // Delay this long period so that the modem is fully on to accept the power-off sequence.
+        HAL_Delay_Milliseconds(12000);
+
+        HAL_GPIO_Write(PWR_UC, 0);
+        // >1.5 seconds on SARA R410M
+        // >1 second on SARA U2
+        // plus a little extra for good luck
+        HAL_Delay_Milliseconds(2000);
+        HAL_GPIO_Write(PWR_UC, 1);
+    }
+#endif
+
     return SYSTEM_ERROR_NONE;
 }
 
--- a/system/src/system_network_compat.cpp
+++ b/system/src/system_network_compat.cpp
@@ -18,7 +18,6 @@
  */
 
 #include "hal_platform.h"
-#include "gpio_hal.h"
 
 /* FIXME: there should be a define that tells whether there is NetworkManager available
  * or not */
@@ -42,7 +41,6 @@ uint32_t wlan_watchdog_duration = 0;
 volatile uint8_t SPARK_WLAN_RESET = 0;
 volatile uint8_t SPARK_WLAN_SLEEP = 0;
 volatile uint8_t SPARK_WLAN_STARTED = 0;
-volatile uint8_t SPARK_WLAN_INITIALIZED = 0; // Used by Electron only
 
 
 #if Wiring_WiFi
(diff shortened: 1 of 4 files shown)


system/src/system_network_compat.cpp
```diff
@@ -147,7 +148,24 @@ void network_off(network_handle_t network, uint32_t flags, uint32_t param, void*
 }
 
 int network_wait_off(network_handle_t network, system_tick_t timeout, void*) {
-    // Temporarily not necessary for Gen2 platforms.
+    /* This is a workaround for the power leakage issue that happens when the
+     * device is powered on and it enters sleep before the modem is initialized.
+     */
+#if PLATFORM_ID == PLATFORM_ELECTRON
+    cellular_pause(nullptr);
+    if (HAL_GPIO_Read(RXD_UC)) {
```
avtolstoy, line 156: I know we've discussed this, but now that I've taken a look at the modem code in the context of #2198, it should probably be fine to use `RI_UC` as long as `AT+URING` was not used.

I propose we should also try to hide as much of the platform-specific behavior as possible. Can we simply modify https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L1114 to perhaps take a look at the current `RI_UC` or some other state and perform the power-off sequence anyway? It should also be fine to introduce an additional cellular hal method if required.

avtolstoy (changes requested):

system/src/system_network_compat.cpp
XuGuohui: I'd like to refactor it after #2198 being merged to reduce conflicts on rebasing.

avtolstoy: You can base this PR on that branch for now.

XuGuohui: It's not just about how we perform the power-off sequence according to the current modem hal, but also the `SPARK_WLAN_STARTED` in the `off()` method in system_network_internal.h. When the power leakage encountered, the `SPARK_WLAN_STARTED` isn't set. Besides, this `off()` implementation is also shared with Photon/P1. I feel like if we're going to introduce a new HAL API for managing the modem power states, that is another approach we mentioned, which will refactor a bunch of files.

avtolstoy: If `SPARK_WLAN_STARTED` isn't set, we can detect that in the system layer and still perhaps perform `off_now()` there? Where the HAL can take care of checking the current state and turning off if needed.

avtolstoy: We can also initialize to by default to something like `UNKNOWN` state and act on it accordingly.

XuGuohui: Its type is `uint8_t`, and any other value except 0 and 1 will be `true` to perform some logic behavior accordingly for those where `if (SPARK_WLAN_STARTED)` is used.

XuGuohui: It's feasible though, we just need to be careful of where `SPARK_WLAN_STARTED` is referenced.

XuGuohui: >If SPARK_WLAN_STARTED isn't set, we can detect that in the system layer and still perhaps perform off_now() there?

If the modem is initialized and powered off by user application, `SPARK_WLAN_STARTED` is cleared as well, while in which case we don't need to perform that particular power-off sequence.

avtolstoy: I mean if we want to be careful with `SPARK_WLAN_STARTED`, okay, we can introduce another flag that would tell us that we are in an unknown state perhaps? :)

XuGuohui: Not bad. We have anyway introduced several global flags for managing the network status for Gen2 already. It can't be worse.

system/src/system_network_compat.cpp
```diff
@@ -41,6 +42,7 @@ uint32_t wlan_watchdog_duration = 0;
 volatile uint8_t SPARK_WLAN_RESET = 0;
 volatile uint8_t SPARK_WLAN_SLEEP = 0;
 volatile uint8_t SPARK_WLAN_STARTED = 0;
+volatile uint8_t SPARK_WLAN_INITIALIZED = 0; // Used by Electron only
```
avtolstoy, line 45: Let's keep it under `ManagedNetworkInterface` instead, as it's used just there.

avtolstoy (changes requested):

avtolstoy (approved):

PR #2198, opened 2020-09-14
[Electron] Recovery mechanics if modem becomes unresponsive

avtolstoy:
### Problem

Electrons can get blocked for substantial amounts of time whenever the modem crashes especially on devices with R410 in conjunction with socket operations which have very high timeouts.

### Solution

This PR attempts to introduce a similar recovery mechanic as on Gen 3 devices:
1. Watches for AT command timeouts
2. Sets a temporary error state
3. Attempts to check whether the modem responds again to a simple command
4. Requests modem to be reset

Apart from that, this PR cleans up the modem code a bit:
1. Hex mode removed
2. `UNLOCK()` removed as we are using RAII anyay
3. UDP `sendto()` should send a single datagram without splitting data into multiple chunks
4. UDP `recvfrom()` always attemps to read out the whole packet and trims it to the appropriate buffer length instead of reading it out potentially in chunks (which is a supported mechanism on the modem), but for simplicity and consistency we'll opt for a safer and more standard behavior
5. Separate timeouts for some socket operations to speed things up on U2-based devices
6. Periodic URC processing

This is a temporary workaround for 2.0.0-rc.2 only for certain modem operations. We should come up with a more generalized solution for any AT command.

### Steps to Test

1. SARA R4 02.00 firmware with 172 socket write error when there is an incoming packet arriving at the same time
2. `wiring/no_fixture_long_running`: `NETWORK_02_network_connection_recovers_after_ncp_failure` -- this emulates a modem crash
3. Socket operations in low level conditions on U2/R4

### Example App

N/A

### References

- [CH62973]
- [CH37789]
- [CH51091]

---

### Completeness

- [ ] Added documentation


avtolstoy:
- [CH37789]
- [CH51091]

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -2836,182 +2867,97 @@ int MDMParser::socketSend(int socket, const char * buf, int len)
...
-    }
-    return (len - bytesLeft);
-error:
-    UNLOCK();
-    return MDM_SOCKET_ERROR;
-#endif // defined(SOCKET_HEX_MODE)
+}
+
+int MDMParser::_socketError(void) {
+    sendFormated("AT+USOER");
```
technobly, line 2878: Gen 2 requires `"\r\n"` trailing most commands.  Without this, the command is not being recognized, and there is no `+USOER: x` URC.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1119,14 +1183,14 @@ bool MDMParser::powerOff(void)
...
         // Try to power off
         for (int i=0; i<3; i++) { // try 3 times
-            if (!_atOk()) {
+            if (_error || !_atOk()) {
                 if (_dev.dev == DEV_SARA_R410) {
                     check_ri = true;
                     // If memory issue is present, ensure we don't force a power off too soon
                     // to avoid hitting the 124 day memory housekeeping issue, AT+CPWROFF will
                     // handle this delay automatically, or timeout after 40s.
                     if (_memoryIssuePresent) {
```
technobly, line 1192: As long as we are going to be potentially entering this block more often, I think it would be safer to default to `_memoryIssuePresent = true` and set it to `false` only when `if (strcmp("L0.0.00.00.05.06,A.02.00", _verExtended) != 0) {`

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1110,7 +1174,7 @@ bool MDMParser::powerOff(void)
     LOCK();
     bool ok = false;
     bool continue_cancel = false;
-    bool check_ri = false;
+    bool check_ri = true;
```
technobly, line 1177: Why was this changed to `true` by default?  It's never set to `false`, only set true if the modem is R410.  IIRC we can only rely on RI_UC input as V_INT for R410, due to its hardware configuration.  This pin might not behave as expected for U2/G3.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1605,8 +1663,9 @@ bool MDMParser::getSignalStrength(NetStatus &status)
         MDM_INFO("\r\n[ Modem::getSignalStrength ] = = = = = = = = = =");
 
         // Ensure modem is responsive
-        if (!_atOk())
+        if (!_checkModem()) {
             goto cleanup;
+        }
```
technobly, line 1668: Could we add braces around the single line statement just below as well please? :D 
```
if (RESP_OK != waitFinalResp(nullptr, nullptr, COPS_TIMEOUT))
            goto cleanup;
```

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1681,16 +1739,16 @@ bool MDMParser::getDataUsage(MDM_DataUsage &data)
...
 }
 
 bool MDMParser::getCellularGlobalIdentity(CellularGlobalIdentity& cgi_) {
     LOCK();
 
     // Ensure modem is responsive
-    if (!_atOk())
+    if (!_checkModem()) {
         goto failure;
+    }
```
technobly, line 1751: More single line statements below to spruce up with braces, thanks! 😸 
```
if (RESP_OK != waitFinalResp(nullptr, nullptr, COPS_TIMEOUT))
        goto failure;
```

```
if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT))
        goto failure;
```

```
if (cgi_.size < sizeof(_net.cgi))
            goto failure;
```

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -339,7 +344,7 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
     return n;
 }
 
-bool MDMParser::_atOk(void)
+int MDMParser::_checkAtResponse(void)
```
technobly, line 347: There are still a number of places in the code where a `_atOk()` check is used instead of a `_checkModem()`, and currently `_atOk()` does not increment `_error` by itself.  We can't add `_error++;` to the `_atOk()` timeout case or it would double increment when `_checkModem()` is called, so should we convert all `_atOk()` uses to `_checkModem()`?

technobly (approved):
Looks good! Please address my comments added separately 😺

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: During testing we saw that we have to wait for the modem to actually turn off, otherwise it's very easy to catch it while it still performs the shutdown and it will respond with an OK to whatever command.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Let's do any unrelated refactoring separately.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: For now we are only handling any functionality that can be triggered by the system or the application (sockets, cellular signal, band select etc). The initialization logic for instance handles timeouts and failures on its own. If we are to do this type of refactoring it's a much larger task with more area of impact that will require significantly more time and not really fit for 2.0.0-rc.2.

We should queue this up for sometime in the future though.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: :+1:

hal/src/electron/modem/mdm_hal.cpp
technobly: Ok, based on our conversation this morning, you will be using the RX pin to check power state on U2/G3 and in that case we can remove this `check_ri` flag and just always check it.

hal/src/electron/modem/mdm_hal.cpp
technobly: I always try to clean up at least within the scope of the function I'm currently making changes in.  If we wait to clean up later, it will likely never happen any time soon.

hal/src/electron/modem/mdm_hal.cpp
technobly: I always try to clean up at least within the scope of the function I'm currently making changes in.  If we wait to clean up later, it will likely never happen any time soon.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: No strong opinions here, but we have some refactoring tasks coming up anyway for example logging as well as https://github.com/particle-iot/device-os/pull/2198#discussion_r490504313.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: I'm going to make that change as part of https://github.com/particle-iot/device-os/pull/2197, since it's primarily focused on the power state detection/management and RX idea original came from there.

I'm also thinking that perhaps it's also fine if we rely on RI_UC for U2 as well, because there is no reliable way to check the modem type without first talking to it. On a related note, I'm going to at least remove the code that resets the device type to UNKNOWN every time we power on.

hal/src/electron/modem/mdm_hal.cpp
technobly: Isn't this just changing 4-5 `_atOk()` statements to `_checkModem()` statements?  If the AT/OK check fails, an error will be accumulated.  Without this, we are fixing most of the cases where we previously decided AT/OK checks ahead of time were important, but leaving a handful of them that can fail without bubbling up to the system layer.  For all of these, we jump to failure, so accumulating an error in those cases makes a lot of sense.

Is the larger task just the testing effort?

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: And the only time RI_UC will become a problem is when `AT+URING` is used or we receive a call or SMS. Which isn't really the normal behavior.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Let's postpone this work until a later release. There shouldn't be any negative consequences leaving things as-is.

hal/src/electron/modem/mdm_hal.cpp
technobly: RI_UC (as RING, not V_INT) also toggles at a slow rate, so we can check it a few times to be sure it's in the right state if need be.

avtolstoy: :+1: Let me look into that.

avtolstoy: So, supposedly we can check two times with 1s in-between them to make sure we are not seeing RI being active as opposed to V_INT being down.

technobly: Sounds good 👍

PR #2203, opened 2020-09-23
Add characteristic discovery to BleService

mgolu:
**This PR is targeting to the feature/ble_cpp_callback/ch65803 branch**

### Problem

Allows for hierarchical discovery of characteristics inside services.

### Solution

Adds a discoverAllCharacteristics function to BleServiceImpl class. 

### Steps to Test

peer->discoverAllServices();
peer.getServiceByUUID(service, <SERVICE TO LOOK FOR>);
service.discoverAllCharacteristics();

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_ble.h
+++ b/wiring/inc/spark_wiring_ble.h
@@ -797,6 +797,9 @@ public:
     bool getCharacteristicByDescription(BleCharacteristic& characteristic, const char* desc) const;
     bool getCharacteristicByDescription(BleCharacteristic& characteristic, const String& desc) const;
     bool getCharacteristicByUUID(BleCharacteristic& characteristic, const BleUuid& uuid) const;
+    // In case that there are several characteristics with the same UUID.
+    Vector<BleCharacteristic> getCharacteristicByUUID(const BleUuid& uuid) const;
+    size_t getCharacteristicByUUID(BleCharacteristic* characteristics, size_t count, const BleUuid& uuid) const;
 
     // Fetch the discovered characteristics under a service.
     Vector<BleCharacteristic> characteristics(const BleService& service) const;
--- a/wiring/inc/spark_wiring_ble.h
+++ b/wiring/inc/spark_wiring_ble.h
@@ -576,13 +576,6 @@ public:
 
     BleUuid UUID() const;
 
-    // Discover all characteristics on this service
-    Vector<BleCharacteristic> discoverAllCharacteristics();
-
-    // Fetch the discovered characteristics on this service
-    Vector<BleCharacteristic> characteristics();
-    bool getCharacteristicByUUID(BleCharacteristic& characteristic, const BleUuid& uuid) const;
-
     BleService& operator=(const BleService& service);
 
     bool operator==(const BleService& service) const;
@@ -786,17 +779,27 @@ public:
     Vector<BleCharacteristic> discoverAllCharacteristics();
     ssize_t discoverAllCharacteristics(BleCharacteristic* characteristics, size_t count);
 
+    // Discover all characteristics of a service
+    Vector<BleCharacteristic> discoverCharacteristicsOfService(const BleService& service);
+
     // Fetch the discovered services on peer device.
-    Vector<BleService> services();
-    size_t services(BleService* services, size_t count);
+    Vector<BleService> services() const;
+    size_t services(BleService* services, size_t count) const;
     bool getServiceByUUID(BleService& service, const BleUuid& uuid) const;
 
     // Fetch the discovered characteristics on peer device.
-    Vector<BleCharacteristic> characteristics();
-    size_t characteristics(BleCharacteristic* characteristics, size_t count);
+    Vector<BleCharacteristic> characteristics() const;
+    size_t characteristics(BleCharacteristic* characteristics, size_t count) const;
     bool getCharacteristicByDescription(BleCharacteristic& characteristic, const char* desc) const;
     bool getCharacteristicByDescription(BleCharacteristic& characteristic, const String& desc) const;
     bool getCharacteristicByUUID(BleCharacteristic& characteristic, const BleUuid& uuid) const;
+
+    // Fetch the discovered characteristics under a service.
+    Vector<BleCharacteristic> characteristics(const BleService& service) const;
+    size_t characteristics(const BleService& service, BleCharacteristic* characteristics, size_t count) const;
+    bool getCharacteristicByDescription(const BleService& service, BleCharacteristic& characteristic, const char* desc) const;
+    bool getCharacteristicByDescription(const BleService& service, BleCharacteristic& characteristic, const String& desc) const;
+    bool getCharacteristicByUUID(const BleService& service, BleCharacteristic& characteristic, const BleUuid& uuid) const;
 
     int connect(const BleAddress& addr, const BleConnectionParams* params, bool automatic = true);
     int connect(const BleAddress& addr, const BleConnectionParams& params, bool automatic = true);
--- a/wiring/inc/spark_wiring_ble.h
+++ b/wiring/inc/spark_wiring_ble.h
@@ -576,6 +576,13 @@ public:
 
     BleUuid UUID() const;
 
+    // Discover all characteristics on this service
+    Vector<BleCharacteristic> discoverAllCharacteristics();
+
+    // Fetch the discovered characteristics on this service
+    Vector<BleCharacteristic> characteristics();
+    bool getCharacteristicByUUID(BleCharacteristic& characteristic, const BleUuid& uuid) const;
+
     BleService& operator=(const BleService& service);
 
     bool operator==(const BleService& service) const;
--- a/wiring/inc/spark_wiring_ble.h
+++ b/wiring/inc/spark_wiring_ble.h
@@ -781,11 +781,15 @@ public:
 
     // Discover all characteristics of a service
     Vector<BleCharacteristic> discoverCharacteristicsOfService(const BleService& service);
+    ssize_t discoverCharacteristicsOfService(const BleService& service, BleCharacteristic* characteristics, size_t count);
 
     // Fetch the discovered services on peer device.
     Vector<BleService> services() const;
     size_t services(BleService* services, size_t count) const;
     bool getServiceByUUID(BleService& service, const BleUuid& uuid) const;
+    // In case that there are several services with the same UUID.
+    Vector<BleService> getServiceByUUID(const BleUuid& uuid) const;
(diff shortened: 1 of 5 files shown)


wiring/src/spark_wiring_ble.cpp
```diff
@@ -883,10 +885,75 @@ class BleServiceImpl {
...
+
+    bool& characteristicsDiscovered() {
+        return characteristicsDiscovered_;
+    }
+
+    Vector<BleCharacteristic>& characteristics() {
+        return characteristics_;
+    }
+
+    int discoverAllCharacteristics(BleService& service) {
```
XuGuohui, line 900: This is almost identical to the same function name inside `BleDiscoveryDelegator`, we can potentially just introduce new method in `BlePeerDeviceImpl` to discover characteristics under a given service.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -883,10 +885,75 @@ class BleServiceImpl {
...
+        return SYSTEM_ERROR_NONE;
+    }
+
 private:
+    BleConnectionHandle connHandle_; // For peer service
     BleUuid uuid_;
     BleAttributeHandle startHandle_;
     BleAttributeHandle endHandle_;
+    bool characteristicsDiscovered_;
+    Vector<BleCharacteristic> characteristics_;
```
XuGuohui, line 928: We can probably share the vector that is created in the `BlePeerDeviceImpl` to save ram and introduce new method in `BlePeerDeviceImpl` to discover characteristics under a given service.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -883,10 +885,75 @@ class BleServiceImpl {
...
+
+    static void onCharacteristicsDiscovered(const hal_ble_char_discovered_evt_t* event, void* context) {
```
XuGuohui, line 930: Same here. Please refer to: https://github.com/particle-iot/device-os/pull/2203/files#r493623290

wiring/inc/spark_wiring_ble.h
```diff
@@ -426,6 +426,13 @@ class BleService {
 
     BleUuid UUID() const;
 
+    // Discover all characteristics on this service
+    Vector<BleCharacteristic> discoverAllCharacteristics();
+
+    // Fetch the discovered characteristics on this service
+    Vector<BleCharacteristic> characteristics();
+    bool getCharacteristicByUUID(BleCharacteristic& characteristic, const BleUuid& uuid) const;
```
XuGuohui, line 434: This method can also implemented in `BlePeerDeviceImpl` potentially.

XuGuohui (changes requested):
Thank you @mgolu for submitting this great PR! There is still something we can improve as I commented. If you'd like, I'm glad to take over this PR to make the final change.

mgolu:
> Thank you @mgolu for submitting this great PR! There is still something we can improve as I commented. If you'd like, I'm glad to take over this PR to make the final change.

Yes, great comments. I agree we can refactor, I just didn't want to touch existing APIs yet. If you have the time to work on it, go for it!

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1549,15 +1468,35 @@ bool BlePeerDevice::getServiceByUUID(BleService& service, const BleUuid& uuid) c
...
+Vector<BleCharacteristic> BlePeerDevice::characteristics() const {
     return impl()->characteristics();
 }
 
-size_t BlePeerDevice::characteristics(BleCharacteristic* chars, size_t count) {
-    CHECK_TRUE(chars && count > 0, SYSTEM_ERROR_INVALID_ARGUMENT);
+Vector<BleCharacteristic> BlePeerDevice::characteristics(const BleService& service) const {
+    Vector<BleCharacteristic> characteristics;
+    for (const auto& characteristic : impl()->characteristics()) {
+        if (characteristic.impl()->svcUUID() == service.impl()->UUID()) {
```
mgolu, line 1478: This checks if a characteristic has a Service UUID that is the same as the one we're looking for. If a GATT server has multiple services with the same UUID, then this would return characteristics from multiple Services, instead of just the Service we're looking for.

XuGuohui: Technically, we should not add more than two services with the same UUID for a BLE peripheral. UUID is shorted for Universal Unique Identifier and BLE introduces it, so we should comply with the way we use it. We can absolutely fix it by checking the attribute handle range, but I'm wondering if it is worthy for us to introduce a wiring API that returns a vector of services with the same UUID for particular and abnormal cases.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1587,6 +1526,31 @@ bool BlePeerDevice::getCharacteristicByUUID(BleCharacteristic& characteristic, c
     return false;
 }
 
+bool BlePeerDevice::getCharacteristicByDescription(const BleService& service, BleCharacteristic& characteristic, const char* desc) const {
+    CHECK_TRUE(desc, false);
+    for (auto& existChar : impl()->characteristics()) {
+        if (existChar.impl()->svcUUID() == service.UUID() && !strcmp(existChar.description().c_str(), desc)) {
```
mgolu, line 1532: This also checks for matching UUIDs, so it will be a problem if GATT server has multiple services with the same UUID. Can we instead check for handles being in range?

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1587,6 +1526,31 @@ bool BlePeerDevice::getCharacteristicByUUID(BleCharacteristic& characteristic, c
...
+    return true;
+}
+
+bool BlePeerDevice::getCharacteristicByDescription(const BleService& service, BleCharacteristic& characteristic, const String& desc) const {
+    return getCharacteristicByDescription(service, characteristic, desc.c_str());
+}
+
+bool BlePeerDevice::getCharacteristicByUUID(const BleService& service, BleCharacteristic& characteristic, const BleUuid& uuid) const {
+    for (auto& existChar : impl()->characteristics()) {
+        if (existChar.impl()->svcUUID() == service.UUID() && existChar.UUID() == uuid) {
```
mgolu, line 1546: Same issue here. Should check by service handles rather than UUID.

mgolu (commented):
This is great to save RAM by keeping just one copy of characteristics. I think the application will have to keep track of the peer associated with each service (for example, when connecting to multiple peers).

Maybe we can add a call to the BleService class that calls the peer's discoverCharacteristicsOfService? That way the application doesn't have to keep track of which peer is for each service.

wiring/src/spark_wiring_ble.cpp
mgolu: Unique means that it identifies a unique service, but there can be multiple copies of the same service on a device (for example, multiple Battery Services with UUID 0x180F on a device with more than one battery).

Here's from the Bluetooth specification:
"A device or higher level specification may have multiple service definitions and may have multiple service definitions with the same service UUID."

From Core specification v5.2, Volume 3, Part G, section 3.1

XuGuohui: 👍 That makes perfect sense! Thanks for diving into the specification and pointing it out! I believe that we are on the same page now. I will make proper changes later on.

XuGuohui:
>Maybe we can add a call to the BleService class that calls the peer's discoverCharacteristicsOfService?

The initial objective of the BLE wiring APIs are designed to be easy to use for non professional developers. Thus, we just provide two main mechanisms for BLE operation, data transaction oriented and non-data transaction oriented. For non-data transaction oriented operation, we have the `BLE` object for local device operation and the `BlePeerDevice` for peer device operation. And for data transaction oriented class, we have the `BleCharacteristic` class. The `BleService` class doesn't need to touch anything about data/non-data transaction operation. I'd prefer not to introduce more APIs under the service class to make things complicate. Could you elaborate how it affects if the central is connected with multiple peers while the `BleService` doesn't associated with a certain peer?

mgolu:
> The initial objective of the BLE wiring APIs are designed to be easy to use for non professional developers. Thus, we just provide two main mechanisms for BLE operation, data transaction oriented and non-data transaction oriented. For non-data transaction oriented operation, we have the `BLE` object for local device operation and the `BlePeerDevice` for peer device operation. And for data transaction oriented class, we have the `BleCharacteristic` class. The `BleService` class doesn't need to touch anything about data/non-data transaction operation. I'd prefer not to introduce more APIs under the service class to make things complicate. Could you elaborate how it affects if the central is connected with multiple peers while the `BleService` doesn't associated with a certain peer?

The use case would be for an Enterprise customer who might have multiple peers connected, which might be the same Profile or type. Then, they would have multiple BleService instances with the same UUID, but related to different peers. If they want to discover characteristics on an instance of BleService, that's currently impossible since there's no API on BleService to discover characteristics, and there's no reference to the BlePeerDevice that the instance is associated with. That means that the developer needs to keep track themselves of each BleService instance association to a BlePeer.

XuGuohui:
I see no reason to keep tracking a BLE service with a peer device in user application. For discovering characteristics under a peer associated service, we can call `peer.discoverCharacteristicsOfService()` that is introduced in this PR. Other than that, we don't need to operate a BLE service. Whenever we want to retrieve a peer associated service, you can call `peer.services()` or `peer.getServiceByUUID()`. So we just need to track the peer devices it connected. As for data transmission, we use characteristic methods, which is already peer associated.

mgolu:
I tested with the new proposed BLE Gateway library (https://github.com/particle-iot/ble-gateway-library) and the new changes are working well. Just submitted a small fix.

XuGuohui (approved):

XuGuohui:
@avtolstoy Do you have any comment here?

mgolu:
Can this branch be merged into develop?

XuGuohui:
> Can this branch be merged into develop?

We will get it merged in this sprint (in two weeks). Cc. @avtolstoy

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1001,10 +1002,24 @@ class BleServiceImpl {
         return endHandle_;
     }
 
+    bool& characteristicsDiscovered() {
```
avtolstoy, line 1005: https://github.com/particle-iot/device-os/pull/1847#discussion_r310971495 :)

Nit: This is a really weird pattern when considering trivial types. I'd rather have this as a public member then or use setters/getters.

avtolstoy (approved):

avtolstoy (approved):

PR #2205, opened 2020-09-29
[wiring] Network: introduce isOn() and isOff() APIs to enquiry the modem power state

XuGuohui:
As the title describes.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void onNetworkStatusUpdated(system_event_t event, int param) {
    if (event == network_status) {
        switch (param) {
            case network_status_on: {
                Log.info("network_status_on");
                break;
            }
            case network_status_off: {
                Log.info("network_status_off");
                break;
            }
            case network_status_connected: {
                Log.info("network_status_connected");
                break;
            }
        }
    }
}

void setup() {
    System.on(network_status, onNetworkStatusUpdated);

    delay(1000);

    if (Cellular.isOn()) {
        Log.info("Modem is on.");
    } else if (Cellular.isOff()) {
        Log.info("Modem is off.");
    } else {
        Log.info("Modem is in an unknown state.");
    }

    Log.info("Turning on the modem...");
    Cellular.on();

    while (Cellular.isOff()) {
        Log.info("modem is off.");
    }
    while (!Cellular.isOn()) {
        Log.info("powering on...");
        delay(500);
    }
    Log.info("Modem is turned on.");

    delay(3s);

    Log.info("Turning off the modem...");
    Cellular.off();

    while (Cellular.isOn()) {
        Log.info("modem is on.");
    }
    while (!Cellular.isOff()) {
        Log.info("powering off...");
        delay(500);
    }
    Log.info("Modem is turned off.");
}

void loop() {
}
```
---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/api/ifapi.h
+++ b/hal/network/api/ifapi.h
@@ -60,7 +60,9 @@ typedef enum if_xflags_t {
     IFXF_DHCP6        = 0x20000,
     IFXF_AUTOIP       = 0x40000,
 
-    IFXF_CANTCHANGE   = 0
+    IFXF_READY        = 0x100000,
+
+    IFXF_CANTCHANGE   = (IFXF_READY)
 } if_xflags_t;
 
 #ifndef IF_NAMESIZE
@@ -141,7 +143,8 @@ typedef enum if_event_type_t {
     IF_EVENT_LINK                 = 0x04,
     IF_EVENT_ADDR                 = 0x05,
     IF_EVENT_LLADDR               = 0x06,
-    IF_EVENT_POWER_STATE          = 0x07
+    IF_EVENT_POWER_STATE          = 0x07,
+    IF_EVENT_PHY_STATE            = 0x08,
 } if_event_type_t;
 
 typedef enum if_state_t {
@@ -164,10 +167,11 @@ typedef enum if_power_state_t {
     IF_POWER_STATE_POWERING_UP = 0x04,
 } if_power_state_t;
 
-typedef enum if_ncp_state_t {
-    IF_NCP_STATE_OFF = 0x00,
-    IF_NCP_STATE_ON = 0x01
-} if_ncp_state_t;
+typedef enum if_phy_state_t {
+    IF_PHY_STATE_UNKNOWN = 0x00,
+    IF_PHY_STATE_OFF = 0x01,
+    IF_PHY_STATE_ON = 0x02
+} if_phy_state_t;
 
 struct if_event_state {
     uint8_t state;
@@ -178,6 +182,10 @@ struct if_event_link_state {
 };
 
 struct if_event_power_state {
+    uint8_t state;
+};
+
+struct if_event_phy_state {
     uint8_t state;
 };
 
@@ -201,6 +209,7 @@ struct if_event {
         struct if_event_addr* ev_if_addr;
         struct if_event_lladdr* ev_if_lladdr;
         struct if_event_power_state* ev_power_state;
+        struct if_event_phy_state* ev_phy_state;
     };
 };
 
@@ -264,7 +273,6 @@ int if_request(if_t iface, int type, void* req, size_t reqsize, void* reserved);
 void if_notify_event(if_t iface, const struct if_event* evt, void* reserved);
 
 int if_get_power_state(if_t iface, if_power_state_t* state);
-int if_get_ncp_state(if_t iface, if_ncp_state_t* state);
 
 #ifdef __cplusplus
 }
--- a/hal/network/api/ifapi.h
+++ b/hal/network/api/ifapi.h
@@ -164,6 +164,11 @@ typedef enum if_power_state_t {
     IF_POWER_STATE_POWERING_UP = 0x04,
 } if_power_state_t;
 
+typedef enum if_ncp_state_t {
+    IF_NCP_STATE_OFF = 0x00,
+    IF_NCP_STATE_ON = 0x01
+} if_ncp_state_t;
+
 struct if_event_state {
     uint8_t state;
 };
@@ -259,6 +264,7 @@ int if_request(if_t iface, int type, void* req, size_t reqsize, void* reserved);
 void if_notify_event(if_t iface, const struct if_event* evt, void* reserved);
 
 int if_get_power_state(if_t iface, if_power_state_t* state);
+int if_get_ncp_state(if_t iface, if_ncp_state_t* state);
 
 #ifdef __cplusplus
 }
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -376,6 +376,7 @@ protected:
 
     virtual int on_now()=0;
     virtual void off_now()=0;
+    virtual bool is_powered()=0;
 
     virtual int process_now()=0;
 
@@ -620,7 +621,7 @@ public:
 
     bool isOff() override
     {
-        return WLAN_INITIALIZED && !SPARK_WLAN_STARTED;
+        return !SPARK_WLAN_STARTED && !is_powered();
     }
 
     void off(bool disconnect_cloud=false) override
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -141,8 +141,14 @@ struct NetworkInterface
     virtual network_interface_t network_interface()=0;
     virtual void setup()=0;
 
+<<<<<<< HEAD
     virtual int on()=0;
+=======
+    virtual int on()=0;
+    virtual bool isOn()=0;
+>>>>>>> [wiring] Network: introduce isOn() and isOff() APIs to enquiry the modem power state.
     virtual void off(bool disconnect_cloud=false)=0;
+    virtual bool isOff()=0;
     virtual void connect(bool listen_enabled=true)=0;
     virtual bool connecting()=0;
     virtual void connect_cancel(bool cancel)=0;
@@ -609,6 +615,24 @@ public:
             return ret;
         }
         return 0;
+    }
+
+    bool isOn() override
+    {
+#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
+        return !WLAN_INITIALIZED || SPARK_WLAN_STARTED;
+#else
+        return SPARK_WLAN_STARTED;
+#endif
+    }
+
+    bool isOff() override
+    {
+#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
+        return WLAN_INITIALIZED && !SPARK_WLAN_STARTED;
+#else
+        return !SPARK_WLAN_STARTED;
+#endif
     }
 
     void off(bool disconnect_cloud=false) override
--- a/system/src/system_network_internal.h
+++ b/system/src/system_network_internal.h
@@ -141,12 +141,8 @@ struct NetworkInterface
     virtual network_interface_t network_interface()=0;
     virtual void setup()=0;
 
-<<<<<<< HEAD
-    virtual int on()=0;
-=======
     virtual int on()=0;
     virtual bool isOn()=0;
->>>>>>> [wiring] Network: introduce isOn() and isOff() APIs to enquiry the modem power state.
     virtual void off(bool disconnect_cloud=false)=0;
     virtual bool isOff()=0;
     virtual void connect(bool listen_enabled=true)=0;
@@ -619,20 +615,12 @@ public:
 
     bool isOn() override
     {
-#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
-        return !WLAN_INITIALIZED || SPARK_WLAN_STARTED;
-#else
-        return SPARK_WLAN_STARTED;
-#endif
+        return WLAN_INITIALIZED && SPARK_WLAN_STARTED;
     }
 
     bool isOff() override
     {
-#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
         return WLAN_INITIALIZED && !SPARK_WLAN_STARTED;
-#else
-        return !SPARK_WLAN_STARTED;
-#endif
     }
 
     void off(bool disconnect_cloud=false) override
--- a/system/src/system_network_manager.cpp
+++ b/system/src/system_network_manager.cpp
@@ -586,6 +586,7 @@ void NetworkManager::handleIfPowerState(if_t iface, const struct if_event* ev) {
     }
 }
 
+// FIXME: this currently works somewhat properly only if there is just 1 network interface.
 void NetworkManager::handleIfPhyState(if_t iface, const struct if_event* ev) {
     if (networkStatus_ == NetworkStatus::NETWORK_STATUS_POWERING_ON && ev->ev_phy_state->state == IF_PHY_STATE_ON) {
         system_notify_event(network_status, network_status_on);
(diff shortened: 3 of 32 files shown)


system/src/system_network_manager.cpp
```diff
@@ -800,6 +800,28 @@ NetworkManager::InterfaceRuntimeState* NetworkManager::getInterfaceRuntimeState(
     return nullptr;
 }
 
+bool NetworkManager::isInterfacePowerState(if_t iface, if_power_state_t state) const {
+    bool ret = false;
+    if_power_state_t pwr = IF_POWER_STATE_NONE;
+    if (!iface) {
+        for_each_iface([&](if_t iface, unsigned int curFlags) {
+            if (if_get_power_state(iface, &pwr) != SYSTEM_ERROR_NONE) {
+                return;
```
technobly, line 809: Would this just return the first network interface found when no specific network IF type is specified?  E.g., Network.isOn() could return on/off when the first IF is found?  I'm not sure if that's good or bad, but some comments added to this function to explain it's behavior would be helpful.

Also, I don't think this should be returning true/false (which is missing) but likely doing something with the following `if (pwr == state) { }`.

technobly (changes requested):
Looking forward to having these essential APIs! Would be nice to see how they can be used with the Future API and waitFor/waitUntil, perhaps in some tests? :D

system/src/system_network_internal.h
```diff
@@ -622,6 +624,24 @@ class ManagedNetworkInterface : public NetworkInterface
         }
     }
 
+    bool isOn() override
+    {
+#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
+        return !WLAN_INITIALIZED || SPARK_WLAN_STARTED;
```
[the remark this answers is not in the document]
avtolstoy: There probably shouldn't be a `PLATFORM_ID == PLATFORM_ELECTRON` check and the condition for on is as Brett says: `SPARK_WLAN_STARTED && WLAN_INITIALIZED`.

system/src/system_network_internal.h
```diff
@@ -622,6 +624,24 @@ class ManagedNetworkInterface : public NetworkInterface
...
+#else
+        return SPARK_WLAN_STARTED;
+#endif
+    }
+
+    bool isOff() override
+    {
+#if PLATFORM_ID == PLATFORM_ELECTRON // Electron
+        return WLAN_INITIALIZED && !SPARK_WLAN_STARTED;
```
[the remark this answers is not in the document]
avtolstoy: Probably let's remove conditional for Electron (unless we know for a fact that BCM on Photon is off on boot).

This looks okay, because technically on Electron `!WLAN_INITIALIZED && !SPARK_WLAN_STARTED` means that the state is unknown. Ideally we'd query the modem for its state, but returning false for both `isOn()` and `isOff()` might also be valid.

system/src/system_network_manager.cpp
avtolstoy: It doesn't take into account just the first interface. Lambda is executed for each of them. If at least one is on - return true.

avtolstoy (changes requested):

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: We need to define what `isOn()`/`isOff()` means. Since the modem may reside in a state that it is powered on, but not initialized. In corresponding to Gen3 platforms, the NCP power state may be on, while the NCP client is not ready to accept AT commands. 

The implementation of the new APIs on Gen3 does return the modem power state, despite of whether the NCP client is ready or not. 

We had a fix for Electron power leakage: https://github.com/particle-iot/device-os/pull/2197, which means that the SARA module is on by default on device powering up. And I just did a test to verify that the BCM module is on by default as well:
```
void setup() {
    pinMode(D0, INPUT_PULLUP);
    if (digitalRead(D0) == HIGH) {
        Particle.connect();
    } else {
        delay(20s);
        WiFi.on();
        delay(20s);
        WiFi.off();
    }
}
```
The power consumption on powering up is higher than after we turn off the module on Photon.

If `isOn()` intends to indicate the power state, than we should return `return !WLAN_INITIALIZED || SPARK_WLAN_STARTED;` for all Gen2 platforms here.

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: If `isOff()` intends to indicate the power state, then we should return return `return WLAN_INITIALIZED && !SPARK_WLAN_STARTED;` for all Gen2 platforms here.

system/src/system_network_manager.cpp
XuGuohui: Here we just return the lambda function with `void`, not returning the `isInterfacePowerState()`.

system/src/system_network_internal.h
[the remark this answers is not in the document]
avtolstoy: If you think from the application perspective you care about two things:
1. Is whatever I need to be off and not consume power really off or do I need to perform some action to turn it off (both logically and physically off)
2. Is whatever I turned on really on and ready to proceed further (in our case - accept commands or initiate connection for example) (both physically and logically on - i.e. passed some low-level initialization)

That would at least be my interpretation.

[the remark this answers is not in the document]
XuGuohui: If I understand correctly, the modem may reside in one of the following states:
**Physically Off <-> Physically On but Logically off <-> Logically On**
So,
1. The `isOff()` should return `true` only if the modem resides in the **Physically Off** state
2. The `isOn()` should return `true` only if the modem resides in the **Logically On** state.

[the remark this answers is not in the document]
avtolstoy: 1. Physically off and logically off (which means that it's not supposed to get turned on at all). For example we encountered some problem and had to hard reset modem: those conditions should not report `isOff()` as `true`.
2. Correct.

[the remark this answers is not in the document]
XuGuohui: That makes sense 👍

XuGuohui:
@technobly @avtolstoy I just revised the PR. Looking forward to your review and any comments.

system/src/system_network_manager.cpp
technobly: I would still love to see some comments added to this function to describe its behavior.  Something like, "This function checks each network interface for the intended power state, and returns true when the first one that matches the desired state is found.  Some remaining network interfaces might not be checked."

system/src/system_network_internal.h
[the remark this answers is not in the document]
technobly: I was still quite confused why isOff() would `return WLAN_INITIALIZED && ...` since it should indicate we are powered off (!SPARK_WLAN_STARTED) and de-initialized (!WLAN_INITIALIZED).  But when I looked through system_network_internal.h I don't see off() setting WLAN_INITIALIZED = 0;  So isOff() would still be false after calling off() with returning `!WLAN_INITIALIZED &&` .  This seems like an error in system_network_internal.h and we should be setting WLAN_INITIALIZED = 0 in off(), which would then allow us to `return !WLAN_INITIALIZED && !SPARK_WLAN_STARTED;` for isOff().

technobly (changes requested):

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: I know why you are confused, as you were expecting the `WLAN_INITIALIZED` flag to indicating the low-level modem state. While the truth is that the `WLAN_INITIALIZED` was introduced in previous PR https://github.com/particle-iot/device-os/pull/2197, which is to fix the power leakage issue on Electron on bootup. The context is that when the Electron is powered up the modem does consuming high current if user application didn't touch any network API and then go to sleep. So this flag only indicates whether the modem is initialized at least once so that the modem power state resides in a well known state and you can see that it is set only in the `on()` function in system_network_internal.h. 

That's why `isOff()` should return `WLAN_INITIALIZED && !SPARK_WLAN_STARTED`, as if `WLAN_INITIALIZED` is not set, we consider that the modem is powered up. As long as the `WLAN_INITIALIZED` is set, the modem state is tracked by `SPARK_WLAN_STARTED`, which is changed along with the `on()`/`off()` API.

system/src/system_network_manager.cpp
XuGuohui: Will do.

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: If `WLAN_INITIALIZED` is not set, the modem is in an unknown state, neither `isOn()` nor `isOff()` should return true.

system/src/system_network_manager.cpp
XuGuohui: >Some remaining network interfaces might not be checked

If you go into the `for_each_iface()` you'll see that every interface is checked. The `return` here only exits the lambda function. I'll still add the comment, but it would be like:
```
This function checks each network interface for the intended power state, and returns true as long as one that matches the desired state is found
```

hal/network/api/ifapi.h
```diff
@@ -141,7 +143,8 @@ typedef enum if_event_type_t {
     IF_EVENT_LINK                 = 0x04,
     IF_EVENT_ADDR                 = 0x05,
     IF_EVENT_LLADDR               = 0x06,
-    IF_EVENT_POWER_STATE          = 0x07
+    IF_EVENT_POWER_STATE          = 0x07,
+    IF_EVENT_PHY_STATE            = 0x08,
```
avtolstoy, line 147: Sounds similar to power state, but perhaps it's probably the closest to what we want to say, so :+1:

wiring/inc/spark_wiring_cellular.h
```diff
@@ -50,6 +50,12 @@ class CellularClass : public NetworkClass
     void off() {
         network_off(*this, 0, 0, NULL);
     }
+    bool isOn() {
```
avtolstoy, line 53: Nit: could potentially just add to `NetworkClass`. Yeah, it's a bit messy right now, but this should be generic and available in the parent class.

system/src/system_network_manager.cpp
```diff
@@ -577,6 +586,14 @@ void NetworkManager::handleIfPowerState(if_t iface, const struct if_event* ev) {
     }
 }
 
+void NetworkManager::handleIfPhyState(if_t iface, const struct if_event* ev) {
+    if (networkStatus_ == NetworkStatus::NETWORK_STATUS_POWERING_ON && ev->ev_phy_state->state == IF_PHY_STATE_ON) {
```
avtolstoy, line 590: Let's add a FIXME that this currently works somewhat properly only if there is just 1 network interface.

system/src/system_network_manager.cpp
```diff
@@ -830,6 +847,51 @@ NetworkManager::InterfaceRuntimeState* NetworkManager::getInterfaceRuntimeState(
...
+    } else {
+        if (if_get_power_state(iface, &pwr) != SYSTEM_ERROR_NONE) {
+            return false;
+        }
+        ret = (pwr == state) ? true : false;
+    }
+    return ret;
+}
+
+bool NetworkManager::isInterfaceReady(if_t iface) const {
```
avtolstoy, line 872: Let's add a comment that it's not same as wiring ready to avoid confusion.

system/src/system_network_manager_api.cpp
```diff
@@ -261,6 +261,36 @@ void network_on(network_handle_t network, uint32_t flags, uint32_t param, void*
     }());
 }
 
+bool network_is_on(network_handle_t network, void* reserved) {
+    if (network == NETWORK_INTERFACE_ALL) {
+        return (NetworkManager::instance()->isInterfacePowerState(nullptr, IF_POWER_STATE_UP) &&
```
avtolstoy, line 266: Nit: could potentially combine these two calls within network manager?

avtolstoy (approved):

avtolstoy:
@XuGuohui We can do this in a separate PR, but let's also make sure that `isOff()` on Electron will actually report `true` on boot if the modem is indeed off, as we talked about it. This should be possible by calling into the modem layer, querying the RI pin state and caching the state, so that the future `isOff()` calls do not have to suffer a 1s delay.

wiring/inc/spark_wiring_cellular.h
XuGuohui: Will remove these from spark_wiring_cellular.h and spark_wiring_wifi.h

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1267,6 +1267,10 @@ bool MDMParser::powerState(void) const {
     state = HAL_GPIO_Read(RI_UC);
     // If RI is still low - we are off
     // If RI is high - we are on
+    if (getModemStateChangeCount() == 0) {
+        _incModemStateChangeCount();
+        _pwr = state;
+    }
```
XuGuohui, line 1273: @avtolstoy Would you like to add this change to the PR: https://github.com/particle-iot/device-os/pull/2217 as well? This could reduce 1 s that is cost on verifying the power state on electron.

system/src/system_network_internal.h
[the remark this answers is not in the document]
technobly: Ok thanks for the explanation.  `WLAN_INITIALIZED` should be commented in firmware with this understanding of how it should be used and what it means, because just looking at it you can't tell that it should not be reset if the modem is uninitialized.  We possibly shouldn't be setting `WLAN_INITIALIZED` back to 0 in `on()` if it's meant to be a static flag that only gets set when the modem is first turned on.  You can't say calling on() twice would turn off the modem, and the second time through it might fail the on_now() call and report that it's off `WLAN_INITIALIZED = 0`, when it might have been turned on the first time and is still on.

technobly (approved):

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: Yeah. I agree with you. @avtolstoy Any reason why we were adding this in: https://github.com/particle-iot/device-os/commit/c6fc6c3da8bef1fe47cb5afadc2056d980109b48#diff-f61a83197bcd7fe3529f7ca6e612a99d1222bf9d8080af2b0c4c7693fb9bf95dR620?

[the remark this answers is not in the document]
avtolstoy: The PR that contains that commit explains why we are doing this. It is reset to keep track of whether the power-on sequence actually succeeded and if not retry it on the subsequent attempts. When we request the modem to turn on we set `SPARK_WLAN_STARTED` to 1. There is a lot of other code that relies on this condition, so we cannot use it to keep track of whether we need to retry or not. So, `WLAN_INITIALIZED` is used for that and will be reset to 0 while powering on or after a failed power on attempt to indicate that we are in an unknown state. This plus `SPARK_WLAN_STARTED == 1` allows us to retry power-on.

system/src/system_network_wifi.h
```diff
@@ -223,6 +236,11 @@ class WiFiNetworkInterface : public ManagedIPNetworkInterface<WLanConfig, WiFiNe
         return wlan_get_hostname(buf, buf_len, NULL);
     }
 
+private:
+    // It's tricky to detect the modem power state.
+    // We assume that every time the device being powered up, the module is powered on.
+    // The the power state then be tracked by the on_now() and off_now() APIs.
+    bool powered = true;
```
avtolstoy, line 243: `volatile`

avtolstoy (approved):

system/src/system_network_wifi.h
avtolstoy, line 243: I think we previously figured out that on boot BCM43362 is powered off, so we should be able to safely set this to `false` by default?

XuGuohui: I think it's on by default: https://github.com/particle-iot/device-os/pull/2205#discussion_r529387030

system/src/system_network_internal.h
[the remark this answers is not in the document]
XuGuohui: That makes sense. Thanks for the explanation! 👍

avtolstoy (approved):

PR #2206, opened 2020-09-29
Disconnect from the cloud before going to sleep

sergeuz:
### Problem

By default, the Sleep 2.0 API leaves the protocol layer in whatever state it was before going to sleep and continues to use it after waking up from sleep without checking whether the session with the server is still valid.

### Solution

Always disconnect from the cloud before going to sleep unless the network interface is used as a wake-up source. In the latter case, send a PING message to the server after waking up from sleep.

### Steps to Test

1. Verify that a sleepy device that doesn't use its network interface as a wake-up source successfully resumes its session when reconnecting to the cloud from a different address.
2. Using one of the network interfaces as a wake-up source, verify that the device sends a ping message to the cloud after waking up from sleep.

### References

- [ch64208]

### DIFF
--- a/system/src/system_cloud_internal.h
+++ b/system/src/system_cloud_internal.h
@@ -29,8 +29,6 @@
 
 void Spark_Signal(bool on, unsigned, void*);
 void Spark_SetTime(unsigned long dateTime);
-void Spark_Sleep(unsigned duration);
-void Spark_Wake();
 int Spark_Save(const void* buffer, size_t length, uint8_t type, void* reserved);
 int Spark_Restore(void* buffer, size_t max_length, uint8_t type, void* reserved);
 
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -79,21 +79,37 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
 
     SystemSleepConfigurationHelper configHelper(config);
 
-    bool cloudResume = false;
+    bool reconnectCloud = false;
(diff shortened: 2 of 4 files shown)


system/src/system_cloud_internal.h
```diff
@@ -29,8 +29,6 @@
 
 void Spark_Signal(bool on, unsigned, void*);
 void Spark_SetTime(unsigned long dateTime);
-void Spark_Sleep(unsigned duration);
```
avtolstoy, line 32: :+1:

system/src/system_sleep.cpp
```diff
@@ -79,21 +79,37 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
...
+            CloudConnectionSettings::instance()->setPendingDisconnectOptions(std::move(opts));
         }
-        cloudResume = spark_cloud_flag_auto_connect();
+        // The CLOUD_DISCONNECT_GRACEFULLY flag doesn't really enable graceful disconnection mode,
+        // it merely indicates that it is ok to disconnect gracefully in this specific case.
+        // Whether the system will actually disconnect from the cloud gracefully or not depends
+        // on the disconnection options that can be set via Particle.setDisconnectOptions() or
+        // Particle.disconnect().
+        //
+        // TODO: Rename CLOUD_DISCONNECT_GRACEFULLY to CLOUD_DISCONNECT_IMMEDIATELY and invert its
```
avtolstoy, line 104: :+1:

avtolstoy (approved):
Thanks for taking care of this!

PR #2208, opened 2020-10-05
[u-blox][Boron / B SoM] PPP workaround for SARA R4 ConfReq behavior

avtolstoy:
### Problem

SARA R4 PPP implementation is peculiar and for unknown reason after IPCP is already up, it sends up an _empty_ ConfReq. If we follow the standard to the T, we are supposed to bring IPCP down and restart negotiation. This increases warm bootup cloud connection times significantly.

### Solution

The solution is to utilize a new PPP option introduced in https://github.com/particle-iot/lwip/pull/8 to slightly go off standard and ACK these Configure-Requests in an already opened state if the protocol handler does not disagree with the options provided in them and we already in an appropriate state (authed, and it's not LCP protocol that we need to renegotiate)

This PR also disables `ATH` command on R4-based devices as it does not function anyway. See https://github.com/particle-iot/device-os/pull/2188

### Steps to Test

1. Build with PPP logs enabled
2. See logs from https://github.com/particle-iot/lwip/pull/8
3. Compare behavior and assess the connection times

### Example App

N/A

### References

- https://github.com/particle-iot/lwip/pull/8
- [CH64715]

---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/network/lwip/ppp_client.cpp
+++ b/hal/network/lwip/ppp_client.cpp
@@ -88,6 +88,22 @@ void Client::init() {
     netif_set_client_data(&if_, netifClientDataIdx_, this);
     UNLOCK_TCPIP_CORE();
 
+    // XXX:
+    if (platform_current_ncp_identifier() == PLATFORM_NCP_SARA_R410) {
+      // SARA R4 PPP implementation is broken and often times we receive
+      // an empty or non-conflicting Configure-Request in an already opened state
+      // which, if we follow the state machine to the T, should restart the negotiation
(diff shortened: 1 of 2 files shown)


avtolstoy:
I found some edge cases, so going to add a couple more things.

sergeuz (approved):
Look good :+1: I can't test this PR unfortunately.

keeramis (approved):

technobly (approved):
Added [u-blox] tag to the title, since we are calling out a bug in a u-blox modem.  Looks good 👍

PR #2209, opened 2020-10-05
[Wiring] Remove warning for publish scope deprecation

keeramis:
### Problem

Deprecation notice exists on Particle.publish without a scope, although there is no need for a scope now.

### Solution

Remove the deprecation notice on Particle.publish without a scope, since there is only one event stream. 

### Example App

Compile the following code, and you should **no longer** see warning of deprecation notice like this
```
warning: 'particle::Future<bool> CloudClass::publish(const char*)' is deprecated: Beginning with 0.8.0 release, Particle.publish() will require event scope to be specified explicitly. Define PARTICLE_USING_DEPRECATED_API macro to avoid this warning. [-Wdeprecated-declarations]
```
or
```
warning: 'bool CloudClass::subscribe(const char*, EventHandler)' is deprecated: Beginning with 0.8.0 release, Particle.subscribe() will require event scope to be specified explicitly. Define PARTICLE_USING_DEPRECATED_API macro to avoid this warning. [-Wdeprecated-declarations]
```

```c
#include "application.h"

void myHandler(const char *event, const char *data)
{
 // code
}

void setup() {
    Particle.publish("msg");
    Particle.subscribe("temp", myHandler);
}

void loop() {
}
```

### References

[ch61262](https://app.clubhouse.io/particle/story/61262/remove-deprecation-notice-for-particle-publish-without-a-scope)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_cloud.h
+++ b/wiring/inc/spark_wiring_cloud.h
@@ -41,9 +41,6 @@
 #include <functional>
 #include <type_traits>
 
-#define PARTICLE_DEPRECATED_API_DEFAULT_SUBSCRIBE_SCOPE \
-        PARTICLE_DEPRECATED_API("Beginning with 0.8.0 release, Particle.subscribe() will require event scope to be specified explicitly.");
-
 typedef std::function<user_function_int_str_t> user_std_function_int_str_t;
 typedef std::function<void (const char*, const char*)> wiring_event_handler_t;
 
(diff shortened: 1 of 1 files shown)


keeramis:
Andrey - Kindly advise that there is no clean up required here - thank you
https://github.com/particle-iot/device-os/blob/develop/wiring/inc/spark_wiring_cloud.h#L565

wiring/inc/spark_wiring_cloud.h
```diff
@@ -41,9 +41,6 @@
 #include <functional>
 #include <type_traits>
 
-#define PARTICLE_DEPRECATED_API_DEFAULT_PUBLISH_SCOPE \
-        PARTICLE_DEPRECATED_API("Beginning with 0.8.0 release, Particle.publish() will require event scope to be specified explicitly.");
-
 #define PARTICLE_DEPRECATED_API_DEFAULT_SUBSCRIBE_SCOPE \
```
avtolstoy, line 44: This should also probably be removed, as I think there is no scope with subscribes either now.

@monkbroc

sergeuz: I think it's fine to remove this deprecation message anyway since it's no longer relevant, i.e. we won't be requiring to specify one of the existing scopes explicitly (`MY_DEVICES`, `ALL_DEVICES`). As for the public event stream, @elijthomas could you please clarify if we finally disabled it?

sergeuz (approved):

wiring/inc/spark_wiring_cloud.h
sergeuz: According to Eli, the public stream is disabled already.

keeramis: Thank you for checking, Sergey. In that case, I believe we can merge this PR.

sergeuz: Sounds good. I've clarified this with Eli just now, and it basically doesn't matter whether you're using `MY_DEVICES` or `ALL_DEVICES` with your subscriptions and, accordingly, `PUBLIC` or `PRIVATE` with published events. Let's keep in mind that we also need to update the reference docs accordingly. cc @rickkas7

keeramis: Got it. thanks!...

PR #2210, opened 2020-10-06
[Gen3] Fix LED behavior in case of network failure in Cloud Connecting state 

keeramis:
### Problem

When cloud _connecting_, the RGB LED doesn't go back to (blinking) green when de-registered from network, or when having DNS test failures

### Solution

The cloud signaling procedure is correct, however, the LED doesn't reflect the same. The LED stays in blinking cyan, although the system state has backed off from "Cloud Connecting" state.

### Steps to Test

- Run any app that allows connecting to cloud, with serial logging enabled helpfully
- As soon as "Cloud connecting" is seen (device must have started blinking cyan), disconnect the module from network (probably by removing antenna). This needs to be done after "Cloud: Connecting" and before "Cloud socket connected" states
- The device struggles at that point to connect, and likely gets a `CEREG: x` deregistered URC
- The device should be blinking green (instead of blinking cyan)

### Example App

```c
void setup() {
}

void loop() {
}
```

### References

[ch64095](https://app.clubhouse.io/particle/story/64095/boron-lte-device-keeps-blinking-cyan-while-it-should-blink-green-with-cereg-2-urc)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_task.cpp
+++ b/system/src/system_task.cpp
@@ -325,6 +325,10 @@ void establish_cloud_connection()
             // the last error diagnostic is used only for communication errors since we cannot mix
             // HAL and communication error codes without specifying the category of an error
             WARN("Cloud socket connection failed: %d", connect_result);
+            if (connect_result == SYSTEM_ERROR_NETWORK) {
+                LED_SIGNAL_STOP(CLOUD_HANDSHAKE);
+                LED_SIGNAL_STOP(CLOUD_CONNECTING);
+            }
             SPARK_CLOUD_SOCKETED = 0;
 
             diag->status(CloudDiagnostics::DISCONNECTED);


system/src/system_task.cpp
```diff
@@ -325,6 +325,10 @@ void establish_cloud_connection()
             // the last error diagnostic is used only for communication errors since we cannot mix
             // HAL and communication error codes without specifying the category of an error
             WARN("Cloud socket connection failed: %d", connect_result);
+            if (connect_result == SYSTEM_ERROR_NETWORK) {
+                LED_SIGNAL_STOP(CLOUD_HANDSHAKE);
+                LED_SIGNAL_STOP(CLOUD_CONNECTING);
+            }
```
keeramis, line 331: A couple of comments: 

- Choosing to change LED behavior only with `SYSTEM_ERROR_NETWORK` because the other errors doesn't necessarily lead to network failure. As per my understanding, network failure should be the one case that should get the LED back to green.

- The LED behavior also largely depends on these states here such as these : https://github.com/particle-iot/device-os/blob/70f4f9a157a83795e86bd644ce97a50c43629ac3/system/src/system_network_manager.cpp#L402

    There is an **edge** case where the device enters "Cloud Connecting" state, and fails DNS test, but it has not yet gotten a deregistration URC, in which case it is in `IP_CONFIGURED` state so the above code makes it _breathe_ green, which is rightfully the correct state because it is in `IP_CONFIGURED` state. Once it gets the deregistration URC, it blinks green as expected. Please see this log as well. Just highlighting an edge case for reference (and it looks OK from my understanding).

keeramis: [ch64095.txt](https://github.com/particle-iot/device-os/files/5331002/ch64095.txt)

sergeuz: IIRC, this is issue is specific to Gen 3 platforms. On Gen 2, the system gets notified that it's no longer connected to the network and proceeds to update the LED (as well as the system state) accordingly via a call to `cloud_disconnect()`: https://github.com/particle-iot/device-os/blob/develop/system/src/system_network_internal.h#L677-L680.

Perhaps, rather than providing a workaround for just one specific case, where creating the cloud socket or "connecting" it to the cloud fails (note that UDP sockets are connection-less), we should investigate why the system doesn't update the cloud connection state accordingly when it's no longer connected to the network? I think we have a bigger problem here rather than just LED indication, and it probably dates back to mesh times when we for some reason couldn't determine reliably which network interface among the available ones was used for the cloud connection.

cc @avtolstoy

keeramis: @sergeuz 
> investigate why the system doesn't update the cloud connection state accordingly

Which state should the system update the cloud connectivity status to? Probably go back to "cloud disconnect" and restart the connection since we see a `+CEREG: 2` URC?

keeramis: `+CEREG: 2` cellular deregistration* URC.

sergeuz: I discussed this PR with Andrey, and whereas it's possible that we're not cancelling the current ongoing connection attempt on network disconnection events in certain cases, trying to investigate and address that in the scope of this PR would be too much. The change introduced in this PR is perfectly fine: the `establish_cloud_connection()` function starts the LED indication and it should perform a cleanup and stop the indication if it wasn't able to establish a connection. I'm going to approve this PR :+1:

sergeuz (approved):

system/src/system_task.cpp
keeramis: okie thanks Sergey. Please let me know if we need to open a separate story for the broader correction.

PR #2211, opened 2020-10-08
[gen3] Allow UMNOPROF SIM_SELECT to default back to SW_DEFAULT

technobly:
### Problem

3rd party SIMs that are not detected with a UMNOPROF, do not default to the SW_DEFAULT profile (0), but get stuck in an init loop in Gen 3.

On Gen 2, it will default to SW_DEFAULT profile, but it will take 8 tries before it does taking up unnecessary time.

### Solution

- [gen2] A repeated UMNOPROF SIM_SELECT allows MDMParser::init to continue early [ch64292]
- [gen3] Allow UMNOPROF SIM_SELECT to default back to SW_DEFAULT [ch64292]

### Steps to Test

- Run the example app
- Let the device connect to the cloud with a Particle SIM
- Triple tap the Mode button and wait for "Cold boot for a Factory Boot Experience!"
- Power off the device
- Switch the SIM for a 3rd party SIM
- Power on the device and it should connect quickly using the SW_DEFAULT profile.

NOTE: You will see a log that says the following if this happens:
> UMNOPROF=1 did not resolve a built-in profile, please check if UMNOPROF=100 is required!

This is up to the user to research and implement.  YMMV on the exact configuration needed.  Please contact your SIM provider.

### Example App

```c
#include "Particle.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
bool reset_umnoprof = false;
{
        if (clicks == 3) {
            reset_umnoprof = true;
        }
    }
}
void setup() {
    // Cellular.setActiveSim(EXTERNAL_SIM); // you might need this
    Particle.connect();
}

void loop() {
    if (reset_umnoprof) {
        reset_umnoprof = false;

        // System.reset();
        Cellular.command(60000, "AT+CFUN=0\r\n");
        Cellular.command(1000, "AT+UMNOPROF=0\r\n");
        Cellular.command(1000, "AT+UMNOPROF?\r\n");
        Cellular.command(1000, "AT+CFUN=15\r\n");
    }
}

```

### References

ch64292

---

### Completeness

- [x] N/A Added documentation

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -1060,11 +1060,11 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
         netConf_ = networkConfigForIccid(buf, lenCcid);
         do {
             // Set UMNOPROF = SIM_SELECT
-            auto resp = parser_.sendCommand("AT+UMNOPROF?");
+            auto respUmnoprof = parser_.sendCommand("AT+UMNOPROF?");
             reset = false;
             int umnoprof = static_cast<int>(UbloxSaraUmnoprof::NONE);
-            auto r = CHECK_PARSER(resp.scanf("+UMNOPROF: %d", &umnoprof));
-            CHECK_PARSER_OK(resp.readResult());
+            auto r = CHECK_PARSER(respUmnoprof.scanf("+UMNOPROF: %d", &umnoprof));
+            CHECK_PARSER_OK(respUmnoprof.readResult());
             if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::SW_DEFAULT) {
                 // Disconnect before making changes to the UMNOPROF
                 auto respCfun = parser_.sendCommand(UBLOX_CFUN_TIMEOUT, "AT+CFUN?");
@@ -1076,9 +1076,16 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                 }
 
                 // This is a persistent setting
-                int umnoprof = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
+                umnoprof = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
                 if (netConf_.netProv() == CellularNetworkProvider::TWILIO) { // if Twilio Super SIM
                     umnoprof = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                } else {
+                    // break out of do-while if we're trying to set SIM_SELECT a second time
+                    if (resetCount >= 1) {
+                        LOG(WARN, "UMNOPROF=1 did not resolve a built-in profile, please check if UMNOPROF=100 is required!");
+                        resetCount = 0;
+                        break;
+                    }
                 }
                 parser_.execCommand(1000, "AT+UMNOPROF=%d", umnoprof);
                 // Not checking for error since we will reset either way
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1083,12 +1083,21 @@ bool MDMParser::init(DevStatus* status)
                 }
             }
             int p = UBLOX_SARA_UMNOPROF_SIM_SELECT;
+            bool continueInit = false;
             if (netProv == CELLULAR_NETPROV_TWILIO) {
                 p = UBLOX_SARA_UMNOPROF_STANDARD_EUROPE;
+            } else {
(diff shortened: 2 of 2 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1076,9 +1076,16 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                 }
 
                 // This is a persistent setting
-                int umnoprof = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
+                umnoprof = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
                 if (netConf_.netProv() == CellularNetworkProvider::TWILIO) { // if Twilio Super SIM
                     umnoprof = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                } else {
+                    // break out of do-while if we're trying to set SIM_SELECT a second time
+                    if (resetCount >= 1) {
```
sergeuz, line 1084: Based on the PR description, shouldn't this workaround only work when the external SIM is used? Or is it not that important to have a correct profile selected for officially supported providers other than Twilio?

sergeuz (approved):

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Currently the only other SIM we only support on R410 is Kore AT&T, which does get picked up via the SIM_SELECT as AT&T profile.  SIM_SELECT is used as a way to somewhat provide 3rd party SIM support if it works... if not... then it will default now to SW_DEFAULT 0 profile.

sergeuz: In other words, it's ok if the client is unable to set SIM_SELECT for Kore on the first try, correct?

technobly: Should we maybe make it try at least twice?

sergeuz: I would stick to the previous behavior as much as possible while still addressing the very specific issue the customer is having. If I'm not mistaken 2.0.0 is the first release that introduces this reset-and-retry loop when setting the operator profile, and, most importantly, it also fails the connection attempt if it wasn't able to set the profile after a number of attempts, which is exactly what is causing the issue.

I'm assuming that there's a good reason why we introduced this loop, so what if after making a few attempts to set the profile we simply ignore the fact that the profile wasn't set correctly and proceed with the initialization just like previous Device OS versions did?

technobly: Yes that's what it's been changed to do now, and to be more like Gen 2 was previously (before 2.0.0) by trying a few times in case of certain errors.  But really this retry loop is in case we set UMNOPROF=x and it keeps ending up in SW_DEFAULT. I'm seeing a couple other cases that need to be resolved automatically as well that I will fix in the next push. Regarding this thread though, I think I will bump the `if (resetCount >= 2) {` to put more emphasis on doing the right thing ™️  for official SIMs, and less emphasis on speeding unofficial SIMs to default to SW_DEFAULT.

sergeuz: The current code is more like
- using Twilio SIM: fail the initialization if unable to set the profile
- using any other SIM including Kore: try setting the profile but ignore the error if it fails

Is there a reason to treat Twilio and Kore differently?

technobly: That's the "bug" part of this fix :D

sergeuz: I would be fine with it, since I'm almost certainly missing some context here :sweat_smile: I just wanted to make sure this is exactly the behavior we wanted

technobly: I'm going to merge this as it worked mostly ok as-is, but with the other UMNOPROF=100 PR and these changes together I'm noticing some more case that need to be addressed.  So we can pick up the review over there :D

sergeuz: Sure. I have an ask: if this is going to land in 2.0.0 as it is implemented now, let's add a comment that would state that it's an intended behavior, as that's not obvious from the code (hence why it caught my eye).

technobly: It will change, but I have two PR's open for similar parts of the code which is making life difficult.  Since that other one has more changes... I'll merge this one, rebase that one and make the changes.  I'll try to make it very obvious what the intended behavior is with comments.

sergeuz: Cool. Thank you :pray:

PR #2212, opened 2020-10-08
[Cellular/Wiring] Remove rssi and qual from Cellular.RSSI()

keeramis:
### Description

Remove `rssi` and `qual` from `Cellular.RSSI()`

Additionally, 
- Printing `Cellular.RSSI()` will now print signalvalue,qualityvalue
- New API call - `Cellular.getSignal()` as an alias for `Cellular.RSSI()`.

### Example App

An app checking for `Cellular.RSSI().rssi` or `Cellular.RSSI().qual` should throw errors.

```c
void setup() {
}

void loop() {
}
```

### References

[ch62644](https://app.clubhouse.io/particle/story/62644/cellular-rssi-rssi-and-qual-should-be-removed-in-device-os-v3-0-0)

---

### Completeness

- [ ] Added documentation


hal/inc/net_hal.h
```diff
@@ -47,8 +47,8 @@ typedef enum {
     NET_ACCESS_TECHNOLOGY_UTRAN = 4,
     NET_ACCESS_TECHNOLOGY_WCDMA = 4,
     NET_ACCESS_TECHNOLOGY_CDMA = 5,
-    NET_ACCESS_TECHNOLOGY_LTE = 6,
-    NET_ACCESS_TECHNOLOGY_IEEE802154 = 7,
+    NET_ACCESS_TECHNOLOGY_IEEE802154 = 6,
+    NET_ACCESS_TECHNOLOGY_LTE = 7,
```
sergeuz, line 51: What's the reason for this change? It breaks the ABI.

sergeuz (changes requested):

hal/inc/net_hal.h
keeramis: This unit test never passes when the RAT=7. 
https://github.com/particle-iot/device-os/blob/1f2d369d22d8af6f9ef25da3a3b0d59df5322ae9/user/tests/wiring/no_fixture/cellular.cpp#L155
Because it looks like it's getting mapped to `NET_ACCESS_TECHNOLOGY_IEEE802154` and never passes the unit test although the values are obtained.

sergeuz: Do you know why? If that's the case then it has to be fixed either in the unit test or in the system code (if, for example, it exposes raw RAT identifiers obtained from the modem to the app code), but, again, we can't change values of an enum that is a part of a dynalib interface

keeramis: Thanks Sergey. Andrey walked me through some of the ABI stuff as well. 

I believe here's why:
The `s.getAccessTechnolog()` here (https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/user/tests/wiring/no_fixture/cellular.cpp#L122) gets the access technology from `sig_.rat` here (https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/wiring/src/spark_wiring_cellular_printable.cpp#L44) where `rat` here is `uint8_t` and is a directly obtained value from the cell radio - https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/hal/inc/cellular_hal_constants.h#L87
And we are directly converting the `uint8_t` to `hal_net_access_tech_t` for unit tests. 

This is not affecting the device-os `Cellular.RSSI()` itself because the `getStrengthValue` for example doesn't check for specific rat. But it affects the unit tests esp. only when RAT=7 on LTE radios (which is for LTE). There are two cases where RAT is `7` - always for LTE Cat-1 , and sometimes for LTE-M1 (ublox radios are strange).

I understand that changing the enum the way I did will break the ABI so I am going to revert this change, but I think this needs to be addressed for the unit tests to pass, if my understanding is correct. Will probably make a separate story/PR (probably having an intermediate translation between `getAccessTechnology` and `hal_net_access_tech_t`

sergeuz: > where rat here is uint8_t and is a directly obtained value from the cell radio

Could you point me to where you think this might be happening in the code? Both [Gen 2](https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/cellular_internal.cpp#L32-L54) and [Gen 3](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp/cellular/cellular_hal.cpp#L357) seem to be converting the modem's RAT identifiers to `hal_net_access_tech_t` rather than using them as is.

keeramis: The [Gen2](https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/cellular_internal.cpp#L32-L54) converts into `hal_net_access_tech_t` but [here](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/hal/src/electron/cellular_internal.cpp#L55) uses this [Act enum](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/hal/shared/cellular_enums_hal.h#L82-L91) from `NetStatus` to make calculations. 
On [Gen3](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp/cellular/cellular_hal.cpp#L357), it uses `s.accessTechnolog()` which there gets values using `CellularAccessTechnolog` [enum](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/hal/network/ncp/cellular/cellular_ncp_client.h#L69-L81). Although they are both converting RAT to `hal_net_access_tech_t`, they are using the other enums for calculations so we don't see any problem there. 

The problem seems to occur in unit tests which I put in [this](https://github.com/particle-iot/device-os/pull/2212/files#r504345008) comment. Hope that helps...

sergeuz: Sorry, maybe I'm missing something, but I still don't understand how an incorrect RAT value (which the change that is being discussed in this thread was supposed to fix) gets returned to the application. Gen 2 uses `AcT` in calculations, that's correct, but it also sets `cellular_signal_t::rat` based on the same value of the `Act` enum. In other words, in case of LTE, the cellular HAL sets `cellular_signal_t::rat` to `NET_ACCESS_TECHNOLOGY_LTE` [here](https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/cellular_internal.cpp#L43) and performs calculations specific to LTE [here](https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/cellular_internal.cpp#L100-L114). Where's the error?

keeramis: The error is while running this unit test. Hope this provides more clarity.... 
While running [this](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/user/tests/wiring/no_fixture/cellular.cpp#L112) unit test on a Boron LTE device when the RAT from the device is `7`:
- Gets the RAT [here](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/user/tests/wiring/no_fixture/cellular.cpp#L122) from s.getAccessTechnoloy()
- This is [getAccessTechnolog()](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/wiring/src/spark_wiring_cellular_printable.cpp#L44) - it reads the `sig_.rat`
- `rat` is `uint8_t` when stored is `7`, so `s.getAccessTechnolog()` gives `7`.
- Back to unit test - in the switch case [here](https://github.com/particle-iot/device-os/blob/6edbe275779e38cb68b8135f876ac1c1248d5758/user/tests/wiring/no_fixture/cellular.cpp#L122), the result of `s.getAccessTechnolog()` is `7`, it looks for `7` , doesn’t see it because `NET_ACCESS_TECHNOLOGY_LTE` is actually `6`

sergeuz: > rat is uint8_t when stored is 7, so s.getAccessTechnolog() gives 7

That's the part I don't get. How 7 (`NET_ACCESS_TECHNOLOGY_IEEE802154`) ends up being stored in `rat` if we're not exposing raw RAT values reported by the modem. On Gen 3, we always convert raw values to `hal_net_access_tech_t` [here](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp/cellular/cellular_hal.cpp#L357) using [this function](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp/cellular/cellular_hal.cpp#L85-L106). As you can see that function never returns `NET_ACCESS_TECHNOLOGY_IEEE802154`.

keeramis: Thank you for helping check this, Sergey. Reverting the change back.

hal/inc/cellular_hal.h
```diff
@@ -159,7 +159,7 @@ void cellular_cancel(bool cancel, bool calledFromISR, void* reserved);
 /**
  * Retrieve cellular signal strength info
  */
-cellular_result_t cellular_signal(CellularSignalHal* signal, cellular_signal_t* reserved);
+cellular_result_t cellular_signal(cellular_signal_t* reserved);
```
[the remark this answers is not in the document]
keeramis: Fixed..

sergeuz (approved):

wiring/src/spark_wiring_cellular.cpp
```diff
@@ -26,30 +26,18 @@
...
-        if (cellular_signal(&sig_hal, &sigext) != 0) {
-            sig.rssi = 1;
+        if (cellular_signal(nullptr, &sigext) != 0) {
             return sig;
         }
-        sig.rssi = sig_hal.rssi;
-        sig.qual = sig_hal.qual;
-        if (sig.rssi == 0) {
-            sig.rssi = 2;
-        }
```
technobly, line 49: I want to make sure we don't get out of sync with our documentation here.  sig.rssi = 2 and sig.rssi = 1 are special error cases.  If we are deprecating those errors as well, we should be sure to update the docs.  Otherwise perhaps we need to keep the errors?  I can't seem to find anything about Cellular.RSSI() errors in our Docs anymore... but that could be it's own issue.  So I'll just ask this question, if the API to request signal strength/quality fails, how does the user know there is an error?

technobly (changes requested):

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -93,21 +93,6 @@ test(CELLULAR_03_resolve) {
...
-        if (s.rssi < 0) {
-            break;
-        }
-        Serial.println(s);
-        delay(5000);
-    }
-    assertLessOrEqual(s.rssi, -20);
-    assertMoreOrEqual(s.rssi, -150);
-}
-
```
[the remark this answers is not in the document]
avtolstoy: https://github.com/particle-iot/device-os/blob/2ec804dd97b36a480272bae45abf6fef36e8ae6e/user/tests/wiring/no_fixture/cellular.cpp#L97

wiring/src/spark_wiring_cellular.cpp
keeramis: Part of documentation already says this but will need to update docs as needed. 
```
Before Device OS 0.8.0, the CellularSignal class only had two member variables, rssi, and qual. These will be removed in a future version of Device OS and you should use getStrengthValue() and getQualityValue() instead.
``` 

I don't think it's a good idea to keep these errors as it could confuse the users even more. Currently, when the signal strength/quality fails we should get `0` on the `getStrengthValue` and `getQualityValue` and a `-1` on `getStrength` and `getQuality`.

sergeuz: Good point, Brett. I suggest we add `operator bool` to `CellularSignal`.

avtolstoy: The problem is that we need to decide on the behavior of that operator:

1. RAT is unknown - a clear failure
2. RAT is valid, strength/quality are invalid/unavailable (temporarily unavailable as modem needs some time to calculate them for example) - ?
3. RAT is valid, strength is valid, quality is invalid/unavailable (isn't supported on some modems with certain RATs for example, or again, the modem needs some time to calculate it) - ?

user/tests/wiring/no_fixture/cellular.cpp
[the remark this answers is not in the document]
technobly: Ahh, good :D Thanks for pointing that out.

wiring/src/spark_wiring_cellular.cpp
technobly: I think it's very possible to have a RAT, but have a missing RSRP/RSRQ value... which results in 0 or -1 in the various APIs.  If we can only use an operator bool like we do in other APIs, perhaps if any error is present this returns false.  This would exclude when quality is not supported.  If we'd like to know exactly what element failed or is not valid, perhaps the user could then turn to individual tests on each one (!= -1 or != 0)... and we'd need to make sure these are all documented (currently I don't see these documented).

sergeuz: Yeah, IMO, that operator should return `true` if the object as a whole is valid, i.e. contains RAT, signal strength and quality, if the latter is applicable. Alternatively, instead of the operator, we can have separate methods like `isRatValid()`, `isStrengthValid()`, etc.

avtolstoy: For consistency, this should probably be `operator bool()` + `isValid()` as with other wiring APIs. Forwarding info about unsupported quality from HAL should be doable as `cellular_signal_t` is extendable.

sergeuz: Then, perhaps, let's give the users full visibility into the info available:

`isAccessTechnologyValid()`
`isStrengthValid()`
`isQualityValid()`
`isQualitySupported()`

?

keeramis: Similar to rssi error codes before, If we can provide one API which returns an `int` and the users can check a look up table?
- `0` - No error
- `1` - Invalid RAT
- `2` - Invalid Strength but valid quality
- `3` - Valid strength but invalid quality
- `4` - InValid strength and quality

avtolstoy: Distinct named methods are the way if we are to add granular checks, because at the end of the day what we are interested in is whether a particular piece of info is valid/available.

technobly: Sounds like an agreement on the bool operator.  I'm wondering what the cost is of adding all of the distinct methods for validity checks?  Once we add them, it would take a while to deprecate them if need be for space or something.  Should we also add them at this time, and/or just document the validity check for now?

sergeuz: Let's also ask @jlovinger-particle as he probably has a better perspective as to which approach would be more convenient/useful from the UX point of view. I'd be fine with any approach as long as we're not forcing our users to deal with special values – API deprecation is a good opportunity to make improvements like this.

wiring/inc/spark_wiring_cellular_printable.h
```diff
@@ -58,6 +51,13 @@ class CellularSignal : public particle::Signal, public Printable {
 
     virtual size_t printTo(Print& p) const;
 
+    virtual bool isAccessTechnologyValid() const;
+    virtual bool isSignalStrengthValid() const;
+    virtual bool isSignalQualityValid() const;
+    virtual bool isSignalQualitySupported() const;
```
sergeuz, line 57: This class is called `CellularSignal` so there's no need to have the word `Signal` in its method names. The existing getter methods for the strength and quality values don't have `Signal` in their names as well.

wiring/inc/spark_wiring_cellular_printable.h
```diff
@@ -58,6 +51,13 @@ class CellularSignal : public particle::Signal, public Printable {
...
+    virtual bool isValid() const;
+    virtual operator bool() const;
```
sergeuz, line 59: No strong opinion here, but I'd prefer to have either `isValid()` or a bunch of methods like `isStrengthValid()`, etc, but not both.

wiring/src/spark_wiring_cellular_printable.cpp
```diff
@@ -80,18 +77,49 @@ float CellularSignal::getQualityValue() const
...
-#pragma GCC diagnostic pop
+
+bool CellularSignal::isAccessTechnologyValid() const
+{
+    return (sig_.rat != NET_ACCESS_TECHNOLOGY_NONE);
+}
+
+bool CellularSignal::isSignalStrengthValid() const
+{
+    return (sig_.rssi != std::numeric_limits<int32_t>::min());
```
sergeuz, line 96: `Cellular.RSSI()` [returns](https://github.com/particle-iot/device-os/blob/ch62644/remove_celluluar_rssi/wiring/src/spark_wiring_cellular.cpp#L31) a default-initialized `CellularSignal` object on certain errors, which has `rssi` and `qual` set to 0. `isSignalStrengthValid()` and `isSignalQualityValid()` called on such an object will incorrectly return `true`.

wiring/src/spark_wiring_cellular_printable.cpp
```diff
@@ -80,18 +77,49 @@ float CellularSignal::getQualityValue() const
...
+
+bool CellularSignal::isSignalQualityValid() const
+{
+    return (sig_.qual != std::numeric_limits<int32_t>::min());
+}
+
+bool CellularSignal::isSignalQualitySupported() const
+{
+#if (PLATFORM_ID == PLATFORM_ELECTRON_PRODUCTION)
+    return (sig_.rat == NET_ACCESS_TECHNOLOGY_GSM) ? false : true;
```
sergeuz, line 107: How accurate is this? The cellular HAL seems to report BER as the signal quality in case of GSM.

wiring/src/spark_wiring_print.cpp
```diff
@@ -170,10 +170,11 @@ size_t Print::printFloat(double number, uint8_t digits)
...
 
-  if (isnan(number)) return print("nan");
-  if (isinf(number)) return print("inf");
-  if (number > 4294967040.0) return print ("ovf");  // constant determined empirically
-  if (number <-4294967040.0) return print ("ovf");  // constant determined empirically
+  // FIXME: Trade off to save memory
+  //if (isnan(number)) return print("nan");
+  //if (isinf(number)) return print("inf");
+  //if (number > 4294967040.0) return print ("ovf");  // constant determined empirically
+  //if (number <-4294967040.0) return print ("ovf");  // constant determined empirically
```
sergeuz, line 177: It's a good question why we use a custom implementation of the double to string conversion here, however, let's keep this code intact for now, and fix the flash overflow in the scope of another PR.

wiring/src/spark_wiring_cellular.cpp
```diff
@@ -26,30 +26,18 @@
 namespace spark {
 
     CellularSignal CellularClass::RSSI() {
```
sergeuz, line 28: RSSI no longer represents the info returned by this method. How about adding an alias for it, e.g. `Cellular.getSignal()`?

sergeuz (changes requested):

wiring/src/spark_wiring_cellular_printable.cpp
```diff
@@ -80,18 +77,30 @@ float CellularSignal::getQualityValue() const
...
     return n;
 }
-#pragma GCC diagnostic pop
+
+bool CellularSignal::isValid() const
+{
+    return (sig_.rat != NET_ACCESS_TECHNOLOGY_NONE &&
+            sig_.rssi != std::numeric_limits<int32_t>::min() &&
+#if (PLATFORM_ID == PLATFORM_ELECTRON_PRODUCTION)
+            (sig_.rat == NET_ACCESS_TECHNOLOGY_GSM) ? true : (sig_.qual != std::numeric_limits<int32_t>::min()));
```
sergeuz, line 94: It would be more straightforward to write this as `(sig_.rat == NET_ACCESS_TECHNOLOGY_GSM || sig_.qual != std::numeric_limits<int32_t>::min())`.

Please add a comment that ublox modems do not support/correctly report signal quality in the case of GSM (if that's why we have this check here).

Also note that in the current code, an object for which `isValid()` returns `true` can hold an invalid signal quality value and there's no way for an app to detect that without relying on magic constants.

sergeuz (approved):

wiring/inc/spark_wiring_cellular_printable.h
technobly: It looks like isValid() was left and the individual methods were removed.  Since we also have the bool return which is basically a duplication of isValid(), would it be more useful to users to have the individual isX() methods?  Especially when the bool return is false, they will have a built in way to determine what has gone wrong.  This can also be very useful for internal testing.

wiring/src/spark_wiring_cellular_printable.cpp
technobly: It seems like we have an issue here with the API where strength/quality values that error are set to 0 instead of -1.  Of course -1 is a valid value for strength/quality so it makes sense.  This is why the previous RSSI().rssi error values were positive values 1 and 2.  We still use this `sig_.qual != std::numeric_limits<int32_t>::min())` in isValid().  Perhaps we need to introduce a new error value for these APIs so that they can be differentiated between default-initialized, valid, and error?

avtolstoy: We touched this topic in one of the comment threads here. Let's try not to introduce any special error values for user facing APIs where possible.

There also shouldn't be any issue as of right either, as one of the requirements for `CellularSignal` or a to be valid is a valid RAT. If we do introduce distinct methods to check if a particular part is valid, we can then make sure that underlying `cellular_signal_t` is initialized by default to invalid values.

technobly: Sounds good, I agree.  I think if we did bring back the isX() methods, we could protect them with `sig_.rat != NET_ACCESS_TECHNOLOGY_NONE &&` like all other APIs use.

wiring/inc/spark_wiring_cellular_printable.h
technobly: @avtolstoy @sergeuz @keeramis any opinions on if we should bring back the isX() methods or leave them out?  My reasoning on adding them back in is users would have to add special code that's not really going to be well documented IMO, or if we documented it it would be dense at best.  Or we'd have to document an exact example for validation code. For example, they would have to know they should qualify the value with `sig_.rat != NET_ACCESS_TECHNOLOGY_NONE && value !=>< x` after looking up what the default/error values are for each API.

avtolstoy (approved):
Looks okay. If I had to make a minor comment is that https://github.com/particle-iot/device-os/pull/2212/files#diff-e3cf88d4bb00cc447ee595431d5ba61076b397be131a25a1f0cb63747d4702d9R94 should probably be hidden somewhere within a HAL

wiring/inc/spark_wiring_cellular_printable.h
avtolstoy: No strong opinions, but I'd probably try to keep things simpler right now and if needed when we get feedback later introduce distinct methods.

However, in general the whole object should be valid and there shouldn't be cases where a particular part of it is invalid. Understandably we do have a quirk on one of the platforms with a particular modem type (which is also documented https://docs.particle.io/reference/device-os/firmware/boron/#getqualityvalue-), but taking all the other platforms into consideration a simpler approach seems more desirable. Thoughts?

technobly (approved):

PR #2213, opened 2020-10-09
[u-blox] older modem firmware does not have UMNOPROF=100 available [ch64670]

technobly:
### Problem

- R410 - Gets stuck in loop where UMNOPROF=100 was not sticking, but erroring out instead, on older modem firmware
- R410 - Mostly an issue for Gen 3 with external SIM support on Boron/B SoM, switching between Kore AT&T vs. Twilio Super vs. 3rd Party SIMs could get stuck on the wrong UMNOPROF.
- Gen 3 Modem version detection was running too late for where it needed to be used, and could potentially return an invalid version number instead of 0.

### Solution

- Allow modem to continue with SW_DEFAULT (UMNOPROF=0) when older modem firmware detected.
- Allow UMNOPROF to be switched accordingly between Kore AT&T vs. Twilio Super vs. 3rd Party SIMs.
- Run Gen 3 modem detection earlier, and use temporary variables to ensure if an error occurs, we don't return the error code as the firmware version.

### Steps to Test

- Make sure your R410 modem has older modem firmware < 05.08
- Run the example app
- Let the device connect to the cloud with a Particle SIM
- Triple tap the Mode button and wait for "Cold boot for a Factory Boot Experience!"
- Power off the device
- Make sure the SIM is a Twilio Super SIM
- Power on the device and it should connect quickly using the SW_DEFAULT profile.
- I also ran through the following Test Matrix, but it is not required to do the same as it will take a long time (spot check if you wish):

```
Boron Old Modem Firmware
- 3rd party SIM tries to connect: OK
- Twilio: OK
- Kore AT&T: OK
- Factory Reset Kore AT&T: OK
- Twilio Again: OK
- Factory Reset Twilio: OK

Boron New Modem Firmware
- 3rd party SIM tries to connect: OK
- Twilio: OK
- Kore AT&T: OK
- Factory Reset Kore AT&T: OK
- Twilio Again: OK
- Factory Reset Twilio: OK

Electron LTE Old Modem Firmware
- 3rd party SIM tries to connect: OK
- Twilio: OK
- Kore AT&T: OK
- Factory Reset Kore AT&T: OK
- Twilio Again: OK
- Factory Reset Twilio: OK

Electron LTE New Modem Firmware
- Kore AT&T: OK
- Factory Reset Kore AT&T: OK
```

### Example App

```c
#include "Particle.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
bool reset_umnoprof = false;
{
        if (clicks == 3) {
            reset_umnoprof = true;
        }
    }
}
void setup() {
    // Cellular.setActiveSim(EXTERNAL_SIM); // you might need this
    Particle.connect();
}

void loop() {
    if (reset_umnoprof) {
        reset_umnoprof = false;

        // System.reset();
        Cellular.command(60000, "AT+CFUN=0\r\n");
        Cellular.command(1000, "AT+UMNOPROF=0\r\n");
        Cellular.command(1000, "AT+UMNOPROF?\r\n");
        Cellular.command(1000, "AT+CFUN=15\r\n");
    }
}
```

### References

---

### Completeness

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -95,6 +95,7 @@ const auto UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_R4 = 460800;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_MEMORY_LEAK_ISSUE = 200;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MIN = 200;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MAX = 203;
+const auto UBLOX_NCP_R4_APP_FW_VERSION_LATEST_02B_01 = 204;
 
 const auto UBLOX_NCP_MAX_MUXER_FRAME_SIZE = 1509;
 const auto UBLOX_NCP_KEEPALIVE_PERIOD = 5000; // milliseconds
@@ -169,7 +170,8 @@ int SaraNcpClient::init(const NcpClientConfig& conf) {
     imsiCheckTime_ = 0;
     powerOnTime_ = 0;
     registeredTime_ = 0;
-    memoryIssuePresent_ = false;
+    memoryIssuePresent_ = true; // default to safe state until we determine modem firmware version
+    oldFirmwarePresent_ = true; // default to safe state until we determine modem firmware version
     parserError_ = 0;
     ready_ = false;
     registrationTimeout_ = REGISTRATION_TIMEOUT;
@@ -1052,20 +1054,70 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
 
     if (ncpId() == PLATFORM_NCP_SARA_R410) {
         int resetCount = 0;
+        // Note: Not failing on AT error on ICCID/IMSI lookup since SIMs have shown strange edge cases
+        // where they error for no reason, and hard resetting the modem or power cycling would not clear it.
+        //
         // Check if we are using a Twilio Super SIM based on ICCID
         char buf[32] = {};
-        auto lenCcid = getIccidImpl(buf, sizeof(buf));
-        CellularNetworkConfig netConfig;
-        CHECK_TRUE(lenCcid > 5, SYSTEM_ERROR_BAD_DATA);
-        netConf_ = networkConfigForIccid(buf, lenCcid);
+        int ccidCount = 0;
         do {
-            // Set UMNOPROF = SIM_SELECT
+            memset(buf, 0, sizeof(buf));
+            auto lenCcid = getIccidImpl(buf, sizeof(buf));
+            if (lenCcid > 5) {
+                netConf_ = networkConfigForIccid(buf, lenCcid);
+                break;
+            }
+        } while (++ccidCount < 2);
+        // If failed above i.e., netConf_ is still not valid, look for network settings based on IMSI
+        if (!netConf_.isValid()) {
+            int imsiCount = 0;
+            do {
+                memset(buf, 0, sizeof(buf));
+                auto resp = parser_.sendCommand(UBLOX_CIMI_TIMEOUT, "AT+CIMI");
+                CHECK_PARSER(resp.readLine(buf, sizeof(buf)));
+                const int r = CHECK_PARSER(resp.readResult());
+                if (r == AtResponse::OK) {
+                    netConf_ = networkConfigForImsi(buf, strlen(buf));
+                    break;
+                }
+            } while (++imsiCount < 2);
+        }
+        do {
+            // Set UMNOPROF and UBANDMASK as appropriate based on SIM
             auto respUmnoprof = parser_.sendCommand("AT+UMNOPROF?");
             reset = false;
-            int umnoprof = static_cast<int>(UbloxSaraUmnoprof::NONE);
-            auto r = CHECK_PARSER(respUmnoprof.scanf("+UMNOPROF: %d", &umnoprof));
+            int curProf = static_cast<int>(UbloxSaraUmnoprof::NONE);
+            auto r = CHECK_PARSER(respUmnoprof.scanf("+UMNOPROF: %d", &curProf));
             CHECK_PARSER_OK(respUmnoprof.readResult());
-            if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::SW_DEFAULT) {
+            // First time setup, or switching between official SIM on wrong profile?
+            if (r == 1 && (static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::SW_DEFAULT ||
+                    (netConf_.netProv() == CellularNetworkProvider::TWILIO &&
+                    static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::STANDARD_EUROPE) ||
+                    (netConf_.netProv() == CellularNetworkProvider::KORE_ATT &&
+                    static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::ATT)) ) {
+                int newProf = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
+                // TWILIO Super SIM
+                if (netConf_.netProv() == CellularNetworkProvider::TWILIO) {
+                    // _oldFirmwarePresent: u-blox firmware 05.06* and 05.07* does not have
+                    // UMNOPROF=100 available. Default to UMNOPROF=0 in that case.
+                    if (oldFirmwarePresent_) {
+                        if (static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::SW_DEFAULT) {
+                            break;
+                        } else {
+                            newProf = static_cast<int>(UbloxSaraUmnoprof::SW_DEFAULT);
+                        }
+                    } else {
+                        newProf = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                    }
+                }
+                // KORE AT&T or 3rd Party SIM
+                else {
+                    // break out of do-while if we're trying to set SIM_SELECT a third time
+                    if (resetCount >= 2) {
+                        LOG(WARN, "UMNOPROF=1 did not resolve a built-in profile, please check if UMNOPROF=100 is required!");
+                        break;
+                    }
+                }
                 // Disconnect before making changes to the UMNOPROF
                 auto respCfun = parser_.sendCommand(UBLOX_CFUN_TIMEOUT, "AT+CFUN?");
                 int cfunVal = -1;
@@ -1076,21 +1128,10 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                 }
 
                 // This is a persistent setting
-                umnoprof = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
-                if (netConf_.netProv() == CellularNetworkProvider::TWILIO) { // if Twilio Super SIM
-                    umnoprof = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
-                } else {
-                    // break out of do-while if we're trying to set SIM_SELECT a second time
-                    if (resetCount >= 1) {
-                        LOG(WARN, "UMNOPROF=1 did not resolve a built-in profile, please check if UMNOPROF=100 is required!");
-                        resetCount = 0;
-                        break;
-                    }
-                }
-                parser_.execCommand(1000, "AT+UMNOPROF=%d", umnoprof);
+                parser_.execCommand(1000, "AT+UMNOPROF=%d", newProf);
                 // Not checking for error since we will reset either way
                 reset = true;
-            } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
+            } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
                 auto respBand = parser_.sendCommand(UBLOX_UBANDMASK_TIMEOUT, "AT+UBANDMASK?");
                 uint64_t ubandUint64 = 0;
                 char ubandStr[24] = {};
@@ -1113,11 +1154,7 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                 HAL_Delay_Milliseconds(10000);
                 CHECK(waitAtResponseFromPowerOn(state));
             }
-        } while (reset && resetCount++ < 4);
-
-        if (resetCount >= 4) {
-            return SYSTEM_ERROR_AT_RESPONSE_UNEXPECTED; // we shouldn't have been resetting this many times
-        }
+        } while (reset && ++resetCount < 4); // Note: Twilio SIMs could take more than 2 tries in some error cases, others <= 2
     }
 
     return SYSTEM_ERROR_NONE;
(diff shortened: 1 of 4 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -261,6 +261,8 @@ MDMParser::MDMParser(void)
     _power_mode = 1; // default power mode is AT+UPSV=1
     _cancel_all_operations = false;
     sms_cb = NULL;
+    _memoryIssuePresent = true;
+    _oldFirmwarePresent = true;
```
sergeuz, line 265: Let's remove the initialization of these member variables from the header file where they're declared – for consistency.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1081,14 +1088,19 @@ bool MDMParser::init(DevStatus* status)
...
-            sendFormated("AT+UMNOPROF=%d\r\n", p);
-            waitFinalResp(nullptr, nullptr, UMNOPROF_TIMEOUT);
-            goto reset_failure; // Not checking for errors above since we will reset either way
+            // _oldFirmwarePresent: u-blox firmware 05.06* and 05.07* does not have
+            // UMNOPROF=100 available. Default to UMNOPROF=0 in that case.
+            if (!(_oldFirmwarePresent && p == UBLOX_SARA_UMNOPROF_STANDARD_EUROPE)) {
+                sendFormated("AT+UMNOPROF=%d\r\n", p);
+                waitFinalResp(nullptr, nullptr, UMNOPROF_TIMEOUT);
+                goto reset_failure; // Not checking for errors above since we will reset either way
+            }
```
sergeuz, line 1103: Wouldn't it be better to avoid setting `CFUN=0` as well if the modem is running an old firmware?

sergeuz (approved):

hal/src/electron/modem/mdm_hal.cpp
technobly: Good call, I'll make that change.

hal/src/electron/modem/mdm_hal.cpp
technobly: Absolutely...

sergeuz:
@technobly Also, shouldn't we implement this check for Gen 3 as well?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1052,20 +1054,71 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
-        CellularNetworkConfig netConfig;
-        CHECK_TRUE(lenCcid > 5, SYSTEM_ERROR_BAD_DATA);
-        netConf_ = networkConfigForIccid(buf, lenCcid);
+        int ccidCount = 0;
         do {
-            // Set UMNOPROF = SIM_SELECT
+            memset(buf, 0, sizeof(buf));
+            auto lenCcid = getIccidImpl(buf, sizeof(buf));
+            if (lenCcid > 5) {
+                netConf_ = networkConfigForIccid(buf, lenCcid);
```
sergeuz, line 1067: I noticed that we're also [overriding](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp_client/sara/sara_ncp_client.cpp#L1485) `netConf_` with the persistently stored (or default) settings in `configureApn()` which is called every time we instruct the client to connect to the network. The same logic with detection of the provider based on the IMSI or ICCID is duplicated there as well, although it only tried once. This duplication clearly needs some refactoring.

Perhaps as a quick fix we can prevent `configureApn()` from overriding `netConf_` if the configuration object passed to the method is empty? cc @avtolstoy

sergeuz (approved):

hal/network/ncp_client/sara/sara_ncp_client.cpp
sergeuz: Alternatively, let's store the provider settings in a local variable in `selectSimCard()` rather than in `netConf_`.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1052,20 +1054,71 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+                            newProf = static_cast<int>(UbloxSaraUmnoprof::SW_DEFAULT);
+                        }
+                    } else {
+                        newProf = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                    }
+                }
+                // KORE AT&T or 3rd Party SIM
+                else {
+                    // break out of do-while if we're trying to set SIM_SELECT a third time
+                    if (resetCount >= 2) {
```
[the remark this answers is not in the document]
technobly: We discussed this one in the other PR that was merged, that we want to error on the side of making sure official SIMs get at least one retry, but 3rd party SIMs can just deal with the extra try even though they likely will end up on default anyway.  In reality, if a 3rd party SIM is going to switch to a supported UMNOPROF (2,3,etc..), having an extra try on SIM_SELECT is good for them as well.

Regarding external though... "external" is different from "3rd party".  I don't consider making decisions about external as a necessary choice, because external SIMs could also be official SIMs.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1052,20 +1054,71 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+            int curProf = static_cast<int>(UbloxSaraUmnoprof::NONE);
+            auto r = CHECK_PARSER(respUmnoprof.scanf("+UMNOPROF: %d", &curProf));
             CHECK_PARSER_OK(respUmnoprof.readResult());
-            if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::SW_DEFAULT) {
+            // First time setup, or switching between official SIM on wrong profile?
+            if (r == 1 && (static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::SW_DEFAULT ||
+                (netConf_.netProv() == CellularNetworkProvider::TWILIO &&
+                static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::STANDARD_EUROPE) ||
+                (netConf_.netProv() == CellularNetworkProvider::KORE_ATT &&
+                static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::ATT)) )
```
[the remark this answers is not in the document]
technobly: Usually, yes.. after a CFUN=15.  But only if the modem detects a matching profile for the SIM provider.

hal/network/ncp_client/sara/sara_ncp_client.cpp
avtolstoy: From what I can see there is no reason in updating `netConf_` here, as it comes into play only when connecting, unless we want to skip detection logic in `configureApn()` if an empty configuration is passed as @sergeuz pointed out. Either solution looks fine :)

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1076,21 +1129,10 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
-                        resetCount = 0;
-                        break;
-                    }
-                }
-                parser_.execCommand(1000, "AT+UMNOPROF=%d", umnoprof);
+                parser_.execCommand(1000, "AT+UMNOPROF=%d", newProf);
                 // Not checking for error since we will reset either way
                 reset = true;
-            } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(umnoprof) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
+            } else if (r == 1 && static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::STANDARD_EUROPE) {
```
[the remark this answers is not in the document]
technobly: Yes that's ok.  We must go through a CFUN=15 reset before we can modify the UBANDMASK.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: It was suggested before to just simplify and use the global variable `netConf_` instead of the local one I had there :D

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1178,16 +1216,22 @@ int SaraNcpClient::getAppFirmwareVersion() {
...
-    return major * 100 + minor;
+    int retMajor = 0, tempMajor = 0;
+    int retMinor = 0, tempMinor = 0;
+    int r = resp.scanf("%*[^,],%*[A.]%d.%d", &tempMajor, &tempMinor);
+    if (r == 2 && (resp.readResult() == AtResponse::OK)) {
+        retMajor = tempMajor;
+        retMinor = tempMinor;
+    }
+    LOG(TRACE, "App firmware: %d", retMajor * 100 + retMinor);
+    return retMajor * 100 + retMinor;
```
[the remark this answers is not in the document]
technobly: Good call on the order. I don't mind adding your ver change as well.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
technobly: Sure why not :D If the external SIM holder is there and a customer wants to switch to a new SIM we offer and not get a new device... seems like a valid option.  We also need this support in our own testing environments.

hal/network/ncp_client/sara/sara_ncp_client.cpp
sergeuz: I have the following suggestion:

- Let store the provider settings determined in `selectSimCard()` to a separate variable, e.g. `simNetConf_`.
- In `configureApn()` let's skip the detection if `conf` (the config passed to the method by the network manager) is empty and `simNetConf_` is not, and reuse `simNetConf_` for APN settings.
- Move the code that detects the provider settings based on IMSI and ICCID to a separate method and use that method in both `selectSimCard()` and `configureApn()` so that there's no code duplication.

I'll leave it up to you guys to decide whether it's something that we can do in this PR. I'm also fine with using a local variable in `selectSimCard()` as a quick fix.

sergeuz: > Move the code that detects the provider settings based on IMSI and ICCID to a separate method and use that method in both selectSimCard() and configureApn() so that there's no code duplication

Actually, we can perform this detection just once in `selectSimCard()`, so there's probably no need to move it to a separate method.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
sergeuz: Then, perhaps, it would be better to break the loop early if we know that it's not an official SIM (i.e. we don't have a provider info for it) and, otherwise, still do all 4 attempts for all official SIMs including Kore?

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: I'd like to keep the code the way it is currently.  I don't see it being duplicated since it's similar, yes... but it fails much differently.  The code in selectSimCard() to detect the SIM provider is designed to not fail completely in case of error, whereas the code in configureApn() is, and rightly so.  The "duplication" is a NIT we can ignore for now IMO.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
sergeuz: I just still don't get this separation: `KORE AT&T or 3rd Party SIM` -> break early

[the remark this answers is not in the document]
technobly: We do not officially support 3rd Party SIMs, so we should not be writing code specifically that enhances them.  If it's good enough for official SIMs, it's good enough 👍

hal/network/ncp_client/sara/sara_ncp_client.cpp
sergeuz: > I don't see it being duplicated since it's similar, yes... but it fails much differently

Given that the detection code is moved to a separate method, it's just a matter of how you're interpreting the return value of that method in `selectSimCard()` or `configureApn()`.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
sergeuz: So 2 attempts is good enough for Kore, but we need 4 for Twilio, is that what you mean? If so, let's be explicit and maybe add a comment?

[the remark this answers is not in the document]
technobly: Twilio will likely never see 4 tries (usually 2 with a new SIM)... it's just there in case some combination of UMNOPROF and UBANDMASK fail.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: I think we should table refactoring configureApn() code for optimization for another PR.  Minimize surface area of fixes for LTS.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
sergeuz: Cool, let's mention this in a comment then :pray:

hal/network/ncp_client/sara/sara_ncp_client.cpp
sergeuz: Sounds good. Please keep in mind that due to `netConf_` being set in multiple places it's generally not safe to rely on it to implement provider-specific workarounds. For example, [this workaround](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp_client/sara/sara_ncp_client.cpp#L1797-L1799) won't work if `netConf_` was overridden in `configureApn()`.

technobly: I don't understand what you mean by "won't work if overridden".  We rely on configureApn() to set the netConf_ properly, so it better work :D

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
technobly: Regarding always reading the final result code "resp.readResult()", is that necessary to clear the parser of responses?

hal/network/ncp_client/sara/sara_ncp_client.cpp
sergeuz: We rely on `configureApn()` to set correct APN settings. The `CellularNetworkConfig` class didn't have a provider ID field previously (`netProv()`) – we added it at some point, but it's set to `NONE` in all objects passed to the NCP client by the network manager.

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
sergeuz: IIRC, the destructor of `AtResponse` clears the context of the current AT command if the final response code wasn't read by the client code, so that the command channel can be reused for other commands.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: When setting the correct APN via networkConfigForIccid()/networkConfigForImsi(), it also now sets the netProv(), so it should be fine, right?

sergeuz: ah, you're right, we're performing the detection once again if the `conf` object is empty in `configureApn()`. I mixed up things, sorry. I guess the only issue then is that we're performing that detection twice, but as you've mentioned already, we can fix that in the scope of another PR :+1:

PR #2215, opened 2020-10-13
[Cellular] Fixes PMIC reducing current on warm boot and causing modem brownouts

avtolstoy:
### Problem

There is a number of problems in the DeviceOS power management subsystem:

1. There is a race condition between DeviceOS and PMIC while in DPDM detection. If DeviceOS modifies 'Input Source Control Register REG00' before DPDM finishes (as it also modifies it), it may be left in an inconsistent state with low input current limit, as updates to the PMIC registers follow the pattern:
    1. Take current register value (which may contain a transient setting)
    2. Change necessary bitfields
    3. Write back
2. When powered by VIN (so USB DP/DM lines are floating), we are still performing DPDM detection on every boot, which will temporarily lower input current limit to 100mA

image

### Solution

1. We are guarding against modifications of 'Input Source Control Register REG00' while DPDM is running to avoid race conditions, or worst case scenario restart DPDM after the modification
2. We aren't going to run DPDM detection on boot if we detect that we are powered by VIN and an appropriate input current limit has already been set (> 100mA)
3. Just in case we also make sure that BATFET is enabled

**NOTE: ignore `HAL_PLATFORM_POWER_MANAGEMENT_PMIC_WATCHDOG`, this is an experimental change that we may enable in 3.x, it's disabled for all the platforms**

### Steps to Test

1. `wiring/power_management` test has been updated to test both warm/cold boot behavior in a somewhat automated manner. Follow the instructions/notes the test presents. It may be run with or without the battery on any platform with a PMIC.
2. Modems on Gen 2 / Gen 3 cellular devices should not brownout on warm boot.

### Example App

N/A

### References

- [CH63632]

---

### Completeness

- [ ] Added documentation


system/src/system_power_manager.cpp
```diff
@@ -357,22 +369,57 @@ void PowerManager::isrHandler() {
...
+    power.enableBATFET();
+  }
+
+  // Enable buck converter
+  bool inDpDm = power.isInDPDM();
+  if (!power.isBuckEnabled()) {
+    power.enableBuck();
+    // Make sure we restart DPDM as otherwise we may get stuck
+    // with an invalid ILIM due to reading back and writing
+    // a transient value by modifying the same register with enableBuck()
```
sergeuz, line 404: Doesn't the same apply to `enableBATFET()` called above?

system/src/system_power_manager.cpp
```diff
@@ -357,22 +369,57 @@ void PowerManager::isrHandler() {
...
+  bool vinWithValidLimit = (power.getInputCurrentLimit() > 100) &&
+      ((powerSource == POWER_SOURCE_VIN) ||
+      ((config_.flags & HAL_POWER_USE_VIN_SETTINGS_WITH_USB_HOST) && powerSource == POWER_SOURCE_USB_HOST));
+
+  if (dpdm) {
+    if (inDpDm || !vinWithValidLimit) {
+      // Force-start input current limit detection
+      // only if we modified REG00 while in DPDM or we are not powered by VIN
+      // and we have a non-100mA limit set
+      power.enableDPDM();
```
sergeuz, line 420: Will this actually restart the detection if REG07[7] is already 1?

avtolstoy: 👍 Let's double check. From what I've seen - yes.

sergeuz (commented):
These tests are intermittently failing on a Boron without a battery that is powered via USB:
```
Assertion failed: (System.powerSource()=3) == ((int)POWER_SOURCE_USB_HOST=2), file tests/wiring/power_management/power.cpp, line 182.
Test POWER_01_PoweredByUsbHostAndBatteryStateIsValid failed.

Assertion failed: (cur=500) >= (prev=1500), file tests/wiring/power_management/power.cpp, line 277.
Test POWER_04_PoweredByUsbHostNoCurrentGlitchColdBootup failed.

Assertion failed: (cur=500) >= (prev=1500), file tests/wiring/power_management/power.cpp, line 402.
Test POWER_10_PoweredByVinNoCurrentGlitchColdBootup failed.
```

sergeuz (approved):
> These tests are intermittently failing on a Boron

The last two commits fixed it for me :+1:

system/src/system_power_manager.cpp
avtolstoy: Doesn't work. A change has been added to wait for it to finish.

system/src/system_power_manager.cpp
```diff
@@ -41,6 +41,8 @@ constexpr system_tick_t DEFAULT_FAULT_SUPPRESSION_PERIOD = 60000;
 
 constexpr system_tick_t DEFAULT_QUEUE_WAIT = 1000;
 constexpr system_tick_t DEFAULT_WATCHDOG_TIMEOUT = 60000;
+// FIXME: make sure to change setWatchdog() call to use the same interval
```
technobly, line 44: When will we make this setWatchdog() change?

technobly (approved):
Tested on U260 and R410 Electrons... with TEST=wiring/power_management 👍

system/src/system_power_manager.cpp
avtolstoy: image

PR #2216, opened 2020-10-14
[wiring] add more operators for BleAddress/BleUuid/IPAddress.

XuGuohui:
As the title describes.
---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_ipaddress.h
+++ b/wiring/inc/spark_wiring_ipaddress.h
@@ -57,7 +57,9 @@ public:
     // Constructors
     IPAddress();
     IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet);
+    // A 32-byte defining the 4 IPv4 octets, in host order.
     IPAddress(uint32_t address);
+    // 4 bytes defining the IP address in network order
     IPAddress(const uint8_t* address);
     IPAddress(const HAL_IPAddress& address);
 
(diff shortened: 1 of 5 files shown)


wiring/inc/spark_wiring_ipaddress.h
```diff
@@ -57,7 +57,9 @@ class IPAddress : public Printable {
     // Constructors
     IPAddress();
     IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet);
+    // A 32-byte defining the 4 IPv4 octets, in host order.
```
avtolstoy, line 60: 32-bits.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -229,23 +238,81 @@ uint8_t BleAddress::operator[](uint8_t i) const {
...
+}
+
 BleAddress& BleAddress::operator=(const uint8_t addr[BLE_SIG_ADDR_LEN]) {
+    // We just update the address value while keep the address type
     memcpy(address_.addr, addr, BLE_SIG_ADDR_LEN);
     return *this;
 }
 
+BleAddress& BleAddress::operator=(const char* address) {
+    // We just update the address value while keep the address type
```
avtolstoy, line 257: Any chance this might result in some inconsistencies? Should we parse the address type somehow?

avtolstoy (approved):

wiring/src/spark_wiring_ble.cpp
XuGuohui: We cannot determine the address type for the given string. For example, if the most significant two bits are `0b11`, it can be either a public address or a static address. Same to thee cases that the most significant two bits are `0x00` or `0x01`. I think it's user's responsibility to check the validity using `isValid()` that will be introduced in another PR.

wiring/inc/spark_wiring_ipaddress.h
XuGuohui: Fixed.

PR #2217, opened 2020-10-14
Gen3: Network management improvement

XuGuohui:
This PR improves the network connection and modem power management for Gen3 platforms. Whenever a network API is called, the final network state intends to be consistent with the final call. And it gracefully transits network state according to the sequence that network APIs are called.

## Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

void setup() {

    Log.info("Type 'o', 'f', 'c', 'd', 'p' to test.");
    Log.info("'o' -- WiFi/Cellular.on()");
    Log.info("'f' -- WiFi/Cellular.off()");
    Log.info("'c' -- WiFi/Cellular.connect()");
    Log.info("'d' -- WiFi/Cellular.disconnect()");
    Log.info("'p' -- Particle.connect()");
    Log.info("'q' -- Particle.disconnect()");
}

void loop() {
    if (Serial.available()) {
        char cmd = Serial.read();
        switch (cmd) {
            case 'o': {
                #if HAL_PLATFORM_CELLULAR
                Log.info("Cellular.on()");
                Cellular.on();
                #else
                Log.info("WiFi.on()");
                WiFi.on();
                #endif
                break;
            }
            case 'f': {
                #if HAL_PLATFORM_CELLULAR
                Log.info("Cellular.off()");
                Cellular.off();
                #else
                Log.info("WiFi.off()");
                WiFi.off();
                #endif
                break;
            }
            case 'c': {
                #if HAL_PLATFORM_CELLULAR
                Log.info("Cellular.connect()");
                Cellular.connect();
                #else
                Log.info("WiFi.connect()");
                WiFi.connect();
                #endif
                break;
            }
            case 'd': {
                #if HAL_PLATFORM_CELLULAR
                Log.info("Cellular.disconnect()");
                Cellular.disconnect();
                #else
                Log.info("WiFi.disconnect()");
                WiFi.disconnect();
                #endif
                break;
            }
            case 'p': {
                Log.info("Particle.connect()");
                Particle.connect();
                break;
            }
            case 'q': {
                Log.info("Particle.disconnect()");
                Particle.disconnect();
                break;
            }
            default: break;
        }
    }
}

```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/cellular/pppncpnetif.cpp
+++ b/hal/network/lwip/cellular/pppncpnetif.cpp
@@ -119,6 +119,8 @@ void PppNcpNetif::loop(void* arg) {
         self->celMan_->ncpClient()->enable(); // Make sure the client is enabled
         os_semaphore_take(self->netifSemaphore_, timeout, false);
 
+        self->setExpectedInternalState(self->lastNetifEvent_);
+
         if (self->expectedNcpState_ == NcpState::ON && self->celMan_->ncpClient()->ncpState() != NcpState::ON) {
             auto r = self->celMan_->ncpClient()->on();
             if (r != SYSTEM_ERROR_NONE && r != SYSTEM_ERROR_ALREADY_EXISTS) {
@@ -169,7 +171,6 @@ int PppNcpNetif::up() {
         return SYSTEM_ERROR_NONE;
     }
     lastNetifEvent_ = NetifEvent::Up;
-    setExpectedInternalState(lastNetifEvent_);
     // It's fine even if we failed to give the semaphore, as we specify a timeout taking the semaphore.
     os_semaphore_give(netifSemaphore_, false);
     return SYSTEM_ERROR_NONE;
@@ -190,7 +191,6 @@ int PppNcpNetif::down() {
         // Disable the client to interrupt its current operation
         client->disable();
     }
-    setExpectedInternalState(lastNetifEvent_);
     // It's fine even if we failed to give the semaphore, as we specify a timeout taking the semaphore.
     os_semaphore_give(netifSemaphore_, false);
     return SYSTEM_ERROR_NONE;
@@ -201,7 +201,6 @@ int PppNcpNetif::powerUp() {
         return SYSTEM_ERROR_NONE;
     }
     lastNetifEvent_ = NetifEvent::PowerOn;
-    setExpectedInternalState(lastNetifEvent_);
     // It's fine even if we failed to give the semaphore, as we specify a timeout taking the semaphore.
     os_semaphore_give(netifSemaphore_, false);
     return SYSTEM_ERROR_NONE;
@@ -216,7 +215,7 @@ int PppNcpNetif::powerDown() {
      * or hardware pins will fail in some case, which will result a hardreset of the modem. That takes
      * longer to turn off the modem. Simply let the NCP initialization complete and then execute the power
      * off sequence. */
-    setExpectedInternalState(lastNetifEvent_);
+
     // It's fine even if we failed to give the semaphore, as we specify a timeout taking the semaphore.
     os_semaphore_give(netifSemaphore_, false);
     return SYSTEM_ERROR_NONE;
--- a/hal/network/lwip/cellular/pppncpnetif.cpp
+++ b/hal/network/lwip/cellular/pppncpnetif.cpp
@@ -45,20 +45,8 @@ using namespace particle::net;
 
 namespace {
 
-enum class NetifEvent {
-    None = 0,
-    Up = 1,
-    Down = 2,
-    Exit = 3,
-    PowerOff = 4,
-    PowerOn = 5
-};
-
 // 5 minutes
 const unsigned PPP_CONNECT_TIMEOUT = 5 * 60 * 1000;
-
-// FIXME: unnecessarily large but helps avoid deadlocks/missing events between LwIP and NCP contexts
-const auto NCP_NETIF_EVENT_QUEUE_SIZE = 32;
 
 } // anonymous
 
@@ -66,11 +54,13 @@ const auto NCP_NETIF_EVENT_QUEUE_SIZE = 32;
 PppNcpNetif::PppNcpNetif()
         : BaseNetif(),
           exit_(false),
-          expectedNetifState_(NETIF_STATE_OFF) {
+          lastNetifEvent_(NetifEvent::None),
+          expectedNcpState_(NcpState::OFF),
+          expectedConnectionState_(NcpConnectionState::DISCONNECTED) {
 
     LOG(INFO, "Creating PppNcpNetif LwIP interface");
-
-    SPARK_ASSERT(os_queue_create(&queue_, sizeof(NetifEvent), NCP_NETIF_EVENT_QUEUE_SIZE, nullptr) == 0);
+    // A boolean semaphore is sufficient to synchronize the internal thread.
+    SPARK_ASSERT(os_semaphore_create(&netifSemaphore_, 1, 0) == 0);
 
     client_.setNotifyCallback(pppEventHandlerCb, this);
     client_.start();
@@ -78,11 +68,10 @@ PppNcpNetif::PppNcpNetif()
 
 PppNcpNetif::~PppNcpNetif() {
     exit_ = true;
-    if (thread_ && queue_) {
-        auto ex = NetifEvent::Exit;
-        os_queue_put(queue_, &ex, 1000, nullptr);
+    if (thread_ && netifSemaphore_) {
+        os_semaphore_give(netifSemaphore_, false);
         os_thread_join(thread_);
-        os_queue_destroy(queue_, nullptr);
+        os_semaphore_destroy(netifSemaphore_);
     }
 }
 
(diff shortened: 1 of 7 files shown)


hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -204,28 +223,21 @@ void Esp32NcpNetif::loop(void* arg) {
...
-        up_ = true;
-    }
-    return r;
+    return os_queue_put(queue_, &ev, 0, nullptr);
 }
 
 int Esp32NcpNetif::down() {
     const auto client = wifiMan_->ncpClient();
-    if (client->connectionState() != NcpConnectionState::CONNECTED) {
+    if (client->connectionState() != NcpConnectionState::CONNECTED && client->ncpState() == NcpState::ON) {
```
avtolstoy, line 231: The idea behind calling disable is to force abort the ongoing (blocking) operation. It may take quite a bit of time for NCP state go to `ON`. I'm not sure it's that good of an idea to add this check. We might be performing some initialization currently which we want to abort.

hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -204,28 +223,21 @@ void Esp32NcpNetif::loop(void* arg) {
...
         client->disable();
     }
     NetifEvent ev = NetifEvent::Down;
     int r = os_queue_put(queue_, &ev, 0, nullptr);
     if (r) {
-        // FIXME: we'll miss this event, just as an attempt to potentially recover
-        // set internal state.
-        up_ = false;
-        client->disable();
+        if (client->ncpState() == NcpState::ON) {
```
avtolstoy, line 238: We've just removed a workaround for a case where we for fail to post an event into the queue. We should still set the intended state here and call `disable()` to unblock the thread that processes these events. Otherwise it will be lost.

hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -204,28 +223,21 @@ void Esp32NcpNetif::loop(void* arg) {
...
 int Esp32NcpNetif::up() {
     NetifEvent ev = NetifEvent::Up;
-    int r = os_queue_put(queue_, &ev, 0, nullptr);
-    if (r) {
-        // FIXME: we'll miss this event, just as an attempt to potentially recover
-        // set internal state.
-        up_ = true;
-    }
-    return r;
+    return os_queue_put(queue_, &ev, 0, nullptr);
```
avtolstoy, line 226: Again, we've just removed the workaround for cases where the queue is filled up and the event will be completely missed.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -361,7 +361,7 @@ int SaraNcpClient::off() {
         // WARN: We assume that the modem can turn off itself reliably.
     } else {
         // Power down using hardware
-        modemPowerOff();
+        CHECK(modemPowerOff());
```
avtolstoy, line 364: This leaves the NCP client in a half-initialized state. Any external interaction with it may bring it into an on state again (see `waitReady()`). If the idea is to make sure we are powered off, we should rely on the power state to decide that we should call `off()` again.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -368,7 +368,7 @@ int QuectelNcpClient::off() {
         // WARN: We assume that the modem can turn off itself reliably.
     } else {
         // Power down using hardware
-        modemPowerOff();
+        CHECK(modemPowerOff());
```
avtolstoy, line 371: This leaves the NCP client in a half-initialized state. Any external interaction with it may bring it into an on state again (see `waitReady()`). If the idea is to make sure we are powered off, we should rely on the power state to decide that we should call `off()` again.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -111,29 +112,55 @@ void PppNcpNetif::loop(void* arg) {
             LOG(TRACE, "PPP netif event from queue: %d", ev);
             switch (ev) {
                 case NetifEvent::Up: {
-                    self->upImpl();
+                    self->expectedNetifState_ = NETIF_STATE_UP;
                     break;
                 }
                 case NetifEvent::Down: {
-                    self->downImpl();
+                    if (self->celMan_->ncpClient()->ncpState() == NcpState::ON) {
```
avtolstoy, line 119: This check should rely on `expectedNetifState_`, not the actual NCP state.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -151,28 +178,26 @@ void PppNcpNetif::loop(void* arg) {
...
 
 int PppNcpNetif::down() {
     const auto client = celMan_->ncpClient();
-    if (client->connectionState() != NcpConnectionState::CONNECTED) {
+    /* Note: It's possible that we have powered on the modem but the NCP client initialization is still in progress.
+     * For example, invokes Cellular.connect() followed by Cellular.disconnect().
+     * If we interrupted the operation, we will probably fail to turn off the modem in enable() that is invoked in the PPP thread.
+     * It would result in a hard reset of the modem, which will take longer to turn off the modem. Simply let the NCP client
+     * finalize the initialization procedure and make it reside in ON state. */
+    if (client->connectionState() != NcpConnectionState::CONNECTED && client->ncpState() == NcpState::ON) {
```
avtolstoy, line 191: The idea behind calling disable is to force abort the ongoing (blocking) operation. It may take quite a bit of time for NCP state go to `ON`. I'm not sure it's that good of an idea to add this check.

The hard reset timing concerns only apply to R410, however they are still lower than waiting for the whole intiialization sequence to complete. If you take a look at it you can see that it may take even several minutes in some conditions.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -151,28 +178,26 @@ void PppNcpNetif::loop(void* arg) {
...
         // Disable the client to interrupt its current operation
         client->disable();
     }
     NetifEvent ev = NetifEvent::Down;
     int r = os_queue_put(queue_, &ev, 0, nullptr);
     if (r) {
-        // FIXME: we'll miss this event, just as an attempt to potentially recover
```
avtolstoy, line 172: We've just removed a workaround for a case where we for fail to post an event into the queue. We should still set the intended state here and call `disable()` to unblock the thread that processes these events. Otherwise it will be lost.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -151,28 +178,26 @@ void PppNcpNetif::loop(void* arg) {
 
 int PppNcpNetif::up() {
     NetifEvent ev = NetifEvent::Up;
-    int r = os_queue_put(queue_, &ev, 0, nullptr);
```
avtolstoy, line 154: Again, we've just removed the workaround for cases where the queue is filled up and the event will be completely missed.

hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -159,40 +160,58 @@ void Esp32NcpNetif::loop(void* arg) {
...
-                    if (self->upImpl()) {
-                        self->wifiMan_->ncpClient()->off();
-                    }
+                    self->expectedNetifState_ = NETIF_STATE_UP;
                     break;
                 }
                 case NetifEvent::Down: {
-                    self->downImpl();
-                    // self->wifiMan_->ncpClient()->off();
+                    if (self->wifiMan_->ncpClient()->ncpState() == NcpState::ON) {
```
avtolstoy, line 170: This check should rely on `expectedNetifState_`, not the actual NCP state.

avtolstoy (changes requested):

hal/network/lwip/esp32/esp32ncpnetif.cpp
XuGuohui: Could elaborate how the previous fix works? If the queue is filled up, the next event popped from the queue will probably overwrite the `up_` flag, and the request is going to be lost as well.

avtolstoy: That's exactly why there is a FIXME there. Since we are refactoring the eventing, we should watch out for this. It's always possible to fill up the queue e.g. just issuing something in a loop accidentally.

```c++
void loop() {
    if (!Cellular.ready()) {
         Cellular.connect();
    } else if (Time.now() == 123123123) {
         // Just an example that this has a potential to get lost
         Cellular.off();
    }
}
```

We either have to live with the fact that we might need to block, or we can do something similar and e.g. calculate and store the intended state at the time of posting an event to the queue and be able to detect any discrepancies.

XuGuohui: I think we can go in the way that blocking with a certain timeout posting an event and discard the event that is the same  with the last posted event.

avtolstoy: That works as well, yeah.

hal/network/lwip/esp32/esp32ncpnetif.cpp
XuGuohui: I think calling `disbale()` in this `down()` just intends to abort the on-going attempt establishing a connection. If the NCP is transiting to the `ON` state, it's beyond the intention of calling the `disable()` here. On the other side, the below logic:
```
void fn() {
    Network.connect();
    Network.disconnect();
}
```
expects the NCP state to be `ON` and ready to accept AT commands. With the current implementation, the NCP will reside in the `OFF` state if transiting to `ON` state is aborted.

avtolstoy: It's not guaranteed that the next operation isn't going to be `Network.off()` and I'd argue that's actually a more likely scenario if we take a look at how https://github.com/particle-iot/device-os/blob/develop/system/src/system_network_manager_api.cpp#L277 is implemented for example.

But I agree, it's probably fine to leave things as-is. Aborting an ongoing power-on/initialization attempt potentially prevents us dropping into a connection attempt if we take a look at https://github.com/particle-iot/device-os/pull/2217/files#diff-333366c726fa5a73345b7c9ced8858345aa5db8b5d22add908e50f7045d83c8eR202, but perhaps it's going to take longer for things to settle, especially if we consider your case where we anyway have to settle on a powered state. Should we perhaps do something in `powerDown()` to abort power-on/initialization sequence to speed things up?

Anyway, no strong preferences, but would be nice to make some comparisons with before/after.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -150,41 +165,61 @@ void PppNcpNetif::loop(void* arg) {
...
-    if (r) {
-        // FIXME: we'll miss this event, just as an attempt to potentially recover
-        // set internal state.
-        up_ = true;
+    if (lastNetifEvent_ == NetifEvent::Up) {
+        return SYSTEM_ERROR_NONE;
     }
-    return r;
+    lastNetifEvent_ = NetifEvent::Up;
+    setExpectedInternalState(lastNetifEvent_);
```
avtolstoy, line 172: Note that three separate sequential atomic operations do not make the whole change atomic. You may end up in an inconsistent state. This needs to be a single `struct` for example and there should be a single modification or we should rely on the NCP thread to at least perform the other two modifications based on the posted event.

This applies to all the other cases.

It's probably fine in our particular use case as all of these events will be coming from system thread though.

hal/network/lwip/esp32/esp32ncpnetif.cpp
```diff
@@ -48,40 +48,33 @@ LOG_SOURCE_CATEGORY("net.esp32ncp")
...
-    Up = 1,
-    Down = 2,
-    Exit = 3,
-    PowerOff = 4,
-    PowerOn = 5
-};
-
-// FIXME: unnecessarily large but helps avoid deadlocks/missing events between LwIP and NCP contexts
-const auto NCP_NETIF_EVENT_QUEUE_SIZE = 32;
+constexpr system_tick_t NETIF_EVENT_WAIT_TIMEOUT_MS = 30000;
```
avtolstoy, line 51: This isn't used anywhere.

avtolstoy (approved):
@XuGuohui Let's add something for wiznetif as well (perhaps in another PR) so that it at least generates somewhat believable power events/state. This is only partially related to this PR, but the way I looked at the implementation it looks like `waitInterfaceOff()` will never succeed for it.

XuGuohui:
>Let's add something for wiznetif as well (perhaps in another PR) so that it at least generates somewhat believable power events/state. This is only partially related to this PR, but the way I looked at the implementation it looks like waitInterfaceOff() will never succeed for it.

It's more like modem power management improvements. I'd like to achieve that in a separated PR.

hal/network/lwip/esp32/esp32ncpnetif.cpp
XuGuohui: I'm going to do some tests to compare the time consumption between 3.0 and pre-3.0.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -99,41 +89,68 @@ if_t PppNcpNetif::interface() {
...
-                    self->celMan_->ncpClient()->off();
-                    break;
-                }
-                case NetifEvent::PowerOn: {
-                    self->celMan_->ncpClient()->on();
-                    break;
-                }
+        os_semaphore_take(self->netifSemaphore_, timeout, false);
+
+        self->setExpectedInternalState(self->lastNetifEvent_);
```
avtolstoy, line 122: Without the queue now this may lose events. Imagine the following two calls:
```c++
Cellular.off();
Cellular.disconnect();
```

If for whatever reason the NCP netif thread doesn't have the time to act on both of these changes (e.g. it's blocked doing something in the meantime), we'll end up in just disconnected state with modem powered on.

avtolstoy (changes requested):

hal/network/lwip/cellular/pppncpnetif.cpp
XuGuohui: Good catch! 👍 The reason being is we don't set the expected state immediately on event requested.

avtolstoy: Yeah, my suggestion was to go with this approach only if there is a queue, but I didn't spell it out. If you do want to fix the original issue with sequential atomic operations an std::atomic struct would probably be better, the only caveat is that you'll need to modify it with compare_exchange_strong to make sure that the thread that modifies accounts for the changes other threads might have done.

avtolstoy: Or just leave things as they were, knowing that these events only come from system thread and it's fine to make sequential state changes.

XuGuohui: Alright, I'll leave things as-was and just add a comment in case that these APIs may be called by other thread.

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -150,41 +165,64 @@ void PppNcpNetif::loop(void* arg) {
...
-        // set internal state.
-        up_ = true;
+    if (lastNetifEvent_ == NetifEvent::Up) {
+        return SYSTEM_ERROR_NONE;
     }
-    return r;
+    // FIXME:
+    // The following separate sequential atomic operations do not make the whole change atomic.
+    // We may end up in an inconsistent state. Same to wherever the changes being made.
+    lastNetifEvent_ = NetifEvent::Up;
```
avtolstoy, line 174: @XuGuohui Let's at least make sure that these operations are completely atomic:
```c++
if (lastNetifEvent_.exchange(NetifEvent::Up, std::memory_order_acq_rel) == NetifEvent::Up) {
    return SYSTEM_ERROR_NONE;
}
```

avtolstoy (changes requested):

avtolstoy (approved):

hal/network/lwip/cellular/pppncpnetif.cpp
XuGuohui: This comment is actually for line 168?

avtolstoy: For 168 to 174. Two operations combined into 1 atomic read-modify-write.

avtolstoy: And also all the other similar cases.

XuGuohui: Done

PR #2218, opened 2020-10-14
[Electron] Workaround for modem HAL relying on system networking code to re-attempt initialization

avtolstoy:
### Problem

There are two problems which have some common roots. If the power-on sequence on Electrons fails (and there are two stages in it: power on and SIM initialization, additional initialization), a specific set of actions needs to be taken by the _system networking layer_ to get the modem HAL into a proper state.

1. We are relying on a hack to get the system layer to restart poweron initialization when `MDMParser::init()` fails to initialize and requires a modem reset e.g. to set UMNO profile. This used to work somewhat until we've introduced reliable modem power off in one of the Sleep 2.0 PRs, which now causes the modem to be hard powercycled almost immediately after we've reset it through an AT command.

```
   219.818 AT send      14 "AT+CFUN=15,0\r\n"
   219.828 AT read OK    6 "\r\nOK\r\n"
0000219828 [system] INFO: Sim Ready
0000219829 [system] INFO: ARM_WLAN_WD 1
0000249833 [system] WARN: Resetting WLAN due to WLAN_WD_TO()
0000249833 [system] INFO: Clearing WLAN WD in disconnect()
0000249838 [system] INFO: Clearing WLAN WD in disconnect()
0000249843 [system] INFO: OFF 1
[ Modem::powerOff ] modem is on unexpectedly. Turning it off...

[ Modem::powerOff ] failed
[ ElectronSerialPipe::end ] pipeTx=0 pipeRx=0

[ Modem::resume ] = = = = = = = = = = = = = = =
[ ElectronSerialPipe::begin ] pipeTx=0 pipeRx=0

[ Modem::powerOn ] = = = = = = = = = = = = = =
   280.173 AT send       4 "AT\r\n"
   283.484 AT send       4 "AT\r\n"
```

2. The second step in poweron (`MDMParser::init()`) was skipped whenever initial SIM initialization in `MDMParser::powerOn()` failed, because for some reason reinitialization is happening in `has_credentials()`

```
0000321278 [system] INFO: OFF 1
[ Modem::powerOff ] modem is on unexpectedly. Turning it off...

[ Modem::powerOff ] failed
[ ElectronSerialPipe::end ] pipeTx=0 pipeRx=0

[ Modem::resume ] = = = = = = = = = = = = = = =
[ ElectronSerialPipe::begin ] pipeTx=0 pipeRx=0

[ Modem::powerOn ] = = = = = = = = = = = = = =
   351.608 AT send       4 "AT\r\n"
   351.613 AT read UNK   3 "AT\r"
   351.614 AT read OK    6 "\r\nOK\r\n"
   351.614 AT send       9 "AT+CGMM\r\n"
   361.819 AT send       6 "ATE0\r\n"
0000371820 [system] INFO: SIM/modem not responsive or SIM not inserted/requires a PIN.

[ Modem::powerOn ] = = = = = = = = = = = = = =
   372.136 AT send       4 "AT\r\n"
   375.447 AT send       4 "AT\r\n"
   378.758 AT send       4 "AT\r\n"
   378.764 AT read UNK   3 "AT\r"
   378.765 AT read OK    6 "\r\nOK\r\n"
   378.765 AT send       9 "AT+CGMM\r\n"
   378.778 AT read UNK   8 "AT+CGMM\r"
   378.779 AT read UNK  18 "\r\nSARA-R410M-02B\r\n"
   378.781 AT read OK    6 "\r\nOK\r\n"
   378.985 AT send       6 "ATE0\r\n"
   378.992 AT read UNK   5 "ATE0\r"
   378.993 AT read OK    6 "\r\nOK\r\n"
   378.993 AT send      11 "AT+CMEE=2\r\n"
   379.003 AT read OK    6 "\r\nOK\r\n"
   379.003 AT send      19 "AT+CMER=1,0,0,2,1\r\n"
   379.012 AT read OK    6 "\r\nOK\r\n"
   379.012 AT send      15 "AT+IPR=115200\r\n"
   379.021 AT read OK    6 "\r\nOK\r\n"
   379.121 AT send      10 "AT+CPIN?\r\n"
   379.129 AT read  +   16 "\r\n+CPIN: READY\r\n"
   379.130 AT read OK    6 "\r\nOK\r\n"
0000379131 [system] INFO: Sim Ready
0000379135 [system] INFO: ARM_WLAN_WD 1
   379.138 AT send       4 "AT\r\n"
   379.148 AT read OK    6 "\r\nOK\r\n"

[ Modem::register ] = = = = = = = = = = = = = =
   379.150 AT send      10 "AT+CFUN?\r\n"
   379.162 AT read  +   12 "\r\n+CFUN: 1\r\n"
   379.163 AT read OK    6 "\r\nOK\r\n"
   379.164 AT send      12 "AT+CEREG=2\r\n"
   379.175 AT read OK    6 "\r\nOK\r\n"
   379.175 AT send      11 "AT+CEREG?\r\n"
   379.186 AT read  +   34 "\r\n+CEREG: 2,1,\"4118\",\"3376211\",8\r\n"
   379.188 AT read OK    6 "\r\nOK\r\n"
   379.190 AT send       4 "AT\r\n"
   379.200 AT read OK    6 "\r\nOK\r\n"
   379.200 AT send       9 "AT+CIMI\r\n"
   379.210 AT read UNK  19 "\r\n732123200003356\r\n"
   379.211 AT read OK    6 "\r\nOK\r\n"
```

(see https://app.clubhouse.io/particle/story/64669, https://app.clubhouse.io/particle/story/64668)

### Solution

Add even more hacks! Ideally all these retries should be contained with the modem code and spread through multiple subsystems but this potentially brings some behavior changes, which we shouldn't be making so far along into LTS.

The main idea behind this PR is to check whether power-on succeeded and if not, rely on [`network_connect()`] (https://github.com/particle-iot/device-os/blob/develop/system/src/system_network_compat.cpp#L262) to trigger reinitialization.

### Steps to Test

Here are a few patches to trigger a particular issue:

1. `MDMParser::init()` failing, to emulate a case where we need a reset to set e.g. a correct UMNO profile:
```diff
diff --git a/hal/src/electron/modem/mdm_hal.cpp b/hal/src/electron/modem/mdm_hal.cpp
index f7c67672c..55edb6bb3 100644
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1160,6 +1160,9 @@ bool MDMParser::init(DevStatus* status)
             goto reset_failure;
         }
     } // if (_dev.dev == DEV_SARA_R410)
+#if 1
+    goto reset_failure;
+#endif // 1
     _resetFailureAttempts = 0;
     return true;
```

There should be 8 attempts made to run `MDMParser::init()` and after that the system should continue to attempt to register

2. `MDMParser::powerOn()` failing due to SIM failure
```diff
diff --git a/hal/src/electron/modem/mdm_hal.cpp b/hal/src/electron/modem/mdm_hal.cpp
index f7d08ec8d..42c15be37 100644
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -935,6 +935,13 @@ bool MDMParser::powerOn(const char* simpin)
     for (int i = 0; (i < 5) && (_dev.sim != SIM_READY) && !_cancel_all_operations; i++) {
         sendFormated("AT+CPIN?\r\n");
         int ret = waitFinalResp(_cbCPIN, &_dev.sim);
+#if 1
+        static int counter = 0;
+        if (counter++ <= 2) {
+            _dev.sim = SIM_MISSING;
+            goto failure;
+        }
+#endif // 1
         // having an error here is ok (sim may still be initializing)
         if ((RESP_OK != ret) && (RESP_ERROR != ret)) {
             goto failure;
```

This should trigger the system to enter listening mode as if the SIM card is not present and after 5 minutes (default listening mode timeout), the device should connect to the network.

3. `MDMParser::powerOn()` failing due to SIM failure (similar to the previous one, but we fail only once)
```diff
diff --git a/hal/src/electron/modem/mdm_hal.cpp b/hal/src/electron/modem/mdm_hal.cpp
index f7d08ec8d..42c15be37 100644
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -935,6 +935,13 @@ bool MDMParser::powerOn(const char* simpin)
     for (int i = 0; (i < 5) && (_dev.sim != SIM_READY) && !_cancel_all_operations; i++) {
         sendFormated("AT+CPIN?\r\n");
         int ret = waitFinalResp(_cbCPIN, &_dev.sim);
+#if 1
+        static int counter = 0;
+        if (counter++ <= 1) {
+            _dev.sim = SIM_MISSING;
+            goto failure;
+        }
+#endif // 1
         // having an error here is ok (sim may still be initializing)
         if ((RESP_OK != ret) && (RESP_ERROR != ret)) {
             goto failure;
```

The device should report SIM failure once and then go through `MDMParser::init()` and successfully connect.

4. Same thing as 1, but with a simulated failure preventing us from completing initailization entirely

```diff
diff --git a/hal/src/electron/modem/mdm_hal.cpp b/hal/src/electron/modem/mdm_hal.cpp
index f7c67672c..9a14d6d5d 100644
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -55,7 +55,7 @@ std::recursive_mutex mdm_mutex;
 #define MDM_URC_POLL_INTERVAL_MS (1000) //!< URC poll interval
 #define MDM_SOCKET_SEND_RETRIES_R4_BUG (3)
 #define MDM_RESET_FAILURE_TIMEOUT_MS (10000)
-#define MDM_RESET_FAILURE_MAX_ATTEMPTS (8)
+#define MDM_RESET_FAILURE_MAX_ATTEMPTS (9999)
 
 // ID of the PDP context used to configure the default EPS bearer when registering in an LTE network
 // Note: There are no PDP contexts in LTE, SARA-R4 uses this naming for the sake of simplicity
@@ -1161,6 +1161,9 @@ bool MDMParser::init(DevStatus* status)
         }
     } // if (_dev.dev == DEV_SARA_R410)
     _resetFailureAttempts = 0;
+#if 1
+    goto reset_failure;
+#endif // 1
     return true;
 
 failure:
```

This should cause a modem reset in 5 minutes and restart the initialization again.

5. These changes should not affect Photon/P1.

### Example App

N/A

### References

- [CH64668]
- [CH64669]

---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -56,6 +56,8 @@ std::recursive_mutex mdm_mutex;
 #define MDM_C_REG_POLL_INTERVAL_MS (30000) //!< CREG/CGREG/CEREG poll interval to check for unexpected modem reset
 #define MDM_SOCKET_SEND_RETRIES_R4_BUG (3)
 #define MDM_MAX_ERRORS (2) //!< max number of errors before action is taken to recover the modem
+#define MDM_RESET_FAILURE_TIMEOUT_MS (10000)
+#define MDM_RESET_FAILURE_MAX_ATTEMPTS (8)
 
 // ID of the PDP context used to configure the default EPS bearer when registering in an LTE network
 // Note: There are no PDP contexts in LTE, SARA-R4 uses this naming for the sake of simplicity
@@ -277,6 +279,8 @@ MDMParser::MDMParser(void)
 #ifdef MDM_DEBUG
     _debugLevel = 3;
 #endif
+    _resetFailureAttempts = 0;
+    _resetFailureTimestamp = 0;
 
     Hal_Pin_Info* pinMap = HAL_Pin_Map();
     RCC_AHB1PeriphClockCmd(RCC_AHB1Periph_GPIOB, ENABLE);
@@ -295,12 +299,16 @@ void MDMParser::setPowerMode(int mode) {
 }
 
 void MDMParser::cancel(void) {
-    MDM_INFO("\r\n[ Modem::cancel ] = = = = = = = = = = = = = = =");
+    if (!_cancel_all_operations) {
+        MDM_INFO("\r\n[ Modem::cancel ] = = = = = = = = = = = = = = =");
+    }
     _cancel_all_operations = true;
 }
 
 void MDMParser::resume(void) {
-    MDM_INFO("\r\n[ Modem::resume ] = = = = = = = = = = = = = = =");
+    if (_cancel_all_operations) {
+        MDM_INFO("\r\n[ Modem::resume ] = = = = = = = = = = = = = = =");
+    }
     _cancel_all_operations = false;
 }
 
@@ -916,11 +924,18 @@ failure:
 bool MDMParser::powerOn(const char* simpin)
 {
     LOCK();
+
     // The modem type won't change that easily
     auto device_type = _dev.dev;
     memset(&_dev, 0, sizeof(_dev));
     _dev.dev = device_type;
     bool retried_after_reset = false;
+
+    if (_resetFailureAttempts > 0) {
+        if (HAL_Timer_Get_Milli_Seconds() - _resetFailureTimestamp < MDM_RESET_FAILURE_TIMEOUT_MS) {
+            goto failure;
+        }
+    }
 
     _error = 0;
 
@@ -998,8 +1013,7 @@ failure:
 bool MDMParser::init(DevStatus* status)
 {
     LOCK();
-    static int resetFailureAttempts = 0;
-    MDM_INFO("\r\n[ Modem::init ] = = = = = = = = = = = = = = =");
+    MDM_INFO("\r\n[ Modem::init (attempt %u/%u) ] = = = = = = = = = =", (unsigned)_resetFailureAttempts + 1, (unsigned)MDM_RESET_FAILURE_MAX_ATTEMPTS);
 
     // Define all cstringhelpers up here, because "goto"s do not like bypassing inits
     CStringHelper str_imei(_dev.imei, sizeof(_dev.imei));
@@ -1193,7 +1207,7 @@ bool MDMParser::init(DevStatus* status)
             goto reset_failure;
         }
     } // if (_dev.dev == DEV_SARA_R410)
-    resetFailureAttempts = 0;
+    _resetFailureAttempts = 0;
     return true;
 
 failure:
@@ -1202,16 +1216,15 @@ failure:
 reset_failure:
     // Don't get stuck in a reset-retry loop
     // eDRX disables can take a couple or more resets, UMNOPROF requires 1 or 2 and UBANDMASK requires 1 or 2 worst case
-    if (resetFailureAttempts++ < 8) {
+    if (_resetFailureAttempts++ <= MDM_RESET_FAILURE_MAX_ATTEMPTS) {
         if (_atOk()) {
             sendFormated("AT+CFUN=15,0\r\n");
             waitFinalResp(nullptr, nullptr, CFUN_TIMEOUT);
-            _init = false; // invalidate init and prevent cellular registration
-            _pwr = false;  //   |
             // When this exits false, ARM_WLAN_WD 1 will fire and timeout after 30s.
             // MDMParser::powerOn and MDMParser::init will then be retried by the system.
             _incModemStateChangeCount();
         }
+        _resetFailureTimestamp = HAL_Timer_Get_Milli_Seconds();
     } else {
         // stop resetting, and try to register.
         // Preventing cellular registration gets us back to retrying init() faster,
@@ -1219,10 +1232,9 @@ reset_failure:
         // connect even with an error, since that error is persisting and might just
         // be a fluke or bug in the modem. Allowing to continue on eventually helps
         // the device recover in odd situations we can't predict.
-        resetFailureAttempts = 0;
-        return true;
+        _resetFailureAttempts = 0;
     }
-    return false;
+    return !_resetFailureAttempts;
 }
 
 void MDMParser::_incModemStateChangeCount(void) {
@@ -1262,6 +1274,8 @@ bool MDMParser::powerOff(void)
 {
     LOCK();
     const char * POWER_OFF_MSG = "\r\n[ Modem::powerOff ]";
+    _resetFailureAttempts = 0;
+    _resetFailureTimestamp = 0;
     bool ok = false;
     bool continue_cancel = false;
     bool check_ri = false;
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -280,7 +280,6 @@ MDMParser::MDMParser(void)
     _debugLevel = 3;
 #endif
     _resetFailureAttempts = 0;
-    _resetFailureTimestamp = 0;
 
     Hal_Pin_Info* pinMap = HAL_Pin_Map();
     RCC_AHB1PeriphClockCmd(RCC_AHB1Periph_GPIOB, ENABLE);
@@ -929,12 +928,6 @@ bool MDMParser::powerOn(const char* simpin)
     auto device_type = _dev.dev;
     bool retried_after_reset = false;
 
-    if (_resetFailureAttempts > 0) {
-        if (HAL_Timer_Get_Milli_Seconds() - _resetFailureTimestamp < MDM_RESET_FAILURE_TIMEOUT_MS) {
-            goto failure;
-        }
-    }
-
     memset(&_dev, 0, sizeof(_dev));
     _dev.dev = device_type;
 
@@ -1141,7 +1134,7 @@ bool MDMParser::init(DevStatus* status)
             // KORE AT&T or 3rd Party SIM
             else {
                 // continue on with init if we are trying to set SIM_SELECT a third time
-                if (resetFailureAttempts >= 2) {
+                if (_resetFailureAttempts >= 2) {
                     LOG(WARN, "UMNOPROF=1 did not resolve a built-in profile, please check if UMNOPROF=100 is required!");
                     continueInit = true;
                 }
@@ -1218,19 +1211,18 @@ failure:
 reset_failure:
     // Don't get stuck in a reset-retry loop
     // eDRX disables can take a couple or more resets, UMNOPROF requires 1 or 2 and UBANDMASK requires 1 or 2 worst case
-    if (_resetFailureAttempts++ <= MDM_RESET_FAILURE_MAX_ATTEMPTS) {
+    if (++_resetFailureAttempts < MDM_RESET_FAILURE_MAX_ATTEMPTS) {
         if (_atOk()) {
             sendFormated("AT+CFUN=15,0\r\n");
             waitFinalResp(nullptr, nullptr, CFUN_TIMEOUT);
-            // When this exits false, ARM_WLAN_WD 1 will fire and timeout after 30s.
-            // MDMParser::powerOn and MDMParser::init will then be retried by the system.
+            // MDMParser::powerOn and MDMParser::init will be retried by the system up to
+            // MDM_RESET_FAILURE_MAX_ATTEMPTS.
             _incModemStateChangeCount();
         }
-        _resetFailureTimestamp = HAL_Timer_Get_Milli_Seconds();
     } else {
         // stop resetting, and try to register.
         // Preventing cellular registration gets us back to retrying init() faster,
-        // and the timeout of 3 attempts is arbitrary but allows us to register and
+        // and the timeout of MDM_RESET_FAILURE_MAX_ATTEMPTS attempts is arbitrary but allows us to register and
         // connect even with an error, since that error is persisting and might just
         // be a fluke or bug in the modem. Allowing to continue on eventually helps
         // the device recover in odd situations we can't predict.
@@ -1277,7 +1269,6 @@ bool MDMParser::powerOff(void)
     LOCK();
     const char * POWER_OFF_MSG = "\r\n[ Modem::powerOff ]";
     _resetFailureAttempts = 0;
-    _resetFailureTimestamp = 0;
     bool ok = false;
     bool continue_cancel = false;
     bool check_ri = false;
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -927,8 +927,6 @@ bool MDMParser::powerOn(const char* simpin)
 
     // The modem type won't change that easily
     auto device_type = _dev.dev;
-    memset(&_dev, 0, sizeof(_dev));
-    _dev.dev = device_type;
     bool retried_after_reset = false;
 
     if (_resetFailureAttempts > 0) {
@@ -936,6 +934,9 @@ bool MDMParser::powerOn(const char* simpin)
             goto failure;
         }
     }
+
+    memset(&_dev, 0, sizeof(_dev));
+    _dev.dev = device_type;
 
     _error = 0;
 
@@ -961,6 +962,7 @@ bool MDMParser::powerOn(const char* simpin)
     for (int i = 0; (i < 5) && (_dev.sim != SIM_READY) && !_cancel_all_operations; i++) {
         sendFormated("AT+CPIN?\r\n");
         int ret = waitFinalResp(_cbCPIN, &_dev.sim);
+
         // having an error here is ok (sim may still be initializing)
         if ((RESP_OK != ret) && (RESP_ERROR != ret)) {
             goto failure;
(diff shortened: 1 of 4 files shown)


system/src/system_network_internal.h
```diff
@@ -599,34 +613,49 @@ class ManagedNetworkInterface : public NetworkInterface
...
         {
+            bool startedPowerOn = !SPARK_WLAN_STARTED;
+            WLAN_INITIALIZED = 0;
             const auto diag = NetworkDiagnostics::instance();
             diag->status(NetworkDiagnostics::TURNING_ON);
-            system_notify_event(network_status, network_status_powering_on);
+            if (startedPowerOn) {
+                // Suppress these events
+                system_notify_event(network_status, network_status_powering_on);
+            }
```
technobly, line 628: Nit: The reverse logic in "Suppress these events" makes my brain hurt.  Can we just omit this comment?

system/src/system_network_internal.h
```diff
@@ -599,34 +613,49 @@ class ManagedNetworkInterface : public NetworkInterface
...
             config_clear();
             const int ret = on_now();
             if (ret != 0) {
                 diag->lastError(ret);
+                if (!wlan_watchdog_duration) {
+                    // Just in case arm WLAN WD
+                    ARM_WLAN_WD(POWER_ON_WD);
+                }
```
technobly, line 636: I think this should at a minimum be a 10 minute timer due to auto COPS and network compliance.

system/src/system_network_internal.h
```diff
@@ -599,34 +613,49 @@ class ManagedNetworkInterface : public NetworkInterface
...
+            } else {
+                WLAN_INITIALIZED = 1;
```
technobly, line 638: If we get past this block and have failed to turn on the NCP `if (ret != 0)`, shouldn't we not be changing the LED to breathing blue yet?  Should that also be moved into the `if (!ret) { }` block?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1153,26 +1169,25 @@ bool MDMParser::init(DevStatus* status)
 reset_failure:
     // Don't get stuck in a reset-retry loop
     // eDRX disables can take a couple or more resets, UMNOPROF requires 1 or 2 and UBANDMASK requires 1 or 2 worst case
-    if (resetFailureAttempts++ < 8) {
+    if (_resetFailureAttempts++ <= MDM_RESET_FAILURE_MAX_ATTEMPTS) {
         if (_atOk()) {
             sendFormated("AT+CFUN=15,0\r\n");
             waitFinalResp(nullptr, nullptr, CFUN_TIMEOUT);
-            _init = false; // invalidate init and prevent cellular registration
-            _pwr = false;  //   |
```
technobly, line 1161: Nit: Update comments below to reflect changes.

technobly (changes requested):

system/src/system_network_internal.h
```diff
@@ -545,6 +554,11 @@ class ManagedNetworkInterface : public NetworkInterface
                 if (ret != 0) {
                     diag->lastError(ret);
                 }
+            } else if (onRet) {
+                // on() failed, we'll rely on the system loop to call connect() again
+                // FIXME: this is a UI workaround, as otherwise the device
+                // will be in a breathing blue state, which might be confusing
+                LED_SIGNAL_START(NETWORK_CONNECTING, NORMAL);
```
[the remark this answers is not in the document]
avtolstoy: Let's not introduce UI changes so far into 2.x as the comment states, especially considering this whole PR is just piling on more hacks. This was broken previously and we were blinking green under these circumstances. I'd like to keep the same behavior in order not to cause any confusion.

system/src/system_network_internal.h
avtolstoy: Again, I don't want to introduce any UI changes so late into 2.x.

system/src/system_network_internal.h
```diff
@@ -514,24 +515,32 @@ class ManagedNetworkInterface : public NetworkInterface
...
             SPARK_WLAN_STARTED = 1;
             SPARK_WLAN_SLEEP = 0;
 
+            // Only if power on succeeded
+            if (!onRet) {
+                CLR_WLAN_WD();
+                connect_init();
+            }
+
+            // This has side-effects on cellular platforms, so we need to run even if we've failed to completely power-on
```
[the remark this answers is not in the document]
avtolstoy: `has_credentials()` checks SIM state and calls `cellular_on()`: https://github.com/particle-iot/device-os/blob/develop/system/src/system_network_cellular.h#L144

system/src/system_network_internal.h
```diff
@@ -514,24 +515,32 @@ class ManagedNetworkInterface : public NetworkInterface
...
             WLAN_DISCONNECT = 0;
-            connect_init();
             SPARK_WLAN_STARTED = 1;
             SPARK_WLAN_SLEEP = 0;
 
+            // Only if power on succeeded
+            if (!onRet) {
+                CLR_WLAN_WD();
+                connect_init();
+            }
```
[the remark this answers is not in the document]
avtolstoy: There shouldn't be any practical difference.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -890,11 +898,19 @@ bool MDMParser::_powerOn(void)
...
+
     // The modem type won't change that easily
     auto device_type = _dev.dev;
+    bool retried_after_reset = false;
+
+    if (_resetFailureAttempts > 0) {
+        if (HAL_Timer_Get_Milli_Seconds() - _resetFailureTimestamp < MDM_RESET_FAILURE_TIMEOUT_MS) {
+            goto failure;
+        }
+    }
```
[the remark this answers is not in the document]
avtolstoy: We were previously relying on a 30s watchdog timeout, so this was an attempt to keep somewhat similar behavior. No strong preferences, we can remove it.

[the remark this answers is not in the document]
technobly: The 30s watchdog was a nuisance before :D So I'd prefer to remove the 10s busy wait.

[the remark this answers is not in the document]
avtolstoy: :+1:

system/src/system_network_internal.h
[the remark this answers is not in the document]
technobly: Could you add this to the comments, because it's not clear what the FIXME should correctly be sometime later.  In fact it says we should be breathing blue, not breathing white.

system/src/system_network_internal.h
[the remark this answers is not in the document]
technobly: Sure, but I guess I'm not seeing what the "side effects" are and why we need to run even if we failed to power-on.  Should this maybe say something like, "we'll end up in Listening mode with no Wi-Fi credentials or SIM card error/cellular power up failure"

system/src/system_network_internal.h
technobly: Right... it's safety code.  But if it ever comes into effect, let's default to 10 minutes... or just remove it.

system/src/system_network_internal.h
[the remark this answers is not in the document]
avtolstoy: 'side effects' as in computer science terms that it modifies the state, despite its name being `has_credentials()`. The 'modifies the state' being that it calls `cellular_on()` (which it shouldn't normally).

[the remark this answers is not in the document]
avtolstoy: :+1: to leaving a comment about the behavior if modem is not responsive or no SIM is present

system/src/system_network_internal.h
technobly: But you moved other code into that new block which changes behavior of system events and diagnostics :D. Also, we are fixing bugs ... not changing the way the UI is interpreted.  Gen 2 breathing white was always booting, breathing blue is NCP on, blinking green network connecting, etc..  If nothing else though, we should add a FIXME note.

avtolstoy: The behavior of system events and diagnostics is unchanged. The changes that had to be made are exactly so that we don't generate a torrent of events if we are waiting for power-on sequence to succeed.

technobly: Before if cellular_on() failed, it would have still generated the `network_status_on` event... now it waits until cellular is really on, right?

avtolstoy: Hm. Perhaps that `if (!ret)` should instead be `if (startedPowerOn)` then, right?

technobly: maybe `if (WLAN_INITIALIZED)` ?  don't we want to generate `network_status_on` after it's fully on, and not just in the process still of `network_status_powering_on` ?

technobly: if WLAN_INITIALIZED is the same as if !ret though.

technobly: Before it would just fail and say it succeeded via system LED and events.  Now we have an opportunity to fix that incorrect behavior.

avtolstoy: We can both agree on the fact that this needs to be fixed, but:

1. Something out there might be relying on this broken behavior (events)
2. The LED indication might be confusing as we might get stuck in breathing white for an extended period of time (up to several minutes)

technobly: We could open a CH ticket regarding this, and reference it in a FIXME comment in this code and move on :D But you'll also want to take the code out of the `if (!ret)` block.  I also don't think being stuck in breathing blue or blinking green is any less confusing if the real issue is the NCP hasn't finished turning on.

avtolstoy: I'll make all the other changes and will play around with this part to see how it's going to behave, so let's have a separate discussion tomorrow and come up with some sort of a decision.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1153,26 +1169,25 @@ bool MDMParser::init(DevStatus* status)
 reset_failure:
     // Don't get stuck in a reset-retry loop
     // eDRX disables can take a couple or more resets, UMNOPROF requires 1 or 2 and UBANDMASK requires 1 or 2 worst case
-    if (resetFailureAttempts++ < 8) {
+    if (_resetFailureAttempts++ <= MDM_RESET_FAILURE_MAX_ATTEMPTS) {
```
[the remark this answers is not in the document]
technobly: Actually, this should be `++_resetFailureAttempts < MDM_RESET_FAILURE_MAX_ATTEMPTS` since it's more of a do-while ;-)

technobly (approved):

system/src/system_network_internal.h
avtolstoy: Any reason we can't keep this at 5 minutes? 10 minutes seems like an awfully long time to wait for the modem to power on, initialize the SIM and perform some additional initialization?

technobly: Since it's a timeout we should never hit, I was thinking we should stay on the safe side of cellular network compliance.  If we do run into it continuously for whatever odd reason, at least we won't be doing anything potentially harmful.  If it would block a user from short circuiting recovery on their own though, 5 minutes might even be too long.

avtolstoy: Changed to 10

sergeuz:
The test #2 fails for me when I call `Cellular.connect()` in the app without establishing a cloud connection. The reason is that `WLAN_CONNECTING` is set to 1 after the `has_credentials()` check in `ManagedNetworkInterface::connect()` and thus the reconnection logic in `ManagedNetworkInterface::process()` is never triggered. Setting that flag earlier in the process causes another issue: `CellularNetworkInterface::on_stop_listening()` force-disconnects the device from the network when exiting the listening mode.

system/src/system_network_internal.h
```diff
@@ -773,6 +825,12 @@ class ManagedNetworkInterface : public NetworkInterface
 
     virtual int process()
     {
+        // Requested to power-on, failed initial power-on, requested to connect as well
+        if ((SPARK_WLAN_STARTED && !WLAN_INITIALIZED) && WLAN_CONNECTING) {
```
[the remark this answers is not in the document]
avtolstoy: I suppose this condition wouldn't happen on Photons. And even if it does, it won't probably hurt?

system/src/system_network_cellular.h
```diff
@@ -144,11 +144,15 @@ class CellularNetworkInterface : public ManagedIPNetworkInterface<CellularConfig
     bool has_credentials() override
     {
         bool rv = cellular_sim_ready(NULL);
-        LOG(INFO,"%s", (rv)?"Sim Ready":"SIM/modem not responsive or SIM not inserted/requires a PIN.");
+        if (!rv) {
+            LOG(INFO, "SIM/modem not responsive or SIM not inserted/requires a PIN.");
```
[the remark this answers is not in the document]
avtolstoy: Fixed.

system/src/system_network_internal.h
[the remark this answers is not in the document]
sergeuz: We discussed the issue described in the comment and decided that we're not going to fix in the scope of this PR as there's a risk of introducing a breaking change in the release.

sergeuz (approved):

PR #2219, opened 2020-10-15
[gen2][electron] modem brownout or unexpected reset causes system to think it still is connected [ch64465]

technobly:
### Problem

Electron was found to suffer from a modem reset while antenna was disconnected in low signal environment on battery only (and also with USB and battery). This leads to `C*REG=2` verbose mode + URCS being reset to `C*REG=0` non-verbose mode + no URCs, which causes the system firmware to not be notified that the modem has deregistered. NO_ACK publishes can continue to be sent into the breeze, and a user's firmware will not know `Particle.connected()` or `Cellular.ready()` are `false`.

#2215 addresses the root cause of the problem, but just in case for unexpected reasons the following solution offers a recovery method.

### Solution

To periodically poll for `C*REG?` every 30 seconds look for changes to CREG verbosity. If there is a valid response, and we see `C*REG: 0,x`, URCs are disabled. Notify the system of disconnect immediately, which will reinitialize the modem.

### Steps to Test

Run unit tests: `TEST=wiring/no_fixture_long_running` CELLULAR_03 & CELLULAR_04

### References

ch64465
#2215 addresses the root cause of the problem
ch63632

---

### Completeness

- [x] N/A Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -53,7 +53,9 @@ std::recursive_mutex mdm_mutex;
 #define MAX_SIZE        1024  //!< max expected messages (used with RX)
 #define USO_MAX_WRITE   1024  //!< maximum number of bytes to write to socket (used with TX)
 #define MDM_URC_POLL_INTERVAL_MS (1000) //!< URC poll interval
+#define MDM_C_REG_POLL_INTERVAL_MS (30000) //!< CREG/CGREG/CEREG poll interval to check for unexpected modem reset
 #define MDM_SOCKET_SEND_RETRIES_R4_BUG (3)
+#define MDM_MAX_ERRORS (2) //!< max number of errors before action is taken to recover the modem
 
 // ID of the PDP context used to configure the default EPS bearer when registering in an LTE network
 // Note: There are no PDP contexts in LTE, SARA-R4 uses this naming for the sake of simplicity
@@ -74,28 +76,28 @@ std::recursive_mutex mdm_mutex;
 #define USOST_TIMEOUT     ( 10 * 1000) /* 10s for R4, 1s for U2/G3, changed to 40s due to R410 firmware */
 #define USOST_TIMEOUT_R4  ( 40 * 1000) /* L0.0.00.00.05.08,A.02.04 background DNS lookup and other factors */
 #define USORF_TIMEOUT     ( 10 * 1000) /* FIXME: 1s for R4/U2/G3, but longer timeouts required in deployments */
-#define CEER_TIMEOUT      ( 10 * 1000)
-#define CEREG_TIMEOUT     ( 60 * 1000)
-#define CGDCONT_TIMEOUT   ( 10 * 1000)
-#define CGREG_TIMEOUT     ( 60 * 1000)
 #define CGATT_TIMEOUT     (180 * 1000)
+#define CEDRXS_TIMEOUT    ( 10 * 1000)
+#define CEER_TIMEOUT      ( 10 * 1000)
+#define CFUN_TIMEOUT      (180 * 1000)
+#define CGDCONT_TIMEOUT   ( 10 * 1000)
+#define CIMI_TIMEOUT      ( 10 * 1000) /* Should be immediate, but have observed 3 seconds occassionally on u-blox and rarely longer times */
 #define CMGS_TIMEOUT      (150 * 1000) /* 180s for R4 (set to 150s to match previous implementation) */
 #define COPS_TIMEOUT      (180 * 1000)
 #define CPWROFF_TIMEOUT   ( 40 * 1000)
-#define CREG_TIMEOUT      ( 60 * 1000)
 #define CSQ_TIMEOUT       ( 10 * 1000)
+#define CREG_TIMEOUT      ( 60 * 1000)
+#define CGREG_TIMEOUT     ( 60 * 1000)
+#define CEREG_TIMEOUT     ( 60 * 1000)
 #define UBANDSEL_TIMEOUT  ( 40 * 1000)
 #define UBANDMASK_TIMEOUT ( 10 * 1000)
+#define UCGED_TIMEOUT     ( 10 * 1000)
 #define UDNSRN_TIMEOUT    ( 30 * 1000) /* 70s for R4 (set to 30s to match previous implementation) */
-#define URAT_TIMEOUT      ( 10 * 1000)
+#define UMNOPROF_TIMEOUT  ( 10 * 1000)
 #define UPSD_TIMEOUT      ( 10 * 1000)
 #define UPSDA_TIMEOUT     (180 * 1000)
 #define UPSND_TIMEOUT     ( 10 * 1000)
-#define UMNOPROF_TIMEOUT  ( 10 * 1000)
-#define CEDRXS_TIMEOUT    ( 10 * 1000)
-#define CFUN_TIMEOUT      (180 * 1000)
-#define UCGED_TIMEOUT     ( 10 * 1000)
-#define CIMI_TIMEOUT      ( 10 * 1000) /* Should be immediate, but have observed 3 seconds occassionally on u-blox and rarely longer times */
+#define URAT_TIMEOUT      ( 10 * 1000)
 
 // num sockets
 #define NUMSOCKETS      ((int)(sizeof(_sockets)/sizeof(*_sockets)))
@@ -248,6 +250,10 @@ MDMParser::MDMParser(void)
     inst = this;
     memset(&_dev, 0, sizeof(_dev));
     memset(&_net, 0, sizeof(_net));
+    memset(_sockets, 0, sizeof(_sockets));
+    for (int socket = 0; socket < NUMSOCKETS; socket ++) {
+        _sockets[socket].handle = MDM_SOCKET_ERROR;
+    }
     _net.cgi.location_area_code = 0xFFFF;
     _net.cgi.cell_id = 0xFFFFFFFF;
     _ip        = NOIP;
@@ -265,11 +271,9 @@ MDMParser::MDMParser(void)
     _oldFirmwarePresent = true; // default to safe state until we determine modem firmware version
     _timePowerOn = 0;
     _timeRegistered = 0;
-    memset(_sockets, 0, sizeof(_sockets));
-    for (int socket = 0; socket < NUMSOCKETS; socket ++)
-        _sockets[socket].handle = MDM_SOCKET_ERROR;
-    _error = 0;
+    _lastVerboseCxregUpdate = 0;
     _lastProcess = 0;
+    _error = 0;
 #ifdef MDM_DEBUG
     _debugLevel = 3;
 #endif
@@ -394,7 +398,7 @@ bool MDMParser::_checkModem(bool force /* = true */) {
     // Either several commands timed-out or we
     // performed an additional AT/OK check here
     // and it also failed.
-    if (_error >= 2) {
+    if (_error >= MDM_MAX_ERRORS) {
         return false;
     }
 
@@ -410,6 +414,7 @@ bool MDMParser::_checkModem(bool force /* = true */) {
 }
 
 int MDMParser::process() {
+
     if (!(_init && _pwr)) {
         return 0;
     }
@@ -421,32 +426,33 @@ int MDMParser::process() {
     {
         LOCK();
         if (!_checkModem(false /* force */)) {
+            MDM_ERROR("[ Modem not responsive ]\r\n");
             // We've notified system layer that we'll require a modem reset
             // This will get reset once the system (or something else) calls on()
             _cancel_all_operations = true;
-            // Short-circuit disconnect() call
-            _ip = NOIP;
+            _ip = NOIP; // Short-circuit disconnect() call
             _attached = false;
             _activated = false;
             HAL_NET_notify_disconnected();
-            MDM_ERROR("[ Modem not responsive ]\r\n");
             return -1;
         }
     }
 
     if ((HAL_Timer_Get_Milli_Seconds() - _lastProcess) >= MDM_URC_POLL_INTERVAL_MS) {
         LOCK();
-        // Process URCs
-        waitFinalResp(nullptr, nullptr, 0);
+        waitFinalResp(nullptr, nullptr, 0); // Process URCs
         _lastProcess = HAL_Timer_Get_Milli_Seconds();
     }
 
     if (_activated && !_attached) {
         LOCK();
         if (!reconnect()) {
-            return -1;
+            return -2;
         }
     }
+
+    // Periodically check if C*REG URCs are still enabled, will cause modem re-init if not.
+    _checkVerboseCxreg();
 
     return 0;
 }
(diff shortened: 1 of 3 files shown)


hal/src/electron/modem/mdm_hal.h
```diff
@@ -661,6 +665,11 @@ class MDMParser
     volatile bool _cancel_all_operations;
     volatile uint32_t _error;
     system_tick_t _lastProcess;
+    system_tick_t _lastVerboseC_regUpdate;
+    bool _verboseCregEnabled;
+    bool _verboseCgregEnabled;
+    bool _verboseCeregEnabled;
+    int _reinitModem;
```
keeramis, line 672: May be Nit: It looks like `_reinitModem` takes these two values `0xFF` and `0x55` and no operations are being done other than assigning and checking. May be can get away with `bool` instead of `int`?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1427,6 +1469,38 @@ int MDMParser::_cbCCID(int type, const char* buf, int len, CStringHelper* str)
...
+    }
+
+    if (_lastVerboseC_regUpdate &&
+        (HAL_Timer_Get_Milli_Seconds() - _lastVerboseC_regUpdate) >= MDM_C_REG_POLL_INTERVAL_MS)
+    {
+        LOCK();
+        if (!_checkModem()) {
+            return;
+        }
+        _lastVerboseC_regUpdate = HAL_Timer_Get_Milli_Seconds();
```
keeramis, line 1484: The following AT commands `AT+C*REG?` have 60 sec timeout. Might it be good to set this _after_ getting a response from the commands below, so that we will actually wait `MDM_C_REG_POLL_INTERVAL_MS` before the next call?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -418,34 +426,38 @@ int MDMParser::process() {
 
     {
         LOCK();
-        if (!_checkModem(false /* force */)) {
+        if (_reinitModem == 0x55 || !_checkModem(false /* force */)) {
+            if (_reinitModem != 0x55) {
+                MDM_ERROR("[ Modem not responsive ]\r\n");
```
keeramis, line 431: May be Nit: Looks like now `_checkModem()` failure will not have any log statement?.... Might be good to have similar log statement if `_checkModem()` fails also?

hal/src/electron/modem/mdm_hal.h
technobly: I made this toggle multiple bits because it's a sensitive operation (re-init modem).  So it's at least somewhat tolerant to cosmic rays toggling bits :D

hal/src/electron/modem/mdm_hal.cpp
technobly: This timer is arbitrarily set to 30 seconds, just to be kind of fast for recovery, but not super duper noisy.  I would not want to set this for the first time after any C*REG response that might never come before a brown out/reset occurs.  This timer can just run asynchronously in the background forever not governed by anything but itself... which will keep it running despite weird edge cases we don't predict now.

hal/src/electron/modem/mdm_hal.cpp
technobly: _checkModem()'s log is "[ Modem not responsive ]", which is entry on != 0x55.  _reinitModem's logs happens in waitFinalResp().

hal/src/electron/modem/mdm_hal.cpp
technobly: We also are guaranteeing recovery within 2 minutes from these failures, by way of CELLUAR_03/04 test assertions. We could better ensure that 2 min by making C*REG polling faster if necessary.  I was seeing recovery in about 60 seconds though, so I think we have a decent margin already.

keeramis: I believe I was referring to the edge-case where the response for `AT+CEREG?` takes a long time like 30 or 40 sec, in which case we will be querying them back to back, because `MDM_C_REG_POLL_INTERVAL_MS` would have elapsed by then. Again, it's a weird edge case that you have a valid point not to account for. 👍

keeramis (approved):

hal/src/electron/modem/mdm_hal.cpp
technobly: This _checkModem() with AT/OK check call will be like 99% defense against calling the C*REG? commands with 60 second timeout when the AT interface is dead.  If the AT interface dies right when C*REG is called, it will take 60 seconds... but then the next _checkModem call will cause it to fail completely as two _error's will have accrue quickly with _checkModem, partially because _lastVerboseC_regUpdate is set after that so that it gets called again right away and fails.

I see what you mean now though.  Typically C*REG calls will be fast.  But if they timeout as mentioned above (after an successful AT/OK check) and we have to wait 60s... I wouldn't want to wait another 30s before the double _checkModem call.  It's good to think these different test cases though, so I appreciate the suggestion :D

keeramis: Cool - thanks for helping understand!..

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -567,13 +579,14 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                         HAL_NET_notify_dhcp(false);
                     }
                 } else {
-                    // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG
-                    // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC
-                    b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1;
-                    r = sscanf(cmd, "%31s %*d,%d,\"%x\",\"%x\",%d",s,&a,&b,&c,&d);
+                    // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG (2,4,5,6 results)
+                    // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC (1,3,4,5 results)
+                    b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1; mode = 2;
```
avtolstoy, line 584: A comment why specifically we default to 2 would be appreciated. If there is no specific reason, we should probably default to something like `-1`.

user/tests/wiring/no_fixture_long_running/cellular.cpp
```diff
@@ -114,7 +114,55 @@ test(CELLULAR_02_device_will_connect_to_the_cloud_when_all_udp_sockets_consumed)
...
+    // Serial.println("the device will reconnect to the cloud when unexpected modem brown out occurs within 2min");
+    // Given the device is currently connected to the Cloud
+    connect_to_cloud(6*60*1000);
+    // When the device modem suffers a simulated brown out (modem reset and powered back on with no initialization)
+    digitalWrite(RESET_UC, 0);
+    delay(10000);
+    digitalWrite(RESET_UC, 1);
+    do {
+        // SARA-U2/LISA-U2 50..80us
+        digitalWrite(PWR_UC, 0); delay(50);
```
avtolstoy, line 133: Nit: single statement on a single line. It's very easy to miss things with this pattern.

avtolstoy (changes requested):

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1453,6 +1495,38 @@ int MDMParser::_cbCCID(int type, const char* buf, int len, CStringHelper* str)
...
+        (HAL_Timer_Get_Milli_Seconds() - _lastVerboseC_regUpdate) >= MDM_C_REG_POLL_INTERVAL_MS)
+    {
+        LOCK();
+        if (!_checkModem()) {
+            return;
+        }
+        _lastVerboseC_regUpdate = HAL_Timer_Get_Milli_Seconds();
+        // Check if C*REG URCs are enabled, handled in waitFinalResp()
+        if (_dev.dev == DEV_SARA_R410) {
+            if (_verboseCeregEnabled) {
```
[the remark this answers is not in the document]
technobly: The flags let us know that the C*REG setting should be expected to be set to verbose mode, so we don't start polling and reset the modem if not expected to be in verbose mode.

hal/src/electron/modem/mdm_hal.h
technobly: It's not a joke... plenty of critical operations use multi-bit/magic numbers.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -586,13 +599,32 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
...
+                                MDM_ERROR("[ Modem unexpectedly reset ]\r\n");
+                                _lastVerboseC_regUpdate = 0;
+                                _verboseCregEnabled = false;
+                                _verboseCgregEnabled = false;
+                                _verboseCeregEnabled = false;
+                                // reinit handled in MDMParser::process()
+                            } else if (_dev.dev == DEV_SARA_R410 && _attached && !REG_OK(*reg) && !strcmp(s, "CEREG:")) {
+                                // R410 : Look out for a CEREG: de-registration URC while attached
+                                MDM_ERROR("[ Cell de-registered ]\r\n");
+                                _reinitModem = 0x55;
```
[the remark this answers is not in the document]
technobly: At the time I didn't see a difference in what was triggered between de-registration and c*reg verbose mode being disabled.  I'll test again to see if there is a difference and change it if so.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
avtolstoy: I see. And we cannot really use something like `_activated` or `_attached`. Still, let's simplify and use just 1 variable to check that we've configured all C*REG to verbose mode.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -265,11 +270,13 @@ MDMParser::MDMParser(void)
     _oldFirmwarePresent = true; // default to safe state until we determine modem firmware version
     _timePowerOn = 0;
     _timeRegistered = 0;
-    memset(_sockets, 0, sizeof(_sockets));
-    for (int socket = 0; socket < NUMSOCKETS; socket ++)
-        _sockets[socket].handle = MDM_SOCKET_ERROR;
-    _error = 0;
+    _lastVerboseC_regUpdate = 0;
```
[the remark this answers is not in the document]
technobly: It seemed clear for me, but I'll change it if it helps you.

hal/src/electron/modem/mdm_hal.cpp
technobly: -1 would be even safer... so I'll go that route.

hal/src/electron/modem/mdm_hal.h
avtolstoy: We have a lot of other critical pieces of code, why only here specifically? And I would argue if we are worried about something flipping bits in SRAM, we have bigger problems on our hands.

technobly: I would prefer to have a few extra bits in between power cycling the modem and normal operation.  I think it's a reasonable thing to add.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
technobly: I changed it back.  When I was originally testing I must have been catching some case that looked like it was power cycling the modem anyway?

user/tests/wiring/no_fixture_long_running/cellular.cpp
```diff
@@ -114,7 +114,59 @@ test(CELLULAR_02_device_will_connect_to_the_cloud_when_all_udp_sockets_consumed)
...
+    skip();
+    return;
+#endif
+    // Serial.println("the device will reconnect to the cloud when unexpected modem reset occurs within 2min");
+    // Given the device is currently connected to the Cloud
+    connect_to_cloud(6*60*1000);
+    // When the device modem suffers a simulated brown out (modem reset and powered back on with no initialization)
+    digitalWrite(RESET_UC, 0);
+    delay(10000);
+    digitalWrite(RESET_UC, 1);
```
avtolstoy, line 163: For some reason a comment got lost, so I'll repeat just in case.

Making sure that we are testing for the right thing: SARA R4 will not boot back up if it's reset through `RESET` pin. SARA U2 will boot up automatically.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -586,12 +596,28 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
...
+                            if (_reinitModem == 0x55) {
+                                MDM_ERROR("[ Modem unexpectedly reset ]\r\n");
+                                _lastVerboseCxregUpdate = 0;
+                                // reinit handled in MDMParser::process()
+                            } else if (_dev.dev == DEV_SARA_R410 && _attached && !REG_OK(*reg) && !strcmp(s, "CEREG:")) {
+                                // R410 : Look out for a CEREG: de-registration URC while attached
+                                MDM_ERROR("[ Cell de-registered ]\r\n");
+                                _ip = NOIP; // Short-circuit disconnect() call
                                 _attached = false;
+                                _activated = false;
```
avtolstoy, line 620: This wasn't here originally. This change will break `reconnect()` call in `process()`.

avtolstoy (changes requested):

hal/src/electron/modem/mdm_hal.h
avtolstoy: @technobly Let's not overcomplicate things. These magic numbers don't protect us against anything. If we are that worried about errors in SRAM, going by the same logic the same thing may happen:
1. In register that holds the value you are comparing with
2. While a constant is being loaded from flash into cache
3. In cache
4. etc etc etc

hal/src/electron/modem/mdm_hal.cpp
technobly: My bad 👍

user/tests/wiring/no_fixture_long_running/cellular.cpp
technobly: If you look a few lines below this it powers on the modem again.  Perhaps not required for U2, but just in case powering on won't hurt.

avtolstoy: That's a different test though. There is a power-on sequence in `CELLULAR_03_device_will_reconnect_to_the_cloud_when_unexpected_modem_brown_out_occurs_within_2min`.

technobly: Gotta fix that comment about brown out on this one.  I'd like to test the case where we power off unexpectedly, and reset was the way to get there easily on R410.  Perhaps I can change this to just power off the modem via the PWR pin.

hal/src/electron/modem/mdm_hal.h
technobly: I ended up implementing your suggestion of using `_error` and made the code a bit more safe by changing the number of `_error`'s that would cause a modem reinit from a magic number of 2, to a define.

avtolstoy (approved):

Issue #1932, opened 2019-10-01
attachInterrupt() fails for pin D7 on Boron


Otterverse:
### Expected Behavior

attachInterrupt(D7, myFunc, RISING) == true;

### Observed Behavior

attachInterrupt(D7, myFunc, RISING) == false;

### Steps to Reproduce

Using firmware 1.4.0, attempt to attach an interrupt to D7 on a Boron device and it will return false.

I'm almost positive this is due to legacy (electron specific if my research is correct) code in spark_wiring_interrupts.cpp which has this catch in the attachInterrupt() definition:

```
#if Wiring_Cellular == 1
  /* safety check that prevents users from attaching an interrupt to D7
   * which is shared with BATT_INT_PC13 for power management */
  if (pin == D7) return false;
#endif 
```
And while the docs for the Electron clearly label this pin as off limits for that device, the docs for Boron state:

```
All A and D pins (including TX, RX, and SPI) on Gen 3 (mesh) devices can be used for interrupts...
```
I'm fairly certain Wiring_Cellular still gets set on the Boron though.

avtolstoy:
Thanks for submitting the issue! Yes, this is an oversight and D7 should be available for `attachInterrupt` on Gen3 devices. This should be fixed before 1.4.1 default.

Otterverse:
Wow! That is the fastest response I've ever had to a bug report! Thank you, I'll look forward to the update.

technobly:
D7 is excluded on Gen 2 because of the shared interrupt with USART 3 or PMIC used for connection the the modem on Electron. Just wanted to double check that D7 is not shared on Gen 3 in any similar way.

PR #2220, opened 2020-10-15
Fixes the issue that BLE.scan() might hang the device.

XuGuohui:
### Problem

BLE may hang the thread that calls `BLE.scan()` due to unexpected scanning failure. Since the `BLE.scan()` is a blocking API waiting until a semaphore is given on scanning timeout event, in some condition the scanning procedure never timeout or a incorrect event is delivered from SoftDevice on scanning failure, in which case the semaphore cannot be taken any more.

### Solution

Introduce a timer to guard the scanning procedure. For example, if the scanning timeout is set to 5s and it starts scanning, the timer will expire after 6s to stop scanning actively. Normally BLE scanning can timeout by SoftDevice and the timer will be stopped by then.

### References

#1932 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -335,7 +335,8 @@ public:
               isScanning_(false),
               scanSemaphore_(nullptr),
               scanResultCallback_(nullptr),
-              context_(nullptr) {
+              context_(nullptr),
+              scanGuardTimer_(nullptr) {
         scanParams_.version = BLE_API_VERSION;
         scanParams_.size = sizeof(hal_ble_scan_params_t);
         scanParams_.active = true;
@@ -371,6 +372,7 @@ private:
     int constructObserverEvent(hal_ble_scan_result_evt_t& result, const ble_gap_evt_adv_report_t& advReport) const;
     void notifyScanResultEvent(const hal_ble_scan_result_evt_t& result);
     static void processObserverEvents(const ble_evt_t* event, void* context);
+    static void onScanGuardTimerExpired(os_timer_t timer);
 
     bool observerInitialized_;
     volatile bool isScanning_;                              /**< If it is scanning or not. */
@@ -380,6 +382,7 @@ private:
     ble_data_t bleScanData_;                                /**< BLE scanned data. */
     hal_ble_on_scan_result_cb_t scanResultCallback_;        /**< Callback function on scan result. */
     void* context_;                                         /**< Context of the scan result callback function. */
+    os_timer_t scanGuardTimer_;                             /**< Timer to guard the scanning procedure is terminated successfully.  */
     Vector<hal_ble_addr_t> cachedDevices_;
     Vector<hal_ble_scan_result_evt_t> pendingResults_;
 };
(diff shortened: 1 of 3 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1354,6 +1389,9 @@ int BleObject::Observer::stopScanning() {
     if (give) {
         os_semaphore_give(scanSemaphore_, false);
     }
+    if (os_timer_is_active(scanGuardTimer_, nullptr)) {
```
avtolstoy, line 1392: Stop before unblocking the thread that requested scanning?

hal/src/nRF52840/ble_hal_impl.h
```diff
@@ -68,9 +68,12 @@
 #define BLE_MAX_TX_POWER                            (8)
 
 /* Default scanning parameters */
+#define BLE_DEFAULT_SCANNING_TIMEOUT_MS             5000
 #define BLE_DEFAULT_SCANNING_INTERVAL               BLE_MSEC_TO_UNITS(100, BLE_UNIT_0_625_MS)   /* The scan interval: 100ms (in units of 0.625 ms). */
 #define BLE_DEFAULT_SCANNING_WINDOW                 BLE_MSEC_TO_UNITS(50, BLE_UNIT_0_625_MS)    /* The scan window: 50ms (in units of 0.625 ms). */
-#define BLE_DEFAULT_SCANNING_TIMEOUT                BLE_MSEC_TO_UNITS(5000, BLE_UNIT_10_MS)     /* The timeout: 5000ms (in units of 10 ms. 0 for scanning forever). */
+#define BLE_DEFAULT_SCANNING_TIMEOUT                BLE_MSEC_TO_UNITS(BLE_DEFAULT_SCANNING_TIMEOUT_MS, BLE_UNIT_10_MS)     /* The timeout: 5000ms (in units of 10 ms. 0 for scanning forever). */
+// Extended timeout for the scanning timeout guard timer
+#define BLE_SCANNING_TIMEOUT_EXT                    1000
```
avtolstoy, line 76: Let's clarify that this is milliseconds as we are dealing with various time units here.

avtolstoy (approved):
@XuGuohui Can we add a synthetic test for this? For example block the BLE thread or block BLE interrupt to simulate this behavior?

hal/src/nRF52840/ble_hal.cpp
XuGuohui: The timer is stopped on entering the function.

hal/src/nRF52840/ble_hal_impl.h
XuGuohui: Changed to `BLE_SCANNING_TIMEOUT_EXT_MS`.

XuGuohui:
> @XuGuohui Can we add a synthetic test for this? For example block the BLE thread or block BLE interrupt to simulate this behavior?

A test is added to `user/tests/wiring/ble_scanner_broadcaster/ble_scanner`. And I ran it manually with success.

hal/src/nRF52840/ble_hal.cpp
avtolstoy: Can you point to where that happens?

XuGuohui: I meant I just moved it up in the same function ahead of giving the scan semaphore.

Issue #2005, opened 2020-01-29
BleAddress could do with a way to check validity and also invalidate an object


ScruffR:
While `BleUuid` features an `isValid()` and `BleCharacteristic` a `valid()` method I was surprised to find that `BleAddress` does *not*.
It would be good practice to have a *consistent* way to check validity *and also* a simple and consistent way to _invalidate_ and object (e.g. like `IPAddress::clear()`)

When I say *"consistent"*, I'm thinking of `BleUuid::isValid()` vs. `BleCharacteristic::valid()` - IMO that is *very poor* style and suggests there were two individuals working on the same project but not speaking with eachother (that reminds me of [this](https://github.com/particle-iot/device-os/issues/1711) blunder with the EtherWing design where obviously the context and preexisting devices were *not* considered in the planning process).
Or if it was only a single coder I'd have that looked at :wink:

ScruffR:
Bump!

avtolstoy:
Fixed in #2222

PR #2222, opened 2020-10-17
Wiring BLE APIs consistency improvements

XuGuohui:
**This PR is targeting to feature/ip_ble_address**

### Problem

1. `BleAddress` doesn't support validity check and invalidating object.
2. Scanning callback doesn't pass scanned result in reference manner.
3. `BleScanResult` should not access members directly

### References

https://github.com/particle-iot/device-os/issues/2005

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -3750,6 +3750,7 @@ int hal_ble_gap_connect_cancel(const hal_ble_addr_t* address, void* reserved) {
 }
 
 int hal_ble_gap_disconnect(hal_ble_conn_handle_t conn_handle, void* reserved) {
+    BleLock lk;
     LOG_DEBUG(TRACE, "hal_ble_gap_disconnect().");
     CHECK_TRUE(BleObject::getInstance().initialized(), SYSTEM_ERROR_INVALID_STATE);
     hal_ble_conn_info_t info = {};
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -3750,7 +3750,6 @@ int hal_ble_gap_connect_cancel(const hal_ble_addr_t* address, void* reserved) {
 }
 
 int hal_ble_gap_disconnect(hal_ble_conn_handle_t conn_handle, void* reserved) {
-    BleLock lk;
     LOG_DEBUG(TRACE, "hal_ble_gap_disconnect().");
     CHECK_TRUE(BleObject::getInstance().initialized(), SYSTEM_ERROR_INVALID_STATE);
     hal_ble_conn_info_t info = {};
--- a/wiring/inc/spark_wiring_ble.h
+++ b/wiring/inc/spark_wiring_ble.h
@@ -146,6 +146,7 @@ typedef hal_ble_attr_handle_t BleAttributeHandle;
 
 typedef void (*BleOnDataReceivedCallback)(const uint8_t* data, size_t len, const BlePeerDevice& peer, void* context);
 typedef void (*BleOnScanResultCallback)(const BleScanResult* device, void* context);
+typedef void (*BleOnScanResultCallbackRef)(const BleScanResult& device, void* context);
 typedef void (*BleOnConnectedCallback)(const BlePeerDevice& peer, void* context);
 typedef void (*BleOnDisconnectedCallback)(const BlePeerDevice& peer, void* context);
 
@@ -577,6 +578,7 @@ public:
 
     // Scanning control
     int scan(BleOnScanResultCallback callback, void* context) const;
+    int scan(BleOnScanResultCallbackRef callback, void* context) const;
     int scan(BleScanResult* results, size_t resultCount) const;
     Vector<BleScanResult> scan() const;
     int stopScanning() const;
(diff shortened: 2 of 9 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -3709,7 +3709,6 @@ int hal_ble_gap_connect_cancel(const hal_ble_addr_t* address, void* reserved) {
 }
 
 int hal_ble_gap_disconnect(hal_ble_conn_handle_t conn_handle, void* reserved) {
-    BleLock lk;
```
avtolstoy, line 3712: Wouldn't this introduces thread-safety issues when accessing/traversing `connections_` ? I also saw that we specifically release wiring lock in `spark_wiring_ble.cpp` before calling `hal_ble_gap_disconnect()`, so any deadlocking issues should already be mitigated?

wiring/inc/spark_wiring_ble.h
```diff
@@ -219,6 +220,9 @@ class BleAddress {
     bool operator!=(const char* address) const;
     bool operator!=(const String& address) const;
 
+    bool valid() const;
```
avtolstoy, line 223: Let's be consistent in our wiring APIs. This should be `isValid()`. Optionally operator bool may also be added.

wiring/inc/spark_wiring_ble.h
```diff
@@ -238,7 +242,8 @@ class BleUuid {
     BleUuid(const char* uuid);
     ~BleUuid() = default;
 
-    bool isValid() const;
+    bool __attribute__((deprecated("Use BleUuid::valid() instead"))) isValid() const;
```
avtolstoy, line 245: Same here. `isValid()` is the correct name.

wiring/inc/spark_wiring_ble.h
```diff
@@ -508,7 +614,10 @@ class BlePeerDevice {
     void bind(const BleAddress& address) const;
     BleAddress address() const;
 
+    bool valid() const;
```
avtolstoy, line 617: `isValid()`.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -700,34 +748,58 @@ String BleAdvertisingData::deviceName() const {
...
+    return selfData_[i];
+}
+
 bool BleAdvertisingData::contains(BleAdvertisingDataType type) const {
     size_t adsOffset;
     return locate(selfData_, selfLen_, type, &adsOffset) > 0;
 }
 
-size_t BleAdvertisingData::serviceUUID(BleAdvertisingDataType type, BleUuid* uuids, size_t count) const {
+size_t BleAdvertisingData::serviceUUID(BleAdvertisingDataType type, Vector<BleUuid>& uuids) const {
```
avtolstoy, line 787: Nit: We can probably just return `Vector<BleUuid>` here. No need to separately track `found`. The vector has the size.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1292,6 +1370,10 @@ bool BleService::operator==(const BleService& service) const {
     return (impl()->UUID() == service.impl()->UUID());
 }
 
+bool BleService::operator!=(const BleService& service) const {
+    return (impl()->UUID() != service.impl()->UUID());
```
avtolstoy, line 1374: Or just `return !(*this == service)` ?

avtolstoy (changes requested):
Let's make sure to make a draft of all the API changes so that they can be properly documented. Ideally this should be a CH story.

hal/src/nRF52840/ble_hal.cpp
XuGuohui: You're right. We need to guard the `connections` in BLE HAL. I reverted this change and added a FIXME in `onBleLinkEvents()` in the wiring to make us be aware of a potential deadlock.

wiring/inc/spark_wiring_ble.h
XuGuohui: Updated and added.

avtolstoy (approved):

wiring/src/spark_wiring_ble.cpp
```diff
@@ -1101,6 +1096,10 @@ class BleLocalDeviceImpl {
                 peer.impl()->address() = event->params.connected.info->address;
                 if (!impl->peers_.append(peer)) {
                     LOG(ERROR, "Failed to append peer Central device.");
+                    // FIXME: It will acquire the BLE HAL lock. If there is a thread currently invoking a
```
avtolstoy, line 1099: If that's the only case can we perhaps return an error from this event handler and let BLE HAL handle this appropriately within its loop or something like that?

XuGuohui: That could be an approach. I'd like to leave it as-is, because the dead lock just happens with a critical use case and it's not part of this PR. This is more like a PR that we need to refactor the BLE HAL and implement non-blocking BLE APIs.

avtolstoy: Yeah, let's leave as-is as this condition shouldn't normally happen, still let's keep a story open, so that we can come of with some kind of a way to deal with these types of situation.

PR #2225, opened 2020-10-19
Export some common stdlib functions from system parts

avtolstoy:
### Problem

1. Some stdlib functions are commonly used between system and user applications. They can be exported from the system through dynalib to avoid code duplication.
2. [wiring] Mutex/RecursiveMutex classes don't follow [Lockable](https://en.cppreference.com/w/cpp/named_req/Lockable) spec

**See [CH65545] for more information and some charts.**

### Solution

1. Move some of them into `rt_dynalib`. On Electron import `services2` dynalib into user application, as they are already exported from there.
2. Add `try_lock()` method to them

### Steps to Test

See CH.

### Example App

N/A

### References

- [CH65545]

---

### Completeness

- [ ] Added documentation


rt-dynalib/inc/rt_dynalib.h
```diff
@@ -59,6 +64,14 @@ DYNALIB_FN(15, rt, __errno, int*())
 #if defined(DYNALIB_EXPORT) || (defined(DYNALIB_IMPORT) && !defined(RT_DYNALIB_NO_DEPENDENCY_BREAKING_IMPORTS))
 DYNALIB_FN(16, rt, __assert_func, void(const char*, int, const char*, const char*))
 DYNALIB_FN(17, rt, newlib_impure_ptr_callback, void(void (*)(struct _reent*, size_t, uint32_t, void*), void*))
+// Electron has some of these exported from a different system part through services2 dynalib
+// Photon/P1 do not have enough space in system-part2
+#if HAL_PLATFORM_EXPORT_STDLIB_RT_DYNALIB
```
sergeuz, line 69: Defining this as a platform feature macro feels a bit weird to me. Perhaps we can define it as a platform-specific hack similarly to `RT_DYNALIB_NO_DEPENDENCY_BREAKING_IMPORTS`?

sergeuz (approved):

rt-dynalib/inc/rt_dynalib.h
avtolstoy: The only problem with that is that we'll have to specify this quirk both when exporting and importing.

sergeuz: ah, I see. Please ignore this comment then :+1:

PR #2227, opened 2020-10-27
[hal] sleep: restore waleup sources' interrupt priority appropriately

XuGuohui:
## Problem

A wakeup source's interrupt priority is increased and never restored if it is duplicated in the wakeup source list or shares the same IRQ with other wakeup sources. Especially on Gen3, for critical SoftDevice region, a user application interrupt that would result in an error in SoftDevice.

### Solution

Bump a wakeup source's interrupt priority only if it hasn't been bumped previously.

### References

N/A

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/sleep_hal.cpp
+++ b/hal/src/nRF52840/sleep_hal.cpp
@@ -49,43 +49,81 @@
 #include "spark_wiring_vector.h"
 
 
-using namespace particle;
+using namespace particle;;
+
+namespace {
 
 typedef struct WakeupSourcePriorityCache {
-    uint32_t gpiotePriority;
-    uint32_t rtc2Priority;
-    uint32_t blePriority;
-    uint32_t lpcompPriority;
-    uint32_t usart0Priority;
-    uint32_t usart1Priority;
+    uint16_t priority;
+    uint16_t bumped;
 } WakeupSourcePriorityCache;
 
-static void bumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources, WakeupSourcePriorityCache* priority, uint32_t newPriority) {
+WakeupSourcePriorityCache gpiotePriority;
+WakeupSourcePriorityCache rtc2Priority;
+WakeupSourcePriorityCache blePriority;
+WakeupSourcePriorityCache lpcompPriority;
+WakeupSourcePriorityCache usart0Priority;
+WakeupSourcePriorityCache usart1Priority;
+
+}
+
+static void bumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources, uint32_t newPriority) {
     auto source = wakeupSources;
     while (source) {
         if (source->type == HAL_WAKEUP_SOURCE_TYPE_GPIO) {
-            priority->gpiotePriority = NVIC_GetPriority(GPIOTE_IRQn);
-            NVIC_SetPriority(GPIOTE_IRQn, newPriority);
+            if (!gpiotePriority.bumped) {
+                gpiotePriority.priority = NVIC_GetPriority(GPIOTE_IRQn);
+                gpiotePriority.bumped = true;
+            }
+            // The lower number, the higher priority
+            if (newPriority < NVIC_GetPriority(GPIOTE_IRQn)) {
+                NVIC_SetPriority(GPIOTE_IRQn, newPriority);
+            }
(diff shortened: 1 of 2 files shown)


hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -49,59 +49,106 @@
...
             }
         }
         source = source->next;
     }
 }
 
-static void unbumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources, const WakeupSourcePriorityCache* priority) {
+static void unbumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources) {
     auto source = wakeupSources;
     while (source) {
```
avtolstoy, line 136: There is probably no need now to walk the list. We can just check the appropriate `bumped` field.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -49,59 +49,106 @@
...
-    uint32_t blePriority;
-    uint32_t lpcompPriority;
-    uint32_t usart0Priority;
-    uint32_t usart1Priority;
+    uint16_t priority;
+    uint16_t bumped;
 } WakeupSourcePriorityCache;
 
-static void bumpWakeupSourcesPriority(const hal_wakeup_source_base_t* wakeupSources, WakeupSourcePriorityCache* priority, uint32_t newPriority) {
+WakeupSourcePriorityCache gpiotePriority;
```
avtolstoy, line 61: These can still be all hidden within `WakeupSourcePriorityCache` and allocated on stack, as they are unused 99% of the time.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -55,6 +55,33 @@ static constexpr uint8_t GPIO_IRQn[extiChannelNum] = {
...
 
+uint8_t exti0Bumped = false;
+uint8_t exti1Bumped = false;
+uint8_t exti2Bumped = false;
+uint8_t exti3Bumped = false;
+uint8_t exti4Bumped = false;
+uint8_t exti95Bumped = false;
+uint8_t exti1510Bumped = false;
+
+static uint8_t* const pExtiBump[extiChannelNum] = {
```
avtolstoy, line 66: Or just simplify this and use the fact that this is 0-15 and use a bitmask. There is no need to introduce two levels of indirection.

avtolstoy (approved):
@XuGuohui I'm approving this, but please look through the comments.

hal/src/nRF52840/sleep_hal.cpp
XuGuohui: I'd persist to traverse the list, as in long term for certain wakeup source it might not need to be bumped the priority if its normal priority is sufficient to wake up device. And for certain wakeup sources we are not just unbumping the priority, but also do something specific to the peripheral. If we depends on the bump flag, we will miss something.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -49,8 +49,12 @@
 #include "spark_wiring_vector.h"
 
 
+#define INVALID_INT_PRIORITY        (0xFFFFFFFF)
```
avtolstoy, line 52: Nit: for C++ prefer to use constants instead of defines.

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -832,7 +869,8 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
         __disable_irq();
 
         // Bump the priority
-        WakeupSourcePriorityCache priorityCache = {};
+        WakeupSourcePriorityCache priorityCache;
+        memset((uint8_t*)&priorityCache, 0xFF, sizeof(WakeupSourcePriorityCache));
```
avtolstoy, line 873: Nit: prefer to initialize things in the constructor for C++ structs.

hal/src/stm32f2xx/sleep_hal.cpp
```diff
@@ -31,6 +31,9 @@
 #include "check.h"
 
 
+#define EXTI9_5_BITS_MASK       (0x03E0)
```
avtolstoy, line 34: Nit: prefer to use constants in C++ code instead of defines.

avtolstoy (approved):

PR #2228, opened 2020-10-27
[wiring] system: configure multiple pins as wakeup source at a time.

XuGuohui:
As the title describes.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {
    delay(300);
    Vector<std::pair<pin_t, InterruptMode>> pins;
    pins.append(std::make_pair(D0, FALLING));
    pins.append(std::make_pair(D1, RISING));

    Vector<pin_t> altPins;
    altPins.append(D2);
    altPins.append(D3);

    uint8_t pinArray[2] = {D4, D5};

    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER)
                                           .gpios(pins)
                                           .gpios(altPins, FALLING)
                                           .gpios(pinArray, sizeof(pinArray), RISING));
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/inc/system_sleep_configuration.h
+++ b/system/inc/system_sleep_configuration.h
@@ -26,9 +26,12 @@
 #include "system_error.h"
 #include "spark_wiring_network.h"
 #include "spark_wiring_usartserial.h"
+#include "spark_wiring_vector.h"
 #include "enumflags.h"
 
 #define SYSTEM_SLEEP_NETWORK_FLAG_SUPPORTED_VER     (3)
+
+using spark::Vector;
 
 namespace particle {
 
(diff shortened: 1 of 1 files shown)


system/inc/system_sleep_configuration.h
```diff
@@ -253,6 +256,27 @@ class SystemSleepConfiguration: protected SystemSleepConfigurationHelper {
         return *this;
     }
 
+    SystemSleepConfiguration& gpios(const Vector<std::pair<pin_t, InterruptMode>>& pins) {
```
eberseth, line 259: In case of an existing entry do you just update the interrupt mode?  Take a look at https://github.com/particle-iot/tracker-edge/blob/develop/src/tracker_sleep.cpp#L115-L126 for reference

XuGuohui: The `gpio(pin_t pin, InterruptMode mode)` overload will take care of that: https://github.com/particle-iot/device-os/blob/feature/sleep_pin_list/ch60648/system/inc/system_sleep_configuration.h#L236

eberseth: Great!

eberseth (approved):
Nice!

PR #2229, opened 2020-10-28
[hal] electron: use AT+CPWROFF to turn off the modem completely

XuGuohui:
### Problem

The Electron with SARA-G350 doesn't support powering off using the PWR_ON pin. The low level pulse of the PWR_ON pin being performed in current powerOff() will actually power on the modem.

### Solution

Use `AT+CPWROFF` completely to power off the modem.

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1282,7 +1282,7 @@ bool MDMParser::softPowerOff(void) {
         }
     }
     for (uint8_t i = 0; i < 3; i++) {
-        if (!_error && _atOk()) {
+        if (_atOk()) {
             // We can talk to the modem getting here.
             MDM_INFO("\r\n[ Modem::softPowerOff ] Powering down the modem using AT+CPWROFF....");
             sendFormated("AT+CPWROFF\r\n");
@@ -1347,7 +1347,7 @@ bool MDMParser::powerOff(void)
             }
         }
         // Skip power off sequence if power is already off
-        if (powerState() && _dev.dev != DEV_SARA_G350 && _dev.dev != DEV_UNKNOWN) {
+        if (powerState() && _dev.dev != DEV_SARA_G350) {
             MDM_INFO("%s Modem not responsive, trying PWR_UC...", POWER_OFF_MSG);
             HAL_GPIO_Write(PWR_UC, 0);
             // >1.5 seconds on SARA R410M
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -828,29 +828,19 @@ bool MDMParser::_powerOn(void)
         _init = true;
     }
 
-    MDM_INFO("\r\n[ Modem::_powerOn ] = = = = = = = = = = = = = =");
+    MDM_INFO("\r\n[ Modem::powerOn ] = = = = = = = = = = = = = =");
     bool continue_cancel = false;
     bool retried_after_reset = false;
 
     int i = 10;
     while (i--) {
-        if (powerState()) {
-            MDM_INFO("\r\n[ Modem::_powerOn ] Modem is already on");
-        } else {
-            // SARA-U2/LISA-U2 50..80us
-            HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(50);
-            HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(10);
-            if (!powerState()) {
-                // SARA-G35 >5ms, LISA-C2 > 150ms, LEON-G2 >5ms, SARA-R4 >= 150ms
-                HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(150);
-                HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(100);
-            }
-            if (powerState()) {
-                MDM_INFO("\r\n[ Modem::_powerOn ] Modem is powered on");
-            } else {
-                MDM_ERROR("\r\n[ Modem::_powerOn ] Failed to power on the modem.");
-            }
-        }
+        // SARA-U2/LISA-U2 50..80us
+        HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(50);
+        HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(10);
+
+        // SARA-G35 >5ms, LISA-C2 > 150ms, LEON-G2 >5ms, SARA-R4 >= 150ms
+        HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(150);
+        HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(100);
 
         // purge any messages
         purge();
@@ -881,6 +871,7 @@ bool MDMParser::_powerOn(void)
             i = 10;
             reset();
         }
+
     }
     if (i < 0) {
         MDM_ERROR("[ No Reply from Modem ]\r\n");
@@ -969,20 +960,13 @@ bool MDMParser::powerOn(const char* simpin)
             goto failure;
         }
         else if (i==4 && (RESP_OK != ret) && !retried_after_reset) {
-            if (_dev.sim == SIM_MISSING) {
-                // No need to try again in a short time.
+            retried_after_reset = true; // only perform reset & retry sequence once
+            i = 0;
+            if(!powerOff())
+                reset();
+            /* Power on the modem and perform basic initialization again */
+            if (!_powerOn())
                 goto failure;
-            } else {
-                retried_after_reset = true; // only perform reset & retry sequence once
-                i = 0;
-                if(!powerOff()) {
-                    reset();
-                }
-                /* Power on the modem and perform basic initialization again */
-                if (!_powerOn()) {
-                    goto failure;
-                }
-            }
         }
         // Enter PIN if needed
         if (_dev.sim == SIM_PIN) {
@@ -1279,12 +1263,12 @@ bool MDMParser::powerState(void) const {
     return state;
 }
 
-bool MDMParser::powerOnOffGracefully(void) {
+bool MDMParser::softPowerOff(void) {
     if (!powerState()) {
         return true;
     }
     if (!(_init && _pwr)) {
-        MDM_INFO("\r\n[ Modem::powerOnOffGracefully ] Perform low level initialization to talk to the modem...");
+        MDM_INFO("\r\n[ Modem::softPowerOff ] Perform low level initialization to talk to the modem...");
         uint8_t i = 0;
         for (i = 0; i < 3; i++) {
             if (_powerOn()) {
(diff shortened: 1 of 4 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1271,85 +1290,90 @@ bool MDMParser::powerOff(void)
...
+    if (!powerState() && !_init && !_pwr) {
+        MDM_INFO("%s Modem is off already", POWER_OFF_MSG);
+        return true;
+    }
+
+    // Make sure we can use the AT parser.
+    // For SARA-G35 it doesn't support powering off the modem using the PWR_ON pin.
+    // But it shares the same code base with other modem, thus, we use AT+CPWROFF generally.
+    if (!(_init && _pwr)) {
+        MDM_INFO("%s Powering on to use AT parser...", POWER_OFF_MSG);
```
avtolstoy, line 1305: We are not really powering on here, we are trying to do some low level initialization to talk to the modem.

avtolstoy (changes requested):

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -960,13 +972,20 @@ bool MDMParser::powerOn(const char* simpin)
             goto failure;
         }
         else if (i==4 && (RESP_OK != ret) && !retried_after_reset) {
-            retried_after_reset = true; // only perform reset & retry sequence once
-            i = 0;
-            if(!powerOff())
-                reset();
-            /* Power on the modem and perform basic initialization again */
-            if (!_powerOn())
+            if (_dev.sim == SIM_MISSING) {
```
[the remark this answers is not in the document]
XuGuohui: On Electron it uses physical SIM. If `_powerOn()` succeeded, which means that the modem is responsive well enough, while it just doesn't detect a physical SIM. In this case, even if we perform a `powerOff()` and then `reset()` and then `powerOn()`, the physical SIM is not there still. Unless user can insert a physical SIM when **seeing** the log in real real quick, then the reset may help, otherwise, I think it's just a waste of time that prevents the device from doing other things afterward and the log is messy.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -828,19 +828,29 @@ bool MDMParser::_powerOn(void)
...
+            HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(10);
+            if (!powerState()) {
+                // SARA-G35 >5ms, LISA-C2 > 150ms, LEON-G2 >5ms, SARA-R4 >= 150ms
+                HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(150);
+                HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(100);
+            }
+            if (powerState()) {
+                MDM_INFO("\r\n[ Modem::_powerOn ] Modem is powered on");
+            } else {
+                MDM_ERROR("\r\n[ Modem::_powerOn ] Failed to power on the modem.");
```
avtolstoy, line 851: This may pollute the logs, as VINT doesn't come up immediately.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -828,19 +828,29 @@ bool MDMParser::_powerOn(void)
...
     while (i--) {
-        // SARA-U2/LISA-U2 50..80us
-        HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(50);
-        HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(10);
-
-        // SARA-G35 >5ms, LISA-C2 > 150ms, LEON-G2 >5ms, SARA-R4 >= 150ms
-        HAL_GPIO_Write(PWR_UC, 0); HAL_Delay_Milliseconds(150);
-        HAL_GPIO_Write(PWR_UC, 1); HAL_Delay_Milliseconds(100);
+        if (powerState()) {
+            MDM_INFO("\r\n[ Modem::_powerOn ] Modem is already on");
```
avtolstoy, line 838: This also may pollute the logs a bit, as modem needs a bit of time to become responsive after VINT is high.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
avtolstoy: Let's revert this. It's a change in the behavior and a missing SIM may also be a temproary condition due to simply loose contacts or something like that. I would even propose we leave `powerOn()` here as-is.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1263,6 +1279,42 @@ bool MDMParser::powerState(void) const {
     return state;
 }
 
+bool MDMParser::powerOnOffGracefully(void) {
```
avtolstoy, line 1282: Nit: the naming is a bit confusing. Let's use `softPowerOff()` or something like that similar to what we use on Gen 3.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1271,85 +1323,62 @@ bool MDMParser::powerOff(void)
...
-        if (_cancel_all_operations) {
-            continue_cancel = true;
-            resume(); // make sure we can use the AT parser
-        }
+
+    MDM_INFO("%s = = = = = = = = = = = = = =", POWER_OFF_MSG);
+
+    if (!powerState() && !_init && !_pwr) {
+        MDM_INFO("%s Modem is off already", POWER_OFF_MSG);
+        return true;
```
avtolstoy, line 1331: This may again skip some deinitialization logic at the bottom.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1271,85 +1323,62 @@ bool MDMParser::powerOff(void)
...
             HAL_GPIO_Write(PWR_UC, 0);
             // >1.5 seconds on SARA R410M
             // >1 second on SARA U2
             // plus a little extra for good luck
-            HAL_Delay_Milliseconds(2000);
+            HAL_Delay_Milliseconds(1600);
+            HAL_GPIO_Write(PWR_UC, 1);
         }
+    } else if (_dev.dev == DEV_SARA_R410) {
+        check_ri = true;
```
avtolstoy, line 1381: I think we should always set it to true.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1271,85 +1323,62 @@ bool MDMParser::powerOff(void)
...
         }
-    } else {
-        if (powerState()) {
-            check_ri = true;
-            MDM_INFO("%s modem is on unexpectedly. Turning it off...", POWER_OFF_MSG);
-            // Delay this long period so that the modem is fully on to accept the power-off sequence.
-            HAL_Delay_Milliseconds(12000);
-
+        // Skip power off sequence if power is already off
+        if (powerState() && _dev.dev != DEV_SARA_G350 && _dev.dev != DEV_UNKNOWN) {
```
avtolstoy, line 1371: We were attempting power-off through pin 3 times previously, now it's just once.

avtolstoy (changes requested):
I'm requesting changes mainly due to `SIM_MISSING` change, but we also should make sure we can as much of the original behavior in `powerOff()` as possible.

hal/src/electron/modem/mdm_hal.cpp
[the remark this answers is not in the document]
XuGuohui: Should we keep the changes to `_powerOn()`. As for SARA-U2, the first low level pulse on PWR_ON may have powered on the modem, while the following 150ms low level pulse might power off the modem immediately.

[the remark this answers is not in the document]
avtolstoy: It shouldn't though. The power-off condition on SARA U2 is low pulse over 1s.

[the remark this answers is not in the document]
avtolstoy: image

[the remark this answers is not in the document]
XuGuohui: I see. Then why just give it the second low level pulse for all Electron modem types? As for SARA-U2, it has to walk through the second low level pulse any way.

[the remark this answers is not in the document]
avtolstoy: We can agree that ideally we should wait properly, but it is what it is. It works and doesn't have any negative consequences as of right now.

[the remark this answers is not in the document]
XuGuohui: Sounds reasonable. I'll revert the changes to the power on sequence.

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: I think we were having one attempt on powering off using pin: https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L1324, right?

avtolstoy: Good catch, yeah, you are right. :+1:

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1271,84 +1307,54 @@ bool MDMParser::powerOff(void)
...
```
avtolstoy, line 1350: I think we should still attempt this even if `_dev.dev == DEV_UNKNOWN`. Again according to the datasheet, we shouldn't be able to accidentally power-on the modem, as power-on condition has a limit much lower than 1600ms.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1263,6 +1263,42 @@ bool MDMParser::powerState(void) const {
...
+            }
+        }
+        if (i >= 3) {
+            // Failed to power on the modem.
+            MDM_INFO("\r\n[ Modem::softPowerOff ] Failed to power on the modem");
+            return false;
+        }
+    }
+    for (uint8_t i = 0; i < 3; i++) {
+        if (!_error && _atOk()) {
```
avtolstoy, line 1285: Should we change this to `_checkModem(true)`?

avtolstoy (approved):

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: I think we can't. Since in `_checkModem()` it checks
```
if (_cancel_all_operations) {
        return false;
    }
```
 while `_cancel_all_operations` might be set after entering `softPowerOff()`, in that case we cannot continue the soft power off procedure.

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: But in the datasheet it doesn't give the maximum low time on PWR_ON to power the modem. As long as the low time exceeds 5ms, the modem can be turned on.

avtolstoy: It does though: 
image

avtolstoy: If I am reading this correctly, pulses over 80uS fall out of the timing requirements for power-on condition.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Should we at least drop `_error` check then perhaps?

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: I was talking about the SARA-G350:
Screen Shot 2020-10-30 at 2 06 11 AM

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: So we're going to power the modem off regardless of any error?

avtolstoy: We should probably at least try to. We still have the AT/OK check in there to make sure the modem is responsive.

XuGuohui: The logic is the same as is: https://github.com/particle-iot/device-os/blob/develop/hal/src/electron/modem/mdm_hal.cpp#L1283. If there is certain error that indicates that the mode is not accepting any AT command, we can probably retry or use PWR_ON going forward.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Okay, double checked R410 as well and G350 is the only one without a maximum. We do have a `powerState()` check though, so supposedly we shouldn't be triggering a power-on, right?

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: No strong preferences, but we cannot use `PWR_ON` on G350, so the suggestion is to ignore any errors that happened previously and just rely on AT/OK, as otherwise there is no reason to also loop at all once we see `_error`, as there is nothing in the loop that can potentially reset it here in place.

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: You're right 👍 . If `powerState()` is high for SARA-G350, it will leaks current anyway, even if we are powering on it again.

XuGuohui: And for the other unknown modules, except for the SARA-G350, they earn a chance to power off themselves.

hal/src/electron/modem/mdm_hal.cpp
XuGuohui: Makes sense. Will drop the `_error` check.

PR #2230, opened 2020-10-29
[system] network diagnostics for Gen 3 platforms

avtolstoy:
### Problem

Network diagnostics/vitals are only enabled on Gen 2 platforms.

### Solution

Fix that and enable them for Gen 3 as well.

### Steps to Test

Request vitals in the console. They should contain network diagnostics. Make the device disconnect e.g. by unplugging antenna, let it connect again, request vitals.

### Example App

N/A

### References

- [CH66610]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_network_manager.cpp
+++ b/system/src/system_network_manager.cpp
@@ -384,7 +384,6 @@ void NetworkManager::transition(State state) {
         case State::IFACE_DOWN: {
             LED_SIGNAL_START(NETWORK_ON, BACKGROUND);
             NetworkDiagnostics::instance()->status(NetworkDiagnostics::DISCONNECTED);
-            NetworkDiagnostics::instance()->resetConnectionAttempts();
             if (state_ == State::IFACE_REQUEST_DOWN) {
                 system_notify_event(network_status, network_status_disconnected);
             } else if (state_ == State::DISABLED) {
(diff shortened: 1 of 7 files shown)


system/src/system_network_manager.cpp
```diff
@@ -757,7 +779,7 @@ int NetworkManager::enableInterface(if_t iface) {
     return 0;
 }
 
-int NetworkManager::disableInterface(if_t iface) {
+int NetworkManager::disableInterface(if_t iface, network_disconnect_reason reason) {
```
sergeuz, line 782: The `reason` argument is not used anywhere in this method.

system/src/system_network_manager.cpp
```diff
@@ -365,13 +375,16 @@ void NetworkManager::transition(State state) {
...
+            // FIXME: turning-off events/diagnostics
             system_notify_event(network_status, network_status_powering_off);
             system_notify_event(network_status, network_status_off);
+            NetworkDiagnostics::instance()->status(NetworkDiagnostics::TURNED_OFF);
             break;
         }
         case State::IFACE_DOWN: {
             LED_SIGNAL_START(NETWORK_ON, BACKGROUND);
+            NetworkDiagnostics::instance()->status(NetworkDiagnostics::DISCONNECTED);
+            NetworkDiagnostics::instance()->resetConnectionAttempts();
```
sergeuz, line 387: Wouldn't this reset the number of connection attempts prematurely? For example, when we're resetting the network interface after a number of unsuccessful attempts to connect to the network.

sergeuz (approved):

system/src/system_network_manager.cpp
avtolstoy: Fixed and moved to `REQUEST_DOWN` state case instead. It may not work entirely correct for cases where Ethernet is also available, but for regular devices this should be fine.

system/src/system_network_manager.cpp
avtolstoy: Added a comment explaining the behavior.

PR #2232, opened 2020-11-12
Query cellular signal while module tries to register

keeramis:
### Problem

The device, at times, do not register and instead stay in "searching" mode. In this case, it's unclear to tell if the device is even receiving a cellular signal which is useful for debugging purposes.

### Solution

Query cellular signal as a part of periodically checking the status of registration. 
- Not checking for an `OK` for the signal queries as sometimes these queries can return `ERROR` while the module is not having cellular coverage 

### Steps to Test

- Watch logs with antenna removed

### References

- [ch64153](https://app.clubhouse.io/particle/story/64153/query-signal-strength-as-a-part-of-registration-intervention-process)

### Completeness

- [x] n/a Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -1713,12 +1713,25 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
     _net.cgi.location_area_code = 0xFFFF;
     _net.cgi.cell_id = 0xFFFFFFFF;
     if (_dev.dev == DEV_SARA_R410) {
+        // Check the signal seen by the module while trying to register
+        // Do not need to check for an OK, as this is just for debugging purpose,
+        // and UCGED may sometimes return CME ERROR with low signal
+        sendFormated("AT+UCGED=5\r\n");
+        waitFinalResp(nullptr, nullptr, UCGED_TIMEOUT);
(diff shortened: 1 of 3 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1713,12 +1713,25 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
...
         // check EPS registration (LTE)
         sendFormated("AT+CEREG?\r\n");
         if (RESP_OK != waitFinalResp(nullptr, nullptr, CEREG_TIMEOUT)) {
             goto failure;
         }
     } else {
+        // Check the signal seen by the module while trying to register
+        // Do not need to check for an OK, as this is just for debugging purpose
+        sendFormated("AT+CSQ\r\n");
+        waitFinalResp(nullptr, nullptr, CSQ_TIMEOUT);
```
avtolstoy, line 1733: We should check for command timeout here (`WAIT`) and `goto failure`. Same with UCGED above.

avtolstoy (changes requested):

hal/src/electron/modem/mdm_hal.cpp
keeramis: Good point.. 👍  Added.

avtolstoy (approved):

technobly (approved):

PR #2235, opened 2020-11-19
[Electron] Build system parts with LTO

avtolstoy:
### Problem

system-part3 (2) overflows on Electron. We either have to move something out of it, or now that we've upgraded GCC, we can potentially enable LTO for a quick win.

### Solution

1. Build everything with LTO and fat objects (CFLAGS/CXXFLAGS need to have -flto)
2. Disable LTO explicitly if needed in linking stage for unaffected platforms
3. Link with LTO when building Electron system-part1/2/3

All the source code changes are designed to make LTO wrapper happy (e.g. fixing ODR violations etc)

### Steps to Test

Make sure that Electron functions as intended.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


XuGuohui (approved):
LGTM.

modules/electron/system-part3/makefile
```diff
@@ -26,6 +26,8 @@ TARGET=elf bin lst hex size
 
 include $(PROJECT_ROOT)/build/arm-tlm.mk
 
+LDFLAGS += -flto -Os -fuse-linker-plugin -Wl,--undefined=vTaskSwitchContext
```
sergeuz, line 29: A comment on `vTaskSwitchContext` would be nice here.

modules/shared/nRF52840/inc/user_part_export.c
```diff
@@ -10,14 +10,14 @@
...
  */
-extern char link_global_data_initial_values;
-extern char link_global_data_start;
-extern char link_global_data_end;
-#define link_global_data_size (&link_global_data_end - &link_global_data_start)
-
-extern char link_bss_location;
-extern char link_bss_end;
-#define link_bss_size (&link_bss_end - &link_bss_location)
+extern uintptr_t link_global_data_initial_values;
```
sergeuz, line 13: Does GCC with LTO enabled complains about these symbols declared as `char`? Using `uintptr_t` here is confusing since you have to take the address of such a symbol in the code rather than use it by value.

build/arm-tools.mk
```diff
@@ -45,9 +45,14 @@ ASFLAGS += -mfloat-abi=hard -mfpu=fpv4-sp-d16
 LDFLAGS += -nostartfiles -Xlinker --gc-sections
 endif
 
+CPPFLAGS += -flto -ffat-lto-objects -DPARTICLE_COMPILE_LTO_FAT
+CONLYFLAGS += -flto -ffat-lto-objects -DPARTICLE_COMPILE_LTO_FAT
```
sergeuz, line 49: A comment clarifying that we're not enabling LTO by default would be nice here. Maybe it's just me, but I didn't know about fat LTO objects.

sergeuz (approved):
Nice! :+1:

modules/shared/nRF52840/inc/user_part_export.c
avtolstoy: Yeah, there was an ODR violation somewhere and we've done a similar change previously when implementing original GCC 9 support, so the intent was to use uintptr_t commonly (https://github.com/particle-iot/device-os/pull/2103#discussion_r425725917)

modules/electron/system-part3/makefile
avtolstoy: Added.

build/arm-tools.mk
avtolstoy: Added

PR #2238, opened 2020-11-23
Gen2: platforms calling sleep20 API to enter stop/ulp mode may block user application.

XuGuohui:
### Problem
As the title describes.
### Solution
Set a flag if cellular/wifi connection need to be restored after waking up and let the background system task to handle the connection establishment, instead of calling the network connect api to before exiting the sleep function.
### Steps to Test
`user/tests/wiring/sleep20`
### References
N/A
---
### Completeness
- [ ] Added documentation

### DIFF
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -55,6 +55,11 @@ static bool system_sleep_network_suspend(network_interface_index index) {
         network_disconnect(index, NETWORK_DISCONNECT_REASON_SLEEP, NULL);
         resume = true;
     }
+#if PLATFORM_GEN == 2
+    if (!SPARK_WLAN_SLEEP) {
+        resume = true;
+    }
+#endif
     // Turn off the modem
     network_off(index, 0, 0, NULL);
     LOG(TRACE, "Waiting interface to be off...");
@@ -64,8 +69,12 @@ static bool system_sleep_network_suspend(network_interface_index index) {
 }
 
 static int system_sleep_network_resume(network_interface_index index) {
+#if PLATFORM_GEN == 2
+    SPARK_WLAN_SLEEP = 0;
+#else
     network_on(index, 0, 0, nullptr);
     network_connect(index, 0, 0, nullptr);
+#endif
     return SYSTEM_ERROR_NONE;
 }
 
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -70,6 +70,10 @@ static bool system_sleep_network_suspend(network_interface_index index) {
 
 static int system_sleep_network_resume(network_interface_index index) {
 #if PLATFORM_GEN == 2
+    /* On Gen2, calling network_on() and network_connect() will block until the connection is established
+     * if single threaded, or this function is invoked synchronously by the system thread if system threading
+     * is enabled. In both case, that would block the user application. Setting a flag here to unblock the user
+     * application and restore the connection later. */
     SPARK_WLAN_SLEEP = 0;
 #else
     network_on(index, 0, 0, nullptr);
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -196,7 +196,6 @@ int system_sleep_ext(const hal_sleep_config_t* config, hal_wakeup_source_base_t*
 
 #if HAL_PLATFORM_WIFI
     if (wifiResume) {
-        SPARK_WLAN_SLEEP = 0;
         system_sleep_network_resume(NETWORK_INTERFACE_WIFI_STA);
     }
 #endif // HAL_PLATFORM_WIFI
--- a/user/tests/wiring/sleep20/sleep20.cpp
+++ b/user/tests/wiring/sleep20/sleep20.cpp
@@ -670,10 +670,9 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
 #endif
 #endif // HAL_PLATFORM_WIFI
 
-#if HAL_PLATFORM_CELLULAR
-test(23_System_Sleep_With_Configuration_Object_Stop_Mode_Execution_Time) {
-    constexpr uint32_t SLEEP_DURATION_MS = 1000;
-    Serial.printf("    >> Device enters stop mode. Please reconnect serial after %ld ms\r\n", SLEEP_DURATION_MS);
+test(27_System_Sleep_With_Configuration_Object_Stop_Mode_Execution_Time) {
+    constexpr uint32_t SLEEP_DURATION_S = 3;
+    Serial.printf("    >> Device enters stop mode. Please reconnect serial after %ld s\r\n", SLEEP_DURATION_S);
     Serial.println("    >> Press any key now");
     while(Serial.available() <= 0);
     while (Serial.available() > 0) {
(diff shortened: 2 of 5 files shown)


system/src/system_sleep.cpp
```diff
@@ -63,8 +68,12 @@ static bool system_sleep_network_suspend(network_interface_index index) {
 }
 
 static int system_sleep_network_resume(network_interface_index index) {
+#if PLATFORM_GEN == 2
+    SPARK_WLAN_SLEEP = 0;
```
avtolstoy, line 72: Let's add a comment explaining why this was needed.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    constexpr uint32_t SLEEP_DURATION_MS = 1000;
+    Serial.printf("    >> Device enters stop mode. Please reconnect serial after %ld ms\r\n", SLEEP_DURATION_MS);
+    Serial.println("    >> Press any key now");
+    while(Serial.available() <= 0);
+    while (Serial.available() > 0) {
+        (void)Serial.read();
+    }
+
+    Serial.println("    >> Connecting to the cloud");
+    Cellular.on();
```
avtolstoy, line 684: Let's generalize this and use `Network` here, so that this can be verified on WiFi devices as well, within a single test.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    Particle.connect();
+    waitUntil(Particle.connected);
+    Serial.println("    >> Connected to the cloud. You'll see the RGB is turned on after waking up.");
+
+    SystemSleepConfiguration config;
+    config.mode(SystemSleepMode::STOP)
+          .duration(SLEEP_DURATION_MS);
+
+    system_tick_t enter = millis();
```
avtolstoy, line 693: This shouldn't work on Gen 2 platforms. `millis()` only count while the MCU is running on Gen 2, whereas on Gen 3 micros, millis and unix time will continue to count while in sleep mode.

We should instead use `Time.now()` universally, which continues to count in STOP mode across all our platforms. In order to get better accuracy, we should just increase sleep time to several seconds.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    SystemSleepResult result = System.sleep(config);
+    system_tick_t exit = millis();
+
+    while (!Serial.isConnected());
+    // It might wait for up to 120s to turn off the modem
+    assertLessOrEqual(exit - enter - SLEEP_DURATION_MS, 121000);
```
avtolstoy, line 699:
```
exit = 123;
enter = 123;
123 - 123 - SLEEP_DURATION_MS = -SLEEP_DURATION_MS
-SLEEP_DURATION_MS <= 121000 -> true
```

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    Serial.printf("Sleep execution time: %ld\r\n", (uint32_t)(exit - enter));
+
+    assertEqual(result.error(), SYSTEM_ERROR_NONE);
+    assertEqual((int)result.wakeupReason(), (int)SystemSleepWakeupReason::BY_RTC);
+}
+
+#if HAL_PLATFORM_GEN == 3
```
avtolstoy, line 706: ULP is supported on Gen 2 as well, right?

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    constexpr uint32_t SLEEP_DURATION_MS = 1000;
+    Serial.printf("    >> Device enters ultra-low power mode. Please reconnect serial after %ld ms\r\n", SLEEP_DURATION_MS);
+    Serial.println("    >> Press any key now");
+    while(Serial.available() <= 0);
+    while (Serial.available() > 0) {
+        (void)Serial.read();
+    }
+
+    Serial.println("    >> Connecting to the cloud");
+    Cellular.on();
```
avtolstoy, line 717: Same, let's make this connectivity-type agnostic.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,139 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    while (!Serial.isConnected());
+    // It might wait for up to 120s to turn off the modem
+    assertLessOrEqual(exit - enter - SLEEP_DURATION_MS, 121000);
+    Serial.printf("Sleep execution time: %ld\r\n", (uint32_t)(exit - enter));
+
+    assertEqual(result.error(), SYSTEM_ERROR_NONE);
+    assertEqual((int)result.wakeupReason(), (int)SystemSleepWakeupReason::BY_RTC);
+}
+
+test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_WiFi) {
```
avtolstoy, line 774: If we make the other two tests network type agnostic, these can be removed.

avtolstoy (changes requested):

system/src/system_sleep.cpp
XuGuohui: Added.

user/tests/wiring/sleep20/sleep20.cpp
XuGuohui: Done.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,69 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    config.mode(SystemSleepMode::STOP)
+          .duration(SLEEP_DURATION_S * 1000);
+
+    time32_t enter = Time.now();
+    SystemSleepResult result = System.sleep(config);
+    time32_t exit = Time.now();
+
+    while (!Serial.isConnected());
+    // It might wait for up to 120s to turn off the modem
+    assertLessOrEqual(exit - enter, 120 + SLEEP_DURATION_S);
```
avtolstoy, line 698: Let's be a bit more strict and also add a check that the sleep was `>= SLEEP_DURATION_S`. Same in the next test.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -661,5 +667,69 @@ test(26_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_By_Wi
...
+    Network.on();
+    Particle.connect();
+    waitUntil(Particle.connected);
+    Serial.println("    >> Connected to the cloud. You'll see the RGB is turned on after waking up.");
+
+    SystemSleepConfiguration config;
+    config.mode(SystemSleepMode::ULTRA_LOW_POWER)
+          .duration(SLEEP_DURATION_S * 1000);
+
+    time32_t enter = millis();
```
avtolstoy, line 724: This needs to use `Time`, as the test above.

avtolstoy (changes requested):

user/tests/wiring/sleep20/sleep20.cpp
XuGuohui: Updated.

avtolstoy (approved):
@XuGuohui Please do review my set of changes here.

PR #2246, opened 2020-12-04
[gen3] bootloader: add workaround for MODULE_DROP_MODULE_INFO

avtolstoy:
### Problem

We've introduced an incompatibility in https://github.com/particle-iot/device-os/pull/2097 between older Gen 3 system-part1 modules and new bootloader in handling of `MODULE_DROP_MODULE_INFO` flag.

Prior to the change the expected behavior for the bootloader was to check the module header in the OTA section and look at the actual flags from it (where `MODULE_DROP_MODULE_INFO` would be present).

After the change, the expected behavior for the bootloader is to only look at the update module slot flags that are stored in DCT, however DeviceOS versions that don't have the change do not set the appropriate slot flag that would instruct the bootloader to drop module info when applying the update.

This breaks SoftDevice updates under following conditions:
1. Gen 3 platforms
2. system-part1 version is < 2.0.0-rc.1
3. bootloader version already on the device is >= 2.0.0-rc.1 (which happens normally during OTA update process due to dependency resolution)

### Solution

The PR brings back an additional compatibility behavior to additionally look into the module header to understand whether module info needs to be dropped or not.

### Steps to Test

1. Get the device into the condition specified in the problem section. E.g. put 1.3.1 system-part1 onto device and flash 2.0.0 bootloader
2. Flash SoftDevice binary either OTA or using YModem
3. The device should apply the update normally and boot back up
4. Update system-part1 to 2.0.0
5. Flash SoftDevice binary either OTA or using YModem
6. The device should apply the update normally and boot back up

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/platform/MCU/nRF52840/src/flash_mal.c
+++ b/platform/MCU/nRF52840/src/flash_mal.c
@@ -411,6 +411,25 @@ int FLASH_CopyMemory(flash_device_t sourceDeviceID, uint32_t sourceAddress,
         return FLASH_ACCESS_RESULT_BADARG;
 #endif // !HAS_COMPRESSED_OTA
     } else {
+        // XXX: Device OS versions < 2.0.0 may not be setting this flag in the module slots!
+        // Bootloader should rely on the actual flags within the module header!
+        if (!(flags & MODULE_DROP_MODULE_INFO) && (flags & MODULE_VERIFY_MASK)) {
+            // We may only check the module header if we've been asked to verify it
+            uint32_t infoOffset = 0;
+            const module_info_t* info = FLASH_ModuleInfo(sourceDeviceID, sourceAddress, &infoOffset);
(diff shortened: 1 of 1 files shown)


XuGuohui (approved):
Just a reminder that the relying on `infoOffset` being `0` to drop the module info is a hacking way, because module info following vector table can be dropped theoretically.

avtolstoy:
> module info following vector table can be dropped theoretically.

Just want to clarify that this has always been the intended behavior. There is no point in having module info not at the front of the module if we are anyway going to drop it (it doesn't matter whether a piece is dropped from the middle of the module or from the front, the end result is still the same, so why not package things in a proper manner from the beginning? `DROP_MODULE_INFO` flag exists in the module header anyway), so this is an additional sanity check to make sure that we are not accidentally doing a wrong thing.

PR #2247, opened 2020-12-08
Use unique Particle prefix for LED defines

tstellanova:
<details>
  <summary><i>Rename simple numbered LED defines to have PARTICLE prefix to avoid possible collision with user applications.</i></summary>

</details>

### Problem

What Bug or Missing Feature does this PR address? (A short summary is preferred over links)

Simple define names such as `LED2` have caused some users problems due to namespace collisions.  

### Solution

By adding a `PARTICLE_` prefix to these defines, we avoid possible namespace collisions. 

### Steps to Test

At the moment tests under `/user/tests` are using `LED_RGB` which is derived from these defines. 

### Example App

Replace any reference to eg `LED2` with `PARTICLE_LED2` in your applications.  

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/platform/MCU/nRF52840/inc/platform_config.h
+++ b/platform/MCU/nRF52840/inc/platform_config.h
@@ -101,9 +101,9 @@
 //RGB LEDs
 #define LED_MIRROR_SUPPORTED                        1
 #define LEDn                                        (4 * (LED_MIRROR_SUPPORTED + 1))
-#define LED_RED                                     LED2  // RED Led
-#define LED_GREEN                                   LED3  // GREEN Led
-#define LED_BLUE                                    LED4  // BLUE Led
+#define PARTICLE_LED_RED                            PARTICLE_LED2  // RED Led
+#define PARTICLE_LED_GREEN                          PARTICLE_LED3  // GREEN Led
+#define PARTICLE_LED_BLUE                           PARTICLE_LED4  // BLUE Led
 #define LED_PIN_USER                                D7
 #define LED_PIN_RED                                 RGBR
 #define LED_PIN_GREEN                               RGBG
--- a/services/inc/rgbled.h
+++ b/services/inc/rgbled.h
@@ -6,9 +6,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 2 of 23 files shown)


platform/MCU/nRF52840/inc/platform_config.h
```diff
@@ -101,9 +101,9 @@
 //RGB LEDs
 #define LED_MIRROR_SUPPORTED                        1
 #define LEDn                                        (4 * (LED_MIRROR_SUPPORTED + 1))
-#define LED_RED                                     LED2  // RED Led
-#define LED_GREEN                                   LED3  // GREEN Led
-#define LED_BLUE                                    LED4  // BLUE Led
+#define LED_RED                                     PARTICLE_LED2  // RED Led
```
avtolstoy, line 104: We could also potentially prefix all of these as they are somewhat too generic, same for the other `platform_config.h`

tstellanova:
I'll call out that I've introduced some content to the `platform_config.h` files for the `gcc` and `newhal` targets.  This is because these targets currently depend upon the LED r,g,b definitions, and they're ordinarily defined by `platform_config.h`.  I realize this may break with tradition for these targets.

services/inc/rgbled.h
```diff
@@ -25,38 +23,23 @@ typedef struct {
...
+#define PARTICLE_LED3_MIRROR        (PARTICLE_LED3 + LED_MIRROR_OFFSET)
+#define PARTICLE_LED4_MIRROR        (PARTICLE_LED4 + LED_MIRROR_OFFSET)
 
 //Extended LED Types
-#define LED_RGB           LED3_LED4_LED2
-#define LED_USER          LED1
-
-// FIXME: These should be pulled from platform_config.h
-#ifndef LED_RED
-#define LED_RED           LED3
```
XuGuohui, line 50: `#error "something"` in case that they are not defined in `platform_config.h`?

XuGuohui (approved):

avtolstoy (approved):

avtolstoy (approved):

PR #2248, opened 2020-12-08
[hal] sleep: allows the network to stay standby in HIBERNATE mode.

XuGuohui:
### Problem
`System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::HIBERNATE).network(SystemSleepNetworkFlag::INACTIVE_STANDBY))` doesn't make the device enter sleep.
### Solution
Removes the restriction in sleep HAL to allow network stay standby in hibernate mode. 
### Example App

```c
#include "Particle.h"
#include "system_threading.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

void setup()
{
    waitFor(Serial.isConnected, 10000);

    Log.info("before sleep");
    config.network(Cellular, SystemSleepNetworkFlag::INACTIVE_STANDBY);
    config.mode(SystemSleepMode::HIBERNATE);
    config.gpio(D0, RISING);
    System.sleep(config);
}

void loop()
{
}
```
---
### Completeness
- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/sleep20/sleep20.cpp
+++ b/user/tests/wiring/sleep20/sleep20.cpp
@@ -755,3 +755,34 @@ test(29_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_Execu
     // Make sure we reconnect back to the cloud
     assertTrue(waitFor(Particle.connected, CLOUD_CONNECT_TIMEOUT));
 }
+
+test(30_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Bypass_Network_Off_Execution_Time) {
+    constexpr uint32_t SLEEP_DURATION_S = 3;
+    Serial.printf("    >> Device enters ultra-low power mode. Please reconnect serial after %ld s\r\n", SLEEP_DURATION_S);
+    Serial.println("    >> Press any key now");
+    while (Serial.available() <= 0);
+    while (Serial.available() > 0) {
+        (void)Serial.read();
+    }
+
+    Particle.disconnect();
+    waitUntil(Particle.disconnected);
+
+    time32_t enter = Time.now();
+    SystemSleepConfiguration config;
+    config.mode(SystemSleepMode::ULTRA_LOW_POWER)
+          .duration(SLEEP_DURATION_S * 1000);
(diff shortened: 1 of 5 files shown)


avtolstoy (approved):

avtolstoy:
@XuGuohui If possible let's add a test at least on Gen 2:

1. With the modem powered on or perhaps off as that's closer to the case we were trying to deal with
2. We take a note of the current RTC time and put it into a retained variable
3. Go into standby with the appropriate flags
4. Bootup again
5. Take a note of how much time has elapsed based on RTC time
6. Make an assertion on that (so that we know that no power-off sequence was triggered)

XuGuohui:
Okay, will do.

user/tests/wiring/sleep20/sleep20.cpp
```diff
@@ -755,3 +755,34 @@ test(29_System_Sleep_With_Configuration_Object_Ultra_Low_Power_Mode_Wakeup_Execu
...
+    while (Serial.available() > 0) {
+        (void)Serial.read();
+    }
+
+    Particle.disconnect();
+    waitUntil(Particle.disconnected);
+
+    time32_t enter = Time.now();
+    SystemSleepConfiguration config;
+    config.mode(SystemSleepMode::ULTRA_LOW_POWER)
```
avtolstoy, line 773: We should ideally be testing HIBERNATE sleep mode here instead, as that's the main thing we've fixed.

avtolstoy (approved):

user/tests/wiring/sleep20/sleep20.cpp
XuGuohui: Hibernate mode will reset the device on waking up and device then waits for user to give instructions to start the test. So the elapsed time is not accurate. If needed, we can add a separated test case to verify that the hibernate mode does accept the network flag. What's your thought?

avtolstoy: 1. You can use a retained variable to keep the start time
2. You can sample Time.now() in `STARTUP()` block as early as possible on boot.

I'd still like to have some sort of a test for HIBERNATE specifically ideally.

avtolstoy (approved):

PR #2249, opened 2020-12-08
Support for DTLS connection IDs

sergeuz:
### Problem

This PR enables the DTLS Connection ID extension (described in [this](https://tools.ietf.org/html/draft-ietf-tls-dtls-connection-id-09) draft RFC).

### Steps to Test

Verify that session resumption works as expected.

### References

- [ch55112]

### DIFF
--- a/communication/src/dtls_message_channel.cpp
+++ b/communication/src/dtls_message_channel.cpp
@@ -23,6 +23,16 @@
 
 LOG_SOURCE_CATEGORY("comm.dtls")
 
+struct mbedtls_ssl_context;
+struct mbedtls_ssl_transform;
+
+extern "C" {
+
+void mbedtls_ssl_update_in_pointers(mbedtls_ssl_context *ssl);
+void mbedtls_ssl_update_out_pointers(mbedtls_ssl_context *ssl, mbedtls_ssl_transform *transform);
+
+} // extern "C"
+
 #include "dtls_message_channel.h"
 
 #if HAL_PLATFORM_CLOUD_UDP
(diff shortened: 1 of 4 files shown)


communication/src/dtls_message_channel.cpp
```diff
@@ -230,31 +258,16 @@ ProtocolError DTLSMessageChannel::init(
 	this->server_public = new uint8_t[server_public_len];
 	memcpy(this->server_public, server_public, server_public_len);
 	this->server_public_len = server_public_len;
+
+	// Records received from the server are not expected to contain a CID
+	mbedtls_ssl_conf_cid(&conf, 0 /* len */, MBEDTLS_SSL_UNEXPECTED_CID_IGNORE);
```
avtolstoy, line 263: Nit: `MBEDTLS_SSL_CID_IN_LEN_MAX` perhaps?

avtolstoy (approved):

avtolstoy:
I rebased and fixed a couple of conflicts around persistent session.

avtolstoy:
One thing to consider (as far as I understand DS side of this feature allows) is to remove server ip:port from the persistent storage and perform DNS resolution all the time to allow devices to resume the session on any DS instance. This should allow us to simplify DS deploys, have devices migrate to current DS instances quicker, increase maximum session times and improve connection times for sleepy devices in some cases.

communication/src/dtls_message_channel.cpp
sergeuz: I'd prefer to use 0 here as it makes the intent clear that we don't expect a CID in server-originated messages. `MBEDTLS_SSL_CID_IN_LEN_MAX` defined as 0 in our config is more a memory optimization.

sergeuz:
> I rebased and fixed a couple of conflicts around persistent session.

@avtolstoy Thanks!

PR #2250, opened 2020-12-09
[Tracker] ESP32 support

avtolstoy:
### Description

This PR adds WiFi scanning support for Tracker. All other APIs except for WiFi interface state management and scanning are not available as of right now.

This feature branch should work with both the stock 0.0.6 ESP32 NCP firmware available from the factory, as well as with https://github.com/particle-iot/argon-ncp-firmware/pull/18.

### Steps to Test

- `wiring/no_fixture` tests have been fixed to allow WiFi scanning on Tracker

1. WiFi scanning should work with stock 0.0.6 ESP32 NCP firmware
2. You should be able to grab a build from https://github.com/particle-iot/argon-ncp-firmware/pull/18 and update update over serial or OTA to 0.0.7 version (`particle serial inspect` should reflect that)
3. WiFi scanning should work with 0.0.7 ESP32 NCP firmware over muxed channel
4. 0.0.7 ESP32 NCP firmware should be updatable. Manually edit version.mk in https://github.com/particle-iot/argon-ncp-firmware/pull/18 and bump version to 0.0.8 and perform a clean build. Update OTA or using serial should work.

### Example App

```c++
SYSTEM_THREAD(ENABLED);

void setup() {
    WiFi.on();
}

void loop() {
    WiFi.scan([](WiFiAccessPoint* result, void* cookie) -> void {
        char ssid[64] = {};
        memcpy(ssid, result->ssid, std::min<size_t>(sizeof(ssid) - 1, result->ssidLength));
        Log.info("Found SSID: %s (%02x:%02x:%02x:%02x:%02x:%02x)", ssid,
                result->bssid[0],
                result->bssid[1],
                result->bssid[2],
                result->bssid[3],
                result->bssid[4],
                result->bssid[5]);
    }, nullptr);
    delay(1000);
}

```

### References

- https://github.com/particle-iot/argon-ncp-firmware/pull/18
- https://github.com/particle-iot/gsm0710muxer/pull/7

---

### Completeness

- [ ] Added documentation


hal/inc/hal_dynalib_wlan.h
```diff
@@ -43,47 +36,11 @@
...
-DYNALIB_FN(24, hal_wlan, wlan_get_credentials, int(wlan_scan_result_t, void*))
-DYNALIB_FN(25, hal_wlan, softap_set_application_page_handler, int(PageProvider* provider, void* reserved))
-DYNALIB_FN(26, hal_wlan, wlan_restart, int(void*))
-DYNALIB_FN(27, hal_wlan, wlan_set_hostname, int(const char*, void*))
-DYNALIB_FN(28, hal_wlan, wlan_get_hostname, int(char*, size_t, void*))
-DYNALIB_FN(29, hal_wlan, wlan_connected_info, int(void*, wlan_connected_info_t*, void*))
-DYNALIB_FN(30, hal_wlan, wlan_get_antenna, WLanSelectAntenna_TypeDef(void*))
-DYNALIB_FN(31, hal_wlan, wlan_get_ipaddress, int(IPConfig*, void*))
-DYNALIB_FN(32, hal_wlan, wlan_get_ipaddress_source, IPAddressSource(void*))
+DYNALIB_FN(0, hal_wlan, wlan_select_antenna, int(WLanSelectAntenna_TypeDef))
```
eberseth, line 39: If `HAL_PLATFORM_WIFI_COMPAT` then expose the legacy WiFi HAL interface; otherwise, only minimal?

hal/inc/hal_dynalib_wlan.h
```diff
@@ -43,47 +36,11 @@
...
-DYNALIB_FN(17, hal_wlan, HAL_NET_SetNetWatchDog, uint32_t(uint32_t))
-DYNALIB_FN(18, hal_wlan, inet_gethostbyname, int(const char*, uint16_t, HAL_IPAddress*, network_interface_t, void*))
-DYNALIB_FN(19, hal_wlan, inet_ping, int(const HAL_IPAddress*, network_interface_t, uint8_t, void*))
-DYNALIB_FN(20, hal_wlan, wlan_select_antenna, int(WLanSelectAntenna_TypeDef))
-DYNALIB_FN(21, hal_wlan, wlan_set_ipaddress, void(const HAL_IPAddress*, const HAL_IPAddress*, const HAL_IPAddress*, const HAL_IPAddress*, const HAL_IPAddress*, void*))
-DYNALIB_FN(22, hal_wlan, wlan_set_ipaddress_source, void(IPAddressSource, bool, void*))
-DYNALIB_FN(23, hal_wlan, wlan_scan, int(wlan_scan_result_t, void*))
-DYNALIB_FN(24, hal_wlan, wlan_get_credentials, int(wlan_scan_result_t, void*))
-DYNALIB_FN(25, hal_wlan, softap_set_application_page_handler, int(PageProvider* provider, void* reserved))
-DYNALIB_FN(26, hal_wlan, wlan_restart, int(void*))
```
eberseth, line 77: Would there be a good reason to include the restart into the minimal implementation?

hal/inc/hal_dynalib_wlan.h
avtolstoy: Correct. We can rename it if needed, but the current HAL is somewhat overloaded and the idea is that we can eventually build up new APIs starting with the minimal one.

eberseth: Are there any special commands to induce sleep?

hal/inc/hal_dynalib_wlan.h
avtolstoy: That's not an API that's normally used by the wiring. It's internal. `WiFi.off()` and `WiFi.on()` are the right type of APIs with the addition of https://github.com/particle-iot/device-os/pull/2205.

In case of any failure conditions, network driver will take care of performing a reset if needed.

hal/inc/hal_dynalib_wlan.h
avtolstoy: There are none. It's still a proper network interface, so it is subject to standard sleep behavior: unless it's specified as a wake-up source or `INACTIVE_STANDBY` flag is enabled it will be turned off before going into sleep.

hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -72,15 +73,15 @@ LOG_SOURCE_CATEGORY("net.en")
             break;                                                              \
         }                                                                       \
         /* I'd really love to get rid of this */                                \
-        os_thread_yield();                                                      \
+        HAL_Delay_Milliseconds(1);                                              \
```
eberseth, line 76: Why switch over to a delay here?

avtolstoy: `os_thread_yield()` yields only to threads with higher priority. `HAL_Delay_Milliseconds(1)` will also yield to threads of lower or same priority if they require CPU time.

hal/network/ncp/wifi/platform_ncp_update.cpp
```diff
@@ -0,0 +1,167 @@
...
+// FIXME: This function accesses the module info via XIP and may fail to parse it correctly under
+// some not entirely clear circumstances. Disabling compiler optimizations helps to work around
+// the problem
+__attribute__((optimize("O0"))) int platform_ncp_update_module(const hal_module_t* module) {
+    const auto ncpClient = particle::wifiNetworkManager()->ncpClient();
+    SPARK_ASSERT(ncpClient);
+    CHECK(ncpClient->on());
+    // we pass only the actual binary after the module info and up to the suffix
+    const uint8_t* start = (const uint8_t*)module->info;
+    static_assert(sizeof(module_info_t)==24, "expected module info size to be 24");
```
eberseth, line 92: What happens in the event of an assert?

eberseth, line 92: This seems like a hard magic number.  any way to derive this?

avtolstoy (commented):
123

hal/network/ncp/wifi/platform_ncp_update.cpp
avtolstoy: If it's calculated, it removes the purpose of this check which is to ensure that a particular structure is indeed of a particular size.

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
```diff
@@ -68,12 +80,18 @@ const auto ESP32_NCP_KEEPALIVE_MAX_MISSED = 5;
...
 
 const auto ESP32_NCP_AT_CHANNEL = 1;
 const auto ESP32_NCP_STA_CHANNEL = 2;
 const auto ESP32_NCP_AP_CHANNEL = 3;
 
+#if !HAL_PLATFORM_WIFI_NCP_SDIO
 const auto ESP32_NCP_MIN_MVER_WITH_CMUX = 4;
+#else
+// FIXME:
+const auto ESP32_NCP_MIN_MVER_WITH_CMUX = 7;
```
eberseth, line 93: 7 or 8?

hal/network/ncp/wifi/platform_ncp_update.cpp
avtolstoy: To clarify this is not a new code, it's https://github.com/particle-iot/device-os/blob/develop/hal/src/argon/platform_ncp_argon.cpp

hal/network/ncp/wifi/platform_ncp_update.cpp
```diff
@@ -0,0 +1,167 @@
...
+                const auto ncpClient = particle::wifiNetworkManager()->ncpClient();
+                SPARK_ASSERT(ncpClient);
+                CHECK(ncpClient->on());
+                uint16_t version;
+                int error = ncpClient->getFirmwareModuleVersion(&version);
+                if (error) {
+                    version = 0;
+                }
+                // todo - we could augment the getFirmwareModuleVersion command to retrieve more details
+                auto info = new module_info_t();
```
eberseth, line 126: nothrow?

hal/network/ncp_client/esp32/esp32_ncp_client.cpp
avtolstoy: We'll go with 7. I should remove the FIXME. The latest (albeit non public) release is 0.0.6. There is no point to jump two version.

hal/network/ncp/wifi/platform_ncp_update.cpp
avtolstoy: There is also no difference between `new` and `new (std::nothrow)` on current platforms other than stylistic as exceptions are disabled and there is a check below. We can add it, but again https://github.com/particle-iot/device-os/pull/2250#discussion_r539603067

hal/src/nRF52840/gsm0710muxer/platform.h
```diff
@@ -45,6 +45,7 @@ const auto taskPriority = OS_THREAD_PRIORITY_NETWORK;
 // #define GSM0710_MODEM_STATUS_SEND_EMPTY_BREAK
 // #define GSM0710_PUMP_INPUT_DATA 50
 // #define GSM0710_ENABLE_DEBUG_LOGGING
+// FIXME:
```
eberseth, line 48: Fix what?

avtolstoy: :+1:
No longer relevant, should be removed.

hal/network/lwip/wiznet/wiznetif.cpp
eberseth: Priority inversion problem?

hal/network/ncp/wifi/platform_ncp_update.cpp
eberseth: Up to you for consistency

hal/network/lwip/wiznet/wiznetif.cpp
avtolstoy: No, that's the documented behavior: https://www.freertos.org/a00020.html#taskYIELD

eberseth: ok, sounds good

hal/network/ncp/wifi/platform_ncp_update.cpp
avtolstoy: Then I'll leave it as-is, since this code block was simply moved and not a new change. We don't have a strong policy on `new (std::nothrow)` either.

eberseth (approved):

YutingYou:
@avtolstoy Great reworking, since you've changed the transport layer, I was wondering have you run the stress test for OTA, this would be helpful for the stability of OTA and SDIO. The easiest way may be manually doing updating with CLI several times. Better to upgrade and downgrade.

avtolstoy:
@YutingYou Yes, I've done that multiple times, you can see that there are some FIXMEs around XModem that were uncovered both here and in the NCP firmware. I've also listed updates as one of the main things to test :)

XuGuohui (changes requested):
This reminds me that in the system_network_manager.cpp, some of the APIs don't support more than one interface. We need to refactor them but we can do that later if they don't break things.

hal/shared/spi_lock.h
```diff
@@ -0,0 +1,136 @@
...
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "spi_hal.h"
+
+namespace particle {
+
+class SpiConfigurationLock {
```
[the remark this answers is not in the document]
avtolstoy: Wiring layer has `SPI.beginTransaction()`, so there is no need to make it available to user application.

I'll export `hal_spi_get_clock_divider()` and fixup how `SPIClass::beginTransaction(const particle::SPISettings& settings)` calculates divider as well. :+1:

hal/src/tracker/esp32_sdio_stream.cpp
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+    return pos;
+}
+
+int Esp32SdioStream::flush() {
+    if (!enabled_) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    return 0;
```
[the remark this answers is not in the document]
avtolstoy: `txBuf_` is only used as a temporary buffer for a single write (split into 512 chunks) to make sure that we are transmitting out of RAM. Once we've submitted as many chunks to ESP32 as possible (`sdio_->write()`) it does not retain any data after that. Hence why `availForWrite()` always returns 512.

hal/src/tracker/esp32_sdio_stream.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+int Esp32SdioStream::init() {
+    CHECK(sdio_->init());
+    enabled_ = true;
+    flush();
+    return SYSTEM_ERROR_NONE;
+}
+
+int Esp32SdioStream::read(char* data, size_t size) {
+    if (!enabled_) {
```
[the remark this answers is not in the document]
avtolstoy: It was a copy-paste from another `Stream` implementation. No strong preferences here.

hal/src/tracker/sdspi/esp32_sdio.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+void Esp32Sdio::txInterruptSupported(bool state) {
+    txInterruptSupported_ = state;
+}
+
+int Esp32Sdio::on(bool on) {
+    if (!on) {
+        HAL_Interrupts_Detach(intrPin_);
+        txInterruptSupported(false);
+    }
```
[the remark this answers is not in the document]
avtolstoy: That happens in the `init()` call due to how ESP32 NCP client works. No need to do it here as we anyway have to go through the full initialization.

modules/shared/nRF52840/inc/system-part1/module_system_part1.inc
```diff
@@ -100,6 +100,9 @@ extern "C" __attribute__((externally_visible)) const void* const system_part1_mo
     , DYNALIB_TABLE_NAME(hal_mesh_deprecated)
 #endif // HAL_PLATFORM_MESH_DEPRECATED
     , DYNALIB_TABLE_NAME(hal_posix_syscall)
+#if HAL_PLATFORM_WIFI && !HAL_PLATFORM_WIFI_COMPAT
+    , DYNALIB_TABLE_NAME(hal_wlan)
```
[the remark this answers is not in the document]
avtolstoy: No strong preferences. We use the same name with `hal_socket` where on some platforms it's POSIX and on others it's different. We should at least add a comment for sure to make the distinction that this is for Tracker to keep the order intact and that it's minimal compared to Argon :+1:

hal/shared/spi_lock.h
```diff
@@ -0,0 +1,136 @@
...
+public:
+    SpiConfigurationLock(hal_spi_interface_t spi)
```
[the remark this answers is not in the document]
avtolstoy: This is to avoid the need to have the configuration readily available at the time of the construction (this is problematic if we make `SpiConfigurationLock` a member variable). Since this is an internal class, we can perhaps add error reporting on `lock()` where this would be caught, or perhaps add an assertion there that it's properly configured.

hal/shared/spi_lock.h
```diff
@@ -0,0 +1,136 @@
...
+    SpiConfigurationLock(hal_spi_interface_t spi, const hal_spi_info_t& conf)
+            : SpiConfigurationLock(spi) {
+        setConfiguration(conf);
+    }
+
+    SpiConfigurationLock() = delete;
+    SpiConfigurationLock(const SpiConfigurationLock&) = delete;
+    SpiConfigurationLock& operator=(const SpiConfigurationLock&) = delete;
+
+    ~SpiConfigurationLock() = default;
```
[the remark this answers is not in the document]
avtolstoy: We could, but we wouldn't be able to use it recursively as it needs to retain the original state that eventually needs to be restored. That's why it's a single object and we use `std::lock_guard`, as disparate instances of it wouldn't be able to share the original state, or they would be restoring things multiple times unnecessarily. Splitting it into two classes: one to store retained configuration, one to do the locking is pretty much reinventing `std::lock_guard`.

hal/inc/hal_platform.h
```diff
@@ -37,7 +37,11 @@
 #define HAL_PLATFORM_WIFI 0
 #endif /* HAL_PLATFORM_WIFI */
 
-#if HAL_PLATFORM_WIFI && HAL_PLATFORM_NCP_AT
+#ifndef HAL_PLATFORM_WIFI_NCP_SDIO
+#define HAL_PLATFORM_WIFI_NCP_SDIO (0)
+#endif // HAL_PLATFORM_WIFI_NCP_SDIO
+
+#if HAL_PLATFORM_WIFI && HAL_PLATFORM_NCP_AT && !HAL_PLATFORM_WIFI_NCP_SDIO
```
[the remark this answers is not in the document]
avtolstoy: :+1: We could also leave this change for later, as of right now it's UART or SDIO.

hal/shared/spi_lock.h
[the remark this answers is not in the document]
avtolstoy: Actually, let's do that in a separate PR. There is a set of other functionality (custom clock reference etc) that complicates things.

hal/inc/hal_dynalib_wlan.h
```diff
@@ -1,38 +1,31 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef HAL_DYNALIB_WLAN_H
-#define	HAL_DYNALIB_WLAN_H
```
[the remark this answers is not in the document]
avtolstoy: Cut too much, fixed.

hal/inc/hal_platform.h
```diff
@@ -426,4 +430,16 @@
 #define HAL_PLATFORM_MULTIPART_SYSTEM (0)
 #endif // HAL_PLATFORM_MULTIPART_SYSTEM
 
+#ifndef HAL_PLATFORM_NCP_COUNT
+#define HAL_PLATFORM_NCP_COUNT (1)
```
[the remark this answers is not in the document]
avtolstoy: Done.

hal/src/tracker/esp32_sdio_stream.cpp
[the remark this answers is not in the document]
avtolstoy: Left as-is for now.

hal/shared/spi_lock.h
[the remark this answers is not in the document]
avtolstoy: Added `SPARK_ASSERT`.

hal/src/tracker/platform_ncp_quectel.cpp
```diff
@@ -52,3 +52,16 @@ PlatformNCPIdentifier platform_current_ncp_identifier() {
     }
     return (PlatformNCPIdentifier)ncpId;
 }
+
+int platform_ncp_get_info(int idx, PlatformNCPInfo* info) {
+    CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
+    CHECK_TRUE(idx > 0 && idx < platform_ncp_count(), SYSTEM_ERROR_INVALID_ARGUMENT);
+    if (idx == 0) {
```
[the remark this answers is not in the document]
avtolstoy: Done.

hal/src/tracker/sdspi/esp32_sdio.h
```diff
@@ -0,0 +1,80 @@
...
+    pin_t csPin_;
+    pin_t intrPin_;
+    os_semaphore_t sem_;
+
+    /* FIXME: SDSPI port functions do not have any kind of state arguments :( */
+    static Esp32Sdio* instance_;
+
+    EventGroupHandle_t evGroup_;
+
+    bool txInterruptSupported_;
```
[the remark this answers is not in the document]
avtolstoy: Done.

modules/shared/nRF52840/inc/system-part1/module_system_part1.inc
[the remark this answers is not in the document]
avtolstoy: Added a comment for now.

XuGuohui (approved):

PR #2251, opened 2020-12-09
Always inline fastPinGetPinmap to speed pin sets

### LINKED ISSUE #2244: Low Level Input/Output Slow on OS 2.0.0-rc4
</details>

## Bug Report
Tested with pinSetFast, the speed at which this is seting the pin high is slower on firmware targeted 2.0.0-rc4 vs the 1.5.2 firmware. Testing on a device running 2.0.0-rc4.
### Expected Behavior

Tested by printing out micros seconds taken to pinsethigh in loop. on 1.5.2 it takes <1 us.
### Observed Behavior

2.0.0rc4 it takes >1us sometimes >2us

### Test App

```c
void setup() {
  // A minimal test case to reproduce the issue is super helpful!
  Serial.begin(9600);
  pinMode(D6, OUTPUT);

}

void loop() {
      unsigned long drawTime = micros();
      pinSetFast(D6);
      drawTime = micros() - drawTime;
      Serial.println(drawTime);
      pinResetFast(D6);
}
```

## Feature/Enhancement Request

Using this to drive a parallel lcd (8-bit), the writing/reseting speed of these low level input/output is considerably slow. Causing the the lcd to run much slower.

tstellanova:
<details>
  <summary><i>Always inline fastPinGetPinmap in order to speed pin set</i></summary>

Also add test to catch unintentional slowing of pin set in the future.  
</details>

### Problem

For some users, calling `pinSetFast` or `pinResetFast` was noticeably slower on 2.0.0-rc.4 or later than on earlier releases. 
 
### Solution

The 2.0.0 development line worked around an issue with initialization order by changing the way the HAL pin map was initialized and subsequently accessed.  This introduced some overhead in each call to fastPinGetPinmap.  We reduce that overhead here by always inlining` fastPinGetPinmap()`.

### References

- Closes #2244 

### Steps to Test

Added a test to `/user/tests/wiring/no_fixture/fastpin.cpp` to set a limit to the allowable ticks.  Tested on Electron 2g. 

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/fastpin.cpp
+++ b/user/tests/wiring/no_fixture/fastpin.cpp
@@ -1,0 +1,68 @@
+
+#include "application.h"
+#include "unit-test/unit-test.h"
+
+
+
+test(FASTPIN_01_MaxDuration_PinSet) {
+    // Attempt pinSetFast and pinResetFast numerous times and check speed.
+    // Duration went from about 62 -> 24 ticks per set on Gen2 with this change list
+
+#if HAL_PLATFORM_GEN == 3
+    // expected max ticks of pinSetFast / pinResetFast on Gen3
+    const uint32_t MAX_DURATION_PINSET_TICKS = 61;
+#elif HAL_PLATFORM_GEN == 2
+    // expected max ticks of pinSetFast / pinResetFast on Gen2
+    const uint32_t MAX_DURATION_PINSET_TICKS = 24;
+#else
+#error "No gpio fastpin timing benchmark yet measured for this platform"
+#endif
+
+    const uint32_t NUM_ITERATIONS = 100;
+    uint32_t start, finish;
+
+    ATOMIC_BLOCK()  {
+        start = System.ticks();
+        for (uint32_t i = 0; i < NUM_ITERATIONS; i++) {
+            pinSetFast(D7);
+            //pinResetFast(D7);
+        }
+        finish = System.ticks();
+    }
+    uint32_t duration = finish - start;
+//    Serial.print("Set duration:");
+//    Serial.println(duration);
+    assertLessOrEqual(duration, NUM_ITERATIONS*MAX_DURATION_PINSET_TICKS);
+}
+
+test(FASTPIN_02_MaxDuration_PinReset) {
+    // Attempt pinResetFast numerous times and check speed.
+
+#if HAL_PLATFORM_GEN == 3
+    // expected max ticks of pinResetFast on Gen3
+    const uint32_t MAX_DURATION_PINRESET_TICKS = 60;
+#elif HAL_PLATFORM_GEN == 2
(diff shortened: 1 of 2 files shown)


user/tests/wiring/no_fixture/fastpin.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "application.h"
+#include "unit-test/unit-test.h"
+
+
+
+test(FASTPIN_01_MaxDuration_SetReset) {
+    // Attempt pinSetFast and pinResetFast numerous times and check speed.
+    // Duration went from 62 -> 24 ticks per set on Gen2 with this change list
+
+#if PLATFORM_ID == PLATFORM_ARGON || \
```
avtolstoy, line 11: Nit: `#if HAL_PLATFORM_GEN == 3` or `#if HAL_PLATFORM_NRF52840`. We add one more platform, forget to change this conditional and the test never runs.

Let's also add an `#else` block with `#error` so that the test fails for any new platforms and we can immediately see that it requires attention.

wiring/inc/fast_pin.h
```diff
@@ -32,7 +32,7 @@ extern "C" {
  * by @pkourany on PR: https://github.com/spark/firmware/pull/556 */
 #define USE_BIT_BAND 0
 
-inline const Hal_Pin_Info* fastPinGetPinmap() {
+__attribute__((always_inline)) inline const Hal_Pin_Info* fastPinGetPinmap() {
     static Hal_Pin_Info* pinMap = HAL_Pin_Map();
```
avtolstoy, line 36: I would change to `static const Hal_Pin_Info*` just in case.

user/tests/wiring/no_fixture/fastpin.cpp
```diff
@@ -0,0 +1,40 @@
...
+    const uint32_t MAX_DURATION_PINSET_TICKS = 24;
+#endif
+
+    const uint32_t NUM_ITERATIONS = 100;
+    uint32_t start, finish;
+
+    ATOMIC_BLOCK()  {
+        start = System.ticks();
+        for (uint32_t i = 0; i < NUM_ITERATIONS; i++) {
+            pinSetFast(D7);
```
avtolstoy, line 30: Nit: I would time these separately. I saw some variation in timings between them, might be good idea to watch out for both separately.

avtolstoy (approved):

avtolstoy:
Some of the timing constraints are too tight on some Gen 3 platforms:
```
        Assertion failed: (duration=6236) <= (NUM_ITERATIONS*MAX_DURATION_PINRESET_TICKS=6000), file tests/integration/wiring/no_fixture/fastpin.cpp, line 64.

        2) FASTPIN_02_MaxDuration_PinReset
```

I'll relax them a bit in develop.

PR #2255, opened 2020-12-14
Change the condition to disable hwfc on b5som

artyLee:
### Problem
During the B523 SoM development, the engineering samples have issues with RTS and CTS hardware flow control, we use the `hwVersion` 0x00 to filter out the PCBA version by disabling HWFC.

Now, we have another new B5SoM variant that has a new modem and the H/W version in OTP is 0x00, to distinguish the B523 and this new B5XX SoM, we have to combine `hwVersion` and `platform` in this case.

### Solution
Changing the condition to filter out the hardware flow control issue for B523 specifically.

### Steps to Test
N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -149,7 +149,7 @@ int QuectelNcpClient::init(const NcpClientConfig& conf) {
 #if PLATFORM_ID == PLATFORM_B5SOM
     uint32_t hwVersion = HW_VERSION_UNDEFINED;
     auto ret = hal_get_device_hw_version(&hwVersion, nullptr);
-    if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_B5SOM_V003) {
+    if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_B5SOM_V003 && ncpId() == PLATFORM_NCP_QUECTEL_EG91_E) {
         sconf = SERIAL_8N1;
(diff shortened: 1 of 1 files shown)


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -149,7 +149,7 @@ int QuectelNcpClient::init(const NcpClientConfig& conf) {
 #if PLATFORM_ID == PLATFORM_B5SOM
     uint32_t hwVersion = HW_VERSION_UNDEFINED;
     auto ret = hal_get_device_hw_version(&hwVersion, nullptr);
-    if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_B5SOM_V003) {
+    if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_B5SOM_V003 && ncpId() == PLATFORM_NCP_QUECTEL_EG91_E) {
```
XuGuohui, line 152: Since we use `&&` here, so just to make sure that ONLY the module that the NCP type is `PLATFORM_NCP_QUECTEL_EG91_E` is going to disable the hardware flow control.

artyLee: Yes, if the NCP type is `PLATFORM_NCP_QUECTEL_EG91_E` and meantime `hwverison` is 0x00 will be blocked hardware flow control to disable.

avtolstoy:
Closing this as OTP will be populated with hw revision > 0x00.

PR #2258, opened 2020-12-22
Wiznetif: report module power state.

XuGuohui:
### Problem
The device may timeout waiting the ethernet interface to be off() when sleep is requested and the ethernet interface is present.
### Solution
Report the Wiznetif power state to system network manager.

Note that the W5500 module is actually on still even if we've called the corresponding API to turn it off. Because there is no hardware pins to control the power to the W5500 module, nor we implement the low power mode for W5500.
### Example App
```c
#include "application.h"

SYSTEM_MODE(MANUAL);

void setup() {

    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).duration(5s));
}

void loop() {
}
```
---
### Completeness
- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/wiznet/wiznetif.cpp
+++ b/hal/network/lwip/wiznet/wiznetif.cpp
@@ -109,6 +109,7 @@ WizNetif::WizNetif(hal_spi_interface_t spi, pin_t cs, pin_t reset, pin_t interru
           cs_(cs),
           reset_(reset),
           interrupt_(interrupt),
+          pwrState_(IF_POWER_STATE_NONE),
           spiLock_(spi, WIZNET_DEFAULT_CONFIG) {
 
     LOG(INFO, "Creating Wiznet LwIP interface");
@@ -265,8 +266,10 @@ err_t WizNetif::initInterface() {
 
     hwReset();
     if (!isPresent()) {
+        pwrState_ = IF_POWER_STATE_DOWN;
         return ERR_IF;
     }
+    pwrState_ = IF_POWER_STATE_UP;
 
     return ERR_OK;
 }
@@ -345,6 +348,12 @@ int WizNetif::up() {
         down_ = false;
     }
 
+    // Just in case. We might call up() directly after powerDown().
+    if (pwrState_ != IF_POWER_STATE_UP) {
+        pwrState_ = IF_POWER_STATE_UP;
+        notifyPowerState();
+    }
+
     return r;
 }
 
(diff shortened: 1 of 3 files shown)


hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -419,16 +428,40 @@ int WizNetif::down() {
...
+    // Notify the system network manager that the module is powered down
+    // to bypass waitInterfaceOff() as required by system sleep.
+    if (pwrState_ != IF_POWER_STATE_DOWN) {
+        pwrState_ = IF_POWER_STATE_DOWN;
+        notifyPowerState();
+    }
+    return ret;
+}
+
+void WizNetif::notifyPowerState() {
```
avtolstoy, line 452: Let's add an argument here and task it with updating internal `pwrState_` as well. There shouldn't be any reason to do that externally and then still call into this method.

avtolstoy (approved):

hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -358,16 +366,39 @@ int WizNetif::down() {
...
+    // Notify the system network manager that the module is powered down
+    // to bypass waitInterfaceOff() as required by system sleep.
+    if (pwrState_ != IF_POWER_STATE_DOWN) {
+        notifyPowerState(IF_POWER_STATE_DOWN);
+    }
+    return ret;
+}
+
+void WizNetif::notifyPowerState(if_power_state_t state) {
+    pwrState_ = state;
```
avtolstoy, line 389: This could probably be:
```c++
if (pwrState_ == state) {
    return;
}
pwrState_ = state;
// Notify
```

So that you don't have to check externally.

avtolstoy (approved):

hal/network/lwip/wiznet/wiznetif.cpp
XuGuohui: That might miss a power state change event. Image that the `pwrState_ ` is changed to UP on initialization, then `powerUp()` won't notify the power state.

avtolstoy: Okay, let's leave it as-is then :+1:

XuGuohui: That's not a good example. If the `pwrState_` is changed to `IF_POWER_STATE_UP` somewhere but without notifying the state, somewhere else call `notifyPowerState(IF_POWER_STATE_UP)` won't notify the power state. Just in case.

XuGuohui: While this remind me that we should always notify the power state if possible, especially in `powerDown()` that may block the `waitInterfaceOff()`. I'll revise it a little bit.

avtolstoy (approved):

PR #2262, opened 2020-12-28
BLE: supports LESC pairing.

XuGuohui:
As the title describes.
### References

BLE core spec 5.0, Vol.3, Part H, Section 2.3.5.6.

---
### Completeness
- [ ] Added documentation

### DIFF
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -121,9 +121,9 @@ typedef enum hal_ble_pairing_io_caps_t {
 } hal_ble_pairing_io_caps_t;
 
 typedef enum hal_ble_pairing_algorithm_t {
-    BLE_PAIRING_ALGORITHM_LEGACY_ONLY = 0,
-    BLE_PAIRING_ALGORITHM_LESC_ONLY = 1,
-    BLE_PAIRING_ALGORITHM_AUTO = 2
+    BLE_PAIRING_ALGORITHM_AUTO = 0,
+    BLE_PAIRING_ALGORITHM_LEGACY_ONLY = 1,
+    BLE_PAIRING_ALGORITHM_LESC_ONLY = 2
 } hal_ble_pairing_algorithm_t;
 
 typedef enum hal_ble_evts_type_t {
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -42,7 +42,7 @@ extern "C" {
  * @{
  */
 
-#define BLE_API_VERSION 1
+#define BLE_API_VERSION 2
 
 // Particle's company ID
 #define PARTICLE_COMPANY_ID     0x0662
@@ -120,6 +120,12 @@ typedef enum hal_ble_pairing_io_caps_t {
     BLE_IO_CAPS_KEYBOARD_DISPLAY = 4
 } hal_ble_pairing_io_caps_t;
 
+typedef enum hal_ble_pairing_algorithm_t {
+    BLE_PAIRING_ALGORITHM_LEGACY_ONLY = 0,
+    BLE_PAIRING_ALGORITHM_LESC_ONLY = 1,
+    BLE_PAIRING_ALGORITHM_AUTO = 2
+} hal_ble_pairing_algorithm_t;
+
 typedef enum hal_ble_evts_type_t {
     BLE_EVT_UNKNOWN = 0x00,
     BLE_EVT_ADV_STOPPED = 0x01,
@@ -137,6 +143,7 @@ typedef enum hal_ble_evts_type_t {
     BLE_EVT_PAIRING_PASSKEY_DISPLAY = 0x0D,
     BLE_EVT_PAIRING_PASSKEY_INPUT = 0x0E,
     BLE_EVT_PAIRING_STATUS_UPDATED = 0x0F,
+    BLE_EVT_PAIRING_NUMERIC_COMPARISON = 0x10,
     BLE_EVT_MAX = 0x7FFFFFFF
 } hal_ble_evts_type_t;
 
@@ -397,7 +404,8 @@ typedef struct hal_ble_pairing_config_t {
     uint16_t version;
     uint16_t size;
     hal_ble_pairing_io_caps_t io_caps;
-    uint8_t reserved[3];
+    hal_ble_pairing_algorithm_t algorithm;
+    uint8_t reserved[2];
 } hal_ble_pairing_config_t;
 
 /**
@@ -834,6 +842,15 @@ int hal_ble_gap_get_rssi(hal_ble_conn_handle_t conn_handle, void* reserved);
 int hal_ble_gap_set_pairing_config(const hal_ble_pairing_config_t* config, void* reserved);
 
 /**
+ * Get pairing configurations
+ *
+ * @param[in]   config      Pointer to where it stores the configuration.
+ *
+ * @returns     0 on success, system_error_t on error.
+ */
+int hal_ble_gap_get_pairing_config(hal_ble_pairing_config_t* config, void* reserved);
+
+/**
  * Start pairing with the peer device.
  *
  * @param[in]   conn_handle BLE connection handle.
@@ -859,6 +876,16 @@ int hal_ble_gap_reject_pairing(hal_ble_conn_handle_t conn_handle, void* reserved
  * @returns     0 on success, system_error_t on error.
  */
 int hal_ble_gap_set_pairing_passkey(hal_ble_conn_handle_t conn_handle, const uint8_t* passkey, void* reserved);
+
+/**
+ * Confirm the LESC numeric comparision result.
+ *
+ * @param[in]   conn_handle BLE connection handle.
+ * @param[in]   equal       True if the numeric comparison is equal, otherwise, false.
+ *
+ * @returns     0 on success, system_error_t on error.
+ */
+int hal_ble_gap_set_lesc_numeric_comparison(hal_ble_conn_handle_t conn_handle, bool equal, void* reserved);
 
 /**
  * Check if pairing with peer device is in progress
--- a/hal/inc/ble_hal.h
+++ b/hal/inc/ble_hal.h
@@ -126,6 +126,13 @@ typedef enum hal_ble_pairing_algorithm_t {
     BLE_PAIRING_ALGORITHM_LESC_ONLY = 2
 } hal_ble_pairing_algorithm_t;
 
+typedef enum hal_ble_pairing_auth_data_type_t {
+    BLE_PAIRING_AUTH_DATA_NUMERIC_COMPARISON = 0,
+    BLE_PAIRING_AUTH_DATA_PASSKEY = 1,
+    BLE_PAIRING_AUTH_DATA_LEGACY_OOB = 2,
+    BLE_PAIRING_AITH_DATA_LESC_OOB = 3
+} hal_ble_pairing_auth_data_type_t;
+
 typedef enum hal_ble_evts_type_t {
     BLE_EVT_UNKNOWN = 0x00,
     BLE_EVT_ADV_STOPPED = 0x01,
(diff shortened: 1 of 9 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -741,6 +775,8 @@ os_thread_return_t BleObject::BleEventDispatcher::processBleEventFromThread(void
                 case BLE_GATTC_EVT_HVX: {
                     BleObject::getInstance().gattc()->processDataNotifiedEventFromThread(event);
                 }
```
mgolu, line 777: When getting a notification, it is calling both `processDataNotifiedEventFromThread(event)` and `processSecurityEventFromThread(event)`

```suggestion
                    break;
                }
```

XuGuohui: Good catch! 👍

mgolu (approved):

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1673,6 +1709,11 @@ int BleObject::ConnectionsManager::init() {
         LOG(ERROR, "os_timer_create() failed.");
         goto error;
     }
+    if (generateEcdhKeyPair() != SYSTEM_ERROR_NONE) {
```
avtolstoy, line 1712: We should reconsider generating the local keypair only once. The newest BLE spec has the following:

> To protect a device's private key, a device should implement a method to
> prevent an attacker from retrieving useful information about the device's private
> key. For this purpose, a device should change its private key after every pairing
> (successful or failed). Otherwise, it should change its private key whenever S +
> 3F > 8, where S is the number of successful pairings and F the number of
> failed attempts since the key was last changed.

(BLUETOOTH CORE SPECIFICATION Version 5.2 | Vol 3, Part H 2.3.6 Repeated attempts)

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -2325,24 +2456,40 @@ int BleObject::ConnectionsManager::processSecurityEventFromThread(const ble_evt_
...
+                keyset.keys_peer.p_pk = connection->peerPublicKey.get();
+                pKeyset = &keyset;
             }
         }
+        if (!reject && connection->info.role == BLE_ROLE_PERIPHERAL) {
+            initSecParams(pairingConfig_, secParams);
+            pSecParams = &secParams;
+        }
+        ret = sd_ble_gap_sec_params_reply(event->evt.gap_evt.conn_handle, reject ? BLE_GAP_SEC_STATUS_PAIRING_NOT_SUPP : BLE_GAP_SEC_STATUS_SUCCESS, pSecParams, pKeyset);
         if (ret != NRF_SUCCESS) {
```
avtolstoy, line 2489: Or if `rejected == true` ?

avtolstoy (changes requested):

avtolstoy:
Good job :tada:

hal/src/nRF52840/ble_hal.cpp
XuGuohui: `ret == NRF_SUCCESS` will result `BLE_GAP_EVT_AUTH_STATUS` event and followed by doing something properly as well.

hal/inc/ble_hal.h
```diff
@@ -860,6 +877,16 @@ int hal_ble_gap_reject_pairing(hal_ble_conn_handle_t conn_handle, void* reserved
...
 
+/**
+ * Confirm the LESC numeric comparision result.
+ *
+ * @param[in]   conn_handle BLE connection handle.
+ * @param[in]   equal       True if the numeric comparison is equal, otherwise, false.
+ *
+ * @returns     0 on success, system_error_t on error.
+ */
+int hal_ble_gap_set_lesc_numeric_comparison(hal_ble_conn_handle_t conn_handle, bool equal, void* reserved);
```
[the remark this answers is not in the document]
XuGuohui: I know that you were thinking we can reject the pairing request if the numeric comparison failed, or leave the device going forward to finish the pairing process. But the fact is that user need to perform action to the numeric comparison explicitly. And without invoking this API, the SoftDevice will consider a timeout and terminate the pairing process. Besides, by invoking this API, the SoftDevice either proceed  or terminate the pairing process immediately.

While I agree that for future OOB authentication requiring OOB data, we can use a more generic API to set passkey or OOB data.

hal/src/nRF52840/ble_hal.cpp
XuGuohui: Will do in this way: 

> a device should change its private key after every pairing
(successful or failed)

hal/inc/ble_hal.h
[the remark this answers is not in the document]
avtolstoy: I was more proposing something along the lines of:
```c++
int hal_ble_gap_set_auth_status(hal_ble_conn_handle_t conn_handle, hal_ble_auth_type_t type, int result, void* data, void* reserved);
```

for future extensibility (take the example above with a grain of salt, I haven't looked into OOB, so don't know how to generalize this best).

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1673,6 +1709,11 @@ int BleObject::ConnectionsManager::init() {
         LOG(ERROR, "os_timer_create() failed.");
         goto error;
     }
+    if (generateEcdhKeyPair() != SYSTEM_ERROR_NONE) {
+        LOG(ERROR, "generateEcdhKeyPair() failed.");
```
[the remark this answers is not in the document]
XuGuohui: Not really if legacy pairing is used.

hal/inc/ble_hal.h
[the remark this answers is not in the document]
XuGuohui: Since we have exposed `hal_ble_gap_set_pairing_passkey()` in the dynalib in previous releases, introducing a more generic API for user set authentication info seams to be duplicated. For the OOB authentication, user need to set the OOB data. It's similar with setting a passkey, but the length is different. Do you think it's fine to deprecate the `hal_ble_gap_set_pairing_passkey()` API?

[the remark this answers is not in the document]
avtolstoy: No strong preferences honestly, but if we can generalize this, we probably should. Not the first time we are deprecating things like this and as long as the old function is kept in the dynalib and delegates to the new function it's all good.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -2545,6 +2536,15 @@ int BleObject::ConnectionsManager::processSecurityEventFromThread(const ble_evt_
...
         linkEvent.params.pairing_status.lesc = authStatus.lesc;
         notifyLinkEvent(linkEvent);
+        if (authStatus.lesc) {
+            /* To protect a device's private key, a device should implement a method to
+            * prevent an attacker from retrieving useful information about the device's private
+            *  key. For this purpose, a device should change its private key after every pairing
+            * (successful or failed). Otherwise, it should change its private key whenever S +
+            * 3F > 8, where S is the number of successful pairings and F the number of
+            * failed attempts since the key was last changed. */
+            generateEcdhKeyPair();
```
avtolstoy, line 2546: This is not going to work because `ecdhContextInit_` is not reset anywhere. We should probably just remove the `ecdhContextInit_` state variable and generate new keys every time in `publicKeysPrepare()` perhaps?

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -2206,9 +2197,9 @@ BleObject::ConnectionsManager::BleConnection* BleObject::ConnectionsManager::fet
     return nullptr;
 }
 
-int BleObject::ConnectionsManager::addConnection(const BleConnection& connection) {
+int BleObject::ConnectionsManager::addConnection(BleConnection& connection) {
```
avtolstoy, line 2200: This is confusing semantically. This method intends to take ownership of `connection`, so it should take it by rvalue reference: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter

avtolstoy (changes requested):

hal/src/nRF52840/ble_hal.cpp
XuGuohui: `ecdhContextInit_` is reset once enter the `generateEcdhKeyPair()` and set if new ECDH key pair is generated successfully. `ecdhContextInit_` is used elsewhere to check the ECDH key pair validity in prior to invoking the key, such as computing the DHKey.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -2035,22 +2033,22 @@ int BleObject::ConnectionsManager::rejectPairing(hal_ble_conn_handle_t connHandl
     CHECK_TRUE(connection, SYSTEM_ERROR_NOT_FOUND);
     switch (connection->pairState) {
         case BLE_PAIRING_STATE_SEC_REQ_RECEIVED: {
-            connection->pairState = BLE_PAIRING_STATE_REJECTED;
             int ret = sd_ble_gap_authenticate(connection->info.conn_handle, nullptr);
             CHECK_NRF_RETURN(ret, nrf_system_error(ret));
+            connection->pairState = BLE_PAIRING_STATE_REJECTED;
```
avtolstoy, line 2038: Can you add a comment on why we are no longer ignoring errors for setting the internal pairing state to rejected?

hal/inc/ble_hal.h
```diff
@@ -312,6 +319,14 @@ typedef struct hal_ble_pairing_status_updated_evt_t {
     uint8_t reserved[3];
 } hal_ble_pairing_status_updated_evt_t;
 
+typedef struct hal_ble_pairing_auth_data_t {
+    hal_ble_pairing_auth_data_type_t type;
```
avtolstoy, line 323: Might be a good idea to add a size field just in case.

hal/inc/ble_hal.h
```diff
@@ -312,6 +319,14 @@ typedef struct hal_ble_pairing_status_updated_evt_t {
...
+    union {
+        bool equal; //For numeric comparison result
+        const uint8_t* data; // For passkey and OOB data
```
avtolstoy, line 326: Might be a good idea to wrap this into a struct or a pointer to a struct?

avtolstoy (approved):

hal/inc/ble_hal.h
XuGuohui: For passkey and legacy OOB data, they are plain data with fixed length, for LESC OOB data, we can static cast the `uint8*` to corresponding struct pointer as well. Maybe extend the union struct like this?:
```
typedef struct hal_ble_pairing_auth_data_t {
    hal_ble_pairing_auth_data_type_t type;
    uint16_t size;
    union {
        bool equal; //For numeric comparison result
        const char* passkey; // For passkey
        const uint8_t* legacy_oob; // For legacy OOB data
        const uint8_t* lesc_oob; // For LESC OOB data
    } params;
} hal_ble_pairing_auth_data_t;
```

hal/src/nRF52840/ble_hal.cpp
XuGuohui: Will add: 
>We should not tamper the pairing state to be out of sync with the SoftDevice. Only if the SD call returns success we then mark the pairing state as rejected.

hal/inc/ble_hal.h
avtolstoy: :+1:

hal/src/nRF52840/ble_hal.cpp
avtolstoy: :+1:

hal/inc/ble_hal.h
```diff
@@ -321,9 +321,12 @@ typedef struct hal_ble_pairing_status_updated_evt_t {
 
 typedef struct hal_ble_pairing_auth_data_t {
     hal_ble_pairing_auth_data_type_t type;
+    uint16_t size; // payload length of params
```
avtolstoy, line 324: Let's rename this to `params_size` or something like that. It's very easy to confuse with the length of `hal_ble_pairing_auth_data_t`, which I was actually initially suggesting for extensibility in case we ever have to add something else to this struct :)

avtolstoy (approved):

hal/inc/ble_hal.h
XuGuohui: In that case, we'd better add the version field as well. Because the new added member is not present in old user application and this member is not initialized, which will lead to unknown behavior.

avtolstoy: We still have the type field which has somewhat a similar purpose.

XuGuohui: Just in case that the new added member is not type related.

avtolstoy (approved):

PR #2267, opened 2021-01-19
AT parser and gsm0710 muxer custom log category support

avtolstoy:
### Problem

Now that we have ESP32 support on Tracker, it's impossible to differentiate between NCP client / gsm0710 muxer logs that come from the cellular modem and ESP32.

### Solution

Add instance-local custom log categories support for AT parser and gsm0710muxer.

This PR also fixes a bug in `ChannelStream` in `waitEvent()` timeout handling and makes sure that we have a single implementation of the gsm0710muxer (by having a common Stream base class that support `eventGroup()` method).

### Steps to Test

Run the test app, watch logs. It should be easy to differentiate NCP/muxer logs that come from the modem and from ESP32.

### Example App

```c++

SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {
    Particle.connect();
}

void loop() {
    WiFiAccessPoint results[5];
    WiFi.scan(results, sizeof(results) / sizeof(results[0]));
    auto sig = Cellular.RSSI();
    (void)sig;
    delay(5000);
}
```

### Logs

```
0000034741 [ncp.at] TRACE: > AT+CCID
0000034743 [ncp.at] TRACE: < +CCID: XXXXXX
0000034743 [ncp.at] TRACE: < OK
0000034745 [ncp.at] TRACE: > AT+CGSN
0000034746 [ncp.at] TRACE: < YYYYYYY
0000034746 [ncp.at] TRACE: < OK
0000034753 [ncp.esp32.at] TRACE: > AT+MVER
0000034763 [ncp.esp32.at] TRACE: < 9
0000034763 [ncp.esp32.at] TRACE: < OK
```

```
0000029981 [mux] INFO: Starting GSM07.10 muxer
0000029981 [mux] INFO: Openning mux channel 0
0000029982 [mux] INFO: GSM07.10 muxer thread started
0000030083 [mux] INFO: Openning mux channel 1
0000022002 [ncp.esp32.mux] INFO: Starting GSM07.10 muxer
0000022004 [ncp.esp32.mux] INFO: Openning mux channel 0
0000022004 [ncp.esp32.mux] INFO: GSM07.10 muxer thread started
0000022010 [ncp.esp32.mux] INFO: Openning mux channel 1
```

### References

- https://github.com/particle-iot/gsm0710muxer/pull/8

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp/at_parser/at_parser.h
+++ b/hal/network/ncp/at_parser/at_parser.h
@@ -18,6 +18,7 @@
 #pragma once
 
 #include <memory>
+#include "c_string.h"
 
 namespace particle {
 
@@ -77,6 +78,12 @@ public:
      * @see `logEnabled()`
      */
     static const auto DEFAULT_LOG_ENABLED = true;
+    /**
+     * Default logging category
+     *
+     * @see `logCategory()`
+     */
+    static constexpr auto DEFAULT_LOG_CATEGORY = "ncp.at";
 
     /**
      * Constructs a settings object with all parameters set to their default values.
@@ -176,6 +183,22 @@ public:
      * @see `DEFAULT_LOG_ENABLED`
      */
     bool logEnabled() const;
+    /**
+     * Sets the logging category.
+     *
+     * @return This settings object.
+     *
+     * @see `DEFAULT_LOG_CATEGORY`
+     */
+    AtParserConfig& logCategory(const char* category);
+    /**
+     * Returns the logging category.
+     *
+     * @return Logging category.
+     *
+     * @see `DEFAULT_LOG_CATEGORY`
+     */
+    const char* logCategory() const;
 
 private:
     Stream* strm_;
@@ -184,6 +207,7 @@ private:
     unsigned strmTimeout_;
     bool echoEnabled_;
     bool logEnabled_;
+    CString logCategory_;
 };
 
 /**
@@ -444,6 +468,15 @@ inline AtParserConfig& AtParserConfig::logEnabled(bool enabled) {
 
 inline bool AtParserConfig::logEnabled() const {
     return logEnabled_;
+}
+
+inline AtParserConfig& AtParserConfig::logCategory(const char* category) {
+    logCategory_ = category;
+    return *this;
+}
+
+inline const char* AtParserConfig::logCategory() const {
+    return logCategory_ ? static_cast<const char*>(logCategory_) : DEFAULT_LOG_CATEGORY;
 }
 
 } // particle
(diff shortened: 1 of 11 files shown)


hal/network/ncp/at_parser/at_parser.h
```diff
@@ -446,4 +470,13 @@ inline bool AtParserConfig::logEnabled() const {
     return logEnabled_;
 }
 
+inline AtParserConfig& AtParserConfig::logCategory(const char* category) {
+    logCategory_ = category;
+    return *this;
+}
+
+inline const char* AtParserConfig::logCategory() const {
+    return logCategory_ ? static_cast<const char*>(logCategory_) : DEFAULT_LOG_CATEGORY;
```
eberseth, line 479: Why not use the parenthesis operator?

avtolstoy: `CString` is not a functor, it doesn't have an `operator()` it has a cast operator to `const char*`: https://en.cppreference.com/w/cpp/language/cast_operator. Ternary if also requires both variants to be of the same type, so an explicit case is required here, otherwise the compiler will throw an error.

hal/src/nRF52840/gsm0710muxer/channel_stream.h
```diff
@@ -217,10 +217,11 @@ inline int MuxerChannelStream<MuxerT>::waitEvent(unsigned flags, unsigned timeou
         if (f &= flags) {
             break;
         }
-        if (timeout > 0 && HAL_Timer_Get_Milli_Seconds() - t >= timeout) {
+        const auto elapsed = HAL_Timer_Get_Milli_Seconds() - t;
+        if (timeout == 0 || elapsed >= timeout) {
```
eberseth, line 221: Was this a bug before?  The logic for this conditional has changed.

avtolstoy: > This PR also fixes a bug in ChannelStream in waitEvent() timeout handling

hal/src/nRF52840/gsm0710muxer/channel_stream.h
```diff
@@ -217,10 +217,11 @@ inline int MuxerChannelStream<MuxerT>::waitEvent(unsigned flags, unsigned timeou
...
             return SYSTEM_ERROR_TIMEOUT;
         }
-        os_semaphore_take(sem_, HAL_Timer_Get_Milli_Seconds() - t, false);
+        os_semaphore_take(sem_, timeout - elapsed, false);
```
eberseth, line 224: What if `timeout == 0` ?  Will the `os_semaphore_take()` call handle this?

avtolstoy: If it's zero, we won't reach this call at all an will return `SYSTEM_ERROR_TIMEOUT` above.

eberseth (approved):

PR #2268, opened 2021-01-19
[gen2] speed up cellular modem powerOn time

technobly:
### Problem

We recently increased the AT_TIMEOUT from 1 to 3 seconds, which is causing a few extra seconds of power on time.

### Solution

- Set AT_TIMEOUT for _powerOn() function back to 1s, and use 3s for everywhere else when modem is on.
- Also increase number of tries from 10 to 25 to keep the power on delay to roughly 30 seconds before attempting to hard reset the modem.

### Steps to Test

- wiring/no_fixture
- can also simulate modem not responsive by changing baudrate in mdm_hal.cpp to something other than 115200
- monitor cold and hard boot times

---

### Completeness

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -63,7 +63,8 @@ std::recursive_mutex mdm_mutex;
 #define PDP_CONTEXT 1
 
 // Timeouts for various AT commands based on R4 & U2/G3 AT command manual as of Jan 2019
-#define AT_TIMEOUT        (  3 * 1000) /* When modem not responsive on boot, AT/OK tried 10x for 30s before hard reset */
+#define AT_TIMEOUT        (  3 * 1000) /* Timeout set a bit lower than 10s for quicker detection of AT/OK failure */
+#define AT_TIMEOUT_POWERON (  1 * 1000) /* When modem not responsive on boot, AT/OK tried 25x (for ~30s) before hard reset */
 #define USOCL_UDP_TIMEOUT ( 10 * 1000) /* 120s for R4 (TCP only, optimizing for UDP with 10s), 1s for U2/G3 */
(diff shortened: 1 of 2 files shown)


hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -367,10 +368,10 @@ int MDMParser::sendFormattedWithArgs(const char* format, va_list args) {
     return n;
 }
 
-int MDMParser::_checkAtResponse(void)
+int MDMParser::_checkAtResponse(bool fastTimeout /* = false */)
```
avtolstoy, line 371: Nit: Perhaps we could simply just take the timeout value here with a default set to `AT_TIMEOUT`

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -838,7 +839,7 @@ bool MDMParser::_powerOn(void)
     bool continue_cancel = false;
     bool retried_after_reset = false;
 
-    int i = 10;
+    int i = 25; // When modem not responsive on boot, AT/OK tried 25x (for ~30s) before hard reset
```
avtolstoy, line 842: Nit: could get red of these magic numbers and use constants since we are touching this, but can leave as-is.

avtolstoy (approved):

hal/src/electron/modem/mdm_hal.cpp
technobly: It's a good callout, and I though about it, but I would have had to move all of the defines to mdm_hal.h... and I thought it might be simpler not to :D

avtolstoy: A default argument can also be specified in the implementation and since these are private methods there shouldn't be any difference (there would be if it was a public method, where the default argument would only be coming into play just in the cpp file that has it and the external code wouldn't be able to use it). Fine to leave as-is :)

PR #2269, opened 2021-01-20
[Gen 3] Fix warm boot; [Tracker] Cache ESP32 firmware version

avtolstoy:
### Problem

1. We've introduced a regression in https://github.com/particle-iot/device-os/pull/2217 and broke warm boot
2. During handshake with the cloud we need to send ESP32 NCP firmware version as it is one of the module, so it's present in the system describe. On Tracker during the cloud connection establishment it's not guaranteed that ESP32 is on, so we have to turn it on in order to query for its version, which prolongs the handshake process. What complicates things even more is that since https://github.com/particle-iot/device-os/pull/2217 we'll also watch the current NCP state closely and power off immediately, so several sequential calls to query ESP32 NCP firmware version may have to power on the ESP32 and initialize it multiple times

### Solution

1. Keep the NCP state as-is unless some command (on, off, up, down) is submitted
2. Cache ESP32 NCP firmware version in non-volatile memory and read it from there

### Steps to Test

1. Use an app with `SEMI_AUTOMATIC` modem and suspend the connection process for some time. Wait until the device connects to the cloud, reset with a button, watch the logs - it should reconnect without powering off the NCP
2. Watch the logs, make sure that ESP32 firmware version is taken out of cache: `Cached ESP32 NCP firmware version`
3. Update ESP32 NCP OTA, watch the logs, make sure that ESP32 firmware version changes and gets taken out of cache.

For step 3, the following 0.0.8 ESP32 NCP firmware can be used: [tracker-esp32-ncp@0.0.8.zip](https://github.com/particle-iot/device-os/files/5839490/tracker-esp32-ncp%400.0.8.zip)

### Example App

N/A

### References

- [CH71350]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp/wifi/platform_ncp_update.cpp
+++ b/hal/network/ncp/wifi/platform_ncp_update.cpp
@@ -98,9 +98,9 @@ int invalidateWifiNcpVersionCache() {
 }
 
 int getWifiNcpFirmwareVersion(uint16_t* ncpVersion) {
+    uint16_t version = 0;
 #if HAL_PLATFORM_NCP_COUNT > 1
     using namespace particle::services;
-    uint16_t version = 0;
     int res = SystemCache::instance().get(SystemCacheKey::WIFI_NCP_FIRMWARE_VERSION, &version, sizeof(version));
     if (res == sizeof(version)) {
         LOG(TRACE, "Cached ESP32 NCP firmware version: %d", (int)version);
--- a/hal/network/ncp/wifi/platform_ncp_update.cpp
+++ b/hal/network/ncp/wifi/platform_ncp_update.cpp
@@ -15,10 +15,13 @@
  * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
+#undef LOG_COMPILE_TIME_LEVEL
+
 #include "platform_ncp.h"
 #include "network/ncp/wifi/ncp.h"
 #include "network/ncp/wifi/wifi_network_manager.h"
 #include "network/ncp/wifi/wifi_ncp_client.h"
+#include "network/ncp/ncp_client.h"
 #include "led_service.h"
 #include "check.h"
 #include "scope_guard.h"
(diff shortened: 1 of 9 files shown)


hal/network/ncp/wifi/platform_ncp_update.cpp
```diff
@@ -114,14 +163,10 @@ int platform_ncp_fetch_module_info(hal_system_info_t* sys_info, bool create) {
...
-                SPARK_ASSERT(ncpClient);
-                CHECK(ncpClient->on());
-                uint16_t version;
-                int error = ncpClient->getFirmwareModuleVersion(&version);
-                if (error) {
-                    version = 0;
-                }
+                uint16_t version = 0;
+                // Defaults to zero in case of failure
+                getWifiNcpFirmwareVersion(&version);
```
XuGuohui, line 168: How about do this in the `Esp32NcpNetif::loop` before entering the while loop like we do to the Argon to fetch the MAC address? Just to be consistence to initialize things in one place.

XuGuohui (approved):

hal/network/ncp/wifi/platform_ncp_update.cpp
avtolstoy: This method will get called multiple times throughout the device runtime, so this is the appropriate place to look into cache first and then fallback to calling into ESP32 itself to get the version (and update the cache).

If we are thinking of checking/updating the cache on early boot as we do with a MAC address on an Argon, this is disabled on Tracker already, so that we don't unnecessarily boot ESP32 every time we boot Device OS. Eventually using the caching that was implemented here we'll do the same for MAC address. So, I'd prefer to keep things on as-needed basis.

services/inc/system_cache.h
```diff
@@ -0,0 +1,50 @@
...
+    WIFI_NCP_FIRMWARE_VERSION = 0x0000
+};
+
+class SystemCache {
+public:
+    static SystemCache& instance();
+
+    // Just straight up proxying to TlvFile
+    // TODO: add support for multiple values for a key?
+    int get(SystemCacheKey key, void* value, size_t length);
```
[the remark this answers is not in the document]
avtolstoy: Can't do `const` as `TlvFile::get()` modifies the internal state and cannot be const either, unless we make some member variables mutable.

PR #2272, opened 2021-01-29
System power management inprovements

XuGuohui:
### Problem
The battery state is not reported consistently.
### Solution
1. When charging is enabled (through the `HAL_POWER_CHARGE_STATE_DISABLE` flag), we debounce the CHARGING/CHARGED/CHARGING/CHARGED events.
2. When charging is disabled, we use the `VCELL` of the fuel gauge to detect the battery state.
### Steps to Test
1. Flashing the attached application to the platforms with PMIC and fuel gauge installed
2. Attach/detach batteries with different capacity and observe the reported battery state.
3. Plug/unplug the power source and observe the reported battery state, as well as the reported power source
### Example App

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

static system_tick_t loopTick = 0;

};
  "usb otg", "battery"
};

void setup() {
  Serial1.begin(115200);
  Serial1.printlnf("Application started.");

  System.on(battery_state, [](system_event_t event, int data) -> void {
    Serial1.printlnf("battery_state: %s", batteryStates[std::max(0, data)]);
  });

  System.on(power_source, [](system_event_t event, int data) -> void {
    Serial1.printlnf("power_source: %s", powerSources[std::max(0, data)]);
  });

  System.on(low_battery, [](system_event_t event, int data) -> void {
    Serial1.printlnf("low_battery: %0.2f", System.batteryCharge());
  });

  Particle.connect();
}

void loop() {
  if (Serial1.available()) {
    char c = (char)Serial1.read();
    // Commands are case sensitive
    switch (c) {
      case 'g': {
        auto cfg = System.getPowerConfiguration();
        Serial1.printlnf("cmd, getconfig, %d, %d, %d, %d, %u, %u, %u, %u",
          cfg.isFeatureSet(SystemPowerFeature::PMIC_DETECTION) ? 1 : 0,
          cfg.isFeatureSet(SystemPowerFeature::USE_VIN_SETTINGS_WITH_USB_HOST) ? 1 : 0,
          cfg.isFeatureSet(SystemPowerFeature::DISABLE) ? 1 : 0,
          cfg.isFeatureSet(SystemPowerFeature::DISABLE_CHARGING) ? 1 : 0,
          cfg.powerSourceMinVoltage(),
          cfg.powerSourceMaxCurrent(),
          cfg.batteryChargeVoltage(),
          cfg.batteryChargeCurrent()
        );
        break;
      }
      case '0': { // Disable charging
        int ret = System.setPowerConfiguration(System.getPowerConfiguration().feature(SystemPowerFeature::DISABLE_CHARGING));
        Serial1.printlnf("Disable charging, %d", ret);
        break;
      }
      case '1': { // Enable charging
        int ret = System.setPowerConfiguration(System.getPowerConfiguration().clearFeature(SystemPowerFeature::DISABLE_CHARGING));
        Serial1.printlnf("Enable charging, %d", ret);
        break;
      }
      default: {
        break;
      }
    }
  }

  auto now = millis();
  if ((now - loopTick) >= 2000) {
    loopTick = now;

    FuelGauge fuel;
    int powerSource = System.powerSource();
    int batteryState = System.batteryState();
    float batterySoc = System.batteryCharge();
    Serial1.printlnf("power: %s, batt: %s, SoC: %0.2f, Vol: %0.2f",
      powerSources[std::max(0, powerSource)], batteryStates[std::max(0, batteryState)],
      batterySoc, fuel.getVCell());
  }
}
```
### Completeness
- [ ] Added documentation

### DIFF
--- a/system/src/system_power_manager.cpp
+++ b/system/src/system_power_manager.cpp
@@ -45,8 +45,8 @@ namespace {
 constexpr uint8_t BQ24195_VERSION = 0x23;
 
 // For deducing battery satte
-constexpr system_tick_t BATTERY_STATE_CHECK_LONG_PERIOD = 10000;
-constexpr system_tick_t BATTERY_STATE_CHECK_SHORT_PERIOD = 1000;
+constexpr system_tick_t BATTERY_STATE_NORMAL_CHECK_PERIOD = 10000;
+constexpr system_tick_t BATTERY_STATE_CHANGE_CHECK_PERIOD = 1000;
 constexpr system_tick_t BATTERY_CHARGED_MUTE_WINDOW = 1000;
 constexpr system_tick_t BATTERY_DISABLE_CHARGING_SUPRESS_PERIOD = 5000; // How long after disable charging the VCELL can reliably reflect the VBAT
 constexpr float BATTERY_CONNECTED_VCELL_THRESHOLD = 2.0; // v
@@ -194,7 +194,7 @@ void PowerManager::handleCharging(bool forceDisable) {
       chargingEnabled_ = false;
       disableChargingTimeStamp_ = millis();
       // Request to shorten the monitor period
-      battMonitorPeriod_ = BATTERY_STATE_CHECK_SHORT_PERIOD;
+      battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
     }
     return;
   }
@@ -206,7 +206,7 @@ void PowerManager::handleCharging(bool forceDisable) {
     power.enableSafetyTimer();
     chargingEnabled_ = true;
     // Request to shorten the monitor period
-    battMonitorPeriod_ = BATTERY_STATE_CHECK_SHORT_PERIOD;
+    battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
   }
 }
 
@@ -233,9 +233,6 @@ void PowerManager::handleUpdate() {
   }
 
   const uint8_t status = power.getSystemStatus();
-
-  battery_state_t state = BATTERY_STATE_UNKNOWN;
-
   const uint8_t pwr_good = (status >> 2) & 0b01;
 
   // Deduce current battery state
@@ -246,15 +243,10 @@ void PowerManager::handleUpdate() {
       if (chrg_stat == 0b11) {
         lastChargedTimeStamp_ = millis();
         batteryStateForwardingTo(BATTERY_STATE_CHARGED);
-      } else if (g_batteryState != BATTERY_STATE_DISCONNECTED && g_batteryState != BATTERY_STATE_UNKNOWN) {
-        state = BATTERY_STATE_CHARGING;
       } else {
         // We might receive continuous interrupt. Counting the charging state may lead to fake charging state.
         batteryStateForwardingTo(BATTERY_STATE_CHARGING, false);
       }
-    } else if (config_.flags & HAL_POWER_CHARGE_STATE_DISABLE) {
-      // The PMIC didn't update the register, we update the battery state manually
-      state = BATTERY_STATE_NOT_CHARGING;
     }
     // Else charging is disabled, the register is just not updated, do nothing
   } else {
@@ -263,24 +255,13 @@ void PowerManager::handleUpdate() {
     const uint8_t bat_fault = (curFault >> 3) & 0b01;
     // const uint8_t ntc_fault = curFault & 0b111;
     if (bat_fault) {
-      state = BATTERY_STATE_FAULT;
+      confirmBatteryState(g_batteryState, BATTERY_STATE_FAULT);
     } else if (!pwr_good) {
-      state = BATTERY_STATE_DISCHARGING;
-    } else if (g_batteryState != BATTERY_STATE_DISCONNECTED && g_batteryState != BATTERY_STATE_UNKNOWN) {
-      state = BATTERY_STATE_NOT_CHARGING;
+      confirmBatteryState(g_batteryState, BATTERY_STATE_DISCHARGING);
     } else {
       // We might receive continuous interrupt. Counting the not charging state may lead to fake not charging state.
       batteryStateForwardingTo(BATTERY_STATE_NOT_CHARGING, false);
     }
-  }
-
-  if (g_batteryState == BATTERY_STATE_DISCONNECTED && state == BATTERY_STATE_NOT_CHARGING) {
-    // We are aware of the fact that charging has been disabled, stay in disconnected state
-    state = BATTERY_STATE_DISCONNECTED;
-  }
-
-  if (state != BATTERY_STATE_UNKNOWN) {
-    confirmBatteryState(g_batteryState, state);
   }
 
   power_source_t src = g_powerSource;
@@ -406,7 +387,7 @@ void PowerManager::loop(void* arg) {
     HAL_USB_Set_State_Change_Callback(usbStateChangeHandler, (void*)self, nullptr);
 #endif
 
-    self->battMonitorPeriod_ = BATTERY_STATE_CHECK_LONG_PERIOD;
+    self->battMonitorPeriod_ = BATTERY_STATE_NORMAL_CHECK_PERIOD;
     self->chargingEnabled_ = power.isChargingEnabled();
   }
 
@@ -577,7 +558,7 @@ void PowerManager::confirmBatteryState(battery_state_t from, battery_state_t to)
 }
 
 void PowerManager::batteryStateForwardingTo(battery_state_t targetState, bool count) {
-  if (targetState == BATTERY_STATE_NOT_CHARGING) {
+  if (targetState == BATTERY_STATE_NOT_CHARGING && g_batteryState != BATTERY_STATE_NOT_CHARGING) {
     if (count) {
       notChargingDebounceCount_++;
     } else {
@@ -586,7 +567,7 @@ void PowerManager::batteryStateForwardingTo(battery_state_t targetState, bool co
     chargedFaultCount_ = 0;
     chargingDebounceCount_ = 0;
     lastChargedTimeStamp_ = 0;
-    battMonitorPeriod_ = BATTERY_STATE_CHECK_SHORT_PERIOD;
+    battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
     DBG_PWR("notChargingDebounceCount_: %d", notChargingDebounceCount_);
     if (notChargingDebounceCount_ >= BATTERY_NOT_CHARGING_DEBOUNCE_COUNT) {
       confirmBatteryState(g_batteryState, BATTERY_STATE_NOT_CHARGING);
@@ -600,7 +581,7 @@ void PowerManager::batteryStateForwardingTo(battery_state_t targetState, bool co
     chargedFaultCount_ = 0;
     notChargingDebounceCount_ = 0;
     lastChargedTimeStamp_ = 0;
-    battMonitorPeriod_ = BATTERY_STATE_CHECK_SHORT_PERIOD;
+    battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
     DBG_PWR("chargingDebounceCount_: %d", chargingDebounceCount_);
     if (chargingDebounceCount_ >= BATTERY_CHARGING_DEBOUNCE_COUNT) {
       confirmBatteryState(g_batteryState, BATTERY_STATE_CHARGING);
@@ -613,7 +594,7 @@ void PowerManager::batteryStateForwardingTo(battery_state_t targetState, bool co
     }
     notChargingDebounceCount_ = 0;
     chargingDebounceCount_ = 0;
-    battMonitorPeriod_ = BATTERY_STATE_CHECK_SHORT_PERIOD;
+    battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
     DBG_PWR("chargedFaultCount_: %d", chargedFaultCount_);
     if (chargedFaultCount_ == 1) {
       DBG_PWR("Set fault recharge threshold: 300mV");
@@ -634,12 +615,12 @@ void PowerManager::clearIntermadiateBatteryState() {
   chargedFaultCount_ = 0;
   notChargingDebounceCount_ = 0;
   lastChargedTimeStamp_ = 0;
-  battMonitorPeriod_ = BATTERY_STATE_CHECK_LONG_PERIOD;
+  battMonitorPeriod_ = BATTERY_STATE_NORMAL_CHECK_PERIOD;
 }
 
 void PowerManager::deduceBatteryStateLoop() {
   // We don't need to deduce the battery state when the battery is used as the main power supplier
-  if (g_batteryState == BATTERY_STATE_DISCHARGING) {
+  if (g_batteryState == BATTERY_STATE_DISCHARGING && battMonitorPeriod_ == BATTERY_STATE_NORMAL_CHECK_PERIOD) {
     return;
   }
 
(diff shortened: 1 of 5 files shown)


system/src/system_power_manager.cpp
```diff
@@ -552,52 +510,87 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+    battery_state_t batteryState;
+    uint8_t status;
+    PMIC power(true);
+    FuelGauge fuel(true);
+    if (chargedTimeStamp_ != 0 && millis() - chargedTimeStamp_ >= BATTERY_CHARED_POSTPONE_PERIOD) {
+      chargedTimeStamp_ = 0;
+      status = power.getSystemStatus();
+      // Battery is charged, then the battery voltage should above VSYS_MIN (3.5v)
+      // The VSYS_STAT bit should be cleared, otherwise, the battery is disconnected
+      if (status & 0x01) {
```
XuGuohui, line 554: We should check that VCELL is above for example 4.0v as well, because when charging is enabled and the battery is disabled the VBAT is floating and may goes above VSYS_MIN.

system/src/system_power_manager.cpp
```diff
@@ -552,52 +526,94 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+      if (chrg_stat == 0b01 || chrg_stat == 0b10) {
+        chargingDebounceCount_++;
+        if (chargingDebounceCount_ >= BATTERY_CHARGING_DEBOUNCE_COUNT) {
+          batteryState = BATTERY_STATE_CHARGING;
+          handleStateChange(g_batteryState, batteryState);
+        }
+      } else if (chrg_stat == 0b00) {
+        chargingDebounceCount_ = 0;
+      } else {
+        // if battery is not attached, the charged event will probably trigger INT and followed by disabling charging
```
avtolstoy, line 610: > and followed by disabling charging

We will also get here if the re-attached battery is above recharge threshold. We should clarify the comment that this state change will correctly be handled by `handleUpdate()`.

system/src/system_power_manager.cpp
```diff
@@ -552,52 +526,94 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+        power.enableSafetyTimer();
+        const uint8_t chrg_stat = (power.getSystemStatus() >> 4) & 0b11;
+        if (chrg_stat == 0b01 || chrg_stat == 0b10) {
+          batteryState = BATTERY_STATE_CHARGING;
+        } else {
+          batteryState = BATTERY_STATE_CHARGED;
+        }
+      }
+      handleStateChange(g_batteryState, batteryState);
+    } else if (g_batteryState == BATTERY_STATE_DISCONNECTED && millis() - disconnectedTimeStamp_ >= BATTERY_STATE_CHECK_PERIOD) {
```
avtolstoy, line 593: Given https://github.com/particle-iot/device-os/pull/2272/files#diff-1e4e4a03c336c7a6fe4cbb1114a4ffdd231581cf432dfa9c20587636679b2febR579, we should increase `BATTERY_STATE_CHECK_PERIOD` from 2 seconds to prevent PMIC from triggering the interrupt line too often.

avtolstoy (changes requested):

system/src/system_power_manager.cpp
```diff
@@ -552,52 +526,94 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+    if (millis() - batMonitorTimeStamp_ > BATTERY_STATE_CHECK_PERIOD) {
+      batMonitorTimeStamp_ = millis();
+      PMIC power(true);
+      FuelGauge fuel(true);
+      // When charging is disabled, the VCELL can reliably refect the voltage on VBAT
+      float vCell = fuel.getVCell();
+      LOG_DEBUG(TRACE, "VCell: %dmV", (uint16_t)(vCell * 1000));
+      if (g_batteryState != BATTERY_STATE_DISCONNECTED) {
+        uint8_t status = power.getSystemStatus();
+        if (vCell < BATTERY_CONNECTED_VCELL_THRESHOLD && !(status & 0x08)/* MUST NOT in DPM mode, in case of sudden sink in suplement mode*/) {
```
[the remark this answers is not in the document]
XuGuohui: We are checking DPM here, not DPDM, which indicates that the PMIC is working in the supplement mode and the battery is discharging.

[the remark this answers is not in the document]
avtolstoy: :sweat_smile: You are right. :+1:

system/src/system_power_manager.cpp
XuGuohui: Since we've disabled charging once we got a charged event, so it won't trigger the interrupt too often. What is the period you would suggest so that it neither overload the CPU, nor take too long to update the battery state?

avtolstoy: Can you point me to where we are disabling charging? I don't see a call to `disableCharging()` anywhere.

XuGuohui: In the `handleUpdate` function where we check the charging state.

avtolstoy: That's a persistent setting (introduced in 3.0) that allows the application to enable or disable charging under specific conditions. Normally charging is enabled.

avtolstoy: This is also the code block that handles the case where charging is configured to be enabled.

XuGuohui: I was mentioning this: https://github.com/particle-iot/device-os/blob/563c3a5120ccb92ab17b1a7798ab56c69755cd1a/system/src/system_power_manager.cpp#L222

avtolstoy: Okay, it seems I misread that check. So, right now what we do is once we get charged event from PMIC and charging is enabled in power manager we are going to disable charging? I don't think we should be controlling that manually, as with disproportionate power supply vs full device current draw we may very well have to drop out of that charged state very soon and I think we should rely on the PMIC for that. The only problem is with fluctuations around recharge-threshold, but these are solvable with changing it temporarily to suppress.

This also becomes a problem for sleepy devices where manually disabling charging like this may lead to a complete discharge if we go into sleep after that.

XuGuohui:
@avtolstoy @eberseth I just updated this PR. I'd like you to review again. If you're going to run some tests with more logs, simply enable `DBG_PWR` in the system power manager. I'll clean up those logs before merging this PR.

system/src/system_power_manager.cpp
```diff
@@ -552,52 +557,188 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
-    initDefault(false);
+void PowerManager::deduceBatteryStateChargeDisabled() {
+  DBG_PWR("deduceBatteryStateChargeDisabled()");
+  // XXX: This is a dirty hack. Using the VCELL to deduce the battery state
+  // is based on the experience that when charging is disabled, the VCELL can
+  // correctly be measured.
+  // When charging is disabled, the VCELL can reliably reflect the voltage on VBAT
+  FuelGauge fuel(true);
+  float vCell = fuel.getVCell();
+  LOG_DEBUG(TRACE, "VCell: %dmV", (uint16_t)(vCell * 1000));
```
avtolstoy, line 657: `DBG_PWR` ?

system/src/system_power_manager.cpp
```diff
@@ -552,52 +557,188 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+    }
+    chargedFaultCount_ = 0;
+    notChargingDebounceCount_ = 0;
+    lastChargedTimeStamp_ = 0;
+    battMonitorPeriod_ = BATTERY_STATE_CHANGE_CHECK_PERIOD;
+    DBG_PWR("chargingDebounceCount_: %d", chargingDebounceCount_);
+    if (chargingDebounceCount_ >= BATTERY_CHARGING_DEBOUNCE_COUNT) {
+      confirmBatteryState(g_batteryState, BATTERY_STATE_CHARGING);
+    }
+  } else if (targetState == BATTERY_STATE_CHARGED) {
```
avtolstoy, line 589: Do we count transitions from CHARGED to CHARGED as well?

system/src/system_power_manager.cpp
```diff
@@ -552,52 +557,188 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+    // Restore normal recharge threshold
+    DBG_PWR("Rstore normal threshold: 100mV");
+    PMIC power(true);
+    power.setRechargeThreshold(PMIC_NORMAL_RECHARGE_THRESHOLD); // 100mV
+    return;
+  }
+
+  if (millis() - batMonitorTimeStamp_ >= battMonitorPeriod_) {
+    batMonitorTimeStamp_ = millis();
+    if (config_.flags & HAL_POWER_CHARGE_STATE_DISABLE) {
```
avtolstoy, line 641: It's not guaranteed that the actual PMIC charging state is going to be what we intend it to be. We should watch out for that.

system/src/system_power_manager.cpp
```diff
@@ -552,52 +557,188 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+    DBG_PWR("In DPM mode");
+    // It's in DPM mode, the battery is discharging
+    confirmBatteryState(g_batteryState, BATTERY_STATE_DISCHARGING);
+  } else if (vCell < BATTERY_CONNECTED_VCELL_THRESHOLD) {
+    DBG_PWR("vCell < BATTERY_CONNECTED_VCELL_THRESHOLD");
+    // If battery falls below 2.0v, it's probably detached
+    confirmBatteryState(g_batteryState, BATTERY_STATE_DISCONNECTED);
+  } else {
+    // Since charging is disabled, it must be in the not charging shate
+    confirmBatteryState(g_batteryState, BATTERY_STATE_NOT_CHARGING);
```
avtolstoy, line 698: Do we additionally need to check 'power good' here or is DPM state enough?, same in `deduceBatteryStateChargeEnabled`

system/src/system_power_manager.cpp
```diff
@@ -29,15 +29,31 @@ LOG_SOURCE_CATEGORY("sys.power");
...
 
-constexpr system_tick_t DEFAULT_FAULT_WINDOW = 1000;
-constexpr uint32_t DEFAULT_FAULT_COUNT_THRESHOLD = HAL_PLATFORM_PMIC_BQ24195_FAULT_COUNT_THRESHOLD;
-constexpr system_tick_t DEFAULT_FAULT_SUPPRESSION_PERIOD = 60000;
+// For deducing battery satte
+constexpr system_tick_t BATTERY_STATE_NORMAL_CHECK_PERIOD = 10000;
+constexpr system_tick_t BATTERY_STATE_CHANGE_CHECK_PERIOD = 1000;
+constexpr system_tick_t BATTERY_CHARGED_MUTE_WINDOW = 1000;
+constexpr system_tick_t BATTERY_DISABLE_CHARGING_SUPRESS_PERIOD = 5000; // How long after disable charging the VCELL can reliably reflect the VBAT
+constexpr float BATTERY_CONNECTED_VCELL_THRESHOLD = 2.0; // v
```
avtolstoy, line 52: Nit: `1.0f`

avtolstoy (changes requested):

system/src/system_power_manager.cpp
```diff
@@ -29,15 +29,31 @@ LOG_SOURCE_CATEGORY("sys.power");
 
 #if (HAL_PLATFORM_PMIC_BQ24195 && HAL_PLATFORM_FUELGAUGE_MAX17043)
 
+#if 0
+#define DBG_PWR(_fmt, ...) \
+        do { \
+            LOG_PRINTF(ERROR, _fmt "\r\n", ##__VA_ARGS__); \
```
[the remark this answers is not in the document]
XuGuohui: It's just a copy/paste from other source file, will change to `LOG()`

system/src/system_power_manager.cpp
```diff
@@ -199,48 +230,40 @@ void PowerManager::handleUpdate() {
   if ((curFault & 0x80 /*watchdog fault*/) || !power.isBuckEnabled()) {
     // Restore parameters
     initDefault();
-  } else {
```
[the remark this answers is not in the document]
XuGuohui: Makes sense 👍

system/src/system_power_manager.cpp
XuGuohui: It does, right? whichever the current state is, as long as the target state is charged we count it. That's how we count the charged event to detect disconnected state.

system/src/system_power_manager.cpp
XuGuohui: It checks the DPM state a few lines above, but yes, we need to check the "power good" as well.

system/src/system_power_manager.cpp
```diff
@@ -552,52 +555,205 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+  }
+
+connected:
+  PMIC power(true);
+  const uint8_t status = power.getSystemStatus();
+  const uint8_t pwrGood = (status >> 2) & 0b01;
+  if ((status & 0x08) || !pwrGood) {
+    DBG_PWR("In DPM mode or power is not good");
+    // It's in DPM mode, the battery is discharging
+    confirmBatteryState(g_batteryState, BATTERY_STATE_DISCHARGING);
```
avtolstoy, line 704: I think this is a bit confusing when battery charging is enabled and we have a power source connected. We will be reporting a sticky `discharging` state even though we are still charging the battery while at the same time occasionally taking some charge off of it to compensate for spikes. I would probably stick with just checking `pwrGood` here.

cc @eberseth

avtolstoy (changes requested):

system/src/system_power_manager.cpp
```diff
@@ -552,52 +555,205 @@ void PowerManager::handleStateChange(battery_state_t from, battery_state_t to, b
...
+  const uint8_t status = power.getSystemStatus();
+  const uint8_t chargeState = (status >> 4) & 0b11;
+  if (chargeState == 0b11 && g_batteryState != BATTERY_STATE_CHARGED) {
+    batteryStateTransitioningTo(BATTERY_STATE_CHARGED);
+  } else if ((chargeState == 0b01 || chargeState == 0b10)) {
+    batteryStateTransitioningTo(BATTERY_STATE_CHARGING);
+  } else if (chargeState == 0b00) {
+    const uint8_t pwrGood = (status >> 2) & 0b01;
+    if ((status & 0x08) || !pwrGood) {
+      DBG_PWR("In DPM mode or power is not good");
```
avtolstoy, line 747: I think this is a bit confusing when battery charging is enabled and we have a power source connected. We will be reporting a sticky `discharging` state even though we are still charging the battery while at the same time occasionally taking some charge off of it to compensate for spikes. I would probably stick with just checking `pwrGood` here, especially considering that we will not be getting an event from the PMIC whenever DPM state changes.

cc @eberseth

system/src/system_power_manager.cpp
```diff
@@ -29,15 +29,33 @@ LOG_SOURCE_CATEGORY("sys.power");
...
 using namespace particle::power;
 
 namespace {
 
 constexpr uint8_t BQ24195_VERSION = 0x23;
 
-constexpr system_tick_t DEFAULT_FAULT_WINDOW = 1000;
-constexpr uint32_t DEFAULT_FAULT_COUNT_THRESHOLD = HAL_PLATFORM_PMIC_BQ24195_FAULT_COUNT_THRESHOLD;
-constexpr system_tick_t DEFAULT_FAULT_SUPPRESSION_PERIOD = 60000;
+// For deducing battery satte
```
eberseth, line 49: state

system/src/system_power_manager.cpp
XuGuohui: > especially considering that we will not be getting an event from the PMIC whenever DPM state changes

That makes sense 👍

avtolstoy (approved):

avtolstoy:
@eberseth Can you please update the review status based on your testing with Asset Tracker?

system/src/system_power_manager.cpp
eberseth: I am ok with this change but understand that it would prevent DPM changes

eberseth (approved):

avtolstoy (approved):

PR #2274, opened 2021-02-04
Develop PoC for device-os-test-runner tool to validate minimum RAM SLO is achieved

joegoggins:
Story details: https://app.clubhouse.io/particle/story/71862


user/tests/integration/communication/startup-slos/p1-and-photon-free-ram-slo.spec.js
```diff
@@ -0,0 +1,12 @@
+// Why this different SLO?
+// The official Device OS calls for 60 kB of free RAM, however, P1 and Photon's have never met that quota and won't
+// so we exclude it here.
+suite('startup free ram SLO for photon and p1');
+platform('photon', 'p1');
+test('slo startup stats', async function () {
+    const target = 45000;
```
avtolstoy, line 7: Let's keep a single spec file for a single suite, otherwise we have code duplication and it's easy to miss things like this when implementing changes later or refactoring. It's possible to get the current DUT from inside the test and use separate limits for specific platforms e.g.:
```js
let target = 60000;
const dut = this.particle.devices[0];
if (dut.platform.name == 'photon' || dut.platform.name == 'p1') {
    target = 45000;
}
```

user/tests/integration/communication/startup-slos/slos.cpp
```diff
@@ -0,0 +1,11 @@
+#include "application.h"
```
avtolstoy, line 1: Let's also move these out of `communication` folder into its own e.g. `slos` to make things neater, as `communication` within Device OS repo specifically means anything dealing with cloud commnucations and cloud primitives: pubsub, functions, variables etc.

avtolstoy (changes requested):

joegoggins:
I took care of both of those suggestions @avtolstoy in 97a2c0d. Test output is accurate for `device-os-test run startup-slos` with a photon and boron plugged into USB.

I decided to name the dir "startup-slos" instead of "slos" because there will be other SLOs later. Also, the minimum flash SLO + "connects quickly" SLO can be implemented easily by just adding to what the C++ app publishes and making a couple of assertions against it in the .js file.

avtolstoy (approved):

PR #2277, opened 2021-02-10
Ch72460/more startup slos

joegoggins:
https://app.clubhouse.io/particle/story/72460/build-device-os-test-runner-integration-test-that-validates-the-minimum-flash-space-and-connects-quickly-slo

### DIFF
--- a/user/tests/integration/startup-slos/startup-slos.cpp
+++ b/user/tests/integration/startup-slos/startup-slos.cpp
@@ -5,9 +5,6 @@
 // Since this test framework does not have a setup() + loop() function, it is somewhat different.
 test(slo_startup_stats)
 {
-    // defined by linker for start of heap (end of bss)
-    extern char link_heap_location;
-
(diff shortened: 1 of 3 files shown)


user/tests/integration/startup-slos/startup-slos.cpp
```diff
@@ -1,11 +1,36 @@
...
+//required for linking FLASH_ModuleLength on Gen3 to work: 
+extern "C" int hal_exflash_read(uintptr_t addr, uint8_t* data_buf, size_t data_size) {
+    return SYSTEM_ERROR_NOT_SUPPORTED;
+}
+
 test(slo_startup_stats)
-{
+{    
+    // get millis_to_connected
+    static unsigned long base_time = millis();
```
avtolstoy, line 16: There is no need for all these to be static.

avtolstoy (approved):
LGTM, although it might be a good idea to separate various SLOs into separate test cases.

PR #2280, opened 2021-02-15
Allow clearing session data and running custom setup code in integration tests

sergeuz:
### Problem

1. The integration test framework defines the `setup()` and `loop()` functions making it impossible for test applications to run custom code in those functions.
2. The system Device OS API doesn't provide a way to force a new session to be established with the cloud which makes it difficult to write tests around session resumption.

### Solution

1. Introduce a couple macros that can be defined in an integration test's `test.mk`: `NO_TEST_APP_INIT` and `NO_TEST_APP_SETUP_AND_LOOP`. The first macro disables automatic execution of the test framework's global initialization code, and the second one prevents the framework from defining the `setup()` and `loop()` functions. In both cases the test application is responsible for initializing the test framework:
```cpp
#include "test.h"

STARTUP({
    testAppInit();
});

void setup() {
    testAppSetup();
}

void loop() {
    testAppLoop();
}

test(my_test_case) {
    ...
}
```

2. Add an option to `CloudDisconnectOptions` that allows clearing the cached session data when disconnecting from the cloud and thus forcing a new session to be established on the next connection attempt:
```cpp
SYSTEM_MODE(SEMI_AUTOMATIC)

void setup() {
    Particle.disconnect(CloudDisconnectOptions().clearSession(true));
    Particle.connect();
}

void loop() {
}
```

This PR also moves the system threading definitions to the `particle` namespace. It's an artifact from the previous implementation of this PR which I decided too keep since it reduces the usage of the global namespace in our code a tiny bit (I originally wanted to extend the `system_internal()` function instead of introducing a new disconnection option).

### References

- [ch72812]

### DIFF
--- a/system/src/system_cloud_connection.cpp
+++ b/system/src/system_cloud_connection.cpp
@@ -91,7 +91,7 @@ int SessionConnection::load(const ServerAddress& addr)
             return -1;
         }
     } else {
-        LOG(ERROR, "Failed to load session data from persistent storage");
+        LOG(WARN, "Failed to load session data from persistent storage");
         discard();
         return -1;
     }
--- a/system/inc/system_cloud.h
+++ b/system/inc/system_cloud.h
@@ -308,8 +308,9 @@ bool spark_cloud_flag_auto_connect(void);
  * Option flags for `spark_cloud_disconnect_options`.
  */
 typedef enum spark_cloud_disconnect_option_flag {
-    SPARK_CLOUD_DISCONNECT_OPTION_GRACEFUL = 0x01, ///< The graceful disconnection option is set.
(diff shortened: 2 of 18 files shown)


system/src/system_cloud_connection.cpp
```diff
@@ -86,7 +86,7 @@ int SessionConnection::load(const ServerAddress& addr)
             return -1;
         }
     } else {
-        LOG(ERROR, "Failed to load session data from persistent storage");
+        LOG(WARN, "Failed to load session data from persistent storage");
```
avtolstoy, line 89: :+1:

avtolstoy (approved):

technobly (approved):

PR #2236, opened 2020-11-21
Signal values to percentage mapping

keeramis:
### Problem

Current signal values to percentage conversion is incorrect depending on the values.

### Solution

Changing the signal percentages reported to the cloud such that they render an accurate representation of the cellular signal.

### References

- [ch66635](https://app.clubhouse.io/particle/story/66635/change-the-how-device-os-signal-quality-percentages-are-calculated-in-device-os)

---

### Completeness

- [x] User is totes amazing for contributing!
- [ ] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [ ] Problem and Solution clearly stated
- [ ] Run unit/integration/application tests on device
- [ ] Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

PR #2285, opened 2021-02-19
Modify mapping for LTE signal values to percentages

keeramis:
### Problem

Current mapping is not reflective of the data and performance obtained from the field units.

### Solution

Present this mapping correctly close to the performance observed in the field devices.

### References

- [ch66647](https://app.clubhouse.io/particle/story/66477/device-details-page-shows-poor-rf-warnings-when-it-s-actually-fine-and-the-platform-behaves-as-expected)
- [ch66635](https://app.clubhouse.io/particle/story/66635/change-the-how-device-os-signal-quality-percentages-are-calculated-in-device-os)
- [PR2236](https://github.com/particle-iot/device-os/pull/2236)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/shared/cellular_sig_perc_mapping.cpp
+++ b/hal/shared/cellular_sig_perc_mapping.cpp
@@ -35,41 +35,43 @@ enum SignalMappingRegions {
     MAX_REGION_THRESHOLDS
 };
 
-/* RF Characterization internal testing. Please see https://docs.google.com/spreadsheets/d/11qpipAc7JWzr8rAk-SOClfhflT-gXabDPGOMQU4be8I/edit?usp=sharing */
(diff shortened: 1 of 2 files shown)


keeramis:
@avtolstoy I have put the script and queries plus results here - please see https://github.com/particle-iot/cellular-connectivity-tests/tree/main/rf-char/field_data

A couple of points:
1. I have roughly put the 60% limits such that most of the "red" region is in the quadrant covered below 60% strength and quality. Also, Yet to generate data about RTT but unsure if that is needed for this PR.
2. Made these changes only for LTE. 3G/2G are still android values.

avtolstoy (approved):

PR #2288, opened 2021-02-24
Enable LTO for Photon system firmware

sergeuz:
### Problem

We're out of flash space in system-part2 on Photon.

### Solution

Enable LTO when building Photon system part modules. We have previously [enabled](https://github.com/particle-iot/device-os/pull/2235) it for Electron, but doing the same for Photon causes various compile- and runtime errors with GCC 9.2.1. Building with GCC 10.2.1 seems to resolve those issues.

Enabling LTO frees up ~3 and ~11K of flash memory in Photon's system-part1 and system-part2 respectively.

### References

- [ch73834]

### DIFF
--- a/bootloader/src/nRF52840/nrf_it.c
+++ b/bootloader/src/nRF52840/nrf_it.c
@@ -83,7 +83,7 @@ void HardFault_Handler(void)
         " mrseq r0, msp                                             \n"
         " mrsne r0, psp                                             \n"
         " ldr r1, [r0, #24]                                         \n"
-        " ldr r2, handler2_address_const                            \n"
+        " ldr r2, =handler2_address_const                           \n"
         " bx r2                                                     \n"
         " handler2_address_const: .word prvGetRegistersFromStack    \n"
     );
--- a/build/arm-tools.mk
+++ b/build/arm-tools.mk
@@ -64,27 +64,21 @@ endif
 CFLAGS += --specs=nano.specs
 
 # Check if the compiler version is the minimum required
-quote="
-lt=\<
-dollar=$$
-arm_gcc_version_str:=$(strip $(shell $(CC) -dumpversion))
-expected_version:=9.2.1
-#$(info result $(shell test $(quote)$(arm_gcc_version_str)$(quote) $(lt) $(quote)$(expected_version)$(quote);echo $$?))
-ifeq ($(shell test $(quote)$(arm_gcc_version_str)$(quote) $(lt) $(quote)$(expected_version)$(quote); echo $$?),0)
-     $(error "ARM gcc version $(expected_version) or later required, but found $(arm_gcc_version_str)")
+version_to_number=$(shell v=$1; v=($${v//./ }); echo $$((v[0] * 10000 + v[1] * 100 + v[2])))
+get_major_version=$(shell v=$1; v=($${v//./ }); echo $${v[0]})
+arm_gcc_version_str:=$(shell $(CC) -dumpversion)
+arm_gcc_version:=$(call version_to_number,$(arm_gcc_version_str))
+expected_version_str:=9.2.1
+ifeq ($(shell test $(arm_gcc_version) -lt $(call version_to_number,$(expected_version_str)); echo $$?),0)
+     $(error "ARM gcc version $(expected_version_str) or later required, but found $(arm_gcc_version_str)")
 endif
 
-arm_gcc_version_major:=$(word 1,$(subst ., ,$(arm_gcc_version_str)))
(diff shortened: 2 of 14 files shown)


bootloader/src/nRF52840/nrf_it.c
```diff
@@ -83,7 +83,7 @@ void HardFault_Handler(void)
         " mrseq r0, msp                                             \n"
         " mrsne r0, psp                                             \n"
         " ldr r1, [r0, #24]                                         \n"
-        " ldr r2, handler2_address_const                            \n"
+        " ldr r2, =handler2_address_const                           \n"
```
sergeuz, line 86: Changing this to a pseudo-instruction fixes warnings that are now generated by the assembler bundled with ARM GCC 10:
```
/tmp/ccfCrEXn.s: Assembler messages:
/tmp/ccfCrEXn.s:65974: Warning: section does not have enough alignment to ensure safe PC-relative loads
```

build/arm-tools.mk
```diff
@@ -64,27 +64,21 @@ endif
...
-ifeq ($(shell test $(quote)$(arm_gcc_version_str)$(quote) $(lt) $(quote)$(expected_version)$(quote); echo $$?),0)
-     $(error "ARM gcc version $(expected_version) or later required, but found $(arm_gcc_version_str)")
+version_to_number=$(shell v=$1; v=($${v//./ }); echo $$((v[0] * 10000 + v[1] * 100 + v[2])))
+get_major_version=$(shell v=$1; v=($${v//./ }); echo $${v[0]})
+arm_gcc_version_str:=$(shell $(CC) -dumpversion)
+arm_gcc_version:=$(call version_to_number,$(arm_gcc_version_str))
+expected_version_str:=9.2.1
+ifeq ($(shell test $(arm_gcc_version) -lt $(call version_to_number,$(expected_version_str)); echo $$?),0)
+     $(error "ARM gcc version $(expected_version_str) or later required, but found $(arm_gcc_version_str)")
 endif
```
technobly, line 74: Should this be pinned to the exact version we require?  For example, v1.5.x Device OS requires GCC 5.3.1, but if you have 9.2.1 installed it doesn't complain.  I guess we will also now have the same problem with 2.x Device OS if GCC 10.x is installed.  We can't go back and fix the checks for older firmware, i.e. Device OS 1.5.x, so it would have to be done when we update the compiler version.

technobly (approved):

build/arm-tools.mk
sergeuz: Why would we want to require a specific compiler version and not allow compiling the code with a more recent one?

avtolstoy:
Added GCC 10.2.1 to `.workbench/manifest.json`. Works ok on Linux/Win/Mac.

cc @XuGuohui once this is merged you should be unblocked

avtolstoy (approved):

busticated (approved):
>updated workbench gcc bundles to gcc 10.2.1

smoke-tested on macOS and looks good 👍

build/arm-tools.mk
busticated: b/c it’s untested and you cannot guarantee it will work.

if you want to allow it for purposes of testing, print out a very clear warning and / or force the user to explicitly override

sergeuz: > b/c it’s untested and you cannot guarantee it will work

This is true for every software project out there, but I personally have never seen a single OSS project that would actively prevent its users from using a newer version of the compiler or runtime with it. We should definitely have the minimum required or recommended version of the toolchain somewhere in the README though.

busticated: your mental model is wrong. the build system is no longer just an internal concern - it drives the local compiler that we encourage end-users to use via Particle Workbench and other tools.

> ...actively prevent its users from using a newer version of the compiler

see the 2nd part of my comment - show a warning and / or force the end-user to explicitly override it such that evidence of them having done so is very obvious in the logs. 

as you are evolving the build system, one of your primary aims should be to anticipate and minimize potential customer support issues.

sergeuz: > your mental model is wrong

All I'm saying is that what you and Brett suggest is unusual for an OSS project such as Device OS. It's a valid point worth considering just like your concerns about Device OS' build system and Workbench.

> as you are evolving the build system, one of your primary aims should be to anticipate and minimize potential customer support issues

Also note that the change that is being discussed in this thread doesn't alter the current behavior in regards to how the build scripts handle the range of supported compiler versions. I'm not "evolving" anything here, I'm fixing a very specific bug with how version strings are compared.

sergeuz:
> Added GCC 10.2.1 to .workbench/manifest.json. Works ok on Linux/Win/Mac.

@avtolstoy Thank you!

PR #2290, opened 2021-02-26
Fixes abnormal battery charged behavior and adds a test for battery state deducing.

XuGuohui:
Add a manual test application for validating battery state. This is a supplement of https://github.com/particle-iot/device-os/pull/2272.

### Completeness
- [ ] Added documentation

### DIFF
--- a/system/src/system_power_manager.cpp
+++ b/system/src/system_power_manager.cpp
@@ -65,6 +65,11 @@ constexpr system_tick_t PMIC_WATCHDOG_INTERVAL = 40000; // 40s
 constexpr uint16_t PMIC_NORMAL_RECHARGE_THRESHOLD = 100; // millivolts
 constexpr uint16_t PMIC_FAULT_RECHARGE_THRESHOLD = 300; // millivolts
 
+constexpr uint16_t PMIC_NORMAL_TERM_CHARGE_CURRENT = 256; // mA
+constexpr uint16_t PMIC_FAULT_TERM_CHARGE_CURRENT = 128; // mA
+constexpr system_tick_t BATTERY_REPEATED_CHARGED_WINDOW = 5000; // ms
+constexpr uint8_t BATTERY_REPEATED_CHARGED_COUNT = 2;
+
 constexpr hal_power_config defaultPowerConfig = {
   .flags = 0,
   .version = 0,
@@ -188,7 +193,7 @@ void PowerManager::handleCharging(bool batteryDisconnected) {
 
   if ((config_.flags & HAL_POWER_CHARGE_STATE_DISABLE) || batteryDisconnected) {
     if (power.isChargingEnabled()) {
-      clearIntermadiateBatteryState();
+      clearIntermadiateBatteryState(STATE_ALL);
       DBG_PWR("Disabled charging.");
       power.disableCharging();
       power.disableSafetyTimer();
@@ -200,7 +205,7 @@ void PowerManager::handleCharging(bool batteryDisconnected) {
   }
 
   if (!power.isChargingEnabled()) {
-    clearIntermadiateBatteryState();
+    clearIntermadiateBatteryState(STATE_ALL);
     DBG_PWR("Enable charging.");
     power.enableCharging();
     power.enableSafetyTimer();
@@ -449,9 +454,10 @@ void PowerManager::initDefault(bool dpdm) {
   power.setChargeVoltage(config_.termination_voltage);
 
   // Set recharge threshold to default value - 100mV
-  DBG_PWR("initDefault, Threshold: 100mV");
+  DBG_PWR("initDefault, recharge threshold: %dmV, term current: %dmA", PMIC_NORMAL_RECHARGE_THRESHOLD, PMIC_NORMAL_TERM_CHARGE_CURRENT);
   power.setRechargeThreshold(PMIC_NORMAL_RECHARGE_THRESHOLD);
   power.setChargeCurrent(config_.charge_current);
+  power.setTermChargeCurrent(PMIC_NORMAL_TERM_CHARGE_CURRENT);
 
   // Enable or disable charging
   if (g_batteryState != BATTERY_STATE_UNKNOWN) {
@@ -500,13 +506,33 @@ void PowerManager::initDefault(bool dpdm) {
 void PowerManager::confirmBatteryState(battery_state_t from, battery_state_t to) {
   // Whenever we confirm the battery state, we re-start the monitor cycle
   batMonitorTimeStamp_ = millis();
-  clearIntermadiateBatteryState();
+  clearIntermadiateBatteryState(STATE_CHARGED | STATE_CHARGING | STATE_NOT_CHARGING);
+  battMonitorPeriod_ = BATTERY_STATE_NORMAL_CHECK_PERIOD;
   
   // Charging may be temporarily enabled when battery is disconnected.
   // If we are still in the disconnected state, we need to disable charging again.
   handleCharging((to == BATTERY_STATE_DISCONNECTED) ? true : false);
 
   if (from == to) {
+    if (to == BATTERY_STATE_CHARGED) {
+      // Deduce the repeated charged state
+      if (millis() - repeatedChargedTimeStamp_ < BATTERY_REPEATED_CHARGED_WINDOW) {
+        repeatedChargedTimeStamp_ = millis();
+        repeatedChargedCount_++;
+        DBG_PWR("Repeated charged: %d", repeatedChargedCount_);
+        if (repeatedChargedCount_ >= BATTERY_REPEATED_CHARGED_COUNT) {
+          repeatedChargedCount_ = 0;
+          DBG_PWR("Set term charge current: %dmA", PMIC_FAULT_TERM_CHARGE_CURRENT);
+          // Now we probably attached a problematic battery, the term charge current
+          // will persist until whenever the initDefault() is called.
+          PMIC power;
+          power.setTermChargeCurrent(PMIC_FAULT_TERM_CHARGE_CURRENT);
+        }
+      } else {
+        repeatedChargedTimeStamp_ = millis();
+        repeatedChargedCount_ = 0;
+      }
+    }
     // No state change occured
     return;
   }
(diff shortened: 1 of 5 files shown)


avtolstoy (changes requested):
This is a good candidate to have as a 'test application', however using the test framework here doesn't bring any benefits and will not allow us to easily migrate this test suite to be run in an automated manner.

Is there any way we can figure out several bite-sized test cases that we can run in a semi-automated manner similar to `wiring/power_management` with appropriate assertions in place, so that we don't have to eyeball the behavior of the device?

XuGuohui:
I did take some time thinking how we can automate the test. My thinking was that the test application prompts tester to set up the hardware, for example, "Please plug the battery and USB", then the application would automatically enable/disable charging and make some assertion according to the current hardware setup. After the test case runs for a period, then prompt tester to set up the hardware in another manner and continue the next test case, and so on. The assertions would be that the power source and battery state should be consistent under certain hardware setup. How does the plan sound?

avtolstoy:
@XuGuohui Yes, that sounds much better:
1. We have bite-sized test cases
2. The parts that cannot be automated as of right now require a human operator to perform a specific action
3. We have proper validation (assertions) that with specific conditions set by human operator a particular test case passes
4. We'll be able to more easily refactor the tests to be performed in an automated manner by just replacing the parts that as of right now require human intervention

eberseth:
I had a similar feeling of wanting it to setup and verify something instead of display results.

eberseth (approved):

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,379 @@
...
+        while (SERIAL.available() <= 0);
+        while (SERIAL.available() > 0) {
+            (void)SERIAL.read();
+        }
+
+        SERIAL.println("> Wait the power management settled...");
+        assertTrue(waitFor([]{
+            return System.batteryState() == BATTERY_STATE_DISCHARGING;
+        }, 10000));
+        assertTrue(System.powerSource() == POWER_SOURCE_BATTERY);
```
avtolstoy, line 82: All comparisons (except of course ones that do check a boolean) should use `assertEqual` otherwise it's impossible to tell what the actual value that caused the assertion to trigger was and we'll just see the result of the comparison - 0 (false).

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,379 @@
...
+    // When the battery is charged and the USB is connected, the battery should
+    // reside in the charged state. While for problematic battery, the voltage
+    // drops too much once stop charging due to charged. If the voltage drops more
+    // than 100mV, the PMIC will start charging again. From the human view, the charging
+    // status LED will blink.
+    SERIAL.println("> Test is running...");
+    time32_t start = Time.now();
+    while ((Time.now() - start) < TEST_DURATION) {
+        if (System.batteryState() == BATTERY_STATE_CHARGING) {
+            SERIAL.println("> WARN: (problematic) battery is charging!");
```
avtolstoy, line 261: I had a Boron stuck in this state with approximately 70% of battery charge, which would indicate that we dropped out of https://github.com/particle-iot/device-os/pull/2290/files#diff-92b482059ee678b515c6002f68f3c2ace005b1ecb29490a6ef2e87b72de3a0c4R228 prematurely and the test above anyway succeeded. We should potentially add some debouncing and/or checks for actual SoC. It may also indicate a problem with power manager.

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+constexpr uint32_t DISCHARGE_STATE1 = 0xc00f;
+constexpr uint32_t DISCHARGE_STATE2 = 0xd00f;
+constexpr uint32_t CHARGING_STATE_DISABLED_CHARGING = 0xbeef;
+constexpr uint32_t CHARGING_STATE_ENABLED_CHARGING = 0xdeed;
+constexpr uint32_t CHARGED_STATE = 0xc0fe;
+constexpr uint32_t DISCONNECTED_STATE = 0xd00d;
+constexpr uint32_t DISCONNECTED_STATE_SLEEP = 0xbead;
+constexpr uint32_t CONNECTED_STATE_SLEEP = 0xb007;
+uint32_t g_state;
```
avtolstoy, line 40: Nit: we normally do this when we need to keep state between the resets as a retained variable. Here it's used only in a single place as a conditional, so perhaps we can just take special care with that specific test case.

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,379 @@
...
+    SystemSleepResult r = System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(SLEEP_DURATION));
+    assertEqual(r.error(), SYSTEM_ERROR_NONE);
+    assertTrue(r.wakeupReason() == SystemSleepWakeupReason::BY_RTC);
+
+    SERIAL.println("> Device wakes up. Please reconnect the Serial to continue the test.");
+    while (!Serial.isConnected());
+    delay(1s);
+
+    SERIAL.println("> Wait the power management settled...");
+    assertTrue(waitFor([&]{
```
avtolstoy, line 375: This one also failed on a Boron and again it's difficult to see what the problem was:
```
Assertion failed: (System.waitCondition([]{ return ([&]{ return System.batteryState() == currState; })(); }, (10000))=0) == (true=1), file tests/wiring/power_management_long_running/power.cpp, line 375.
```

The device was blinking charge LED in a pattern of `.-.-.-.-.-.-.-.----------------------------------.-.-.-.-.-.-.-.-.-----------------------------` for a couple of minutes after which the charge LED settled on full-on (at this point the test has already failed).

avtolstoy (changes requested):

user/tests/wiring/power_management_long_running/power.cpp
avtolstoy: This happened close to termination voltage, so I assume most likely our hack with re-charge threshold may not be working entirely correctly or we cannot do anything about this potentially and simply need to account for this behavior in the test.

user/tests/wiring/power_management_long_running/power.cpp
XuGuohui: It is a global variable being used in every test case to track the testing state, if certain test case fails, then the following test cases will be skipped.

user/tests/wiring/power_management_long_running/power.cpp
XuGuohui: I've figured out the potential fake battery state in power manager. I'll add an assertion in the previous test case to check if the SoC complies with the charged state.

user/tests/wiring/power_management_long_running/power.cpp
XuGuohui: This symptom is probably caused by the recharged threshold and happened on part of batteries. Unless we keep the 100mV recharge threshold all the time, we may see this occasionally.

avtolstoy (approved):

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,384 @@
...
+    while (SERIAL.available() > 0) {
+        (void)SERIAL.read();
+    }
+
+    SERIAL.println("> Wait the power management settled...");
+    assertTrue(waitFor([]{
+        return (System.batteryState()!= BATTERY_STATE_UNKNOWN) && (System.powerSource() == POWER_SOURCE_USB_HOST);
+    }, 10000));
+    PMIC power;
+    assertTrue(power.isChargingEnabled());
```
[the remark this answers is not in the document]
XuGuohui: This fails only if the `SystemPowerFeature::DISABLE_CHARGING` is `true` or the battery is disconnected. Otherwise, the PMIC might have tweaked its register for some reason.

system/src/system_power_manager.cpp
```diff
@@ -499,13 +504,33 @@ void PowerManager::initDefault(bool dpdm) {
...
+      // Deduce the repeated charged state
+      if (millis() - repeatedChargedTimeStamp_ < BATTERY_REPEATED_CHARGED_WINDOW) {
+        repeatedChargedTimeStamp_ = millis();
+        repeatedChargedCount_++;
+        DBG_PWR("Repeated charged: %d", repeatedChargedCount_);
+        if (repeatedChargedCount_ >= BATTERY_REPEATED_CHARGED_COUNT) {
+          repeatedChargedCount_ = 0;
+          DBG_PWR("Set term charge current: %dmA", PMIC_FAULT_TERM_CHARGE_CURRENT);
+          // Now we probably attached a problematic battery, the term charge current
+          // will persist until whenever the initDefault() is called.
```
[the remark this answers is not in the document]
XuGuohui: Reset it back to normal term current on charged.

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+} // anonymous
+
+test(power_00_setup) {
+    SERIAL.begin(115200);
+    SERIAL.println("\r\npower_00_setup");
+
+    constexpr float DISCHARGE_VCELL = 4.1f;
+
+    auto cfg = System.getPowerConfiguration();
```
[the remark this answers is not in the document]
XuGuohui: Set the desired power configuration followed by checking if they are applied.

user/tests/wiring/power_management_long_running/power.cpp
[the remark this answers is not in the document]
XuGuohui: It's probably that on startup we get continuous charged event even if when we raise the recharge threshold to be 300mV

user/tests/wiring/power_management_long_running/power.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+        SERIAL.println("> Device is publishing message in every second to discharge the battery...");
+        FuelGauge fuel;
+        float vcell = fuel.getVCell();
+        SERIAL.printlnf("> VBAT: %0.2fv", vcell);
+
+        SERIAL.println("> Test is running...");
+        uint8_t count = 0;
+        time32_t now = Time.now();
+        while (vcell > DISCHARGE_TERMINATION_VCELL || count < 10) {
```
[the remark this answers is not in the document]
XuGuohui: Done.

avtolstoy (approved):

avtolstoy (approved):

Issue #1112, opened 2016-09-07
[0.6.0] I2C - provide flexibility so that larger than 32 bytes can be read


hramadan:
Is it possible to increase the buffer length used in I2C requestFrom() so that 64 bytes of data can be read at once? The buffer size is currently 32.

### Am having an issue with the NXP PN532 NFC reader IC when there is a  more than 32 bytes of data to be transferred. I have not been able to find a work around. 

Completeness:
- [x] Minimum test case added
- [x] Device, system and user firmware versions stated (0.5.3/0.6.0)

hramadan:
A better idea has been suggested by @pkourany - rather than increase the internal buffer to some arbitrary size, let the application specify its own buffers. This is a far better solution.

hramadan:
Refer to the following post:

https://community.particle.io/t/i2c-requestfrom-how-to-read-32-bytes/25615

Would like similar lower level control as offered by SoftI2CMaster lib, not a larger buffer per se. Am changing the title accordingly.

worldburger:
Has there been a simple implementation to reading buffer lengths longer than 32?

hramadan:
@worldburger, to my knowledge this request has not been addressed. Particle documentation, https://docs.particle.io/reference/device-os/firmware/photon/#wire-i2c-, still states a **32 byte buffer**. 

Is your problem in relation to reading NFC cards as this was (still is) my issue.

worldburger:
@hramadan No, I've got a usage case with a particulate monitor serving up more than 32 bits.

I recognize your icon! You're active in the Particle thread:
https://community.particle.io/t/solved-i2c-requestfrom-how-to-read-32-bytes/25615/18}

The particle library I'm trying to get to work is the SPS30. I had looked at your solution with SoftI2CMaster/SoftI2CLibrary.  But to be honest, it's a bit over my head (am rather new-ish to coding).  Have you found a more turnkey solution to adjusting the buffer?

hramadan:
@worldburger, have you looked at the Sensiron source code on GitHub?

https://github.com/Sensirion/embedded-sps
https://github.com/Sensirion/embedded-common

Am pretty sure that you will be able to solve your issue by implementing their source code.

avtolstoy:
Fixed in #2035 #2022

PR #2292, opened 2021-03-03
Added i2c buffer size to Wire interface

perotom:
### Problem

If a custom buffer for i2c is used these functions will give libraries a better ability to determin the i2c buffer size.

### Solution

Added two functions to return the assigned buffer size.

### Example App

```c

constexpr size_t I2C_BUFFER_SIZE = 128;

hal_i2c_config_t acquireWireBuffer() {
    hal_i2c_config_t config = {
        .rx_buffer = new (std::nothrow) uint8_t[I2C_BUFFER_SIZE],
        .rx_buffer_size = I2C_BUFFER_SIZE,
        .tx_buffer = new (std::nothrow) uint8_t[I2C_BUFFER_SIZE],
        .tx_buffer_size = I2C_BUFFER_SIZE
    };
    return config;
}

void setup() {
   Serial.begin();
   Wire.begin();

   waitFor (Serial.isConnected, 10000);
   delay(1000);
   Serial.println(Wire.rxBufferSize());
}

void loop() {

}
```

Tested on B-Som platform.

### References

Adds to #1112 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/hal_dynalib_i2c.h
+++ b/hal/inc/hal_dynalib_i2c.h
@@ -87,6 +87,8 @@ DYNALIB_FN(BASE_IDX + 18, hal_i2c, hal_i2c_lock, int32_t(hal_i2c_interface_t, vo
 DYNALIB_FN(BASE_IDX + 19, hal_i2c, hal_i2c_unlock, int32_t(hal_i2c_interface_t, void*))
 DYNALIB_FN(BASE_IDX + 20, hal_i2c, hal_i2c_request_ex, int32_t(hal_i2c_interface_t, const hal_i2c_transmission_config_t*, void*))
 DYNALIB_FN(BASE_IDX + 21, hal_i2c, hal_i2c_sleep, int(hal_i2c_interface_t i2c, bool sleep, void* reserved))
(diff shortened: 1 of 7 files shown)


jlovinger-particle:
Without getting into the wider PR, lock usage in your `hal_i2c_rx_buffer_size` and `hal_i2c_tx_buffer_size` functions is unnecessary as a read with no modify is already atomic.

perotom:
@jlovinger-particle Should I close the pull request and open another one?

jlovinger-particle:
> @jlovinger-particle Should I close the pull request and open another one?

I'm actually not on the engineering team so am not sure on the process. Just saw it pop up on my feed and had a quick comment.

avtolstoy:
@perotom You can just update your branch that is the source of this PR at any time while the PR is open.

I'll take a look at this later in more detail, but one suggestion is to potentially have a single HAL getter method that would fill in `hal_i2c_config_t` structure instead of discrete methods to just get the sizes of the buffers. Wiring API can be more specific for convenience and pick whatever is needed from `hal_i2c_config_t`.

PR #2293, opened 2021-03-03
Particle.unsubscribe() doesn't preserve system subscriptions

sergeuz:
### Problem

`Particle.unsubscribe()` clears system subscriptions to cloud events along with the application subscriptions and disrupts functioning of IOTA updates.

### Solution

The list of registered event handlers is managed by the protocol layer that doesn't distinguish between system and application events. As a workaround, the system layer now re-adds system subscriptions after clearing them via the protocol's `spark_protocol_remove_event_handlers()`.

### Steps to Test

Verify that a product application based on the example code below can be updated OTA. Alternatively, add some logging to the `systemEventHandler()` function and verify that the system correctly receives `particle/*` events during handshake.

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC)
SYSTEM_THREAD(ENABLED)

// PRODUCT_ID(...)
// PRODUCT_VERSION(...)

namespace {

bool updatePending = false;

} // namespace

void setup() {
    System.disableUpdates();
    Particle.unsubscribe();
    Particle.connect();
}

void loop() {
    if (!updatePending && System.updatesPending()) {
        updatePending = true;
        System.enableUpdates();
    }
}
```

### References

- [ch74478]

### DIFF
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -183,8 +183,9 @@ int sendApplicationDescription() {
 }
 
 void registerSystemSubscriptions() {
-    if (!Particle.subscribe("particle", systemEventHandler, MY_DEVICES) ||
-            !Particle.subscribe("spark", systemEventHandler, MY_DEVICES)) {
+    bool ok = Particle.subscribe("particle", systemEventHandler, MY_DEVICES);
+    ok = Particle.subscribe("spark", systemEventHandler, MY_DEVICES) && ok;
+    if (!ok) {
         LOG(ERROR, "Particle.subscribe() failed");
     }
 }
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -61,9 +61,96 @@
 #include <stdio.h>
 #include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


system/src/system_cloud_internal.cpp
```diff
@@ -95,6 +182,13 @@ int sendApplicationDescription() {
     return 0;
 }
 
+void registerSystemSubscriptions() {
+    if (!Particle.subscribe("particle", systemEventHandler, MY_DEVICES) ||
+            !Particle.subscribe("spark", systemEventHandler, MY_DEVICES)) {
+        LOG(ERROR, "Particle.subscribe() failed");
+    }
+}
```
technobly, line 190: Since there is no error handling other than logging a failure, would it make sense to break these out into separate calls like they were before so `spark` is always attempted to be registered, even if `particle` fails?  We could probably also change the LOG to LOG_DEBUG to keep the file size down.

technobly (approved):

avtolstoy (approved):
Looks good. Shame that we cannot easily trigger `spark/` or `particle/` event to add some kind of a test.

PR #2295, opened 2021-03-05
Inhibit Cellular URCs before going into sleep

XuGuohui:
### Problem
When we specify `.network(Cellular)` as a wakeup source when calling the system sleep API, we leave the modem running. When there is URC coming from the modem, like when Cellular moving from tower to another, it wakes up the device, which is unwanted. From the user's view, the device may be woken up frequently.

This PR also fixes issues:
1. Tracker specifically, which is when Cellular is specified as wakeup source, on waking up the on-board ESP32 module is turned on unexpectedly.

### Solution
Inhibit URCs before going into sleep.
### Steps to Test
To verify the patch, you need to make the comparison:
1. Checkout and build the branch within this PR and observe the log from `Serial1`. 
2. Comment out `cellular_urcs(false, nullptr);` in `system_sleep.cpp` and build again. Observe the log from `Serial1`.

You'll see some URCs relevant logs on waking up if you  comment out `cellular_urcs(false, nullptr);`, but not if you build the branch as-is.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(AUTOMATIC);

void setup() {
    LOG(INFO, "Application started.");

    LOG(INFO, "Cloud connected.");

    delay(10s);
}

void loop() {
    delay(3s);
    LOG(INFO, "\r\n\r\n%d:%d:%d Device entering sleep mode...\r\n", Time.hour(), Time.minute(), Time.second());
    SystemSleepResult result = System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).network(Cellular));
    LOG(INFO, "\r\n\r\n%d:%d:%d Device wakes up, error: %d, reason: %d\r\n", Time.hour(), Time.minute(), Time.second(), result.error(), result.wakeupReason());
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -1505,6 +1505,14 @@ int QuectelNcpClient::getMtu() {
     return 0;
 }
 
+int QuectelNcpClient::enterSleepyState() {
+    return muxer_.suspendChannel(QUECTEL_NCP_AT_CHANNEL);
+}
+
+int QuectelNcpClient::exitSleepyState() {
+    return muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL);
+}
+
 void QuectelNcpClient::connectionState(NcpConnectionState state) {
     if (ncpState_ == NcpState::DISABLED) {
         return;
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -1505,12 +1505,12 @@ int QuectelNcpClient::getMtu() {
     return 0;
 }
 
-int QuectelNcpClient::enterSleepyState() {
-    return muxer_.suspendChannel(QUECTEL_NCP_AT_CHANNEL);
-}
-
-int QuectelNcpClient::exitSleepyState() {
-    return muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL);
+int QuectelNcpClient::urcs(bool enable) {
+    if (enable) {
+        return muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL);
+    } else {
+        return muxer_.suspendChannel(QUECTEL_NCP_AT_CHANNEL);
+    }
 }
 
 void QuectelNcpClient::connectionState(NcpConnectionState state) {
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -1507,10 +1507,11 @@ int QuectelNcpClient::getMtu() {
 
 int QuectelNcpClient::urcs(bool enable) {
     if (enable) {
-        return muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL);
+        CHECK_TRUE(muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
     } else {
-        return muxer_.suspendChannel(QUECTEL_NCP_AT_CHANNEL);
+        CHECK_TRUE(muxer_.suspendChannel(QUECTEL_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
     }
+    return SYSTEM_ERROR_NONE;
 }
 
 void QuectelNcpClient::connectionState(NcpConnectionState state) {
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -1760,6 +1760,14 @@ int SaraNcpClient::getMtu() {
     return 0;
 }
 
+int SaraNcpClient::enterSleepyState() {
+    return muxer_.suspendChannel(UBLOX_NCP_AT_CHANNEL);
+}
+
+int SaraNcpClient::exitSleepyState() {
+    return muxer_.resumeChannel(UBLOX_NCP_AT_CHANNEL);
+}
+
 void SaraNcpClient::connectionState(NcpConnectionState state) {
     if (ncpState_ == NcpState::DISABLED) {
         return;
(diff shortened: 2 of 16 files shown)


keeramis:
@XuGuohui It seems to be helping with this PR. I checked on my Electron LTE, and there's difference with and without URCs disabled; with URCs disabled, it does not wake up before the sleep timer ends, but with URCs enabled, it wakes up every 2s,41s, etc instead of configured 60s. Nice change! I had to explicitly set `cellular_urcs()` to `true/false` though - because the AT commands `AT+UCIND` and `AT+CMER` stick through reboots.

keeramis (approved):

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1402,6 +1402,14 @@ int QuectelNcpClient::enterDataMode() {
     return r;
 }
 
+int QuectelNcpClient::urcs(bool enable) {
+    if (enable) {
+        return muxer_.resumeChannel(QUECTEL_NCP_AT_CHANNEL);
```
avtolstoy, line 1407: Convert to `system_error_t`?

system/src/system_sleep.cpp
```diff
@@ -47,19 +47,29 @@ using namespace particle;
...
 static bool system_sleep_network_suspend(network_interface_index index) {
     bool resume = false;
-    // Disconnect from network
-    if (network_connecting(index, 0, NULL) || network_ready(index, 0, NULL)) {
-        if (network_connecting(index, 0, NULL)) {
-            network_connect_cancel(index, 1, 0, 0);
+
+// FIXME: Dirty hack. For platforms support both Wi-Fi and Cellular interfaces,
+// we simply bypass the Wi-Fi interface 
+#if HAL_PLATFORM_WIFI && HAL_PLATFORM_CELLULAR
```
avtolstoy, line 53: Can you comment on why we need this? `network_connecting(index, 0, NULL) || network_ready(index, 0, NULL)` will be `false` on e.g. Tracker.

avtolstoy (approved):

system/src/system_sleep.cpp
XuGuohui: It actually return `true` if the Cellular is connected to the tower, i.e. IP is obtained in system network manager and is specified as a wakeup source, in which case `network_ready()` returns `true`.

avtolstoy: I think that's not it. The problem is in how `network_connecting` is implemented on Gen 3 platforms. We will return `true` on a Tracker for WiFi interface if Cellular is connected or connecting or any other interface e.g. Ethernet.

avtolstoy: I think we can add a simple method to network manager to check interface flags:

`IFF_UP && !IFF_LOWER_UP` - would indicate `connecting` state.

XuGuohui: Yes, you're right. Also I just find an inconsistent state before/after sleep, which is when the application just turns on the modem to send AT commands manually and then make device enter sleep mode, after waking up the modem won't be turned on, since the `resume` flag isn't set before entering sleep mode. Considering fix it in the PR as well?

avtolstoy: Yes, let's add the fix in this PR.

XuGuohui: Fixed.

avtolstoy:
This needs to be rebased and reviewed in the context of #2308 and #2309

avtolstoy (changes requested):
We should add a test for this.

On Gen 3 we can simply run `Cellular.command()` and make sure that it fails when `cellular_urcs(false)` is called and succeeds when `cellular_urcs(true)` is called.

On Gen 2 we can easily validate that this does work as intended by connecting to the cellular network and performing e.g. a disconnection with `Cellular.command()`. If URCs are disabled, modem driver should fail to notice that we no longer have an active connection and should still continue reporting `Cellular.ready()`.

All of this needs to be done while connected to the cellular network.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1760,6 +1760,15 @@ int SaraNcpClient::getMtu() {
     return 0;
 }
 
+int SaraNcpClient::urcs(bool enable) {
+    if (enable) {
+        CHECK_TRUE(muxer_.resumeChannel(UBLOX_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
+    } else {
+        CHECK_TRUE(muxer_.suspendChannel(UBLOX_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
```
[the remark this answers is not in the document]
XuGuohui: Can we always update `localModemState` and set `update` to be `true` in `resumeChannel()`?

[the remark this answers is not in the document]
avtolstoy: Might be a good idea to default `localModemState` to empty state, or yes, always default to `update = true`. :+1:

user/tests/wiring/no_fixture/cellular.cpp
```diff
@@ -196,6 +196,74 @@ test(CELLULAR_06_on_off_validity_check) {
...
+    }
+#elif HAL_PLATFORM_GEN == 3
+    assertEqual(Cellular.command("AT\r\n"), (int)RESP_OK);
+
+    assertEqual(cellular_urcs(false, nullptr), (int)SYSTEM_ERROR_NONE);
+    assertNotEqual(Cellular.command("AT\r\n"), (int)RESP_OK);
+
+    assertEqual(cellular_urcs(true, nullptr), (int)SYSTEM_ERROR_NONE);
+    assertEqual(Cellular.command("AT\r\n"), (int)RESP_OK);
+#endif
```
avtolstoy, line 264: Nit:
```c++
#else
#error Unsupported platform
#endif
```

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -1295,6 +1295,59 @@ bool MDMParser::powerState(void) {
     return state;
 }
 
+bool MDMParser::urcs(bool enable) {
+    if (enable) {
+        sendFormated("AT+UCIND=4095\r\n");
```
avtolstoy, line 1300: Nit: might be a good idea to wrap these in `CHECK_TIMEOUT` on wake-up so that if the modem for whatever reason is unresponsive we'll go into an error state much quicker.

avtolstoy (approved):

avtolstoy (approved):

PR #2297, opened 2021-03-10
Querying the value of an empty string variable causes an error

sergeuz:
### Problem

Querying the value of an empty string variable via the HTTP API causes an error.

### Solution

Do not attempt to allocate an intermediate buffer for an empty string variable. The implementation of `malloc()` that we use returns `NULL` if the size of the buffer to be allocated is 0. Currently, the system layer misinterprets that as a memory allocation error.

### Steps to Test

Run the `communication/variables` integration test.

### References

- [ch71196]


system/src/system_cloud_internal.cpp
```diff
@@ -778,12 +778,14 @@ void getUserVarImpl(User_Var_Lookup_Table_t* item, SparkDescriptor::GetVariableC
             data = item->userVar;
         }
         size = variableDataSize(data, item->userVarType);
-        copy = malloc(size);
-        if (!copy) {
-            getUserVarResult(ProtocolError::NO_MEMORY, 0, nullptr, 0, callback, context);
-            return;
+        if (size > 0) {
```
avtolstoy, line 781: :+1: Perhaps we can add a small check into our `malloc` routines to modify size 0 allocations to 1 to avoid these problems. I understand that normally this is implementation-specific, but I personally find getting a `NULL` in this case a tad inconvenient having to implement these checks everywhere before calling into malloc.

avtolstoy (approved):

system/src/system_cloud_internal.cpp
sergeuz: Agreed. It also looks like the implementation of `pvPortMalloc()` that we use in our `malloc()` can be easily modified to actually perform "0-byte" allocations.

technobly: Is `size` calculated correctly?  Shouldn't a null string be 1 byte = '\0' ? In this case maybe `size` is more equated to data `length` ?  For a null terminated string, instead of allocating zero bytes, should it allocate 1 byte?  I think this might be more along the lines of what Andrey was saying.

sergeuz: It's by design that `size` doesn't include `\0` here. Also, this code allocates memory for variables of all other types too (`int`, `double`, etc). What Andrey meant is that, typically, `malloc(0)` doesn't return `NULL` and instead returns a valid pointer that can be passed to `free()`, which simplifies error handling and helps to avoid issues like this.

sergeuz: Do you mean the fixed `malloc()`? No, the resulting pointer will literally reference "0-byte" allocation somewhere on the heap, i.e. it will be an address immediately following some `malloc()`'s internal header structure in memory. The only guarantee is that it's not going to be `NULL` and that it's safe to pass it to `free()`.

technobly: My thought was that a String variable that is assigned a null string would be queryable/printable/loggable as a proper null terminated c_str() string, which should be an actual 1 byte memory allocation containing `\0`?  If malloc(0) points to a memory address that is not a nullptr, can/should it still be interpreted as a null terminated string?  I understand the free() part, and free() would also be ok if it was a 1-byte allocation.

technobly: Oh I see now, the data value will default to a nullptr, assuming `memcpy(copy, data, size);` is not run. 
`getUserVarResult(ProtocolError::NO_ERROR, type, copy, size, callback, context);`

I was thinking if the variable contains a null string, why would we not return that actual variable value?

sergeuz: I probably don't understand the question, but, just in case, it's irrelevant for the protocol layer whether the data is null-terminated or not. All it cares about is `size`: if it's 0, the CoAP message carrying the variable's value won't have payload data and the cloud will interpret it as an empty string.

technobly (approved):

keeramis (approved):

PR #2287, opened 2021-02-20
Add support for BLE 5 PHY_CODED (long range) scanning and advertising

markleavitt:
<details>
  <summary><i>submission notes</i></summary>

</details>

### Problem

Obtaining adequate wireless transmission range between BLE devices is a challenge in many use cases. BLE 5 introduced a new transmission mode called PHY_CODED which trades off speed in order to increase noise and error resistance. The current version of Particle DeviceOS does not provide any API to engage the PHY_CODED mode when advertising or scanning.

### Solution

For advertising, added a new parameter to the BleAdvertisingParams structure, called primary_phy. 
For scanning, added a new parameter to the BleScanParams structure, called scan_phys. 

Enumerated options for primary_phy and scan_phys include:

- BLE_PHYS_1MBPS  // this is the standard, 1 MBPS format supported by all BLE devices
- BLE_PHYS_CODED // this is a 125 KBPS format introduced in BLE 5 for longer range

Note: for scanning only, these two modes can be combined (see the example in the test application below). 

### Steps to Test

To fully test this feature, two Particle Boron devices are needed. One is programmed as a peripheral that advertises, the other as a central that scans. The test applications provided will advertise and scan in both PHY_CODED and 1 MBPS modes, alternating at timed intervals.

First, flash the ble_5_phy_coded_advert_test.ino code below onto the first device, confirming correct operation with the Serial Monitor over USB. Next, flash the ble_5_phy_coded_scan_test.ino code onto the second device, again confirming correct operation with the Serial Monitor over USB. You can then move the advertising device to various distances from the scanning device to explore the range benefit for the PHYS_CODED mode .

### ble_5_phy_coded_advert_test.ino

```c
/*
 * ble_5_phy_coded_advert_test.ino
 * Description: Test app demonstrates ADVERTISING for as a long range (Coded Phy) device
 * Author: Mark Leavitt    mark@markleavitt.com
 * Date: Mar 3 2021
 * 
 * This library is free software; you can redistribute it and/or
 * modify it under the terms of the GNU Lesser General Public
 * License as published by the Free Software Foundation, either
 * version 3 of the License, or (at your option) any later version.
 *
 * This library is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
 * Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public
[...]

PR #2298, opened 2021-03-10
Ble5 long range wiring API

mgolu:
This builds on the community-submitted PR(https://github.com/particle-iot/device-os/pull/2287) to add BLE5 long range and does the following:

HAL:
- Add check when setting Advertising parameters, that the PHY is properly selected (for advertising, only one PHY can be chosen).
- When setting the extended scan parameter, check if Coded PHY is selected either by itself or bitwise with another PHY. In either of those cases, extended must be set to 1.

Wiring:
- Add setScanPhy() API
- Add setAdvertisingPhy() API

Integration/Application tests run on 2 Tracker Ones, which can communicate with each other and switch PHYs

Usage example:

Scanner:
BLE.setScanPhy(BlePhy::CODED | BlePhy::MBPS1);

Advertiser:
BLE.setAdvertisingPhy(BlePhy::CODED);

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -74,6 +74,7 @@ using namespace particle::ble;
 
 #define BLE_API_VERSION_1   1
 #define BLE_API_VERSION_2   2
+#define BLE_API_VERSION_3   3
 
 //anonymous namespace
 namespace {
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -1033,6 +1033,7 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
         tempParams.interval = BLE_DEFAULT_ADVERTISING_INTERVAL;
         tempParams.timeout = BLE_DEFAULT_ADVERTISING_TIMEOUT;
         tempParams.inc_tx_power = false;
+        tempParams.primary_phy = BLE_PHYS_AUTO;
     } else {
         memcpy(&tempParams, params, std::min(tempParams.size, params->size));
         if (tempParams.primary_phy != BLE_PHYS_AUTO && tempParams.primary_phy != BLE_PHYS_1MBPS && tempParams.primary_phy != BLE_PHYS_CODED) {
@@ -1219,8 +1220,12 @@ int BleObject::Broadcaster::resume() {
 ble_gap_adv_params_t BleObject::Broadcaster::toPlatformAdvParams(const hal_ble_adv_params_t* halParams) {
     ble_gap_adv_params_t params = {};
     if (halParams->primary_phy == BLE_PHYS_CODED) {
-        // For Coded Phy advertising, type must be extended, nonscannable
-        params.properties.type = BLE_GAP_ADV_TYPE_EXTENDED_CONNECTABLE_NONSCANNABLE_UNDIRECTED;
+        if (halParams->type == BLE_ADV_SCANABLE_UNDIRECTED_EVT) {
+            params.properties.type = BLE_GAP_ADV_TYPE_EXTENDED_NONCONNECTABLE_NONSCANNABLE_UNDIRECTED;
+        } else {
+            // For Coded Phy advertising, type must be extended, nonscannable
+            params.properties.type = BLE_GAP_ADV_TYPE_EXTENDED_CONNECTABLE_NONSCANNABLE_UNDIRECTED;
+        }
         params.primary_phy = BLE_GAP_PHY_CODED;
     } else {
         // For 1 MBPS advertising, use whatever type is specified
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -1034,7 +1034,7 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
         tempParams.inc_tx_power = false;
     } else {
         memcpy(&tempParams, params, std::min(tempParams.size, params->size));
-        if (tempParams.primary_phy != BLE_PHYS_1MBPS && tempParams.primary_phy != BLE_PHYS_CODED) {
+        if (tempParams.primary_phy != BLE_PHYS_AUTO && tempParams.primary_phy != BLE_PHYS_1MBPS && tempParams.primary_phy != BLE_PHYS_CODED) {
             LOG(ERROR, "primary_phy value not supported");
             return SYSTEM_ERROR_NOT_SUPPORTED;
         }
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -1034,6 +1034,10 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
         tempParams.inc_tx_power = false;
     } else {
         memcpy(&tempParams, params, std::min(tempParams.size, params->size));
+        if (tempParams.primary_phy != BLE_PHYS_1MBPS && tempParams.primary_phy != BLE_PHYS_CODED) {
+            LOG(ERROR, "primary_phy value not supported");
+            return SYSTEM_ERROR_NOT_SUPPORTED;
+        }
     }
     CHECK(suspend());
     if (connHandle_ != BLE_INVALID_CONN_HANDLE) {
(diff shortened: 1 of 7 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1034,6 +1034,10 @@ int BleObject::Broadcaster::setAdvertisingParams(const hal_ble_adv_params_t* par
         tempParams.inc_tx_power = false;
     } else {
         memcpy(&tempParams, params, std::min(tempParams.size, params->size));
+        if (tempParams.primary_phy != BLE_PHYS_1MBPS && tempParams.primary_phy != BLE_PHYS_CODED) {
```
XuGuohui, line 1037: `BLE_PHYS_AUTO` is allowed and the SoftDevice will apply `BLE_PHYS_1MBPS` implicitly. For old user application, the `primary_phy` field is presented as the reserved field, which is initialized to 0 (now corresponding to `BLE_PHYS_AUTO`), the new Device OS should be backward compatible and accept the `BLE_PHYS_AUTO`.

And we should remove the `reserved` field in `hal_ble_adv_params_t` to make sure the struct is word aligned still. And we should bump the `BLE_API_VERSION` in ble_hal.h, since we have changed the struct filed.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1458,7 +1462,7 @@ int BleObject::Observer::stopScanning() {
 
 ble_gap_scan_params_t BleObject::Observer::toPlatformScanParams() const {
     ble_gap_scan_params_t params = {};
-    params.extended = (scanParams_.scan_phys == BLE_PHYS_1MBPS) ? 0x00 : 0x01; /**< Extended required if other than PHYS_1MBPS */
+    params.extended = ( (scanParams_.scan_phys & BLE_PHYS_CODED) != 0) ? 0x01 : 0x00; /**< Extended must be 1 if using Coded PHY */
```
XuGuohui, line 1465: I'd like to add a `uint8_t reserved[3]` filed in `hal_ble_scan_params_t` to make sure the struct is word aligned.

wiring/inc/spark_wiring_ble.h
```diff
@@ -128,6 +128,14 @@ enum class BleAdvertisingEventType : uint8_t {
     SCANABLE_DIRECTED                       = BLE_ADV_SCANABLE_DIRECTED_EVT
 };
 
+enum class BlePhy : uint8_t {
+    AUTO        = 0x00,
```
XuGuohui, line 132: Nit: use enum names in ble_hal.h to make sure the values are consistent.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2109,6 +2109,15 @@ int BleLocalDevice::setAdvertisingType(BleAdvertisingEventType type) const {
     return hal_ble_gap_set_advertising_parameters(&advParams, nullptr);
 }
 
+int BleLocalDevice::setAdvertisingPhy(EnumFlags<BlePhy> phy) const {
+    if(phy != BlePhy::CODED && phy != BlePhy::MBPS1) return SYSTEM_ERROR_NOT_SUPPORTED;
```
XuGuohui, line 2113: We can remove this line, since we have the BLE HAL to check the validity already.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2109,6 +2109,15 @@ int BleLocalDevice::setAdvertisingType(BleAdvertisingEventType type) const {
...
+    hal_ble_adv_params_t advParams = {};
+    advParams.size = sizeof(hal_ble_adv_params_t);
```
XuGuohui, line 2115: We should also set the version field with `BLE_API_VERSION` just in case for backward compatibility. The BLE HAL may make use of this field to execute code conditionally. Please help also init this field in other wiring APIs those need to set/get HAL struct.

XuGuohui (changes requested):
Good job 👍! We also need to add API test and on-device unit test for QA. I'm glad to help with this if you're not familiar with the test framework.

hal/src/nRF52840/ble_hal.cpp
mgolu: Good catch! Added the check for `BLE_PHYS_AUTO`. Thought about checking for `BLE_PHYS_AUTO` and changing to `BLE_PHYS_1MBPS`, but if the SoftDevice already does it, we don't need to.

I removed the `reserved` field and bumped `BLE_API_VERSION`

hal/src/nRF52840/ble_hal.cpp
mgolu: Done

wiring/inc/spark_wiring_ble.h
mgolu: Done

wiring/src/spark_wiring_ble.cpp
mgolu: Done.

wiring/src/spark_wiring_ble.cpp
mgolu: The first thing that `BleObject::Broadcaster::setAdvertisingParams()` does is set the version field to `BLE_API_VERSION`, so it seems like if we set it in the Wiring function, it'll get overwritten. 

Are you thinking about changing `setAdvertisingParams()` to only set version if it's not set, that way Wiring APIs can set different values of version? Currently that's not possible.

wiring/inc/spark_wiring_ble.h
XuGuohui: I can see the enum values are still magic numbers, instead of using `BLE_PHYS_AUTO` for example.

wiring/src/spark_wiring_ble.cpp
XuGuohui: We just set the version for the temporary advertising params in `setAdvertisingParams()`. The internal cached advertising params version can be the latest so that the relevant logic in HAL can always handle the latest params properly, for example, we just introduce the new field `primary_phy` and the BLE HAL should check (although we don't check as of right now) the version and handle this field properly. While the version filed set in an old user application (which means that user built the application against an old Device OS version, but the on-device Device OS has been updated to the latest) may be lower then the version in Device OS, in that case the BLE HAL can check the version field that is passed in by user application and init/tweak some fields' value and then finally update the cached params for later usage. I know that we don't do this for now, but for better practice we should set the version field in wiring just in case we make use of this filed in BLE HAL in the future.

XuGuohui: There is an example in place: https://github.com/particle-iot/device-os/blob/develop/hal/src/nRF52840/spi_hal.cpp#L479

XuGuohui: Basically, the version field we set in BLE HAL is to make sure the internal logic comply with the latest struct, while the version field set by user application determines how we exchange the params between BLE HAL and wiring.

wiring/inc/spark_wiring_ble.h
mgolu: ah, understood! See new commit

wiring/src/spark_wiring_ble.cpp
mgolu: Ok, I think I have it now.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2089,6 +2089,7 @@ int BleLocalDevice::setAdvertisingInterval(uint16_t interval) const {
     hal_ble_adv_params_t advParams = {};
     advParams.size = sizeof(hal_ble_adv_params_t);
     CHECK(hal_ble_gap_get_advertising_parameters(&advParams, nullptr));
+    advParams.version = BLE_API_VERSION;
```
XuGuohui, line 2092: The version should be set before getting the current params. Same to elsewhere the version is set.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2474,6 +2487,14 @@ int BleLocalDevice::setScanTimeout(uint16_t timeout) const {
     return hal_ble_gap_set_scan_parameters(&scanParams, nullptr);
 }
 
+int BleLocalDevice::setScanPhy(EnumFlags<BlePhy> phy) const {
+    hal_ble_scan_params_t scanParams = {};
+    scanParams.size = sizeof(hal_ble_scan_params_t);
```
XuGuohui, line 2492: Set the version field.

XuGuohui (changes requested):

XuGuohui (approved):

avtolstoy (approved):

Issue #2291, opened 2021-02-27
Random I2C resets caused by non-monotonic HAL_Timer_Get_Micro_Seconds()


perotom:
### Expected Behavior

I2C timeout should work.

### Observed Behavior

It doesn't. I did the following observation with my logic analyzer:

![Screenshot 2021-02-27 at 19 42 04](https://user-images.githubusercontent.com/3322872/109396881-0e4b2c80-7934-11eb-8c00-13dc46faa061.png)

You can see the I2C reset happens already after 0.15ms which is very bad for some devices. I further dig into device-os and found this:

https://github.com/particle-iot/device-os/blob/c9fed1b2db0ee1d4462fae98b8bc14280958f092/hal/src/nRF52840/i2c_hal.cpp#L42-L54

It seemed like on some rare occasions the timeout is skipped. I added a debug log like:

```
#define WAIT_TIMED(timeout_ms, what) ({ \
    system_tick_t _micros = HAL_Timer_Get_Micro_Seconds();                      \
    bool res = true;                                                            \
    while ((what)) {                                                            \
        system_tick_t tmp = HAL_Timer_Get_Micro_Seconds();                          \
        system_tick_t dt = tmp - _micros;                                     \
        bool nok = ((timeout_ms * 1000 < dt) && (what));                               \
        if (nok) {                                                              \
            LOG_DEBUG(TRACE, "Timed out %lu, %lu, %lu", dt, _micros, tmp);          \
            res = false;                                                        \
            break;                                                              \
        }                                                                       \
    }                                                                           \
    res;                                                                        \
})
```

and I got this log: `TRACE: Timed out 4294966523, 670721873, 670721100`. It seems like `HAL_Timer_Get_Micro_Seconds()` is not monotonic which results in a negative value which wraps around in `system_tick_t` and results in an immediate timeout. To proof my point I changed to a signed int like:

```
#define WAIT_TIMED(timeout_ms, what) ({ \
    system_tick_t _micros = HAL_Timer_Get_Micro_Seconds();                      \
    bool res = true;                                                            \
    while ((what)) {                                                            \
        system_tick_t tmp = HAL_Timer_Get_Micro_Seconds();                          \
        int64_t dt = ((int64_t) tmp -  (int64_t)_micros);                                     \
        bool nok = ((timeout_ms * 1000 < dt) && (what));                               \
        if (nok) {                                                              \
[...]

PR #2303, opened 2021-04-06
[Gen 3] Fix micros/millis/unixtime becoming non-monotonic / Make I2C bus reset less destructive if performed in the middle of the transaction

avtolstoy:
### Problem

1. [Gen 3] Micros/millis/unixtime may become non-monotonic (especially with hot/cold ambient temperature) due to unaccounted clock drift between 32KHz RTC and `DWT->CYCCNT`
2. [Gen 3] Non-monotonic micros may cause internal I2C HAL timeouts to expire early and cause I2C HAL to perform bus reset in the middle of the transaction.
3. [Gen 2 / Gen 3] Current I2C bus reset procedure is destructive if happens in the middle of the unfinished transaction and may cause erroneous data to be written into the I2C slave. Devices populated with BQ24195 PMIC (cellular) or AM1805 RTC (tracker) may very well end up in a broken state (e.g. powered off)

See also https://github.com/particle-iot/device-os/issues/2291

### Solution

1. [Gen 3] Properly account for positive and negative clock drift between RTC and `DWT->CYCCNT`, so that the micros/millis/etc counters stay monotonic.
2. [Gen 3] Every 512 seconds (RTC overflow) clocks are properly synchronized keeping the accumulated drift/bias between them intact, so there is no sudden jump in any direction after RTC overflow.
3. [Gen 2 / Gen 3] I2C reset checks the state of the I2C bus to ensure the reset sequence is only sent when the bus is held busy by one of the slaves and attempts to finalize a potentially interrupted I2C transaction as soon as possible.
4. [HAL] Add new `OUTPUT_OPEN_DRAIN_PULLUP` pin mode.
5. [Gen 3] Add workaround for anomaly [219] TWIM: I2C timing spec is violated at 400 kHz
6. [Cellular / Tracker] Perform PMIC/RTC bus reset on boot to make to avoid clocking erroneous data if booting after e.g. a hard reset

### Steps to Test

- `wiring/no_fixture`: `I2C_07_bus_reset_is_not_destructive`
- `wiring/timer_stress` (requires temperature controlled environment, see README.md)

### Example App

N/A

### References

- Closes https://github.com/particle-iot/device-os/issues/2291

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -218,7 +218,7 @@ static void twimHandler(nrfx_twim_evt_t const * p_event, void * p_context) {
     }
 }
 
-static int twiUninit(hal_i2c_interface_t i2c) {
+static int twiUninit(hal_i2c_interface_t i2c, bool reset_pin_configuration = true) {
     if (i2cMap[i2c].state != HAL_I2C_STATE_ENABLED) {
         return SYSTEM_ERROR_INVALID_STATE;
     }
@@ -229,11 +229,13 @@ static int twiUninit(hal_i2c_interface_t i2c) {
         nrfx_twis_uninit(i2cMap[i2c].slave);
     }
 
-    // Reset pin function
-    HAL_Pin_Mode(i2cMap[i2c].scl_pin, PIN_MODE_NONE);
-    HAL_Pin_Mode(i2cMap[i2c].sda_pin, PIN_MODE_NONE);
-    HAL_Set_Pin_Function(i2cMap[i2c].scl_pin, PF_NONE);
-    HAL_Set_Pin_Function(i2cMap[i2c].sda_pin, PF_NONE);
+    if (reset_pin_configuration) {
+        // Reset pin function
+        HAL_Pin_Mode(i2cMap[i2c].scl_pin, PIN_MODE_NONE);
+        HAL_Pin_Mode(i2cMap[i2c].sda_pin, PIN_MODE_NONE);
+        HAL_Set_Pin_Function(i2cMap[i2c].scl_pin, PF_NONE);
+        HAL_Set_Pin_Function(i2cMap[i2c].sda_pin, PF_NONE);
+    }
 
     return SYSTEM_ERROR_NONE;
 }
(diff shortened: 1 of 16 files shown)


user/tests/wiring/no_fixture/i2c.cpp
```diff
@@ -115,4 +331,60 @@ test(I2C_06_I2c_Sleep_FuelGauge) {
...
+    hal_i2c_reset(HAL_PLATFORM_FUELGAUGE_MAX17043_I2C, 0, nullptr);
+    fuel.reset();
+
+    // Validate that fuel gauge has default configuration
+    uint8_t buf[2] = {0xff, 0xff};
+    assertEqual(0, fuel.readConfigRegister(buf[0], buf[1]));
+    uint16_t config = buf[0] << 8 | buf[1];
+    assertEqual(config, MAX17043_DEFAULT_CONFIG);
+
+    // Validate that we can read exactly the same configuration with soft i2c implementation
```
XuGuohui, line 370: Could you help me understand why we need to use software I2C to validate the communication with the fuel gauge?

XuGuohui (approved):

user/tests/wiring/no_fixture/i2c.cpp
avtolstoy: We use software I2C to create a situation of an interrupted transaction at a particular location e.g. clock 3 bits out 8 or clock 1 byte and 5 bits and then check whether `hal_i2c_reset` can reliably terminate this ongoing transaction without writing garbage data into the slave (previously it would have just clocked 9 high bits).

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -636,29 +648,57 @@ uint8_t hal_i2c_reset(hal_i2c_interface_t i2c, uint32_t reserved, void* reserved
...
-        //Generate 9 pulses on SCL to tell slave to release the bus
-        for (int i = 0; i < 9; i++) {
+        if (HAL_GPIO_Read(i2cMap[i2c].sda_pin) == 0) {
             HAL_GPIO_Write(i2cMap[i2c].scl_pin, 0);
-            HAL_Delay_Microseconds(100);
+            HAL_Delay_Microseconds(50);
             HAL_GPIO_Write(i2cMap[i2c].scl_pin, 1);
-            HAL_Delay_Microseconds(100);
+            HAL_Delay_Microseconds(50);
+            HAL_GPIO_Write(i2cMap[i2c].scl_pin, 0);
```
eberseth, line 682: Does it matter if this is left low as the last transition?

hal/src/stm32f2xx/gpio_hal.c
```diff
@@ -348,6 +348,15 @@ int HAL_Pin_Configure(pin_t pin, const hal_gpio_config_t* conf, void* reserved)
             PIN_MAP[pin].pin_mode = AN_OUTPUT;
             break;
 
+        case OUTPUT_OPEN_DRAIN_PULLUP:
+            GPIO_InitStructure.GPIO_Mode = GPIO_Mode_OUT;
+            GPIO_InitStructure.GPIO_OType = GPIO_OType_OD;
+            GPIO_InitStructure.GPIO_Speed = GPIO_Speed_50MHz;
```
eberseth, line 354: Generally, I2C busses don't need a really fast fall time.  I am a little concerned with generating electrical noise and ringing on the output.  Can we set this to 2MHz instead?

avtolstoy: This is not for I2C specifically. It's a general purpose IO mode. I'd prefer we keep the default the same with the existing `AF_OUTPUT_DRAIN` mode, which is similar.

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: Doesn't, because generating a stop condition below will anyway pull it correctly high.

eberseth (approved):

PR #2308, opened 2021-04-20
Restores the modem state after device is woken up.

XuGuohui:
### Problem
For sleep 2.0 API, if the modem is powered on, but not establishing connection, it won't restore the modem power state after device is woken up.
### Solution
Tracks the modem power state and connection state respectively in system sleep.
### Steps to Test
Running the example below building against develop branch you won't see the modem is on after device is woken up, but you will if you build it against the current branch.
### Example App
```c
#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);

Serial1LogHandler l(115200, LOG_LEVEL_INFO);

void setup() {
    Log.info("Turning on the modem...");
    Cellular.on();

    waitUntil(Cellular.isOn);
    Log.info("Done");
    delay(2s);

    Log.info("Entering sleep...");
    SystemSleepResult r = System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(5s));

    if (r.error() == SYSTEM_ERROR_NONE) {
        Log.info("Exited sleep.");
    } else {
        Log.error("Failed to enter sleep.");
    }

    Log.info("Waiting for the modem to be turned on...");
    waitFor(Cellular.isOn, 60000);
    Log.info("Modem is %s", Cellular.isOn() ? "on" : "off");
}

void loop() {
}
```
### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -39,8 +39,10 @@ LOG_SOURCE_CATEGORY("system.sleep");
 #include "cellular_hal.h"
 #endif // HAL_PLATFORM_CELLULAR
 #include "check.h"
+#include "system_network_manager.h"
 
 using namespace particle;
+using namespace particle::system;
 
 #undef LOG_COMPILE_TIME_LEVEL
 #define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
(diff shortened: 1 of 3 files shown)


system/src/system_sleep.cpp
```diff
@@ -45,45 +45,67 @@ using namespace particle;
...
     }
 #endif
     // Turn off the modem
-    network_off(index, 0, 0, NULL);
-    LOG(TRACE, "Waiting interface %d to be off...", (int)index);
-    // There might be up to 30s delay to turn off the modem for particular platforms.
-    network_wait_off(index, 120000/*ms*/, nullptr);
-    return resume;
+    if (!network_is_off(index, nullptr)) {
+        status.on = true;
```
avtolstoy, line 75: What if it's being powered off right now (`IF_POWER_STATE_POWERING_DOWN` on Gen 3)? This might need to have separate implementation for Gen 2 and Gen 3: `SPARK_WLAN_XXX` for Gen 2 and going into network manager on Gen 3 for `pwrState` value.

system/src/system_sleep.cpp
```diff
@@ -45,45 +45,67 @@ using namespace particle;
...
+        network_wait_off(index, 120000/*ms*/, nullptr);
+    } else {
+        LOG(TRACE, "Interface %d is off already", (int)index);
+    }
+    
+    return status;
 }
 
-static int system_sleep_network_resume(network_interface_index index) {
+int system_sleep_network_resume(network_interface_index index, network_status_t& status) {
```
avtolstoy, line 87: There is no need to pass `network_status_t` by reference here, which semantically means that it may be modified. Either by value (preferred) or at least const reference.

avtolstoy (changes requested):
Rejecting because of `status.on` logic.

`wiring/sleep20` also needs to be updated to have a set of tests to validate this behavior.

system/src/system_sleep.cpp
XuGuohui: Good catch! Will work around it.

avtolstoy (approved):

PR #2310, opened 2021-04-26
System: make sure that the power source being USB host is well detected

XuGuohui:
### Problem
This issue can be easily replicated on Tracker. 
1. Attached both battery and connect device to computer via USB cable
2. Unplug the USB cable and plug back again
3. The power source is detected as `VIN`, instead of `USB Host`

If there is a power source attached to the DC jack, pluging back USB cable won't update the power source either.
### Solution
When it detects the power source is `VIN` and the USB peripheral is powered, try to run DPDM in the system power management loop to make sure the USB host is correctly reported by PMIC
### Steps to Test
As steps to replicate the issue.
### Example App
```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

};
  "usb otg", "battery"
};

void setup() {
#if PLATFORM_ID == PLATFORM_TRACKER
  // We need to enable the CAN_5V to connect the Serial1 to M8.
  pinMode(CAN_PWR, OUTPUT);
  digitalWrite(CAN_PWR, HIGH);
#endif

  Serial1.begin(115200);
  Serial1.printlnf("Application started.");

  System.on(battery_state, [](system_event_t event, int data) -> void {
    Serial1.printlnf("battery_state: %s", batteryStates[std::max(0, data)]);
  });

  System.on(power_source, [](system_event_t event, int data) -> void {
    Serial1.printlnf("power_source: %s", powerSources[std::max(0, data)]);
  });
}

void loop() {
}
```

### References
- ch77765

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_power_manager.cpp
+++ b/system/src/system_power_manager.cpp
@@ -70,6 +70,9 @@ constexpr uint16_t PMIC_FAULT_TERM_CHARGE_CURRENT = 128; // mA
 constexpr system_tick_t BATTERY_REPEATED_CHARGED_WINDOW = 5000; // ms
 constexpr uint8_t BATTERY_REPEATED_CHARGED_COUNT = 2;
 
+constexpr system_tick_t DPDM_PERIOD = 1000;
+constexpr uint8_t DPDM_RETRY_COUNT = 3;
+
 constexpr hal_power_config defaultPowerConfig = {
   .flags = 0,
   .version = 0,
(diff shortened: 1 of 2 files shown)


avtolstoy (changes requested):

system/src/system_power_manager.cpp
```diff
@@ -505,6 +510,31 @@ void PowerManager::initDefault(bool dpdm) {
...
 
+void PowerManager::runDpdm() {
+// FIXME: maybe introduce a new flag?
+#if HAL_PLATFORM_POWER_WORKAROUND_USB_HOST_VIN_SOURCE
+  static system_tick_t lastRun = 0;
+  static uint8_t dpdmRetry = 0;
+
+  if (millis() - lastRun >= DPDM_PERIOD) {
+    lastRun = millis();
+    if (g_powerSource == POWER_SOURCE_VIN && HAL_USB_Get_State(nullptr) >= HAL_USB_STATE_POWERED) {
```
eberseth, line 521: what other states are caught by `HAL_USB_Get_State(nullptr) >= HAL_USB_STATE_POWERED`?  Can you make this more explicit such as a switch/case statement?

eberseth:
There are cases where the status register is showing 0x24 and the device is being powered through USB.  Is there something not configured completely for USB detection?

PR #2300, opened 2021-03-12
[Release] v3.0.0-rc.2

avtolstoy:
Review only

PR #2315, opened 2021-05-12
Add an API to get the maximum supported size of event data

sergeuz:
### Problem

1. The maximum supported sizes of an event's payload data, variable value and a function call argument no longer depend solely on the Device OS version and there's no way to determine their values at runtime in the application code.
3. The constants defined in protocol_defs.h (`MAX_EVENT_DATA_LENGTH`, etc.), although undocumented, not necessarily reflect the actual limits either.

### Solution

Add an API to query the actual values of the aforementioned parameters:
```cpp
Particle.maxEventDataSize()
Particle.maxVariableValueSize()
Particle.maxFunctionArgumentSize()
```

**Note:** This PR also fixes the maximum supported size of event data at 782 bytes on Boron/BSoM devices running an old modem firmware (02.03 and older). [Previously](https://github.com/particle-iot/device-os/pull/2300), it was possible to send events with up to 782-845 bytes of payload data – depending on the length of the event name.

### Steps to Test

Run `communication/*` integration tests.

### References

- [ch75605]


communication/src/spark_protocol_functions.cpp
```diff
@@ -196,41 +196,72 @@ int spark_protocol_set_connection_property(ProtocolFacade* protocol, unsigned pr
...
     }
     default:
-        return particle::protocol::ProtocolError::NOT_IMPLEMENTED;
+        return ProtocolError::NOT_IMPLEMENTED;
+    }
+}
+
+int spark_protocol_get_connection_property(ProtocolFacade* protocol, unsigned property, void* data, size_t* size, void* reserved)
+{
+    ASSERT_ON_SYSTEM_THREAD();
```
avtolstoy, line 239: This shouldn't work for threaded applications as of right with how this is used in `CloudClass`. Add `spark_get_connection_property()` to schedule to run on system thread?

avtolstoy (approved):

PR #2316, opened 2021-05-17
Add system watchdog API for user applications

eberseth:
### Problem

Applications (tracker) must instantiate their own version of RTC drivers to manage the watchdog.  While the device OS handles synchronization of time and general RTC functions, it does not expose the RTC watchdog features, if any.  Application code size may be reduced by not having the additional/redundant driver for things that device OS may already handle.

### Solution

Expose the watchdog portions of the HAL RTC driver to the application so that the device OS can offer it without a separate application driver.  Four functions would be essential to support the watchdog:
1. Enable/configure
2. Disable
3. Reset/kick
4. Check limits of watchdog timeouts

### Steps to Test

1. Setup the watchdog per the user app below, observe that the device does not reset.
2. Comment out the inner `loop()` portion of code and observe the device resetting after 60 seconds.

### Example App

```c
#include "exrtc_hal.h"

constexpr system_tick_t WatchdogTimeout = 60 * 1000;  // Watchdog will expire after 60 seconds
constexpr unsigned int WatchdogKickTime = 10;  // Watchdog will be kicked every 10 seconds
static unsigned int kickTime = 0;

void setup() {
  hal_exrtc_enable_watchdog(WatchdogTimeout, nullptr);
  hal_exrtc_feed_watchdog(nullptr);
}

void loop() {
  if ((System.uptime() - kickTime) >= WatchdogKickTime) {
    kickTime = System.uptime();
    hal_exrtc_feed_watchdog(nullptr);
  }
}
```

### References

- [ch79352]

### DIFF
--- a/hal/inc/hal_dynalib.h
+++ b/hal/inc/hal_dynalib.h
@@ -24,6 +24,8 @@
 
 #include "dynalib.h"
 
+#include "hal_platform.h"
+
 #include "time_compat.h"
 
 #ifdef DYNALIB_EXPORT
@@ -33,6 +35,11 @@
 #include "timer_hal.h"
 #include "rtc_hal.h"
 #include "interrupts_hal.h"
+
+#if PLATFORM_ID == PLATFORM_TRACKER
+#include "exrtc_hal.h"
+#endif // PLATFORM_ID
+
 #endif
 
 /**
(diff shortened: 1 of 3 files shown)


hal/inc/hal_dynalib.h
```diff
@@ -33,6 +35,11 @@
 #include "timer_hal.h"
 #include "rtc_hal.h"
 #include "interrupts_hal.h"
+
+#if PLATFORM_ID == PLATFORM_TRACKER
```
avtolstoy, line 39: Nit: might be `HAL_PLATFORM_EXTERNAL_RTC`

avtolstoy (approved):

PR #2317, opened 2021-05-18
[Gen3] Support for SARA R410 - 05.12 firmware

keeramis:
### Differences between 05.12 and older versions

1. The SIM select functionality for Kore ATT defaults the MNO profile to 198 (which does not support ATT band 5) which for older radio firmware defaults to UMNOPROF:2 (which supports Band 5 along with other ATT bands) [ch78344](https://app.clubhouse.io/particle/story/78344/r410-05-12-umnoprof-s-kore-att-equivalent-now-defaults-to-198-instead-of-2)
2. Radio does not respond over data channel after some cold boot runs [ch78322](https://app.clubhouse.io/particle/story/78322/r410-05-12-radio-does-not-respond-over-data-channel-after-some-cold-boot-runs)
3. PSM/eDRX are enabled by default with UMNOPROF=100. Disable it before (it's idle timer expires when) AT becomes unresponsive [ch78529](https://app.clubhouse.io/particle/story/78529/r410-05-12-psm-enabled-by-default-with-umnoprof-100-disable-it-before-it-s-idle-timer-expires-when-at-becomes-unresponsive)

### Changes / Features

1. Hard code UMNOPROF to 2 (ATT) if Kore ATT is detected on this radio firmware
2. There is some kind of a bug in 02.19 R410 modem firmware in where it does not accept any AT commands over the second muxed channel, unless we send something over channel 1 first. Separately, [fixed resumeChannel in gsm0710muxer](https://github.com/particle-iot/gsm0710muxer/pull/9/files) which is potentially related.
3. Refactor sim selection and umnoprof code such that we are setting the hardware flow control (using IFC=2,2) and the run time baud rate as soon as possible (after sim is ready), which prevents longer wait times when the radio recovers from a reset, potentially prevents PSM taking over (and AT becoming unresponsive)

### Steps to Test

#### 1. Kore ATT SIMs
Test with Kore ATT SIM and ensure UMNOPROF: 2 is selected

#### 2. Cold boot data channel responsiveness
Use any tinker-like app, cold boot the device, check if the modem is responding to AT interface over data channel every single run. May need to test multiple times as it's not reproducible every run.

#### 3. Verify PSM does not take over after EFS backup/restore
1. Use Twilio SIM which sets the UMNOPROF=100 (which enables PSM by default)
2. Use three clicks (backup) followed by four clicks (restore), and ensure the modem recovers without getting stuck in AT unresponsive state.

```c
#include "Particle.h"

Serial1LogHandler log1Handler(115200, LOG_LEVEL_ALL);

SYSTEM_MODE(SEMI_AUTOMATIC);

bool clicks_3 = false;
bool clicks_4 = false;
bool clicks_5 = false;

void button_handler(system_event_t event, int clicks) {
        if (clicks == 3) {
            clicks_3 = true;
        } else if (clicks == 4) {
            clicks_4 = true;
        } else if (clicks == 5) {
            clicks_5 = true;
        }
    }
}

void setup() {
    // Cellular.setActiveSim(EXTERNAL_SIM);
    Particle.connect();
}

void loop() {
    if (clicks_3 || clicks_4 || clicks_5) {
        if (clicks_3) {
            Log.info("\r\n\r\nAT+UBKUPDATA?");
            Cellular.command(1000, "AT+UBKUPDATA?\r\n");
        } else if (clicks_4) {
            Log.info("\r\n\r\nAT+UBKUPDATA=r");
            Cellular.command(1000, "AT+UBKUPDATA=r\r\n");
        } else if (clicks_5) {
            Log.info("\r\n\r\nAT+UBKUPDATA");
            Cellular.command(1000, "AT+UBKUPDATA\r\n");
        }
        clicks_3 = false;
        clicks_4 = false;
        clicks_5 = false;

        uint32_t start = millis();
        while (millis() - start < 20000UL) {
            if (RESP_OK == Cellular.command(1000, "AT\r\n")) {
                break;
            }
        }
    }
}
```

### References

- [ch78529](https://app.clubhouse.io/particle/story/78529/r410-05-12-psm-enabled-by-default-with-umnoprof-100-disable-it-before-it-s-idle-timer-expires-when-at-becomes-unresponsive)
- [ch78322](https://app.clubhouse.io/particle/story/78322/r410-05-12-radio-does-not-respond-over-data-channel-after-some-cold-boot-runs)
- [ch78344](https://app.clubhouse.io/particle/story/78344/r410-05-12-umnoprof-s-kore-att-equivalent-now-defaults-to-198-instead-of-2)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -96,6 +96,7 @@ const auto UBLOX_NCP_R4_APP_FW_VERSION_MEMORY_LEAK_ISSUE = 200;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MIN = 200;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_NO_HW_FLOW_CONTROL_MAX = 203;
 const auto UBLOX_NCP_R4_APP_FW_VERSION_LATEST_02B_01 = 204;
+const auto UBLOX_NCP_R4_APP_FW_VERSION_0512 = 219;
 
 const auto UBLOX_NCP_MAX_MUXER_FRAME_SIZE = 1509;
 const auto UBLOX_NCP_KEEPALIVE_PERIOD = 5000; // milliseconds
@@ -1118,6 +1119,12 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                     (netConf_.netProv() == CellularNetworkProvider::KORE_ATT &&
                     static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::ATT)) ) {
                 int newProf = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
+                // Hard code ATT for 05.12 firmware versions
+                if (netConf_.netProv() == CellularNetworkProvider::KORE_ATT) {
+                    if (getAppFirmwareVersion() == UBLOX_NCP_R4_APP_FW_VERSION_0512) {
+                        newProf = static_cast<int>(UbloxSaraUmnoprof::ATT);
+                    }
+                }
                 // TWILIO Super SIM
                 if (netConf_.netProv() == CellularNetworkProvider::TWILIO) {
                     // _oldFirmwarePresent: u-blox firmware 05.06* and 05.07* does not have
@@ -1176,9 +1183,9 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
                 CHECK_PARSER_OK(parser_.execCommand("AT+CFUN=15,0"));
                 HAL_Delay_Milliseconds(2000);
 				// Radio sometimes takes a while to resetfrom CFUN=15
-                CHECK(waitAtResponseFromPowerOn(state, 20000));
+                CHECK(waitAtResponseFromPowerOn(state, 6000));
 				// Prevent modem from immediately dropping into PSM/eDRX modes
-				// which (on 05.12) are enabled as soon as the UMNOPROF has taken effect
+				// which (on 05.12) may be enabled as soon as the UMNOPROF has taken effect
                 if (disableLowPowerModes) {
                     disablePsmEdrx();
                 }
@@ -1191,6 +1198,20 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
 }
 
 int SaraNcpClient::disablePsmEdrx() {
+
+    // Check SIM state as PSM command's readiness could depend on it
+    int simState = 0;
+    for (unsigned attempts = 0; attempts < 10; attempts++) {
+        simState = checkSimCard();
+        if (!simState) {
+            break;
+        }
+        HAL_Delay_Milliseconds(1000);
+    }
+    if (simState != SYSTEM_ERROR_NONE) {
+        return simState;
+    }
+
     // Force Power Saving mode to be disabled
     //
     // TODO: if we enable this feature in the future add logic to CHECK_PARSER macro(s)
@@ -1234,11 +1255,11 @@ int SaraNcpClient::waitAtResponseFromPowerOn(ModemState& modemState, unsigned in
     // Make sure we reset back to using hardware serial port @ default baudrate
     muxer_.stop();
 
-    unsigned int defaultBaudrate = ncpId() != PLATFORM_NCP_SARA_R410 ?
+    unsigned int attemptedBaudRate = ncpId() != PLATFORM_NCP_SARA_R410 ?
             UBLOX_NCP_DEFAULT_SERIAL_BAUDRATE :
             UBLOX_NCP_RUNTIME_SERIAL_BAUDRATE_R4;
 
-    CHECK(serial_->setBaudRate(defaultBaudrate));
+    CHECK(serial_->setBaudRate(attemptedBaudRate));
     CHECK(initParser(serial_.get()));
     skipAll(serial_.get(), 1000);
     parser_.reset();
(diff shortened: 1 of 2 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1054,129 +1175,55 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+    // that sim readiness is already checked
+    // CHECK(checkSimReadiness());
+
+    // Force Power Saving mode to be disabled
+    //
+    // TODO: if we enable this feature in the future add logic to CHECK_PARSER macro(s)
+    // to wait longer for device to become active (see MDMParser::_atOk)
+    auto resp = parser_.sendCommand("AT+CPSMS?");
+    unsigned psmState = 1;
+    auto r = resp.scanf("+CPSMS:%d", &psmState);
```
avtolstoy, line 1198: Nit: `+CPSMS: %u` ? space + unsigned/int type mismatch. Or `psmState` should be `int`.

Also `CHECK_PARSER(resp.scanf(...));`, return value can be ignored if you are defaulting `psmState = 1`.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1054,129 +1175,55 @@ int SaraNcpClient::selectSimCard(ModemState& state) {
...
+    resp.readResult();
```
avtolstoy, line 1199: `CHECK_PARSER(resp.readResult());`

We don't need to check for OK response, but generic errors should still be handled (e.g. a case where we've completely failed to get a response, or low level communication error).

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -966,6 +967,126 @@ int SaraNcpClient::checkNetConfForImsi() {
     return SYSTEM_ERROR_TIMEOUT;
 }
 
+int SaraNcpClient::selectNetworkProf(ModemState& state) {
+    bool reset = false;
```
avtolstoy, line 971: Nit: consider moving this and `disableLowPowerModes` a bit further down before the loop/inside the loop where it becomes relevant?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -966,6 +967,126 @@ int SaraNcpClient::checkNetConfForImsi() {
...
+        if (r == 1 && (static_cast<UbloxSaraUmnoprof>(curProf) == UbloxSaraUmnoprof::SW_DEFAULT ||
+                (netConf_.netProv() == CellularNetworkProvider::TWILIO &&
+                static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::STANDARD_EUROPE) ||
+                (netConf_.netProv() == CellularNetworkProvider::KORE_ATT &&
+                static_cast<UbloxSaraUmnoprof>(curProf) != UbloxSaraUmnoprof::ATT)) ) {
+            int newProf = static_cast<int>(UbloxSaraUmnoprof::SIM_SELECT);
+
+            // Hard code ATT for 05.12 firmware versions
+            if (netConf_.netProv() == CellularNetworkProvider::KORE_ATT) {
+                if (getAppFirmwareVersion() == UBLOX_NCP_R4_APP_FW_VERSION_0512) {
```
avtolstoy, line 1010: Nit: `fwVersion_`, it has already been cached.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -966,6 +967,126 @@ int SaraNcpClient::checkNetConfForImsi() {
...
+                        break;
+                    } else {
+                        newProf = static_cast<int>(UbloxSaraUmnoprof::SW_DEFAULT);
+                    }
+                } else {
+                    newProf = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                }
+            }
+            // KORE AT&T or 3rd Party SIM
+            else {
```
avtolstoy, line 1029: @keeramis Should this apply to the 05.12 workaround as well?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -966,6 +967,126 @@ int SaraNcpClient::checkNetConfForImsi() {
...
+
+            CHECK(waitAtResponseFromPowerOn(state));
+
+            // Checking for SIM readiness ensures that other related commands
+            // like IFC or PSM/eDRX won't error out
+            CHECK(checkSimReadiness());
+            // Prevent modem from immediately dropping into PSM/eDRX modes
+            // which (on 05.12) may be enabled as soon as the UMNOPROF has taken effect
+            if (disableLowPowerModes) {
+                // Not checking the error
```
avtolstoy, line 1081: Any specific reason? If something goes wrong when we try to disable PSM/eDRX we may accidentally continue initialization with either of those enabled.

avtolstoy (changes requested):
:tada: Requesting changes only because of `CHECK_PARSER` in `disablePsmEdrx()`.

avtolstoy:
@keeramis Make sure to rebase on develop, otherwise CI will fail to build (we've had a gcovr dependency problem recently, should be fixed now on current `develop`)

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -966,6 +967,126 @@ int SaraNcpClient::checkNetConfForImsi() {
...
+            HAL_Delay_Milliseconds(2000);
+
+            CHECK(waitAtResponseFromPowerOn(state));
+
+            // Checking for SIM readiness ensures that other related commands
+            // like IFC or PSM/eDRX won't error out
+            CHECK(checkSimReadiness());
+            // Prevent modem from immediately dropping into PSM/eDRX modes
+            // which (on 05.12) may be enabled as soon as the UMNOPROF has taken effect
+            if (disableLowPowerModes) {
```
keeramis, line 1080: @avtolstoy Could you also please shed some light on this?
If for some reason, UMNOPROF is set correctly with the first attempt, it will not try to set UMNOPROF again. But during the first time's reset, if any of the `checkSimReadiness` or `waitAtResponseFromPowerOn` have failed, we won't disable psm/edrx and since the CHECK() fails, we exit with "Failed to perform earlier init..". Next, as a part of init, we will check UMNOPROF, but it's already set to the desired value so it won't enter `if (reset)`, in which case disablePsmEdrx() will not be called. It's a very corner case, but just wanted to put a thought out there.

So here's a sequence for example (may be Twilio):
```
1. UMNOPROF reads as 0
2. UMNOPROF set to 100
3. Reset using CFUN=15.
4. Enters `if (reset)`. Let's say `waitAtResponseFromPowerOn` or `checkSimReadiness` has failed for some reason. We go no further and exit
5. Next round, we check UMNOPROF, it's already 100, so we don't enter the `if (reset)` loop which has `disablePsmEdrx` , hence potentially not disabling psm/edrx. 
```
This may be a corner case that I don't know if it can happen.

avtolstoy: @keeramis It will still happen [here](https://github.com/particle-iot/device-os/pull/2317/files#diff-a8e99f7ec884e049b049347a881b289b8edd5641af45e8d407cb97a78fdf7550R1368), right?

hal/network/ncp_client/sara/sara_ncp_client.cpp
keeramis: I have seen some runs fail PSM with something like the following
```
AT+CPSMS=0
+CME ERROR: SIM Failure
```
, and I thought not to make this a blocking call. But then, we are looking for `CHECK(checkSimReadiness())` so it should be OK. I will add that in.

hal/network/ncp_client/sara/sara_ncp_client.cpp
keeramis: Yes, that is correct. It cannot escape from disablement - that is why, kept those calls [there](https://github.com/particle-iot/device-os/pull/2317/files#diff-a8e99f7ec884e049b049347a881b289b8edd5641af45e8d407cb97a78fdf7550R1368) intact. Cool. That should be OK.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1541,6 +1569,34 @@ int SaraNcpClient::checkSimCard() {
...
+    unsigned attempts = 0;
+    for (attempts = 0; attempts < 10; attempts++) {
+        simState = checkSimCard();
+        if (!simState) {
+            break;
+        }
+        HAL_Delay_Milliseconds(1000);
+    }
+
+    if (checkForRfReset && attempts != 0 && ncpId() == PLATFORM_NCP_SARA_R410) {
```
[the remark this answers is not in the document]
keeramis: That's my observation too. Have seen it not work the first time many times. This code though, I just moved it over from [HERE](https://github.com/particle-iot/device-os/blob/develop/hal/network/ncp_client/sara/sara_ncp_client.cpp#L1075-L1083) as a part of refactoring.....

hal/network/ncp_client/sara/sara_ncp_client.cpp
keeramis: Correct. Great catch. thanks! Noticed that 2 tries is an optimal number for ATT too. Added

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1193,10 +1199,11 @@ int SaraNcpClient::disablePsmEdrx() {
     //
     // TODO: if we enable this feature in the future add logic to CHECK_PARSER macro(s)
     // to wait longer for device to become active (see MDMParser::_atOk)
+    // Format is +CPSMS:0,,,"00010011","00000011"
```
avtolstoy, line 1202: :+1:

avtolstoy (approved):

PR #2322, opened 2021-05-26
[Gen 3] 256KB application support

avtolstoy:
### Description

Refer to the [SDD](https://docs.google.com/document/d/1DvOFlKDMDqB1JotDDHTFnWLNPyqYZ9ZADfSXr4W8P9U/edit)

NOTE: this PR also enables LTO on all Gen 3 platforms when building system parts.

**NOTE: do not use current version of CLI to flash new 256KB binaries over DFU, this will fail**

### Steps to Test

NOTE: `3.1.0-alpha.1` and `3.1.0-alpha.2` releases available on staging can be used for testing (`test/v3.1.0-alpha.1` and `test/v3.1.0-alpha.2` are the respective branches). The releases are fully available on staging and appropriate version of DS is also deployed there.

1. Building larger than 128KB applications should be possible with this branch. They should also run just fine
2. 128KB older applications should still work correctly
3. `particle serial inspect` should correctly report either type of an application (index = 1, 128KB max) (index = 2, 256KB max)
4. Flashing 128KB older app should override the current 256KB app, even if they don't overlap
5. Flashing 256KB app should override (invalidate) the older 128KB  app residing on the device even if they don't overlap
6. Bootloader should function correctly with older versions of system-part1 on the device (mainly the bootloader updates through MBR are a concern and application updates)

### Example App

N/A

### References

- [CH60338]
- There is an appropriate PR in DS repo

---

### Completeness

- [ ] Added documentation


platform/MCU/nRF52840/src/flash_mal.c
```diff
@@ -368,23 +368,31 @@ int FLASH_CopyMemory(flash_device_t sourceDeviceID, uint32_t sourceAddress,
...
 #if SOFTDEVICE_MBR_UPDATES
-    if (module_function == MODULE_FUNCTION_BOOTLOADER && destinationAddress == USER_FIRMWARE_IMAGE_LOCATION) {
+    if (module_function == MODULE_FUNCTION_BOOTLOADER && destinationAddress == USER_FIRMWARE_IMAGE_LOCATION_COMPAT) {
         // Backup user firmware
         dest_size = CEIL_DIV(length, INTERNAL_FLASH_PAGE_SIZE) * INTERNAL_FLASH_PAGE_SIZE;
         if (!FLASH_EraseMemory(FLASH_SERIAL, EXTERNAL_FLASH_RESERVED_ADDRESS, dest_size)) {
             return FLASH_ACCESS_RESULT_ERROR;
         }
-        if (!flash_copy(FLASH_INTERNAL, USER_FIRMWARE_IMAGE_LOCATION, FLASH_SERIAL, EXTERNAL_FLASH_RESERVED_ADDRESS, dest_size)) {
+        if (!flash_copy(FLASH_INTERNAL, USER_FIRMWARE_IMAGE_LOCATION_COMPAT, FLASH_SERIAL, EXTERNAL_FLASH_RESERVED_ADDRESS, dest_size)) {
```
XuGuohui, line 377: If we are running the application@0xb4000 and its size is larger than 128KB, we need to backup it instead?

XuGuohui (changes requested):

bootloader/src/nRF52840/usbd_dfu_mal.cpp
```diff
@@ -73,6 +74,16 @@ int InternalFlashMal::write(const uint8_t* buf, uintptr_t addr, size_t len) {
         return 1;
     }
 
+    if (addr == USER_FIRMWARE_IMAGE_LOCATION && len >= sizeof(module_info_t)) {
```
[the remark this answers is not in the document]
avtolstoy: We can. This is guaranteed by DFU interface `wTransferSize` which is set to 4096. For all intents and purposes there cannot be a case where a valid binary is flashed and a transfer size is less than `module_info_t`.

platform/MCU/nRF52840/src/flash_mal.c
avtolstoy: We don't need to do that. This change simply keeps the existing behavior where the part of the 0xd40000 application (and now potentially 0xb40000 application) is temporarily used for storing the bootloader binary for MBR update. For all intents and purposes it does not matter where we take the chunk out of temporarily (it's anyway going to be restored after the bootloader update is applied). This is just specifically kept at the same locations as < 3.1.0 Device OS "just in case".

XuGuohui (approved):

XuGuohui:
- [ ] They should also run just fine 128KB older applications should still work correctly
- [ ] Building larger than 128KB applications should be possible with this branch.
```cpp
constexpr uint32_t BUF_SIZE = 250 * 1024;
const uint8_t buf[BUF_SIZE] = {};

void setup() {
    for (uint32_t i = 0; i < BUF_SIZE; i++) {
        Serial.printf("%d", buf[i]);
    }
}
```
- [ ] particle serial inspect should correctly report either type of an application (index = 1, 128KB max) (index = 2, 256KB max)
- [ ] Flashing 128KB older app should override the current 256KB app, even if they don't overlap
- [ ] Flashing 256KB app should override (invalidate) the older 128KB app residing on the device even if they don't overlap
- [ ] Bootloader should function correctly with older versions of system-part1 on the device. This may surprise user that the device enters safe mode after downgrading the pre Device OS versions. This is an expected behavior and it would be better to document this somewhere.

Also, after updating the bootloader, system part1 and the 256K user application, their module version isn't bumped. Should we bump the versions in this branch or postpone to when we cut a release?

modules/shared/nRF52840/inc/system_part1_loader.c
```diff
@@ -1,17 +1,18 @@
...
 
 /**
  * Determines if the user module is present and valid.
  * @return
  */
-bool is_user_module_valid()
+bool run_user_module()
 {
-    return module_user_part_validated;
+    return system_mode() != SAFE_MODE;
```
sergeuz, line 15: :+1:

platform/MCU/nRF52840/src/flash_mal.c
```diff
@@ -368,23 +368,31 @@ int FLASH_CopyMemory(flash_device_t sourceDeviceID, uint32_t sourceAddress,
         return FLASH_ACCESS_RESULT_BADARG;
     }
 #if SOFTDEVICE_MBR_UPDATES
-    if (module_function == MODULE_FUNCTION_BOOTLOADER && destinationAddress == USER_FIRMWARE_IMAGE_LOCATION) {
+    if (module_function == MODULE_FUNCTION_BOOTLOADER && destinationAddress == USER_FIRMWARE_IMAGE_LOCATION_COMPAT) {
```
sergeuz, line 371: Note that `verify_module()` still uses `USER_FIRMWARE_IMAGE_LOCATION` and not `USER_FIRMWARE_IMAGE_LOCATION_COMPAT`.

sergeuz (changes requested):

platform/MCU/nRF52840/src/flash_mal.c
avtolstoy: Good catch, thanks!

hal/src/nRF52840/ota_flash_hal.cpp
```diff
@@ -173,10 +173,21 @@ void HAL_System_Info(hal_system_info_t* info, bool construct, void* reserved)
...
 #endif
+            bool user_module_found = false;
+            memset(info->modules, 0, sizeof(hal_module_t) * count);
             for (unsigned i = 0; i < count; i++) {
                 const auto bounds = module_bounds[i];
                 const auto module = info->modules + i;
-                fetch_module(module, bounds, false, MODULE_VALIDATION_INTEGRITY);
+                if (bounds->module_function == MODULE_FUNCTION_USER_PART && user_module_found) {
+                    // Make sure that we report only single user part (either 128KB or 256KB) in the
+                    // list of modules.
```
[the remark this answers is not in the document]
avtolstoy: Added

platform/MCU/nRF52840/src/flash_mal.c
avtolstoy: Double checked bootloader updates. Works fine with system-part out of this branch and older releases as well.

sergeuz (approved):

PR #2331, opened 2021-06-25
Add ability to set and scan extended advertisement size (when using Coded PHY)

mgolu:
### Problem

Now that 3.1 supports Coded PHY, we should also support scanning for advertisements that are up to 255 bytes instead of 31 bytes. 

### Solution

This PR changes `BleAdvertisingData's` buffer for advertisement data from an array with fixed size of 31, to a Vector. I selected this rather than an array of 255 bytes to save RAM, since an instance of `BleAdvertisingData` is created for every `BleScanResult`. 

For the Broadcaster, this PR changes the fixed size array from 31 (BLE_MAX_ADV_DATA_LEN) to 255 (BLE_MAX_SCAN_REPORT_BUF_LEN). Since there's a single instance of the Broadcaster, I think this is ok, although we could make this a Vector as well.

### Steps to Test

Added tests in `user/tests/wiring/ble_scanner_broadcaster` for both the Broadcaster and the Scanner.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -327,7 +327,7 @@ private:
     volatile hal_ble_auto_adv_cfg_t autoAdvCfg_;    /**< Automatic advertising configuration. */
     uint8_t advHandle_;                             /**< Advertising handle. */
     hal_ble_adv_params_t advParams_;                /**< Current advertising parameters. */
-    uint8_t advData_[BLE_MAX_SCAN_REPORT_BUF_LEN];  /**< Current advertising data. */
+    uint8_t advData_[BLE_MAX_ADV_DATA_LEN_EXT];     /**< Current advertising data. */
     size_t advDataLen_;                             /**< Current advertising data length. */
     uint8_t scanRespData_[BLE_MAX_ADV_DATA_LEN];    /**< Current scan response data. */
     size_t scanRespDataLen_;                        /**< Current scan response data length. */
@@ -391,7 +391,7 @@ private:
     volatile bool isScanning_;                              /**< If it is scanning or not. */
     hal_ble_scan_params_t scanParams_;                      /**< BLE scanning parameters. */
     os_semaphore_t scanSemaphore_;                          /**< Semaphore to wait until the scan procedure completed. */
-    uint8_t scanReportBuff_[BLE_MAX_SCAN_REPORT_BUF_LEN];   /**< Buffer to hold the scanned report data. */
+    uint8_t scanReportBuff_[BLE_MAX_ADV_DATA_LEN_EXT];   /**< Buffer to hold the scanned report data. */
     ble_data_t bleScanData_;                                /**< BLE scanned data. */
     hal_ble_on_scan_result_cb_t scanResultCallback_;        /**< Callback function on scan result. */
     void* context_;                                         /**< Context of the scan result callback function. */
@@ -1090,7 +1090,7 @@ int BleObject::Broadcaster::setAdvertisingData(const uint8_t* buf, size_t len) {
     // It is invalid to provide the same data buffers while advertising.
     CHECK(suspend());
     if (buf != nullptr) {
-        len = std::min(len, (size_t)BLE_MAX_SCAN_REPORT_BUF_LEN);
+        len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN_EXT);
         memcpy(advData_, buf, len);
     } else {
         len = 0;
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -391,7 +391,7 @@ private:
     volatile bool isScanning_;                              /**< If it is scanning or not. */
     hal_ble_scan_params_t scanParams_;                      /**< BLE scanning parameters. */
     os_semaphore_t scanSemaphore_;                          /**< Semaphore to wait until the scan procedure completed. */
-    uint8_t scanReportBuff_[BLE_MAX_ADV_DATA_LEN_EXT];   /**< Buffer to hold the scanned report data. */
+    uint8_t scanReportBuff_[BLE_MAX_SCAN_REPORT_BUF_LEN];   /**< Buffer to hold the scanned report data. */
     ble_data_t bleScanData_;                                /**< BLE scanned data. */
     hal_ble_on_scan_result_cb_t scanResultCallback_;        /**< Callback function on scan result. */
     void* context_;                                         /**< Context of the scan result callback function. */
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -329,7 +329,7 @@ private:
     hal_ble_adv_params_t advParams_;                /**< Current advertising parameters. */
     uint8_t advData_[BLE_MAX_ADV_DATA_LEN_EXT];     /**< Current advertising data. */
     size_t advDataLen_;                             /**< Current advertising data length. */
-    uint8_t scanRespData_[BLE_MAX_ADV_DATA_LEN];    /**< Current scan response data. */
+    uint8_t scanRespData_[BLE_MAX_SCAN_REPORT_BUF_LEN];    /**< Current scan response data. */
     size_t scanRespDataLen_;                        /**< Current scan response data length. */
     int8_t txPower_;                                /**< TX Power. */
     bool advPending_;                               /**< Advertising is pending. */
@@ -391,7 +391,7 @@ private:
     volatile bool isScanning_;                              /**< If it is scanning or not. */
     hal_ble_scan_params_t scanParams_;                      /**< BLE scanning parameters. */
     os_semaphore_t scanSemaphore_;                          /**< Semaphore to wait until the scan procedure completed. */
-    uint8_t scanReportBuff_[BLE_MAX_SCAN_REPORT_BUF_LEN];   /**< Buffer to hold the scanned report data. */
+    uint8_t scanReportBuff_[BLE_MAX_ADV_DATA_LEN_EXT];      /**< Buffer to hold the scanned report data. */
     ble_data_t bleScanData_;                                /**< BLE scanned data. */
     hal_ble_on_scan_result_cb_t scanResultCallback_;        /**< Callback function on scan result. */
     void* context_;                                         /**< Context of the scan result callback function. */
@@ -1429,7 +1429,7 @@ int BleObject::Observer::startScanning(hal_ble_on_scan_result_cb_t callback, voi
     scanResultCallback_ = callback;
     context_ = context;
     int ret = sd_ble_gap_scan_start(&bleGapScanParams, &bleScanData_);
-    // LOG_DEBUG(TRACE,"Ret code sd_ble_scan_start: %d", ret);  // Uncomment to get error code from starting scan
+    LOG_DEBUG(TRACE,"Ret code sd_ble_scan_start: %d", ret);  // Uncomment to get error code from starting scan
     CHECK_NRF_RETURN(ret, nrf_system_error(ret));
     isScanning_ = true;
     // If timeout is set to 0, it should scan indefinitely
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -327,7 +327,7 @@ private:
     volatile hal_ble_auto_adv_cfg_t autoAdvCfg_;    /**< Automatic advertising configuration. */
     uint8_t advHandle_;                             /**< Advertising handle. */
     hal_ble_adv_params_t advParams_;                /**< Current advertising parameters. */
-    uint8_t advData_[BLE_MAX_ADV_DATA_LEN];         /**< Current advertising data. */
+    uint8_t advData_[BLE_MAX_SCAN_REPORT_BUF_LEN];  /**< Current advertising data. */
     size_t advDataLen_;                             /**< Current advertising data length. */
     uint8_t scanRespData_[BLE_MAX_ADV_DATA_LEN];    /**< Current scan response data. */
     size_t scanRespDataLen_;                        /**< Current scan response data length. */
@@ -1090,7 +1090,7 @@ int BleObject::Broadcaster::setAdvertisingData(const uint8_t* buf, size_t len) {
     // It is invalid to provide the same data buffers while advertising.
     CHECK(suspend());
     if (buf != nullptr) {
-        len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN);
+        len = std::min(len, (size_t)BLE_MAX_SCAN_REPORT_BUF_LEN);
         memcpy(advData_, buf, len);
     } else {
         len = 0;
--- a/hal/src/nRF52840/ble_hal_impl.h
+++ b/hal/src/nRF52840/ble_hal_impl.h
@@ -81,6 +81,8 @@
 /* Maximum length of the buffer to store scan report data */
 #define BLE_MAX_SCAN_REPORT_BUF_LEN                 BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
 
+#define BLE_MAX_ADV_DATA_SIZE_EXTENDED_CONNECTABLE  BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
+
 /* Connection Parameters limits */
 #define BLE_CONN_PARAMS_SLAVE_LATENCY_ERR           5
 #define BLE_CONN_PARAMS_TIMEOUT_ERR                 100
--- a/hal/src/nRF52840/ble_hal_impl.h
+++ b/hal/src/nRF52840/ble_hal_impl.h
@@ -79,9 +79,9 @@
 #define BLE_MAX_ADV_DATA_LEN                        BLE_GAP_ADV_SET_DATA_SIZE_MAX
 
 /* Maximum length of the buffer to store scan report data */
-#define BLE_MAX_SCAN_REPORT_BUF_LEN                 BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
+#define BLE_MAX_ADV_DATA_LEN_EXT                    BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
 
-#define BLE_MAX_ADV_DATA_SIZE_EXTENDED_CONNECTABLE  BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
+#define BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE        BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
 
 /* Connection Parameters limits */
 #define BLE_CONN_PARAMS_SLAVE_LATENCY_ERR           5
--- a/hal/src/nRF52840/ble_hal_impl.h
+++ b/hal/src/nRF52840/ble_hal_impl.h
@@ -79,8 +79,10 @@
 #define BLE_MAX_ADV_DATA_LEN                        BLE_GAP_ADV_SET_DATA_SIZE_MAX
 
 /* Maximum length of the buffer to store scan report data */
-#define BLE_MAX_ADV_DATA_LEN_EXT                    BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
+#define BLE_MAX_SCAN_REPORT_BUF_LEN                 BLE_GAP_ADV_SET_DATA_SIZE_MAX
 
+/* Maximum length for CODED_PHY */
+#define BLE_MAX_ADV_DATA_LEN_EXT                    BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED
 #define BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE        BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
 
 /* Connection Parameters limits */
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -564,8 +564,7 @@ constexpr uint8_t BleUuid::UUID16_HI;
  * BleAdvertisingData class
  */
 BleAdvertisingData::BleAdvertisingData()
-        : selfData_(),
-          selfLen_(0) {
+        : selfData_() {
     uint8_t flag = BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE;
     append(BleAdvertisingDataType::FLAGS, &flag, sizeof(uint8_t));
 }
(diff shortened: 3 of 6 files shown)


wiring/src/spark_wiring_ble.cpp
```diff
@@ -2177,8 +2168,11 @@ ssize_t BleLocalDevice::getAdvertisingData(BleAdvertisingData* advertisingData)
     if (advertisingData == nullptr) {
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
+    hal_ble_adv_params_t advParams = {};
+    hal_ble_gap_get_advertising_parameters(&advParams, nullptr);
     advertisingData->clear();
-    size_t len = CHECK(hal_ble_gap_get_advertising_data(advertisingData->data(), BLE_MAX_ADV_DATA_LEN, nullptr));
+    advertisingData->resize((advParams.primary_phy == BLE_PHYS_CODED) ? BLE_MAX_SCAN_REPORT_BUF_LEN : BLE_MAX_ADV_DATA_LEN);
```
avtolstoy, line 2174: This can fail to allocate and cause `hal_ble_gap_get_advertising_data()` to write out of bounds.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2192,6 +2186,7 @@ ssize_t BleLocalDevice::getScanResponseData(BleAdvertisingData* scanResponse) co
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
     scanResponse->clear();
+    scanResponse->resize(BLE_MAX_ADV_DATA_LEN);
```
avtolstoy, line 2189: Can potentially fail, `hal_ble_gap_get_scan_response_data` may write out of bounds.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2177,8 +2168,11 @@ ssize_t BleLocalDevice::getAdvertisingData(BleAdvertisingData* advertisingData)
     if (advertisingData == nullptr) {
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
+    hal_ble_adv_params_t advParams = {};
```
avtolstoy, line 2171:
```c++
advParams.size = sizeof(hal_ble_adv_params_t);
```

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,76 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
+            // Firstly, remove the existing AD structure.
+            selfData_.removeAt(offset, adsLen);
             // Secondly, Update the AD structure.
             // The Length field is the total length of Type field and Data field.
-            selfData_[offset] = len + 1;
-            memcpy(&selfData_[offset + 2], buf, len);
+            selfData_.insert(offset, len + 1);
             // An AD structure is composed of one byte length field, one byte Type field and Data field.
-            selfLen_ = staLen + len + 2;
+            selfData_.insert(offset+1, static_cast<uint8_t>(type));
```
avtolstoy, line 638: Nit: spaces :)

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,76 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-        return selfLen_;
+        return selfData_.size();
     }
 
-    selfData_[selfLen_++] = 0x02;
-    selfData_[selfLen_++] = BLE_SIG_AD_TYPE_FLAGS;
-    selfData_[selfLen_++] = BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE;
-    selfData_[selfLen_++] = 0x1a; // length
-    selfData_[selfLen_++] = BLE_SIG_AD_TYPE_MANUFACTURER_SPECIFIC_DATA;
+    selfData_.append(0x02);
```
avtolstoy, line 595: Any append in theory can fail. Not sure how exactly we want to deal with allocation failures here. Reserving `BLE_MAX_ADV_DATA_LEN` might be a good idea just in case as well. cc @XuGuohui

avtolstoy (changes requested):

wiring/src/spark_wiring_ble.cpp
mgolu: Added a `reserve` for `BLE_MAX_ADV_DATA_LEN` before all the appends, returning 0 if it fails (since the `set` function doesn't return negative). Also in `advertise(iBeacon)` added a check that the length of the advert is larger than 0.

wiring/src/spark_wiring_ble.cpp
mgolu: Added check for resize failure.

wiring/src/spark_wiring_ble.cpp
mgolu: Added check for resize failure

wiring/src/spark_wiring_ble.cpp
XuGuohui: You just added a reserve here only. It might fail to append wherever `append()` is used. Adding a reserve is one approach, but it still waste ram. Here are two candidates I can put up:

1. Reserve `BLE_MAX_SCAN_REPORT_BUF_LEN` ram before appending and simply return if it fails. If succeeds, use `trimToSize()` to free unused ram.
2. New a temp vector and fill it with the new data. If it fails at certain append, we abort the data, otherwise, we finally append the temp vector to the `selfData_` vector.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,77 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-            uint16_t moveLen = selfLen_ - offset - adsLen;
-            memmove(&selfData_[offset + len + 2], &selfData_[offset + adsLen], moveLen);
+        if ((selfData_.size() - adsLen + len + 2) <= BLE_MAX_SCAN_REPORT_BUF_LEN) {
+            // Firstly, remove the existing AD structure.
+            selfData_.removeAt(offset, adsLen);
             // Secondly, Update the AD structure.
             // The Length field is the total length of Type field and Data field.
-            selfData_[offset] = len + 1;
-            memcpy(&selfData_[offset + 2], buf, len);
+            selfData_.insert(offset, len + 1);
```
XuGuohui, line 637: It may fail as well. Same to wherever the `insert()` is used.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -327,7 +327,7 @@ class BleObject::Broadcaster {
     volatile hal_ble_auto_adv_cfg_t autoAdvCfg_;    /**< Automatic advertising configuration. */
     uint8_t advHandle_;                             /**< Advertising handle. */
     hal_ble_adv_params_t advParams_;                /**< Current advertising parameters. */
-    uint8_t advData_[BLE_MAX_ADV_DATA_LEN];         /**< Current advertising data. */
+    uint8_t advData_[BLE_MAX_SCAN_REPORT_BUF_LEN];  /**< Current advertising data. */
```
XuGuohui, line 330: I'd suggest to introduce dedicate macros for adv data, e.g. 
```
#define BLE_MAX_ADV_DATA_LEN_EXT BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_MAX_SUPPORTED
#define BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
```
and use `BLE_MAX_ADV_DATA_LEN_EXT` for general adv data buffer, while use `BLE_MAX_SCAN_REPORT_BUF_LEN` for scan report buffer.

wiring/src/spark_wiring_ble.cpp
XuGuohui: And initializing the `.version` field would be even better.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2192,7 +2188,8 @@ ssize_t BleLocalDevice::getScanResponseData(BleAdvertisingData* scanResponse) co
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
     scanResponse->clear();
-    size_t len = CHECK(hal_ble_gap_get_scan_response_data(scanResponse->data(), BLE_MAX_ADV_DATA_LEN, nullptr));
+    CHECK_TRUE(scanResponse->resize(BLE_MAX_ADV_DATA_LEN), SYSTEM_ERROR_NO_MEMORY);
```
XuGuohui, line 2191: Not sure but we should probably need to resize according to the PHY as what we do to `getAdvertisingData()`?

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2436,8 +2434,8 @@ class BleScanDelegator {
         size_t filterCustomDatalen;
         const uint8_t* filterCustomData = filter_.customData(&filterCustomDatalen);
         if (filterCustomData != nullptr && filterCustomDatalen > 0) {
-            size_t srLen = result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, sizeof(size_t));
-            size_t advLen = result.advertisingData().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, sizeof(size_t));
+            size_t srLen = result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, filterCustomDatalen);
```
XuGuohui, line 2437: We should not try to get length of the custom data from the actual adv data. Think of this scenario:

- The custom data we're going to filter is `01 02 03 04`
- The actual custom data is the adv data is `01 02 03 04 aa bb cc`

Can we say that they are identical if we only compare the first `filterCustomDatalen` of data?

XuGuohui (changes requested):

wiring/src/spark_wiring_ble.cpp
mgolu: Good point! I will set the length maximum to the maximum for each one.

wiring/src/spark_wiring_ble.cpp
mgolu: Scan response is only available when using 1MBPS PHY, so I think that keeping the maximum resize to `BLE_MAX_ADV_DATA_LEN` which is 31 bytes is ok.

hal/src/nRF52840/ble_hal.cpp
mgolu: 👍  Added

wiring/src/spark_wiring_ble.cpp
mgolu: Reserved the right amount of RAM (and return if fail) prior to the `insert()` calls

wiring/src/spark_wiring_ble.cpp
mgolu: Added `trimToSize()` after done appending.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -389,7 +389,7 @@ class BleObject::Observer {
     volatile bool isScanning_;                              /**< If it is scanning or not. */
     hal_ble_scan_params_t scanParams_;                      /**< BLE scanning parameters. */
     os_semaphore_t scanSemaphore_;                          /**< Semaphore to wait until the scan procedure completed. */
-    uint8_t scanReportBuff_[BLE_MAX_SCAN_REPORT_BUF_LEN];   /**< Buffer to hold the scanned report data. */
+    uint8_t scanReportBuff_[BLE_MAX_ADV_DATA_LEN_EXT];   /**< Buffer to hold the scanned report data. */
```
XuGuohui, line 392: We should keep `BLE_MAX_SCAN_REPORT_BUF_LEN` as-is and use it for scan buffers.

hal/src/nRF52840/ble_hal_impl.h
```diff
@@ -79,7 +79,9 @@
 #define BLE_MAX_ADV_DATA_LEN                        BLE_GAP_ADV_SET_DATA_SIZE_MAX
 
 /* Maximum length of the buffer to store scan report data */
-#define BLE_MAX_SCAN_REPORT_BUF_LEN                 BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
+#define BLE_MAX_ADV_DATA_LEN_EXT                    BLE_GAP_SCAN_BUFFER_EXTENDED_MAX_SUPPORTED  /* Must support extended length for CODED_PHY scanning */
```
XuGuohui, line 82: We should keep `BLE_MAX_SCAN_REPORT_BUF_LEN` as-is and use it for scan buffers. And add:
```
#define BLE_MAX_ADV_DATA_LEN_EXT BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_MAX_SUPPORTED
#define BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE BLE_GAP_ADV_SET_DATA_SIZE_EXTENDED_CONNECTABLE_MAX_SUPPORTED
```
Let's use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` for advertising data buffers for now and leave `BLE_MAX_ADV_DATA_LEN_EXT` alone.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1087,7 +1087,7 @@ int BleObject::Broadcaster::setAdvertisingData(const uint8_t* buf, size_t len) {
     // It is invalid to provide the same data buffers while advertising.
     CHECK(suspend());
     if (buf != nullptr) {
-        len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN);
+        len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN_EXT);
```
XuGuohui, line 1090: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` for the advertising data buffer.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -327,7 +327,7 @@ class BleObject::Broadcaster {
     volatile hal_ble_auto_adv_cfg_t autoAdvCfg_;    /**< Automatic advertising configuration. */
     uint8_t advHandle_;                             /**< Advertising handle. */
     hal_ble_adv_params_t advParams_;                /**< Current advertising parameters. */
-    uint8_t advData_[BLE_MAX_ADV_DATA_LEN];         /**< Current advertising data. */
+    uint8_t advData_[BLE_MAX_ADV_DATA_LEN_EXT];     /**< Current advertising data. */
```
XuGuohui, line 330: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` for the advertising data buffer.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,81 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-        return selfLen_;
+        selfData_.clear();
+        return selfData_.size();
     }
-    len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN);
-    memcpy(selfData_, buf, len);
-    selfLen_ = len;
-    return selfLen_;
+    selfData_.clear();
+    len = std::min(len, (size_t)BLE_MAX_ADV_DATA_LEN_EXT);
```
XuGuohui, line 583: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,81 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
+    selfData_.append(buf, len);
```
XuGuohui, line 584: `append()` may fail.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,81 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-        return selfLen_;
+        return selfData_.size();
     }
 
-    selfData_[selfLen_++] = 0x02;
-    selfData_[selfLen_++] = BLE_SIG_AD_TYPE_FLAGS;
-    selfData_[selfLen_++] = BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE;
-    selfData_[selfLen_++] = 0x1a; // length
-    selfData_[selfLen_++] = BLE_SIG_AD_TYPE_MANUFACTURER_SPECIFIC_DATA;
+    CHECK_TRUE(selfData_.reserve(BLE_MAX_ADV_DATA_LEN), 0);
```
XuGuohui, line 595: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,81 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-    size_t adsLen = locate(selfData_, selfLen_, type, &offset);
+    size_t adsLen = locate(selfData_.data(), selfData_.size(), type, &offset);
     if (!force && adsLen > 0) {
         // Update the existing AD structure.
-        size_t staLen = selfLen_ - adsLen;
-        if ((staLen + len + 2) <= BLE_MAX_ADV_DATA_LEN) {
-            // Firstly, move the last consistent block.
-            uint16_t moveLen = selfLen_ - offset - adsLen;
-            memmove(&selfData_[offset + len + 2], &selfData_[offset + adsLen], moveLen);
+        if ((selfData_.size() - adsLen + len + 2) <= BLE_MAX_ADV_DATA_LEN_EXT) {
```
XuGuohui, line 633: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -577,81 +576,81 @@ BleAdvertisingData::BleAdvertisingData(const iBeacon& beacon)
...
-            memcpy(&selfData_[offset + 2], buf, len);
+            selfData_.insert(offset, len + 1);
             // An AD structure is composed of one byte length field, one byte Type field and Data field.
-            selfLen_ = staLen + len + 2;
+            selfData_.insert(offset + 1, static_cast<uint8_t>(type));
+            selfData_.insert(offset + 2, buf, len);
         }
     }
-    else if ((selfLen_ + len + 2) <= BLE_MAX_ADV_DATA_LEN) {
+    else if ((selfData_.size() + len + 2) <= BLE_MAX_ADV_DATA_LEN_EXT) {
```
XuGuohui, line 646: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -671,39 +670,36 @@ size_t BleAdvertisingData::appendAppearance(ble_sig_appearance_t appearance) {
 }
 
 size_t BleAdvertisingData::resize(size_t size) {
-    selfLen_ = std::min(size, (size_t)BLE_MAX_ADV_DATA_LEN);
-    return selfLen_;
+    selfData_.resize(std::min(size, (size_t)BLE_MAX_ADV_DATA_LEN_EXT));
```
XuGuohui, line 673: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2177,8 +2173,12 @@ ssize_t BleLocalDevice::getAdvertisingData(BleAdvertisingData* advertisingData)
     if (advertisingData == nullptr) {
         return SYSTEM_ERROR_INVALID_ARGUMENT;
     }
+    hal_ble_adv_params_t advParams = {};
+    advParams.size = sizeof(hal_ble_adv_params_t);
+    hal_ble_gap_get_advertising_parameters(&advParams, nullptr);
     advertisingData->clear();
-    size_t len = CHECK(hal_ble_gap_get_advertising_data(advertisingData->data(), BLE_MAX_ADV_DATA_LEN, nullptr));
+    CHECK_TRUE(advertisingData->resize((advParams.primary_phy == BLE_PHYS_CODED) ? BLE_MAX_ADV_DATA_LEN_EXT : BLE_MAX_ADV_DATA_LEN), SYSTEM_ERROR_NO_MEMORY);
```
XuGuohui, line 2180: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2436,8 +2438,8 @@ class BleScanDelegator {
         size_t filterCustomDatalen;
         const uint8_t* filterCustomData = filter_.customData(&filterCustomDatalen);
         if (filterCustomData != nullptr && filterCustomDatalen > 0) {
-            size_t srLen = result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, sizeof(size_t));
-            size_t advLen = result.advertisingData().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, sizeof(size_t));
+            size_t srLen = result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, BLE_MAX_ADV_DATA_LEN);
+            size_t advLen = result.advertisingData().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, nullptr, BLE_MAX_ADV_DATA_LEN_EXT);
```
XuGuohui, line 2442: We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` here.

XuGuohui (changes requested):

hal/src/nRF52840/ble_hal.cpp
mgolu: > We should use `BLE_MAX_ADV_DATA_LEN_EXT_CONNECTABLE` for the advertising data buffer.

What if a customer wants to make an application that advertises the full 255 bytes in non-connectable mode?

hal/src/nRF52840/ble_hal_impl.h
mgolu: 👍

hal/src/nRF52840/ble_hal.cpp
mgolu: 👍

hal/src/nRF52840/ble_hal.cpp
mgolu: Same issue as above, I think this is an issue if customer application is non-connectable.

wiring/src/spark_wiring_ble.cpp
mgolu: Added check

wiring/src/spark_wiring_ble.cpp
mgolu: By specification, iBeacon data is 30 bytes in length.

wiring/src/spark_wiring_ble.cpp
mgolu: Application could be non-connectable.

wiring/src/spark_wiring_ble.cpp
mgolu: See above.

wiring/src/spark_wiring_ble.cpp
mgolu: See above (connectable application)

wiring/src/spark_wiring_ble.cpp
mgolu: See above (non-connectable application)

wiring/src/spark_wiring_ble.cpp
mgolu: Peripheral could use all 255 bytes if non-connectable

wiring/src/spark_wiring_ble.cpp
mgolu: See above (non-connectable application).

XuGuohui: Yeah. you're right 👍 . Let's use `BLE_MAX_ADV_DATA_LEN_EXT` as the maximum length of adv data buffers.

XuGuohui (approved):

PR #2332, opened 2021-06-25
Free malloc'ed memory

mgolu:
### Problem

Seems like a bug that the memory is allocated with malloc but never freed. I didn't run a test to check if there was an actual memory leak, so I might have missed how it's freed. I did test that this new code compiles and the function works as expected (re-ran the `ble_scanner_broadcaster` tests)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -2450,8 +2450,10 @@ private:
                 }
                 result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, buf, srLen);
                 if (!memcmp(buf, filterCustomData, srLen)) {
+                    free(buf);
                     return true;
                 }
+                free(buf);
             }
             if (advLen == filterCustomDatalen) {
                 uint8_t* buf = (uint8_t*)malloc(advLen);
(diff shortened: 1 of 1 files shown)


wiring/src/spark_wiring_ble.cpp
```diff
@@ -2450,8 +2450,10 @@ class BleScanDelegator {
                 }
                 result.scanResponse().get(BleAdvertisingDataType::MANUFACTURER_SPECIFIC_DATA, buf, srLen);
                 if (!memcmp(buf, filterCustomData, srLen)) {
+                    free(buf);
```
avtolstoy, line 2453: :sweat: Ooof. Thanks, we overlooked this.

Better to use 
```c++
uint8_t* buf = (uint8_t*)malloc(srLen);
SCOPE_GUARD({
    if (buf) {
        free(buf);
    }
});
```

or even

```c++
std::unique_ptr<uint8_t[]> buf(srLen);
// ...
```

avtolstoy (approved):

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2444,6 +2444,11 @@ class BleScanDelegator {
             }
             if (srLen == filterCustomDatalen) {
                 uint8_t* buf = (uint8_t*)malloc(srLen);
+                SCOPE_GUARD({
```
XuGuohui, line 2447: 💯

XuGuohui (approved):

PR #2334, opened 2021-07-01
[u-blox] Handle erroneous -1 AcT response from CEREG and COPS commands on R510

scott-brust:
### Problem

The SARAR510 modem can potentially report a -1 value in the Radio Access Technology (RAT or AcT) field of at least two AT commands: `AT+COPS?` and `AT+CEREG?`
This value results in an incorrectly parsed cellular access technology that breaks the signal strength reporting in particle console.

### Solution

Ublox confirmed this is a bug and will be fixed in later modem firmware versions, but a workaround should still detect this problem. 
If a -1 value is parsed from the `CEREG` URC or `COPS` command result, it will be reset to the fixed R510 value of 7. This should ensure the correct cellular access technology value. 

### Steps to Test
You can manually introduce the problem by substituting the COPS and CEREG strings with the -1 values. IE:
```
+CEREG: 2,5,"6a41","07f06011",-1
+COPS: 0,0,"AT&T",-1
```

### Example App

Problem is not introducible via user app, its part of device OS cellular HAL

### References

See [clubhouse ticket](https://app.clubhouse.io/particle/story/79160/r510-rat-value-switches-to-an-invalid-value-1-a-few-minutes-after-registration)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -317,6 +317,10 @@ int SaraNcpClient::initParser(Stream* stream) {
         // self->checkRegistrationState();
         // Cellular Global Identity (partial)
         if (r >= 3) {
+            // Check for erroneous R510 AcT value of -1, change to R510 specific Cat-M1 value of 7
+            if (r >= 4 && val[3] == 0xFFFFFFFF) {
+                val[3] = 7; 
+            }
             auto rat = r >= 4 ? static_cast<CellularAccessTechnology>(val[3]) : self->act_;
             switch (rat) {
                 case CellularAccessTechnology::LTE:
@@ -642,7 +646,8 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
     cgi_.mobile_network_code = static_cast<uint16_t>(::atoi(mobileNetworkCode));
 
     if (ncpId() == PLATFORM_NCP_SARA_R410 || ncpId() == PLATFORM_NCP_SARA_R510) {
-        if (act == particle::to_underlying(CellularAccessTechnology::LTE)) {
+        if (act == particle::to_underlying(CellularAccessTechnology::LTE) || 
+            (act == -1 && ncpId() == PLATFORM_NCP_SARA_R510)) {
             act = particle::to_underlying(CellularAccessTechnology::LTE_CAT_M1);
         }
     }
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -307,6 +307,12 @@ bool MDMParser::modemIsSaraRxFamily() {
     return ((_dev.dev == DEV_SARA_R410) || (_dev.dev == DEV_SARA_R510));
 }
 
+void MDMParser::_fixSaraR510AccessTechnology(void) {
+    if (_dev.dev == DEV_SARA_R510 && _net.act == ACT_UNKNOWN) {
+        _net.act = ACT_LTE_CAT_M1;
+    }
+}
+
 void MDMParser::cancel(void) {
     if (!_cancel_all_operations) {
         MDM_INFO("\r\n[ Modem::cancel ] = = = = = = = = = = = = = = =");
@@ -598,6 +604,7 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                 } else {
                     // +CREG|CGREG: <n>,<stat>[,<lac>,<ci>[,AcT[,<rac>]]] // reply to AT+CREG|AT+CGREG (2,4,5,6 results)
                     // +CREG|CGREG: <stat>[,<lac>,<ci>[,AcT[,<rac>]]]     // URC (1,3,4,5 results)
+                    // +CEREG: <n>,<stat>[,[<tac>],[<ci>],[<AcT>]]
                     b = (int)0xFFFF; c = (int)0xFFFFFFFF; d = -1; mode = -1; // default mode to -1 for safety
                     r = sscanf(cmd, "%31s %d,%d,\"%x\",\"%x\",%d",s,&mode,&a,&b,&c,&d);
                     if (r <= 2) {
@@ -658,6 +665,7 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                             // access technology
                             if (r >= 5) {
                                 _net.act = toCellularAccessTechnology(d);
+                                _fixSaraR510AccessTechnology();
                             }
                         }
                     }
@@ -1913,6 +1921,8 @@ bool MDMParser::registerNet(const char* apn, NetStatus* status, system_tick_t ti
             if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
                 goto failure;
             }
+            _fixSaraR510AccessTechnology();
+
             // Only run AT+COPS=0 if currently de-registered, to avoid PLMN reselection
             if (_net.cops != 0 && _net.cops != 1) {
                 sendFormated("AT+COPS=0,2\r\n");
@@ -2031,6 +2041,7 @@ bool MDMParser::checkNetStatus(NetStatus* status /*= NULL*/)
         if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
             goto failure;
         }
+        _fixSaraR510AccessTechnology();
         // AT command used to collect signal stregnth is different for R410M radio
         if (_dev.dev == DEV_SARA_R410) {
             sendFormated("AT+UCGED=5\r\n");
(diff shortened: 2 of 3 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -642,9 +645,7 @@ int SaraNcpClient::queryAndParseAtCops(CellularSignalQuality* qual) {
     cgi_.mobile_network_code = static_cast<uint16_t>(::atoi(mobileNetworkCode));
 
     if (ncpId() == PLATFORM_NCP_SARA_R410 || ncpId() == PLATFORM_NCP_SARA_R510) {
-        if (act == particle::to_underlying(CellularAccessTechnology::LTE)) {
-            act = particle::to_underlying(CellularAccessTechnology::LTE_CAT_M1);
-        }
+        act = particle::to_underlying(CellularAccessTechnology::LTE_CAT_M1);
```
scott-brust, line 648: Seems like the easiest way to avoid this problem is to ignore the AcT parameter for R410/510 and always fix it to LTE_CAT_M1

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -2402,6 +2406,9 @@ int MDMParser::_cbCOPS(int type, const char* buf, int len, NetStatus* status)
             status->cgi.mobile_country_code = static_cast<uint16_t>(::atoi(mobileCountryCode));
             status->cgi.mobile_network_code = static_cast<uint16_t>(::atoi(mobileNetworkCode));
 
+            if( act == -1 ) {
+                act = UBLOX_SARA_RAT_LTE_CAT_M1;
```
scott-brust, line 2410: Its not clear to me if non SARA R modules use these static callbacks. I'm assuming not in this case, or at least that if we see a -1 value then we can assume its a R510 and should be changed

scott-brust:
Also, coincidentally `-1` maps to `UBLOX_SARA_RAT_NONE` in the ublox cell tech enum. Should we handle that rat in the `toCellularAccessTechnology` function in `mdm_hal.cpp`. Perhaps add it to the default case in [this line](https://github.com/particle-iot/device-os/blob/3ca99dfcfbf15b9af92cece665083c7a69a4e6b3/hal/src/electron/modem/mdm_hal.cpp#L168)

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -317,6 +317,9 @@ int SaraNcpClient::initParser(Stream* stream) {
         // self->checkRegistrationState();
         // Cellular Global Identity (partial)
         if (r >= 3) {
+            if(r >= 4 && val[3] == 0xFFFFFFFF) {
```
scott-brust, line 320: since `val` is unsigned, the sscanf of -1 will result in `0xFFFFFFFF` Should we change `val` to signed int to make the comparison to -1 more obvious? Or leave as is?

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: This somewhat hard codes the RAT for R410 as well which doesn't have the -1 bug.  The bug also will be resolved in a future update from u-blox for the R510.  If we hard code it this way, would it show that we are always connected to Cat-M1, even when we might not have signal?  For the previous R410 code, it checked if the AcT was LTE first, before setting to Cat-M1.

I believe the fix in the parser should be good enough to resolve this for R510
```
if (modemIsSaraRxFamily() && d == -1 ) {
    d = UBLOX_SARA_RAT_LTE_CAT_M1;
}
```

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -657,7 +658,10 @@ int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                             if ((r >= 4) && (c != (int)0xFFFFFFFF))  _net.cgi.cell_id  = c;
                             // access technology
                             if (r >= 5) {
-                                _net.act = toCellularAccessTechnology(d);
+                                if(modemIsSaraRxFamily() && d == -1 ) {
```
technobly, line 661: This is probably ok to include R410, just in case?  At a minimum though we should have an R510 check here.

NIT: spacing for `if (`

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -317,6 +317,9 @@ int SaraNcpClient::initParser(Stream* stream) {
...
+                val[3] = 7; // Force SARA R510 fixed value
```
tstellanova, line 321: Is there a handy meaningful constant we can use instead of `7` ?

hal/src/electron/modem/mdm_hal.cpp
technobly: We should probably add a check here for R510 and/or R410 like above, since this callback is used for all u-blox modem types we support.

NIT: spacing for `if (`

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Hmm, good question.  I think the other values that val[] is capturing would potentially be messier if val[] was changed to an int.  So possibly add a comment that we are checking for `RAT == -1` on that line.

technobly (changes requested):

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Follow up... just caught that this was Gen 3 vs. Gen 2 code I referenced.  Now I'm also wondering if we have this R410 "fix" in Gen 2.

So I'm thinking we should not change this code here for Gen 3, but instead test for -1 and hard code to Cat-M1 for R510 only.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Unfortunately this AcT for CEREG doesn't line up with the Enums we have for AcT/RaT for u-blox.  We have enums that work with COPS, but they are slightly different for C*REG.  Could update this comment to `Cat-M1` instead of `fixed value`.

hal/src/electron/modem/mdm_hal.cpp
scott-brust: I had the same instinct, but it appears the COPS handler is a static function that is invoked from `waitFinalResp` like so
```
sendFormated("AT+COPS?\r\n");
if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
```
```
int MDMParser::waitFinalResp(_CALLBACKPTR cb /* = NULL*/,
                             void* param /* = NULL*/,
                             system_tick_t timeout_ms /*= 10000*/)
{
...
            if (cb) {
                int len = LENGTH(ret);
                int ret = cb(type, buf, len, param);
                if (WAIT != ret)
                    return ret;
            }
...
}
```
`modemIsSaraRxFamily()` relies on the `_dev.dev` class member to check for modem type. I'm not sure how to do this device check from a static context just yet

hal/network/ncp_client/sara/sara_ncp_client.cpp
scott-brust: Changed to explicitly check for act == -1 and R510 modem

hal/src/electron/modem/mdm_hal.cpp
scott-brust: I thought it was cleaner to just use `modemIsSaraRxFamily()` but I can change it to `(_dev.dev == DEV_SARA_R510)` if you would like

hal/network/ncp_client/sara/sara_ncp_client.cpp
scott-brust: Made comment hopefully clearer

hal/network/ncp_client/sara/sara_ncp_client.cpp
scott-brust: Updated my comment

hal/src/electron/modem/mdm_hal.cpp
technobly: Oh, yeah that is a problem... so perhaps just post process it after these calls:
```
sendFormated("AT+COPS?\r\n");
if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
    goto failure;
}
```

scott-brust: I can do something like this after each `AT+COPS?` query:
```
            sendFormated("AT+COPS?\r\n");
            if (RESP_OK != waitFinalResp(_cbCOPS, &_net, COPS_TIMEOUT)) {
                goto failure;
            }
            if (_net.act == ACT_UNKNOWN && modemIsSaraRxFamily()){
                _net.act = ACT_LTE_CAT_M1;
            }
```
There are 4 instances in `mdm_hal.cpp` where `AT+COPS?` is sent, should I do this for each? or make another helper function to check and modify `_net.act` after the command is sent from the non-static context?

technobly: Another helper function sounds appropriate.  I think we should standardize on only making this change for R510 though.  So far we have a few places where the change is made for R510 only and some for R510 & R410.

hal/src/electron/modem/mdm_hal.cpp
technobly: Yeah since this is a R510 bug, we should probably not alter R410 behavior.

technobly (approved):

PR #2339, opened 2021-07-15
Use ncp_id from OTP to determine modem type

scott-brust:
### Problem

Early boards with the SARA R510 did not have the Network Coprocessor Identifier programed into the OTP memory locations. To work around this SARA R510 was hard coded in the GEN2 modem hal and the GEN3 `platform_primary_ncp_identifier`. 

The latest gen 2 / 3 SARAR510 EVT devices have this data so the workaround should not be needed going forward
 
### Solution

For gen 3, the `platform_primary_ncp_identifier` function already reads the NCP id OTP data.
For gen 2, there is no `platform_ncp_electron` file, but rather it appears the other OTP data (device serial number, secret, etc) is accessible via `deviceid_hal`. I opted to extend the existing gen 2 deviceid hal, but there may be reasons we dont want to do that.  

### Steps to Test

Run example app, some hardware may not have R510 OTP programmed. This case still needs to be detected and handled

### Example App

```c
void setup() {
}

void loop() {
    delay(1000);

    CellularDevice device = {};
    device.size = sizeof(device);
    cellular_device_info(&device, NULL);
    Log.info("Cellular device info: %X IMEI: %s ICCID: %s", device.dev, device.imei, device.iccid);

    #if PLATFORM_GEN == 2
        uint8_t ncpId = 0;
        const size_t OTP_NCP_ID_ADDRESS = 32;
        int err_code = FLASH_ReadOTP(OTP_NCP_ID_ADDRESS, (uint8_t*)&ncpId, sizeof(ncpId));
        if (err_code) {
            Log.info("ERROR: %d", err_code);
        }
        Log.info("NCP ID: 0x%x", ncpId);
    #endif
}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/electron/modem/mdm_hal.cpp
+++ b/hal/src/electron/modem/mdm_hal.cpp
@@ -41,6 +41,8 @@
 #include <mutex>
 #include "net_hal.h"
 #include <limits>
+#include "platform_ncp.h"
+#include "cellular_ncp_dev_mapping.h"
 
 std::recursive_mutex mdm_mutex;
 
@@ -882,9 +884,9 @@ bool MDMParser::_powerOn(void)
     bool retried_after_reset = false;
     int i = MDM_POWER_ON_MAX_ATTEMPTS_BEFORE_RESET; // When modem not responsive on boot, AT/OK tries 25x (for ~30s) before hard reset
 
-    // FIXME: REMOVE THIS BEFORE MERGING!!!!!!
-    _dev.dev = DEV_SARA_R510;
-    MDM_ERROR("Forcing to R510\r\n");
+    auto otp_ncp_id = platform_primary_ncp_identifier();
+    _dev.dev = cellular_dev_from_ncp(otp_ncp_id);
+    LOG(INFO, "NCP ID OTP value: 0x%X _dev.dev set to 0x%X", otp_ncp_id, _dev.dev);
 
     while (i--) {
 
--- a/hal/src/electron/platform_ncp_electron.cpp
+++ b/hal/src/electron/platform_ncp_electron.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Particle Industries, Inc.  All rights reserved.
+ * Copyright (c) 2021 Particle Industries, Inc.  All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -16,8 +16,7 @@
  */
 
 #include "platform_ncp.h"
-#include "exflash_hal.h"
-#include "dct.h"
+#include "flash_mal.h"
 #include "system_error.h"
 
 namespace {
@@ -29,6 +28,8 @@ bool isValidNcpId(uint8_t id) {
     case PlatformNCPIdentifier::PLATFORM_NCP_SARA_U201:
     case PlatformNCPIdentifier::PLATFORM_NCP_SARA_G350:
     case PlatformNCPIdentifier::PLATFORM_NCP_SARA_R410:
+    case PlatformNCPIdentifier::PLATFORM_NCP_SARA_U260:
+    case PlatformNCPIdentifier::PLATFORM_NCP_SARA_U270:
     case PlatformNCPIdentifier::PLATFORM_NCP_SARA_R510:
         return true;
     default:
@@ -39,17 +40,15 @@ bool isValidNcpId(uint8_t id) {
 } // unnamed
 
 PlatformNCPIdentifier platform_primary_ncp_identifier() {
-    // Check the DCT
-    uint8_t ncpId = 0;
-    int r = dct_read_app_data_copy(DCT_NCP_ID_OFFSET, &ncpId, 1);
-    if (r < 0 || !isValidNcpId(ncpId)) {
-        // Check the OTP flash
-        r = hal_exflash_read_special(HAL_EXFLASH_SPECIAL_SECTOR_OTP, NCP_ID_OTP_ADDRESS, &ncpId, 1);
-        if (r < 0 || !isValidNcpId(ncpId)) {
-            ncpId = PlatformNCPIdentifier::PLATFORM_NCP_UNKNOWN;
-        }
+    uint8_t otp_ncp_id;
+    PlatformNCPIdentifier identifier = PlatformNCPIdentifier::PLATFORM_NCP_UNKNOWN;
+
+    int ret = FLASH_ReadOTP(NCP_ID_OTP_ADDRESS, (uint8_t*)&otp_ncp_id, sizeof(otp_ncp_id));
+    if (ret == 0 && isValidNcpId(otp_ncp_id)) {
+        identifier = (PlatformNCPIdentifier)otp_ncp_id;
     }
-    return (PlatformNCPIdentifier)ncpId;
+
+    return identifier;
 }
 
 int platform_ncp_get_info(int idx, PlatformNCPInfo* info) {
--- a/hal/src/photon/platform_ncp_photon.cpp
+++ b/hal/src/photon/platform_ncp_photon.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Particle Industries, Inc.  All rights reserved.
+ * Copyright (c) 2021 Particle Industries, Inc.  All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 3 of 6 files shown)


scott-brust:
This is a somewhat naive attempt, but I wanted to get feedback if this approach is how we want to go for gen 2 OTP support. I imagine there are dynalib implications to extended the deviceid hal. Perhaps implementing a `platform_ncp_electron.cpp` would be preferable. 

 There are still error cases that need to be handled as well as potentially removing other hardcoded R510 behavior.

hal/src/stm32f2xx/deviceid_hal.cpp
```diff
@@ -126,6 +127,20 @@ int hal_get_device_secret(char* data, size_t size, void* reserved)
 }
 #endif /* HAL_DEVICE_ID_NO_DCT */
 
+#ifndef HAL_DEVICE_ID_NO_DCT
+int hal_get_device_ncp_id(uint8_t * ncpid)
+{
+    uint8_t otp_ncp_id;
+    int ret = FLASH_ReadOTP(DEVICE_NCP_ID_OTP_ADDRESS, (uint8_t*)&otp_ncp_id, sizeof(otp_ncp_id));
```
avtolstoy, line 134: `FLASH_ReadOTP` may very well return garbage for devices with unpopulated OTP area, the output should be sanitized, hence the above proposal to instead implement `platform_ncp_primary_identifier()` which provides a concrete NCP identifier or `PLATFORM_NCP_UNKNOWN`.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -882,9 +884,26 @@ bool MDMParser::_powerOn(void)
...
 
-    // FIXME: REMOVE THIS BEFORE MERGING!!!!!!
-    _dev.dev = DEV_SARA_R510;
-    MDM_ERROR("Forcing to R510\r\n");
+    // TODO: 
+    // Implement `platform_ncp_electron` + `platform_primary_ncp_identifier` to determine NCP id? 
+    // -OR-
+    // Extend `deviceid_hal.h` ? 
+    uint8_t otp_ncp_id = 0;
+    int result = hal_get_device_ncp_id(&otp_ncp_id);
```
avtolstoy, line 892: Nit: might be a good idea to introduce a dedicated member method to get the modem type similar to `ncpId()` in cellular NCP classes to hide the logic of looking either at runtime `_dev.dev` or `platform_ncp_primary_indentifier()` to clean things up a bit.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -882,9 +884,26 @@ bool MDMParser::_powerOn(void)
...
+        case PLATFORM_NCP_SARA_R510:
+            _dev.dev = DEV_SARA_R510;
+            break;
+        default:
+            _dev.dev = DEV_UNKNOWN;
+            break;        
+        }
+    }
+
+    MDM_ERROR("_dev.dev SET AS = 0x%X otp value: 0x%X result code: %d\r\n", _dev.dev, otp_ncp_id, result);
```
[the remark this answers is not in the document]
scott-brust: All of `mdm_hal.cpp` uses the `MDM*` logging macros, I just tried to match that. I think `INFO` is ok to leave? I'm still figuring out how verbose we would like to be in logs

hal/src/stm32f2xx/deviceid_hal.cpp
```diff
@@ -126,6 +127,20 @@ int hal_get_device_secret(char* data, size_t size, void* reserved)
 }
 #endif /* HAL_DEVICE_ID_NO_DCT */
 
+#ifndef HAL_DEVICE_ID_NO_DCT
+int hal_get_device_ncp_id(uint8_t * ncpid)
```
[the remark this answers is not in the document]
scott-brust: reverted this

hal/src/stm32f2xx/deviceid_hal.cpp
scott-brust: moved this logic to `platform_ncp_electron.cpp`

hal/src/electron/platform_ncp_electron.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+} // unnamed
+
+PlatformNCPIdentifier platform_primary_ncp_identifier() {
+    uint8_t otp_ncp_id;
+    PlatformNCPIdentifier identifier = PlatformNCPIdentifier::PLATFORM_NCP_UNKNOWN;
+
+    int ret = FLASH_ReadOTP(NCP_ID_OTP_ADDRESS, (uint8_t*)&otp_ncp_id, sizeof(otp_ncp_id));
```
scott-brust, line 46: `platform_ncp_boron` attempts to read NCP info from "DCT", I'm not sure if gen2 should operate in the same way? Or just generally what DCT data is there

modules/electron/system-part1/src/cellular_hal.cpp
```diff
@@ -8,3 +8,4 @@
 #include "../src/electron/inet_hal_new.cpp"
 #include "../src/electron/socket_hal.cpp"
 #include "../shared/cellular_sig_perc_mapping.cpp"
+#include "../src/electron/platform_ncp_electron.cpp"
```
scott-brust, line 11: I still dont fully understand the appropriate way to work with dynalib, particularly with Gen2 devices, but this is how I was able to get my builds to link successfully. If this is not the correct way to do things let me know

hal/src/electron/platform_ncp_electron.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+int platform_ncp_get_info(int idx, PlatformNCPInfo* info) {
+    if (idx == 0 && info) {
+        info->identifier = platform_primary_ncp_identifier();
+        info->updatable = false;
+        return 0;
+    }
+
+    return SYSTEM_ERROR_INVALID_ARGUMENT;
+}
```
scott-brust, line 62: Are there other functions in `platform_ncp.h` that I should implement for electron?

hal/src/electron/platform_ncp_electron.cpp
```diff
@@ -0,0 +1,62 @@
+/*
```
scott-brust, line 1: Should I create a similar file for Photon just to be consistent?

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -173,6 +174,40 @@ AcT toCellularAccessTechnology(int rat) {
 // see 3GPP TS 45.008 [20] subclause 8.2.4
 const char compatQualMap[] = { 49, 43, 37, 25, 19, 13, 7, 0 };
 
+Dev fromPlatformNcpIdentifer(PlatformNCPIdentifier identifer) {
```
scott-brust, line 177: This kind of duplicates some of the code from `cellular_device_info()` in `/hal/network/ncp/cellular/cellular_hal.cpp` but I struggled to find an appropriate place to make a helper function to map `PlatformNCPIdentifier` to `Dev` enum values. I settled on this approach since it seems `mdm_hal` looks to be the only file that uses the `Dev` enum extensively.

hal/src/electron/platform_ncp_electron.cpp
avtolstoy: Probably no need to implement this for Gen 2, unless we are going to use the runtime modem type detection and then cache this data in DCT, so that devices manufactured before we've introduced OTP support on Gen 2 can kinda use this.

hal/src/electron/modem/mdm_hal.cpp
```diff
@@ -882,9 +917,9 @@ bool MDMParser::_powerOn(void)
     bool retried_after_reset = false;
     int i = MDM_POWER_ON_MAX_ATTEMPTS_BEFORE_RESET; // When modem not responsive on boot, AT/OK tries 25x (for ~30s) before hard reset
 
-    // FIXME: REMOVE THIS BEFORE MERGING!!!!!!
-    _dev.dev = DEV_SARA_R510;
-    MDM_ERROR("Forcing to R510\r\n");
+    auto otp_ncp_id = platform_primary_ncp_identifier();
+    _dev.dev = fromPlatformNcpIdentifer(otp_ncp_id);
+    MDM_INFO("ncp_id otp value: 0x%X _dev.dev SET AS = 0x%X \r\n", otp_ncp_id, _dev.dev);
```
avtolstoy, line 922: My original comment about `MDM_` macros still stands. For any new code let's use `LOG()`. We also have a long standing story to refactor this: see CH50252.

hal/src/electron/platform_ncp_electron.cpp
avtolstoy: Just for consistency, but you might want to just hardcode `USI09` for Photon and `USI14` for P1.

hal/src/electron/platform_ncp_electron.cpp
avtolstoy: This should be it :+1:

modules/electron/system-part1/src/cellular_hal.cpp
avtolstoy: :+1: Due to the weirdness in dependency order, this is the right hack.

hal/src/electron/modem/mdm_hal.cpp
avtolstoy: Or `cellular_internal.h/cpp` might also be another place for this and the implementation there can be used in both `mdm_hal` and `cellular_device_info`.

hal/src/electron/platform_ncp_electron.cpp
avtolstoy: I'd proably leave this as-is.

hal/src/electron/platform_ncp_electron.cpp
scott-brust: From the documentation, the Photon and P1 use a BCM43362 module. How does this relate to `PLATFORM_NCP_BROADCOM_BCM9WCDUSI09` and `PLATFORM_NCP_BROADCOM_BCM9WCDUSI14`?

It makes sense to add `/hal/src/photon/platform_ncp_photon.cpp` for photon, but what about P1? Will it use the same code? Should I check for `PLATFORM` in `platform_primary_ncp_identifier()` and return `USI09` on photon platform and `USI14` on P1 platform? (Also, I dont have a photon or P1 board to test with)

avtolstoy: BCM43362 is the Cypress/Broadcom IC. What is used on Photon and P1s are modules with additional components and their own differences. The one on Photon is called `BCM9WCDUSI09` the one on P1 is `BCM9WCDUSI14`.

Yes, you should implement it once for both Photon and P1 and use a conditional on e.g. `PLATFORM_ID == PLATFORM_PHOTON`.

hal/src/electron/modem/mdm_hal.cpp
scott-brust: I was trying to find a shared spot that could be included from gen 2 `mdm_hal.cpp` as well as gen3 `cellular_hal.cpp`. `cellular_internal.cpp` seems to be only used by electron? Should I try to include this for gen3 builds?

avtolstoy: In that case `hal/shared` would be your best candidate. Similar to how `cellular_sig_perc_mapping`-stuff is implemented there and shared across Gen 2 and Gen 3.

hal/src/photon/platform_ncp_photon.cpp
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "platform_ncp.h"
+#include "system_error.h"
+
+PlatformNCPIdentifier platform_primary_ncp_identifier() {
+    #if PLATFORM_ID == 6
```
avtolstoy, line 22: Magic number. There is `platforms.h` header with proper platform id definitions.

avtolstoy (approved):

PR #2340, opened 2021-07-15
[gcc] Enable -Wextra

avtolstoy:
<details>

```

```
</details>

### Problem

There are many more diagnostics available in GCC which can be enabled with `-Wextra`, which can protect us against common pitfalls e.g. forgetting to initialize on of the fields of a struct, forgetting to `break` out of `switch-case` etc.

### Solution

Enable `-Wextra` for all the modules except application firmware and disable some of the more annoying warnings e.g. unused function argument.

### Steps to Test

N/A. CI should build all the platforms.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/photon/wlan_hal.cpp
```diff
@@ -1567,8 +1567,8 @@ int wlan_scan(wlan_scan_result_t callback, void* cookie)
     memset(&info, 0, sizeof(info));
     info.callback = callback;
     info.callback_data = cookie;
-    int result =  sniff_security(&info);
-    return result < 0 ? result : info.count;
+    auto result = sniff_security(&info);
+    return result != WICED_SUCCESS ? -result : info.count;
```
sergeuz, line 1571: :+1:

hal/src/electron/ota_module_bounds.c
```diff
@@ -1,25 +1,25 @@
...
-const module_bounds_t module_factory_mono = { 0x60000, 0x8080000, 0x80E0000, MODULE_FUNCTION_MONO_FIRMWARE, 0, MODULE_STORE_FACTORY };
-const module_bounds_t module_ota_mono = { 0x60000, 0x8080000, 0x80E0000, MODULE_FUNCTION_NONE, 0, MODULE_STORE_SCRATCHPAD };
+const module_bounds_t module_user_mono = { 0x60000, 0x8020000, 0x8080000, MODULE_FUNCTION_MONO_FIRMWARE, 0, MODULE_STORE_MAIN, 0, MODULE_BOUNDS_LOC_INTERNAL_FLASH };
+const module_bounds_t module_factory_mono = { 0x60000, 0x8080000, 0x80E0000, MODULE_FUNCTION_MONO_FIRMWARE, 0, MODULE_STORE_FACTORY, 0, MODULE_BOUNDS_LOC_INTERNAL_FLASH };
+const module_bounds_t module_ota_mono = { 0x60000, 0x8080000, 0x80E0000, MODULE_FUNCTION_NONE, 0, MODULE_STORE_SCRATCHPAD, 0, MODULE_BOUNDS_LOC_INTERNAL_FLASH };
 
 #if defined(MODULAR_FIRMWARE) && MODULAR_FIRMWARE
-const module_bounds_t* module_bounds[] = { &module_bootloader, &module_system_part1, &module_system_part2, &module_system_part3, &module_user, &module_factory };
+const module_bounds_t* module_bounds[] = { &module_bootloader, &module_system_part1, &module_system_part2, &module_system_part3, &module_user, &module_factory, 0, MODULE_BOUNDS_LOC_INTERNAL_FLASH };
```
XuGuohui, line 20: This is a pointer array and don't need to change it. Same to where `module_bounds` is defined.

wiring/src/spark_wiring_json.cpp
```diff
@@ -542,7 +543,7 @@ void spark::JSONWriter::writeEscaped(const char *str, size_t size) {
     const char *s = str;
     while (s != end) {
         const char c = *s;
-        if (c == '"' || c == '\\' || (c >= 0 && c <= 0x1f)) {
+        if (c == '"' || c == '\\' || std::isprint(c) || std::iscntrl(c)) {
```
sergeuz, line 546: `std::isprint(c)` here will cause all printable characters to be escaped. Also, I believe you should explicitly cast the argument to `unsigned char` for portability when using functions from ctype.h.

system/src/system_network_diagnostics.cpp
```diff
@@ -109,8 +109,7 @@ class NetworkCache
 private:
 #if Wiring_Cellular
     CellularSignal sig_;
-    CellularGlobalIdentity cgi_{.size = sizeof(CellularGlobalIdentity),
-                                .version = CGI_VERSION_LATEST};
+    CellularGlobalIdentity cgi_ = {};
```
XuGuohui, line 112: We're clearing the `.size` and `.version` fields.

wiring/src/spark_wiring_stream.cpp
```diff
@@ -192,7 +193,7 @@ float Stream::parseFloat(char skipChar){
 
   c = peekNextDigit();
     // ignore non numeric leading characters
-  if(c < 0)
+  if(!std::isprint(c))
```
sergeuz, line 196: I suggest instead of using `isprint()` here we store the result of the `peekNextDigit()` call in a separate `int` variable and assign it to `c` if it's non-negative.

XuGuohui (changes requested):

hal/src/electron/ota_module_bounds.c
avtolstoy: :+1: Edited too much.

system/src/system_network_diagnostics.cpp
avtolstoy: :+1:

avtolstoy: Actually, let's leave it as-is. `getCGI()` anyway configures them correctly.

services/inc/enumflags.h
```diff
@@ -264,6 +267,12 @@ inline bool particle::EnumFlags<T, typename std::enable_if_t<std::is_enum<T>::va
     return (value_ <= flags.value_);
 }
 
+template<typename T>
+inline particle::EnumFlags<T>& particle::EnumFlags<T, typename std::enable_if_t<std::is_enum<T>::value>>::operator=(EnumFlags<T> flags) {
+    std::swap(*this, flags);
+    return *this;
+}
```
sergeuz, line 274: Nit: A common idiom for swapping objects in C++ is
```
using std::swap; // Enable ADL
swap(obj1, obj2);
```

sergeuz (changes requested):

hal/src/nRF52840/ota_flash_hal.cpp
```diff
@@ -794,7 +794,7 @@ void HAL_FLASH_Write_ServerAddress(const uint8_t *buf, bool udp)
 int HAL_Set_System_Config(hal_system_config_t config_item, const void* data, unsigned data_length)
 {
     unsigned offset = 0;
-    unsigned length = -1;
+    int length = -1;
```
tstellanova, line 797: could we keep the `unsigned length` and just check for length > 0 below?

wiring/src/spark_wiring_json.cpp
avtolstoy: I think I misread that as `0x7f` which would be matching `isprint()` + `iscntrl()`. :+1:

wiring/src/spark_wiring_stream.cpp
avtolstoy: :+1:

hal/src/nRF52840/ota_flash_hal.cpp
avtolstoy: Yeah, might be an option :+1:

hal/src/stm32f2xx/ota_flash_hal_stm32f2xx.cpp
```diff
@@ -616,7 +616,7 @@ const uint8_t* fetch_device_public_key(uint8_t lock)
 int HAL_Set_System_Config(hal_system_config_t config_item, const void* data, unsigned data_length)
 {
     unsigned offset = 0;
-    unsigned length = -1;
+    int length = -1;
```
tstellanova, line 619: keep `unsigned length` and use `length > 0` below?

hal/src/stm32f2xx/bkpreg_hal.c
```diff
@@ -17,7 +17,7 @@ int32_t HAL_Core_Backup_Register(uint32_t BKP_DR)
 
 void HAL_Core_Write_Backup_Register(uint32_t BKP_DR, uint32_t Data)
 {
-    uint32_t BKP_DR_Address = HAL_Core_Backup_Register(BKP_DR);
+    int32_t BKP_DR_Address = HAL_Core_Backup_Register(BKP_DR);
```
tstellanova, line 20: I guess the backup register addresses at this point are known to be less than `0x7FFFFFFF`  -- but it does seem weird that `HAL_Core_Backup_Register` returns an `int32_t`

tstellanova (approved):
Only minor suggestions

wiring/src/spark_wiring_json.cpp
avtolstoy: Fixed

wiring/src/spark_wiring_stream.cpp
avtolstoy: Fixed

services/inc/enumflags.h
avtolstoy: Fixed

hal/src/nRF52840/ota_flash_hal.cpp
avtolstoy: Fixed

hal/src/stm32f2xx/ota_flash_hal_stm32f2xx.cpp
avtolstoy: Fixed

hal/src/stm32f2xx/bkpreg_hal.c
avtolstoy: I think there is a bit of a confusion in these APIs and what is meant to be returned here is actually a backup register 'index' if you look at `RTC_BKP_DRX` definitions and also how it's implemented in Gen 3 (index in a lookup table), so it does make sense that `-1` would be 'unknown'.

sergeuz (approved):

PR #2344, opened 2021-08-03
Additional state for firmware update checks

sergeuz:
### Problem

`System.updatesPending()` can only return `true` or `false` and it's impossible to tell whether `false` means that an update is indeed not available or that the system hasn't performed a check for updates yet.

### Solution

Introduce a separate method, `System.updateStatus()`, that returns one of the values defined by the `UpdateStatus` enum:
- `UNKNOWN` – The system will check for firmware updates when the device connects to the cloud.
- `NOT_AVAILABLE` – No firmware update available.
- `PENDING` – A firmware update is available.
- `IN_PROGRESS` – A firmware update is in progress.

In addition to `System.updateStatus()`, update status changes are also reported via `firmware_update_status` system events.

### Steps to Test

1. Flash the example product app to a device and connect to the device via the CLI's serial console. Make sure to set the app's `PRODUCT_ID` to your product ID.
2. Verify via the log that `System.updateStatus()` initially returns `UNKNOWN` and that the `NOT_AVAILABLE` status is reported via both `System.updateStatus()` and a `firmware_update_status` event shortly after the cloud connection is established.
3. Release a new version of the product firmware with IOTA enabled. Verify via the log that the update status has changed to `PENDING`.
4. Enable updates by clicking the setup button. Verify via the log that the update status has changed to `IN_PROGRESS`.

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC)
SYSTEM_THREAD(ENABLED)

PRODUCT_ID(...) // FIXME
PRODUCT_VERSION(1)

namespace {

const SerialLogHandler logHandler(LOG_LEVEL_WARN, {
    { "app", LOG_LEVEL_ALL }
});

int lastStatus = std::numeric_limits<int>::min();
volatile bool toggleUpdates = false;

void systemEventHandler(system_event_t event, int value, void* data)
{
    switch (event) {
    case firmware_update_status: {
        switch (value) {
        case (int)UpdateStatus::UNKNOWN:
            Log.info("firmware_update_status: UNKNOWN");
            break;
        case (int)UpdateStatus::NOT_AVAILABLE:
            Log.info("firmware_update_status: NOT_AVAILABLE");
            break;
        case (int)UpdateStatus::PENDING:
            Log.info("firmware_update_status: PENDING");
            break;
        case (int)UpdateStatus::IN_PROGRESS:
            Log.info("firmware_update_status: IN_PROGRESS");
            break;
        default:
            break;
        }
        break;
    }
    case button_final_click: {
        if (value == 1) { // 1 click
            toggleUpdates = true;
        }
        break;
    }
    default:
        break;
    }
}

} // namespace

void setup() {
    System.on(firmware_update_status, systemEventHandler);
    System.on(button_final_click, systemEventHandler);
    System.disableUpdates();
    Particle.connect();
}

void loop() {
    const int status = System.updateStatus();
    if (lastStatus != status) {
        lastStatus = status;
        if (status < 0) {
            Log.error("System.updateStatus() failed: %d", status);
        } else {
            switch (status) {
            case (int)UpdateStatus::UNKNOWN:
                Log.info("System.updateStatus(): UNKNOWN");
                break;
            case (int)UpdateStatus::NOT_AVAILABLE:
                Log.info("System.updateStatus(): NOT_AVAILABLE");
                break;
            case (int)UpdateStatus::PENDING:
                Log.info("System.updateStatus(): PENDING");
                break;
            case (int)UpdateStatus::IN_PROGRESS:
                Log.info("System.updateStatus(): IN_PROGRESS");
                break;
            default:
                break;
            }
        }
    }
    if (toggleUpdates) {
        toggleUpdates = false;
        if (System.updatesEnabled()) {
            System.disableUpdates();
            Log.info("Firmware updates disabled");
        } else {
            System.enableUpdates();
            Log.info("Firmware updates enabled");
        }
    }
}
```

### References

- [ch62095]

### DIFF
--- a/system/inc/system_dynalib.h
+++ b/system/inc/system_dynalib.h
@@ -117,6 +117,8 @@ DYNALIB_FN(BASE_IDX1 + 2, system, system_power_management_get_config, int(hal_po
 #define BASE_IDX2 (BASE_IDX1 + 2)
 #endif  // HAL_PLATFORM_POWER_MANAGEMENT
 
+DYNALIB_FN(BASE_IDX2 + 0, system, system_get_update_status, int(void*))
+
 DYNALIB_END(system)
 
 #undef BASE_IDX
--- a/system/inc/system_event.h
+++ b/system/inc/system_event.h
@@ -47,20 +47,21 @@ enum SystemEvents {
     setup_all = wifi_listen,
     network_credentials = 1<<4,
     network_status = 1<<5,
-    cloud_status = 1<<6,             // parameter is 0 for disconnected, 1 for connecting, 8 for connected, 9 for disconnecting. other values reserved.
-    button_status = 1<<7,            // parameter is >0 for time pressed in ms (when released) or 0 for just pressed.
-    firmware_update = 1<<8,          // parameter is 0 for begin, 1 for OTA complete, -1 for error.
-    firmware_update_pending = 1<<9,	// notifies the application that an OTA update is pending and will be delivered when updates are enabled
+    cloud_status = 1<<6,            // parameter is 0 for disconnected, 1 for connecting, 8 for connected, 9 for disconnecting. other values reserved.
+    button_status = 1<<7,           // parameter is >0 for time pressed in ms (when released) or 0 for just pressed.
+    firmware_update = 1<<8,         // parameter is 0 for begin, 1 for OTA complete, -1 for error.
+    firmware_update_pending = 1<<9, // notifies the application that an OTA update is pending and will be delivered when updates are enabled
     reset_pending = 1<<10,          // notifies that the system would like to shutdown (System.resetPending() return true)
     // todo - rename to system_reset, or otherwise avoid common name clashes
-	reset = 1<<11,                  // notifies that the system will now reset on return from this event.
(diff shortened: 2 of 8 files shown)


keeramis:
I gave it a test on my bsom and it works as expected @sergeuz

If I am trying to do a v2 -> v1 on my product though, like below - it wouldn't ota even though ota updated is enabled. Perhaps it's expected coz it's a lower version? Posting a screenshot nevertheless. 

image

```
0000828250 [app] INFO: Firmware updates enabled
0000837909 [app] INFO: System.updateStatus(): NOT_AVAILABLE
0000837909 [app] INFO: firmware_update_status: NOT_AVAILABLE
0000843030 [app] INFO: System.updateStatus(): PENDING
0000843030 [app] INFO: firmware_update_status: PENDING
// waited here but ota didn't come through.
```

keeramis (approved):

wraithan:
Looks like you triggered it as a regular OTA instead of an IOTA, regular OTA is mostly an after reboot or new session thing, IOTA would have flashed it immediately since the device was available for updates.

avtolstoy (approved):

technobly (approved):

PR #2346, opened 2021-08-19
Clear module slots in DCT when preparing for an OTA update

sergeuz:
### Problem

When system resets are disabled via `System.disableReset()`, it is possible to flash the device OTA multiple times without it resetting itself to apply the first received update. Module slots that are stored in the DCT and trigger the bootloader to copy the module's data from the OTA section to its destination section in the internal flash are not cleared in this case, even if the received modules occupy the same area in the OTA section, which may result in a corrupted firmware update.

### Solution

Clear all non-factory module slots in the DCT at the beginning of an OTA update.

### Steps to Test

1. Flash the example application to a Gen 2 or Gen 3 device. Let the device connect to the cloud.
2. Flash the main system part module that is already running on the device OTA. The device will not reset after the update and stay connected to the cloud.
3. Flash some firmware module that has unmet dependencies to the device OTA. One way to get such a module is to bump the `BOOTLOADER_DEPENDENCY` version of the main system part module that is running on the device.
4. Reset the device via the reset button and verify that the bootloader doesn't try to apply any updates. Running this test without the fix in place would result in the bootloader copying the module with unmet dependencies to its destination section in the internal flash.

### Example App

```c
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC)
SYSTEM_THREAD(ENABLED)

void setup() {
    System.disableReset();
    Particle.connect();
}

void loop() {
}
```

### References

- [ch84957]

### DIFF
--- a/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/flash_mal.c
+++ b/platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/flash_mal.c
@@ -873,16 +873,43 @@ uint32_t FLASH_PagesMask(uint32_t imageSize, uint32_t pageSize)
     return numPages;
 }
 
-void FLASH_Begin(uint32_t FLASH_Address, uint32_t imageSize)
+int FLASH_Begin(uint32_t FLASH_Address, uint32_t imageSize)
 {
     system_flags.OTA_FLASHED_Status_SysFlag = 0x0000;
     Save_SystemFlags();
 
+    // Clear all non-factory module slots in the DCT
+    const size_t slot_count = MAX_MODULES_SLOT - GEN_START_SLOT;
+    size_t slot_offs = DCT_FLASH_MODULES_OFFSET + sizeof(platform_flash_modules_t) * GEN_START_SLOT;
+    for (size_t i = 0; i < slot_count; ++i) {
+        uint16_t magic_num = 0;
+        int r = dct_read_app_data_copy(slot_offs + offsetof(platform_flash_modules_t, magicNumber), &magic_num,
+                sizeof(magic_num));
+        if (r != 0) {
+            return FLASH_ACCESS_RESULT_ERROR;
+        }
+        if (magic_num == 0xabcd) {
+            // Mark the slot as unused
+            magic_num = 0xffff;
+            r = dct_write_app_data(&magic_num, slot_offs + offsetof(platform_flash_modules_t, magicNumber),
+                    sizeof(magic_num));
+            if (r != 0) {
+                return FLASH_ACCESS_RESULT_ERROR;
+            }
+        }
+        slot_offs += sizeof(platform_flash_modules_t);
+    }
+
 #ifdef USE_SERIAL_FLASH
-    FLASH_EraseMemory(FLASH_SERIAL, FLASH_Address, imageSize);
+    const bool ok = FLASH_EraseMemory(FLASH_SERIAL, FLASH_Address, imageSize);
 #else
-    FLASH_EraseMemory(FLASH_INTERNAL, FLASH_Address, imageSize);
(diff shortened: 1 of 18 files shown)


avtolstoy (approved):

avtolstoy:
Looks good, we have a path forward for an automated test for this as well but dependent on hal_system_info changes.

platform/MCU/STM32F2xx/SPARK_Firmware_Driver/src/flash_mal.c
```diff
@@ -873,16 +873,43 @@ uint32_t FLASH_PagesMask(uint32_t imageSize, uint32_t pageSize)
...
+    const bool ok = FLASH_EraseMemory(FLASH_SERIAL, FLASH_Address, imageSize);
 #else
-    FLASH_EraseMemory(FLASH_INTERNAL, FLASH_Address, imageSize);
+    const bool ok = FLASH_EraseMemory(FLASH_INTERNAL, FLASH_Address, imageSize);
 #endif
+    if (!ok) {
+        return FLASH_ACCESS_RESULT_ERROR;
+    }
+
+    return 0;
```
technobly, line 912: Suggestion: `return FLASH_ACCESS_RESULT_OK;`

platform/MCU/nRF52840/src/flash_mal.c
```diff
@@ -952,16 +952,43 @@ void FLASH_Restore(uint32_t FLASH_Address)
...
+    const bool ok = FLASH_EraseMemory(FLASH_SERIAL, FLASH_Address, imageSize);
 #else
-    FLASH_EraseMemory(FLASH_INTERNAL, FLASH_Address, imageSize);
+    const bool ok = FLASH_EraseMemory(FLASH_INTERNAL, FLASH_Address, imageSize);
 #endif
+    if (!ok) {
+        return FLASH_ACCESS_RESULT_ERROR;
+    }
+
+    return 0;
```
technobly, line 991: Suggestion: `return FLASH_ACCESS_RESULT_OK;`

technobly (approved):
Looks good, Sergey!  Clearing this on every chunk received seems necessary since an async reset can happen at any time.  Nice error handling ;-)

sergeuz:
@technobly The slots are cleared just once at the beginning of the update. Clearing them on every received chunk is not necessary as they're populated with the module details only after the module binary has been fully received and validated in `HAL_FLASH_End()`.

technobly:
@sergeuz Ah, I see that now... that's even better, thanks :D

system/src/system_cloud_internal.cpp
```diff
@@ -1079,12 +1079,12 @@ void Spark_Protocol_Init(void)
...
 
         // Enable device-initiated describe messages
-        spark_protocol_set_connection_property(sp, protocol::Connection::DEVICE_INITIATED_DESCRIBE, 0, nullptr, nullptr);
+        spark_protocol_set_connection_property(sp, protocol::Connection::ENABLE_DEVICE_INITIATED_DESCRIBE, 0, nullptr, nullptr);
 
 #if HAL_PLATFORM_COMPRESSED_OTA
         // Enable compressed/combined OTA updates
         if (bootloader_get_version() >= COMPRESSED_OTA_MIN_BOOTLOADER_VERSION) {
-            spark_protocol_set_connection_property(sp, protocol::Connection::COMPRESSED_OTA, 0, nullptr, nullptr);
+            spark_protocol_set_connection_property(sp, protocol::Connection::COMPRESSED_OTA, 1, nullptr, nullptr);
```
sergeuz, line 1087: This is a safe change as compressed updates are only supported on Gen 3 where the comms library and the system are in the same system part.

sergeuz:
I added an integration test for this fix: `ota/multiple_ota_no_reset`. The test runner 0.3.13 is required in order to run it.

avtolstoy:
Tested for now manually on Tracker and Electron :heavy_check_mark:

PR #2348, opened 2021-08-23
[services] Fixes jsmnerr_t issues in jsmn

avtolstoy:
### Problem

(based on https://github.com/particle-iot/device-os/tree/fix/jsmn-library-updates)

Due to usage of `-fshort-enums` we were exporting `jsmn_parse()` which in fact was only returning an `int8_t`, preventing parsing of relatively large JSONs. This is now fixed in the upstream version of jsmn library (returns in `int`) and also in our "fork" contained within Device OS sources in services.

We will sync up with the upstream jsmn at a later point.

### Solution

1. Update `jsmn` fork sources to return `int` and not use `jsmnerr_t` directly
2. Add `jsmn_parse_deprecated()` with old signature for dynalib compatibility purposes

### Steps to Test

N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/services/inc/jsmn.h
+++ b/services/inc/jsmn.h
@@ -1,28 +1,26 @@
-/*
- * MIT License
- *
- * Copyright (c) 2010 Serge Zaitsev
- *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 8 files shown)


sergeuz (approved):

avtolstoy:
@eberseth Can you post an example of how you triggered this with config service?

eberseth:
> @eberseth Can you post an example of how you triggered this with config service?

I think your test adequately checks for the condition.

eberseth:
Tested out on device as well with real data.  it works.

eberseth (approved):

PR #2356, opened 2021-09-14
Avoid freeing memory in detachInterrupt

metcalf:
### Problem
When `attachHandler` is called with a function pointer (or a pointer to a member function and instance), it allocates a `std::function`. When `detachHandler` is then called on the same pin, it calls `delete` to free that memory. That causes a heap error if you try to call `detachHandler` on this pin from an ISR. The same is true for attaching and detaching system interrupts.

### Solution
This PR removes the `delete`s and leaves the memory allocated to be re-used the next time `attach*` is called. This prevents the application from reclaiming the memory `std::function` is allocating on the heap, but I suspect that's OK since there can be no more allocations than the number of pins. 

I think the alternative is to update the documentation to specify when `detachInterrupt` can be called. It's probably still worth updating the docs to specify that `attachInterrupt` and `attachSystemInterrupt` cannot be called from an ISR when used with a function pointer / class+member.

### Steps to Test
Wrote and ran an integration test on my Photon.

### Example App
Deferring to the integration test, but can write something if necessary.

### References
The `delete` in `detachSystemInterrupt` was added in #951. I think the primary problem fixed in that PR was an unbounded memory leak repeatedly calling `attachSystemInterrupt`.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/interrupts.cpp
+++ b/user/tests/wiring/no_fixture/interrupts.cpp
@@ -81,6 +81,26 @@ public:
 	static int count;
 };
 
+class DetachingHandler
+{
+	public:
+		DetachingHandler(pin_t p): pin(p) {}
+
+		void handler()
+		{
+			cnt++;
+			detachInterrupt(pin);
+		}
+
+		int count() {
+			return cnt;
+		}
+
+	private:
+		pin_t pin;
+		volatile int cnt = 0;
+};
+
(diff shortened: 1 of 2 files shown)


metcalf:
Checking in on this since it's blocking me from using Cloud Flash. Is this likely to get merged or should I rewrite my application to avoid this issue?

PR #2357, opened 2021-09-16
Add missing DYNALIB_FNs for os_thread_{wait,notify}

metcalf:
### Problem
```
/Users/andrew/code/supply_ventilation//src/DisplayTask.cpp:51: undefined reference to `os_thread_notify'
/Users/andrew/.particle/toolchains/gcc-arm/10.2.1/bin/../lib/gcc/arm-none-eabi/10.2.1/../../../../arm-none-eabi/bin/ld: ../../../build/target/user/platform-6-m/supply_ventilation//libuser.a(DisplayTask.o): in function `DisplayTask::_taskFun()':
/Users/andrew/code/supply_ventilation//src/DisplayTask.cpp:59: undefined reference to `os_thread_wait'
```
### Solution
Added missing `DYNALIB_FN` entries for `os_thread_{wait,notify}`

### Steps to Test
I assume existing compilation/tests would break if I

### Example App
N/A

### Completeness

- [ ] Added documentation


metcalf:
Hmm, maybe this isn't an API that's intended to be exposed directly given the existence of the `Thread` class? Should `Thread` support these methods?

PR #2370, opened 2021-10-22
Enabling building of GCC platform as a standalon shared library with no dependencies on e.g. Boost

avtolstoy:
### TODO

1. Add a cmake file (potentially generated in runtime) which can simplify using this in libraries/applications
2. Update CI to generate ready-to-use containers during release with Device OS prebuilt as a shared library

### Problem

We want to simplify unit testing of libraries or applications that rely on Device OS APIs.

### Solution

Build complete GCC platform for the host machine without any external dependencies on e.g. Boost, so that only stubs are generated, which can be easily mocked e.g. with Hippomocks.

NOTE: this requires GCC >= 6, as we need `-fno-plt` to enable mocking of symbols coming from shared library.

### Steps to Test

See https://github.com/particle-iot-inc/device-os-unit-test-demo

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


eberseth (approved):

eberseth:
The "steps to test" repo is not public.  Can you provide snippets here?

avtolstoy:
@eberseth It's in a different org, you should have permissions for it though?

eberseth:
@avtolstoy Could we merge this for 6.x?  There is some future work that would benefit from not having Boost dependencies.

avtolstoy:
@eberseth This is only useful for unit testing/mocking.
> Build complete GCC platform for the host machine without any external dependencies on e.g. Boost, so that only stubs are generated, which can be easily mocked e.g. with Hippomocks.

GCC platform to work does require Boost, this will simply replace proper implementations with empty stubs.

PR #2374, opened 2021-11-03
[gen3] hal: zero out invalid user modules

avtolstoy:
### Problem

Devices may present invalid user modules in system describe which can appear as some other type of module with potentially a set of invalid dependencies which cannot be satisfied preventing OTA from working correctly.

This happens because we ignore integrity or any other checks when making a decision on whether to include read-out module data in system describe. This is more problematic with Gen 3 devices running 3.1.0+ with 256KB applications as the compat (128KB) module header location would be in the middle of the larger binary. If by chance bytes at that location contain something seemingly valid (e.g. correct module type), this erroneous module would be reported in the describe message. If there is additionally a seemingly valid dependency, this is going to prevent product updates from happening, as DS will not be able to satisfy it.

### Solution

Zero out module data if we've fetched an invalid module from `module_user` or `module_user_compat` locations.

### Steps to Test

`wiring/gen3_invalid_compat_user_app`

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ota_flash_hal.cpp
+++ b/hal/src/nRF52840/ota_flash_hal.cpp
@@ -188,8 +188,17 @@ int HAL_System_Info(hal_system_info_t* info, bool construct, void* reserved)
                 }
                 bool valid = fetch_module(module, bounds, false, MODULE_VALIDATION_INTEGRITY);
                 valid = valid && (module->validity_checked == module->validity_result);
-                if (valid && bounds->module_function == MODULE_FUNCTION_USER_PART) {
-                    user_module_found = true;
+                if (bounds->module_function == MODULE_FUNCTION_USER_PART) {
+                    if (valid) {
+                        user_module_found = true;
+                    } else {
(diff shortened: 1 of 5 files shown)


keeramis (approved):

keeramis:
Do we also need to modify these tests accordingly? Perhaps the `info` collected from `HAL_System_Info` is now different(?)
```
Assertion failed: (factory == nullptr=1) == (false=0), file tests/integration/wiring/no_fixture/system.cpp, line 168.
        3) SYSTEM_06_system_describe_is_not_overflowed_when_factory_module_present
        Assertion failed: (factory == nullptr=1) == (false=0), file tests/integration/wiring/no_fixture/system.cpp, line 261.
        4) SYSTEM_07_system_describe_is_not_overflowed_when_factory_module_present_but_invalid
```

avtolstoy:
@keeramis Good catch, we should actually also check that we are checking against `MODULE_STORE_MAIN`. Thanks

avtolstoy:
Fixed

Issue #2380, opened 2021-12-04
WiFi dropout causes hardfault, hardfault handler is broken and P1 doesn't SOS and reset


elcojacobs:
### Expected Behavior

Hardfault handler should blink RED status LED and reboot

### Observed Behavior

Hardfault results in complete system hangup and system doesn't reset.

I built my app with a non-modular build with SWD enabled in an attempt to find the cause of complete system hangups when the wifi becomes unavailable.
The cause of the hangup is inaccessible to me because it is in the closed source WICED files.
However, I noticed that I couldn't break the debugger on hardfaults anymore in 'prvGetRegistersFromStack'. I did looked at the history of the hardfault handler and reverted a change. This made SOS and hardfault work again.

### Steps to Reproduce
- Clone our repo
- In the <repo root>/build directory, run `make program-dfu MODULAR=n PLATFORM=p1 PARTICLE_DEVELOP=y USE_SWD=y APP=brewblox`
- Turn WiFi on and off on router until turning it off causes a hardfault

### Test App
https://github.com/BrewBlox/brewblox-firmware/tree/develop/app/brewblox-particle

### Analysis
Since recently, I had issues with complete hangups of our temperature controllers. Even the watchdog didn't reset the system anymore.

While trying to step debug using SWD, I noticed the system hung in the `HardFault_Handler` function, without jumping into `prvGetRegistersFromStack.

In this commit:

https://github.com/particle-iot/device-os/commit/7b6e2086af907668d395c63404ecc754ee36fe41#diff-cd3f7ac0caef6976249cf63ff8b1dbcfe1c3150d09e3b694a45cfbc3354c85d7

```cpp
void HardFault_Handler(void)
{
    __asm volatile
    (
        " tst lr, #4                                                \n"
        " ite eq                                                    \n"
        " mrseq r0, msp                                             \n"
        " mrsne r0, psp                                             \n"
        " ldr r1, [r0, #24]                                         \n"
        " ldr r2, handler2_address_const                           \n"
        " bx r2                                                     \n"
        " handler2_address_const: .word prvGetRegistersFromStack    \n"
    );
}
```

was changed into:

```
void HardFault_Handler(void)
{
    __asm volatile
    (
        " tst lr, #4                                                \n"
        " ite eq                                                    \n"
        " mrseq r0, msp                                             \n"
        " mrsne r0, psp                                             \n"
        " ldr r1, [r0, #24]                                         \n"
[...]

PR #2381, opened 2021-12-04
Fix hardfault handler to SOS again

elcojacobs:
This fixes the broken hardfault handler as reported in #2380 

I added the `.align 4` line, which I think might be the actual fix for the compiler warning that the commit that introduced the bug intended to address.

It is very easy to test the hard fault handler, so it seems that the compiler warning was 'fixed' in the commit below without any testing! :fearful: 

https://github.com/particle-iot/device-os/commit/7b6e2086af907668d395c63404ecc754ee36fe41

I think that you are actually causing a hardfault in the hardfault handler now, resulting in an endless loop! @avtolstoy @sergeuz 

## Test app:

```cpp
void setup()
{
    int (*bad_instruction)(void) = nullptr;
    bad_instruction();
}

void loop()
{
    HAL_Delay_Milliseconds(1);
}
```

### DIFF
--- a/bootloader/src/nRF52840/nrf_it.c
+++ b/bootloader/src/nRF52840/nrf_it.c
@@ -83,8 +83,9 @@ void HardFault_Handler(void)
         " mrseq r0, msp                                             \n"
         " mrsne r0, psp                                             \n"
         " ldr r1, [r0, #24]                                         \n"
-        " ldr r2, handler2_address_const                           \n"
+        " ldr r2, handler2_address_const                            \n"
         " bx r2                                                     \n"
+        " .align 4                                                  \n"
         " handler2_address_const: .word prvGetRegistersFromStack    \n"
     );
 }
--- a/bootloader/src/nRF52840/nrf_it.c
+++ b/bootloader/src/nRF52840/nrf_it.c
@@ -83,7 +83,7 @@ void HardFault_Handler(void)
         " mrseq r0, msp                                             \n"
         " mrsne r0, psp                                             \n"
         " ldr r1, [r0, #24]                                         \n"
-        " ldr r2, =handler2_address_const                           \n"
+        " ldr r2, handler2_address_const                           \n"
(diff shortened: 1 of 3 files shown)


elcojacobs:
It seems that the hardfault handler is part of the system layer, so it cannot be easily hotfixed without changing how the system layer is downloaded to the device. 
Because this is not fixable in the user app, we depend on Particle to do a hotfix release ASAP.

Or does the hardfault handler have weak linkage so we can override it?

sergeuz (approved):

sergeuz:
Thanks for the PR, @elcojacobs. I'm not sure how this issue sneaked through testing, but your fix resolves it for me :+1:

elcojacobs:
Thanks for merging the fix. Will build this into a release soon too?

sergeuz:
> Will build this into a release soon too?

We will include this fix in the next 3.x release that we're hoping to get out this month.

elcojacobs:
Can you please do better than 'this month'?

This is a critical bug that causes a complete device freeze. It causes large batches of beer to be ruined for my customers. Note that even the watchdog stops working with this bug.

You don't have to fast track 3.2.0 for me, just merge only the hardfault fix into the last release candidate.

elcojacobs:
Can you discuss a hotfix release internally and inform us of the outcome?
This fix doesn't conflict with any other changes and should be an easy merge and release. I think the type of bug definitely warrants a hotfix release.

Waiting more than a week is unacceptable for my customers. If you cannot do a hotfix release, we have two options:

- Force all our customers to do a temporary flash over usb with hotfixed system images we compile ourselves. We already flash the user app over local wifi using ymodem, but the system layer comes from the cloud.
- Change the app to download the system images from our servers instead of particle. Where would we make that change and how should we host the binaries?

sergeuz:
That's the only timeline I have at the moment, but the possibility of a hotfix release is being discussed in the context of your ticket for the original issue.

PR #2379, opened 2021-12-03
[Gen3] Feature to disable listening mode

keeramis:
### Description

This PR introduces a feature flag that disables listening mode on Gen3 platform. This feature is originally being introduced to separate listening mode from BLE provisioning mode (a feature that supports BLE connections with custom mobile apps - for more info on provisioning mode, see this \<insert link here when available>). This flag is disabled by default, hence the users who do not wish to use this flag can continue using listening mode without any changes to their user apps and BLE setup processes. Enabling this feature flag can be done similar to any other feature flags. 

An example app that enables this feature looks like the following

```c
#include "Particle.h"

Serial1LogHandler log1Handler(115200, LOG_LEVEL_ALL);
SerialLogHandler logHandler(LOG_LEVEL_ALL);

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);
STARTUP(System.enableFeature(FEATURE_DISABLE_LISTENING_MODE));

void setup() {
  // A call to 'WiFi.listen()' for example, will not work 
}

void loop() {
}

```

### Additional Notes
- There will be a blip in the cloud connection in the following scenarios when attempting to enter listening mode while the listening mode is disabled through the feature flag
    - User presses SETUP button
    - USB control request to enter listening mode
- The device continues into user app when the flag is disabled irrespective of the status of the setup-done flag. This may not be concerning as the flag is disabled through a custom app
- If the device is already in listening mode, and the user app enables the `FEATURE_DISABLE_LISTENING_MODE` flag, the device comes out of listening mode.

### Steps to Test

Listening mode is expected to be entered in the following scenarios:

1. User presses SETUP button
2. Application calls Wiring function to enter listening mode
3. USB control request to enter listening mode
4. If there are no credentials in the device
5. If setup done is not set

### 1. User presses SETUP button
Assume this device has been previously claimed
```c
#include "Particle.h"

Serial1LogHandler log1Handler(115200, LOG_LEVEL_ALL);
SerialLogHandler logHandler(LOG_LEVEL_ALL);

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);
STARTUP(System.enableFeature(FEATURE_DISABLE_LISTENING_MODE));
void setup() {
    Particle.connect();
    waitUntil(Particle.connected);
    // User presses SETUP button to enter listening mode
    // Device will not enter listening mode and internally returns a `NOT_ALLOWED` system error
[...]

PR #2382, opened 2021-12-08
[Gen3] BLE Provisioning Mode 

keeramis:
# BLE provisioning mode

Provisioning mode allows the device to enter BLE setup process. When the device enters provisioning mode, it should be able to connect and pair with the mobile app. The module's cloud connectivity does not get disrupted while entering / in provisioning mode. The provisioning mode may run-side-by-side at all times with the user application. The user app must use `SYSTEM_THREAD(ENABLED)` but may use any of the app modes: `MANUAL`, `SEMI_AUTOMATIC`, `AUTOMATIC`. Provisioning mode and its APIs do not work when listening mode is enabled
This #2379 PR contains the changes to use the flag to disable listening mode

### CUSTOM SERVICE AND CHARACTERISTIC UUIDs
- Provisioning mode allows the device to connect using custom service and characteristic UUIDs instead of the [default ones](https://github.com/particle-iot/device-os/blob/develop/system/src/ble_control_request_channel.cpp#L95-L105)
- The custom UUIDs must be set before entering provisioning mode, and can be set only once throughout the runtime of the program (**TODO**: Investigating whether dynamic re-config of UUIDs is possible)
- Provisioning mode BLE coexists with other BLE advertising a user app is doing, within the resource constraints of the device. The resource constraints include the size of the advertisement packet, dynamic reconfiguration of the UUIDs to name a few
- Provisioning custom UUIDs will stick even after provisioning mode has exited

### SETUP CODE
- BLE module advertises setup code in place of manufacturing data when the device is in provisioning mode
- When module exits provisioning mode, the default manufacturing data is advertised
- This may not be overwritten in provisioning mode but multiple advertising data can be displayed simultaneously in provisioning mode

### CUSTOM MOBILE SECRET
- Supports setting a custom mobile secret through HAL API which is exposed to the user app through dynalib
- The custom mobile secret gets written into DCT
- Supports clearing the custom mobile secret through HAL API to restore the mobile secret that is programmed during manufacturing

### SYSTEM EVENTS
- Provides system events for the following
    - BLE connected / disconnected
    - Handshake done / failed / error
    - Other events such as network credentials set/clear are handled through the network_credentials system events

### DEVICE NAME
- Device name set by the user sticks throughout provisioning mode. (Device name previously would revert to default name ARGON-XXXXXX when listening mode was entered)

### CONTROL REQUEST FILTER
Processes control requests that are input through a vector. The filter can be cleared through `System.clearControlRequestFilter()` where all the requests are processed. Inputs are obtained through `system_control_req_filter` from the user app which contains the requests to be executed. 
```
0000044925 [system.ctrl.ble] TRACE: Received a request message; type: 506, ID: 8
0000044927 [system] TRACE: Control request 506 filtered out
0000269916 [wiring.ble] TRACE: Disconnected by remote device.
```

# Example app
```c
#include "Particle.h"
SYSTEM_MODE(SEMI_AUTOMATIC);

const char* serviceUuid = "6E400021-B5A3-F393-E0A9-E50E24DCCA9E";
const char* rxUuid = "6E400022-B5A3-F393-E0A9-E50E24DCCA9E";
const char* txUuid = "6E400023-B5A3-F393-E0A9-E50E24DCCA9E";
const char* versionUuid = "6E400024-B5A3-F393-E0A9-E50E24DCCA9E";

// 16-bit should also work
// BleUuid serviceUuid(0x1234);
// BleUuid txUuid(0x5678);
// BleUuid rxUuid(0x9123);
// BleUuid versionUuid(0x8123);

void ble_prov_mode_handler(system_event_t evt, int param) {
    if (param == ble_prov_mode_connected) {
        Log.info("BLE Event detected: ble_prov_mode_connected");
    }
    if (param == ble_prov_mode_disconnected) {
        Log.info("BLE Event detected: ble_prov_mode_disconnected");
    }
    if (param == ble_prov_mode_handshake_failed) {
        Log.info("BLE Event detected: ble_prov_mode_handshake_failed");
    }
    if (param == ble_prov_mode_handshake_done) {
        Log.info("BLE Event detected: ble_prov_mode_handshake_done");
    }
}

void nw_creds_handler(system_event_t evt, int param) {
    if (param == network_credentials_added) {
        Log.info("BLE Event detected: network_crendetials_added");
    }
}

void setup() {
    // ---------System Events---------
    System.on(ble_prov_mode, ble_prov_mode_handler);
    System.on(network_credentials, nw_creds_handler);
    
    // ---------Control request filter---------
    // System.setControlRequestFilter(SystemControlRequestAclAction::ACCEPT, {
    //     {CTRL_REQUEST_DFU_MODE, SystemControlRequestAclAction::DENY},
    //     {CTRL_REQUEST_RESET, SystemControlRequestAclAction::DENY}
    // });
    System.setControlRequestFilter(SystemControlRequestAclAction::ACCEPT);

    // ---------Provisioning Service and Characteristic UUIDs---------
    // Provisioning UUIDs must be set before initialising BLE for the first time
    // Even better to call in STARTUP()
    BLE.setProvisioningSvcUuid(serviceUuid);
    BLE.setProvisioningTxUuid(txUuid);
    BLE.setProvisioningRxUuid(rxUuid);
    BLE.setProvisioningVerUuid(versionUuid);
     
    // ---------Custom mobile secret---------
    // Uncomment the following three lines to test setting up custom mobile secret
    // char arr[HAL_DEVICE_SECRET_SIZE] = {};
    // memcpy(arr, "0123456789abcde", HAL_DEVICE_SECRET_SIZE);
    // hal_set_device_secret(arr, sizeof(arr), nullptr);
    // To clear device secret and use default, run -> hal_clear_device_secret(nullptr); 

    // ---------Setup device name---------
    BLE.setDeviceName("aabbccdd", 8); 

    // ---------Set company ID---------
    BLE.setCompanyId(0x1234);

    // ---------BLE provisioning mode---------
    BLE.provisioningMode(true);
    LOG(TRACE, "BLE prov mode status: %d", BLE.getProvisioningStatus());
    // To exit provisioning mode -> BLE.provisioningMode(false);
    
}

void loop() {
}
```
### Result
- Check in the Particle BLE app that the module can be paired with in provisioning mode, and the user can enter or change wifi creds while in provisioning mode
- Verify that the service and characteristic UUIDs are changed only once as expected
- Verify that the advertising service UUID can be changed every time the provisioning mode is entered
- Verify that the user configured device name sticks throughout provisioning mode
- Verify that an incorrect mobile secret will cause the handshake to fail and using the correct one will succeed handshake
- Verify that all the above APIs except control request filtering are only available when in provisioning mode

## Edge cases / Notes

1. Disable the listening mode flag while listening mode is in progress, and immediately enter provisioning mode
```c
void setup() {
    // some BLE advertising etc
    WiFi.listen();
    BLE.provisioningMode(true);
}
```
Log->
```c
0000035604 [system.listen] INFO: Entering listening mode
0000035605 [system] INFO: Cloud: disconnecting
0000035610 [system] INFO: Cloud: disconnected
.
.
0000060808 [system] TRACE: Ble prov mode enabled
0000060808 [app] INFO: BLE Prov mode status: 1
0000060808 [system.listen] INFO: Exiting listening mode.   // this is caused by the 'manage_listening_mode_flag'
... // no more advertising
```
In this case, since the listening mode disablement happens in the Spark_Idle_Events(), it's noticed in some cases that the provisioning mode turns ON first, and then the listening mode gets disabled causing the whole BLE advertising to stop. Adding a delay between setting the flag and turning on prov mode helps, OR checking that listening mode if off before entering prov mode helps.

2. Setting `BLE.off()` and `BLE.on()` do not reset the characteristic UUIDs, and UUIDs are able to be set only once. 

### References

#2379 
[SC-epic](https://app.shortcut.com/particle/epic/89844/fine-grained-user-app-control-over-ble-provisioning-listening-mode?cf_workflow=500127125&ct_workflow=all&vc_group_by=day)

---

### Completeness

- [ ] Added documentation


avtolstoy:
Note this branch fails to build for non-Gen3 platforms as of right now.

system/src/system_control_internal.cpp
```diff
@@ -81,30 +83,38 @@ SystemControl g_systemControl;
...
 
 void SystemControl::run() {
 #if HAL_PLATFORM_BLE
-    bleChannel_.run();
+    BleControlRequestChannel::instance(this)->run();
 #endif
 }
 
 void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
+    if (std::find(particle::system::SystemControl::vecCtrlReq.begin(), particle::system::SystemControl::vecCtrlReq.end(), req->type) != particle::system::SystemControl::vecCtrlReq.end()) {
```
keeramis, line 114: Would we like to make this usable only during prov mode?

system/src/system_ble_prov.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
+bool system_get_ble_prov_status(void* reserved) {
+    return particle::system::BleListeningModeHandler::instance()->getProvModeStatus();
+}
+
+int system_set_prov_svc_uuid(const uint8_t* svcUuid, const uint8_t* txUuid, const uint8_t* rxUuid, size_t len, void* reserved) {
```
XuGuohui, line 59: Use type `hal_ble_uuid_t*`

system/src/system_ble_prov.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (particle::system::BleControlRequestChannel::instance(particle::system::SystemControl::instance())->getProfInitStatus()) {
+        LOG(TRACE, "Ble control req channel prov UUIDs already initialized");
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    particle::system::BleControlRequestChannel::instance(particle::system::SystemControl::instance())->setProvUuids(svcUuid, txUuid, rxUuid, len);
+    return SYSTEM_ERROR_NONE;
+}
+
+int system_set_prov_adv_svc_uuid(const uint8_t* buf, size_t len, void* reserved) {
```
XuGuohui, line 72: Considering introduce a method in BleControlRequestChannel to get the current service UUID and use it in advertising data.

system/src/system_control_internal.cpp
XuGuohui: `vecCtrlReq` is a member variable of `SystemControl`, you don't need to prepend the namespace. Besides, `std::find()` cannot guarantee that whether the given type is found or not when the last element is equal to the given type. Let's do it in this way: 
```
ctrl_request_type type = CTRL_REQUEST_INVALID;
for (auto t : vecCtrlReq) {
    if (t == req->type) {
        type = req->type
        break;
    }
}
if (type == CTRL_REQUEST_INVALID) {
    return;
}
```

system/src/system_control_internal.cpp
```diff
@@ -81,30 +83,38 @@ SystemControl g_systemControl;
...
+        LOG(TRACE, "Control request %u filtered out", req->type);
+        return;
```
XuGuohui, line 116: I think the types being added to the filter should be processed instead. Others are dismissed.

system/src/system_control_internal.cpp
```diff
@@ -389,6 +399,19 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
     particle::system::SystemControl::instance()->setResult(req, result, handler, data);
 }
 
+int system_set_control_request_filter(uint16_t reqType, void* reserved) {
+    if (std::find(particle::system::SystemControl::vecCtrlReq.begin(), particle::system::SystemControl::vecCtrlReq.end(), reqType) == particle::system::SystemControl::vecCtrlReq.end()) {
```
XuGuohui, line 403: You will add a duplicated type if the last element in the vector is equal to the given type

system/src/system_control_internal.cpp
```diff
@@ -389,6 +399,19 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
...
+        LOG_DEBUG(TRACE, "Adding a new req to control req filter: %u", reqType);
+        particle::system::SystemControl::vecCtrlReq.insert(particle::system::SystemControl::vecCtrlReq.size(),reqType);
```
XuGuohui, line 405: Simply use `append()` and check the return value. If it failed to append, which means that there is probably no heap available, the return `SYSTEM_ERROR_NO_MEMORY`

system/src/system_control_internal.cpp
```diff
@@ -389,6 +399,19 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
     particle::system::SystemControl::instance()->setResult(req, result, handler, data);
 }
 
+int system_set_control_request_filter(uint16_t reqType, void* reserved) {
```
XuGuohui, line 402: Let's return `ssize_t` to indicate error.

system/src/system_control_internal.h
```diff
@@ -51,12 +51,11 @@ class SystemControl: public ControlRequestHandler {
 
     static SystemControl* instance();
 
+    static Vector<uint16_t> vecCtrlReq;
```
XuGuohui, line 57: It doesn't have to be static, since it is singleton. Nit: `vecCtrlReq_`

XuGuohui (changes requested):

system/src/ble_listening_mode_handler.cpp
```diff
@@ -72,15 +95,37 @@ int BleListeningModeHandler::constructControlRequestAdvData() {
...
+        CHECK_TRUE(tempAdvData.append(sizeof(code) + 1), SYSTEM_ERROR_NO_MEMORY);
+        CHECK_TRUE(tempAdvData.append(BLE_SIG_AD_TYPE_MANUFACTURER_SPECIFIC_DATA), SYSTEM_ERROR_NO_MEMORY);
+        CHECK_TRUE(tempAdvData.append((uint8_t*)&code, 6), SYSTEM_ERROR_NO_MEMORY);
+    }
 
     // Particle Control Request Service 128-bits UUID
-    CHECK_TRUE(tempSrData.append(sizeof(BLE_CTRL_REQ_SVC_UUID) + 1), SYSTEM_ERROR_NO_MEMORY);
-    CHECK_TRUE(tempSrData.append(BLE_SIG_AD_TYPE_128BIT_SERVICE_UUID_COMPLETE), SYSTEM_ERROR_NO_MEMORY);
-    CHECK_TRUE(tempSrData.append(BLE_CTRL_REQ_SVC_UUID, sizeof(BLE_CTRL_REQ_SVC_UUID)), SYSTEM_ERROR_NO_MEMORY);
+    if (provMode_ && memcmp(provBleCtrlReqSvcUuid_, zeros, sizeof(provBleCtrlReqSvcUuid_)) != 0) {
```
XuGuohui, line 115: Introduce an API in `BleControlRequestChannel` to get the current provisioning service UUID, despite of `provMode_`.

system/src/ble_listening_mode_handler.h
```diff
@@ -68,6 +76,9 @@ class BleListeningModeHandler {
     bool userAdv_;
     bool restoreUserConfig_;
     static bool exited_;
+    static bool provMode_;
```
XuGuohui, line 79: It doesn't have to be static, since it is singleton.

system/inc/system_control.h
```diff
@@ -161,6 +161,12 @@ void system_ctrl_free_request_data(ctrl_request* req, void* reserved);
 // while sending the reply
 void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handler_fn handler, void* data, void* reserved);
 
+// The control requests in the list are filtered out from executing
```
XuGuohui, line 164: not in the list

system/src/system_control_internal.cpp
```diff
@@ -81,30 +81,48 @@ SystemControl g_systemControl;
...
+    BleControlRequestChannel::instance(this)->run();
 #endif
 }
 
 void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
+    // All requests are allowed if vector is empty
+    // If vector is non-empty, only those requests in the vector are allowed. Others are filtered out
+    if (!vecCtrlReq_.isEmpty()) {
+        bool reqFound = false;
+        for (auto i=0; i<vecCtrlReq_.size() && !reqFound; i++) {
```
XuGuohui, line 116: Nit:
```
for (auto& type : vecCtrlReq_) {
    if (type == req->type) {
        reqFound = true;
        break;
    }
}
```

system/src/system_control_internal.cpp
```diff
@@ -389,6 +407,28 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
     particle::system::SystemControl::instance()->setResult(req, result, handler, data);
 }
 
+int system_ctrl_set_request_filter(uint16_t reqType, void* reserved) {
```
XuGuohui, line 410: `system_ctrl_add_request_filter` is more accurate. "set" sounds like we're going to remove existing elements. Alternatively, pass in an array and element count as its argument to set the filter for once.

wiring/inc/spark_wiring_ble.h
```diff
@@ -927,6 +928,26 @@ class BleLocalDevice {
...
+    template<typename T1, typename T2, typename T3>
+    int setProvisioningUuids(T1 svcUuid, T2 txUuid, T3 rxUuid) const {
+        BleUuid tempSvcUUID(svcUuid);
+        BleUuid tempTxUUID(txUuid);
+        BleUuid tempRxUUID(rxUuid);
+        return system_set_prov_svc_uuid(tempSvcUUID.halUUID(), tempTxUUID.halUUID(), tempRxUUID.halUUID(), nullptr);
+    }
+
+    template<typename T>
+    int setProvisioningAdvServiceUuid(T uuid) const {
```
XuGuohui, line 944: We don't need this API. System layer can fetch the provisioning service UUID from the BLE control request class.

XuGuohui (changes requested):

system/src/ble_listening_mode_handler.cpp
```diff
@@ -72,15 +95,37 @@ int BleListeningModeHandler::constructControlRequestAdvData() {
...
+        CHECK_TRUE(tempAdvData.append((uint8_t*)&platformID, sizeof(platformID)), SYSTEM_ERROR_NO_MEMORY);
+    } else {
+        LOG_DEBUG(TRACE, "Using setup code");
+        char code[SETUP_CODE_SIZE] = {};
+        CHECK(get_device_setup_code(code));
+        // FIX the below
+        CHECK_TRUE(tempAdvData.append(sizeof(code) + 1), SYSTEM_ERROR_NO_MEMORY);
+        CHECK_TRUE(tempAdvData.append(BLE_SIG_AD_TYPE_MANUFACTURER_SPECIFIC_DATA), SYSTEM_ERROR_NO_MEMORY);
+        CHECK_TRUE(tempAdvData.append((uint8_t*)&code, 6), SYSTEM_ERROR_NO_MEMORY);
+    }
```
monkbroc, line 112: @keeramis let's simplify this to always include the same information in the manufacturer data in listening mode and provisioning mode. There's no strong reason to have it be different and it means that an app that can work with provisioning mode would also work with listening mode and vice-versa.

```suggestion
    char code[SETUP_CODE_SIZE] = {};
    CHECK(get_device_setup_code(code));
    CHECK_TRUE(tempAdvData.append(sizeof(platformID) + sizeof(companyID) + sizeof(code) + 1), SYSTEM_ERROR_NO_MEMORY);
    CHECK_TRUE(tempAdvData.append(BLE_SIG_AD_TYPE_MANUFACTURER_SPECIFIC_DATA), SYSTEM_ERROR_NO_MEMORY);
    CHECK_TRUE(tempAdvData.append((uint8_t*)&companyID, 2), SYSTEM_ERROR_NO_MEMORY);
    CHECK_TRUE(tempAdvData.append((uint8_t*)&platformID, sizeof(platformID)), SYSTEM_ERROR_NO_MEMORY);
    CHECK_TRUE(tempAdvData.append((uint8_t*)&code, 6), SYSTEM_ERROR_NO_MEMORY);
```

hal/inc/device_code.h
```diff
@@ -32,6 +33,9 @@ typedef struct device_code_t {
     uint8_t value[32];
 } device_code_t;
 
+// XXX: This constant is obtained from dct.h
+#define SETUP_CODE_SIZE 6
```
avtolstoy, line 37: Nit: let's prefix with `HAL_`

hal/src/nRF52840/device_code.cpp
```diff
@@ -38,29 +38,28 @@ const auto DEVICE_NAME_DCT_OFFSET = DCT_SSID_PREFIX_OFFSET;
...
 const auto SETUP_CODE_DCT_OFFSET = DCT_DEVICE_CODE_OFFSET;
 
-int getDeviceSetupCode(char* code) {
+} // ::
+
+int get_device_setup_code(char* code, size_t size) {
     // Check if the device setup code is initialized in the DCT
-    int ret = dct_read_app_data_copy(SETUP_CODE_DCT_OFFSET, code, SETUP_CODE_SIZE);
-    if (ret < 0 || !isPrintable(code, SETUP_CODE_SIZE)) {
+    int ret = dct_read_app_data_copy(SETUP_CODE_DCT_OFFSET, code, size);
```
avtolstoy, line 47: Why the change to use `size` instead of fixed constant? If buffer provided is larger than `SETUP_CODE_SIZE` we'll read garbage data. The existing code was saner.

hal/src/nRF52840/device_code.cpp
```diff
@@ -85,7 +84,7 @@ int get_device_name(char* buf, size_t size) {
     if (nameSize == 0 || nameSize > DEVICE_NAME_MAX_SIZE || !isPrintable(name, nameSize)) {
         // Get device setup code
         char code[SETUP_CODE_SIZE] = {};
-        ret = getDeviceSetupCode(code);
+        ret = get_device_setup_code(code, SETUP_CODE_SIZE);
```
avtolstoy, line 87: Nit: prefer `sizeof(code)`. When refactoring it's easy to forget to replace the size in both places: declaration and function call.

hal/src/nRF52840/deviceid_hal.cpp
```diff
@@ -137,5 +137,23 @@ int hal_get_device_secret(char* data, size_t size, void* reserved)
     }
     return HAL_DEVICE_SECRET_SIZE;
 }
+
+int hal_set_device_secret(char* data, size_t size, void* reserved) {
+    if (!data || size != HAL_DEVICE_SECRET_SIZE || !isPrintable(data, HAL_DEVICE_SECRET_SIZE)) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    int ret = dct_write_app_data(data, DCT_DEVICE_SECRET_OFFSET, size);
```
avtolstoy, line 145: `size` needs to be trimmed to `HAL_DEVICE_SECRET_SIZE`

system/inc/system_ble_prov.h
```diff
@@ -0,0 +1,23 @@
...
+#include "ble_hal.h"
+
+#if HAL_PLATFORM_BLE
+
+int system_ble_prov_mode(bool enabled, void* reserved);
+bool system_get_ble_prov_status(void* reserved);
+int system_set_custom_prov_svc_uuid(hal_ble_uuid_t svcUuid, void* reserved);
+int system_set_custom_prov_tx_uuid(hal_ble_uuid_t txUuid, void* reserved);
+int system_set_custom_prov_rx_uuid(hal_ble_uuid_t rxUuid, void* reserved);
+int system_set_custom_prov_ver_uuid(hal_ble_uuid_t verUuid, void* reserved);
```
avtolstoy, line 17: Pass `hal_ble_uuid_t` by constant pointer for ABI stability.

system/inc/system_ble_prov.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+#include "ble_hal.h"
+
+#if HAL_PLATFORM_BLE
+
+int system_ble_prov_mode(bool enabled, void* reserved);
```
avtolstoy, line 12: Nit: let's make sure all of these have a common prefix e.g. `system_ble_prov_xxx`

system/src/ble_control_request_channel.cpp
```diff
@@ -92,18 +95,6 @@ struct __attribute__((packed)) HandshakeHeader {
 // Device setup protocol version
 const unsigned PROTOCOL_VERSION = 0x02;
 
-// UUID of the control request service
-const uint8_t CTRL_SERVICE_UUID[] = { 0xfc, 0x36, 0x6f, 0x54, 0x30, 0x80, 0xf4, 0x94, 0xa8, 0x48, 0x4e, 0x5c, 0x01, 0x00, 0xa9, 0x6f };
```
avtolstoy, line 96: Let's keep this as-is. This looks clearer with the comments and named constants. `hal_ble_uuid` can be initialized from this data, or even higher level `BleUuid` can be used to simplify things.

system/src/ble_control_request_channel.cpp
```diff
@@ -560,6 +551,11 @@ BleControlRequestChannel::BleControlRequestChannel(ControlRequestHandler* handle
         maxPacketSize_(0),
         subscribed_(false),
         writable_(false),
+        initialized_(false),
+        bleCtrlSvcUuid_{.uuid128 = { 0xfc, 0x36, 0x6f, 0x54, 0x30, 0x80, 0xf4, 0x94, 0xa8, 0x48, 0x4e, 0x5c, 0x01, 0x00, 0xa9, 0x6f },.type=BLE_UUID_TYPE_128BIT, 0},
```
avtolstoy, line 555: See comment above. This does not look readable now IMO.

system/src/ble_control_request_channel.cpp
```diff
@@ -1029,7 +1053,7 @@ int BleControlRequestChannel::initProfile() {
     recvCharHandle_ = attrHandles.value_handle;
 
     CHECK(hal_ble_set_callback_on_periph_link_events(onBleLinkEvents, this, nullptr));
-
+    initialized_ = true;
```
avtolstoy, line 1056: Should this be reset somewhere e.g. in `destroy()`?

system/src/ble_control_request_channel.h
```diff
@@ -123,6 +131,13 @@ class BleControlRequestChannel: public ControlRequestChannel {
     volatile bool subscribed_; // Set to `true` if the client is subscribed to the notifications
     volatile bool writable_; // Set to `true` if the TX characteristic is writable
 
+    bool initialized_;
+
+    hal_ble_uuid_t bleCtrlSvcUuid_{};
```
avtolstoy, line 136: Keep initializations uniform. Either in-class initialization or initialization in constructor initializer list.

system/src/ble_provisioning_mode_handler.cpp
```diff
@@ -60,8 +77,8 @@ int BleListeningModeHandler::constructControlRequestAdvData() {
     CHECK_TRUE(tempAdvData.append(BLE_SIG_AD_TYPE_FLAGS), SYSTEM_ERROR_NO_MEMORY);
     CHECK_TRUE(tempAdvData.append(BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE), SYSTEM_ERROR_NO_MEMORY);
 
-    char devName[32] = {};
-    CHECK(get_device_name(devName, sizeof(devName)));
+    char devName[BLE_MAX_DEV_NAME_LEN] = {};
+    CHECK(hal_ble_gap_get_device_name(devName, BLE_MAX_DEV_NAME_LEN, nullptr));
```
avtolstoy, line 81: Nit: prefer `sizeof(devName)`.

system/src/ble_provisioning_mode_handler.cpp
```diff
@@ -60,8 +77,8 @@ int BleListeningModeHandler::constructControlRequestAdvData() {
     CHECK_TRUE(tempAdvData.append(BLE_SIG_AD_TYPE_FLAGS), SYSTEM_ERROR_NO_MEMORY);
     CHECK_TRUE(tempAdvData.append(BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE), SYSTEM_ERROR_NO_MEMORY);
 
-    char devName[32] = {};
-    CHECK(get_device_name(devName, sizeof(devName)));
+    char devName[BLE_MAX_DEV_NAME_LEN] = {};
```
avtolstoy, line 80: Should probably be `BLE_MAX_DEV_NAME_LEN + 1`. Additional care probably needs to be taken that it's null-terminated (in case returned name for some reason is > buffer size), see https://github.com/particle-iot/device-os/pull/2382/files#diff-522277e93ac030d207929e5d20501837ba30f5d94daaf8759714f2910cec970dR84

system/src/ble_provisioning_mode_handler.cpp
```diff
@@ -72,22 +89,33 @@ int BleListeningModeHandler::constructControlRequestAdvData() {
     uint16_t companyID = PARTICLE_COMPANY_ID;
 
     // Manufacturing specific data
-    CHECK_TRUE(tempAdvData.append(sizeof(platformID) + sizeof(companyID) + 1), SYSTEM_ERROR_NO_MEMORY);
+    char code[SETUP_CODE_SIZE] = {};
+    CHECK(get_device_setup_code(code, SETUP_CODE_SIZE));
```
avtolstoy, line 93: Nit: prefer `sizeof(code)`.

system/src/control/wifi_new.cpp
```diff
@@ -115,6 +116,8 @@ int joinNewNetwork(ctrl_request* req) {
         network_disconnect(NETWORK_INTERFACE_WIFI_STA, NETWORK_DISCONNECT_REASON_USER, nullptr);
     });
     CHECK(wifiMgr->connect(dSsid.data));
+    // TODO: Not adding NetworkCredentials for now as this object needs to be allocated on heap and then cleaned up
+    system_notify_event(network_credentials, network_credentials_added);
```
avtolstoy, line 120: :+1:

wiring/inc/spark_wiring_system.h
```diff
@@ -764,6 +772,16 @@ class SystemClass {
 
 #endif // HAL_PLATFORM_POWER_MANAGEMENT
 
+    int setControlRequestFilter(system_control_req_filter* filter) {
```
avtolstoy, line 775: This needs to take a wrapper around `system_control_req_filter`. A suggestion is to take a similar approach to sleep API for example with the class that allows chaining `.call(123).call(345)` etc.

system/src/ble_control_request_channel.cpp
```diff
@@ -603,12 +604,14 @@ void BleControlRequestChannel::run() {
         resetChannel();
```
avtolstoy, line 604: This method should probably check for initialization state and return early if not initialized.

system/src/system_control_internal.cpp
```diff
@@ -389,6 +408,32 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
     particle::system::SystemControl::instance()->setResult(req, result, handler, data);
 }
 
+int system_ctrl_add_request_filter(uint16_t* reqItems, size_t cnt, void* reserved) {
+    // check if the element is already there. Check if the element is a valid element
+    for (unsigned i=0; i<cnt; i++) {
+        bool eleExists = false;
+        for (auto t : particle::system::SystemControl::instance()->vecCtrlReq_) {
```
avtolstoy, line 415: `vecCtrlReq_.contains()`. Let's also not access the vector directly from here and delegate to a class method instead.

system/src/system_control_internal.cpp
```diff
@@ -389,6 +408,32 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
...
+        if (!ret) {
+            return SYSTEM_ERROR_NO_MEMORY;
+        }
+        LOG(TRACE, "Added a new req to control req filter: %u", reqItems[i]);
+    }
+    return particle::system::SystemControl::instance()->vecCtrlReq_.size();
+}
+
+int system_ctrl_clear_request_filter(void* reserved) {
+    particle::system::SystemControl::instance()->vecCtrlReq_.clear();
```
avtolstoy, line 433: Delegate to class method.

system/src/system_control_internal.cpp
```diff
@@ -81,30 +81,49 @@ SystemControl g_systemControl;
...
+    BleControlRequestChannel::instance(this)->run();
 #endif
 }
 
 void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
+    // All requests are allowed if vector is empty
+    // If vector is non-empty, only those requests in the vector are allowed. Others are filtered out
+    if (!vecCtrlReq_.isEmpty()) {
+        bool reqFound = false;
+        for (auto& type : vecCtrlReq_) {
```
avtolstoy, line 116: `vecCtrlReq_.contains()`.

system/src/system_control_internal.cpp
```diff
@@ -81,30 +81,49 @@ SystemControl g_systemControl;
...
 #if HAL_PLATFORM_BLE
-    bleChannel_.run();
+    BleControlRequestChannel::instance(this)->run();
 #endif
 }
 
 void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
+    // All requests are allowed if vector is empty
+    // If vector is non-empty, only those requests in the vector are allowed. Others are filtered out
+    if (!vecCtrlReq_.isEmpty()) {
```
avtolstoy, line 114: There should be an easy way to disable all requests. A special request type (SOMETHING_SOMETHING_ANY) can be used.

avtolstoy (changes requested):

wiring/inc/spark_wiring_system.h
avtolstoy: + some additional overloads that allow to pass an array/vector of ids.

avtolstoy: I gave this a bit more thought and now the API from a first glance looks a bit confusing. Does the filter specify enabled or disabled requests? I think we should make this more explicit. Let's do something along the lines of :
```c++
enum class SystemControlRequestAclAction {
    NONE = 0,
    ACCEPT = 1,
    DENY = 2
};
class SystemControlRequestAcl {
public:
    SystemControlRequestAcl(ctrl_request_type type, SystemControlRequestAclAction action);

    ctrl_request_type type() const;
    SystemControlRequestAclAction action() const;

private:
    ctrl_request_type type_;
    SystemControlRequestAclAction action_;
};

int setControlRequestFilter(SystemControlRequestAclAction defaultAction, Vector<SystemControlRequestAcl> filters);
```

With this you can do the following:
```c++
// Accept by default, deny some
System.setControlRequestFilter(SystemControlRequestAclAction::ACCEPT, { {CTRL_REQUEST_RESET, SystemControlRequestAclAction::DENY} });

// Deny by default, accept some
System.setControlRequestFilter(SystemControlRequestAclAction::DENY, { {CTRL_REQUEST_RESET, SystemControlRequestAclAction::ACCEPT} });
```

The appropriate system C APIs/structs will need to be added as well.

avtolstoy: @keeramis

monkbroc:
The sample code generates a warning. It's not a big deal here but for future sample code, use another variable name like `logHandler` to avoid this warning.
```
 warning: built-in function 'log' declared as non-function [-Wbuiltin-declaration-mismatch]
    3 | SerialLogHandler log(115200, LOG_LEVEL_ALL);
```

system/src/control/wifi_new.cpp
```diff
@@ -106,15 +108,29 @@ int joinNewNetwork(ctrl_request* req) {
...
+            network_ready(NETWORK_INTERFACE_WIFI_STA, NETWORK_READY_TYPE_ANY, nullptr);
+    // To unblock
+    ncpClient->disable();
+    CHECK(ncpClient->enable());
+    // These two are in sync now
+    ncpClient->disconnect(); // ignore the error
     network_disconnect(NETWORK_INTERFACE_WIFI_STA, NETWORK_DISCONNECT_REASON_USER, nullptr);
-    CHECK(ncpClient->disconnect());
+    // FIXME: We are wiating for ncpNetif to potentially fully disconnect
+    HAL_Delay_Milliseconds(200);
```
avtolstoy, line 120: As we discussed, delay here should not be needed due to `NcpClientLock` and `processEvents()` call that requires a lock in esp32ncpnetif.

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2070,6 +2074,14 @@ int BleLocalDevice::selectAntenna(BleAntennaType antenna) const {
     return hal_ble_select_antenna(static_cast<hal_ble_ant_type_t>(antenna), nullptr);
 }
 
+int BleLocalDevice::provisioningMode(bool enabled) const{
```
avtolstoy, line 2077: Nit/style: space

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2070,6 +2074,14 @@ int BleLocalDevice::selectAntenna(BleAntennaType antenna) const {
...
+    return system_ble_prov_mode(enabled, nullptr);
+}
+
+bool BleLocalDevice::getProvisioningStatus() const{
```
avtolstoy, line 2081: Nit/style: space

wiring/inc/spark_wiring_system.h
```diff
@@ -893,6 +953,7 @@ class SystemClass {
 };
 
 extern SystemClass System;
+extern SystemControlRequestAclClass SystemControlRequestAcl;
```
avtolstoy, line 956: This is not needed.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+
+        for(int i=0; i<filters.size(); i++) {
+            if (first == nullptr) {
+                first = (system_control_filter*)malloc(sizeof(system_control_filter));
+                if (!first) {
+                    return SYSTEM_ERROR_NO_MEMORY;
+                }
+                first->req = filters[i].type();
+                first->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
+                LOG(TRACE, "TP1 - action: %d", first->action);
```
avtolstoy, line 892: Extra log.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                first->next = nullptr;
+                current = first;
+            } else {
+                system_control_filter* temp = (system_control_filter*)malloc(sizeof(system_control_filter));
+                current->next = temp;
+                temp->req = filters[i].type();
+                temp->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
+                LOG(TRACE, "TP2 - action: %d", temp->action);
```
avtolstoy, line 900: Extra log.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
 
 #endif // HAL_PLATFORM_POWER_MANAGEMENT
 
+    int setControlRequestFilter(SystemControlRequestAclAction defaultAction, Vector<SystemControlRequestAclClass> filters = {}) {
+		// Add requests and their actions to the control filter
+        system_control_acl default_action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(defaultAction);
```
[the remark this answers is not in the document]
avtolstoy: `SystemControlRequestAcl` class itself can also perform conversion to `system_control_acl`.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+        system_control_filter* first = nullptr;
+        system_control_filter* current = nullptr;
+
+        for(int i=0; i<filters.size(); i++) {
```
avtolstoy, line 884: Nit: spaces around operators.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                current->next = temp;
+                temp->req = filters[i].type();
+                temp->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
+                LOG(TRACE, "TP2 - action: %d", temp->action);
+                temp->next = nullptr;
+                current = temp;
+            }
+        }
+
+        return system_ctrl_add_request_filter(default_action, first, nullptr);
```
avtolstoy, line 906: Memory allocated for list of filters is never freed. As proposed in `system_ctrl_add_request_filter`, take ownership of the linked list of filters in the SystemControl. Here, free memory only on errors.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                if (!first) {
+                    return SYSTEM_ERROR_NO_MEMORY;
+                }
+                first->req = filters[i].type();
+                first->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
+                LOG(TRACE, "TP1 - action: %d", first->action);
+                first->next = nullptr;
+                current = first;
+            } else {
+                system_control_filter* temp = (system_control_filter*)malloc(sizeof(system_control_filter));
```
avtolstoy, line 896: allocation failure is not handled.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+        system_control_filter* first = nullptr;
+        system_control_filter* current = nullptr;
+
+        for(int i=0; i<filters.size(); i++) {
+            if (first == nullptr) {
+                first = (system_control_filter*)malloc(sizeof(system_control_filter));
+                if (!first) {
+                    return SYSTEM_ERROR_NO_MEMORY;
+                }
+                first->req = filters[i].type();
```
avtolstoy, line 890: `size` is not set.

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                first->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
```
avtolstoy, line 891: `enumclass.h`

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                LOG(TRACE, "TP1 - action: %d", first->action);
+                first->next = nullptr;
+                current = first;
+            } else {
+                system_control_filter* temp = (system_control_filter*)malloc(sizeof(system_control_filter));
+                current->next = temp;
+                temp->req = filters[i].type();
```
avtolstoy, line 898: `size` is not set

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +875,37 @@ class SystemClass {
...
+                temp->action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(filters[i].action());
```
avtolstoy, line 899: `enumclass.h`

system/src/system_control_internal.cpp
```diff
@@ -389,6 +415,44 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
     particle::system::SystemControl::instance()->setResult(req, result, handler, data);
 }
 
+int system_ctrl_add_request_filter(system_control_acl default_action, system_control_filter* filters, void* reserved) {
+    if (default_action == SYSTEM_CONTROL_ACL_NONE) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    particle::system::SystemControl::instance()->filterCtrlReq_.vecCtrlReq_.clear();
```
avtolstoy, line 422: I think we've discussed the same thing in a different context: functions should not access internal members of classes. Add an accessor method to the class instead.

system/src/system_control_internal.cpp
```diff
@@ -389,6 +415,44 @@ void system_ctrl_set_result(ctrl_request* req, int result, ctrl_completion_handl
...
+
+    // If default action is to accept requests, vecCtrlReq_ will contain requests that need to be rejected
+    if (default_action == SYSTEM_CONTROL_ACL_ACCEPT) {
```
avtolstoy, line 425: No need for such complexity. To simplify just pass default action and `filters` as-is to the class and set two internal member variables: `defaultAction_` and keep `filters` struct as-is. When filtering check the default action and traverse the list, no need for an extra Vector. When replacing the list of filters make sure to free the memory of the list members.

system/src/system_ble_prov.cpp
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+    return SYSTEM_ERROR_NONE;
+}
+
+bool system_ble_prov_get_status(void* reserved) {
+    return BleProvisioningModeHandler::instance()->getProvModeStatus();
+}
+
+int system_ble_prov_set_custom_svc_uuid(hal_ble_uuid_t* svcUuid, void* reserved) {
```
avtolstoy, line 62: passed `hal_ble_uuid_t*` being `nullptr` is not handled. There will be a null pointer dereference in `setProvSvcUuid` and others.

avtolstoy (changes requested):

system/src/system_ble_prov.cpp
```diff
@@ -0,0 +1,99 @@
+#include "ble_provisioning_mode_handler.h"
```
XuGuohui, line 1: Let‘s add copyright here.

system/src/system_control_internal.cpp
avtolstoy: @XuGuohui See https://github.com/particle-iot/device-os/pull/2382#discussion_r806670907

hal/src/nRF52840/deviceid_hal.cpp
```diff
@@ -137,5 +137,19 @@ int hal_get_device_secret(char* data, size_t size, void* reserved)
     }
     return HAL_DEVICE_SECRET_SIZE;
 }
+
+int hal_set_device_secret(char* data, size_t size, void* reserved) {
+    if (!data || size != HAL_DEVICE_SECRET_SIZE || !isPrintable(data, HAL_DEVICE_SECRET_SIZE)) {
```
XuGuohui, line 142: As per Andrey's comment, we could check `data` and when it is `nullptr` we then write zeros to DCT to clear the mobile secret.

hal/src/nRF52840/device_code.cpp
```diff
@@ -38,29 +38,31 @@ const auto DEVICE_NAME_DCT_OFFSET = DCT_SSID_PREFIX_OFFSET;
...
+    int ret = dct_read_app_data_copy(SETUP_CODE_DCT_OFFSET, setupCode, sizeof(setupCode));
+    if (ret < 0 || !particle::isPrintable(setupCode, sizeof(setupCode))) {
         // Check the OTP memory
         char serial[HAL_DEVICE_SERIAL_NUMBER_SIZE] = {};
         ret = hal_get_device_serial_number(serial, sizeof(serial), nullptr);
-        if (ret >= (int)SETUP_CODE_SIZE && (size_t)ret <= sizeof(serial)) {
+        if ((size_t)ret >= sizeof(setupCode) && (size_t)ret <= sizeof(serial)) {
             // Use last characters of the serial number as the setup code
-            memcpy(code, &serial[ret - SETUP_CODE_SIZE], SETUP_CODE_SIZE);
+            memcpy(code, &serial[ret - size], size);
```
XuGuohui, line 55: before copying, `size = std::min(size, HAL_SETUP_CODE_SIZE)`

system/inc/system_ble_prov.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2018 Particle Industries, Inc.  All rights reserved.
```
XuGuohui, line 2: `Copyright (c) 2022`

system/src/ble_control_request_channel.cpp
```diff
@@ -659,6 +689,52 @@ void BleControlRequestChannel::run() {
     resetChannel();
 }
 
+int BleControlRequestChannel::setProvSvcUuid(hal_ble_uuid_t* svcUuid) {
+    if (!svcUuid) {
```
XuGuohui, line 693: Prefer to use `CHECK_TRUE()`.

system/src/ble_control_request_channel.cpp
```diff
@@ -659,6 +689,52 @@ void BleControlRequestChannel::run() {
...
+    bleVerCharUuid_ = *verUuid;
+    return SYSTEM_ERROR_NONE;
+}
+
+hal_ble_uuid_t BleControlRequestChannel::getBleCtrlSvcUuid() {
+    return bleCtrlSvcUuid_;
+}
+
+int BleControlRequestChannel::setCompanyId(uint16_t companyId) {
+    // TODO: Add error checking for companyId
```
XuGuohui, line 729: This API should belong to the `BleProvisioningModeHandler` class, as the BLE control request doesn't care what the company ID is, it is used in advertising data only. We probably don't need to add any check here. It's all up to who invokes this API.

system/src/system_ble_prov.cpp
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright (c) 2018 Particle Industries, Inc.  All rights reserved.
```
XuGuohui, line 2: `Copyright (c) 2022`

XuGuohui (changes requested):

system/src/ble_control_request_channel.h
```diff
@@ -64,6 +64,16 @@ class BleControlRequestChannel: public ControlRequestChannel {
...
     virtual void setResult(ctrl_request* ctrlReq, int result, ctrl_completion_handler_fn handler, void* data) override;
 
+    int setProvSvcUuid(hal_ble_uuid_t* svcUuid);
+    int setProvTxUuid(hal_ble_uuid_t* txUuid);
+    int setProvRxUuid(hal_ble_uuid_t* rxUuid);
+    int setProvVerUuid(hal_ble_uuid_t* verUuid);
+
+    hal_ble_uuid_t getBleCtrlSvcUuid();
+
+    int setCompanyId(uint16_t companyId);
```
XuGuohui, line 74: The declaration of `setCompanyId()` and `getCompanyId()` in this class are no more needed.

XuGuohui (changes requested):

XuGuohui (approved):

wiring/inc/spark_wiring_system.h
```diff
@@ -846,6 +877,50 @@ class SystemClass {
...
 #endif // HAL_PLATFORM_POWER_MANAGEMENT
 
+    int setControlRequestFilter(SystemControlRequestAclAction defaultAction, Vector<SystemControlRequestAcl> filters = {}) {
+		// Add requests and their actions to the control filter
+        system_control_acl default_action = (system_control_acl)static_cast<typename std::underlying_type<SystemControlRequestAclAction>::type>(defaultAction);
+        system_control_filter* first = nullptr;
+        system_control_filter* current = nullptr;
+
+        for(int i=0; i < filters.size(); i++) {
+            if (first == nullptr) {
```
avtolstoy, line 887: Let's try to get rid of code duplication here.

```c++
NAMED_SCOPE_GUARD(cleanup, {
    while (first) {
        auto tmp = first->next;
        free(first);
        first = tmp;
    }
});
for (const auto& filter: filters) {
    auto f = (system_control_filter*)malloc(sizeof(system_control_filter));
    if (!f) {
         return SYSTEM_ERROR_NOT_MEMORY;
    }
    memset(f, 0, sizeof(system_control_filter));
    f->size = sizeof(system_control_filter);
    f->req = filter.type();
    f->action = to_underlying(filter.action());
    if (!current) {
        first = current = f;
    } else {
        current->next = f;
        current = f;
    }
}

auto ret = system_ctrl_set_request_filter(default_action, first, nullptr);
if (!ret) {
     cleanup.dismiss();
}
return ret;
```

avtolstoy (changes requested):

hal/src/nRF52840/device_code.cpp
```diff
@@ -38,29 +38,37 @@ const auto DEVICE_NAME_DCT_OFFSET = DCT_SSID_PREFIX_OFFSET;
...
-    int ret = dct_read_app_data_copy(SETUP_CODE_DCT_OFFSET, code, SETUP_CODE_SIZE);
-    if (ret < 0 || !isPrintable(code, SETUP_CODE_SIZE)) {
+    char setupCode[HAL_SETUP_CODE_SIZE] = {};
+    int ret = dct_read_app_data_copy(SETUP_CODE_DCT_OFFSET, setupCode, sizeof(setupCode));
+    if (ret < 0 || !particle::isPrintable(setupCode, sizeof(setupCode))) {
         // Check the OTP memory
         char serial[HAL_DEVICE_SERIAL_NUMBER_SIZE] = {};
         ret = hal_get_device_serial_number(serial, sizeof(serial), nullptr);
-        if (ret >= (int)SETUP_CODE_SIZE && (size_t)ret <= sizeof(serial)) {
+        if ((size_t)ret >= sizeof(setupCode) && (size_t)ret <= sizeof(serial)) {
```
avtolstoy, line 53: We can't cast `int` to `size_t` before we check whether it's positive or not. Add `ret > 0` check first into the `if`.

hal/src/nRF52840/device_code.cpp
```diff
@@ -38,29 +38,37 @@ const auto DEVICE_NAME_DCT_OFFSET = DCT_SSID_PREFIX_OFFSET;
...
             // Use last characters of the serial number as the setup code
-            memcpy(code, &serial[ret - SETUP_CODE_SIZE], SETUP_CODE_SIZE);
+            size = std::min(size, (size_t)HAL_SETUP_CODE_SIZE);
```
avtolstoy, line 55: Probably better use `std::min(size, (size_t)ret)`

hal/src/nRF52840/device_code.cpp
```diff
@@ -38,29 +38,37 @@ const auto DEVICE_NAME_DCT_OFFSET = DCT_SSID_PREFIX_OFFSET;
...
-            memset(code, 'X', SETUP_CODE_SIZE);
+            size = std::min(sizeof(setupCode), size);
+            memset(code, 'X', size);
         }
+    } else {
+        size = std::min(sizeof(setupCode), size);
+        memcpy(code, setupCode, size);
     }
-    return 0;
+    if (size < sizeof(setupCode)) {
```
avtolstoy, line 66: Use a different variable to store the actual resulting size and keep `size` argument intact. Use that to infer whether we can null-terminate or not.

system/src/system_control_internal.cpp
```diff
@@ -105,6 +112,23 @@ void SystemControl::run() {
...
 
 void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* channel */) {
+
+    // Filter requests based on the setControlRequestFilter system API
+    system_ctrl_filter* filter = getFiltersList();
+
+    while(filter != nullptr) {
+        if (filter->req == req->type) {
+            break;
+        }
```
avtolstoy, line 122: `filter = filter->next` :)

hal/src/nRF52840/device_code.cpp
```diff
@@ -45,28 +45,30 @@ const auto SETUP_CODE_DCT_OFFSET = DCT_DEVICE_CODE_OFFSET;
...
     } else {
-        size = std::min(sizeof(setupCode), size);
-        memcpy(code, setupCode, size);
+        codeSize = std::min(sizeof(setupCode), size);
+        memcpy(code, setupCode, codeSize);
     }
-    if (size < sizeof(setupCode)) {
-        code[size] = '\0';
+    // Null terminate if size of the code is less than the size allocated by DCT
+    if (codeSize < sizeof(setupCode)) {
```
avtolstoy, line 68:
```c++
if (codeSize < size) {
    code[codeSize] = '\0';
}
```

 right?

avtolstoy (approved):

XuGuohui (approved):

user/tests/wiring/no_fixture/ble_provisioning.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "application.h"
+#include "unit-test/unit-test.h"
+
+#if HAL_PLATFORM_BLE
+
+SYSTEM_THREAD(ENABLED);
```
avtolstoy, line 23: This should be removed. Tests can be built with or without threading and should function in both modes.

user/tests/wiring/no_fixture/ble_provisioning.cpp
```diff
@@ -0,0 +1,101 @@
...
+    System.enableFeature(FEATURE_DISABLE_LISTENING_MODE);
+    HAL_Delay_Milliseconds(1500); // Time for system thread to process the flag
+    assertFalse(Network.listening());
+    BLE.provisioningMode(true);
+    HAL_Delay_Milliseconds(100);
+    assertTrue(BLE.getProvisioningStatus());
+    BLE.provisioningMode(false);
+    HAL_Delay_Milliseconds(100);
+    assertFalse(BLE.getProvisioningStatus());
+    System.disableFeature(FEATURE_DISABLE_LISTENING_MODE);
```
avtolstoy, line 98: Call at the beginning of the test with `SCOPE_GUARD`, so that no matter if the test fails we restore the feature flag.

PR #2384, opened 2021-12-22
Panic handler updates

mrlambchop:
Add in the ability to override the panic handler to change the behaviour of assert/panic in the system

<details>
<summary>
<I>
(1) add in dynalib binding for user space apps (unstable API)
(2) refactor some of the panic code that wasn't being used
</I>
</summary>
</details>

### Problem

Applications want to have the ability to override the panic handler such that customer panic handlers can be implemented in user space.

### Solution

Add in panic_hook_set() function and dynalib bindings


services/inc/panic.h
```diff
@@ -1,31 +1,51 @@
...
+    StackOverflow,
+    HeapError,
 } ePanicCode;
-#undef def_panic_codes
 
-void panic_(ePanicCode code, void* extraInfo, void (*HAL_Delay_Microseconds)(uint32_t));
+typedef void (*PanicHook)(const ePanicCode code, const void* extraInfo);
+
+//optional function to set a hook that replaces the core body of the panic function
+void panic_set_override(const PanicHook panicHook);
```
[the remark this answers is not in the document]
avtolstoy: I propose we also wrap this function under `#if !defined(PARTICLE_USER_MODULE) || defined(PARTICLE_USE_UNSTABLE_API)` similar to `storage_hal.h`, to indicate that this is not a stable API and you should explicitly enable it with `PARTICLE_USE_UNSTABLE_API` and unrerstand what you are doing.

avtolstoy (changes requested):
Mainly `panic_set_override` needs to have a `void* reserved` argument to allow for extensibility.

services/inc/logging.h
```diff
@@ -433,7 +431,7 @@ static const char* const _log_category = NULL;
 #define PANIC(_code, _fmt, ...) \
         do { \
             LOG_DEBUG(PANIC, _fmt, ##__VA_ARGS__); \
-            panic_(_code, NULL, HAL_Delay_Microseconds); \
```
avtolstoy, line 436: A second thought/result of private discussion: this needs to be kept as-is for Photon/P1/Electron. The problem is that services dynalib exists in system-part1 and timer_hal implementation in part2. This is the opposite way of the dependency tree, the imports go from part1 into part2 into user app, hence why the pointer to micros function is passed here. Fine on other platforms, but going cause issues for Gen 2.

elcojacobs:
Nice to see this picked up. I would like to see a possibility to store a hardfault address in backup ram so it can be sent to a log and doesn't require a hardware debugger to capture. This however happens before the panic handler is called. Is this possible?

services/src/panic.c
```diff
@@ -16,89 +26,79 @@
...
 
+static void panic_internal(const ePanicCode code, const void* extraInfo);
 
-#define def_panic_codes(_class,led,code) {led, code},
-static const flash_codes_t flash_codes[] = {
-                 {0,0}, //NotUsedPanicCode
-#include "panic_codes.h"
-};
-#undef def_panic_codes
+static PanicHook panicHook = panic_internal;
```
avtolstoy, line 37: This should be `nullptr` after refactoring, otherwise we'll run `panic_internal` twice.

services/inc/services_dynalib.h
```diff
@@ -114,6 +114,7 @@ DYNALIB_FN(BASE_IDX + 0, services, set_system_error_message, void(const char*, .
 DYNALIB_FN(BASE_IDX + 1, services, clear_system_error_message, void())
 DYNALIB_FN(BASE_IDX + 2, services, get_system_error_message, const char*(int))
 DYNALIB_FN(BASE_IDX + 3, services, jsmn_parse, int(jsmn_parser*, const char*, size_t, jsmntok_t*, unsigned int, void*))
+DYNALIB_FN(BASE_IDX + 4, services, panic_set_hook, void(PanicHook, void*))
```
avtolstoy, line 117: Signature does not match what's in `panic.h` (`const` vs non-`const` `PanicHook`)

avtolstoy (changes requested):

avtolstoy (approved):

elcojacobs:
Can you pass the register info extracted in prvGetRegistersFromStack to the panic hook?
That would make it possible to actually get some useful info on the panic location in code, instead of just some different LED behaviour or just logging the panic reason.

Getting a log of the panic reason is already possible through the reset reason that is stored in backup RAM.

elcojacobs:
I am not using the Particle cloud and I don't want to use it for publishing errors either. So please keep the mechanisms open, so I can report the error to a local server too.

perotom:
> Hey elcojacobs, plan is (a) implement the panic handler (b) test using this repo to capture / report bugs (https://github.com/particle-iot/errorflinger-library) (c) add in stack retrieval functions on a crash (d) release this in conjunction with a partnership with memfault.io as a third party crash analysis tool. Feedback welcome on the strategy!

Sounds great. I am with @elcojacobs: would be great if these mechanisms are open for writing these information to SD card or showing them on a display on reboot.

PR #2386, opened 2022-01-03
Remove WiFiTesting mode and Setup over Serial1

mrlambchop:
Removes the following features:

1. Wi-Fi testing mode
2. Setup over serial1

This commit will be proceded by a commit to enable similar features over standard USB commands

### Problem

Dead code used on Photon (no longer supported on head of Device OS) was hanging around, so it got a haircut

### Solution

Quick trim of the code.


wiring/src/spark_wiring_system.cpp
```diff
@@ -101,8 +100,7 @@ bool SystemClass::enableFeature(LoggingFeature) {
 #endif
 
 bool SystemClass::enableFeature(const WiFiTesterFeature feature) {
```
tstellanova, line 102: Can we remove WiFiTesterFeature altogether?

tstellanova (approved):
Looks good, minor question about keeping or removing the WiFiTesterFeature altogether. Haven’t tried building or testing this PR.

PR #2390, opened 2022-02-01
Thread/interrupt safety for system_event_t subscriptions, support for unsubscribing

avtolstoy:
### Problem

1. `std::vector` is used for storing subscriptions in the system layer
2. `system_subscribe_event()` is not thread/interrupt safe and can easily trigger an issue if for example a button interrupt (which is generated directly from an ISR) is generated while the subscriptions list is generated
3. There is no way to disable a subscription (`System.off()` exists, but does not do anything)

### Solution

1. Migrate to `spark::Vector` which does not throw exceptions guaranteed
2. Add thread/interrupt safety whenever accessing or modifying subscription list
3. Implement `system_unsubscribe_event`, add appropriate wiring methods/classes to manage subscriptions for more complex handler types (`SystemEventSubscription`)

### Steps to Test

- `wiring/no_fixture`: SYSTEM_08 SYSTEM_09 tests
- `wiring/api`

### Example App

N/A

### References

- [SC-64736]

---

### Completeness

- [ ] Added documentation


user/tests/wiring/no_fixture/system.cpp
```diff
@@ -299,3 +299,95 @@ test(SYSTEM_07_system_describe_is_not_overflowed_when_factory_module_present_but
...
+    auto handler = [](system_event_t ev, int data, void*) {
+        sLastEvent = ev;
+    };
+    System.on(cloud_status, handler);
+    Particle.disconnect();
+    assertTrue(waitFor(Particle.disconnected, 5000));
+    assertEqual(sLastEvent, (int)cloud_status);
+
+    // System.off(all_events, handler)
+    System.off(cloud_status);
```
technobly, line 348: Should this be `System.off(all_events, handler);` as the comment suggests, since `System.off(cloud_status);` is tested below?

technobly (approved):

PR #2391, opened 2022-02-04
Add GCC feature 'build-id' into the system part1 and user apps for Ge…

mrlambchop:


modules/shared/nRF52840/part1.ld
```diff
@@ -95,6 +95,12 @@ SECTIONS
         /* This is used by the startup in order to initialize the .data secion */
     }>APP_FLASH  AT> APP_FLASH
 
+    .note.gnu.build-id :
+    {
+        __start_gnu_build_id_start = .;
```
avtolstoy, line 100: Same as above comment.

modules/shared/nRF52840/user.ld
```diff
@@ -54,6 +54,12 @@ SECTIONS
 
     }>APP_FLASH  AT> APP_FLASH
 
+    .note.gnu.build-id :
+    {
+        __start_gnu_build_id_start = .;
```
avtolstoy, line 59: Same as above comment.

build/arm/linker/linker_nrf52840_common.ld
```diff
@@ -81,6 +81,11 @@ SECTIONS
         __exidx_end = .;
     } >APP_FLASH
 
+    .note.gnu.build-id :
+    {
+        __start_gnu_build_id_start = .;
```
avtolstoy, line 86: Let's keep the same format with the rest the linker file and name this `link_gnu_build_id_start`.

avtolstoy (changes requested):
@mrlambchop As discussed, let's make sure we rebase on develop before merging and not merge `develop` into the feature branch or merge as-is. This complicates the history a lot.

build/arm/linker/linker_nrf52840_common.ld
mrlambchop: note.gnu.build-id is the output from the built in GCC ld feature so it can't be changed sadly

---

The .note.gnu.build-id section in an [ELF](https://en.wikipedia.org/wiki/Executable_and_Linkable_Format) binary contains a "strongly unique embedded identifier".

[binutils](https://www.gnu.org/software/binutils/)' ld has supported the --build-id=... option since version 2.18 (released 2007). When used, with a sha1 or md5 argument it directs ld to insert an ELF section .note.gnu.build-id into the binary containing a hash of the normative parts of the output—that is, an identifier that uniquely identifies the output file.

Its originally intended purpose (described [here](https://fedoraproject.org/wiki/Releases/FeatureBuildId)) is to simplify and improve debugging tools, but it is occasionally useful for a program to be able to read its own build-id.

PR #2397, opened 2022-02-16
Feature/sc 94643

tstellanova:
Some applications wish to manage their own battery model using the Maxim fuel gauge part included with some device platforms.  In order to allow the application and  DVOS to stay in sync, this PR provides a way for the user app to inform DVOS as to which bits precision (18 or 19) should be used when calculating the State of Charge.  (By default the current FuelGauge implementation uses 18 bits, so unless your app requires 19 bits, there is no reason to use this call.)

This can be used by eg:

```'

void setup() {
    FuelGauge fg(true);
    fg.setCustomSoCPrecision(19);
}

void loop() {
    static uint32_t loop_count = 0;

    FuelGauge fg(true);
    float soc = fg.getSoC();
    float nsoc = fg.getNormalizedSoC();
    Log.info("loop: %lu soc19:  %0.2f nsoc19: %0.2f", loop_count, soc, nsoc);

    delay(3000);
    loop_count++;
}
```

Note that this change is limited in scope to the current Maxim fuel gauge driver, and only applicable on device platforms that use this driver. This is not more broadly applicable to any generalized fuel gauge class we may release in the future.


avtolstoy:
@tstellanova This is not going to work with modular builds. Both system-part and application have its own instance of `FuelGauge` class, so setting the member variable in one will have no affect on the other.

tstellanova:
> @tstellanova This is not going to work with modular builds. Both system-part and application have its own instance of `FuelGauge` class, so setting the member variable in one will have no effect on the other.

I can see from testing with a live device that is doesn't work (vitals value is different from app measured value), but since this is a static class member variable (not an instance variable), I imagine this is due to thread-local storage?

avtolstoy:
@tstellanova Nothing to do with thread local storage. Each binary has its own distinct compiled in variant of `FuelGauge`. `/* user part */ &FuelGauge::soc_bits_ != /* system part */ &FuelGauge::soc_bits_`

tstellanova:
Won’t work with modular builds since each module gets its own copy of FuelGauge class definition.

PR #2398, opened 2022-02-17
[gen3] BLE: introduce wiring APIs to change ATT MTU.

XuGuohui:
### Problem

Currently the desired ATT MTU is set to maximum by default. The ATT MTU exchange procedure is performed automattically once connection is established. But the procedure may failed for some reason and user application has no way to update the ATT MTU to a higher value.

### Solution

1. DO NOT perform the ATT MTU exchange procedure automatically and leaves the user application to perform this procedure when needed.
2. Introduce event callback to notify user application when ATT MTU is updated. This is useful when local device doesn't initiate the ATT MTU exchange procedure, while the peer device does.

### Steps to Test

Build and run the `user/tests/wiring/ble_central_peripheral` unit test.

### Example App

N/A

### References

https://community.particle.io/t/how-to-change-mtu-size-of-ble/61902

---

### Completeness

- [ ] Added documentation


hal/inc/hal_dynalib_ble.h
```diff
@@ -106,6 +106,9 @@ DYNALIB_FN(71, hal_ble, hal_ble_gap_is_paired, bool(hal_ble_conn_handle_t, void*
 DYNALIB_FN(72, hal_ble, hal_ble_gap_set_pairing_auth_data, int(hal_ble_conn_handle_t, const hal_ble_pairing_auth_data_t*, void*))
 DYNALIB_FN(73, hal_ble, hal_ble_gap_get_pairing_config, int(hal_ble_pairing_config_t*, void*))
 
+DYNALIB_FN(74, hal_ble, hal_ble_gatt_get_att_mtu, ssize_t(hal_ble_conn_handle_t, void*))
+DYNALIB_FN(75, hal_ble, hal_ble_gatt_client_att_mtu_exchange, int(hal_ble_conn_handle_t, void*))
```
tstellanova, line 110: Should this function name include a `get_` as well?

tstellanova (changes requested):
Minor function name changes suggested

hal/inc/hal_dynalib_ble.h
```diff
@@ -81,7 +81,7 @@ DYNALIB_FN(48, hal_ble, hal_ble_gatt_client_discover_service_by_uuid, int(hal_bl
 DYNALIB_FN(49, hal_ble, hal_ble_gatt_client_discover_characteristics, int(hal_ble_conn_handle_t, const hal_ble_svc_t*, hal_ble_on_disc_char_cb_t, void*, void*))
 DYNALIB_FN(50, hal_ble, hal_ble_gatt_client_discover_characteristics_by_uuid, int(hal_ble_conn_handle_t, const hal_ble_svc_t*, const hal_ble_uuid_t*, hal_ble_on_disc_char_cb_t, void*, void*))
 DYNALIB_FN(51, hal_ble, hal_ble_gatt_client_is_discovering, bool(hal_ble_conn_handle_t, void*))
-DYNALIB_FN(52, hal_ble, hal_ble_gatt_set_att_mtu, int(size_t, void*))
+DYNALIB_FN(52, hal_ble, hal_ble_gatt_server_set_desire_att_mtu, int(size_t, void*))
```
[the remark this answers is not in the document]
XuGuohui: This is a typo. `desired` would make more sense, as the att mtu to be set may not be the maximum value.

hal/inc/hal_dynalib_ble.h
XuGuohui: This api is for initiating the att mtu exchange procedure, not really getting something.

user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
```diff
@@ -540,5 +541,28 @@ test(BLE_31_Pairing_Receiption_Reject) {
     }
 }
 
+test(BLE_32_Att_Mtu_Exchange) {
+#define LOCAL_DESIRED_ATT_MTU 100
```
avtolstoy, line 545: Nit: use consts.

user/tests/wiring/ble_central_peripheral/ble_peripheral/peripheral.cpp
```diff
@@ -463,5 +464,29 @@ test(BLE_31_Pairing_Receiption_Reject) {
     }
 }
 
+test(BLE_32_Att_Mtu_Exchange) {
+#define LOCAL_DESIRED_ATT_MTU 123
```
avtolstoy, line 468: Nit: use constants.

user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
```diff
@@ -540,5 +541,28 @@ test(BLE_31_Pairing_Receiption_Reject) {
...
+#define PEER_DESIRED_ATT_MTU 123
+
+    BLE.onAttMtuExchanged([&](const BlePeerDevice& p, size_t attMtu) {
```
avtolstoy, line 548: This is a capturing lambda, might be a good idea to reset the callback after exiting the test.

user/tests/wiring/ble_central_peripheral/ble_peripheral/peripheral.cpp
```diff
@@ -463,5 +464,29 @@ test(BLE_31_Pairing_Receiption_Reject) {
...
+#define PEER_DESIRED_ATT_MTU 100
+
+    BLE.onAttMtuExchanged([&](const BlePeerDevice& p, size_t attMtu) {
```
avtolstoy, line 471: This is a capturing lambda, might be a good idea to reset the callback after exiting the test.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -2366,13 +2332,6 @@ int BleObject::ConnectionsManager::processConnectedEventFromThread(const ble_evt
         linkEvent.params.connected.info = &connection.info;
         notifyLinkEvent(linkEvent);
     }
-    if (desiredAttMtu_ > BLE_DEFAULT_ATT_MTU_SIZE) {
```
avtolstoy, line 2369: @XuGuohui This seems like a major behavior change where we now require the caller to use `hal_ble_gatt_client_att_mtu_exchange()`. Can we keep this automatic?

XuGuohui: What about keeping the existing automatic ATT MTU exchanging scheme, plus probably adding retries? And if it kept failing, this API still enable user to perform the exchange.

avtolstoy: Yeah, I think that'll work best. You get good out of the box behavior (we attempt to negotiate max possible MTU) and the app can always trigger it manually if required.

user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
```diff
@@ -438,7 +442,7 @@ static void pairingTestRoutine(bool request, BlePairingAlgorithm algorithm,
             assertTrue(waitFor([&]{ return pairingRequested; }, 5000));
         }
         assertTrue(BLE.isPairing(peer));
-        assertTrue(waitFor([&]{ return !BLE.isPairing(peer); }, 60000));
+        assertTrue(waitFor([&]{ return !BLE.isPairing(peer); }, 20000));
```
avtolstoy, line 445: @XuGuohui Can we leave this high just in case?

avtolstoy (approved):

user/tests/wiring/ble_central_peripheral/ble_central/central.cpp
XuGuohui: It is probably changed to 6000 since last commit for debugging purpose and is reverted in next commit. It had passed the tests anyway after the final change.

tstellanova (approved):

PR #2399, opened 2022-02-20
Make system_version.h compliant with C; Coalesce FreeRTOS task list internals into single location

mrlambchop:
Remove initialisation of size field and move this into the code to remain compliant with the API.

Allows C files to access the constants and version API in system_version.h

### DIFF
--- a/modules/shared/nRF52840/part1.ld
+++ b/modules/shared/nRF52840/part1.ld
@@ -232,9 +232,15 @@ SECTIONS
         _sbss = ALIGN(., 4);
         link_bss_location = ALIGN(., 4);
         __bss_start__ = _sbss;
+        /* Store the BSS sections for the RTOS first */
+        __rtos_capture_start = .;
+         *tasks.o(.bss COMMON .bss*)
+         *timers*.o(.bss COMMON .bss*)
+        __rtos_capture_end = .;
         *(.bss*)
         *(COMMON)
         link_bss_end = .;
+
         . = ALIGN(4);
         /* This is used by the startup in order to initialize the .bss secion */
         _ebss = . ;
--- a/system/inc/system_version.h
+++ b/system/inc/system_version.h
@@ -347,8 +347,6 @@ typedef struct __attribute__((packed)) SystemVersionInfo
 
 } SystemVersionInfo;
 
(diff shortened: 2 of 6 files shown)


tstellanova (changes requested):
Consider using a designated initializer instead, eg:
```
SystemVersionInfo sys_ver = { .size = sizeof(SystemVersionInfo) };
```

modules/shared/nRF52840/part1.ld
```diff
@@ -232,9 +232,15 @@ SECTIONS
         _sbss = ALIGN(., 4);
         link_bss_location = ALIGN(., 4);
         __bss_start__ = _sbss;
+        /* Store the BSS sections for the RTOS first */
+        __rtos_capture_start = .;
+         *tasks.o(.bss COMMON .bss*)
+         *timers*.o(.bss COMMON .bss*)
```
tstellanova, line 238: I’m not sure what this ld file change has to do with the other system_version changes. Suggest splitting into a separate PR.

user/tests/unit/string_interpolate.cpp
```diff
@@ -109,7 +109,8 @@ class Mocks {
                 String deviceID = spark_deviceID();
                 size_t id_len = deviceID.length();
 
-                SystemVersionInfo sys_ver;
+                SystemVersionInfo sys_ver = { sizeof(SystemVersionInfo), 0, 0, {0} };
```
avtolstoy, line 112: @mrlambchop I'd suggest against using this pattern in C++ and especially for structs that may very well be modified and grow and be explicit about what is initialized:

Either (preferred):
```c++
// Zero-initialize completely
SystemVersionInfo sys_ver = {};
sys_ver.size = sizeof(sys_ver);
```

or if the size field is the first member:

```c++
// Designated initializer for size, rest gets zero-initialized.
SystemVersionInfo sys_ver = {
    .size = sizeof(sys_ver)
};
```

avtolstoy (approved):

modules/shared/nRF52840/part1.ld
avtolstoy: This is for memfault FreeRTOS introspection support, all task lists etc which are static variables are collected in a single coalescent location.

PR #2405, opened 2022-03-14
Listening mode should be allowed to enter when safe mode is requested.

XuGuohui:
### Problem
If Wi-Fi credential is missing and safe mode is requested, device cannot enter listening mode to allow user to configure Wi-Fi credentials. As result device cannot finally connect to the cloud and stay in safe mode.
### Solution
Do not process scenarios that is based on  the `FEATURE_DISABLE_LISTENING_MODE` flag when safe mode is requested.
### Example App

```c
STARTUP( System.enableFeature(FEATURE_DISABLE_LISTENING_MODE) );

void setup() {
  WiFi.clearCredentials();
}

void loop() {

}
```

### References

N/A

---

### Completeness

- [x] Added documentation

### DIFF
--- a/system/src/system_control_internal.cpp
+++ b/system/src/system_control_internal.cpp
@@ -95,7 +95,7 @@ int SystemControl::init() {
     // is enabled, we need to make sure to call characteristic UUIDs in prov
     // mode system calls for the first time instead of here.
     // See system_ble_prov_mode()
-    if (!HAL_Feature_Get(FEATURE_DISABLE_LISTENING_MODE)) {
+    if (system_mode() == SAFE_MODE || !HAL_Feature_Get(FEATURE_DISABLE_LISTENING_MODE)) {
         const int ret = bleChannel_.init();
         if (ret != 0) {
             return ret;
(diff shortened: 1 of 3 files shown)


keeramis (approved):
Tested with my Argon at my desk - Works as expected when device enters safe mode and when it is not in safe mode 👍

technobly:
@keeramis Is there a use case where a user might be in safe mode due to a dependency issue, but need to then also use BLE provisioning mode to get the device connected to the cloud to resolve those dependencies?  If so I think part of this PR will block that use case.  If BLE provisioning mode can only be started from a user app, then that means dependency issues must be resolved before BLE provisioning can be used.  Please merge for 3.3.0 when you feel this is complete, thanks!

keeramis:
@technobly About https://github.com/particle-iot/device-os/pull/2405#issuecomment-1069301834

The user can then enter listening mode and won't enter provisioning mode to setup credentials. This PR especially addresses how the users can enter listening mode when in safe mode and makes this use case possible.

PR #2412, opened 2022-03-29
[test] fixes integration slo/connect_time

technobly:
### Problem

Integration test `connect_time/slo` was failing intermittently on various devices for a number of reasons:

1. The test is designed to run cold boot and warm boot 10 times each, and take the 75th percentile result to compare to the SLO for each test case.  If any individual test failed, it would count towards the total number of tests as a failure, even if the individual test was one of the 25th percentile outliers.
2. Every test in the test runner suite has a 10 minute timeout value, so if the device `waitFor()` timeouts exceed that, the test will result in a failure and we won't know what was really failing in the test.
3. The test was running 10 test cases, and looking for a 75th percentile SLO which is not evenly divisible within 10 test cases, so the math was actually calculating the 70th percentile instead of 75th.
4. If a test failed before setting the `stats` object, the resulting test would be saved as 0s which is technically passing.
5. `stats.coldBootCount` and `stats.warmBootCount` were not incremented if the test exited early.
6. Javascript `.sort()` defaults to alphanumeric string sort and will not sort integers properly
7. There was a potential index error with where the 75th percentile was calculated `const i = Math.floor((values.length-1) * p / 100)`
8. The cold boot test was attempting to power off the NCP before it was on.  This was causing longer power off delays due to no AT interface being available.  R510 was resorting to a lengthy emergency hard power off sequence.

### Solution

1. Do not assertTrue() on each individual test in the connect_time test suite.  Each test will show as pass, and log a failure statement if something should fail.  assertTrue only on the overall 75th percentile calculation.
2. Change the `WAIT_FOR_TIMEOUT` to an overall `TEST_MAX_TIMEOUT` and make sure it's just shy of the 10 minute test timeout.  As the test progresses, successive waitFor()'s should use whatever time is remaining for the test, up to `TEST_MAX_TIMEOUT`.
3. Change the number of tests to be evenly divisible for our required `PERCENTILE` (e.g. `CONNECT_COUNT * PERCENTILE / 100` should be a whole number).  In this case we'll reduce the number of tests to 8 which will speed it up a little as well.
4. Set the test result to `TEST_MAX_TIMEOUT` at the beginning of the test in case of early exit.
5. increment `stats.coldBootCount` and `stats.warmBootCount` at the beginning of the test in case of early exit.
6. Add `.sort()` compare function for integers to properly sort the array
7. Fix index error by calculating percentage first, then adjust for zero based index last. `const i = Math.floor((values.length) * p / 100) - 1`
8. Power on the NCP first for the cold boot test, then turn it off to prepare for cold boot timing.  This speeds up the cold boot time on R510 by about 30 seconds.  There is a noticeable speed up on R410 as well.

### Steps to Test

- run `slo/connect_time` integration tests and cause more or less failures by removing the antenna (faster if you force the tests to exit early in code)
- e.g.
    ```
    device-os/user/tests/integration $ device-os-test build bsom slo/connect_time -v
    device-os/user/tests/integration $ device-os-test run bsom slo/connect_time -v
    ```

---

### Completeness

### DIFF
--- a/user/tests/integration/slo/connect_time/connect_time.cpp
+++ b/user/tests/integration/slo/connect_time/connect_time.cpp
@@ -6,13 +6,13 @@ namespace {
 // Number of connection time measurements to make. When changing this parameter, make sure to change
 // the number of cloud_connect_time_from_cold_boot_XX and cloud_connect_time_from_warm_boot_XX tests
 // and update the spec file accordingly
-const unsigned CONNECT_COUNT = 10;
+const unsigned CONNECT_COUNT = 8;
 
 // Delay before resetting the device after a test
 const system_tick_t DELAY_AFTER_TEST = 2000;
 
-// Default timeout for waitFor()
-const system_tick_t WAIT_FOR_TIMEOUT = 10 * 60 * 1000;
+// Test timeout (we have to complete the entire test in 10 minutes, so set this just shy of that with a little overhead)
+const system_tick_t TEST_MAX_TIMEOUT = 9 * 60 * 1000;
 
 struct Stats {
     struct SignalStats {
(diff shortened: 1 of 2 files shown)


user/tests/integration/slo/connect_time/connect_time.cpp
```diff
@@ -6,13 +6,13 @@ namespace {
...
-const unsigned CONNECT_COUNT = 10;
+const unsigned CONNECT_COUNT = 8;
 
 // Delay before resetting the device after a test
 const system_tick_t DELAY_AFTER_TEST = 2000;
 
-// Default timeout for waitFor()
-const system_tick_t WAIT_FOR_TIMEOUT = 10 * 60 * 1000;
+// Test timeout (we have to complete the entire test in 10 minutes, so set this just shy of that with a little overhead)
+const system_tick_t TEST_MAX_TIMEOUT = 9 * 60 * 1000;
```
avtolstoy, line 15: Note that in `.spec.js` you can increase overall timeout driven by test runner with `timeout(xxxx)`.

avtolstoy (approved):
LGTM

user/tests/integration/slo/connect_time/connect_time.cpp
technobly: Thanks, I'll keep that in mind for other tests.  I think for this test though, 9 minutes should be long enough to power on/off/connect.  Especially when the test fails if it's more than 30s or 60s to connect.  That would mean we are waiting about 8 minutes to power on/off which is more than enough.

joegoggins (approved):
🤯 🤯 LGTM💥 💥 

Thanks for including me on this PR review and writing up such an awesome PR body; even though I'm cheezier with C++ than I'd like to be, I had all of the context I needed to grok the code and understand how your code solved the stated problems.

PR #2417, opened 2022-03-31
[test] communication/functions communication/variables fixes for large describes

avtolstoy:
It may take a while for the device to finish transmitting a large describe with many (50-100) functions/variables and API to report that. Retry a few times in JS side of the tests.


technobly (approved):
Looks good!

PR #2418, opened 2022-03-31
feature/sc-100324/ci-build-cross-platform

busticated:
### Problem

Device OS' build system drives local compilation of user firmware in a number of Particle's user-facing and internal developer tools - most notably Particle Workbench. Currently prior to releasing, the build system is tested on Linux / Ubuntu which means bugs and regressions are much harder to proactively catch on other supported platforms (Windows and macOS).

### Solution

Add a series of checks to the CI/CD system such that key local compilation tasks are regularly tested to ensure they work across all supported environments (Windows, Linux / Ubuntu, and macOS)

### Steps to Test

1. Using macOS, run the test script included in this PR: `./scripts/test-build-tasks.sh ~/path/to/device-os`
2. ...do the same thing using a compatible Ubuntu environment
3. Using Windows, run the test script: `.\scripts/test-build-tasks.ps1 $home\project`
4. Review the [updated CI/CD pipeline](https://github.com/particle-iot/device-os/blob/feature/sc-100324/ci-build-cross-platform/.circleci/config.yml#L18-L42)
5. Observe the [latest CI run in CircleCI](https://app.circleci.com/pipelines/github/particle-iot/device-os/431/workflows/4552f97f-3607-470c-871b-b896c1920826)

**outcome**

Scripts should run properly - exiting on error and otherwise reporting success. CI configuration changes should make sense and do what you expect.

### References

https://app.shortcut.com/particle/story/100324/add-ci-job-s-to-validate-device-os-builds-on-all-supported-oses-windows-linux-ubuntu-and-macos

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,9 +9,49 @@ tag_filters: &tag_filters
       only:
         - /.+/
 
+orbs:
+  win: circleci/windows@2.4.0
+
 # Define a job to be invoked later in a workflow.
 # See: https://circleci.com/docs/2.0/configuration-reference/#jobs
 jobs:
+  build-for-windows:
+    executor:
+      name: win/default
+      size: xlarge
+    parameters:
+      tasks:
+        type: string
+    steps:
+      - checkout
+      - run: git submodule update --init --recursive
+      - install-prtcl-windows
+      - test-localcompiler-windows:
+          tasks: << parameters.tasks >>
+  build-for-darwin:
+    macos:
+      xcode: "13.2.1"
+    parameters:
+      tasks:
+        type: string
+    steps:
+      - checkout
+      - run: git submodule update --init --recursive
+      - install-prtcl-darwin
+      - test-localcompiler-nix:
+          tasks: << parameters.tasks >>
+  build-for-ubuntu:
+    machine:
+      image: ubuntu-2004:current
+    parameters:
+      tasks:
+        type: string
+    steps:
+      - checkout
+      - run: git submodule update --init --recursive
+      - install-prtcl-ubuntu
+      - test-localcompiler-nix:
+          tasks: << parameters.tasks >>
   combine-binaries:
     docker:
       - image: alpine:latest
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -9,9 +9,49 @@ tag_filters: &tag_filters
...
+      size: xlarge
+    parameters:
+      tasks:
+        type: string
+    steps:
+      - checkout
+      - run: git submodule update --init --recursive
+      - install-prtcl-windows
+      - test-localcompiler-windows:
+          tasks: << parameters.tasks >>
```
busticated, line 30: elephant in the room is the total runtime of `1h 25m 48s` 👀 😅 😬 

mainly this is due to windows but i was surprised to see longer times over in macOS as well (`53m 16s`)

next level of optimization here would be to split out steps for individual platforms. i don't have a clear path forward there atm as part of what these jobs do is load and verify that the platforms listed in the manifest can be built. so effectively that list of platforms is not known until the `test-localcompiler-windows` runs (see [here](https://github.com/particle-iot/device-os/pull/2418/files#diff-a86f76a16d6c5e0195845416ff9a0fcdcda32d2aac25849ae15ac10765dafbf8R48)) and i don't know how to create steps dynamically from a previous step.

we could consider more focused triggers and / or running these jobs asynchronously like via cron (or equivalent). the later obviously makes it harder to catch regression _before_ merging but maybe that's ok near-term.

ultimately, the big take-away here is that we have room to improve these task run times 👍

avtolstoy: >  i don't have a clear path forward there atm as part of what these jobs do is load and verify that the platforms listed in the manifest can be built

@busticated We could potentially just pass the platforms as a matrix (since we are anyway listing them out for other tasks) to parallelize the builds based on a platform and validate that all of them are present in the manifest.json separately perhaps even as one of the first steps after a checkout?

```yaml
parameters:
  manifest-expected-platforms:
    type: enum
    default: ["photon", "p1" ...]
# Pass ^^ as matrix to the build jobs to parallelize and also reference pipeline.parameters.manifest-expected-platforms to check with jq presence of all in manifest.json
```

avtolstoy: As an alternative, perhaps enable these jobs only for `release/xxx` and `test/xxx` branches that we use during the release process, as 1.5 hour CI builds during development are impossible to deal with :sweat_smile:

busticated: > We could potentially just pass the platforms as a matrix

yep, i'm aware :+1:

the issue there is how do you ensure that the manifest does not include _more_ than what is specified? for example, the [list here](https://github.com/particle-iot/device-os/pull/2418/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R292) includes Particle platforms not available in the manifest and omits platforms that were in the manifest ([example](https://github.com/particle-iot/device-os/pull/2418/commits/7700150e9414e1997bbd829a380ee02461d83069)). ultimately manifest as source-of-truth is best.

>As an alternative, perhaps enable these jobs only for `release/xxx` and `test/xxx` branches that we use during the release process

that seems like a reasonable first-step - i'll take a shot at making that change. the obvious downside of course is that in that scenario `deviceOS@source` isn't explicitly known-good. we'd only really know if things have regressed at release time which is a big improvement but not ideal.

avtolstoy: > ultimately manifest as source-of-truth is best.

Yup, but we also have a 1.5 hour issue. What I'm proposing is to match that the list we have in circleci.yml matches 1-1 entries present in the manifest.json. Pretty much for example make a JSON array out of the platforms in manifest, make a JSON array out of `manifest-expected-platforms`, subtract them using `jq` and if it's empty == pass. You get both objectives achieved (parallelized build and manifest validate) with some minor hardcoded crust that would easily trigger a failure if something does not add up.

busticated: @avtolstoy ok, here's what i came up with 👉 8405492a24bb90013e4bfb2497b49ad0380353b8 ([link](https://github.com/particle-iot/device-os/pull/2418/commits/8405492a24bb90013e4bfb2497b49ad0380353b8))

[CI: cb02117](https://app.circleci.com/pipelines/github/particle-iot/device-os/527/workflows/0bcc85f1-ab88-48dd-9039-72eb75e99dcf) demonstrates the error state when a platform is missing from the CI yaml

[CI: fd6f3bc](https://app.circleci.com/pipelines/github/particle-iot/device-os/528/workflows/5f162db6-6a78-4d94-bf9f-46d54e8ad06b) demonstrates the error state when a platform is missing from the manifest

avtolstoy (approved):

busticated:
i've run this pipeline now an absolutely ungodly number of times and of course just as i'm about to merge i hit some weirdness 😅🤦. over in an ubuntu job, i just saw a job fail during the `prtcl` install step due to:

```
Could not connect to prtcl.s3-website-us-east-1.amazonaws.com:80 (52.216.177.162), connection timed out
```
([log](https://app.circleci.com/pipelines/github/particle-iot/device-os/530/workflows/16c39ac8-7739-48d5-8d5e-70907d7143c7/jobs/5062?invite=true#step-103-148))

i just re-ran the step and it worked as expected so it's almost certainly just a blip. let's keep an eye on it though - if we see more failures, we can dig in more 👍

PR #2419, opened 2022-04-06
Return relevant error from control request to enter listening mode

keeramis:
### Problem

In the case where listening mode is not allowed in the user firmware, the device-os returns `SYSTEM_ERROR_NONE` instead of the relevant error, which when propagated to the particle-cli results in a timeout.

### Solution

`processRequest()` which processes the control requests returns `SYSTEM_ERROR_NONE`. Added a blocking system call to return the relevant error from the `CTRL_REQUEST_START_LISTENING` control request. 

### Steps to Test

1. Run the following application on the module
2. Run `particle usb listen` on CLI and ensure "Not allowed" error is returned
```
test@Loaners-Macbook-Pro particle-usb (sc-92655/listening-mode-bug) $ particle usb listen
Not allowed
```

### Example App

```c
#include "application.h"
SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {

}
```

### References

[SC-92655](https://app.shortcut.com/particle/story/92655/particle-cli-s-response-when-listening-mode-is-disabled-on-the-device-side)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_control_internal.cpp
+++ b/system/src/system_control_internal.cpp
@@ -188,8 +188,8 @@ void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* 
         break;
     }
     case CTRL_REQUEST_START_LISTENING: {
-        network_listen(0, 0, 0);
-        setResult(req, SYSTEM_ERROR_NONE);
+        auto ret = network_listen_sync(0, 0, 0);
+        setResult(req, ret);
         break;
     }
     case CTRL_REQUEST_STOP_LISTENING: {
--- a/system/src/system_control_internal.cpp
+++ b/system/src/system_control_internal.cpp
@@ -160,27 +160,31 @@ void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /* 
         break;
     }
     case CTRL_REQUEST_RESET: {
-        HAL_Delay_Milliseconds(1000);
-        auto res = System.reset();
-        setResult(req, res);
+        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
+            HAL_Delay_Milliseconds(1000);
+            System.reset();
+        });
         break;
     }
     case CTRL_REQUEST_FACTORY_RESET: {
-        HAL_Delay_Milliseconds(1000);
-        auto res = System.factoryReset();
-        setResult(req, res);
+        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
+            HAL_Delay_Milliseconds(1000);
+            System.factoryReset();
+        });
         break;
     }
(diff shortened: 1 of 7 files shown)


avtolstoy (approved):

system/src/system_control_internal.cpp
```diff
@@ -160,36 +160,32 @@ void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /*
         break;
     }
     case CTRL_REQUEST_RESET: {
-        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
-            HAL_Delay_Milliseconds(1000);
-            System.reset();
-        });
+        HAL_Delay_Milliseconds(1000);
+        auto res = System.reset();
+        setResult(req, res);
```
keeramis, line 165: I removed lambda functions here to collect the return code. Please let me know if there is more C++ way to do this.

system/src/system_control_internal.cpp
```diff
@@ -160,36 +160,32 @@ void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /*
...
         break;
     }
     case CTRL_REQUEST_SAFE_MODE: {
-        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
-            HAL_Delay_Milliseconds(1000);
-            System.enterSafeMode();
-        });
+        HAL_Delay_Milliseconds(1000);
+        auto res = System.enterSafeMode();
```
avtolstoy, line 182: Nit: might want to just `setResult(req, System.enterSafeMode());` etc everywhere just to be consistent with the rest of the cases.

avtolstoy (approved):

system/src/system_control_internal.cpp
```diff
@@ -160,36 +160,32 @@ void SystemControl::processRequest(ctrl_request* req, ControlRequestChannel* /*
...
-            HAL_Delay_Milliseconds(1000);
-            System.reset();
-        });
+        HAL_Delay_Milliseconds(1000);
+        auto res = System.reset();
+        setResult(req, res);
         break;
     }
     case CTRL_REQUEST_FACTORY_RESET: {
-        setResult(req, SYSTEM_ERROR_NONE, [](int result, void* data) {
```
avtolstoy, line 170: Actually, no, we can't change these. This will affect the behavior. Note that the lambda is passed as third argument which is a completion handler, so we enter DFU only after the request finishes, otherwise if we return a value here we will immediately execute it and e.g. CLI will fail to get a request response from the device.

avtolstoy (changes requested):

avtolstoy (approved):

PR #2443, opened 2022-04-26
[gen3][esomx] Define E404X platform

scott-brust:
### Problem
E404X Hardware needs to be supported in DVOS

### Solution
Add new `PLATFORM_ID` for `esomx` platform, largely based off of boron module/platform definitions

### Steps to Test
Build using new platform and test with boron hardware:
```
make PLATFORM=esomx APP=tinker clean all
```

### Example App
Use tinker

---

### Completeness

- [ ] Added documentation


platform/MCU/nRF52840/inc/platform_config.h
```diff
@@ -37,7 +37,8 @@
     PLATFORM_ID == PLATFORM_ASOM || \
     PLATFORM_ID == PLATFORM_BSOM || \
     PLATFORM_ID == PLATFORM_B5SOM || \
-    PLATFORM_ID == PLATFORM_TRACKER
+    PLATFORM_ID == PLATFORM_TRACKER || \
+    PLATFORM_ID == PLATFORM_ESOMX
```
XuGuohui, line 41: If the E404X is gping to adopt an 8MB external flash, we need to modify line 52.

scott-brust: I think there is still discussion about if E404X will support only 4mb filesystem or not, regardless of flash part. I updated `flash_mal` to support 8mb for now

scott-brust:
> I think the following files need to be modified as well:
> 
> 1. platform/MCU/nRF52840/src/hw_config.c, line 197
> 2. user/applications/tinker/application.cpp
> 3. user/tests/wiring/no_fixture_long_running/network.cpp, pwm.cpp
> 4. user/tests/wiring/spi_master_slave

I'm working on pinmap next and will update tinker and tests then

hal/src/boron/hal_platform_config.h
```diff
@@ -28,6 +28,18 @@
...
+
+#define HAL_PLATFORM_POWER_MANAGEMENT (1)
+#define HAL_PLATFORM_PMIC_BQ24195 (1)
+#define HAL_PLATFORM_PMIC_BQ24195_FAULT_COUNT_THRESHOLD (4)
+#define HAL_PLATFORM_FUELGAUGE_MAX17043 (1)
+#define HAL_PLATFORM_PMIC_BQ24195_I2C (HAL_I2C_INTERFACE2)
+#define HAL_PLATFORM_FUELGAUGE_MAX17043_I2C (HAL_I2C_INTERFACE2)
+#define HAL_PLATFORM_RADIO_ANTENNA_INTERNAL (0)
+#define HAL_PLATFORM_RADIO_ANTENNA_EXTERNAL (1)
+#define HAL_PLATFORM_FLASH_MX25L3233F (1)
```
artyLee, line 41: If there is using 8M flash, need to change macro `#define HAL_PLATFORM_FLASH_MX25L3233F (1)` to `#define HAL_PLATFORM_FLASH_MX25R6435FZNIL0 (1)`

hal/src/boron/pinmap_hal.c
```diff
@@ -123,6 +123,58 @@ const uint8_t NRF_PIN_LOOKUP_TABLE[48] = {
 
 #endif // PLATFORM_ID == PLATFORM_BSOM
 
+#if PLATFORM_ID == PLATFORM_ESOMX
+
+static Hal_Pin_Info s_pin_map[TOTAL_PINS] = {
```
artyLee, line 128: This pin map looks not match our Evo board schematic. Maybe you will change it later. Please reference: [evo schematic](https://github.com/particle-iot/Evo/tree/main/Eagle/v0.02)

hal/src/boron/pinmap_defines.h
```diff
@@ -167,3 +167,75 @@
 #define UBVINT 37
 
 #endif // PLATFORM_ID == PLATFORM_BSOM
+
+#if PLATFORM_ID == PLATFORM_ESOMX
```
artyLee, line 171: Synchronization update from pin map.

platform/MCU/nRF52840/inc/platform_config.h
scott-brust: E404X will support 8mb flash, and expose 4mb to user as FS, like tracker and b5som here.

XuGuohui (approved):

PR #2447, opened 2022-04-28
[deprecation] remove setup_done flag and add deprecation notice

keeramis:
### Description

This PR removes the setup-done flag, and adds deprecation notice when the dct setup-done flag is used.

#### Starting Device OS 4.x:
1. Attempting to set this bit via particle CLI or BLE will not have any impact on the device behavior. The setup_done bit can be modified only via user app.
2. Setting this via user app causes the following deprecation warning. Starting Device OS 5.x, the setup_done bit and the relevant HAL flag will be completely removed and will cause a build error.
```
Setup Done flag has been deprecated since Device OS 4.0 and has no effect on device behavior. This will be removed in Device OS 5.x causing build error!
```

### Steps to Test

Run an app that reads/writes `DCT_SETUP_DONE_OFFSET` and verify that the following warning occurs during compiling
`Setup Done flag has been deprecated since Device OS 4.0 and has no effect on device behavior`

### Example App
```
#include "Particle.h"
#include "dct.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
void setup() {
	// 0x01 = setup done
	// 0xff = setup not done (go into listening mode at boot)
	const uint8_t val = 0x01;
	dct_write_app_data(&val, DCT_SETUP_DONE_OFFSET, 1);

	// This is just so you know the operation is complete
	pinMode(D7, OUTPUT);
	digitalWrite(D7, HIGH);
}

void loop() {
}
```

2. The following commands shouldn't break when run through CLI
Screen Shot 2022-04-29 at 12 15 38 PM

3. Mobile app should setup OK on factory fresh module

### References

[SC-72116](https://app.shortcut.com/particle/story/72116/remove-the-need-for-setup-done-setupdone-flag-in-device-os-across-the-board-for-all-devices-sims)

---

### Completeness

- [ ] Added documentation


platform/MCU/nRF52840/inc/dct.h
```diff
@@ -151,50 +151,57 @@ typedef struct __attribute__((packed)) application_dct {
...
 #define STATIC_ASSERT_DCT_OFFSET(field, expected) PARTICLE_STATIC_ASSERT( dct_##field, offsetof(application_dct_t, field)==expected)
 #define STATIC_ASSERT_FLAGS_OFFSET(field, expected) PARTICLE_STATIC_ASSERT( dct_sysflag_##field, offsetof(platform_system_flags_t, field)==expected)
 
 #define DCT_OTA_UPDATE_FLAG_SET (0xA5)
 #define DCT_OTA_UPDATE_FLAG_CLEAR (0XFF)
 
+typedef enum platform_dct_deprecated {
+    DCT_SETUP_DONE_OFFSET __attribute__((deprecated("Setup Done flag has been deprecated since Device OS 4.0 and has no effect on device behavior")))
+            = (offsetof(application_dct_t, setup_done)),
+    DCT_SETUP_DONE_SIZE __attribute__((deprecated("Setup Done flag has been deprecated since Device OS 4.0 and has no effect on device behavior")))
```
technobly, line 201: Since we are planning on removing SETUP DONE in 5.x completely (which will cause a build error), can we add that notification to these deprecation messages?  Something like "Will be removed in 5.x and cause a build error!" ?

technobly (changes requested):
I tested this on Boron and it allows the device to exit Listening Mode without clearing the SETUP DONE flag as expected.  I also don't see the warning when not including this in user code as expected.  👍🏼 

FIXED: ~Can we suppress the deprecation warnings when building system firmware though?~

avtolstoy:
@keeramis As discussed I think control requests returning NOT_SUPPORTED will mess up existing mobile applications (to be tested). I'd propose for now in 4.x we stub them out to always return setup done - yes and on changes to 1 report success as well. On changes to 0 report an error.

keeramis:
@avtolstoy A couple of questions: 

1. What will happen if the user ACTUALLY NEEDS to use `particle usb setup-done`? Must they use it via user app and not the control requests?
2. system doesn't manifest deprecations as errors (just tested) - it still keeps them as warnings during compilation - just a ton of warnings though for all references of `DCT_SETUP_DONE_OFFSET` in system

keeramis:
Another reference of `DCT_SETUP_DONE_OFFSET` still exists in `testAndClearListeningModeFlag()`

technobly (approved):
LGTM!

keeramis:
Results for reference: 

1. Running the control requests from CLI 
image

2. Setup through Particle mobile app run on a Boron module with setup-done set to 0x00, and cleared credentials. Mobile setup goes OK ✔️ 

4. Setup through Particle mobile app run on a factory fresh module (actual factory fresh module) with tinker/system from the PR branch. Mobile setup goes OK ✔️  (Currently stuck activating SIM but otherwise OK)

system/src/control/config.cpp
```diff
@@ -171,41 +171,25 @@ int getDeviceMode(ctrl_request* req) {
 }
 
 int setDeviceSetupDone(ctrl_request* req) {
-#if HAL_PLATFORM_DCT_SETUP_DONE
+    // This functionality is currently deprecated.
+    // Do not perform any DCT accesses. Instead return an appropriate error code
```
avtolstoy, line 175: I'd probably leave `HAL_PLATFORM_DCT_SETUP_DONE` conditional, or get rid of it completely everywhere.

avtolstoy (approved):

keeramis:
> I'd probably leave HAL_PLATFORM_DCT_SETUP_DONE conditional, or get rid of it completely everywhere.


@avtolstoy Since the code under `HAL_PLATFORM_DCT_SETUP_DONE` is not compiled, may be you meant to leave the conditional as a comment? Either way, I have removed the flag altogether.

avtolstoy (approved):

PR #2456, opened 2022-05-24
[E404X] Determine flash part at runtime

scott-brust:
### Problem

The ESOMX, B404X, B524X, T404X and other SKUs will need to support multiple different flash chips in order to avoid dependencies on a single supplier. Historically we have handled hardware variations using OTP, but we cannot use that as we will not know the flash variant at boot. The exflash driver will need to discover what flash part is attached, and update the driver to communicate with it accordingly.

### Solution

This PR reads the flash Device ID from the part, then modifies the driver to account for any differences/quirks of that part. Right now the following flash parts are supported
`MX25R6435FZNIL0`
`GD25WQ64EQFG` -- as of yet untested

### Steps to Test

For ESOMX / other boron/NRF52840 platforms, this PR should be backwards compatible with the current 8 and 4mb MXIC parts (`MX25R6435FZNIL0` and `MX25L3233F`)

### Example App

Any, tinker is fine

### References

See AVL Components List in sheets

---

### Completeness

- [ ] Added documentation


hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+
+    QSPI_GD25_CMD_WRSR2          = 0x31, // Write Status Register 2
+    QSPI_GD25_CMD_WRSR3          = 0x11, // Write Status Register 3
+    QSPI_GD25_CMD_SEC_ERASE      = 0x44, // Security Register Erase (ie Erase OTP)
+    QSPI_GD25_CMD_SEC_READ       = 0x48, // Security Register Read
+    QSPI_GD25_CMD_SEC_PROGRAM    = 0x42, // Security Register Program (ie Write OTP)
+    QSPI_GD25_CMD_PGMERS_SUSPEND = 0x75, // Program/Erase Suspend
+};
+
+enum qspi_type_t {
```
scott-brust, line 70: I should consider changing the order of this enum to search for the more likely flash part

scott-brust:
@XuGuohui Here is where I am with the GD25 flash part support. Feel free to use this PR if it is helpful to bring up the part. I wont have hardware to test on for a few more days

hal/src/nRF52840/exflash_hal.c
```diff
@@ -388,64 +563,96 @@ int hal_exflash_copy_sector(uintptr_t src_addr, uintptr_t dest_addr, size_t data
...
+    int ret = 0;
 
     hal_exflash_lock();
 
-    /* Enter Secure OTP mode */
-    int ret = enter_secure_otp();
-    if (ret) {
-        goto hal_exflash_write_special_done;
-    }
+    if (hal_exflash_driver.type == QSPI_TYPE_GD25WQ64EQFG) {
```
scott-brust, line 617: The GD25 OTP register read/writes are copy/paste of each other basically. Could refactor/consolidate this if this isnt ok

hal/src/nRF52840/exflash_hal.c
```diff
@@ -487,20 +695,11 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
...
-            cinstr_cfg.opcode = QSPI_STD_CMD_WRSR;
-            cinstr_cfg.length = NRF_QSPI_CINSTR_LEN_2B;
-            ret = exflash_qspi_cinstr_xfer(&cinstr_cfg, temporary, NULL);
-#else
-#error "Unsupported platform for external flash"
-#endif
+
+            if (hal_exflash_driver.init) {
+                // TODO: Consolidate init here?
+                ret = hal_exflash_driver.init();
```
scott-brust, line 701: The actual flash part init procedures aren't that complicated. We could just do an if/else here and do it inline rather than this func pointer if thats easier.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -464,21 +671,22 @@ int hal_exflash_special_command(hal_exflash_special_sector_t sp, hal_exflash_com
         };
 
         if (cmd == HAL_EXFLASH_COMMAND_SLEEP) {
-            cinstr_cfg.opcode = QSPI_MX25_CMD_SLEEP;
-        } else if (cmd == HAL_EXFLASH_COMMAND_WAKEUP) {
-            /* Wake up external flash from deep power down modeby sending read id command */
-            cinstr_cfg.opcode = QSPI_MX25_CMD_READ_ID;
+            cinstr_cfg.opcode = QSPI_STD_CMD_SLEEP;
+        } else if (cmd == HAL_EXFLASH_COMMAND_WAKEUP || cmd == HAL_EXFLASH_COMMAND_READID) {
```
scott-brust, line 675: Do we want to remove `HAL_EXFLASH_COMMAND_WAKEUP` and make it clear that to wake up the chip we use `HAL_EXFLASH_COMMAND_READID`?

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
-static const size_t MX25_OTP_SECTOR_SIZE = 4096 / 8; /* 4Kb or 512B */
+// For any variations between flash parts, set member vars at runtime
+typedef struct  {
+    int type;                      // qspi_type_t to identify what part is present
+    bool part_detected;            // Have we positively identified the flash part yet
+    uint8_t write_opcode;          // Which opcode to use for quad write
+    uint8_t pgmers_suspend_opcode; // Which opcode to use when interrupting pending program/erase command
+    uint8_t sck_delay;             // tSHSL, tWHSL, and tSHWL in number of 16 MHz periods (62.5ns)
+    uint32_t otp_size;             // Size of OTP area in bytes
+    int (*init)();                 // Function to call once chip is identified. IE enter QSPI mode
```
XuGuohui, line 85: Prefer use typedef.

@YutingYou As discussed, we're going to go with this way:
1. define `hal_qspi_flash_params_t`, which contains flash-specific parameters.
2. create an static `hal_qspi_flash_params_t` array that has all kind of flash parameters initialized
3. define `hal_exflash_t`:
```c++
typedef struct hal_exflash_t {
    hal_qspi_flash_type_t type;
    uintt8_t reserved[3];
    hal_qspi_flash_parames* params;
} hal_exflash_t;
``` 
4. use default QSPI config to read chip ID and initialize the `params` field accordingly followed by initializing QSPI with  new config.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+                                   // TODO: GD25: CLK and CSN pin drive strength? 
+} exflash_driver_parameters_t;
```
XuGuohui, line 87: Make sure the size of this struct is word aligned by adding `reserved` field.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+
+static exflash_driver_parameters_t hal_exflash_driver = {};
 
 static hal_exflash_state_t qspi_state = HAL_EXFLASH_STATE_DISABLED;
 
+static const size_t MX25R6435F_OTP_SECTOR_SIZE = 1024;
```
XuGuohui, line 93: We can probably add the following values to the flash parameter struct.

XuGuohui (changes requested):
We need to also remove `HAL_PLATFORM_FLASH_MX25L3233F` for Boron and the ESOM after rebasing on top of `develop`.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+{                                           \
+    .opcode    = (opc),                     \
+    .length    = (len),                     \
+    .io2_level = true,                      \
+    .io3_level = true,                      \
+    .wipwait   = true,                      \
+    .wren      = true                       \
+}
 
 enum qspi_cmds_t {
```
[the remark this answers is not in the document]
YutingYou: This platform-related code to configure nrf52's qspi, it would be better to keep it in the source file.

hal/src/nRF52840/exflash_hal.c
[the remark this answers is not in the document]
YutingYou: The flash driver header shouldn't care about the exact vendor of the flash chip since we can detect it and apply the corresponding driver in the source code.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+    QSPI_GD25_CMD_WRSR2          = 0x31, // Write Status Register 2
+    QSPI_GD25_CMD_WRSR3          = 0x11, // Write Status Register 3
+    QSPI_GD25_CMD_SEC_ERASE      = 0x44, // Security Register Erase (ie Erase OTP)
+    QSPI_GD25_CMD_SEC_READ       = 0x48, // Security Register Read
+    QSPI_GD25_CMD_SEC_PROGRAM    = 0x42, // Security Register Program (ie Write OTP)
+    QSPI_GD25_CMD_PGMERS_SUSPEND = 0x75, // Program/Erase Suspend
+};
+
+enum qspi_type_t {
+    QSPI_TYPE_MX25L3233F,
```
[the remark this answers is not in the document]
YutingYou: 👍

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+    QSPI_TYPE_MX25R6435FZNIL0,
+    QSPI_TYPE_GD25WQ64EQFG,
+    QSPI_TYPE_END
 };
 
-static const size_t MX25_OTP_SECTOR_SIZE = 4096 / 8; /* 4Kb or 512B */
+// For any variations between flash parts, set member vars at runtime
+typedef struct  {
+    int type;                      // qspi_type_t to identify what part is present
```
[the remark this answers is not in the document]
YutingYou: 👍

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+    bool part_detected;            // Have we positively identified the flash part yet
```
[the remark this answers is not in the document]
YutingYou: Yeah, we need to add an unknown type in the flash type define

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
+enum qspi_type_t {
+    QSPI_TYPE_MX25L3233F,
+    QSPI_TYPE_MX25R6435FZNIL0,
+    QSPI_TYPE_GD25WQ64EQFG,
+    QSPI_TYPE_END
 };
 
-static const size_t MX25_OTP_SECTOR_SIZE = 4096 / 8; /* 4Kb or 512B */
+// For any variations between flash parts, set member vars at runtime
+typedef struct  {
```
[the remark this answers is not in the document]
YutingYou: As discussed, we can abandon the custom structure, and set dedicated configuration in the nrf52 qspi init structure and store them into an array.

hal/src/nRF52840/exflash_hal.c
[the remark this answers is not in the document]
XuGuohui: They are not platform-specific, they are flash-specific and will be shared with P2 as well.

hal/src/nRF52840/exflash_hal.c
```diff
@@ -35,24 +33,76 @@
...
-    QSPI_STD_CMD_WREN     = 0x06,
-    QSPI_STD_CMD_RSTEN    = 0x66,
-    QSPI_STD_CMD_RST      = 0x99,
-    QSPI_MX25_CMD_ENSO    = 0xb1,
-    QSPI_MX25_CMD_EXSO    = 0xc1,
-    QSPI_MX25_CMD_WRSCUR  = 0x2f,
-    QSPI_MX25_CMD_SLEEP   = 0xB9,
-    QSPI_MX25_CMD_READ_ID = 0xAB,
-    QSPI_STD_CMD_PGMERS_SUSPEND = 0xB0
+    QSPI_STD_CMD_WRSR            = 0x01, // Write Status Register
```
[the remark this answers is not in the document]
XuGuohui: Sure. I meant we don't need to add `STD` for common commands and suffix it with `MX25` or `GD25` with flash-specific commands.

hal/src/nRF52840/exflash_hal.c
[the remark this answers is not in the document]
XuGuohui: We still need it, see what we do in uart hal. And then define an array like `static hal_qspi_flash_params_t flash_params[MAX_SUPPORTED_FLASH_VARIANT]`.

YutingYou:
To get rid of `goto` and use RAII lock in the driver, I change the c source file to cpp but keep most of the original driver. The OTP read/write passed the test. I'll continue to run an EEPROM test which is based on the file system.

@XuGuohui @avtolstoy Could you please help re-review the driver to make sure it is implemented as we discussed yesterday?

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,702 @@
...
+        .length    = NRF_QSPI_CINSTR_LEN_4B,
+        .io2_level = true,
+        .io3_level = true,
+        .wipwait   = true,
+        .wren      = true
+    };
+    CHECK(exflash_qspi_cinstr_xfer(&cinstr_cfg, NULL, chip_id));
+
+    flash_type = hal_exflash_get_type(chip_id);
+    hal_exflash_params_t flash_params = {};
```
XuGuohui, line 327: We can simply define a pointer `hal_exflash_params_t*` to get  the corresponding parameters.

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,76 @@
...
+#include <stdint.h>
+#include "nrfx_qspi.h"
+
+#define MX25R6435F_OTP_SECTOR_SIZE      1024
+#define MX25L3233F_OTP_SECTOR_SIZE      512
+#define GD25_SECURITY_REGISTER_COUNT    3
+#define GD25_SECURITY_REGISTER_SIZE     1024
+#define GD25_OTP_SECTOR_SIZE            (GD25_SECURITY_REGISTER_SIZE * GD25_SECURITY_REGISTER_COUNT)
+
+typedef enum hal_exflash_type_t {
```
XuGuohui, line 29: Prefer `hal_qspi_flash_type`

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,76 @@
...
+    uint8_t             reserved[3];     // Reserved padding
+    uint32_t            otp_size;        // Size of OTP area in bytes
+} hal_exflash_params_t;
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+
+hal_exflash_type_t hal_exflash_get_type(const uint8_t* id_buf);
+int hal_exflash_get_params(hal_exflash_type_t type, hal_exflash_params_t* params);
```
XuGuohui, line 72: return a pointer to the parameters so that we don't have to make a copy.

XuGuohui (changes requested):
Again, we need to remove `HAL_PLATFORM_FLASH_MX25L3233F` for Boron and ESOMX after rebasing on top of `develop`.

YutingYou:
@scott-brust Welcome back and congratulations!

It is probably a concern because flash initialization is called in the bootloader `main()` function while the C++ constructor initialization is before the `main()`. So I keep most of the original c-style driver, there are no global variables would call the constructor.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,703 @@
...
+        .io2_level = true,
+        .io3_level = true,
+        .wipwait   = true,
+        .wren      = true
+    };
+    CHECK(exflash_qspi_cinstr_xfer(&cinstr_cfg, nullptr, chip_id));
+
+    flash_type = hal_exflash_get_type(chip_id);
+    const hal_exflash_params_t* flash_params = hal_exflash_get_params(flash_type);
+    CHECK_TRUE(flash_params, SYSTEM_ERROR_NOT_FOUND);
```
XuGuohui, line 354: We should probably assert it, instead of returning an error.

XuGuohui (approved):

scott-brust:
> @scott-brust Welcome back and congratulations!
> 
> It is probably a concern because flash initialization is called in the bootloader `main()` function while the C++ constructor initialization is before the `main()`. So I keep most of the original c-style driver, there are no global variables would call the constructor.

Thanks! 
Ill keep in mind that we must be careful to not use globals that require a constructor in this driver. I have the code running on my boron with the GD25 flash, I will probe it and confirm all the commands are working as expected. I will also do more OTP write/read testing (ie across security register boundaries). I will add any comments by end of day tomorrow... apologies for the delay

hal/src/nRF52840/exflash_hal.cpp
YutingYou: It's fine, we've already had assert for `hal_exflash_init()` in `hw_config.c`.

scott-brust:
A couple of points:

1) I did some testing and verified that I can read/write across OTP register boundaries on the GD25 flash (with my latest changes). [See attached patch file I used to test](https://github.com/particle-iot/device-os/files/8882729/exflash_hal_test.txt)

2) I verified that we can re-write OTP data, and that it will only clear the bits, so the data will be AND'd together. Just as an FYI

3) I found a problem with `HAL_EXFLASH_COMMAND_SLEEP` though. For GD25 this is `Deep Power-Down (DP) (B9H)`. Which according to documentation requires a special command to wake it back up: `Release from Deep Power-Down and Read Device ID (RDI) (ABH)`

Our current init process uses the `Read Identification (RDID) (9FH)` command, which specifically is not supposed to be used while GD25 is in deep sleep 🙂 
```
The Read Identification (RDID) command should not be issued while the device is in Deep Power-Down Mode.
```

Working on a fix for this now, so please dont merge just yet

scott-brust:
Pushed a fix to wake GD25 from sleep mode. Seems to work ok.
I still havent tested
1. `HAL_QSPI_CMD_GD25_PGMERS_SUSPEND`
2. `HAL_QSPI_CMD_GD25_SEC_ERASE` To see if this even works
3. `HAL_EXFLASH_COMMAND_LOCK_ENTIRE_OTP` Is also still unimplemented for GD25, not sure if this is high priority though

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    {
+        nrf_qspi_cinstr_conf_t wake_command = {
+            .opcode    = HAL_QSPI_CMD_GD25_WAKE,
+            .length    = NRF_QSPI_CINSTR_LEN_1B,
+            .io2_level = true,
+            .io3_level = true,
+            .wipwait   = true,
+            .wren      = true
+        };
+        CHECK(exflash_qspi_cinstr_xfer(&wake_command, nullptr, NULL));
```
YutingYou, line 361: Nit/Style: nullptr

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+        // Read ID again
+        exflash_qspi_cinstr_xfer(&read_id_command, nullptr, chip_id);
```
YutingYou, line 363: `CHECK()` is needed.

hal/src/nRF52840/exflash_hal.cpp
YutingYou: A 20us delay is needed before reading ID to ensure the flash is woken up.

Screen Shot 2022-06-13 at 10 42 19

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,77 @@
...
+#pragma once
+
+#include <stdint.h>
+#include "nrfx_qspi.h"
+
+#define MX25R6435F_OTP_SECTOR_SIZE      1024
+#define MX25L3233F_OTP_SECTOR_SIZE      512
+#define GD25_SECURITY_REGISTER_COUNT    3
+#define GD25_SECURITY_REGISTER_SIZE     1024
+#define GD25_OTP_SECTOR_SIZE            (GD25_SECURITY_REGISTER_SIZE * GD25_SECURITY_REGISTER_COUNT)
```
YutingYou, line 27: Should we reverse `GD25_OTP_SECTOR_SIZE` and `GD25_SECURITY_REGISTER_SIZE`? The way we define the sector size is different from it in our cognition.

hal/inc/exflash_hal.h
```diff
@@ -55,6 +56,7 @@ int hal_exflash_read(uintptr_t addr, uint8_t* data_buf, size_t data_size);
 int hal_exflash_copy_sector(uintptr_t src_addr, size_t dest_addr, size_t data_size);
 int hal_exflash_lock(void);
 int hal_exflash_unlock(void);
+int hal_exflash_otp_size(void);
```
avtolstoy, line 59: This should be done within `hal_exflash_special_command()` or some other generic method taking `hal_exflash_special_sector_t` as an argument. No need to introduce OTP-specific APIs.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+{                                           \
+    .opcode    = (opc),                     \
+    .length    = (len),                     \
+    .io2_level = true,                      \
+    .io3_level = true,                      \
+    .wipwait   = true,                      \
+    .wren      = true                       \
+}
+
+#define NRF_QSPI_QUICK_CFG(WRITEOC, READOC, FREQUENCY)      \
```
avtolstoy, line 50: Nit/style: don't use ALLCAPS for macro argument names.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,702 @@
...
+    return exflash_qspi_cinstr_xfer(&config, p_tx_buffer, NULL);
+}
+
+static int perform_write(uintptr_t addr, const uint8_t* data, size_t size) {
+    return nrf_system_error(nrfx_qspi_write(data, size, addr));
+}
+
+static nrf_qspi_cinstr_len_t get_nrf_cinstr_length(size_t length) {
+    nrf_qspi_cinstr_len_t cinstrLen = NRF_QSPI_CINSTR_LEN_1B;
+    switch (length) {
```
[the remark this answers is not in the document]
avtolstoy: Keep it as-is. Add `constexpr` as well just in case.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+#include "flash_common.h"
+#include "nrf_nvic.h"
+#include "concurrent_hal.h"
+#include "gpio_hal.h"
+#include "system_error.h"
+#include "nrf_system_error.h"
+#include "check.h"
+#include "exflash_hal_params.h"
+
+#define PRTCL_QSPI_DEFAULT_CINSTR(opc, len) \
```
avtolstoy, line 40: Nit/style: `PRTCL_` is not a common prefix that we use.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    nrf_qspi_cinstr_conf_t cinstr_cfg = PRTCL_QSPI_DEFAULT_CINSTR(HAL_QSPI_CMD_STD_WRSR, NRF_QSPI_CINSTR_LEN_2B);
+    uint8_t status_register[1] = { 0x40 };
+    return exflash_qspi_cinstr_xfer(&cinstr_cfg, status_register, nullptr);
+}
+
+int gd25_switch_to_qspi() {
+    nrf_qspi_cinstr_conf_t cinstr_cfg = PRTCL_QSPI_DEFAULT_CINSTR(HAL_QSPI_CMD_STD_WRSR, NRF_QSPI_CINSTR_LEN_2B);
+    uint8_t status_registers[3] = {0x00, 0x02, 0x00};
+
+    CHECK_FALSE(exflash_qspi_cinstr_xfer(&cinstr_cfg, &status_registers[0], nullptr), SYSTEM_ERROR_FLASH_IO);
```
avtolstoy, line 228: Why `CHECK_FALSE()` for a function returning `int` system error codes already? Same below.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    uint8_t addrBuf[4] = {}; 
+    uint32_t address = ((security_registers_index + 1) << 12) + security_registers_addr;
+    addrBuf[0] = (uint8_t)((address>>16) & 0xFF);
+    addrBuf[1] = (uint8_t)((address>>8) & 0xFF);
+    addrBuf[2] = (uint8_t)((address>>0) & 0xFF);
+    addrBuf[3] = 0x00;
+
+    // start long frame transfer with "Read Security Register" opcode
+    nrfx_err_t ret = nrfx_qspi_lfm_start(&readSecurityRegister);
+    CHECK_TRUE(ret == NRFX_SUCCESS, nrf_system_error(ret));
```
avtolstoy, line 255: Nit: `CHECK_NRF`? Same below.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    int err_code = NRF_SUCCESS;
+
+    const size_t block_length = len == NRF_QSPI_ERASE_LEN_4KB ? 4096 : 64 * 1024;
+
+    /* Round down to the nearest block */
+    start_addr = ((start_addr / block_length) * block_length);
+
+    for (size_t i = 0; i < num_blocks; i++) {
+        err_code = nrfx_qspi_erase(len, start_addr);
+        CHECK_TRUE(err_code == NRFX_SUCCESS, nrf_system_error(err_code));
```
avtolstoy, line 303: `CHECK_NRF`

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    nrfx_qspi_config_t default_config = NRF_QSPI_QUICK_CFG(NRF_QSPI_WRITEOC_PP, NRF_QSPI_READOC_FASTREAD, NRF_QSPI_FREQ_32MDIV2);
+    int ret = nrfx_qspi_init(&default_config, nullptr, nullptr);
+    if (ret) {
+        if (ret == NRFX_ERROR_TIMEOUT) {
+            // Could have timed out because the flash chip might be in an ongoing program/erase operation.
+            // Suspend it.
+            LOG_DEBUG(TRACE, "QSPI NRFX timeout error. Suspend pgm/ers op.");
+            ret = hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, nullptr, nullptr, 0);
+        }
+        return nrf_system_error(ret);
```
avtolstoy, line 336: This is a change in behavior. Previously in case of timeout and having to execute `SUSPEND_PGMERS` we were only breaking the execution if `hal_exflash_special_command` fails.

https://github.com/particle-iot/device-os/pull/2456/files#diff-9f1bf2f5c73c5d1dfb77046821a95dd7f47f367cc24f08f378ec9c7595d623b5R611 Also in general this will no longer work as we don't yet know the type of the flash?

See `hal_exflash_get_params` as well.

hal/src/nRF52840/exflash_hal_params.cpp
```diff
@@ -0,0 +1,72 @@
...
+        .suspend_opcode = HAL_QSPI_CMD_GD25_PGMERS_SUSPEND,
+        .reserved       = {},
+        .otp_size       = GD25_OTP_SECTOR_SIZE
+    }
+};
+
+} // Anonymous
+
+const hal_exflash_params_t* hal_exflash_get_params(hal_qspi_flash_type_t type) {
+    CHECK_TRUE(type != HAL_QSPI_FLASH_TYPE_UNKNOWN, nullptr);
```
avtolstoy, line 58: We should probably default to some 'generic' type that supports just standard common commands.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    // Determine chip type / wake up from potential sleep
+    uint8_t chip_id[3] = {};
+    nrf_qspi_cinstr_conf_t read_id_command = {
```
avtolstoy, line 341: Nit: let's use some default generic type before full initialization and delegate to `hal_exflash_special_command` here.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+        .opcode    = HAL_QSPI_CMD_STD_READ_ID,
+        .length    = NRF_QSPI_CINSTR_LEN_4B,
+        .io2_level = true,
+        .io3_level = true,
+        .wipwait   = true,
+        .wren      = true
+    };
+    
+    // If read chip_id fails, it may be a GD25 that is asleep
```
avtolstoy, line 350: Why not switch type (`flash_type`) to GD25 temporarily and delegate to `hal_exflash_special_command` ?

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+
+    flash_type = hal_exflash_get_type(chip_id);
+    const hal_exflash_params_t* flash_params = hal_exflash_get_params(flash_type);
+    CHECK_TRUE(flash_params, SYSTEM_ERROR_NOT_FOUND);
+
+    // Re-initialize QSPI using corresponding flash parameters
+    nrfx_qspi_uninit();
+    nrfx_qspi_config_t new_config = NRF_QSPI_QUICK_CFG(flash_params->write_opcode, flash_params->read_opcode, NRF_QSPI_FREQ_32MDIV2);
+    ret = nrfx_qspi_init(&new_config, nullptr, nullptr);
+    CHECK_TRUE(ret == NRFX_SUCCESS, nrf_system_error(ret));
```
avtolstoy, line 374: Nit: `CHECK_NRF`

hal/src/nRF52840/exflash_hal.cpp
avtolstoy: Let's refactor this whole logic.

1. Read id
2. If succeeds - set the type and proceed further with flash-specific commands/logic
3. If fails, try a number of (or all) supported flash types and perform wake-up/op suspend etc commands
4. Try reading id again

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+        uint8_t* dst_aligned = (uint8_t*)ADDR_ALIGN_WORD_RIGHT((uintptr_t)data_buf);
+        const size_t unaligned_bytes = std::min(data_size, (size_t)(dst_aligned - data_buf));
+        const size_t size_aligned = ADDR_ALIGN_WORD(data_size - unaligned_bytes);
+
+        if (size_aligned > 0) {
+            /* Read-out the aligned portion into an aligned address in the buffer
+             * with an adjusted size in multiples of 4.
+             */
+            ret = nrfx_qspi_read(dst_aligned, size_aligned, src_aligned);
+            CHECK_TRUE(ret == NRFX_SUCCESS, nrf_system_error(ret));
```
avtolstoy, line 423: `CHECK_NRF`

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+    FlashLock lk;
+
+    if (flash_type == HAL_QSPI_FLASH_TYPE_GD25WQ64E) {
+        // Determine which OTP Register to start reading from
+        size_t otp_page = addr / GD25_SECURITY_REGISTER_SIZE;
+        size_t data_to_read = data_size;
+        size_t data_read = 0;
+        size_t read_amount = 0;
+
+        while (data_to_read && !ret) {
```
avtolstoy, line 518: Nit: move to some common code as it's reused below.

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+        }
+    } else if (sp == HAL_EXFLASH_SPECIAL_SECTOR_OTP) {
+        /* OTP-specific commands */
+        if (cmd == HAL_EXFLASH_COMMAND_LOCK_ENTIRE_OTP) {
+            /* Secured OTP Indicator bit (4K-bit Secured OTP)
+             * 1 = factory lock
+             * This will block the whole OTP region.
+             */
+            uint8_t v = 0x01;
+            if (flash_type == HAL_QSPI_FLASH_TYPE_MX25L3233F || flash_type == HAL_QSPI_FLASH_TYPE_MX25R6435F) {
```
avtolstoy, line 667: Error otherwise, add TODO.

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,77 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <stdint.h>
+#include "nrfx_qspi.h"
+
+#define MX25R6435F_OTP_SECTOR_SIZE      1024
```
avtolstoy, line 23: Nit/style: let's use some commond prefix e.g. `HAL_QSPI` as the rest of the file.

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,77 @@
...
+    HAL_QSPI_CMD_GD25_SEC_ERASE      = 0x44, // Security Register Erase (ie Erase OTP)
+    HAL_QSPI_CMD_GD25_SEC_READ       = 0x48, // Security Register Read
+    HAL_QSPI_CMD_GD25_SEC_PROGRAM    = 0x42, // Security Register Program (ie Write OTP)
+    HAL_QSPI_CMD_GD25_PGMERS_SUSPEND = 0x75, // Program/Erase Suspend
+    HAL_QSPI_CMD_GD25_WAKE           = 0xAB, // Release from Deep Power-Down and Read Device ID
+} hal_qspi_flash_cmd_t;
+
+typedef struct  {
+    hal_qspi_flash_type_t   type;            // hal_qspi_flash_type_t to identify what part is present
+    nrf_qspi_writeoc_t      write_opcode;    // Which opcode to use for quad write
```
avtolstoy, line 61: Nit: we'll need to refactor this later to be non-Gen3-specific.

avtolstoy (changes requested):

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,717 @@
...
+
+    // Use single data line SPI to read ID
+    nrfx_qspi_config_t default_config = NRF_QSPI_QUICK_CFG(NRF_QSPI_WRITEOC_PP, NRF_QSPI_READOC_FASTREAD, NRF_QSPI_FREQ_32MDIV2);
+    int ret = nrfx_qspi_init(&default_config, nullptr, nullptr);
+    if (ret) {
+        if (ret == NRFX_ERROR_TIMEOUT) {
+            // Could have timed out because the flash chip might be in an ongoing program/erase operation.
+            // Suspend it.
+            LOG_DEBUG(TRACE, "QSPI NRFX timeout error. Suspend pgm/ers op.");
+            ret = hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, nullptr, nullptr, 0);
```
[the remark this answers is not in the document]
YutingYou: Suspend operation is still needed, we should add the reset operation after it.

hal/src/nRF52840/exflash_hal.cpp
YutingYou: @scott-brust Read ID should be supported in the Deep Power-Down Mode.

>[GD25Q32C Datasheet]
>The Deep Power-Down Mode can only be entered by executing the Deep Power-Down (DP) command. Once the device has entered the Deep Power-Down Mode, all commands are ignored except the Release from Deep Power-Down and Read Device ID (RDI) (ABH) or Enable Reset (66H) and Reset (99H) commands.

By the way, I would consider that the read status command(RDSR), which is used in `nrfx_qspi_init()` would fail in Deep Power-Down Mode.

hal/src/nRF52840/exflash_hal.cpp
scott-brust: Addressed this, let me know if refactor is ok

hal/src/nRF52840/exflash_hal_params.h
scott-brust: Addressed this to be non-NRF specific

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+                    CHECK(exflash_qspi_cinstr_quick_send(HAL_QSPI_CMD_MX25_WRSCUR, get_nrf_cinstr_length(2), &v));
+                } else {
+                    // TODO: Implement for GD25 flash chip
+                    return SYSTEM_ERROR_NOT_SUPPORTED;
+                }
+                break;
+            };
+            case HAL_EXFLASH_COMMAND_GET_OTP_SIZE: {
+                const hal_exflash_params_t* flash_params = hal_exflash_get_params(flash_type);
+                return flash_params->otp_size;
```
scott-brust, line 693: Is it ok to return the OTP size directly from the function? Or should I set the value to `result`?

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+    CHECK_TRUE(sp == HAL_EXFLASH_SPECIAL_SECTOR_OTP, SYSTEM_ERROR_INVALID_ARGUMENT);
+    CHECK_TRUE(addr + data_size <= flash_params->otp_size, SYSTEM_ERROR_OUT_OF_RANGE);
+    CHECK_TRUE(data_buf != nullptr, SYSTEM_ERROR_INVALID_ARGUMENT);
+    int ret = 0;
+
+    FlashLock lk;
+
+    if (flash_type == HAL_QSPI_FLASH_TYPE_GD25WQ64E) {
+       ret = gd25_security_register_operation(gd25_secure_register_write, addr, (uint8_t *)data_buf, data_size);
```
scott-brust, line 587: I had to drop the `const` in order to provide a common function signature for the common security register read/write function, but re-cast the data buffer back to const when passing it to the NRF SDK function that uses it to write the data

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+    hal_qspi_flash_type_t all_flashes[] = {
+        HAL_QSPI_FLASH_TYPE_MX25L3233F,
+        HAL_QSPI_FLASH_TYPE_MX25R6435F,
+        HAL_QSPI_FLASH_TYPE_GD25WQ64E
+    };
+
+    // Use single data line SPI to read ID
+    nrfx_qspi_config_t default_config = NRF_QSPI_QUICK_CFG(NRF_QSPI_WRITEOC_PP, NRF_QSPI_READOC_FASTREAD, NRF_QSPI_FREQ_32MDIV2);
+    int ret = nrfx_qspi_init(&default_config, nullptr, nullptr);
+    if (ret == NRFX_ERROR_TIMEOUT) {
```
scott-brust, line 386: When the GD25 is in deep sleep, `nrfx_qspi_init` does not return an error code. The first indication of the chip being in deep sleep is when the `HAL_EXFLASH_COMMAND_READID` command below fails. 
I believe the `NRFX_ERROR_TIMEOUT` error condition only happens when the chip really is doing a program/erase operation, so if this happens do the `SUSPEND_PGMERS` operation for all known chips

scott-brust:
Tested latest changes with boron + GD25 flash, read/write OTP works, sleep/wake works.
Will test with boron +MXIC flash as well

hal/src/nRF52840/exflash_hal.cpp
YutingYou: 👍 @scott-brust I reproduced this behavior, it is different from the document, the document shows the status register can be read at any time. Also agree that we don't need to return an error since we have `SUSPEND_PGMERS` + `RESET`.

Should we handle other error conditions of `nrfx_qspi_init()`?

YutingYou (approved):

hal/src/nRF52840/exflash_hal.cpp
avtolstoy: Probably better to use the result.

hal/src/nRF52840/exflash_hal_params.h
```diff
@@ -0,0 +1,85 @@
...
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include <stdint.h>
+#include "nrfx_qspi.h"
+
+// Default flash type to support at least standard flash commands
+#define DEFAULT_SPI_FLASH_TYPE                      HAL_QSPI_FLASH_TYPE_MX25R6435F
```
avtolstoy, line 24: Prefix with `HAL_QSPI`

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+
+int hal_exflash_init(void) {
+    // PATCH: Initialize CS pin, external memory discharge
+    nrf_gpio_cfg_output(QSPI_FLASH_CSN_PIN);
+    nrf_gpio_pin_set(QSPI_FLASH_CSN_PIN);
+    nrf_gpio_cfg_input(QSPI_FLASH_IO1_PIN, NRF_GPIO_PIN_PULLDOWN);
+
+    FlashLock lk;
+
+    hal_qspi_flash_type_t all_flashes[] = {
```
avtolstoy, line 377: Nit: This should probably be coming from exflash_hal_params or even perhaps the platform config file?

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+            nrf_delay_us(20); // Need delay to ensure the flash is suspended
+        }
+    }
+
+    // Determine chip type / wake up from potential sleep
+    uint8_t chip_id[3] = {};
+
+    // Temporarily set the flash type to make use of hal_exflash_special_command()
+    flash_type = DEFAULT_SPI_FLASH_TYPE;
+    if (!hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_READID, nullptr, chip_id, 0)) {
```
avtolstoy, line 404: `size` as `sizeof(chip_id)` ? Same below.

`if (!` is checking for success, it should be the opposite, right?

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+    if (ret == NRFX_ERROR_TIMEOUT) {
+        // Could have timed out because the flash chip might be in an ongoing program/erase operation or Deep Power-Down mode.
+        // Suspend it or wake it up.
+        LOG_DEBUG(TRACE, "QSPI NRFX timeout error. Suspend pgm/ers op or wake it up.");
+
+        // Try all the supported suspend command
+        for (auto flash: all_flashes) {
+            flash_type = flash;
+            hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_SUSPEND_PGMERS, nullptr, nullptr, 0);
+            nrf_delay_us(20); // Need delay to ensure the flash is suspended
```
avtolstoy, line 395: Might perhaps be better doing this inside `hal_exflash_special_command`. Same below

hal/src/nRF52840/exflash_hal.cpp
```diff
@@ -0,0 +1,739 @@
...
+            flash_type = flash;
+            hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_WAKEUP, nullptr, nullptr, 0);
+            nrf_delay_us(20); // Need delay to ensure the flash is woken up.
+        }
+
+        // Read ID must succeed at this point
+        CHECK(hal_exflash_special_command(HAL_EXFLASH_SPECIAL_SECTOR_NONE, HAL_EXFLASH_COMMAND_READID, nullptr, chip_id, 0));
+    }
+
+    flash_type = hal_exflash_get_type(chip_id);
```
avtolstoy, line 416: Unchecked value, we should probably error out in case of unknown flash type.

scott-brust:
Tested both MXIC and GD25 flash parts with Sleep, OTP read/write, and new `HAL_EXFLASH_COMMAND_GET_OTP_SIZE`. 

Had to tweak the `init` again, `hal_exflash_special_command` returns success even if the flash part is asleep and does not return a non-null chip id (I suppose the NRF driver cannot distinguish if a custom instruction failed or legitimately returned all `00` in response).

hal/src/nRF52840/exflash_hal.cpp
scott-brust: Thanks for catching this. `size` was also unused and unchecked. Tweaked that in last commit

scott-brust: To be clear, the `hal_exflash_special_command` does not return a failure code, but rather we assume it failed if we get a NULL device_id, then try the wake routine and query device ID again.

hal/src/nRF52840/exflash_hal.cpp
avtolstoy: Yup, it does put things into the right state and it's exactly how you imagine: we've seen resets (either with a button or during OTA or especially brownouts) triggering a condition where the flash is busy on next boot and we have to explicitly abort the operation.

avtolstoy:
@scott-brust One additional thing to test is sleep: the flash should correctly be put into suspended/low power mode and correctly re-initialized on wake-up.

avtolstoy (approved):

scott-brust:
> @scott-brust One additional thing to test is sleep: the flash should correctly be put into suspended/low power mode and correctly re-initialized on wake-up (STOP or ULP modes).

Ok, I tested sleep using `hal_exflash_sleep()`, which is what appears to be used in `sleep_hal.cpp`, so I'm assuming that is good enough.

avtolstoy:
@scott-brust This is a pretty major change affecting not only new devices with GD25 part, but all other Gen 3 platforms as well, so let's do a proper test with STOP/ULP modes. I'd suggest going through `wiring/sleep20` suite on Boron/BSoM/Argon (any one), B5 SoM/Tracker (any one) and devices with new GD25 flash.

scott-brust:
> @scott-brust This is a pretty major change affecting not only new devices with GD25 part, but all other Gen 3 platforms as well, so let's do a proper test with STOP/ULP modes. I'd suggest going through `wiring/sleep20` suite on Boron/BSoM/Argon (any one), B5 SoM/Tracker (any one) and devices with new GD25 flash.

Good point, ill run `wiring/sleep20` on both my boards locally. If that looks ok ill merge and we can run on wider fleet of devices on HIL

avtolstoy:
@scott-brust Note that sleep tests are manual and not runnable in automated fashion.

scott-brust:
Tested the following permutations with this branch

- BORON404X - flash-detect branch bootloader + OS + app
    - GD25 Sleep20: All passed except final cellular test
    -  MXIC Sleep20:  All passed except final cellular test
- Argon (wake from GPIO)
    - Flash detect bootloader + 3.3.0 System part
        - HIBERNATE, ULTRA_LOW_POWER,STOP : OK
    - Flash detect bootloader + 2.3.0 System part:
        - HIBERNATE, ULTRA_LOW_POWER,STOP : OK
- Tracker (Wake from duration)
    - Flash detect bootloader + 3.3.0 System part
        - HIBERNATE, ULTRA_LOW_POWER,STOP : OK
    - Flash detect bootloader + 2.3.0 System part:
        - HIBERNATE, ULTRA_LOW_POWER,STOP : OK

PR #1953, opened 2019-10-28
Refactors link scripts for all Particle platforms

XuGuohui:
#### Structure
```
device-os
    |---- build > arm > linker                   : Shared auxiliary linker scripts for linking bootloader, monolithic and modular firmware for all platforms
    |                      |-- (MCU)            : Shared auxiliary linker scripts for linking bootloader, monolithic and modular firmware for core-based platforms
    |                             |-- (platform) : Auxiliary linker scripts for linking bootloader, monolithic and modular firmware for specific platform
    |
    |---- bootloader > src > (MCU)              : Linker script entrance for linking the bootloader for core-based platforms
    |
    |---- hal > src
    |            |- (MCU)                       : Linker scripts entrance for linking the monolithic firmware for core-based platforms
    |            |- (platform)                   : Linker scripts entrance for linking the monolithic firmware for specific platform
    |
    |---- modules
             |--- shared > (MCU)                : Shared auxiliary linker scripts for linking the modular firmware for core-based platforms
             |--- (platform)
                       |---- system-part1        : Linker script entrance for linking the system part1 firmware for specific platform, and glue linker scripts for linking other parts of modular firmware.
                       |---- system-part2        : Linker script entrance for linking the system part2 firmware for specific platform, and glue linker scripts for linking other parts of modular firmware.
                       |---- system-part3        : Linker script entrance for linking the system part3 firmware for specific platform, and glue linker scripts for linking other parts of modular firmware.
                       |---- user-part           : Linker script entrance for linking the user part firmware for specific platform, and glue linker scripts for linking other parts of modular firmware.
```
#### Filename 

- Linker script entrance is always named `linker.ld`.
- Auxiliary linker scripts those defined output section's memory region are prefixed with `memory_`, e.g. `memory_backup_ram.ld`.
- Auxiliary linker scripts those defined platform's flash memory map are named `platform_flash.ld`.
- Auxiliary linker scripts those defined platform's RAM memory map are named `platform_ram.ld`.
- Other auxiliary linker scripts are prefixed with `linker_`, e.g. `linker_module_start.ld`.
- Glue linker scripts for modular building are prefixed with `module_`, e.g. `module_system_part1_export.ld`.

#### Symbols in linker script

- Symbols those are calculated by linker or referenced by source code should be prefixed with `link_`, e.g. `link_global_data_start`. This kind of symbols can also be referenced by other linker scripts if necessary, such as the symbols defined in glue linker scripts.
[...]

PR #2472, opened 2022-07-14
[linker] make sure that FreeRTOS task/timer symbols are coalesced in bss section, disable LTO for FreeRTOS sources as well

avtolstoy:
### Problem

Due to refactoring of linker files in #1953 (and those changes actually coming after P2 merge) we lost a quirk in linker files which coalesced FreeRTOS task/timer symbols into a single location in .bss for some of the tooling that we use (#2399). This also wasn't really working with LTO enabled as is the case with Gen 3 system part builds.

### Solution

Fix that.

### Steps to Test

Build out of this branch, check system-part map files, `pxCurrentTCB` and other FreeRTOS symbols should be located in `.bss` section between `__rtos_capture_start` and `__rtos_capture_end`.

### Example App

N/A

### References

- #2399
- #1852

---

### Completeness

- [x] Added documentation

### DIFF
--- a/build/arm/linker/linker_freertos_task_symbols.ld
+++ b/build/arm/linker/linker_freertos_task_symbols.ld
@@ -1,0 +1,4 @@
+__rtos_capture_start = .;
+*tasks.o(.bss COMMON .bss*)
+*timers*.o(.bss COMMON .bss*)
+__rtos_capture_end = .;
--- a/build/arm-tools.mk
+++ b/build/arm-tools.mk
@@ -95,8 +95,9 @@ CFLAGS += -DARM_CPU_$(shell echo $(ARM_CPU) | tr '-' '_' | tr 'a-z' 'A-Z')
 #
 # -fno-use-cxa-atexit makes sure that destructors for statically created C++ objects are never called,
 # which saves us some flash space.
-CPPFLAGS += -flto -ffat-lto-objects -DPARTICLE_COMPILE_LTO_FAT -fno-use-cxa-atexit
-CONLYFLAGS += -flto -ffat-lto-objects -DPARTICLE_COMPILE_LTO_FAT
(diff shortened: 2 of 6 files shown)


build/arm/linker/linker_freertos_task_symbols.ld
```diff
@@ -0,0 +1,4 @@
+__rtos_capture_start = .;
+*tasks.o(.bss COMMON .bss*)
+*timers*.o(.bss COMMON .bss*)
+__rtos_capture_end = .;
```
eberseth, line 4: add `croutines.o` as well?

build/arm/linker/linker_freertos_task_symbols.ld
```diff
@@ -0,0 +1,4 @@
+__rtos_capture_start = .;
```
eberseth, line 1: How about `.data` objects as well?

eberseth (approved):

build/arm/linker/linker_freertos_task_symbols.ld
avtolstoy: There is nothing non-zero-initialized in there:
```
 .bss.xTimerTaskHandle
                0x0000000020006ac0        0x4 ../../../build/target/freertos/platform-26-m//libfreertos.a(timers.o)
                0x0000000020006ac4                __rtos_capture_data = .
 *tasks.o(.data .data*)
 *timers*.o(.data .data*)
                0x0000000020006ac4                __rtos_capture_end = .
```

We can't really coalesce things into two different locations with different initialization schemes.

This is also not going to be changed in FreeRTOS as there is a `PRIVILEGED_DATA` functionality for MPUs that relies on the fact that everything is pretty much in bss as well.

build/arm/linker/linker_freertos_task_symbols.ld
eberseth: probably not

PR #2476, opened 2022-07-18
[rtl872x] DCache alignment requirement

XuGuohui:
A clone of https://github.com/particle-iot/firmware-private/pull/381.

### Problem
Buffers being used by DMA or between KM0 and KM4 should be 32-byte aligned, and the length of it should be multiple of 32 bytes, otherwise, data in SRAM might be corrupted.

### References
N/A

Tested:

1. Flash new prebootloadder and bootloader agianst the branch in this PR
2. Flash old system firmware that is against the `develop` branch.
3. Verified sleep functionalities.
4. Verified bootloader updating using `particle flash --serial`

"Old prebootloader/bootloader + new system firmware" is broken, but this normally won't happen given that users wil go 
 through the safe mode healer to update system firmware.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/spi_hal.cpp
+++ b/hal/src/rtl872x/spi_hal.cpp
@@ -231,6 +231,7 @@ public:
         // Start address of chunk buffer should be 32-byte aligned
         SPARK_ASSERT(((uint32_t)chunkBuffer_.txBuf & 0x1f) == 0);
         SPARK_ASSERT(((uint32_t)chunkBuffer_.rxBuf & 0x1f) == 0);
+        SPARK_ASSERT((CFG_CHUNK_BUF_SIZE) % 32 == 0);
         CHECK_TRUE(validateConfig(rtlSpiIndex_, config), SYSTEM_ERROR_INVALID_ARGUMENT);
 
         if (isEnabled()) {
@@ -485,7 +486,7 @@ public:
         if (chunkBuffer_.rxIndex >= bufferConfig_.rxLength) {
             chunkBuffer_.rxIndex = 0;
 
-            if (config_.spiMode == SPI_MODE_MASTER) {
+            if (config_.spiMode == SPI_MODE_MASTER && bufferConfig_.rxBuf) {
                 // FIXME: For SPI slave, the user callback will be called after CS pin is pulled high.
                 if (callbackConfig_.dmaUserCb) {
                     (*callbackConfig_.dmaUserCb)();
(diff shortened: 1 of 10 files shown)


hal/inc/usart_hal.h
```diff
@@ -30,13 +30,18 @@
...
 #include <unistd.h>
 #include "platforms.h"
+#include "hal_platform.h"
 
 /* Includes ------------------------------------------------------------------*/
 // #include "pinmap_hal.h"
 
 /* Exported defines ----------------------------------------------------------*/
 
+#if HAL_PLATFORM_UART_BUFFER_ALIGNMENT > 4
```
avtolstoy, line 40: This shouldn't be needed anymore and in general due to `portBYTE_ALIGNMENT` being 32 on P2.

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -542,8 +542,12 @@ class Spi {
                 bytesToCopy = std::min(bufferConfig_.rxLength - chunkBuffer_.rxIndex, dmaRxCount + fifoRxCount);
                 if (bufferConfig_.rxBuf) {
                     DCache_Invalidate((u32) chunkBuffer_.rxBuf, chunkBuffer_.rxLength);
+                    int primask = HAL_disable_irq();
+                    __ISB();
```
avtolstoy, line 546: This shouldn't be needed?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -708,8 +713,12 @@ class Spi {
         copyLength = std::min(bufferConfig_.rxLength - chunkBuffer_.rxIndex, (size_t)chunkBuffer_.rxLength);
         if (bufferConfig_.rxBuf) {
             DCache_Invalidate((u32) chunkBuffer_.rxBuf, chunkBuffer_.rxLength);
+            int primask = HAL_disable_irq();
+            __ISB();
```
avtolstoy, line 717: Same here.

avtolstoy (approved):
Rebase on `develop`, remove UART changes.

hal/src/rtl872x/spi_hal.cpp
XuGuohui: Only this line?

avtolstoy: There is no need for DSB/ISB barriers after CPSID. It's not a register write, it's a CPU instruction that is not pipelined.

avtolstoy: image

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -461,7 +461,10 @@ class Usart {
                     uart->receiving_ = false;
                     uint8_t temp[MAX_UART_FIFO_SIZE];
                     uint32_t inFifo = UART_ReceiveDataTO(uartInstance, temp, MAX_UART_FIFO_SIZE, 1);
-                    uart->rxBuffer_.put(temp, inFifo);
+                    const ssize_t canWrite = uart->rxBuffer_.space();
```
avtolstoy, line 464: This doesn't look right, we should problably:
```suggestion
                    const ssize_t canWrite = uart->rxBuffer_.space();
                    if (canWrite > 0) {
                        uint8_t temp[MAX_UART_FIFO_SIZE];
                        uint32_t inFifo = UART_ReceiveDataTO(uartInstance, temp, std::min<size_t>(canWrite, MAX_UART_FIFO_SIZE), 1);
                        uart->rxBuffer_.put(temp, inFifo);
                    }
```

avtolstoy (approved):

PR #2477, opened 2022-07-18
[rtl872x] usart: flush DMA FIFO data when required

avtolstoy:
### TODO

Review TX path as well.

### Problem

RTL872x DMA peripheral is not transferring data out of its internal FIFO (previously having transferred it in there from UART) into the destination buffer if FIFO contains less than 4 bytes, no matter the burst size, FIFO or other setting.

### Solution

Calculate how much data has been already transferred into destination buffer while DMA transaction is still ongoing, how much is still in DMA FIFO, commit into RX buffer when possible and whenever required force-flush < 4 bytes of data out of DMA FIFO into dest buffer using DMA channel suspend.

Read up on similar DMA peripheral here: https://www.infineon.com/dgdl/xmc4100_xmc4200_rm_v1.5_2014_04.pdf?fileId=db3a30433afc7e3e013b3c44ccd35c20&ack=t

### Steps to Test

N/A

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/rtl872x/usart_hal.cpp
```diff
@@ -585,16 +644,21 @@ class Usart {
             // Disable Rx interrupt
             UART_INTConfig(uartInstance, (RUART_IER_ERBI | RUART_IER_ELSI | RUART_IER_ETOI), DISABLE);
```
avtolstoy, line 645: Might want to add `RX_THRESHOLD` check similar to Gen 3: https://github.com/particle-iot/device-os/blob/develop/hal/src/nRF52840/usart_hal.cpp#L557

to avoid getting into weird ring buffer configurations with wraparound.

avtolstoy: Added to keep consistent between Gen 3 and RTL872x.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -279,44 +322,59 @@ class Usart {
...
-                const uint32_t receivedDma = curAddr - rxDmaInitStruct_.GDMA_DstAddr;
-                uint32_t inFifo = UART_ReceiveDataTO(uartInstance, (uint8_t*)curAddr, rxDmaInitStruct_.GDMA_BlockSize - receivedDma, 1);
-                uint32_t receivedTotal = receivedDma + inFifo;
-                // Clean Auto Reload Bit
-                GDMA_ChCleanAutoReload(rxDmaInitStruct_.GDMA_Index, rxDmaInitStruct_.GDMA_ChNum, CLEAN_RELOAD_DST);
-                // Clear Pending ISR
-                GDMA_ClearINT(rxDmaInitStruct_.GDMA_Index, rxDmaInitStruct_.GDMA_ChNum);
-                rxBuffer_.acquireCommit(receivedTotal, rxDmaInitStruct_.GDMA_BlockSize - receivedTotal/* cancel */);
-                len += receivedTotal;
-            } else {
```
XuGuohui, line 301: This code block is for interrupt mode.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -577,24 +640,29 @@ class Usart {
...
                 return;
             }
             rxSize = std::min(MAX_DMA_BLOCK_SIZE, rxSize);
             // Disable Rx interrupt
             UART_INTConfig(uartInstance, (RUART_IER_ERBI | RUART_IER_ELSI | RUART_IER_ETOI), DISABLE);
             auto ptr = rxBuffer_.acquire(rxSize);
-            DCache_CleanInvalidate((uint32_t)ptr, rxSize);
+
+            UART_RXDMAConfig(uartInstance, 4);
+	        UART_RXDMACmd(uartInstance, ENABLE);
```
XuGuohui, line 653: Nit: spaces

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -630,15 +698,15 @@ class Usart {
...
             return 0;
         }
-        uart->receiving_ = false;
         auto rxDmaInitStruct = &uart->rxDmaInitStruct_;
-        DCache_Invalidate((uint32_t)rxDmaInitStruct->GDMA_DstAddr, rxDmaInitStruct->GDMA_BlockSize);
         // Clean Auto Reload Bit
 	    GDMA_ChCleanAutoReload(rxDmaInitStruct->GDMA_Index, rxDmaInitStruct->GDMA_ChNum, CLEAN_RELOAD_SRC);
+        GDMA_Cmd(rxDmaInitStruct->GDMA_Index, rxDmaInitStruct->GDMA_ChNum, DISABLE);
         // Clear Pending ISR
 	    GDMA_ClearINT(rxDmaInitStruct->GDMA_Index, rxDmaInitStruct->GDMA_ChNum);
```
XuGuohui, line 706: Nit: spaces

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -630,15 +698,15 @@ class Usart {
...
-        GDMA_Cmd(rxDmaInitStruct->GDMA_Index, rxDmaInitStruct->GDMA_ChNum, DISABLE);
-        uart->rxBuffer_.acquireCommit(rxDmaInitStruct->GDMA_BlockSize);
+        UART_RXDMACmd(uartTable_[uart->index_].UARTx, DISABLE);
+        uart->dataInFlight(true /* commit */);
```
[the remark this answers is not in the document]
avtolstoy: I think instead we should add `cancel` argument to `dataInFlight()` and cancel-commit the extra buffer size. This way we keep the logic of dealing with transfers in `dataInFlight()` universally.

hal/src/rtl872x/usart_hal.cpp
avtolstoy: We don't normally transfer data out of internal peripheral FIFO in `data()`, so I'd argue we keep this as-is and let the interrupts put the data into the RX buffer and here just deal with whatever is already in it.

hal/src/rtl872x/usart_hal.cpp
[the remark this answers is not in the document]
XuGuohui: We should not cancel the pending in `dataInFlight()`, cos it might be called to commit data but a DMA transaction is not completed, e.g. in `read()`, unless we add one more argument to differentiate it.

hal/src/rtl872x/usart_hal.cpp
XuGuohui: I that case, should we call `UART_SetRxLevel(uartInstance, UART_RX_FIFOTRIG_LEVEL_1BYTES);` explicitly so that `Serial.available()` has value returned if there is data in FIFO but the ISR is not serviced.

hal/src/rtl872x/usart_hal.cpp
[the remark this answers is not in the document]
avtolstoy: Yes, that's the idea to add an argument to differentiate between the two cases but keep the functionality in a single place.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -279,44 +322,59 @@ class Usart {
...
 
+    size_t dataInFlight(bool commit = false) {
+        // Must be called under RX lock!
+        if (receiving_ && !useInterrupt()) {
+            auto uartInstance = uartTable_[index_].UARTx;
+            size_t transferredToDmaFromUart = uartInstance->RX_BYTE_CNT & RUART_RX_READ_BYTE_CNTER;
+            size_t dmaAvailableInBuffer = GDMA_GetDstAddr(rxDmaInitStruct_.GDMA_Index, rxDmaInitStruct_.GDMA_ChNum) - rxDmaInitStruct_.GDMA_DstAddr;
+            size_t blockSize = rxDmaInitStruct_.GDMA_BlockSize;
+            size_t alreadyCommitted = blockSize - rxBuffer_.acquirePending();
+            size_t toConsume = std::max(dmaAvailableInBuffer, transferredToDmaFromUart) - alreadyCommitted;
```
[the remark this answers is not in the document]
avtolstoy: It won't help though, we can add an assert here on negative value, but this generally should be impossible to trigger. Data transferred from UART FIFO into DMA FIFO and from DMA FIFO into destination buffer cannot exceed block size.

XuGuohui (changes requested):

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -322,15 +323,16 @@ class Usart {
         return 0;
     }
 
-    size_t dataInFlight(bool commit = false) {
+    size_t dataInFlight(bool commit = false, bool cancel = true) {
```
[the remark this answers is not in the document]
avtolstoy: ah, my bad, yeah

PR #2481, opened 2022-07-19
[tinker] only include fqc functionality if APP=tinker

scott-brust:
### Problem

Ensure FQC functionality is only compiled into tinker if built with `APP=tinker PLATFORM=p2`
All other builds will exclude it
### Steps to Test
CI should pass
tinker release builds should include FQC functionality on P2 platform
tinker builds with APPDIR=  should NOT include FQC 

### Example App

tinker
### References

---

### Completeness

- [ ] Added documentation


user/applications/tinker/src/fqc_test.cpp
```diff
@@ -1,3 +1,5 @@
+#include "hal_platform.h"
```
avtolstoy, line 1: Add license header

avtolstoy (approved):

avtolstoy (approved):

PR #2482, opened 2022-07-20
[p2][photon2] Add VBAT_MEAS pin and charging indication pin for Photon2

YutingYou:
### Problem

This PR is to add `VBAT_MEAS` and charging pin to P2 pinmap.

 - `VBAT_MEAS` is an ADC-dedicated pin, it can measure the input voltage with 5V tolerance,
 - Charging pin is only used on Lightcycle

### Solution

Add VBAT_MEAS as the A6 pin, note that the reference voltage is 5V instead of 3.3V

### Steps to Test

1. Re-compile and download bootloader/system-part1/user-app
2. Measure different voltage with A6 pin

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {
    int value = analogRead(A6);
    Log.info("A6 value: %d, volt: %f", value, value / 4096.0 * 5.0);

    int A0Value = analogRead(A0);
    Log.info("A0 Value: %d, volt: %f", A0Value, A0Value / 4096.0 * 3.3);
    delay(1000);
}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/gpio_hal.cpp
+++ b/hal/src/rtl872x/gpio_hal.cpp
@@ -68,6 +68,10 @@ bool isCachePin(hal_pin_t pin) {
     return false;
 }
 
+bool isVbatMeasPin(hal_pin_t pin) {
+    return(pin == A6) ? true : false;
+}
+
 bool isCachePinSetToOutput(hal_pin_t pin) {
     hal_pin_info_t* pinInfo = hal_pin_map() + pin;
     if ((pinInfo->pin_mode == OUTPUT) ||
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/gpio_hal.cpp
```diff
@@ -68,6 +68,10 @@ bool isCachePin(hal_pin_t pin) {
     return false;
 }
 
+bool isVbatMeasPin(hal_pin_t pin) {
```
avtolstoy, line 71: Perhaps generalize this to something like `isGpioPin()` and check for `!= RTL_PORT_NONE`?

avtolstoy (approved):

hal/src/rtl872x/gpio_hal.cpp
YutingYou: Updated.

tstellanova (approved):

PR #2484, opened 2022-07-20
[rtl872x] usart: UART_DEV2 (Serial1) actually works with DMA, no need to use interrupt mode

avtolstoy:
### Description

Well, as title says. For some reason the SDK is missing definitions for DMA hardware handshake identifiers for UART2_DEV (Serial1), so we had to use interrupt mode for Serial1. Apparently if we just add them (there is an obvious gap in the numbers), everything works just fine.

**NOTE: If shorting TX/RX together on boot the device will enter boot ROM. Is there any way to disable this?**

### Steps to Test

- `wiring/serial_loopback` ?
- `wiring/serial_loopback2` ?

### Example App

N/A

### References

- #2264
- https://github.com/particle-iot/realtek_ambd_sdk_public/compare/1dffc25c890010c4da1f7e5cfb8a77d817363e9a...8de686dcb70781d61d86f2c7554514d608112f8f

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/usart_hal.cpp
+++ b/hal/src/rtl872x/usart_hal.cpp
@@ -233,15 +233,10 @@ public:
             uartInstance->MCR &= ~ BIT(5);
         }
 
-        if (!useInterrupt()) {
-            // Configuring DMA
-            if (!initDmaChannels()) {
-                end();
-                return SYSTEM_ERROR_INTERNAL;
-            }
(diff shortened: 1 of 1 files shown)


avtolstoy:
Doesn't actually work after further testing. A dirty build previously just ran in interrupt mode :shrug:

PR #2486, opened 2022-07-20
[p2] fixes INTERRUPTS_01_isisr_willpreempt_servicedirqn test

scott-brust:
### Problem

`INTERRUPTS_01_isisr_willpreempt_servicedirqn` test fails on P2 due to unimplemented HAL

### Solution

Use an array to backup any overridden ISR handlers. When override is removed, restore the backup.

There are other parts of the HAL that are still unimplemented, namely `HAL_Core_Restore_Interrupt()` and the `link_interrupt_vectors_location` RAM table in general during `HAL_Core_Config()`.

This PR is still a WIP to facilitate discussion on how to approach these problems. 

### Steps to Test

Run `wiring/no_fixture`, specifically `INTERRUPTS_01_isisr_willpreempt_servicedirqn`

### Example App

Run `wiring/no_fixture`

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/interrupts_hal.cpp
+++ b/hal/src/rtl872x/interrupts_hal.cpp
@@ -258,36 +258,39 @@ int hal_interrupt_set_direct_handler(IRQn_Type irqn, hal_interrupt_direct_handle
         return 1;
     }
 
+    int error = 0;
     int32_t state = HAL_disable_irq();
 
     if (handler == nullptr && (flags & HAL_INTERRUPT_DIRECT_FLAG_RESTORE)) {
-        // Restore
+        // Restore old handler only if one was backed up
         uint32_t old_handler = hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)];
-        __NVIC_SetVector(irqn, (uint32_t)old_handler);
-
-        hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = 0;
-    } else {
-
-        // If there is currently a backup: Return error
-        if(hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)]){
-            return 1;
+        if (old_handler) {
+            __NVIC_SetVector(irqn, (uint32_t)old_handler);
+            hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = 0;
         }
-
-        // If there is a current handler, back it up
-        uint32_t current_handler = __NVIC_GetVector(irqn);
-        if (current_handler) {
-            hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = current_handler;    
+    } else {
+        // If there is currently a handler backup: Return error
+        if (hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)]) {
+            error = SYSTEM_ERROR_NOT_SUPPORTED;
+        } else {
+            // If there is a current handler, back it up
+            uint32_t current_handler = __NVIC_GetVector(irqn);
+            if (current_handler) {
+                hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = current_handler;    
+            }    
         }
     }
 
-    if (flags & HAL_INTERRUPT_DIRECT_FLAG_DISABLE) {
-        // Disable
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -258,20 +259,32 @@ int hal_interrupt_set_direct_handler(IRQn_Type irqn, hal_interrupt_direct_handle
...
+        uint32_t current_handler = __NVIC_GetVector(irqn);
+        if (current_handler) {
+            hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = current_handler;    
+        }
     }
 
     if (flags & HAL_INTERRUPT_DIRECT_FLAG_DISABLE) {
         // Disable
-        // sd_nvic_DisableIRQ(irqn);
+        //__NVIC_DisableIRQ(irqn); // actually disable the interrupt (ie like systick!?)
```
avtolstoy, line 285: SysTick is non-maskable, it's a special case. Ignore it, uncomment NVIC_DisableIRQ() here.

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -258,20 +259,32 @@ int hal_interrupt_set_direct_handler(IRQn_Type irqn, hal_interrupt_direct_handle
     }
 
     int32_t state = HAL_disable_irq();
-    volatile uint32_t* isrs = (volatile uint32_t*)&link_ram_interrupt_vectors_location;
 
     if (handler == nullptr && (flags & HAL_INTERRUPT_DIRECT_FLAG_RESTORE)) {
         // Restore
-        // HAL_Core_Restore_Interrupt(irqn);
+        uint32_t old_handler = hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)];
```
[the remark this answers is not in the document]
XuGuohui: Or considering use `std::pair(irqn, handler)` along with `vector()`.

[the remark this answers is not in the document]
avtolstoy: I'd avoid using `vector` here due to dynamic allocations. VTOR size is known and negligible, so we can keep the whole thing preallocated.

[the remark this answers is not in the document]
scott-brust: Is there ever a valid use case to have a NULL interrupt handler? Can I just check if the backed up handler is not null, and only restore it if something has been saved?

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -257,24 +258,39 @@ int hal_interrupt_set_direct_handler(IRQn_Type irqn, hal_interrupt_direct_handle
...
+        if (old_handler) {
+            __NVIC_SetVector(irqn, (uint32_t)old_handler);
+            hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)] = 0;
+        }
     } else {
-        isrs[IRQN_TO_IDX(irqn)] = (uint32_t)handler;
+        // If there is currently a handler backup: Return error
+        if (hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)]) {
+            error = SYSTEM_ERROR_NOT_SUPPORTED;
```
scott-brust, line 274: It might actually simplify the logic if I use `goto` to bail out rather than have to check for error conditions? I cant `return` here because I need to re-enable interrupts.

avtolstoy: Use `CHECK_TRUE`/`CHECK_FALSE` then, much better than `goto`, so:

```c++
CHECK_FALSE(hal_interrupts_handler_backup[IRQN_TO_IDX(irqn)], SYSTEM_ERROR_ALREADY_EXISTS);
```

avtolstoy (approved):

hal/src/rtl872x/interrupts_hal.cpp
scott-brust: Ah yes, those macros are handy. Unfortunately I need to re-enable interrupts before exiting. ill just do that and return `SYSTEM_ERROR_ALREADY_EXISTS`. I think thats cleaner than adding a return variable and checking it, etc.

avtolstoy (approved):

PR #2491, opened 2022-07-25
[p2] Simplify burn in LED blink, work around System.millis() reset

scott-brust:
### Problem

The P2 burn in process is supposed to indicate how long it has run by blinking the LED every 1s. An alternating green/blue pattern indicates the tests are still running and passing.
 
For burn in hours X:
x >= 72h : 6 green blinks 4 blue
72h > x >= 48h : 7 green blinks 3 blue 
48h > x >= 24h : 8 green blinks, 2 blue
x < 24h : 9 green blinks, 1 blue

The current LED blink logic is convoluted, and relies on the `System.millis()` value monotonically increasing. Unfortunately there is also a bug right now where it resets sometimes, which causes the LED logic to break and stop blinking all together. 

### Solution

Simplify LED blink logic. Just use `delay()` since LED loop has its own thread. No need to rely on `System.millis()`

### Steps to Test

Build tinker, run burnin

### Example App

Tinker

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/applications/tinker/src/burnin_test.cpp
+++ b/user/applications/tinker/src/burnin_test.cpp
@@ -109,9 +109,6 @@ void BurninTest::setup(bool forceEnable) {
 	}
 
 	RGB.control(true);
-	start_time_millis_ = System.millis();
-	next_blink_millis_ = start_time_millis_ + BLINK_PERIOD_MILLIS;
-	next_status_report_millis_ = start_time_millis_ + BLINK_PERIOD_STATUS_REPORT_MILLIS;
     os_thread_create(&led_thread_, "led", OS_THREAD_PRIORITY_DEFAULT, ledLoop, this, OS_THREAD_STACK_SIZE_DEFAULT);
 }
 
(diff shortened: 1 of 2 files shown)


user/applications/tinker/src/burnin_test.cpp
```diff
@@ -182,56 +179,40 @@ static int blueBlinksFromRuntime(uint64_t run_time_millis) {
...
+	RGB.color(red, green, blue);
+	delay(blink_time / 2);
+	RGB.color(0, 0, 0);
+	delay(blink_time / 2);
 }
 
 // Blink LED / signal uptime / failure state
 void BurninTest::ledLoop(void * arg) {
-    BurninTest* self = static_cast<BurninTest*>(arg);
+	uint32_t total_runtime_millis = 0;
```
avtolstoy, line 191: Nit: `system_tick_t` (32-bit) or `uint64_t`

avtolstoy (approved):

PR #2493, opened 2022-07-27
RTL872x: Implement HAL event group for UART

XuGuohui:
### Problem
For platforms, like TrackerM, rely on the UART HAL event group APIs, but they are not implemented yet.

### Solution
Implements the UART HAL event group APIs. This PR also implements the UART HAL sleep API.

### Example App
 This test app is for verifying the HAL sleep API.
```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

#define SERIAL                  Serial2
#define SERIAL_INTERFACE        HAL_USART_SERIAL2

void setup() {
    SERIAL.begin(115200);
    pinMode(D11, INPUT_PULLUP);
}

void loop() {
    if (digitalRead(D11) == LOW) {
        delay(1s);
        SERIAL.printlnf("available1: %d", SERIAL.available());
        hal_usart_sleep(SERIAL_INTERFACE, true, nullptr);
        while (digitalRead(D11) == LOW);
        hal_usart_sleep(SERIAL_INTERFACE, false, nullptr);
        delay(1s);
        SERIAL.printlnf("available2: %d", SERIAL.available());
        while (SERIAL.available()) {
            SERIAL.write(SERIAL.read());
        }
        SERIAL.println(" ");
    }

    SERIAL.println("This is loop()");
    delay(1s);
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/usart_hal.cpp
+++ b/hal/src/rtl872x/usart_hal.cpp
@@ -412,6 +412,51 @@ public:
         return SYSTEM_ERROR_NONE;
     }
 
+    EventGroupHandle_t eventGroup() {
+        return evGroup_;
+    }
+
+    int enableEvent(HAL_USART_Pvt_Events event) {
+        if (event & HAL_USART_PVT_EVENT_READABLE) {
+            if (data() <= 0) {
+                rxLock(false);
+            } else {
+                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
+            }
+        }
+
+        if (event & HAL_USART_PVT_EVENT_WRITABLE) {
+            if (space() <= 0) {
+                txLock(false);
+            } else {
+                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
+            }
+        }
+
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int disableEvent(HAL_USART_Pvt_Events event) {
+        if (event & HAL_USART_PVT_EVENT_READABLE) {
+            rxLock(true);
+            xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
+        }
+
+        if (event & HAL_USART_PVT_EVENT_WRITABLE) {
+            txLock(true);
+            xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
+        }
+
+        return SYSTEM_ERROR_NONE;
+    }
+
+    int waitEvent(uint32_t events, system_tick_t timeout) {
+        CHECK(enableEvent((HAL_USART_Pvt_Events)events));
+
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/usart_hal.cpp
```diff
@@ -717,6 +826,10 @@ class Usart {
         uart->dataInFlight(true /* commit */, true /* cancel */);
         uart->receiving_ = false;
         uart->startReceiver();
+        BaseType_t yield = pdFALSE;
+        if (xEventGroupSetBitsFromISR(uart->evGroup_, HAL_USART_PVT_EVENT_READABLE, &yield) != pdFAIL) {
```
avtolstoy, line 830: This doesn't look right. This interrupt handler only generates at the end of the full DMA transfer. This event should be generated once we receive even just 1 byte.

XuGuohui: In that case, we need to enable rx interrupt when using  DMA mode and only set this event bit in the ISR. Sounds good?

avtolstoy: Yes, that's the idea (it's the same on Gen 3). And disable immediately once it happens. Also this interrupt should ideally be triggered on transfer out of UART FIFO not on when the byte appears in UART FIFO. + We need to syncup the data in flight on getting unblocked by this event.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -437,9 +440,7 @@ class Usart {
 
     int disableEvent(HAL_USART_Pvt_Events event) {
         if (event & HAL_USART_PVT_EVENT_READABLE) {
-            RxLock lk(this);
-            auto uartInstance = uartTable_[index_].UARTx;
-            uartInstance->MISCR &= ~(RUART_RXDMA_OWNER);
+            rxLock(true);
```
avtolstoy, line 443: This is not correct. This should acquire RX lock, and
```c++
UART_RXDMACmd(uartInstance, ENABLE);
```

avtolstoy (changes requested):

keeramis (approved):
Verified on test runner and through user app that the new event group implementation is working 🎉

avtolstoy (changes requested):

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -412,6 +408,68 @@ class Usart {
...
+                UART_RXDMACmd(uartInstance, DISABLE);
+            } else {
+                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
+            }
+        }
+
+        if (event & HAL_USART_PVT_EVENT_WRITABLE) {
+            if (space() <= 0) {
+                TxLock lk(this);
+                UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
```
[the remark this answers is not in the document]
XuGuohui: Needs to check `receiving_` as well?

avtolstoy (approved):

PR #2498, opened 2022-08-03
Improve logging

sergeuz:
### Problem

This PR improves logging of CoAP messages.

### References

- sc-107290


technobly (approved):
LGTM, do we want to enable this via DEBUG_BUILD=y or always have it enabled by default (most debugging gets enabled as TRACE or ALL, this is enabled with level >= TRACE).

PR #2499, opened 2022-08-03
[rtl872x] Poll micros instead of rom DelayUs function. Tweak 1ms test

scott-brust:
### Problem

`DELAY_01_delay_1ms_is_within_tolerance` fails occasionally on p2 `no_fixture` tests, ie the 1ms delay ends up being out of the 6% margin of error ( > 1060 us). 

### Solution

Changing the `HAL_Delay_Microseconds()` implementation to a simple poll of `HAL_Timer_Get_Micro_Seconds()` seems to be no worse than using the rom `DelayUs()` function.
Also tweaked the `system_delay_pump()` logic when doing a 1ms delay, which is what happens in the DELAY_01 test. 

However, the 6% threshold is still too short in some situations. Doubling it seems to ensure tests pass a higher % of the time. I still see an occasional failure where the delay took >1120 us, but this happened once in 300 test iterations. Not sure if this is infrequent enough to ignore or if the threshold should be raised even higher.

### Steps to Test

Run wiring/no_fixture tests

### Example App

wiring/no_fixture

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_task.cpp
+++ b/system/src/system_task.cpp
@@ -544,16 +544,20 @@ void system_delay_pump(unsigned long ms, bool force_no_background_loop=false)
             break;
         }
         else if (elapsed_millis >= (ms-1)) {
-            // on the last millisecond, resolve using millis - we don't know how far in that millisecond had come
+            // on the last millisecond, resolve using micros - we don't know how far in that millisecond had come
             // have to be careful with wrap around since start_micros can be greater than end_micros.
 
-            for (;;)
-            {
-                system_tick_t delay = end_micros-HAL_Timer_Get_Micro_Seconds();
-                if (delay>100000)
-                    return;
-                HAL_Delay_Microseconds(min(delay/2, 1u));
+            system_tick_t start_micros = HAL_Timer_Get_Micro_Seconds();
+            system_tick_t delay = end_micros - start_micros;
+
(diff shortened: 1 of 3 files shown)


system/src/system_task.cpp
```diff
@@ -544,16 +544,20 @@ void system_delay_pump(unsigned long ms, bool force_no_background_loop=false)
...
-                if (delay>100000)
-                    return;
-                HAL_Delay_Microseconds(min(delay/2, 1u));
+            system_tick_t start_micros = HAL_Timer_Get_Micro_Seconds();
+            system_tick_t delay = end_micros - start_micros;
+
+            // If delay duration extends across the micro rollover, 
+            // account for micros remaining before rollover as well as micros after rollover
+            if (end_micros < start_micros) {
+                delay = (0xFFFFFFFF - start_micros) + end_micros;
```
scott-brust, line 556: Is there a better symbol for `system_tick_t` MAX value?

avtolstoy: Might use `hal_timer_micros(nullptr)` which is 64-bit. Alternatively probably `std::numeric_limits<system_tick_t>::max()`

avtolstoy (approved):

PR #2501, opened 2022-08-04
[rtl872x] Fix millis() rollover

scott-brust:
### Problem

`HAL_Timer_Get_Milli_Seconds()` / `HAL_Timer_Get_Micro_Seconds()` Eventually reset after ~1 hour of execution. This results in `millis()` rolling over to zero. 

### Solution

It appears that the problem is due to integer promotion (in this case, demotion?) when calculating elapsed micro seconds:
```
constexpr int US_PER_OVERFLOW = 65536;                 // 65536us
...
return usSystemTimeMicrosBase_ + state.overflowCounter * US_PER_OVERFLOW + ticksToTime(state.curTimerTicks);
```

Making `US_PER_OVERFLOW` a uint64_t seems to solve the problem.

### Steps to Test

Modify `timer_hal.cpp` `init()` and `start()` functions to start the `overflowCounter_` at a value close to 65536
```
        overflowCounter_ = 65000;
```

Run the app and observe the uptime millis incrementing beyond `0004294659`

### Example App

tinker or any app with timestamp logging

### References

---

### Completeness

- [ ] Added documentation


avtolstoy (approved):
:tada:

PR #2502, opened 2022-08-11
[rtl872x] hal: fix issue that uart rx dma may hang up.

XuGuohui:
### Problem

When running the Serial2 stress test app as attached below, device may hang up over time.

### Solution

1. Implement recursive TX/RX lock
2. Fix minor bug in ring buffer
3. Do not suspend DMA unless necessary.

### Steps to Test

Build and run the below test app.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);
SYSTEM_THREAD(ENABLED);

Serial1LogHandler log(115200, LOG_LEVEL_ALL, {
    {"system", LOG_LEVEL_WARN}
});

constexpr uint16_t buffLen = 64;
uint8_t txBuff[buffLen];
uint8_t rxBuff[buffLen] = {};

system_tick_t was = 0;

void setup() {
    Serial2.begin(115200);
    for (int i = 0; i < buffLen; i++) {
        txBuff[i] = (uint8_t)i;
    }
    was = millis();
}

bool done = true;
size_t txLen = 0;
size_t rxCnt = 0;

void loop() {
    if (done) {
        txLen = random(1, buffLen);
        txLen = Serial2.write(txBuff, txLen);
        Log.printf("TX Done: %d\r\n", txLen);
        done = false;
        was = millis();
        rxCnt = 0;
    }

    size_t len = Serial2.available();
    if (len) {
        while (len--) {
            rxBuff[rxCnt] = Serial2.read();
            Log.printf("%02d ", rxBuff[rxCnt]);
            rxCnt++;
        }
    }

    if (millis() - was > 800) {
        was = millis();
        Log.printf("Done \r\n");
        done = true;
        if (rxCnt == txLen && !memcmp(txBuff, rxBuff, txLen)) {
            Log.info("Test Passed! len: %d", rxCnt);
        } else {
            Log.info("Test Failed! TX: %d, RX: %d", txLen, rxCnt);
            Serial2.breakTx();
            while (1);
        }
    }
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/usart_hal.cpp
+++ b/hal/src/rtl872x/usart_hal.cpp
@@ -81,27 +81,39 @@ public:
     class RxLock {
     public:
         RxLock(Usart* instance)
-                : uart_(instance) {
+                : uart_(instance),
+                  locked_(0) {
             uart_->rxLock(true);
+            locked_++;
         }
         ~RxLock() {
-            uart_->rxLock(false);
+            locked_--;
+            if (locked_ == 0) {
+                uart_->rxLock(false);
+            }
         }
     private:
         Usart* uart_;
+        uint32_t locked_;
     };
 
     class TxLock {
     public:
         TxLock(Usart* instance)
-                : uart_(instance) {
+                : uart_(instance),
+                  locked_(0) {
             uart_->txLock(true);
+            locked_++;
         }
         ~TxLock() {
-            uart_->txLock(false);
+            locked_--;
+            if (locked_ == 0) {
+                uart_->txLock(false);
+            }
         }
     private:
         Usart* uart_;
+        uint32_t locked_;
     };
 
     bool isEnabled() const {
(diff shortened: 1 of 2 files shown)


services/inc/ringbuffer.h
```diff
@@ -310,7 +310,7 @@ inline ssize_t RingBuffer<T>::acquireCommit(size_t size, size_t cancel) {
 
     headPending_ -= (size + cancel);
     head_ = wrap(head_ + size, curSize_);
-    full_ = ((head_ == tail_) && size > 0);
+    full_ = ((head_ == tail_) && size > 0) || full_;
```
avtolstoy, line 313: Nit: let's change this for readability:
```c++
if (size > 0) {
    full_ = (head_ == tail_);
}
```

We should only be updating `full_` flag if we are modifying `head_`.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -327,27 +339,30 @@ class Usart {
...
-                            GDMA_BASE->CH[rxDmaInitStruct_.GDMA_ChNum].CFG_LOW &= ~(BIT_CFGX_LO_CH_SUSP);
-                            __DSB();
-                            __ISB();
+                if (hal_interrupt_is_isr()) {
+                    // This method is called from DMA RX ISR
+                    toConsume = transferredToDmaFromUart - alreadyCommitted;
+                    flushDmaRxFiFo(transferredToDmaFromUart);
+                } else {
+                    toConsume = dmaAvailableInBuffer - alreadyCommitted;
+                    // Try not suspending DMA unless necessary
```
avtolstoy, line 361: :+1:

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -327,27 +339,30 @@ class Usart {
...
             auto uartInstance = uartTable_[index_].UARTx;
-            size_t transferredToDmaFromUart = uartInstance->RX_BYTE_CNT & RUART_RX_READ_BYTE_CNTER;
             size_t dmaAvailableInBuffer = GDMA_GetDstAddr(rxDmaInitStruct_.GDMA_Index, rxDmaInitStruct_.GDMA_ChNum) - rxDmaInitStruct_.GDMA_DstAddr;
             size_t blockSize = rxDmaInitStruct_.GDMA_BlockSize;
             size_t alreadyCommitted = blockSize - rxBuffer_.acquirePending();
+            size_t transferredToDmaFromUart = uartInstance->RX_BYTE_CNT & RUART_RX_READ_BYTE_CNTER;
+            if (hal_interrupt_is_isr()) {
+                // This method is called from DMA RX ISR
+                // Clear UART RX FIFO read counter ASAP, because UART might be still receiving data 
+                uartInstance->RX_BYTE_CNT |= RUART_RX_BYTE_CNTER_CLEAR;
```
avtolstoy, line 349: This should not be necessary to perform here. DMA will not be reading data out of UART FIFO if the transaction just finished. The counter will stay the same. This is not data put into UART FIFO, this is data transmitted from UART FIFO into DMA FIFO.

avtolstoy (approved):

hal/src/rtl872x/usart_hal.cpp
XuGuohui: It is possible that other interrupts with higher priority may interrupt the executtion of the UART RX DMA ISR, while the UART is receiving data. If we clear the counter later in the DMA ISR, we might lose some bytes, especially with higher baudrate.

avtolstoy: We won't. We are resetting RX_BYTE_CNT only at the end of the DMA transaction. Once DMA transaction finishes RX_BYTE_CNT will no longer increase because DMA no longer transfers data out of UART FIFO into its internal FIFO.

XuGuohui: I  see 👍 . Accordding to the `REG_RX_BYTE_CNT`'s description:
>Counting the byte number of data reading from Rx FIFO

avtolstoy: Yes, it's an actual number of bytes already _transferred from UART FIFO into DMA FIFO_, so already in DMA FIFO.

avtolstoy (approved):

PR #2503, opened 2022-08-11
[p2][gen3] fixes for attachInterrupt() / noInterrupts() / interrupts() APIs

technobly:
### Problem

- [P2] attachInterrupt() failing to call interrupt handler after using noInterrupts()/interrupts()
- [gen3] interrupts() API should not clear pending interrupts

### Solution

- [P2] The fact that the interrupts() API cleared the pending interrupts also disabled the currently attached interrupt upon re-enabling global interrupts.  Simply removing the clearPending call resolves it.
- [gen3] pending interrupts was being cleared unintentionally.  Simply removing the clearPending call resolves it.

### Steps to Test

- run `TEST=app/interrupts` on all platforms, jumpering a wire from A5 to D0 ~ D6 one at a time and resetting the device between tests.

### References

- [sc-107555](https://app.shortcut.com/particle/story/107555/p2-attachinterrupt-fails-to-call-interrupt-handler)

#### BUGFIXES

- [P2] fixes attachInterrupt() failing to call interrupt handler after using noInterrupts()/interrupts()
- [gen3] fixes interrupts() API, should not clear pending interrupts

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/app/interrupts/interrupts.cpp
+++ b/user/tests/app/interrupts/interrupts.cpp
@@ -1,0 +1,271 @@
+#include "Particle.h"
+SYSTEM_MODE(SEMI_AUTOMATIC);
+SerialLogHandler logHandler(LOG_LEVEL_ALL);
+int count = 0;
+int error = 0;
+int test_errors = 0;
+int state = 1;
+int COUNT_MULTIPLIER = 1;
+system_tick_t lastState = 0;
+const system_tick_t TIME_FOR_5_PULSES = (20 * 5);
+InterruptMode intType = RISING;
+//
+// We will set most pins to INPUT_x modes, but only test the first 7
+// since PM_INT takes up one of the 8, and after that RISING and CHANGE pending interrupt tests will fail.
+//
+#if (PLATFORM_ID == PLATFORM_P2)
+    const int PIN_MAX = S6; // TEST: D0 ~ D6 (however all will work on P2)
+#elif (PLATFORM_ID == PLATFORM_ARGON) || (PLATFORM_ID == PLATFORM_BORON)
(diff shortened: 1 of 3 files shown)


user/tests/app/interrupts/interrupts.cpp
```diff
@@ -0,0 +1,271 @@
+#include "Particle.h"
```
avtolstoy, line 1: Let's turn this into a test that uses our testing framework? We can then run this on HIL if required with a fixture. Moreover there is no user interaction here, so another argument for not having this as a regular application.

avtolstoy (approved):

user/tests/app/interrupts/interrupts.cpp
technobly: There is user interaction because you have to move the wire from A5 to D0, then A5 to D1... etc.. resetting or re-running the test each time.  This test has served its purpose of testing all platforms for this issue as it stands.  It's easy to compile and run.  I don't think we need to reconfigure it for a "fixtured" test framework until we have that framework in place.  Right now all "fixtured" tests are run manually.

avtolstoy: Ok, fair enough. Even though we have examples with user interaction (wiring/sleep20) this is more akin to `sleep_multiple_pins`, so let's leave as-is :+1:

(Ugh, wish we had something like a CPLD/FPGA that we could control from test runner to dynamically 'connect' pins at least for digital IO)

technobly: Yeah, we'll get there and can then automate all of these manual tests 👍🏼 💯

PR #2505, opened 2022-08-11
[esomx] Remove undefined pins from esomx pinmap

scott-brust:
### Problem

There are defined pins that dont exist on the esomx hardware. These are causing failures when they are used, and masking problems when users use this pins and compliation succeeds. 

### Solution
Remove the pins from the pinmap, force explicit compilation failures to prompt fixing usage of the non existent pins.
This PR also fixes compliation problems with
- WIznet ethernet chip: If we want to use this with esomx we should reconsider the pin usage
- Tinker app: Specified the available pins for ESOMX hardware
- The 'User LED' for boron is on `D7` which doesnt exist on esomx. If we want to support this concept, I used `D5` instead
- The `wiring/api` test for Arduino expects a `LED_BUILTIN` define
- The `product_id_and_version` app expects a D7 pin to be available. 

### Steps to Test

Build for esomx platform

### Example App
use tinker

### References

---

### Completeness

- [ ] Added documentation


hal/src/boron/network/network.cpp
```diff
@@ -125,7 +125,7 @@ int if_init_platform(void*) {
 #if PLATFORM_ID == PLATFORM_BORON
         en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
 #elif PLATFORM_ID == PLATFORM_ESOMX
-        en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);        
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, SS1, MOSI1, MISO1, mac);
```
avtolstoy, line 128: I'd propose we don't use pins that overlap with a peripheral, it is that way for other Gen 3 platforms, but shouldn't be the case here if avoidable.

hal/src/boron/pinmap_defines.h
```diff
@@ -170,41 +170,37 @@
...
+#define D15 11
+#define D16 12
+#define D17 13
+#define D18 14
+#define D19 15
+#define D20 16
+#define D21 17
+#define D22 18
+#define D23 19
+#define D24 20
```
avtolstoy, line 198: This doesn't look right
```
/* D21 / B3      - 17 */ { NRF_PORT_1,    13,          PIN_MODE_NONE, PF_NONE, ADC_CHANNEL_NONE, 2,                 3,                8, EXTI_CHANNEL_NONE, 0},
/* D24 / C0      - 18 */ { NRF_PORT_1,    8,           PIN_MODE_NONE, PF_NONE, ADC_CHANNEL_NONE, PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 8, EXTI_CHANNEL_NONE, 0},
```

user/applications/product_id_and_version/application.cpp
```diff
@@ -5,13 +5,19 @@
 //
 PRODUCT_VERSION(3);
 
+#if PLATFORM_ID == PLATFORM_ESOMX
+	hal_pin_t toggle_pin = LED_PIN_USER;
```
avtolstoy, line 9: This should be defined for all platforms and this is an internal app. Can use `LED_PIN_USER` universally.

avtolstoy (changes requested):

user/applications/product_id_and_version/application.cpp
scott-brust: Ok, I didnt know if this define was available on gen2 platforms when we backport this PR to 4.x

avtolstoy: Well, good news is that there is no Gen 2 support in 4.x :)

scott-brust: oh, right 🙃

avtolstoy (approved):

technobly (approved):
LGTM!  After you merge to develop for 5.x, could you please make a new PR to target develop-4.x with the same commits?

hal/src/boron/pinmap_defines.h
technobly: D22/D23 is still missing in the pinmap_hal.c

hal/src/boron/network/network.cpp
technobly: Do we need to test this functionality with the new pins?

scott-brust: Ideally yes, but since there is no ethernet wing for the esomx it might take some extra effort to wire up. I'm not even sure this feature will be used for this platform.

PR #2507, opened 2022-08-13
[p2] attachInterrupt() should not configure input pullup/pulldown/nopull

technobly:
### Problem

From [Docs on attachInterrupt](https://docs.particle.io/reference/device-os/api/interrupts/attachinterrupt/):

> NOTE: pinMode() MUST be called prior to calling attachInterrupt() to set the desired mode for the interrupt pin (INPUT, INPUT_PULLUP or INPUT_PULLDOWN).

P2 implementation currently does not respect this and modifies the pin pull up based on the InterruptMode type (RISING/FALLING/CHANGE).

### Solution

- The Realtek routines that modify pullup/pulldown resistors are `GPIO_Init()` and `GPIO_INTMode()`.  Since these are compiled libraries, we have to reimplement them without the calls to `PAD_PullCtrl()` for use only in `hal_interrupt_attach()`
- I moved the placement/timing of `GPIO_Init_HAL()` based on guidance from Realtek examples and documentation.

### Steps to Test

- Run `TEST=wiring/no_fixture` (INTERRUPTS* - automated test)
- Run `TEST=app/interrupts` (follow logging prompts - manual test)
- Run `TEST=app/input_pinmode` (follow logging prompts - manual test)
```
Attach a scope to A5, and test input D0 ~ D6 one at a time.
Check that test input mode only changes between the first two pulses on A5,
and not the second and third. Press the MODE button to cycle through all 9
configurations (INPUT/INPUT_PULLUP/INPUT_PULLDOWN) with (RISING/FALLING/CHANGE).
```

### References

- [sc-107554](https://app.shortcut.com/particle/story/107554/p2-attachinterrupt-api-should-not-add-pull-up-down-resistors)

---

### Completeness

- [x] Added documentation

### DIFF
--- a/hal/src/rtl872x/interrupts_hal.cpp
+++ b/hal/src/rtl872x/interrupts_hal.cpp
@@ -87,6 +87,64 @@ void gpioIntHandler(void* data) {
     }
 }
 
+// Particle implementation of GPIO_INTMode (without the PAD_PullCtrl() call)
+void GPIO_INTMode_HAL(u32 GPIO_Pin, u32 GPIO_Port, u32 GPIO_ITEnable, u32 GPIO_ITTrigger, u32 GPIO_ITPolarity, u32 GPIO_ITDebounce) {
+    uint32_t pinMask = 1 << (GPIO_Pin & 0x1f);
+    uint32_t portBase = (GPIO_Port == RTL_PORT_A) ? (uint32_t)GPIOA_BASE : (uint32_t)GPIOB_BASE;
+    uint32_t tempReg = 0;
+
+    // Explicitly do NOT set PULL UP/PULL DOWN PADCTL register
+
+    // SET GPIO_INT_BOTHEDGE reg
+    tempReg = HAL_READ32(portBase, 0x68);
+    if (GPIO_ITTrigger == GPIO_INT_Trigger_BOTHEDGE) {
+        HAL_WRITE32(portBase, 0x68, tempReg | pinMask);
+    } else {
+        HAL_WRITE32(portBase, 0x68, tempReg & ~pinMask);
+        // SET GPIO_INTTYPE_LEVEL reg only if not BOTH EDGE
+        tempReg = HAL_READ32(portBase, 0x38);
+        if (GPIO_ITTrigger == GPIO_INT_Trigger_EDGE) {
+            HAL_WRITE32(portBase, 0x38, tempReg | pinMask);
+        } else {
+            HAL_WRITE32(portBase, 0x38, tempReg & ~pinMask);
+        }
+    }
+    // SET GPIO_INT_POLARITY reg
+    tempReg = HAL_READ32(portBase, 0x3c);
+    if (GPIO_ITPolarity == GPIO_INT_POLARITY_ACTIVE_HIGH) {
+        HAL_WRITE32(portBase, 0x3c, tempReg | pinMask);
+    } else {
+        HAL_WRITE32(portBase, 0x3c, tempReg & ~pinMask);
+    }
+    // SET GPIO_DEBOUNCE reg
+    tempReg = HAL_READ32(portBase, 0x48);
+    if (GPIO_ITDebounce == GPIO_INT_DEBOUNCE_ENABLE) {
+        HAL_WRITE32(portBase, 0x48, tempReg | pinMask);
+    } else {
+        HAL_WRITE32(portBase, 0x48, tempReg & ~pinMask);
+    }
+    // SET GPIO_INTEN reg
+    tempReg = HAL_READ32(portBase, 0x30);
+    if (GPIO_ITEnable == ENABLE) {
+        HAL_WRITE32(portBase, 0x30, tempReg | pinMask);
+    } else {
+        HAL_WRITE32(portBase, 0x30, tempReg & ~pinMask);
+    }
+}
+
+// Particle implementation of GPIO_Init (which calls our GPIO_INTMode_HAL to avoid PAD_PullCtrl() call)
+// NOTE: Only for use with GPIO_Mode_INT, do not use for GPIO_Mode_IN/GPIO_Mode_OUT
+// NOTE: hal_pin_is_valid(pin) must be called prior to this!
+void GPIO_Init_HAL(GPIO_InitTypeDef *GPIO_InitStruct, u32 GPIO_Port) {
+    if (GPIO_InitStruct->GPIO_Mode != GPIO_Mode_INT) {
+        return;
+    }
+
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -87,6 +87,64 @@ void gpioIntHandler(void* data) {
...
+
+// Particle implementation of GPIO_Init (which calls our GPIO_INTMode_HAL to avoid PAD_PullCtrl() call)
+// NOTE: Only for use with GPIO_Mode_INT, do not use for GPIO_Mode_IN/GPIO_Mode_OUT
+// NOTE: hal_pin_is_valid(pin) must be called prior to this!
+void GPIO_Init_HAL(GPIO_InitTypeDef *GPIO_InitStruct, u32 GPIO_Port) {
+    if (GPIO_InitStruct->GPIO_Mode != GPIO_Mode_INT) {
+        return;
+    }
+
+    Pinmux_Config(GPIO_InitStruct->GPIO_Pin, GPIO_Mode_IN); // force INPUT mode
```
avtolstoy, line 143: This doesn't look right. `Pinmux_Config` configures the pin function, pin functions are:
```
#define PINMUX_FUNCTION_GPIO				(0)
#define PINMUX_FUNCTION_UART				(1)

#define PINMUX_FUNCTION_UART_RTSCTS		(2)
#define PINMUX_FUNCTION_LOGUART			(2)

#define PINMUX_FUNCTION_SPIM				(3)
#define PINMUX_FUNCTION_SPIS				(3)

#define PINMUX_FUNCTION_RTC				(4)
#define PINMUX_FUNCTION_TIMINPUT			(4)
#define PINMUX_FUNCTION_EXT32K			(28)
#define PINMUX_FUNCTION_RTCOUT			(28)
#define PINMUX_FUNCTION_TIMINPUT_HS		(22)

#define PINMUX_FUNCTION_IR					(5)
#define PINMUX_FUNCTION_SPIF				(6)
#define PINMUX_FUNCTION_I2C				(7)

#define PINMUX_FUNCTION_SDIOD				(8)
#define PINMUX_FUNCTION_SDIOH				(8)

#define PINMUX_FUNCTION_PWM				(9)
#define PINMUX_FUNCTION_PWM_HS			(9)
#define PINMUX_FUNCTION_PWM_LP			(10)
#define PINMUX_FUNCTION_SWD				(11)

#define PINMUX_FUNCTION_I2S				(12)
#define PINMUX_FUNCTION_DMIC				(12)

#define PINMUX_FUNCTION_KEYSCAN_ROW		(29)
#define PINMUX_FUNCTION_KEYSCAN_COL		(30)PAD_BIT_MASK_FUNCTION_ID
#define PINMUX_FUNCTION_LCD				(13)
#define PINMUX_FUNCTION_USB				(14)
#define PINMUX_FUNCTION_QDEC				(15)
#define PINMUX_FUNCTION_SGPIO				(16)
#define PINMUX_FUNCTION_RFE				(18)
#define PINMUX_FUNCTION_BTCOEX			(19)

#define PINMUX_FUNCTION_WIFIFW			(20)
#define PINMUX_FUNCTION_EXT_PCM			(20)
#define PINMUX_FUNCTION_BB_PIN			(21)
#define PINMUX_FUNCTION_SIC				(22)
#define PINMUX_FUNCTION_WAKEUP			(31)
#define PINMUX_FUNCTION_DBGPORT			(23)
#define PINMUX_FUNCTION_BBDBG				(25)
```

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -150,9 +207,9 @@ int hal_interrupt_attach(uint16_t pin, hal_interrupt_handler_t handler, void* da
             return SYSTEM_ERROR_INVALID_ARGUMENT;
         }
 
+        GPIO_Init_HAL(&GPIO_InitStruct, pinInfo->gpio_port);
         GPIO_UserRegIrq(rtlPin, (VOID*)gpioIntHandler, (void*)((uint32_t)pin));
-
-        GPIO_INTMode(rtlPin, ENABLE, GPIO_InitStruct.GPIO_ITTrigger, GPIO_InitStruct.GPIO_ITPolarity, GPIO_INT_DEBOUNCE_ENABLE);
+        GPIO_INTMode_HAL(rtlPin, pinInfo->gpio_port, ENABLE, GPIO_InitStruct.GPIO_ITTrigger, GPIO_InitStruct.GPIO_ITPolarity, GPIO_INT_DEBOUNCE_ENABLE);
```
avtolstoy, line 212: Why do we call `GPIO_INTMode_HAL` twice here? Once in `GPIO_Init_HAL` once here.

user/tests/app/input_pinmode/input_pinmode.cpp
```diff
@@ -0,0 +1,122 @@
+#include "Particle.h"
+SYSTEM_MODE(SEMI_AUTOMATIC);
```
avtolstoy, line 2: @technobly App-based tests are fine, but we all know that they are never going to be run. Let's also come up with an automated test that verifies that the pin configuration was unaffected, even if something simple as e.g.

```c++
pinMode(pin, OUTPUT_OPEN_DRAIN_PULLUP);
digitalWrite(pin, 0);
assertEqual(digitalRead(pin), 0);
digitalWrite(pin, 1);
assertEqual(digitalRead(pin), 1);
//
attachInterrupt(pin, ...);
//
digitalWrite(pin, 0);
assertEqual(digitalRead(pin), 0);
digitalWrite(pin, 1);
assertEqual(digitalRead(pin), 1);
```

avtolstoy (changes requested):

hal/src/rtl872x/interrupts_hal.cpp
technobly: Good catch!  I had documented this, but missed it in the implementation.  They are both 0 though so it still worked :D

user/tests/app/input_pinmode/input_pinmode.cpp
technobly: I added a test that checks the pullup/down setting which is the main this affected by this PR.  The INPUT/OUTPUT mode was actually not changed in attachInterrupt() before or now.

hal/src/rtl872x/interrupts_hal.cpp
technobly: It's called this way because of the Realtek SDK docs, and all SDK examples.  [See here](https://github.com/particle-iot/realtek_ambd_sdk_public/blob/main/component/soc/realtek/amebad/fwlib/include/rtl8721d_gpio.h#L59-L82)  Maybe this isn't fully correct, but I didn't want to make any assumptions since I don't know everything that `GPIO_UserRegIrq()` does.

avtolstoy (approved):

PR #2511, opened 2022-08-16
P2 PWM fixture tests. PWM HAL fixes

scott-brust:
### Problem

P2 cannot read a PWM pin itself using `pulse_in()`. This PR adds a PWM fixture test that uses one pin to generate a PWM output and another to measure it via `pulse_in()`. 

### Solution

Port the `no_fixture_long_running/pwm` tests into a `wiring/pwm` fixture test app. 
- In general, tests configure one pin as the "output" pin, which a PWM signal is generated on. Each other PWM pin is configured as in "input" pin, and `pulse_in` is used to measure the output pins signal. Each permutation of output/input pins is configured and run for each test. 
- Capacitance doesnt seem to be an issue but if it is, the test can be modified to only test certain pairs/combinations of pins

Implementing these tests resulted in some PWM hal fixes/changes:
- Default PWM resolution changed to "8 bit" resolution (The KM4 PWM timer Auto Reload and Count registers are always 16 bit values, and vary depending on the prescaler value as well. But when the PWM resolution is changed via `hal_pwm_set_resolution()` the register values are calculated to match the new "resolution".)
- `pwmInfo` state structure refactored a bit. Each channel has its own state to make pwm sleep/wake easier
- PWM Frequency calculations improved by use of prescaler
- GPIO HAL / `pinMode()` will disable PWM configuration from pin. 
- `pwmPinDeinit()` will not clear pin from pwmInfo state, allowing `hal_pwm_sleep()` to work as intended

### Steps to Test

- To build:  
`make PLATFORM=p2 TEST=wiring/pwm`
- To setup: connect all P2 PWM pins together: D1, A2, A5, S0, S1
- To run: open USB serial connection, hit `t` to start all tests

### Example App

use `wiring/pwm`

Test output (verified with logic analyer)
```
Running tests
Test PWM_01_LowDCAnalogWriteOnPinResultsInCorrectPulseWidth passed.
Test PWM_02_LowFrequencyAnalogWriteOnPinResultsInCorrectPulseWidth passed.
Test PWM_03_HighFrequencyAnalogWriteOnPinResultsInCorrectPulseWidth passed.
Test PWM_04_PwmSleep passed.
Test PWM_05_CompherensiveResolutionFrequency passed.
Test summary: 5 passed, 0 failed, and 0 skipped, out of 5 test(s).
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/pwm_hal.cpp
+++ b/hal/src/rtl872x/pwm_hal.cpp
@@ -39,7 +39,7 @@ extern "C" {
 
 #define MAX_PRESCALER               256
 
-#define DEFAULT_RESOLUTION          10
+#define DEFAULT_RESOLUTION          8
 
 namespace {
 
@@ -50,35 +50,27 @@ typedef enum pwm_state_t {
     PWM_STATE_SUSPENDED
 } pwm_state_t;
 
+typedef struct rtl_pwm_channel_state_t {
+    uint8_t resolution;  // Desired resolution in bits
+    uint8_t pin;         // Index in pinmap for this channel
+    uint32_t value;      // Desired PWM value, in relation to the resolution
+    pwm_state_t state;
+} rtl_pwm_channel_state_t;
+
 typedef struct rtl_pwm_info_t {
     RTIM_TypeDef*           tim;
     IRQn                    irqn;
     uint8_t                 timIdx;
-    uint8_t                 resolution[MAX_PWM_CHANNEL_NUM]; // Desired resolution
     pwm_state_t             state;
-    uint32_t                frequency; // Actual output PWM frequency
-    uint8_t                 pins[MAX_PWM_CHANNEL_NUM];
-    uint32_t                values[MAX_PWM_CHANNEL_NUM];
-    uint32_t                arr; // 16-bits Auto-Reload register
+    uint32_t                frequency; // Desired output PWM frequency in Hz
+    uint32_t                arr;       // 16-bits Auto-Reload register
+    rtl_pwm_channel_state_t channels[MAX_PWM_CHANNEL_NUM];
 } rtl_pwm_info_t;
 
-#define PWM_DEFAULT_RESOLUTION_INITIALIZER() { \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION, \
-    DEFAULT_RESOLUTION \
-    }
-
 rtl_pwm_info_t pwmInfo[RTL_PWM_INSTANCE_NUM] = {
-//   tim,    irqn,       timIdx, resolution,                        state,             frequency
-    {TIMM05, TIMER5_IRQ, 5,      PWM_DEFAULT_RESOLUTION_INITIALIZER(), PWM_STATE_UNKNOWN, DEFAULT_PWM_FREQ, {0}, {0}, 0},
-    {TIM5,   TIMER5_IRQ, 5,      PWM_DEFAULT_RESOLUTION_INITIALIZER(), PWM_STATE_UNKNOWN, DEFAULT_PWM_FREQ, {0}, {0}, 0}
+//   tim,    irqn,       timIdx, state,             frequency,        arr, channels
+    {TIMM05, TIMER5_IRQ, 5,      PWM_STATE_UNKNOWN, DEFAULT_PWM_FREQ, 0,   {}},
+    {TIM5,   TIMER5_IRQ, 5,      PWM_STATE_UNKNOWN, DEFAULT_PWM_FREQ, 0,   {}}
 };
 
 bool isPwmPin(uint16_t pin) {
@@ -99,17 +91,19 @@ bool isPwmPin(uint16_t pin) {
 }
 
 void calculateArrAndPrescaler(uint32_t frequency, uint32_t* arr, uint8_t* prescaler) {
-    // Figure out the maximum ARR that meets the frequency requirment
-    // The higher ARR, the higher resolution we can set
+    // The PWM output frequency is calculated by the following formula
+    // Frequency = (Xtal/prescaler) / arr
     *arr = RTL_XTAL_CLOCK_HZ / frequency;
-    if (*arr > ((1 << RTL_PWM_TIM_MAX_RESOLUTION) - 1)) {
-        *arr = (1 << RTL_PWM_TIM_MAX_RESOLUTION) - 1;
+    uint8_t calculated_prescaler = 0;
+
+    // If the calculated ARR value is larger than will fit in the 16 bit ARR register
+    // Increase the prescaler and recalculate until we get a value that fits
+    while (*arr > ((1 << RTL_PWM_TIM_MAX_RESOLUTION) - 1) && (calculated_prescaler < MAX_PRESCALER-1)) {
+        calculated_prescaler++;
+        *arr = (RTL_XTAL_CLOCK_HZ / (calculated_prescaler+1)) / frequency;
     }
-    float temp = (float)RTL_XTAL_CLOCK_HZ / frequency / *arr;
-    *prescaler = (uint8_t)(temp + 0.5);
-    if (*prescaler != 0) {
-        *prescaler -= 1;
-    }
+
+    *prescaler = calculated_prescaler;
 }
 
 uint32_t minFrequency() {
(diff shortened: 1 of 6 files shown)


hal/src/rtl872x/pwm_hal.cpp
```diff
@@ -373,10 +371,10 @@ void hal_pwm_set_resolution(uint16_t pin, uint8_t resolution) {
         }
     }
 
-    uint32_t prePeriodCycles = (1 << pwmInfo[instance].resolution[channel]);
+    uint32_t prePeriodCycles = (1 << pwmInfo[instance].channels[channel].resolution);
     uint32_t newPeriodCycles = (1 << resolution);
-    pwmInfo[instance].resolution[channel] = resolution;
-    pwmInfo[instance].values[channel] = ((float)pwmInfo[instance].values[channel] / prePeriodCycles) * newPeriodCycles;
+    pwmInfo[instance].channels[channel].resolution = resolution;
+    pwmInfo[instance].channels[channel].value = ((float)pwmInfo[instance].channels[channel].value / prePeriodCycles) * newPeriodCycles;
```
scott-brust, line 377: When a new resolution is set, should we recalculate the ARR and CCMRX register values (ie duty cycle) to update the new values?

hal/src/rtl872x/pwm_hal.cpp
```diff
@@ -182,21 +179,22 @@ int pwmPinDeinit(uint16_t pin) {
...
     const uint8_t instance = pinInfo->pwm_instance;
     const uint8_t channel = pinInfo->pwm_channel;
+    const pwm_state_t channelState = pwmInfo[instance].channels[channel].state;
 
-    if (pwmInfo[instance].pins[channel] != PIN_INVALID) {
+    if (channelState == PWM_STATE_ENABLED) {
         const uint32_t rtlPin = hal_pin_to_rtl_pin(pin);
         Pinmux_Config(rtlPin, PINMUX_FUNCTION_GPIO);
 
-        pwmInfo[instance].pins[channel] = PIN_INVALID;
```
scott-brust, line 190: Setting the pin to `PIN_INVALID` made it impossible to re-generate the PWM signal on the suspended PWM channel when `hal_pwm_sleep(true) / hal_pwm_sleep(false)` was called. Thats why I changed pwm hal to track each channel state. When a pin is init'd, the pin number will always be preserved

XuGuohui (changes requested):

hal/src/rtl872x/pwm_hal.cpp
```diff
@@ -99,17 +91,19 @@ bool isPwmPin(uint16_t pin) {
...
-    }
-    float temp = (float)RTL_XTAL_CLOCK_HZ / frequency / *arr;
-    *prescaler = (uint8_t)(temp + 0.5);
-    if (*prescaler != 0) {
-        *prescaler -= 1;
+    uint8_t calculated_prescaler = 0;
+
+    // If the calculated ARR value is larger than will fit in the 16 bit ARR register
+    // Increase the prescaler and recalculate until we get a value that fits
+    while (*arr > ((1 << RTL_PWM_TIM_MAX_RESOLUTION) - 1) && (calculated_prescaler < MAX_PRESCALER-1)) {
```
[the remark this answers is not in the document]
technobly: I think it was because of a bug found with lower resolutions.  As a compromise can we pre-calculate things, and then drop into the while() loop to make a quick tweak, instead of starting from `calculated_prescaler = 0` every time?  I'm sure this is pretty fast already, and also I guess this does get called on every call to analogWrite* so the time it takes could be important to some.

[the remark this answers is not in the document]
scott-brust: When the calculated `arr` value was larger than 16 bits, it was truncated at 16 bits. 
Then the truncated value was used to calculate the prescaler value. This would lead to incorrect PWM frequencies, especially at low frequencies. 

For example, the default PWM frequency is 500hz. that means `arr` is calculated to be:
`40mhz / 500 hz = 80000`
which is greater than `1<<16 -1` (ie 0xFFFF) and is truncated to `0xFFFF`

When using the truncated value to calculate prescaler:
`40mhz / 500 hz / 0xFFFF = 1.22`
`1.22 + 0.5 = 1.72`
Casting to `uint8_t` rounds to `1`. Subtracting 1 from that means the prescaler should be 0.

`arr` set to `0xFFFF` and prescaler `0` gives a frequency calculation of:
`40mhz / 0xFFFF = 610 hz`
Which is not the desired 500hz output. 

This method is somewhat slower but should be more accurate.

user/tests/wiring/pwm/pwm.cpp
```diff
@@ -0,0 +1,400 @@
...
+            assertMoreOrEqual(avgPulseHigh, 150);
+            assertLessOrEqual(avgPulseHigh, 250);
+        }, pwm_output_pin);
+        pinMode(pwm_output_pin, INPUT);
+    });
+}
+
+// 10hz
+test(PWM_02_LowFrequencyAnalogWriteOnPinResultsInCorrectPulseWidth) {
+    static const uint32_t TEST_FREQUENCY_HZ = 10;
```
technobly, line 135: Since we guarantee operation down to 5Hz, should we make this test run at 5Hz?  I did a quick test locally doubling the assert values and it passed.

technobly (approved):
Tests passed several times for me 👍🏼 
After the 3rd test run I noticed this failure:
```
Assertion failed: (avgPulseHigh=342) <= (316=316), file tests/wiring/pwm/pwm.cpp, line 107.
Test PWM_01_LowDCAnalogWriteOnPinResultsInCorrectPulseWidth failed.
```

So I ran it 100 more times in a loop, only failed 3 times:
```
Assertion failed: (avgPulseHigh=283) <= (250=250), file tests/wiring/pwm/pwm.cpp, line 119.
Assertion failed: (avgPulseHigh=278) <= (250=250), file tests/wiring/pwm/pwm.cpp, line 119.
Assertion failed: (avgPulseHigh=300) <= (250=250), file tests/wiring/pwm/pwm.cpp, line 119.
```

Would it be ok to increase the tolerance here and just check for really blatant errors?

hal/src/rtl872x/pwm_hal.cpp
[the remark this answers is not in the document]
XuGuohui: That makes perfect sense. 👍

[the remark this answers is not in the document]
technobly: Can we maybe cache the frequency value and only re-calculate if the frequency for the pin changes?

hal/src/rtl872x/pwm_hal.cpp
```diff
@@ -129,11 +123,13 @@ int pwmTimBaseInit(rtl_pwm_info_t* info) {
...
-        memset(info->resolution, DEFAULT_RESOLUTION, sizeof(info->resolution));
-        // Juts in case
+        for( int i = 0; i < MAX_PWM_CHANNEL_NUM; i++){
+            info->channels[i].pin = PIN_INVALID;
+            info->channels[i].resolution = DEFAULT_RESOLUTION;
+            info->channels[i].state = PWM_STATE_DISABLED;
+        }
+        // Just in case
         RTIM_Cmd(info->tim, DISABLE);
-        info->state = PWM_STATE_DISABLED;
```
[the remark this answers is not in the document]
XuGuohui: I recall that updating the state here right after the PWM is disabled is just to avoid potential race condition. But yeah, it doesn't hurt even if we don't update the state herein.

XuGuohui (approved):

hal/src/rtl872x/pwm_hal.cpp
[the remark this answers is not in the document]
scott-brust: Determining the prescaler is only done when changing the frequency. Do we think this will be done frequently enough to need to improve performance here? 

Modifying duty cycle via `hal_pwm_update_duty_cycle_ext()` only updates the CCR register, which should be fast.

I would prefer to not make this more complicated unless needed

hal/src/rtl872x/pwm_hal.cpp
technobly: If a pin is already running a non-zero PWM output, and the resolution is changed... we don't have docs that say what is expected to happen.  I don't think the duty cycle value should be re-adjusted though (that's up to the user), but the frequency probably should remain the same (if it already doesn't).  We probably should have a note that says something like, "It's recommended to set the resolution before calculating the duty cycle value and setting the PWM to a non-zero value." ?

user/tests/wiring/pwm/pwm.cpp
scott-brust: Probably? This is just what was in `no_fixture_long_running`

hal/src/rtl872x/pwm_hal.cpp
scott-brust: We will store the new resolution and duty cycle value in the PWM HAL, but it wont be applied to the timer registers and the PWM output duty cycle / frequency wont change. 
The user will have to call `analogWrite()` to get their duty cycle/frequency changes to take effect. So there is no new behavior here, but yeah could be documented better?

scott-brust:
> Would it be ok to increase the tolerance here and just check for really blatant errors?
Since this test is manual I think its probably fine to leave as is and just accept that a tight tolerance means some false failures? I dont really know how the margins of error were originally determined anyways.

hal/src/rtl872x/pwm_hal.cpp
avtolstoy: > We will store the new resolution and duty cycle value in the PWM HAL, but it wont be applied to the timer registers and the PWM output duty cycle / frequency wont change.
> The user will have to call analogWrite() to get their duty cycle/frequency changes to take effect. So there is no new behavior here, but yeah could be documented better?

That's the expected behavior.

avtolstoy (approved):

hal/src/rtl872x/pwm_hal.cpp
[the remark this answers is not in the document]
technobly: It looked to me that `calculateArrAndPrescaler` was called from every call to `hal_pwm_write_with_frequency_ext` which also gets called by `hal_pwm_write*`.

user/tests/wiring/pwm/pwm.cpp
technobly: I would advocate for changing it to 5Hz in `no_fixture_long_running` as well (maybe as a separate PR later).  If we going to test it, might as well test the limits?

scott-brust: [I agree with that, made a story for it here](https://app.shortcut.com/particle/story/108164/update-pwm-tests-to-use-lowest-supported-frequency)

PR #2513, opened 2022-08-17
[test] fixes race condition with wiring/no_fixture TIME_17

technobly:
### Problem

TIME_17 was failing, presumably due to a race condition with time sync arriving late

### Solution

Run TIME_17 offline and use the RTC as-is (it might have sync'd time, it might not... doesn't matter for the test)

### Steps to Test

- Run `TEST=wiring/no_fixture` (TIME*) tests

---

### Completeness

- [x] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/time.cpp
+++ b/user/tests/wiring/no_fixture/time.cpp
@@ -27,6 +27,7 @@
 #include "unit-test/unit-test.h"
 #include "rtc_hal.h"
 #include "simple_ntp_client.h"
+#include "scope_guard.h"
 
 test(TIME_01_NowReturnsCorrectUnixTime) {
     // when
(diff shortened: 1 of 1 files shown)


user/tests/wiring/no_fixture/time.cpp
```diff
@@ -270,11 +271,12 @@ test(TIME_16_TimeChangedEvent) {
 }
 
 test(TIME_17_RtcAlarmFiresCorrectly) {
-    if (Particle.syncTimePending()) {
-        waitFor(Particle.syncTimeDone, 60000);
-        assertTrue(Particle.syncTimeDone());
-        assertTrue(Time.isValid());
-    }
+    SCOPE_GUARD ({
+            Particle.connect();
```
avtolstoy, line 275: Nit: extra indentation.

We can perhaps even skip connecting here, `TIME_19_LocalTimeIsCloseToNtpTime` will connect us back anyway.

avtolstoy (approved):

PR #2526, opened 2022-08-25
[wiring] System.hardwareInfo()

avtolstoy:
### TODO

When backporting to 4.x, make sure to avoid calling into `hal_get_device_hw_info()` if system module version is >= 5000 and <= 5002`

### Steps to Test

- `wiring/api`
- `wiring/no_fixture` SYSTEM_08

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/tracker/hal_platform_config.h
```diff
@@ -48,3 +48,9 @@
 #define HAL_PLATFORM_MAX_CLOUD_CONNECT_TIME (9*60*1000)
 
 #define PRODUCT_SERIES "boron"
+
+#define PLATFORM_TRACKER_MODEL_BARE_SOM_DEFAULT        (0xffff)
```
eberseth, line 52: These macros are application specific and likely redundant

avtolstoy: As we discussed yesterday, I'm adding these here "just in case" for non-tracker-edge use-cases. Doesn't cost us anything, they don't overlap with previous edge-specific defines and are properly namespaced (`PLATFORM_`). No strong opinions on this though.

eberseth: They can stay.  Would you mind adding another macro like this or similiar?
```cpp
#define PLATFORM_TRACKER_MODEL_LAST    (PLATFORM_TRACKER_MODEL_MONITORONE)
```

hal/src/nRF52840/deviceid_hal.cpp
```diff
@@ -111,6 +112,21 @@ int hal_get_device_hw_model(uint32_t* model, uint32_t* variant, void* reserved)
     return SYSTEM_ERROR_NONE;
 }
 
+int hal_get_device_hw_info(hal_device_hw_info* info, void* reserved) {
```
eberseth, line 115: Does `info->size` need to be handled in this function?

avtolstoy: Not as of right now. Only once this struct is extended.

hal/src/tracker/hal_platform_config.h
avtolstoy: Sure, we can add it.

tstellanova (changes requested):
Adding minor #defines suggested by Eric would be good. Otherwise looks great

hal/src/rtl872x/deviceid_hal.cpp
```diff
@@ -165,6 +165,21 @@ int hal_get_device_hw_model(uint32_t* model, uint32_t* variant, void* reserved)
     return SYSTEM_ERROR_NONE;
 }
 
+int hal_get_device_hw_info(hal_device_hw_info* info, void* reserved) {
+    CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
+    // HW Data format: | NCP_ID (LSB) | HW_VERSION | HW Feature Flags |
+    //                 |    byte 0    |   byte 1   |    byte 2/3      |
+    uint8_t hw_data[4] = {};
```
scott-brust, line 172: It would be helpful to return the NCP ID in the `hal_device_hw_info` struct here as well

hal/shared/platform_ncp.h
```diff
@@ -62,7 +66,11 @@ enum PlatformNCPIdentifier {
     PLATFORM_NCP_BROADCOM_BCM9WCDUSI09 = PLATFORM_NCP_IDENTIFIER(PLATFORM_NCP_MANUFACTURER_BROADCOM, 1),
     PLATFORM_NCP_BROADCOM_BCM9WCDUSI14 = PLATFORM_NCP_IDENTIFIER(PLATFORM_NCP_MANUFACTURER_BROADCOM, 2),
     PLATFORM_NCP_REALTEK_RTL872X = PLATFORM_NCP_IDENTIFIER(PLATFORM_NCP_MANUFACTURER_REALTEK, 1)
```
scott-brust, line 68: I realized all existing P2s have been provisioned with an NCPID of `A0`, which unfortunately doesnt match this (`PLATFORM_NCP_REALTEK_RTL872X` should be `A1` right?). 
I think it should be changed to the following if we want to identify P2s as they are now:
```
PLATFORM_NCP_REALTEK_RTL872X = PLATFORM_NCP_IDENTIFIER(PLATFORM_NCP_MANUFACTURER_REALTEK, 0)
```

hal/src/rtl872x/deviceid_hal.cpp
```diff
@@ -165,6 +150,28 @@ int hal_get_device_hw_model(uint32_t* model, uint32_t* variant, void* reserved)
...
+    CHECK(readLogicalEfuse(HARDWARE_DATA_OFFSET, (uint8_t*)hw_data, sizeof(hw_data)));
+    CHECK(hal_get_device_hw_model(&info->model, &info->variant, nullptr));
+    info->revision = hw_data[1];
+    if (info->revision == 0xff) {
+        info->revision = 0xffffffff;
+    }
+    info->features = ((uint32_t)hw_data[3] << 8) | (uint32_t)hw_data[2];
+    memset(info->ncp, 0xff, sizeof(info->ncp));
+    for (int i = 0; i < HAL_PLATFORM_NCP_COUNT; i++) {
```
scott-brust, line 166: For `trackerm` we are going to not count the P2 wifi as one of the NCPs? I see its defined as only having the Quectel modem here
```
#define HAL_PLATFORM_NCP_COUNT (1)
```

avtolstoy: Not necessarily. It's one of the potentially components that can be updated, so it doesn't need to be in the list of NCPs.

hal/shared/platform_ncp.h
avtolstoy: Let's fix this then.

hal/src/rtl872x/deviceid_hal.cpp
scott-brust: Ok, we only have one byte in efuse/otp for NCPID, so this should be the primary NCP right? So for trackerm this should be the appropriate Quectel NCPID right? 
I guess this also begs the question of what happens if/when we have a product with two updateable/"primary" NCPs, but I guess thats not really relevant (or likely?) right now

avtolstoy: See latest updates. This should probably be correct for all supported platforms.

scott-brust (approved):
LGTM.
Only thing I would point out is that currently with `trackerm` calling `platform_primary_ncp_identifier()` will give you `PlatformNCPIdentifier::PLATFORM_NCP_UNKNOWN` since there is nothing in OTP and the NCPID burned in efuse is the RTK WIFI ID, which isnt considered a valid primary NCPID. 
If you want `hal_get_device_hw_info()`/``platform_primary_ncp_identifier()` to be correct, users will have to either 1) Program NCPID into OTP now that we can do that on rtl872x platform or 2) re-provision efuse with a new NCPID, which isnt really supported yet.

avtolstoy:
@scott-brust Good point. We should expect this to be correctly provisioned for production devices and we can manually provision existing once as well.

PR #2527, opened 2022-08-26
[tests] fix pins for interrupt tests on trackerm

keeramis:
### Description

Fix the pins defined for running interrupt tests on TrackerM

### Example App

Run `wiring/no_fixture` user tests

### References

[SC-108882](https://app.shortcut.com/particle/story/108882/fix-interrupts-02-attachintterupt-does-not-affect-pullup-pulldown-nopull-test-for-trackerm)

---

### Completeness

- [ ] Added documentation


user/tests/app/input_pinmode/input_pinmode.cpp
```diff
@@ -14,7 +14,7 @@ InterruptMode intType = CHANGE;
 // We will set most pins to INPUT_x modes, but only test the first 7
 // since PM_INT takes up one of the 8, and after that RISING and CHANGE pending interrupt tests will fail.
 //
-#if (PLATFORM_ID == PLATFORM_P2)
+#if (PLATFORM_ID == PLATFORM_P2) || (PLATFORM_ID == PLATFORM_TRACKERM)
```
technobly, line 17: We should limit TRACKERM to use only D8 & D9.  Unfortunately this test starts always at D0, so it would require a PIN_MAX => START_PIN / END_PIN change.

user/tests/wiring/no_fixture/interrupts.cpp
```diff
@@ -68,6 +68,9 @@ test(INTERRUPTS_02_attachintterupt_does_not_affect_pullup_pulldown_nopull)
 #if PLATFORM_ID == PLATFORM_ESOMX
 	const pin_t START_PIN = D0;
     const pin_t END_PIN = D2;
+#elif PLATFORM_ID == PLATFORM_TRACKERM
+    const pin_t START_PIN = D8;
+    const pin_t END_PIN = D9;
```
technobly, line 73: Do you want to add the other changes we discussed as well if they pass ok?

technobly (changes requested):

user/tests/app/input_pinmode/input_pinmode.cpp
keeramis: There are multiple ways these pins can be added but looks like adding START_PIN is probably the simplest way.

technobly (approved):

PR #2528, opened 2022-08-27
Update demux driver and pinmap

jmwilson:
### Description

Fix demux driver on `trackerm`; the bit masks only worked with the nRF platform. Since all the pins are now on the same port, the output can be set atomically.

### Steps to Test

Run `wiring/no_fixture` after cherry-picking `fix/trackerm_interrupt_tests`.

### References

https://app.shortcut.com/particle/story/108998/

---

### Completeness

- [ ] Added documentation


hal/shared/demux.cpp
```diff
@@ -100,9 +110,10 @@ void Demux::init() {
     hal_gpio_mode(DEMUX_B, OUTPUT);
     hal_gpio_mode(DEMUX_C, OUTPUT);
     // Select Y0 by default.
-    hal_gpio_write(DEMUX_C, 0);
-    hal_gpio_write(DEMUX_B, 0);
-    hal_gpio_write(DEMUX_A, 0);
+    auto pinmap {hal_pin_map()};
+    GPIO_PortWrite(pinmap[DEMUX_A].gpio_port,
```
avtolstoy, line 114: This should ideally follow the same pattern as for nrf52840-based platforms: set the pin states in port output register, configure those pins as output as a single step if possible to avoid a glitch.

jmwilson: That is actually what this change does, which I have been looking forward to. `GPIO_PortWrite` is the rtl872x function that sets the outputs in a single step.

avtolstoy: We still have `hal_gpio_mode()` calls above though: https://github.com/particle-iot/device-os/pull/2528/files#diff-b5d1c90b79a5b05127a4447a1bc0844a98e9d5b1582b675d21040e7a0f6e697bR109. I'm emphasizing that configuring pins as push-pull and then setting the pin states may cause the glitch as well. The output state should be preconfigured before changing the pins to push-pull and changing to push-pull (OUTPUT) should ideally also be done in a single step similar. Now that I've looked at this more attentively we actually have the same problem with Gen 3, it should probably also be `nrf_gpio_port_out_clear`, `nrf_gpio_port_dir_output_set`, not the opposite.

jmwilson: I see what you mean. There is a similar call in the rtl872x SDK for this, but I had some problems with it and went back to individual `hal_gpio_mode` calls. I can look for more docs and examples, and then test it out.

jmwilson: @avtolstoy - I tried the following again based on the RTL8721x SDK examples, but it doesn't work. I think there is some secret sauce that `GPIO_Init` does that is not mentioned. I have not seen any glitching at reset time with these lines viewed under a logic analyzer.

```
GPIO_PortDirection(pinmap[DEMUX_A].gpio_port,
                   1u << pinmap[DEMUX_C].gpio_pin | 1u << pinmap[DEMUX_B].gpio_pin | 1u << pinmap[DEMUX_A].gpio_pin,
                   GPIO_Mode_OUT);
PAD_PullCtrl((pinmap[DEMUX_A].gpio_port << 5 | pinmap[DEMUX_A].gpio_pin), GPIO_PuPd_NOPULL);
PAD_PullCtrl((pinmap[DEMUX_A].gpio_port << 5 | pinmap[DEMUX_B].gpio_pin), GPIO_PuPd_NOPULL);
PAD_PullCtrl((pinmap[DEMUX_A].gpio_port << 5 | pinmap[DEMUX_C].gpio_pin), GPIO_PuPd_NOPULL);
```

avtolstoy: Okay, if there isn't any glitch good enough for us. Let's leave as-is.

keeramis (approved):
Approving to unblock merging this to 5.0.1 release.

PR #2530, opened 2022-08-31
Secures DCT initialization from getting interrupted between creating DCT file and filling it with 0xff to default state

avtolstoy:
### Description

As title says. Also refactors things a bit to have a common `dct_file.h` implementation shared between Gen 3 / P2 / Tracker M.

### Steps to Test

Force the reset:

```c++
diff --git a/services/inc/dct_file.h b/services/inc/dct_file.h
index 7bd291911..e6f905dc6 100644
--- a/services/inc/dct_file.h
+++ b/services/inc/dct_file.h
@@ -25,6 +25,8 @@
 #include <algorithm>
 #include <mutex>
 
+#include "core_hal.h"
+
 namespace particle {
 namespace dct {
 
@@ -133,19 +135,25 @@ private:
         int r = lfs_mkdir(lfs(), "/sys");
         SPARK_ASSERT((r == 0 || r == LFS_ERR_EXIST));
 
+        lfs_remove(lfs(), "/sys/dct.bin");
+
         /* Check that /sys/dct.bin exists */
         struct lfs_info info;
         r = lfs_stat(lfs(), path_, &info);
 
         int flags = LFS_O_RDWR;
 
-        if (r || info.size == 0) {
+        if (r /*|| info.size == 0*/) {
             /* Create as well if doesn't exist or was just created and not written */
             flags |= LFS_O_CREAT;
         }
 
         SPARK_ASSERT(open(flags));
 
+        if (flags & LFS_O_CREAT) {
+            HAL_Core_Enter_Bootloader(0);
+        }
+
         if (flags & LFS_O_CREAT) {
             LOG_DEBUG(INFO, "Initializing empty DCT");
             /* Fill with 0xff for compatibility with raw flash DCD */
```

The device will reset into DFU without blinking LED. Flash vanilla 5.0.0 or `develop`. It will boot-up with no LEDs etc and broken DCT filled with zeros.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


scott-brust (approved):
My only comments are about renaming from `class DcdFile` to `class DctFile`. I'm assuming DCD was something that predated DCT? Could we delete more stuff like 
`/services/inc/dcd.h`?

technobly:
Actually DCT predated DCD.  DCT came from Photon/P1 WICED SDK, and when we implemented our own for Electron it was named DCD to avoid confusion with WICED's DCT.  I don't feel super strongly about it, but it would be nice if it was uniformly one way or the other.

avtolstoy:
Now that we don't have WICED, `DCD` (Electron) let's just streamline everything as APIs are `dct_xxx` and when we talk about this we just say DCT. The class also has a concrete name which describes what it does. I'm going to drop `services/inc/dcd.h` as well. No need to keep it around anymore in 4.x+.

hal/src/rtl872x/dct_hal.cpp
```diff
@@ -138,8 +138,8 @@ class DcdFile {
 
         int flags = LFS_O_RDWR;
 
-        if (r) {
-            /* Create as well */
+        if (r || info.size == 0) {
```
technobly: Could we also check for `!= or < an expected size` in case a device is reset in the middle of the 0xFF writing loop?

technobly: `info.size < sizeof(application_dct_t)` ?

avtolstoy: That's guaranteed to either complete entirely or not at all. Also basing things on size is going to cause problems if we ever have to extend `application_dct_t`. We'll just jump straight on this landmine.

technobly: Lemme run your PR test... and I'll try to reset mid-0xFF write and see if that size is still 0.  If it's guaranteed, I'm good with that.

avtolstoy: (If it wasn't, we'd have all sorts of consistency issues everywhere not just DCT)

services/inc/dct_file.h
```diff
@@ -0,0 +1,180 @@
...
+        int r = lfs_mkdir(lfs(), "/sys");
+        SPARK_ASSERT((r == 0 || r == LFS_ERR_EXIST));
+
+        /* Check that /sys/dct.bin exists */
+        struct lfs_info info;
+        r = lfs_stat(lfs(), path_, &info);
+
+        int flags = LFS_O_RDWR;
+
+        if (r /* || info.size == 0*/) {
```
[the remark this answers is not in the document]
avtolstoy: Ooof. That shouldn't have gone in. That was for a test. Good catch. Fixed.

technobly (approved):
LGTM! 👍🏼 I tested:
- Recreating the issue in `develop`
- Recovering from the issue with this PR
- Recreating the issue in a different way by resetting after 1000 0xFF's were written, and it also will recover from that.

PR #2533, opened 2022-09-07
Fix MCP23S17 driver to allow mirrored (shared) interrupts

jmwilson:
### Description

Tracker M only has one of the MCP23S17 interrupt pins connected to the MCU. In proto 2, some of the interrupt pins are moved to port B, so the driver must configure the expander to connect the interrupt pins together. This configuration was made as a compile time flag for definition in `hal_platform_config.h`.

image

### Steps to Test

Manual testing with the CAN-FD controller that is connected to port B. Receive is not working without this change, but interrupts are handled with this change.


avtolstoy (approved):

hal/src/tracker/hal_platform_config.h
```diff
@@ -31,6 +31,7 @@
 #define HAL_PLATFORM_MCP23S17 (1)
 #define HAL_PLATFORM_MCP23S17_SPI (HAL_SPI_INTERFACE2)
 #define HAL_PLATFORM_MCP23S17_SPI_CLOCK (16000000)
+#define HAL_PLATFORM_MCP23S17_MIRROR_INTERRUPTS (0)
```
avtolstoy, line 34: A default needs to be added to `hal_platform.h`, no need to set to 0 here.

avtolstoy (changes requested):

avtolstoy (approved):

PR #2534, opened 2022-09-11
[tests] Fix listening mode tests on trackerM

keeramis:
### Problem

Listening mode tests are unstable on trackerM with threading enabled.
1. Unable to process feature flag as expected
2. Radio is repeatedly turned off indefinitely

### Solution

1. Some of the tests depend on a certain feature flag being processed by Spark_Idle_Events(), hence calling Particle.process() unblocks the system thread as needed and processes the feature flag
3. Network.listen() calls could potentially interfere with the state of ncpClient() and renders undefined behavior and doesn't detect the cell radio being powered off, hence calling Particle.disconnect() at the beginning of the tests keeps the ncpClient() states as expected

### Steps to Test

Run `wiring/no_fixture` tests

### References

[SC-107222](https://app.shortcut.com/particle/story/107222/tracker-m-invoke-particle-connect-inside-listening-tests-cause-blocking-behavior-with-threading-enabled)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/ble_provisioning.cpp
+++ b/user/tests/wiring/no_fixture/ble_provisioning.cpp
@@ -19,14 +19,9 @@
 #include "unit-test/unit-test.h"
 
 #if HAL_PLATFORM_BLE
-
+const system_tick_t WAIT_TIMEOUT = 15 * 60 * 1000;
 // Do not enter listening mode based on the flag
 test(LISTENING_00_DISABLE_LISTENING_MODE) {
-    // If Particle.disconnect() is not called here, the Network.listen() calls
-    // could potentially interfere with ncpClient and rendering undefined
-    // behavior for some of the following tests 
-    Particle.disconnect();
-    waitUntil(Particle.disconnected);
     System.enableFeature(FEATURE_DISABLE_LISTENING_MODE);
     Network.listen();
     SCOPE_GUARD({
@@ -51,6 +46,7 @@ test(LISTENING_01_ENABLE_LISTENING_MODE) {
     Network.listen(false);
     delay(1000); // Time for system thread to exit listening mode
     assertFalse(Network.listening());
+    waitFor(Particle.connected, WAIT_TIMEOUT);  // Wait for Particle connection to come back
 }
 
 // If the flag is enabled while device is in listening mode,
@@ -70,6 +66,7 @@ test(LISTENING_02_DISABLE_FLAG_WHILE_IN_LISTENING_MODE) {
     });
     delay(1500); // Time for system thread to process the flag
     assertFalse(Network.listening());
+    waitFor(Particle.connected, WAIT_TIMEOUT);  // Wait for Particle connection to come back
 }
 
 test(LISTENING_03_ENABLE_BLE_PROV_MODE_WHEN_FLAG_SET) {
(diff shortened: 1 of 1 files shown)


user/tests/wiring/no_fixture/ble_provisioning.cpp
```diff
@@ -22,6 +22,11 @@
 
 // Do not enter listening mode based on the flag
 test(LISTENING_00_DISABLE_LISTENING_MODE) {
+    // If Particle.disconnect() is not called here, the Network.listen() calls
+    // could potentially interfere with ncpClient and rendering undefined
+    // behavior for some of the following tests 
+    Particle.disconnect();
```
avtolstoy, line 28: I think this somewhat negates the purpose of the test and hides an underlying problem.

> Network.listen() calls could potentially interfere with ncpClient and rendering undefined behavior for some of the following tests 

This should not be happening. This call should break cloud connection and enter listening mode no matter what. If 1s timing cannot be guaranteed on some platforms due to modem behavior etc, this test needs to be refactored to wait longer and/or use system events to understand that an entry into listening mode occurred.

avtolstoy (changes requested):

user/tests/wiring/no_fixture/ble_provisioning.cpp
keeramis: Strangely, this is only reproducible on proto1 hardware. Proto2 hardware doesn't have this problem.

avtolstoy: Perhaps post the log from when this test runs? That can help us see what's happening in the background and perhaps also augment the test with additional log statements around various calls.

keeramis: Posted the log here 🙇  https://app.shortcut.com/particle/story/107222/tracker-m-invoke-particle-connect-inside-listening-tests-cause-blocking-behavior-with-threading-enabled#activity-109610

keeramis: > If 1s timing cannot be guaranteed on some platforms due to modem behavior etc, this test needs to be refactored to wait longer 

This helped. Added the change.

avtolstoy: can we perhaps just change these https://github.com/particle-iot/device-os/pull/2534/commits/7f10992f503d01215193c40375c968958c2ab675#diff-2798bda0c6b54d595cbad836b0e6bff170c69d0894c9e90b0195c2d547bf6537R32 to busy-ish loops waiting for listening mode system events or just checking `Network.listening()` up to a certain large timeout until the state change actually occurs?

keeramis: Which `state` should we monitor? `Network.listening()` is being checked where ever needed [like here](https://github.com/particle-iot/device-os/pull/2534/files#diff-2798bda0c6b54d595cbad836b0e6bff170c69d0894c9e90b0195c2d547bf6537R68) - After `Network.listening()` state is changed, previously established cloud connection restores which I believe is taking time.

user/tests/wiring/no_fixture/ble_provisioning.cpp
```diff
@@ -96,24 +106,22 @@ test(LISTENING_04_ENABLE_BLE_PROV_MODE_WHEN_FLAG_CLEARED) {
...
+        assertTrue(waitListening(false));
         // Make sure we restore cloud connection after exiting this test because we entered listening mode
         Particle.connect();
         waitFor(Particle.connected, WAIT_TIMEOUT);
         assertTrue(Particle.connected());
     });
     System.enableFeature(FEATURE_DISABLE_LISTENING_MODE);
-    delay(1500); // Time for system thread to process the flag
-    assertFalse(Network.listening());
+    Particle.process();
```
avtolstoy, line 123: This is implicitly done in `waitFor`, so probably can be omitted.

avtolstoy (approved):

PR #2535, opened 2022-09-12
Increase power management thread stack size for Tracker M

jmwilson:
### Description

Increase the power management thread stack size to prevent crash. `1024 + 128` crashes, `1024 + 256` is ok, so choosing `1024 + 512` for extra margin.

### Steps to Test

Manual testing with minimum crash test case on Tracker M proto board:

```
SYSTEM_MODE(MANUAL);

SerialLogHandler logHandler;

void setup() {
  Wire.setSpeed(CLOCK_SPEED_400KHZ);
}

void loop() {
}
```

### Completeness

- [ ] Added documentation


hal/src/trackerm/hal_platform_config.h
```diff
@@ -28,6 +28,7 @@
 #define HAL_PLATFORM_PMIC_BQ24195_I2C (HAL_I2C_INTERFACE1)
 #define HAL_PLATFORM_FUELGAUGE_MAX17043 (1)
 #define HAL_PLATFORM_FUELGAUGE_MAX17043_I2C (HAL_I2C_INTERFACE1)
+#define HAL_PLATFORM_POWER_MANAGEMENT_STACK_SIZE (1536)
```
avtolstoy, line 31: Also set for P2 to the same value.

avtolstoy (approved):

hal/src/trackerm/hal_platform_config.h
avtolstoy: Actually, probably set generally in `hal_platform_rtl872x_config.h`

PR #2536, opened 2022-09-13
[wiring] ApplicationWatchdog: fixes potential 2x timeout required  to fire

avtolstoy:
### Problem

See [SC](https://app.shortcut.com/particle/story/109275/application-watchdog-running-for-double-time)

This is a bug/problematic behavior in ApplicationWatchdog: https://github.com/particle-iot/device-os/blob/develop/wiring/inc/spark_wiring_watchdog.h#L52
https://github.com/particle-iot/device-os/blob/develop/wiring_globals/src/spark_wiring_watchdog.cpp#L22

`checkin()` gets called after `ApplicationWatchdog` thread has been started and the thread went to sleep. This updates `last_checkin` and upon wake-up in 10 minutes in this case the condition `now - last_checkin` will not be `>= timeout`, requiring another sleep cycle.

### Solution

Update `last_checkin` first, use `last_checkin` + `timeout` as initial timeout in watchdog thread.

### Steps to Test

- `wiring/no_fixture` `APPLICATION*`

### Example App

See SC

### References

- [SC109275]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/thread.cpp
+++ b/user/tests/wiring/no_fixture/thread.cpp
@@ -262,8 +262,8 @@ test(APPLICATION_WATCHDOG_01_fires_timeout)
 {
 	s_ram_free_before = System.freeMemory();
 	timeout_called = 0;
-	ApplicationWatchdog wd(5, timeout);
-	HAL_Delay_Milliseconds(10);
+	ApplicationWatchdog wd(50, timeout);
+	HAL_Delay_Milliseconds(60);
 
 	assertEqual((int)timeout_called, 1);
 	waitForComplete(wd);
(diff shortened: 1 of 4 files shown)


avtolstoy:
@keeramis @technobly ping

user/tests/wiring/no_fixture/thread.cpp
```diff
@@ -262,8 +262,8 @@ test(APPLICATION_WATCHDOG_01_fires_timeout)
 {
 	s_ram_free_before = System.freeMemory();
 	timeout_called = 0;
-	ApplicationWatchdog wd(5, timeout);
-	HAL_Delay_Milliseconds(10);
+	ApplicationWatchdog wd(50, timeout);
+	HAL_Delay_Milliseconds(60);
```
keeramis, line 266: @avtolstoy Is it reasonable to increase the timeout to 100msec? This particular test on my trackerM passes more consistently with increased delay.

keeramis (changes requested):

user/tests/wiring/no_fixture/thread.cpp
avtolstoy: No, this value needs to be under 2 * timeout value close to the actual timeout, this is the exact issue we are trying to fix.

keeramis (approved):

PR #2539, opened 2022-09-20
[hal] wifi: add generic 'world' country code as not every country code is exposed through API

avtolstoy:
### Problem

As title says. 2.4G channels 13-14 are not available by default and 14 is also not available in current country codes, same story with 5G. Even if only for testing allow explicitly allow all channels. The default is still US.

### References

N/A

---

### Completeness

- [ ] Added documentation


technobly (approved):
LGTM!

scott-brust (approved):
The RTK `channel_plan` values are defined [here](https://docs.google.com/spreadsheets/d/1pj02-m84nz5Irjv4O68Y8kxs2MVhpAz-/edit#gid=2054984874), correct? If so, looks good to me

Specifically the 'world' changes allow 2.4ghz channels 12,13 right?

PR #2551, opened 2022-10-07
[gen3][rtl872x]increase i2c buffer for tracker[m]. fix acquireWireBuffer

scott-brust:
### Problem
The tracker/trackerm platforms communicate with a handful of i2c devices on system startup. The system allocates and provides the default 32 byte sized I2C Wire buffers to the I2C HAL. The I2C hal is able to communicate and initialize the peripherals, all is well.

However, we have a new Quectel GNSS device that returns data longer than 32 bytes in its I2C responses. 

Normally, a user can allocate their own larger I2C buffers via defining a `acquireWireBuffer()` function. However, when the user app constructs its own `TwoWire` object and attempts to re-initialize the i2c HAL, the system code rejects it (in the case of the NRF hal), since it has already been initialized. 

see [SC-109914](https://app.shortcut.com/particle/story/109914/allow-user-app-to-override-i2c-buffer-on-tracker-trackerm-other-platforms-that-may-have-rtc-and-pmic) for details
### Solution

There should be 2 changes:
1. The default system i2c buffer should be increased for the tracker/trackerm platforms in order to allow the system to read any data it wishes from the GNSS device
2. The `acquireWireBuffer()` function should work on any platform, even if the system has allocated buffers already. The only caveat to this is that if a user wishes to override the system buffers, they must be larger than the currently allocated ones. 

### Steps to Test

Run example app

### Example App

```
SerialLogHandler g_logSerial(LOG_LEVEL_ALL);

hal_i2c_config_t allocateWireConfig(uint32_t bufferSize) {
    hal_i2c_config_t config = {
        .rx_buffer = new (std::nothrow) uint8_t[bufferSize],
        .rx_buffer_size = bufferSize,
        .tx_buffer = new (std::nothrow) uint8_t[bufferSize],
        .tx_buffer_size = bufferSize,
        .flags = HAL_I2C_CONFIG_FLAG_FREEABLE
    };
    return config;
}

void setup() {
  Wire.begin();
}

void loop() {

  hal_i2c_config_t biggerBuffer = allocateWireConfig(128);
  int reinitLargerResult = hal_i2c_init(HAL_I2C_INTERFACE1, &biggerBuffer);
  Log.info("I2C Hal Reinit bigger buffer result %d", reinitLargerResult);

  hal_i2c_config_t smallerBuffer = allocateWireConfig(10);
  int reinitSmallerResult = hal_i2c_init(HAL_I2C_INTERFACE1, &smallerBuffer);
  Log.info("I2C Hal Reinit smaller buffer result %d", reinitSmallerResult);

 delay(3000);
}
```

The first `hal_i2c_init()` call should succeed, as it is increasing the default i2c buffer from 32 -> 128
The second `hal_i2c_init()` call should fail, as it is trying to shrink the i2c buffer from 128 -> 10
Something like this:
```
0000009039 [app] INFO: I2C Hal Reinit bigger buffer result 0
0000009041 [app] INFO: I2C Hal Reinit smaller buffer result -191
```

### References

---

### Completeness

- [ ] Added documentation


hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -310,28 +313,46 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
...
+    } 
 
     // Capture the mutex and re-enable threading
     I2cLock lk(i2c);
     os_thread_scheduling(true, nullptr);
 
-    // Initialize internal data structure
+    if (i2cMap[i2c].configured) {
+        // Configured, but new buffers are invalid
+        if (!isConfigValid(config)){
```
avtolstoy, line 324: Nit: Can probably do this just once and short-circuit early on before configuring/reconfiguring.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -310,28 +313,46 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
...
+            return SYSTEM_ERROR_INVALID_ARGUMENT;
+        }
+        // Configured, but new buffers are smaller
+        if (config->rx_buffer_size < i2cMap[i2c].rx_buf_size ||
```
avtolstoy, line 328: `<=`, no point reassigning same-sized buffers.

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -94,6 +94,15 @@ class I2cClass {
         lock();
         os_thread_scheduling(true, nullptr);
         if (isConfigured()) {
+            // Configured, but new buffers are invalid
+            if (!isConfigValid(conf)){
+                return SYSTEM_ERROR_INVALID_ARGUMENT;
+            }
+            // Configured, but new buffers are smaller
+            if (conf->rx_buffer_size < rxBuffer_.size() ||
```
avtolstoy, line 102: `<=`

hal/inc/i2c_hal.h
```diff
@@ -52,6 +53,7 @@ typedef struct hal_i2c_config_t {
     uint32_t rx_buffer_size;
     uint8_t* tx_buffer;
     uint32_t tx_buffer_size;
+    uint8_t  freeable;
```
avtolstoy, line 56: Perhaps consider using `flags` instead of a single member variable for future extensibility.

hal/src/nRF52840/i2c_hal.cpp
scott-brust: The previous logic allowed initializing the HAL with a NULL config (ie it should use default buffers). Do we want to disallow this?

hal/inc/hal_platform.h
```diff
@@ -530,4 +530,8 @@
 #define HAL_PLATFORM_BLE_ACTIVE_EVENT (0)
 #endif // HAL_PLATFORM_BLE_ACTIVE_EVENT
 
+#ifndef HAL_PLATFORM_I2C_BUFFER_SIZE
+#define HAL_PLATFORM_I2C_BUFFER_SIZE (32)
```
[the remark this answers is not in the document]
scott-brust: I refactored things a bit, let me know if this is closer to what you imagined

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -94,14 +94,25 @@ class I2cClass {
...
             CHECK(deInit());
         }
         if (isConfigValid(conf)) {
             rxBuffer_.init((uint8_t*)conf->rx_buffer, conf->rx_buffer_size);
             txBuffer_.init((uint8_t*)conf->tx_buffer, conf->tx_buffer_size);
+            heapBuffer_ = ((conf->version >= HAL_I2C_CONFIG_VERSION_2) && (conf->flags & HAL_I2C_CONFIG_FLAG_FREEABLE));
         } else {
-            rxBuffer_.init((uint8_t*)malloc(I2C_BUFFER_LENGTH), I2C_BUFFER_LENGTH);
-            txBuffer_.init((uint8_t*)malloc(I2C_BUFFER_LENGTH), I2C_BUFFER_LENGTH);
+            int buffer_size = HAL_PLATFORM_I2C_BUFFER_SIZE(HAL_I2C_INTERFACE1);
```
scott-brust, line 113: Kind of cheated here. rtl872x platforms only have the one i2c instance

hal/src/nRF52840/i2c_hal.cpp
avtolstoy: Should still be allowed, yeah, let's leave as-is :)

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -310,28 +313,43 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
...
-        return SYSTEM_ERROR_NONE;
-    }
+    } 
 
     // Capture the mutex and re-enable threading
     I2cLock lk(i2c);
     os_thread_scheduling(true, nullptr);
 
-    // Initialize internal data structure
+    if (i2cMap[i2c].configured) {
```
avtolstoy, line 322: I think one additional thing here so that we somewhat don't leak memory is either to free the passed buffers in config if `HAL_I2C_CONFIG_FLAG_FREEABLE` here on errors or at least specifically on `SYSTEM_ERROR_NOT_ENOUGH_DATA`, or do this in `TwoWire` perhaps.

avtolstoy (commented):
@scott-brust We should also be able to add an automated test for this for tracker/trackerm and read over 32 bytes from one of the available components like PMIC/FG/RTC.

hal/inc/hal_platform.h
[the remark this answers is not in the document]
avtolstoy: Yeah, looks better for Tracker now :+1:

hal/src/trackerm/hal_platform_config.h
```diff
@@ -9,7 +9,7 @@
 #define HAL_PLATFORM_CELLULAR_SERIAL (HAL_USART_SERIAL2)
 #define HAL_PLATFORM_SETUP_BUTTON_UX (1)
 #define HAL_PLATFORM_SPI_NUM (2)
-#define HAL_PLATFORM_I2C_NUM (2)
```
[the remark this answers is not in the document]
scott-brust: I was curious about `HAL_PLATFORM_I2C_NUM`. Its only used in the nrf i2c/sleep HALs. 
I also saw in `hal_platform.h` that there is this comment:
```
// FIXME: HAL_PLATFORM_I2C_NUM
#ifndef HAL_PLATFORM_I2C1
#define HAL_PLATFORM_I2C1 (1)
#endif /* HAL_PLATFORM_I2C1 */
```
I can see that `HAL_PLATFORM_I2CX` is what results in defining `Wiring_WireX` and Ultimately `WireX`. 

Which macros should be defined for a given platform? `HAL_PLATFORM_I2C_NUM (X)`?
 or `HAL_PLATFORM_I2C1`, `HAL_PLATFORM_I2C2`, etc ?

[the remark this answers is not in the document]
avtolstoy: `HAL_PLATFORM_I2C_NUM` - number of supported i2c peripherals.
`HAL_PLATFORM_I2C1`, `HAL_PLATFORM_I2C2` ... for each i2c peripheral supposed to be enabled in Wiring.

We do the same thing with e.g. `HAL_PLATFORM_USART2` on cellular platforms where `HAL_PLATFORM_USART_NUM` is 2, but we only expose `Serial1` and don't set `HAL_PLATFORM_USART2`.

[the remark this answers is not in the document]
avtolstoy: What I'm trying to say: `HAL_PLATFORM_XXX_NUM` - low level capabilities and what is supported in HAL`. `HAL_PLATFORM_PERIPHERALNAMEX` - enables a particular peripheral in higher level APIs etc (e.g. wiring).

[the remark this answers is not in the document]
avtolstoy: It's useful to have these for tests as well.

scott-brust:
> @scott-brust We should also be able to add an automated test for this for tracker/trackerm and read over 32 bytes from one of the available components like PMIC/FG/RTC.

Ok, ill try and read >32 bytes from one of the components and see what they return. 
Should this be a new test under `wiring/integration/no_fixture_i2c/`? Or a standalone test app? 
Either way it will only support tracker/trackerm platforms.

avtolstoy:
@scott-brust New test under `no_fixture_i2c` :+1: Let's avoid standalone apps whenever possible :) This is the case where this can be a simple automated test.

scott-brust:
@avtolstoy Hmm there is `wiring/integration/i2c.cpp` that has some tests already. I was looking at the RTC and it seems like I should be able to do a ~64 byte register read from it and check the values as a good test of the I2C HAL. Should I add tests here or make my own standalone `wiring/no_fixture_i2c`?

avtolstoy:
If you are testing `acquireWire1Buffer` behavior as well, it makes sense to move this into a separate test suite, something like `wiring/i2c_buffers`, as it gets called only once on boot, so you'd need a pretty much standalone test application for that.

scott-brust:
Are there any more tests that I should add to `no_fixture_i2c`?

wiring/src/spark_wiring_i2c.cpp
```diff
@@ -27,13 +27,20 @@
...
 
 // Constructors ////////////////////////////////////////////////////////////////
 
 TwoWire::TwoWire(hal_i2c_interface_t i2c, const hal_i2c_config_t& conf)
 {
   _i2c = i2c;
-  hal_i2c_init(_i2c, &conf);
+  int result = hal_i2c_init(_i2c, &conf);
+  if (result == SYSTEM_ERROR_NOT_ENOUGH_DATA && 
+    (conf.version >= HAL_I2C_CONFIG_VERSION_2) &&
```
avtolstoy, line 39: No need to check version here, application always has the latest structures.

avtolstoy: Nit/style: add extra indent for all newlined if expressions

user/tests/integration/wiring/no_fixture_i2c/application.cpp
```diff
@@ -0,0 +1,40 @@
...
+// });
+
+UNIT_TEST_APP();
+
+// Enable threading if compiled with "USE_THREADING=y"
+#if PLATFORM_THREADING == 1 && USE_THREADING == 1
+SYSTEM_THREAD(ENABLED);
+#endif
+
+SYSTEM_MODE(MANUAL);
```
avtolstoy, line 38: `SEMI_AUTOMATIC`.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,131 @@
+
```
avtolstoy, line 1: Nit: copyright header

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,131 @@
...
+test(I2C_00_hal_init_default_buffer)
+{
+    // Initializing the HAL with a null config is allowed and should allocate system defaults
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, nullptr), (int)SYSTEM_ERROR_NONE);
+
+#if (PLATFORM_ID == PLATFORM_TRACKER) || (PLATFORM_ID == PLATFORM_TRACKERM)
+    // Tracker platforms should have buffers at least 512 bytes large, allocated by the system on startup
+    // IE trying to re-initialize the HAL with same size should fail
+    hal_i2c_config_t wire1_small_buffers = allocateWireConfig(WIRE1_TRACKER_MINIMUM_BUFFER_SIZE);
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE2, &wire1_small_buffers), (int)SYSTEM_ERROR_NOT_ENOUGH_DATA);
```
avtolstoy, line 50: This is not going to work on Tracker M, there is only 1 I2C peripheral. You should use `HAL_PLATFORM_EXTERNAL_RTC_I2C` if that's what is being used for tests on these platforms.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // Larger sized buffers should succeed
+    hal_i2c_config_t wire_larger_buffers = allocateWireConfig(WIRE_ACQUIRE_BUFFER_SIZE + 1);
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, &wire_larger_buffers), (int)SYSTEM_ERROR_NONE);
+    freeWireBuffers(&wire_larger_buffers);
+
+    // Reinitializing with a NULL config is not allowed
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, nullptr), (int)SYSTEM_ERROR_INVALID_ARGUMENT);
+
+    #if Wiring_Wire1
```
avtolstoy, line 74: Nit: no indent

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+test(I2C_01_test_acquire_wire_buffer) {
+    // Construct TwoWire object with user acquireWireBuffer() configuration structures, which should re-initialize the I2C HAL
+    Wire.begin();
+
+    // Attempt to re-initialize the HAL with different sized buffers to implicitly determine if the user buffers were used by the HAL
+    // System Defaults < User values from acquireWireBuffer < Values used here to re-initialize the HAL
+
+    // Equal size buffers should fail
+    hal_i2c_config_t wire_smaller_buffers = allocateWireConfig(WIRE_ACQUIRE_BUFFER_SIZE);
```
avtolstoy, line 62: Nit/style: camelCase

avtolstoy (changes requested):

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
scott-brust: My understanding of the requirements from the original ticket [sc-109914](https://app.shortcut.com/particle/story/109914/allow-user-app-to-override-i2c-buffer-on-tracker-trackerm-other-platforms-that-may-have-rtc-and-pmic) is that the i2c busses that connect to peripherals that return >32 bytes of data should have 512 byte buffers by default. 
The only device that returns >32 bytes of data currently is the new Quectel GNSS chip.
This chip is only present on Trackerm with its single I2C interface, and I think it will be on a new tracker variant on I2C2. So this test should be testing those two specific instances.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,152 @@
...
+    uint8_t longReadBuffer[REGISTER_RANGE] = {};
+    uint8_t readLength = REGISTER_RANGE; 
+
+    // Check read values against ID registers
+    const uint8_t ID0_REGISTER = 0x28;
+    const uint8_t ID0_REGISTER_VALUE = 0x18;
+    const uint8_t ID1_REGISTER = 0x29;
+    const uint8_t ID1_REGISTER_VALUE = 0x05;
+
+    hal_i2c_interface_t wireInterface = HAL_PLATFORM_EXTERNAL_RTC_I2C;
```
scott-brust, line 125: It looks like only Tracker platform has the external RTC (not trackerm). If there is another platform that does incorporate an RTC, it just needs to define `HAL_PLATFORM_EXTERNAL_RTC`, `HAL_PLATFORM_EXTERNAL_RTC_I2C`, `HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR` and this test should work.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // Larger sized buffers should succeed
+    hal_i2c_config_t wireLargerBuffers = allocateWireConfig(WIRE_ACQUIRE_BUFFER_SIZE + 1);
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, &wireLargerBuffers), (int)SYSTEM_ERROR_NONE);
+    freeWireBuffers(&wireLargerBuffers);
+
+    // Reinitializing with a NULL config is not allowed
+    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, nullptr), (int)SYSTEM_ERROR_INVALID_ARGUMENT);
+
+#if Wiring_Wire1
```
avtolstoy, line 93: Nit: I'd probably move to a different test.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
avtolstoy: What about using PMIC or FG?

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,152 @@
...
+    uint8_t rtcRegister = 0x00;
+
+    hal_i2c_begin_transmission(wireInterface, rtcAddress, NULL);
+    hal_i2c_write(wireInterface, rtcRegister, NULL);
+    hal_i2c_end_transmission(wireInterface, true, NULL);
+    auto config = WireTransmission(rtcAddress).quantity(readLength).stop(true).halConfig();
+    hal_i2c_request_ex(wireInterface, &config, nullptr);
+
+    int32_t size = hal_i2c_available(wireInterface, NULL);
+    assertMore(size, 0);
```
avtolstoy, line 136: > readLength?

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    // Log.info("I2C_02_long_read result %d\n", size);
+    // Log.dump(long_read_buffer, readLength);
+    // Log.info(" \n");
+
+    assertEqual(longReadBuffer[ID0_REGISTER], ID0_REGISTER_VALUE);
+    assertEqual(longReadBuffer[ID1_REGISTER], ID1_REGISTER_VALUE);
+
+    assertTrue(true);
```
avtolstoy, line 149: Doesn't do anything? :)

avtolstoy (approved):

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
scott-brust: The MAX17043 fuel gauge only has [15 bytes of register data](https://datasheets.maximintegrated.com/en/ds/MAX17043-MAX17044.pdf)
The BQ24195 PMIC only has [10 bytes of register data](https://www.ti.com/lit/ds/symlink/bq24195.pdf?ts=1666034756902&ref_url=https%253A%252F%252Fwww.ti.com%252Fproduct%252FBQ24195)
So only the RTC has >32 bytes of contiguous register data to conveniently read (or the actual Quectel GNSS device on trackerm, which I dont have available to test against)

avtolstoy: I think most of the times you just wrap around at the maximum. Might be worth trying, all we need for the IC to not NACK

scott-brust: I can indeed read more bytes than the device has available from the BQ24195 PMIC:
```
0000002665 [app] INFO: I2C_03_long_read_2 result 64                                                                                                                                                                                                                                                                                   
021b18119a8a034b6c0023ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```
I will keep the RTC test and add one for the PMIC? Is it ok to have both, or are they too similar/redundant?

avtolstoy: I'd say add both. Make sure to enable PMIC only for platforms with `HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL == 0`

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
scott-brust: ok, should I have a similar test for `acquireWire3Buffer()` just to be complete?

avtolstoy: Perhaps makes sense, yeah, for platforms with `Wiring_Wire3` which would as of right now be only Tracker.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
scott-brust: Ok, checked for `HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL == 0`, which limits this to tracker it seems.

user/tests/integration/wiring/no_fixture_i2c/i2c.cpp
scott-brust: I can test `acquireWire3Buffer()` and `Wire3.begin()`, but then the `Wire1` tests fail with PMIC + RTC. 
I know Wire3 on tracker conflicts with UART pins and the `Wire` TWI handler, but it shouldn't affect `Wire1` right? IE `Wire1` and `Wire3` on tracker should be independently usable? 
I tried `Wire3.end()` before the RTC tests but they still fail. Not sure whats going wrong

avtolstoy: Yeah, sounds weird. Wire1 is TWIM1 and Wire3/Wire are TWIM0, so not sure what's going on.

PR #2559, opened 2022-10-18
Fix no_fixture_i2c for esomx,boron,bsom platforms

scott-brust:
### Problem

`no_fixture_i2c` tests were only validated against `p2` and `tracker` platforms. Unfortunately other platforms fail the tests.

### Solution

`boron`: For some reason, the PMIC on boron NAKs bytes after 34 reads, where the same part on tracker will allow 64 byte reads.
`bsom`: The BSOM initializes the PMIC on boot, so the `Wire` interface is already initialized when attempting to init with a null buffer

### Steps to Test

`device-os-test --test-dir ~/develop/device-os/user/tests/integration --device-os-dir ~/develop/device-os run bsom wiring/no_fixture_i2c -v `

### Example App

`no_fixture_i2c`

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture_i2c/i2c.cpp
+++ b/user/tests/wiring/no_fixture_i2c/i2c.cpp
@@ -131,9 +131,7 @@ test(I2C_02_test_acquire_wire1_buffer) {
 test(I2C_03_long_read_rtc)
 {
     // Test > 32 byte reads by reading from AM18x5 RTC peripheral
-    Wire1.begin();
-
-    // Read full register range
+    // Read beyond full register range
     const uint8_t REGISTER_RANGE = 64;
     uint8_t longReadBuffer[REGISTER_RANGE] = {};
     uint8_t readLength = REGISTER_RANGE; 
@@ -147,6 +145,8 @@ test(I2C_03_long_read_rtc)
     hal_i2c_interface_t wireInterface = HAL_PLATFORM_EXTERNAL_RTC_I2C;
     uint8_t rtcAddress = HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR;
     uint8_t rtcRegister = 0x00;
+
+    hal_i2c_begin(wireInterface, I2C_MODE_MASTER, 0x00, NULL);
 
     hal_i2c_lock(wireInterface, nullptr);
     SCOPE_GUARD({
--- a/user/tests/wiring/no_fixture_i2c/i2c.cpp
+++ b/user/tests/wiring/no_fixture_i2c/i2c.cpp
@@ -57,7 +57,11 @@ hal_i2c_config_t acquireWire1Buffer()
 test(I2C_00_hal_init_default_buffer)
 {
     // Initializing the HAL with a null config is allowed and should allocate system defaults
-    assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, nullptr), (int)SYSTEM_ERROR_NONE);
+    if (!hal_i2c_is_enabled(HAL_I2C_INTERFACE1, nullptr)) {
+        assertEqual(hal_i2c_init(HAL_I2C_INTERFACE1, nullptr), (int)SYSTEM_ERROR_NONE);    
+    } else {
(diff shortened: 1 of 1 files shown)


avtolstoy:
As discussed:

1. enums can't be used in preprocessor directives
2. Make sure to lock i2c peripherals while messing with reinitialization/buffers just in case in order not to confuse Device OS when talking to PMIC/FG/RTC/etc, make sure to Wire.begin() on test completion e.g. with scope guard

scott-brust:
Retested branch on these platforms to confirm tests pass:
P2
BSOM
Boron
Esomx
Tracker
Argon

Only platform untested is trackerm

scott-brust:
> As discussed:
> 
> 1. enums can't be used in preprocessor directives
> 2. Make sure to lock i2c peripherals while messing with reinitialization/buffers just in case in order not to confuse Device OS when talking to PMIC/FG/RTC/etc, make sure to Wire.begin() on test completion e.g. with scope guard

@avtolstoy Let me know if my changes are in line with these requests

user/tests/wiring/no_fixture_i2c/i2c.cpp
```diff
@@ -105,13 +124,11 @@ test(I2C_02_test_acquire_wire1_buffer) {
...
-
-    hal_i2c_end(HAL_I2C_INTERFACE2, nullptr);
 }
 #endif
 
 #if HAL_PLATFORM_EXTERNAL_RTC
-test(I2C_04_long_read_rtc)
+test(I2C_03_long_read_rtc)
 {
     // Test > 32 byte reads by reading from AM18x5 RTC peripheral
```
avtolstoy, line 133: This needs to look at `HAL_PLATFORM_EXTERNAL_RTC_I2C`, right?

avtolstoy (approved):

user/tests/wiring/no_fixture_i2c/i2c.cpp
scott-brust: It does. if `HAL_PLATFORM_EXTERNAL_RTC` is defined, then `HAL_PLATFORM_EXTERNAL_RTC_I2C` must also be defined by these checks in `hal_platform.h`

```
#if HAL_PLATFORM_EXTERNAL_RTC
# ifndef HAL_PLATFORM_EXTERNAL_RTC_I2C
#  error "HAL_PLATFORM_EXTERNAL_RTC_I2C is not defined"
# endif /* HAL_PLATFORM_EXTERNAL_RTC_I2C */
# ifndef HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR
#  error "HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR is not defined"
# endif /* HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR */
# ifndef HAL_PLATFORM_EXTERNAL_RTC_CAL_XT
#  define HAL_PLATFORM_EXTERNAL_RTC_CAL_XT (0)
# endif /* HAL_PLATFORM_EXTERNAL_RTC_CAL_XT */
#endif /* HAL_PLATFORM_EXTERNAL_RTC */
```
I use these defines in the test itself
```
    hal_i2c_interface_t wireInterface = HAL_PLATFORM_EXTERNAL_RTC_I2C;
    uint8_t rtcAddress = HAL_PLATFORM_EXTERNAL_RTC_I2C_ADDR;
```

avtolstoy: I guess the only thing then is removing this `Wire1.begin()` then: https://github.com/particle-iot/device-os/pull/2559/files/98a9bd5aee8f26bcea23c47ead5e206b8a3a1285#diff-bf15c43f7b50354a9dee0d64fbb7cd3bd2b5bfcfd706d009369069cce9f4cfddR134

scott-brust: Ah good point, I will replace it with `hal_i2c_init()` / `hal_i2c_begin()` using the defined interface

PR #2568, opened 2022-10-25
Turn off cell modem on Tracker M sleep

jmwilson:
### Description

`system_sleep_network_suspend` is effectively a no-op on P2 platforms, but this should only apply to the WiFi interface. 

### Steps to Test

Manual testing that calling `System.sleep` will automatically turn off the cell modem on Tracker M proto boards.

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/system_sleep.cpp
+++ b/system/src/system_sleep.cpp
@@ -81,13 +81,17 @@ network_status_t system_sleep_network_suspend(network_interface_index index) {
         }
 #endif
 #if HAL_PLATFORM_RTL872X
-        // P2 doesn't need to turn off the modem manually
-        system_notify_event(network_status, network_status_off);
-#else
+        if (index == NETWORK_INTERFACE_WIFI_STA) {
+            // P2 doesn't need to turn off the modem manually
+            system_notify_event(network_status, network_status_off);
+        } else {
+#endif
         network_off(index, 0, 0, NULL);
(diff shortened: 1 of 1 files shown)


system/src/system_sleep.cpp
```diff
@@ -81,13 +81,17 @@ network_status_t system_sleep_network_suspend(network_interface_index index) {
         }
 #endif
 #if HAL_PLATFORM_RTL872X
-        // P2 doesn't need to turn off the modem manually
-        system_notify_event(network_status, network_status_off);
-#else
+        if (index == NETWORK_INTERFACE_WIFI_STA) {
```
avtolstoy, line 84: This is a platform-specific hack which should not be here in generic system code. I understand it's kind of already here and the reason for that was to be able to bring up P2 stuff quicker, but I'd prefer that we fixed the underlying problem, which is in RtlNcpNetif/RtlNcpClient with regards to how power on/off requests/events are managed. cc @XuGuohui

avtolstoy (changes requested):

system/src/system_sleep.cpp
XuGuohui: I agree. I will fix the power leakage issue, along with refactoring this alittle bit.

jmwilson: OK, I will close this PR and assign the shortcut ticket to @XuGuohui

PR #2569, opened 2022-10-28
[rtl872x] prebootloader: enable BOR with lowest available thresholds

avtolstoy:
As title says.


XuGuohui (approved):

scott-brust:
FWIW I tested this branch with a power supply. I added some code to toggle GPIO pins when 1) The `bodIrqHandler()` ISR is triggered and 2) when KM0 detects `BIT_AON_BOD_RESET_FLAG` on boot. I can see the brown out ISR firing when lowering voltage to ~2.9-3.0v, and then when powered up to >4.0v the prebootloader detects the brownout bit. When the system part boots it reports `RESET_REASON_POWER_BROWNOUT` via `System.resetReason()`
See my annotated capture here
annotated capture here
e

PR #2570, opened 2022-11-01
Server key rotation

sergeuz:
### Problem

This PR extends the cloud protocol to add the possibility for the server to remotely update the public key and address that are used by the device to establish a connection with the cloud. The details on the implementation can be found in the related story and SDD.

This PR also reverts some of the changes from https://github.com/particle-iot/device-os/pull/1300 and adds a more reliable detection of corrupted server settings. The settings are validated on startup and restored persistently if necessary.

### Steps to Test

See the PR with the server-side implementation of this feature.

### References

- sc-62023


avtolstoy:
@sergeuz I've squashed this branch as there were a lot of conflicts to resolve when rebasing on develop, squashing made that a bit easier. There is a backup in https://github.com/particle-iot/device-os/tree/server_key_rotation/sc-62023-bak

I've also disabled `NRFX_ASSERT()` for nRF SDK to save quite a lot on flash space.

avtolstoy (approved):

PR #2572, opened 2022-11-02
Fix i2c hal deadlock

scott-brust:
### Problem

James reported a trackerm device that would not breathe cyan when running on battery power. The device appeared deadlocked. After investigation, the deadlock was traced to the i2c HAL mutex. The power management thread held the I2C mutex and the user application thread was blocked waiting on it. However, the scheduler was disabled so the power management thread could never execute and release the lock, resulting in a deadlock. 

### Solution

Re-enable scheduling when acquiring the i2c hal lock in order to allow any other tasks using it to execute and release their lock

### Steps to Test

Running any user app that uses the PMIC/Fuel gauge should work on trackerm hardware

### Example App

```c
void setup() {
   */
}

void loop() {
    FuelGauge fuel;
    Log.info( "voltage=%.2f", fuel.getVCell() );
    delay(5000);
}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/i2c_hal.cpp
+++ b/hal/src/rtl872x/i2c_hal.cpp
@@ -91,16 +91,18 @@ public:
         if (!mutex_) {
             os_mutex_recursive_create(&mutex_);
         }
-        lock();
         os_thread_scheduling(true, nullptr);
+        lock();
         if (isConfigured()) {
             // Configured, but new buffers are invalid
             if (!isConfigValid(conf)){
+                unlock();
                 return SYSTEM_ERROR_INVALID_ARGUMENT;
             }
             // Configured, but new buffers are smaller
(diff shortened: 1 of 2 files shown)


XuGuohui (approved):

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -91,16 +91,18 @@ class I2cClass {
         if (!mutex_) {
             os_mutex_recursive_create(&mutex_);
         }
-        lock();
         os_thread_scheduling(true, nullptr);
+        lock();
```
avtolstoy, line 95: Let's switch this to RAII `I2cLock` as well.

avtolstoy (approved):

PR #2580, opened 2022-11-15
Updates for Tracker M proto 3

jmwilson:
### Explanation

Tracker M proto 3 has the wakeup pin to _or_ the PMIC and IO expander interrupts so either can wake it from deep sleep. This means the IO expander should set its interrupt output as open drain.

In proto 2, the cell reset line was disconnected via a DNP jumper. In proto 3, is connected to the cell modem reset (which is internally connected to modem power). Since the cell modem pins are level-shifted from the MCU pins via open drain MOSFETs it is ok to have the MCU pins driven separately and will mean the pinmap doesn't have to change for an EG91 variant.

### Completeness

- [ ] Added documentation


hal/shared/mcp23s17.cpp
```diff
@@ -135,9 +135,19 @@ int Mcp23s17::reset(bool verify) {
...
 
+#if HAL_PLATFORM_MCP23S17_MIRROR_INTERRUPTS || HAL_PLATFORM_MCP23S17_OPEN_DRAIN_INT
+    uint8_t iocon {};
+
 #if HAL_PLATFORM_MCP23S17_MIRROR_INTERRUPTS
-    CHECK(writeRegister(IOCON_ADDR[0], 1u << 6));
+    iocon |= 1u << 6;
+#endif
+#if HAL_PLATFORM_MCP23S17_OPEN_DRAIN_INT
+    iocon |= 1u << 2;
```
avtolstoy, line 145: Nit: magic numbers. We have plenty of those here, but if we can define self-describing constants for these would be great.

avtolstoy (approved):

hal/shared/mcp23s17.cpp
jmwilson: I generally agree, but in a case where the descriptive flag is right next to the magic number that's used once, it's not adding much. 🤷‍♂️

PR #2581, opened 2022-11-16
[freertos] [experimental] multi-step priority disinheritance

avtolstoy:
### Description

This PR (or more precisely https://github.com/particle-iot/freertos/pull/2) implements and enables (for now on RTL872x-based platforms) an optional multi-step priority disinheritance mechanism for FreeRTOS mutexes.

The default implementation in FreeRTOS only demotes the priority of a thread, which had its priority raised previously due to priority inheritance mechanism, once the thread releases its last held mutex.

This is problematic for cases where the thread acquires a set of mutexes one after another and continues holding at least one of them. With the default implementation such thread may be runnning with a raised priority for prolonged amounts of time, which has caused us various problems in the past and even now again (on P2 with listening mode and serial setup console + BLE).

**This PR also "fixes" a problem with RTL872x SDK trying to free memory under a critical section (disabled interrupts/scheduling).**

This PR also adds base and current priorities into `os_thread_dump()` introspection function and enables it on RTL872x-based platforms and by default in general.

#### A simple example of priority inheritance mechanism

inheritance1

#### FreeRTOS "simple" priority disinheritance
inheritance2

#### Multi-step priority disinheritance
inheritance3

### Caveats

The implementation is pretty simple **and may not cover some corner cases** but should be good enough to resolve issues for the common pattern that Device OS may run into: acquiring one mutex and making other calls that also require mutexes.

To make this work we simply need to keep a priority-sorted list of held mutexes (by default FreeRTOS only keeps track of the number of held mutexes) and occasionally (when timeouting to acquire the mutex or new threads are waiting for us to release a held mutex) update the list: update the inherited/required priority and re-insert into the list so that the highest priority entry is on top. See https://github.com/particle-iot/freertos/pull/2 for the implementation itself.

### Tests

- wiring/no_fixture: `CONCURRENT_MUTEX_01_priority_inheritance_two_threads`, `CONCURRENT_MUTEX_02_priority_inheritance_three_threads`
- TODO: more complex cases?


scott-brust:
Posted some questions in the [freertos PR](https://github.com/particle-iot/freertos/pull/2)
Also had a question about the realtek sdk changes:
We re-enable interrupts in the critical section in order to allow the `vPortFree(pbuf);` call to succeed? Why do we need to do this, does this call require hardware interrupts?

avtolstoy:
Heap allocations/deallocations are protected by a mutex, you can't acquire it under disabled interrupts/scheduling, that's going to cause a deadlock if it's being held by some other thread.

scott-brust:
> 

ok so `portDISABLE_INTERRUPTS()` also disabled the scheduler?

avtolstoy:
The scheduler is disabled as a consequence when interrupts are disabled. You cannot have context switches if your `PendSV` interrupt cannot fire.

scott-brust:
> The scheduler is disabled as a consequence when interrupts are disabled. You cannot have context switches if your `PendSV` interrupt cannot fire.

Ah, yes, totally makes sense now. So if we free memory and block on the malloc semaphore, we need to be able to schedule the task that is holding it, which will require the pendsv interrupt to switch task contexts. 

Are we at risk of other interrupts happening here while we have them enabled? I guess that is a risk we have to take?

avtolstoy:
@XuGuohui is reporting this to Realtek folks, so hopefully this is going to be resolved properly, for now we are sticking with this hacky workaround, so yes, there is a chance that something goes wrong but it's better than having a chance of a complete deadlock.

XuGuohui (approved):

YutingYou (approved):

PR #2597, opened 2022-12-01
Use OTP Feature flag to change ADC reference source

scott-brust:
### Problem

There are potentially many different platforms that may end up using the richtek DCDC part with the excessive ripple voltage. For already shipped `BRN404X` platforms, the `HW_VERSION` OTP field can be used to identify them. For other SKUs, we need a different method.

### Solution

The `HW Feature Flags` bytes of HW INFO in OTP have bits available. However, the `tracker` platform [uses all but one of these bits](https://docs.google.com/document/d/1NWEVTn1BI07rdKzzvx_HsGqpiBn-hIN0Kc91B989_3c/edit#heading=h.35xkyity7v7y). We will use the last OTP feature flag bit to denote the presence of the RIchtek part, and thus the need to switch to the internal ADC reference voltage. 

### Steps to Test

1. Clear the OTP bit
2. Reboot, and the ADC should use the internal reference

### Example App

```
#define PARTICLE_USE_UNSTABLE_API
#include "storage_hal.h"

void burnOTP(){
    const uintptr_t HW_DATA_OTP_ADDRESS = 0x00000022;

    uint8_t hwFeatures[] = {0x00, 0x00};
    int r = hal_storage_read(HAL_STORAGE_ID_OTP, HW_DATA_OTP_ADDRESS, hwFeatures, 2);
    Log.info("Read result: %d / %x %x", r, hwFeatures[0], hwFeatures[1]);

    // Clear bit 15 of the HW Feature Flags bytes
    hwFeatures[1] &= 0x7F;
    r = hal_storage_write(HAL_STORAGE_ID_OTP, HW_DATA_OTP_ADDRESS, hwFeatures, 2);
    Log.info("Write result: %d", r);

    r = hal_storage_read(HAL_STORAGE_ID_OTP, HW_DATA_OTP_ADDRESS, hwFeatures, 2);
    Log.info("Read result after writing: %d / %x %x", r, hwFeatures[0], hwFeatures[1]);
}

void setup()
{
    // Init the ADC HAL to force reading of the OTP bits
    hal_adc_read(D0);
}

void loop()
{
    if (!Serial.available()) {
        return;
    }

    auto ch {Serial.read()};

    switch (ch) {
    case '1':
        burnOTP();
        break;
    default:
        Serial.printf("Invalid selection\r\n");
        break;
    }

    hal_device_hw_info hwInfo = {};
    hal_get_device_hw_info(&hwInfo, nullptr);
    Log.info("Feature Bits: %08x Top byte %02x Low Byte %02x", 
        hwInfo.features, 
        (hwInfo.features & 0x0000FF00) >> 8,
        hwInfo.features & 0x000000FF);
}
```

Burning the bit should show this
```
0000017308 [app] INFO: Feature Bits: 0000ebfe Top byte eb Low Byte fe
0000017308 [app] INFO: Read result: 2 / fe eb
0000017309 [app] INFO: Write result: 2
0000017309 [app] INFO: Read result after writing: 2 / fe 6b
0000022312 [app] INFO: Feature Bits: 00006bfe Top byte 6b Low Byte fe
```

After reboot, the ADC source will be internal. 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/adc_hal.cpp
+++ b/hal/src/nRF52840/adc_hal.cpp
@@ -141,6 +141,7 @@ void hal_adc_dma_init() {
 #endif // PLATFORM_ID == PLATFORM_BORON
 
     hal_device_hw_info hwInfo = {};
+    hwInfo.size = sizeof(hwInfo);
     hal_get_device_hw_info(&hwInfo, nullptr);
     if ((hwInfo.features & HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT) == 0) {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
--- a/hal/src/nRF52840/adc_hal.cpp
+++ b/hal/src/nRF52840/adc_hal.cpp
@@ -38,6 +38,7 @@ static const nrfx_saadc_config_t saadcConfig = {
 
 static hal_adc_reference_t adcReference = HAL_ADC_REFERENCE_EXTERNAL;
 static constexpr uint32_t HAL_VERSION_BRN404X_V003 = 0x03;
+static constexpr uint32_t HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT = (1<<15);
 
 static void analog_in_event_handler(nrfx_saadc_evt_t const *p_event) {
     (void) p_event;
@@ -138,7 +139,11 @@ void hal_adc_dma_init() {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
 #else
-    (void)adcReference;
+    hal_device_hw_info hwInfo = {};
+    hal_get_device_hw_info(&hwInfo, nullptr);
+    if ((hwInfo.features & HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT) == 0) {
+        adcReference = HAL_ADC_REFERENCE_INTERNAL;
+    }
 #endif // PLATFORM_ID == PLATFORM_BORON
 }
 
--- a/hal/src/nRF52840/adc_hal.cpp
+++ b/hal/src/nRF52840/adc_hal.cpp
@@ -138,13 +138,13 @@ void hal_adc_dma_init() {
     if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_BRN404X_V003) {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
-#else
+#endif // PLATFORM_ID == PLATFORM_BORON
+
     hal_device_hw_info hwInfo = {};
     hal_get_device_hw_info(&hwInfo, nullptr);
     if ((hwInfo.features & HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT) == 0) {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
-#endif // PLATFORM_ID == PLATFORM_BORON
 }
 
 /*


hal/src/nRF52840/adc_hal.cpp
```diff
@@ -138,7 +139,11 @@ void hal_adc_dma_init() {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
 #else
-    (void)adcReference;
+    hal_device_hw_info hwInfo = {};
+    hal_get_device_hw_info(&hwInfo, nullptr);
+    if ((hwInfo.features & HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT) == 0) {
+        adcReference = HAL_ADC_REFERENCE_INTERNAL;
+    }
```
scott-brust, line 146: Do we want to check this feature flag on `PLATFORM_BORON` as well? In case we move forward with a HW version 4 or something?

hal/src/nRF52840/adc_hal.cpp
```diff
@@ -138,7 +139,11 @@ void hal_adc_dma_init() {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
 #else
-    (void)adcReference;
+    hal_device_hw_info hwInfo = {};
```
avtolstoy, line 142:
```c++
#if PLATFORM_ID == PLATFORM_BORON
    uint32_t hwVersion = 0xFF;
    auto ret = hal_get_device_hw_version(&hwVersion, nullptr);
    if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_BRN404X_V003) {
        adcReference = HAL_ADC_REFERENCE_INTERNAL;
        return;
    }
#endif // PLATFORM_ID == PLATFORM_BORON

    hal_device_hw_info hwInfo = {};
    hwInfo.size = sizeof(hwInfo);
    hal_get_device_hw_info(&hwInfo, nullptr);
    if ((hwInfo.features & HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT) == 0) {
        adcReference = HAL_ADC_REFERENCE_INTERNAL;
    }
```

hal/src/nRF52840/adc_hal.cpp
```diff
@@ -38,6 +38,7 @@ static const nrfx_saadc_config_t saadcConfig = {
 
 static hal_adc_reference_t adcReference = HAL_ADC_REFERENCE_EXTERNAL;
 static constexpr uint32_t HAL_VERSION_BRN404X_V003 = 0x03;
+static constexpr uint32_t HW_FEATURE_FLAG_USE_INTERNAL_ADC_REFERENCE_BIT = (1<<15);
```
avtolstoy, line 41: Double check endianness.

avtolstoy (changes requested):

hal/src/nRF52840/adc_hal.cpp
scott-brust: `hal_get_device_hw_info()` Puts the bytes in the right spots, I can examine `adcReference` with debugger after HAL init and can see its detecting the cleared bit in the right spot:
```
(gdb) p adcReference
$1 = HAL_ADC_REFERENCE_INTERNAL
```

hal/src/nRF52840/adc_hal.cpp
```diff
@@ -137,9 +138,13 @@ void hal_adc_dma_init() {
     if (ret == SYSTEM_ERROR_NONE && hwVersion == HAL_VERSION_BRN404X_V003) {
         adcReference = HAL_ADC_REFERENCE_INTERNAL;
     }
-#else
-    (void)adcReference;
 #endif // PLATFORM_ID == PLATFORM_BORON
+
+    hal_device_hw_info hwInfo = {};
```
avtolstoy, line 143: size needs to be set, see resolved comment

scott-brust: ok, is this checked anywhere? What about version? Or is this just implicitly version 0?

avtolstoy: It's going to be checked later on if that struct is extended and one is going to forget to update this particular instance of a call to `hal_device_hw_info()` and it will silently stop working. All sized structs need to have size initialized. Version can be kept as-is/default.

scott-brust: Ok got it 👍 will keep that in mind for similar structs in the future

avtolstoy (approved):

PR #2599, opened 2022-12-07
Free memory from rtl sdk in SystemISRTaskQueue

scott-brust:
### Problem

See ticket [here](https://app.shortcut.com/particle/story/112961/deadlock-on-p2-trackerm-when-blocking-on-mutex-with-interrupts-disabled)

### Solution

Instead of freeing memory in a critical section with interrupts disabled, defer the execution to the system ISR queue
See realtek SDK PR [here](https://github.com/particle-iot/realtek_ambd_sdk_public/pull/2)

### Steps to Test

Run on device tests

### Example App

Device tests

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/rtl_sdk_support.cpp
+++ b/hal/src/rtl872x/rtl_sdk_support.cpp
@@ -34,7 +34,6 @@ extern "C" {
 #include "interrupts_hal.h"
 #include "osdep_service.h"
 #include "concurrent_hal.h"
-#include "system_task.h"
 
 extern "C" {
 
@@ -223,14 +222,6 @@ extern "C" void HAL_Core_System_Reset(void) {
 }
 
 extern "C" {
-void _freertos_mfree(uint8_t *pbuf, u32 sz) {
-    if (__get_BASEPRI() != 0) {
-        // Defer freeing of memory to the SystemISRTaskQueue to be processed outside of this critical section
-        SPARK_ASSERT(system_isr_task_queue_free_memory((void*)pbuf) == 0);
-    } else {
-        free(pbuf);
-    }
-}
 
 #if 0 // Enable to get btgap logs
 
--- a/hal/src/rtl872x/rtl_osdep.cpp
+++ b/hal/src/rtl872x/rtl_osdep.cpp
@@ -15,6 +15,19 @@
  * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
 
+extern "C" {
+#include "rtl8721d.h"
+}
+#include "system_task.h"
+
+extern "C" void _freertos_mfree(u8 *pbuf, u32 sz) {
+    if (__get_BASEPRI() != 0) {
+        // Defer freeing of memory to the SystemISRTaskQueue to be processed outside of this critical section
+        SPARK_ASSERT(system_isr_task_queue_free_memory((void*)pbuf) == 0);
+    } else {
+        free(pbuf);
+    }
+}
 // namespace osdep {
 
 // u8* malloc(u32 sz) {


system/src/system_task.cpp
```diff
@@ -467,6 +467,26 @@ void manage_ble_prov_mode() {
...
 
+extern "C" void system_isr_task_queue_free_memory(uint8_t *ptrToFree) {
+    struct FreeTask : ISRTaskQueue::Task {
+        uint8_t *arg;
+    };
+
+    auto task = static_cast<FreeTask*>(system_pool_alloc(sizeof(FreeTask), nullptr));
+    if (!task)
+    {
+        return;
```
scott-brust, line 478: This will result in a memory leak, but I think we have no other recourse here?

system/src/system_task.cpp
```diff
@@ -467,6 +467,26 @@ void manage_ble_prov_mode() {
 #endif
 }
 
+extern "C" void system_isr_task_queue_free_memory(uint8_t *ptrToFree) {
+    struct FreeTask : ISRTaskQueue::Task {
+        uint8_t *arg;
```
scott-brust, line 472: Could this be simpler if we added some kind of argument variable or `void *` context pointer to the base `ISRTaskQueue::Task` instead? I think some other use cases could take advantage of this

system/src/system_task.cpp
```diff
@@ -467,6 +467,26 @@ void manage_ble_prov_mode() {
...
+    auto task = static_cast<FreeTask*>(system_pool_alloc(sizeof(FreeTask), nullptr));
+    if (!task)
+    {
+        return;
+    }
+    task->arg = ptrToFree;
+    task->func = [](ISRTaskQueue::Task * task) {
+       auto freeTask = reinterpret_cast<FreeTask*>(task);
+       free(freeTask->arg);
+       system_pool_free(task, nullptr);
```
scott-brust, line 484: Does the order of freeing matter here?

system/src/system_task.cpp
```diff
@@ -467,6 +467,26 @@ void manage_ble_prov_mode() {
 #endif
 }
 
+extern "C" void system_isr_task_queue_free_memory(uint8_t *ptrToFree) {
```
avtolstoy, line 470: This is going to be unused on all the other platforms except for rtl872x-based which is going to cause the compiler to complain that this is unused. Perhaps move into rtl872x-specific source files under HAL? Or perhaps let's do a proper thing and add this function into the system_task.h header, implement it here, disable unused warning for it.

avtolstoy, line 470: `void*` argument, not `uint8_t*`

system/src/system_task.cpp
avtolstoy, line 472: `void*`

avtolstoy: Potentially, but I'd leave this as-is for now, as all the other usages of ISR task queue already do the same thing anyway.

system/src/system_task.cpp
avtolstoy: Well, of course. If you free `task` first, you'll access garbage in `freeTask->arg`.

system/src/system_task.cpp
avtolstoy: Let's also not use this directly in the RTL SDK, perhaps. On second thought, let's mark `_freertos_mfree` as weak instead, revert all the changes, and reimplement it with system_isr_task_queue_free_memory()` when interrupts are disabled inside `rtl_sdk_support.cpp`.

system/src/system_task.cpp
avtolstoy: Make this return `int` and return `SYSTEM_ERROR_NO_MEMORY`. Handle it in `_freertos_mfree()` or more precisely its strong override in rtl_sdk_support.cpp and SPARK_ASSERT.

system/src/system_task.cpp
scott-brust: Ok, just sanity checking 👍

hal/src/rtl872x/rtl_sdk_support.cpp
```diff
@@ -221,6 +222,15 @@ extern "C" void HAL_Core_System_Reset(void) {
     }
 }
 
+extern "C" void _freertos_mfree(uint8_t *pbuf, u32 sz) {
```
avtolstoy, line 225: Nit: or can just move below, there is a whole extern "C" block there already :)

hal/src/rtl872x/rtl_sdk_support.cpp
```diff
@@ -221,6 +222,15 @@ extern "C" void HAL_Core_System_Reset(void) {
...
+    if (__get_BASEPRI() != 0) {
+        // Defer freeing of memory to the SystemISRTaskQueue to be processed outside of this critical section
+        SPARK_ASSERT(system_isr_task_queue_free_memory((void*)pbuf) == 0);
+    } else {
+        vPortFree(pbuf);
```
avtolstoy, line 230: Use the regular `free()`, we don't normally go to heap implementation directly, only through libc malloc/free/realloc/calloc.

avtolstoy (approved):

PR #2600, opened 2022-12-09
[rtl872x] fixes System.ticks()

technobly:
### Problem

- System.ticks() returns 0 currently on  P2 and TrackerM platforms.
- System.ticksDelay() also doesn't work.
- System.ticksPerMicrosecond() is wrong.

### Solution

- Ideally we'd have a hardware timer to use to hook up to System.ticks() like the rest of Gen 3 devices do, but the only one with a fast enough clock speed is currently being used for the RTC implementation and only has a 16-bit register anyway.
- Instead of just disabling System.ticks() we have implemented a stop gap of tying it to the micros() counter when the system does not support a system tick hardware counter.
- System.ticks() and micros() are essentially the same value on rtl872x based platforms now.
- System.ticksDelay() works like delayMicroseconds()
- System.ticksPerMicrosecond() returns 1

### Steps to Test

- Run TEST=wiring/no_fixture SYSTEM_08/SYSTEM_09

---

### Completeness

- [ ] Added documentation


platform/MCU/newhal-mcu/inc/hw_ticks.h
```diff
@@ -28,10 +28,10 @@ extern "C" {
  * The number of ticks per microsecond of the system counter.
  * SYSTEM_TICK_COUNTER
  */
-#define SYSTEM_US_TICKS		100     // cycles per microsecond
+#define SYSTEM_US_TICKS         (1)     // cycles per microsecond
```
scott-brust, line 31: Is there any specific reason to change this for `newhal-mcu`?

wiring/inc/spark_wiring_system.h
```diff
@@ -434,20 +434,28 @@ class SystemClass {
...
     static inline uint32_t ticksPerMicrosecond() {
         return SYSTEM_US_TICKS;
     }
 
     static inline uint32_t ticks() {
         return SYSTEM_TICK_COUNTER;
     }
+#else
+    static inline uint32_t ticksPerMicrosecond() {
+        return 1;
```
scott-brust, line 447: Shouldn't this still be `SYSTEM_US_TICKS`?

platform/MCU/rtl872x/inc/hw_ticks.h
```diff
@@ -23,8 +23,8 @@
 #include "platform_config.h"
 #include "system_tick_hal.h"
 
-#define SYSTEM_US_TICKS         (SystemCoreClock / 1000000) //cycles per microsecond
```
scott-brust, line 26: My initial reaction is that `SystemCoreClock` is different on KM0 (ie prebootloader) vs KM4, but I dont think `SYSTEM_US_TICKS` is used in prebootloader, so I think this change is ok.

platform/MCU/newhal-mcu/inc/hw_ticks.h
technobly: PLATFORM`newhal` is a placeholder for new platforms.  For the most part the numbers associated with it should be some sane default values.  This is why I've changed it to 1, as a default value when a platform doesn't support a hardware based system tick counter.

wiring/inc/spark_wiring_system.h
technobly: I guess it does seem like this another magic number, but on the off chance we define a new platform without defining SYSTEM_US_TICKS properly, this will ensure the API ticksPerMicrosecond() is not affected.

platform/MCU/rtl872x/inc/hw_ticks.h
technobly: Right, SYSTEM_US_TICKS is really only used by the System.ticks* APIs and one test case SPIX, which could/should probably be converted to use micros() later in a separate PR.

user/tests/wiring/no_fixture/system.cpp
```diff
@@ -439,3 +439,51 @@ test(SYSTEM_08_hardware_info) {
     }
 #endif // HAL_PLATFORM_NCP
 }
+
+test(SYSTEM_09_system_ticks) {
+    Serial.printlnf("ticksPerMicrosecond: %lu %lu", System.ticksPerMicrosecond(), System.ticksPerMicrosecond() * 100);
```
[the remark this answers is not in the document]
technobly: Thanks, I meant to comment these back out though.  Will fix that.

avtolstoy (approved):

PR #2604, opened 2022-12-15
[rtl872x] Support IO wakeup sources through IO expander.

XuGuohui:
As the title describes.

### Example App

```c
void setup() {
   */
}

void loop() {

}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
+++ b/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
@@ -126,7 +126,6 @@ void onSleepRequestReceived(km0_km4_ipc_msg_t* msg, void* context) {
 
 void configureSleepWakeupSource(const hal_sleep_config_t* config) {
     sleepDuration = 0;
-    hal_pin_info_t* halPinMap = hal_pin_map();
 
 #if HAL_PLATFORM_IO_EXTENSION && HAL_PLATFORM_MCP23S17
     bool ioExpanderIntConfigured = false;
@@ -145,6 +144,7 @@ void configureSleepWakeupSource(const hal_sleep_config_t* config) {
             uint32_t rtlPin;
             InterruptMode mode;
 #if HAL_PLATFORM_IO_EXTENSION && HAL_PLATFORM_MCP23S17
+            const hal_pin_info_t* halPinMap = hal_pin_map();
             if (halPinMap[gpioWakeup->pin].type == HAL_PIN_TYPE_MCU)
 #endif
             {
--- a/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
+++ b/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
@@ -126,6 +126,11 @@ void onSleepRequestReceived(km0_km4_ipc_msg_t* msg, void* context) {
 
 void configureSleepWakeupSource(const hal_sleep_config_t* config) {
     sleepDuration = 0;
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -88,6 +88,55 @@ class SleepClass {
...
+                        }
+                    }
+                    gpIntEn[halPinMap[gpioWakeup->pin].gpio_port] |= (0x01 << halPinMap[gpioWakeup->pin].gpio_pin);
+                }
+            }
+            source = source->next;
+        }
+
+        if (config) {
+            // TODO: check the returned result
```
jmwilson, line 128: There's no return value from `Mcp23s17::writeRegister` other than `SYSTEM_ERROR_NONE` so this can be removed. There doesn't seem to be any meaningful error recovery anyway since this function is `void` return.

Same thing in the nRF copy of this code

hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -88,6 +88,55 @@ class SleepClass {
         return SYSTEM_ERROR_NONE;
     }
 
+#if HAL_PLATFORM_IO_EXTENSION && MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+#if HAL_PLATFORM_MCP23S17
+    void configGpioWakeupSourceExt(const hal_wakeup_source_base_t* wakeupSources, hal_sleep_mode_t sleepMode) {
+        hal_pin_info_t* halPinMap = hal_pin_map();
```
jmwilson, line 94: `const`?

hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -120,6 +169,16 @@ class SleepClass {
             }
         }
 
+        hal_interrupt_suspend();
+
+        /* We neeed to configure the IO expander interrupt before disabling SPI interface. */
+#if HAL_PLATFORM_IO_EXTENSION && MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+#if HAL_PLATFORM_MCP23S17
+        Mcp23s17::getInstance().interruptsSuspend();
```
jmwilson, line 177: danger! - no corresponding call to `Mcp23s17::interruptsRestore`

hal/src/rtl872x/sleep_hal.cpp
jmwilson: If this function can't be shared/de-deuped between the nRF version of it, add a comment above the function that these two versions are coupled

hal/src/rtl872x/sleep_hal.cpp
XuGuohui: Good catch!

jmwilson (approved):

bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
```diff
@@ -126,6 +126,11 @@ void onSleepRequestReceived(km0_km4_ipc_msg_t* msg, void* context) {
 
 void configureSleepWakeupSource(const hal_sleep_config_t* config) {
     sleepDuration = 0;
+    hal_pin_info_t* halPinMap = hal_pin_map();
```
avtolstoy, line 129: Unused on P2 build.

avtolstoy (approved):

mehmood-n (approved):

PR #2595, opened 2022-12-01
Feature: supports hardware watchdog

XuGuohui:
As the title describes

### Example App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);

Serial1LogHandler l(115200, LOG_LEVEL_ALL);

static retained uint32_t magick = 0;

void setup() {
    Log.info("Application started: %d, running: %d, reason: %d\r\n", magick, Watchdog.started(), System.resetReason());

    Watchdog.init(WatchdogConfiguration().timeout(20s));
    Watchdog.onExpired([]() {
        magick++;
    });
    Watchdog.start();
}

void loop() {
    delay(3s);
    // Watchdog.refresh();
    Log.info("running: %d", Watchdog.started());
}
```

### References

Links to the Community, Docs, Other Issues, etc..

---

### Completeness

- [x] User is totes amazing for contributing!
- [x] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [ ] Problem and Solution clearly stated
- [ ] Run unit/integration/application tests on device
- [ ] Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

PR #2606, opened 2022-12-19
fix overflow error in HAL_Delay_Microseconds

jmwilson:
### Problem

`HAL_Delay_Microseconds` on the P2 has this main loop:

```
    system_tick_t start_micros = HAL_Timer_Get_Micro_Seconds();
    system_tick_t current_micros = start_micros;
    system_tick_t end_micros = start_micros + uSec;

    ...

    while (current_micros < end_micros) {
        current_micros = HAL_Timer_Get_Micro_Seconds();
    }
```

If `end_micros` is close to `UINT_MAX` then it is possible that updating `current_micros` can wrap the counter and reset it to a low value, extending the wait by > 71 minutes. The closer `end_micros` is to precisely `UINT_MAX`, the less likely the loop will ever terminate.

### Solution

Check if we've waited long enough by using unsigned difference instead of comparing to `end_micros`. Watchdog isn't implemented yet but is expected later and makes this mirror the nRF code.

### Steps to Test

`wiring/no_fixture` runs without errors.
Run publish in Tracker Edge with sleep enabled to test the startup delay call.

### References

https://app.shortcut.com/particle/story/113600/overflow-in-p2-hal-delay-microseconds-causes-hangs-in-tracker-edge

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/delay_hal.cpp
+++ b/hal/src/rtl872x/delay_hal.cpp
@@ -34,21 +34,9 @@ void HAL_Delay_Milliseconds(uint32_t nTime)
 
 void HAL_Delay_Microseconds(uint32_t uSec)
 {
-    system_tick_t start_micros = HAL_Timer_Get_Micro_Seconds();
-    system_tick_t current_micros = start_micros;
-    system_tick_t end_micros = start_micros + uSec;
+    const system_tick_t start_micros = HAL_Timer_Get_Micro_Seconds();
 
-    // Handle case where micros rolls over
-    if (end_micros < start_micros) {
-        while(1) {
-            current_micros = HAL_Timer_Get_Micro_Seconds();
-            if (current_micros < start_micros) {
-                break;
-            }
-        }
-    }
-
-    while (current_micros < end_micros) {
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/delay_hal.cpp
```diff
@@ -34,21 +34,9 @@ void HAL_Delay_Milliseconds(uint32_t nTime)
...
-            if (current_micros < start_micros) {
-                break;
-            }
-        }
-    }
-
-    while (current_micros < end_micros) {
-        current_micros = HAL_Timer_Get_Micro_Seconds();
+    while (HAL_Timer_Get_Micro_Seconds() - start_micros < uSec) {
+        HAL_Notify_WDT();
```
avtolstoy, line 40: As of #2595 watchdog is not going to be kicked from Device OS, completely under application control. This needs to be removed.

avtolstoy (changes requested):

avtolstoy (approved):

XuGuohui (approved):

PR #2608, opened 2023-01-03
implements shared interrupt pin for Trackerm.

XuGuohui:
As the title describes.

### Example App

```c
void handler1(void* data) {
    LOG(TRACE, "handler1: %08X", (uint32_t)data);
}

void handler2(void* data) {
    LOG(TRACE, "handler2: %08X", (uint32_t)data);
}

void handler3(void* data) {
    LOG(TRACE, "handler3: %08X", (uint32_t)data);
}

void test() {
    hal_gpio_mode(D0, INPUT_PULLUP);
    hal_interrupt_extra_configuration_t config = {};
    config.appendHandler = true;
    config.priority = 1;
    hal_interrupt_attach(D0, handler1, (void*)0x12345678, FALLING, &config);
    config.priority = 3;
    hal_interrupt_attach(D0, handler2, (void*)0xDEADBEEF, FALLING, &config);
    config.priority = 2;
    hal_interrupt_attach(D0, handler3, (void*)0x55AA55AA, FALLING, &config);

    // hal_interrupt_detach_ext(D0, false, (void*)handler3);
}
```

---

### Completeness

- [ ] Added documentation


scott-brust:
@XuGuohui Are you testing your example code on a non form factor trackerm?
I am trying to test this on a V0.0.3 form factor board, and I think the only two IO I have easily available are `D10` IE `P2_D10_IOEX_INT` and `A1`. I cant get the handlers to execute when I short either of those pins to ground. Not sure what im doing wrong

hal/inc/interrupts_hal.h
```diff
@@ -83,6 +84,10 @@ typedef struct hal_interrupt_extra_configuration_t {
       uint8_t keepHandler  : 1;
     };
   };
+  struct {
+      bool appendHandler;
```
avtolstoy, line 88: Perhaps this can be added as a flag into the above flags/union? The unused bits are 0 right now in every call.

hal/inc/interrupts_hal.h
```diff
@@ -83,6 +84,10 @@ typedef struct hal_interrupt_extra_configuration_t {
...
+      uint8_t priority;
```
avtolstoy, line 89: Let's give a more verbose name e.g. `chainPriority` or something like so that we are not mixing this up with IRQ priority etc.

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -48,12 +53,87 @@ typedef enum interrupt_state_t {
...
+    void executeHandlers() const {
+        for (auto& callback : callbacks_) {
+            if (callback.handler) {
+                callback.handler(callback.data);
+            }
+        }
+    }
+
+public:
+    interrupt_state_t           state;
```
avtolstoy, line 118: Nit: formatting, no indents between type/name

system/src/system_power_manager.cpp
```diff
@@ -375,8 +375,16 @@ void PowerManager::loop(void* arg) {
     // IMPORTANT: attach the interrupt handler first
 #if HAL_PLATFORM_PMIC_INT_PIN_PRESENT
     hal_gpio_mode(PMIC_INT, INPUT_PULLUP);
+#if PLATFORM_ID == PLATFORM_TRACKERM
```
avtolstoy, line 378: `HAL_PLATFORM_` feature flag.

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -48,12 +53,87 @@ typedef enum interrupt_state_t {
     INT_STATE_SUSPENDED
 } interrupt_state_t;
 
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+
+class InterruptConfig {
+public:
+    struct InterruptCallback {
+        bool operator==(const InterruptCallback& callback) {
```
avtolstoy, line 61: When adding `operator==` make sure to add `operator!=` as well

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -48,12 +53,87 @@ typedef enum interrupt_state_t {
...
+    };
+
+    InterruptConfig()
+            : state(INT_STATE_DISABLED),
+              callbacks_() {
+    }
+    ~InterruptConfig() = default;
+
+    int appendHandler(hal_interrupt_handler_t handler, void* data, uint32_t priority) {
+        uint32_t idx = 0;
```
avtolstoy, line 77: Nit: for things like these and needing unsigned type prefer `unsigned` or `size_t`

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -48,12 +53,87 @@ typedef enum interrupt_state_t {
...
+            }
+        }
+    }
+
+public:
+    interrupt_state_t           state;
+    InterruptMode               mode;
+
+private:
+    Vector<InterruptCallback>   callbacks_;
```
avtolstoy, line 122: There is no thread safety in interrupts hal and with existing implementation it all works out okay, but adding a Vector here is going to cause thread/interrupt-safety issues when modifying handlers/trying to execute them.

So, we either have to add some form of interrupt/thread-safety e.g. create a new modified vector of handlers and swap it with existing one when modifying while disabling interrupts.

avtolstoy (changes requested):

hal/src/rtl872x/interrupts_hal.cpp
XuGuohui: Can we simply disable interrupt while modifying the verctor?

avtolstoy: No, we unfortunately cannot, as modifying a vector involves heap allocations and those cannot be done under disabled interrupts. You could use an `IntrusiveList` perhaps and perform allocations outside of the list pointer modifications as an option if we are to step away from a vector here.

avtolstoy: You can manually change `next` pointer though. `AtomicSimpleStaticPool` is an overkill for this honestly, better create a whole new vector and swap atomically in that case then.

user/tests/wiring/shared_interrupt/shared_interrupt.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    hal_gpio_mode(trigPin, OUTPUT);
+    hal_gpio_write(trigPin, HIGH);
+
+    hal_gpio_mode(intPin, INPUT_PULLUP);
+    hal_interrupt_extra_configuration_t config = {};
+    config.flags = HAL_INTERRUPT_FLAG_APPEND_HANDLER;
+
+    config.chainPriority = 1;
+    hal_interrupt_attach(intPin, handler, (void*)0x12345678, mode, &config);
```
scott-brust, line 55: Multiple handlers can be attached to a pin via `hal_interrupt_attach()` in system code, but users cannot currently append multiple handlers using `attachInterrupt()`.
Is this intentional? Do we only want to have this functionality available on the system side?

scott-brust (commented):
I was able to test this with a P2 and could see the three handlers hit in priority order using the debugger. So it seems to work as intended in basic use cases.

user/tests/wiring/shared_interrupt/shared_interrupt.cpp
avtolstoy: We are not exposing this for general use in wiring as of right now, correct.

technobly (commented):
Tested this on TrackerM and charging appears to work fine, and also with the 100mV threshold for recharging.

One thing I noticed though was that it doesn't seem to indicate the charging source properly.

BSoM running 5.2.0 shows USB HOST
```
0000060001 [app] TRACE: Current PMIC settings:
0000060007 [app] TRACE: VIN Vmin: 3880
0000060009 [app] TRACE: VIN Imax: 500
0000060013 [app] TRACE: Ichg: 896
0000060018 [app] TRACE: Vterm: 4112
0000060022 [app] TRACE: Power source: usb host
0000060023 [app] TRACE: Battery state: charging
```

TrackerM running this PR shows UNKNOWN
```
0031446006 [app] TRACE: Current PMIC settings:
0031446020 [app] TRACE: VIN Vmin: 3880
0031446035 [app] TRACE: VIN Imax: 500
0031446050 [app] TRACE: Ichg: 896
0031446061 [app] TRACE: Vterm: 4112
0031446077 [app] TRACE: Power source: unknown
0031446094 [app] TRACE: Battery state: charging
```

For comparison, here's the BSoM also running this branch, which seems to work:
```
0001190000 [app] TRACE: Current PMIC settings:
0001190006 [app] TRACE: VIN Vmin: 3880
0001190011 [app] TRACE: VIN Imax: 500
0001190013 [app] TRACE: Ichg: 896
0001190017 [app] TRACE: Vterm: 4112
0001190020 [app] TRACE: Power source: usb host
0001190022 [app] TRACE: Battery state: charged
```

hal/inc/interrupts_hal.h
```diff
@@ -57,6 +57,10 @@ typedef enum {
   HAL_INTERRUPT_DIRECT_FLAG_ENABLE  = 0x04
 } hal_interrupt_direct_flags_t;
 
+typedef enum hal_interrupt_flags_t {
+    HAL_INTERRUPT_FLAG_APPEND_HANDLER = 0x01,
+} hal_interrupt_flags_t;
```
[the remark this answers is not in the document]
scott-brust: We discussed in team meeting today to add another bit in order to keep backwards compatibility with `keepHandler` which is still used

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -212,9 +317,16 @@ int hal_interrupt_attach(uint16_t pin, hal_interrupt_handler_t handler, void* da
         GPIO_UserRegIrq(rtlPin, (VOID*)gpioIntHandler, (void*)((uint32_t)pin));
         GPIO_INTMode_HAL(rtlPin, pinInfo->gpio_port, ENABLE, GPIO_InitStruct.GPIO_ITTrigger, GPIO_InitStruct.GPIO_ITPolarity, GPIO_INT_DEBOUNCE_ENABLE);
 
-        interruptsConfig[pin].state = INT_STATE_ENABLED;
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+        if (!config || (config && !(config->appendHandler))) {
+            interruptsConfig[pin].clearHandlers();
+        }
+        interruptsConfig[pin].appendHandler(handler, data, config->chainPriority);
```
avtolstoy, line 324: Ideally `config->chainPriority` can only be accessed if there is such a field, which as of right now indicated by `config->version == HAL_INTERRUPT_EXTRA_CONFIGURATION_VERSION_3`. Presence of `config->appendHandler` does not necessarily mean that we are not going to access out of bounds.

avtolstoy (approved):

PR #2609, opened 2023-01-05
A workaround to fix the AMD USB issue

YutingYou:
### Problem

We found the USB request `Get Line Coding` couldn't get a response on the AMD Windows PC, the symptom could be that we can't open the com port on a serial tool software on the AMD Windows PC.

### Solution

This is a workaround, not an ideal solution, create this PR for further discussion.

A straightforward fix is to add a bit of delay after `usbd_ep0_transmit()`, then the response of `Get Line Coding` can be sent. The transmit operation seems to be blocked. 

The FAE said `usbd_ep0_transmit()` only enables `NPTXFEM` interrupt, the real transfer is finished in the USB thread `usbd_isr_...` for which we configured its priority to `RTL_USBD_ISR_PRIORITY=7`.

But the callstack shows `usbd_ep0_transmit()` is called in `setupCb()` which has already been running in the USB thread.
Screen Shot 2023-01-04 at 13 58 01

### Steps to Test

1. Checkout this branch, compile firmware for P2
2. Install a serial tool software and try to open/close the com port
Screen Shot 2023-01-05 at 17 58 57

### References

[CH109872]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/usbd_driver.cpp
+++ b/hal/src/rtl872x/usbd_driver.cpp
@@ -263,8 +263,10 @@ int RtlUsbDriver::setupReply(SetupRequest* r, const uint8_t* data, size_t size) 
             // some common cases of returning not a lot of data generated on stack
             memcpy(tempBuffer_, data, size);
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, tempBuffer_, size));
+            HAL_Delay_Milliseconds(2);
         } else {
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, (uint8_t*)data, size));
+            HAL_Delay_Milliseconds(2);
         }
     } else if (data == nullptr && size == 0) {
         // FIXME: doesn't work
--- a/hal/src/rtl872x/usbd_driver.cpp
+++ b/hal/src/rtl872x/usbd_driver.cpp
@@ -263,10 +263,14 @@ int RtlUsbDriver::setupReply(SetupRequest* r, const uint8_t* data, size_t size) 
             // some common cases of returning not a lot of data generated on stack
             memcpy(tempBuffer_, data, size);
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, tempBuffer_, size));
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
             HAL_Delay_Milliseconds(2);
+#endif // MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
         } else {
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, (uint8_t*)data, size));
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
             HAL_Delay_Milliseconds(2);
+#endif // MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
         }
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/usbd_driver.cpp
```diff
@@ -263,8 +263,10 @@ int RtlUsbDriver::setupReply(SetupRequest* r, const uint8_t* data, size_t size)
             // some common cases of returning not a lot of data generated on stack
             memcpy(tempBuffer_, data, size);
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, tempBuffer_, size));
+            HAL_Delay_Milliseconds(2);
```
scott-brust, line 266: The build is currently failing because the bootloader doesnt include `HAL_Delay_Milliseconds()`. If we are ok without this change in the bootloader (I think we are?) I think we need to wrap this with 
```
#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
...
#endif
```

YutingYou: Yeah, `HAL_Delay_Milliseconds()` is based on FreeRTOS, actually `HAL_Delay_Microseconds()` also works for me, but let's use your method to keep the bootloader as-is.

hal/src/rtl872x/usbd_driver.cpp
```diff
@@ -263,8 +263,14 @@ int RtlUsbDriver::setupReply(SetupRequest* r, const uint8_t* data, size_t size)
             // some common cases of returning not a lot of data generated on stack
             memcpy(tempBuffer_, data, size);
             CHECK_RTL_USB_TO_SYSTEM(usbd_ep0_transmit(rtlDev_, tempBuffer_, size));
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
+            HAL_Delay_Milliseconds(2);
```
avtolstoy, line 267: @YutingYou I think this just hides the problem and we can still run into it. Can we look into why the delay here is required? I would assume only two possible reasons:
1. Thread priority issues
2. Previous `usbd_ep0_transmit()` needs to complete which probably means we have to wait for `ep0DataInCb()` and queue things up

YutingYou: @avtolstoy I tried several things before

1. ❌ improve the USB thread priority to the highest
2. ✅ HAL_Delay_Microseconds(2000)
3. ❌ Disable the scheduler
```
os_thread_scheduling(false nullptr);
HAL_Delay_Milliseconds(2);
```

So it might not be a multi-thread issue, I also confirmed `usbd_ep0_transmit ()` and `ep0DataInCb()` are running in the USB thread.

We're calling `usbd_ep0_transmit()` inside `ep0DataInCb()`, I'm not sure if this causes issues. In their CDC example, they don't event `use ep0_data_in` (we register this callback with `ep0DataInCb()`).
```
/* CDC ACM Class Driver */
usbd_class_driver_t usbd_cdc_driver = {
	.get_descriptor = cdc_acm_get_descriptor,
	.set_config = cdc_acm_set_config,
	.clear_config = cdc_acm_clear_config,
	.setup = cdc_acm_setup,
	.ep0_data_out = cdc_acm_handle_ep0_data_out,
	.ep_data_in = cdc_acm_handle_ep_data_in,
	.ep_data_out = cdc_acm_handle_ep_data_out,
};
```

YutingYou:
Fixed in PR2625, close this PR.

PR #2610, opened 2023-01-10
[photon2] Update FQC test with photon 2 pinout

scott-brust:
### Problem

Photon 2 devices are going into production. The feather pinout of the Photon 2 differs from the bare P2 module. The FQC test will need to be updated to test the feather pins

### Solution

Update the GPIO pin mapping to reflect which pins are available and adjacent on the feather pins

### Steps to Test

Run FQC tests with photon 2 (NOTE: Current pre DVT photon 2 boards do not have the efuse model variant settings programed, so they will not be detected as Photon 2s. In order to test FQC, you will have to make local modifications to tinker to use the photon 2 pins, then jumper those pins together). 

### Example App
Use Tinker

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/deviceid_hal_impl.h
+++ b/hal/src/rtl872x/deviceid_hal_impl.h
@@ -46,6 +46,8 @@
 #define WIFI_OUID_SIZE                  3
 #define DEVICE_ID_PREFIX_SIZE           6
 
+#define MODEL_VARIANT_PHOTON_2 1
+
 #ifdef	__cplusplus
 extern "C" {
 #endif
--- a/hal/src/rtl872x/deviceid_hal_impl.h
+++ b/hal/src/rtl872x/deviceid_hal_impl.h
@@ -46,8 +46,6 @@
 #define WIFI_OUID_SIZE                  3
 #define DEVICE_ID_PREFIX_SIZE           6
 
-#define MODEL_VARIANT_PHOTON_2 1
-
 #ifdef	__cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/deviceid_hal_impl.h
```diff
@@ -46,6 +46,8 @@
 #define WIFI_OUID_SIZE                  3
 #define DEVICE_ID_PREFIX_SIZE           6
 
+#define MODEL_VARIANT_PHOTON_2 1
```
scott-brust, line 49: Is this the correct spot for this model variant definition to be defined? Are we ok starting with `1` and not `0`?

user/applications/tinker/src/fqc_test.cpp
```diff
@@ -258,15 +258,18 @@ bool FqcTest::bleScan(JSONValue req) {
...
 // The pin order largely reflects which pins are physically next to each other on the P2 jig and can be easily jumpered together. 
-// Two sets of pins are not adjacent and require a hookup wire:
-// - Pin19 (PA0) + Pin10 (PA15 ie WKP/D10)
-// - Pin14 (PB4) + Pin21 (PB31)
-static const Vector<uint16_t> p2_digital_pins = {14, 21, 10, 19, 0, 1, 15, 16, 17, 12, 18, 2, 4, 5, 3, 11, 13, 20 };
+// Two sets of pins on the P2 jig are not adjacent and require a hookup wire:
+// - S4 (PA0) + WKP/D10 (PA15)
+// - A5 (PB4) + S6 (PB31)
+static Vector<uint16_t> p2_gpio_test_pins = { A5, S6, WKP, S4, D0, D1, S0, S1, S2, A1, S3, D2, D4, D5, D3, A0, A2, S5 };
+static Vector<uint16_t> photon2_gpio_test_pins = {A0, A1, A2, A5, S4, S3, S2, S0, S1, D0, D1, D2, D3, D4, D5, D10 };
```
scott-brust, line 267: These pin pairs are a best guess based on the adjacent feather pins. The real test jig may route pins differently and require changes to the pairing. But this is a start for now

avtolstoy (approved):

PR #2611, opened 2023-01-12
[test] mailbox support and support for resets within tests

avtolstoy:
### Problem

Currently there is no sane way of writing automated tests using our test framework that:
1. Generate messages during the test that the JS side can act upon
2. Notify device-os-test-runner about on-device tests triggering a reset/detach

### Solution

Add a device->host mailbox interface to send either specific notifications (e.g. device is going to reset) or generic data e.g. statistics from SLO tests.

NOTE: requires https://github.com/particle-iot/device-os-test-runner/pull/21

### Steps to Test

- runner/mailbox
- slo/connect_time
- wiring/no_fixture: system reset test

### Example App

See runner/mailbox test.

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/integration/application/src/request_handler.cpp
+++ b/user/tests/integration/application/src/request_handler.cpp
@@ -171,6 +171,8 @@ int RequestHandler::request(Request* req) {
         return getStatus(req);
     } else if (cmd == "L") { // Get log
         return getLog(req);
+    } else if (cmd == "M") { // Read device->host mailbox
+        return readMailbox(req);
     } else if (cmd == "r") { // Reset
         return reset(req);
     } else if (req->isEmpty()) { // Ping request
@@ -254,6 +256,38 @@ int RequestHandler::getLog(Request* req) {
         const auto runner = TestRunner::instance();
         w.value(runner->logBuffer(), runner->logSize());
     }));
+    return 0;
+}
+
+int RequestHandler::readMailbox(Request* req) {
+    const auto runner = TestRunner::instance();
+    auto m = runner->peekOutboundMailbox();
+    if (!m) {
+        return 0;
+    }
+
+    CHECK(req->reply([&](JSONWriter& w) {
+        w.beginObject();
+        w.name("t").value((int)m->type());
+        const auto data = m->data();
+        if (data.size() > 0) {
+            w.name("d").value(data.data(), data.size());
+        }
+        w.endObject();
+    }));
+    if (m->waitedOn()) {
+        req->done(0, [](int, void* data) -> void {
+            auto m = static_cast<TestRunner::MailboxEntry*>(data);
+            if (m->waitedOn()) {
+                m->completed();
+            } else {
+                const auto runner = TestRunner::instance();
+                runner->popOutboundMailbox();
+            }
+        }, m);
+    } else {
+        CHECK(runner->popOutboundMailbox());
+    }
     return 0;
 }
 
--- a/user/tests/integration/application/src/request_handler.h
+++ b/user/tests/integration/application/src/request_handler.h
@@ -47,6 +47,7 @@ private:
     int getStatus(Request* req);
     int getLog(Request* req);
     int reset(Request* req);
+    int readMailbox(Request* req);
 };
 
 } // namespace particle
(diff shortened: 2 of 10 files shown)


user/tests/integration/application/src/request_handler.cpp
```diff
@@ -257,6 +259,38 @@ int RequestHandler::getLog(Request* req) {
...
+        w.beginObject();
+        w.name("t").value((int)m->type());
+        const auto data = m->data();
+        if (data.size() > 0) {
+            w.name("d").value(data.data(), data.size());
+        }
+        w.endObject();
+    }));
+    if (m->waitedOn()) {
+        req->done(0, [](int, void* data) -> void {
```
scott-brust, line 279: Why do we explicitly call `req->done()` here rather than let it be called in `RequestHandler::request(ctrl_request* ctrlReq)` like all the other request messages?

scott-brust:
This is my understanding of how the mailbox is supposed to work, please correct where I am wrong:

On device tests can put messages to the host into a mailbox in TestRunner. Right now the messages are one way (device --> host) and can only contain log messages or notifications that the device is wanting to reset. 

Messages can have a `wait` time associated with them. The reset message uses this specifically in order to allow the test runner to detect the pending reset and start polling for the device response
When a message with a wait time (ie a reset message) is enqueued into the mailbox, it will call the `MailboxEntry::wait()` function which will `delay(1)` until it is "completed".
Meanwhile, the test runner host will read from the mailbox and see that the message is in the waiting loop via `waitedOn()`. It will signal to the wait() loop that the message has been received and is "completed". The device can continue execution, and reset itself at that point.

From the host side, the Test Runner runs the tests as usual
While the test runs, Test Runner sends USB request `{ c : 'M'}` to check for potential mailbox messages
Device checks mailbox for messages to respond with. Messages are stored in an IntrusiveQueue `outMailbox_`
If there is a message in the queue, it will reply with a response `{ t : 1 }` for reset pending and `{ t : 2 , 'd' : 'string_data'}` for general "log" messages

scott-brust:
Just FYI, I ran this and it worked as expected:
```
Test directory: /Users/scottbrust/develop/device-os/user/tests/integration
Device OS directory: /Users/scottbrust/develop/device-os
Firmware binaries directory: /Users/scottbrust/develop/device-os/build/integration
Generating test matrix
Initializing API client
URL: https://api.particle.io
Authenticating with access token
Signed in as device-os-ci@particle.io
Subscribing to device events
Initializing device manager
Enumerating USB devices
Retrieving devices from the cloud
Device pool:
0a10aced202194944a001418 (scott-trackerm-proto3ff-1418)
0a10aced202194944a02c5f8
No fixtures configured
Running tests


  Test runner mailbox
    p2
    Target device: 0a10aced202194944a02c5f8
    Flashing application: mailbox.bin
    Getting device tests
      systemThread=enabled
        Initializing device test suite
        Running device test: 01_mailbox
mailbox msg { t: 2, d: 'test' }
mailbox msg { t: 2, d: 'deadbeef' }
mailbox msg { t: 2, d: '\u0000\u0001\u0002\u0003' }
        Device test passed
        ✓ 01_mailbox (7.2s)
        Running device test: 02_mailbox_reset
mailbox msg { t: 1 }
        Device test notified about expected reset
        Device test passed
        ✓ 02_mailbox_reset (4.8s)
        Running device test: 03_mailbox
mailbox msg { t: 2, d: 'test' }
mailbox msg { t: 2, d: 'deadbeef' }
mailbox msg { t: 2, d: '\u0000\u0001\u0002\u0003' }
        Device test passed
        ✓ 03_mailbox (1.1s)
        Running device test: 04_mailbox_reset_postponed
mailbox msg { t: 1 }
        Device test notified about expected reset
        Device test passed
        ✓ 04_mailbox_reset_postponed (9.9s)
        Running device test: 05_mailbox
mailbox msg { t: 2, d: 'test' }
mailbox msg { t: 2, d: 'deadbeef' }
mailbox msg { t: 2, d: '\u0000\u0001\u0002\u0003' }
        Device test passed
        ✓ 05_mailbox (1.6s)
    Resetting device


  5 passing (38s)

Waiting for the debugger to disconnect...
```

user/tests/integration/application/src/request_handler.cpp
avtolstoy: It's not ideal, but just to keep the logic contained within `readMailbox()` and pass our own completion handler to `system_ctrl_set_result()`. If the mailbox entry is being actively waited on, we only want to unblock the waiter once the host has received it.

The second req->done() call in `request()` will just be no-op.

avtolstoy:
One key thing to mention is that when the device generates a reset request to the test runner and doesn't actually perform the reset that will trigger an error within the test runner, so that case should also be covered here.

scott-brust (approved):

PR #2612, opened 2023-01-18
[rtl872x] enable factory reset feature

avtolstoy:
### Problem

Factory reset through bootloader and long MODE button press until blinking WHITE is only available as of right now on nRF52840 platforms.

### Solution

Fix that, enable on P2 and Tracker M as well.

List of things that are reset:
1. Wifi configuration
2. Cellular configuration (SIM external/internal/APN etc)
3. Everything in DCT except for device keys
4. Power management settings

### TODO

1. ~Reset power management configuration?~ In DCT, already reset.
2. Reset system cache (NCP fw version, NCP MAC, ADC calibration offset)?
3. Reset OTA transfer state?

### Steps to Test

1. Flash system part out of this branch
5. Hold RESET and MODE, release RESET, continue holding until flashing WHITE (after MAGENTA and YELLOW)
6. Release MODE
7. Device should boot and lose its WiFi configuration, feature flags etc.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/src/main.cpp
+++ b/system/src/main.cpp
@@ -648,7 +648,6 @@ int resetSettingsToFactoryDefaultsIfNeeded() {
     }
     SYSTEM_FLAG(NVMEM_SPARK_Reset_SysFlag) = 0x0000;
     Save_SystemFlags();
-#if HAL_PLATFORM_NRF52840
     LOG(WARN, "Resetting all settings to factory defaults");
 #if HAL_PLATFORM_WIFI
     // Clear WiFi credentials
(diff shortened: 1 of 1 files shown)


mrlambchop (approved):

scott-brust:
```
TODO
Reset power management configuration?
Reset system cache (NCP fw version, NCP MAC, ADC calibration offset)?
Reset OTA transfer state?
```
If we reset the device and clear DCT during a factory reset, how are these not reset as well? Are we storing state somewhere besides RAM/DCT? (retained ram?)

avtolstoy:
@scott-brust Separate files in fs.

scott-brust:
> @scott-brust Separate files in fs.

Ok, interesting. I had not come across system cache yet. It seems very similar to DCT
The PMIC settings are stored in DCT, am I missing other power settings stored elsewhere?

avtolstoy:
Oh, that's true. Power management is still in DCT. Good catch, so that's a non-issue and a good thing that they are already reset. Edited.

PR #2615, opened 2023-01-31
[nRF52] BLE plus RTC sleep causes hardfault

XuGuohui:
### Problem
When `.ble()` and `.duration(...)` are specified as wakeup sources, device will crash on entering sleep mode.

### Solution
Wait until RTC is ticking before `__disable_irq()`

### Steps to Test

Build and flash the below app to Argon(Boron)

### Example App

```c
SYSTEM_MODE(MANUAL);

void setup() {
    delay(5s);

    BLE.advertise();
    delay(1s);

    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::ULTRA_LOW_POWER).ble().duration(10s));
}
```

### References
https://community.particle.io/t/device-wake-up-via-ble-and-duration-getting-sos-hard-fault/63466

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/sleep_hal.cpp
+++ b/hal/src/nRF52840/sleep_hal.cpp
@@ -463,13 +463,6 @@ static void configRtcWakeupSource(const hal_wakeup_source_base_t* wakeupSources)
             // Reconfigure RTC2 for wake-up
             NVIC_ClearPendingIRQ(RTC2_IRQn);
 
-            nrf_rtc_event_clear(NRF_RTC2, NRF_RTC_EVENT_TICK);
-            nrf_rtc_event_enable(NRF_RTC2, RTC_EVTEN_TICK_Msk);
-            // Make sure that RTC is ticking
-            // See 'TASK and EVENT jitter/delay'
-            // http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52840.ps%2Frtc.html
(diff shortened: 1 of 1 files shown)


hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -973,6 +966,17 @@ static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_sour
...
     __ISB();
 
+    // WARNING: This is going to fix the issue that when both .ble() and .duration()
+    // are specified as wakeup sources, device will run into hardfault if the following
+    // check comes after __disable_irq(), such as in configRtcWakeupSource() previously.
+    nrf_rtc_event_enable(NRF_RTC2, RTC_EVTEN_TICK_Msk);
+    // Make sure that RTC is ticking
+    // See 'TASK and EVENT jitter/delay'
+    // http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52840.ps%2Frtc.html
+    while (!nrf_rtc_event_pending(NRF_RTC2, NRF_RTC_EVENT_TICK));
```
avtolstoy, line 976: I guess this one takes a long time to complete?

XuGuohui: yup. I replaced the `while()` loop with `HAL_Delay_Microseconds(1000000)` and it crashed as well. If I shorten the delay, it worked as expected. So it's weird that why the `NRF_RTC_EVENT_TICK` event is not asserted after `__disable_irq()` when softdevice is enabled.

PR #2616, opened 2023-02-02
Ethernet GPIO config

technobly:
### Problem

Devices that support the WizNet Ethernet controller have 3 hard coded GPIO pins in system firmware that collide with SPI1 usage.  If a user wants to use Ethernet which requires SPI and D3, D4 and D5 to be reserved on Argon/Boron/P2/Photon2/TrackerM for example.  If they also need to use hardware SPI1 for some other peripheral, they cannot.

### Solution

- Make Ethernet GPIO pins configurable for all platforms Argon/Boron/BSoM/B5SoM/ESoMX/P2/Tracker/TrackerM.
- The API is not official and will change in a future release, but once set the device shall remember the configuration and the API need not be called on every boot.
- The system must be reset after setting the pin configuration, it will not work in STARTUP without a reset so it's best to use it as in the example below
- `System.enableFeature(FEATURE_ETHERNET_DETECTION)` must be used, and if Ethernet is detected, `Ethernet.isOn()` will return `true`.  If the system boots with FEATURE_ETHERNET_DETECTION disabled (default), the Ethernet API's will not work and `Ethernet.isOn()` will return `false`, which can be used to set the correct pin configuration, enabled FEATURE_ETHERNET_DETECTION and reset the system.  Upon successive boots the Ethernet controller will be found and pin config and reset will not need to be called again.  NOTE: ensure the device does not get caught up in a reset loop with a `retained` variable, or some other NVM method.  Another important note is when following this sequence of DISABLED, pin config change, ENABLE, the Ethernet GPIO pins will not inadvertently be toggled upon first boot.  Namely these are the CS and RESET OUTPUT pins, whereas the INT pin is an INPUT with external 10k pull-up.
- Pins defined as PIN_INVALID will use their default value.

### Steps to Test

- Run `TEST=app/ethernet` with an Ethernet Wing and Argon/Boron/Photon2.
- Test moving all 3 pins
- Test moving only 1 or 2 pins while keeping the others set to PIN_INVALID
- Test cold booting the device in Safe Mode
- Test OTA updates on the device

## How to change the pins:

Set the pins to whatever available GPIO is required.  All can be explicitly set, or if you would like one or more to be in their default location you may use `PIN_INVALID`. Enable the Ethernet detection feature, and reset the system.  Avoid calling this again on successive boots by using the Ethernet.isOn() API. (See example snippet and full example below)

```
remap.cs_pin = D2;
remap.reset_pin = D0;
remap.int_pin = D1;
System.reset();
```

### Example App

```c
#include "Particle.h"
SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);
Serial1LogHandler log1Handler(115200, LOG_LEVEL_ALL);

retained uint8_t resetRetry = 0;

void setup() {
#if HAL_PLATFORM_WIFI && !HAL_PLATFORM_WIFI_SCAN_ONLY
    // To force Ethernet only, clear Wi-Fi credentials
    Log.info("Clear Wi-Fi credentionals...");
    WiFi.clearCredentials();
#endif // HAL_PLATFORM_WIFI && !HAL_PLATFORM_WIFI_SCAN_ONLY

    // Disable Listening Mode if not required
    if (System.featureEnabled(FEATURE_DISABLE_LISTENING_MODE)) {
        Log.info("FEATURE_DISABLE_LISTENING_MODE enabled");
    } else {
        Log.info("Disabling Listening Mode...");
    }

    Log.info("Checking if Ethernet is on...");
    if (Ethernet.isOn()) {
        Log.info("Ethernet is on");
        uint8_t macAddrBuf[8] = {};
        uint8_t* macAddr = Ethernet.macAddress(macAddrBuf);
        if (macAddr != nullptr) {
            Log.info("Ethernet MAC: %02x %02x %02x %02x %02x %02x",
                    macAddr[0], macAddr[1], macAddr[2], macAddr[3], macAddr[4], macAddr[5]);
        }
        Ethernet.connect();
        waitFor(Ethernet.ready, 30000);
        Log.info("Ethernet.ready: %d", Ethernet.ready());
        resetRetry = 0;
    } else if (++resetRetry <= 3) {
        Log.info("Ethernet is off or not detected, attmpting to remap pins: %d/3", resetRetry);

        remap.cs_pin = PIN_INVALID; // default
        remap.reset_pin = PIN_INVALID; // default
        remap.int_pin = PIN_INVALID; // default

        // remap.cs_pin = D5;
        // remap.reset_pin = D3;
        // remap.int_pin = D4;

        // remap.cs_pin = D2;
        // remap.reset_pin = D0;
        // remap.int_pin = D1;

        // remap.cs_pin = D10;
        // remap.reset_pin = D6;
        // remap.int_pin = D7;

        // remap.cs_pin = A0;
        // remap.reset_pin = A2;
        // remap.int_pin = A1;

        // remap.cs_pin = 50; // bad pin config
        // remap.reset_pin = 100;
        // remap.int_pin = 150;

        auto ret = if_request(nullptr, IF_REQ_DRIVER_SPECIFIC, &remap, sizeof(remap), nullptr);
        if (ret != SYSTEM_ERROR_NONE) {
            Log.error("Ethernet GPIO config error: %d", ret);
        } else {
            if (System.featureEnabled(FEATURE_ETHERNET_DETECTION)) {
                Log.info("FEATURE_ETHERNET_DETECTION enabled");
            } else {
                Log.info("Enabling Ethernet...");
            }
            delay(500);
            System.reset();
        }
    }

    Particle.connect();
}

void loop() {
    static system_tick_t lastPublish = millis();
    static int count = 0;
    static bool reconnect = false;

    if (Particle.connected()) {
        reconnect = false;
        if (millis() - lastPublish >= 10000UL) {
            Particle.publish("mytest", String(++count), PRIVATE, WITH_ACK);
            lastPublish = millis();
        }
    }

    // Detect a network dropout and reconnect quickly
    if (!reconnect && !Ethernet.ready()) {
        Log.info("Particle disconnect...");
        Particle.disconnect();
        waitFor(Particle.disconnected, 5000);
        Log.info("Network disconnect...");
        Network.disconnect();
        Particle.connect();
        reconnect = true;
    }
}
```

### References

- sc-105146

---

### Completeness

- [x] Added documentation


hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+LOG_SOURCE_CATEGORY("net.en.cfg")
+
+#include "system_error.h"
+
+#include "check.h"
+#include "delay_hal.h"
+
+#if HAL_PLATFORM_NCP
+#include "system_cache.h"
+#endif // HAL_PLATFORM_NCP
```
scott-brust, line 30: Whats the point of this check? I think the only platform that doesnt have `HAL_PLATFORM_NCP` is the gcc platform, which I dont think should be building the wiznet stuff?

hal/network/lwip/ifapi.cpp
```diff
@@ -1153,11 +1155,38 @@ int if_event_handler_del(if_event_handler_cookie_t cookie) {
...
+            if_wiznet_pin_remap* wzpr = (if_wiznet_pin_remap*)req;
+            if (wzpr->base.type == IF_WIZNET_DRIVER_SPECIFIC_PIN_REMAP) {
+                // LOG(INFO, "IF_REQUEST cs_pin: %u, reset_pin: %u, int_pin: %u",
+                //         wzpr->cs_pin, wzpr->reset_pin, wzpr->int_pin);
+                WizNetifConfigData wizNetifConfigData;
+                wizNetifConfigData.size = sizeof(WizNetifConfigData);
+                wizNetifConfigData.cs_pin = wzpr->cs_pin;
+                wizNetifConfigData.reset_pin = wzpr->reset_pin;
+                wizNetifConfigData.int_pin = wzpr->int_pin;
+                WizNetifConfig::instance()->setConfigData(&wizNetifConfigData);
```
scott-brust, line 1186: If we are not going to implement pin remapping at runtime (which I think is fine), then effectively all we need to do is store this structure in system cache. (and then when we bring up the ethernet interface, check if there are cached pins present)
So is using `if_request()` and `IF_REQ_DRIVER_SPECIFIC` message type just a way to leverage an already user available function/api?

hal/network/api/ifapi.h
```diff
@@ -247,9 +247,25 @@ typedef struct if_event_power_state if_req_power;
...
     IF_REQ_NONE        = 0,
     IF_REQ_POWER_STATE = 1,
-    IF_REQ_DHCP_STATE  = 2
+    IF_REQ_DHCP_STATE  = 2,
+    IF_REQ_DRIVER_SPECIFIC = 3,
 } if_req_t;
 
+typedef struct if_req_driver_specific {
+    uint32_t type;
+} if_req_driver_specific;
```
scott-brust, line 256: Why do we need this struct? Cant we just add a `type` field to `if_wiznet_pin_remap` below?

hal/src/tron/network/network.cpp
```diff
@@ -111,7 +112,13 @@ int if_init_platform(void*) {
...
 
     if (HAL_Feature_Get(FEATURE_ETHERNET_DETECTION)) {
-        en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
+        WizNetifConfigData wizNetifConfigData;
+        WizNetifConfig::instance()->getConfigData(&wizNetifConfigData);
+        hal_pin_t cs_pin = (wizNetifConfigData.cs_pin == PIN_INVALID) ? D5 : wizNetifConfigData.cs_pin;
+        hal_pin_t reset_pin = (wizNetifConfigData.reset_pin == PIN_INVALID) ? D3 : wizNetifConfigData.reset_pin;
+        hal_pin_t int_pin = (wizNetifConfigData.int_pin == PIN_INVALID) ? D4 : wizNetifConfigData.int_pin;
+        // Default: cs_pin = D5, reset_pin = D3, int_pin = D4
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, cs_pin, reset_pin, int_pin, mac);
```
scott-brust, line 121: I think the conclusion was that we would not allow switching to `HAL_SPI_INTERFACE2`, but if we did, is it as simple as storing that to the cache and using it here like the other pins?

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: I borrowed it from R510 FOTA stuff, which has Unit Tests... which needed that.  It probably doesn't hurt to leave it?

hal/network/lwip/ifapi.cpp
technobly: Yep, that's how it works.  `if_request()` was an available API that already has a DYNALIB entry, since we can't access the WizNetifConfig::instance() by itself.

hal/network/api/ifapi.h
technobly: It's so you can dereference the pointer and check the `type`, before you dereference the pointer as the unique struct.  I didn't end up using it the way it was originally intended, but if we ever add more unique types it will be handy to convert to using the `if_req_driver_specific` struct to check the type first, since `type` will always be the first 4 bytes.

hal/src/tron/network/network.cpp
technobly: Yeah, that should work.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+int WizNetifConfig::saveWizNetifConfigData() {
+    WizNetifConfigData tempData = {};
+    int result = SystemCache::instance().get(SystemCacheKey::WIZNET_CONFIG_DATA, (uint8_t*)&tempData, sizeof(tempData));
+    if (result != sizeof(tempData) ||
+            /* memcmp(tempData, wizNetifConfigData_, sizeof(tempData) != 0)) { // not reliable */
+            tempData.size != wizNetifConfigData_.size ||
+            tempData.cs_pin != wizNetifConfigData_.cs_pin ||
+            tempData.reset_pin != wizNetifConfigData_.reset_pin ||
+            tempData.int_pin != wizNetifConfigData_.int_pin ) {
```
XuGuohui, line 67: Use `memcmp()` instead?

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+//             configData.cs_pin,
+//             configData.reset_pin,
+//             configData.int_pin);
+// }
+
+int WizNetifConfig::saveWizNetifConfigData() {
+    WizNetifConfigData tempData = {};
+    int result = SystemCache::instance().get(SystemCacheKey::WIZNET_CONFIG_DATA, (uint8_t*)&tempData, sizeof(tempData));
+    if (result != sizeof(tempData) ||
+            /* memcmp(tempData, wizNetifConfigData_, sizeof(tempData) != 0)) { // not reliable */
```
XuGuohui, line 63: Could you elaborate it?

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+int WizNetifConfig::recallWizNetifConfigData() {
+    WizNetifConfigData tempData;
+    int result = SystemCache::instance().get(SystemCacheKey::WIZNET_CONFIG_DATA, (uint8_t*)&tempData, sizeof(tempData));
+    if (result != sizeof(tempData)) {
+        return SYSTEM_ERROR_NOT_FOUND;
+    }
+    // LOG(INFO, "Recalling cached wizNetifConfigData");
+    // logWizNetifConfigData(tempData);
+    memcpy(&wizNetifConfigData_, &tempData, sizeof(wizNetifConfigData_));
```
XuGuohui, line 83: Might need to check if the read data is valid or not. Pins should be in valid range.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+        recallWizNetifConfigData();
+        memcpy(configData, &wizNetifConfigData_, sizeof(WizNetifConfigData));
+        // logWizNetifConfigData(wizNetifConfigData_);
+    }
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int WizNetifConfig::validateWizNetifConfigData() {
+    if (wizNetifConfigData_.size != sizeof(WizNetifConfigData)) {
```
XuGuohui, line 132: We need to make sure all pins  are in valid range  as well.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+WizNetifConfig::WizNetifConfig() :
+        initialized_(false) {
+}
+
+void WizNetifConfig::init() {
+    memset(&wizNetifConfigData_, 0, sizeof(wizNetifConfigData_));
+    recallWizNetifConfigData();
+    if (validateWizNetifConfigData() != SYSTEM_ERROR_NONE) {
+        recallWizNetifConfigData();
```
XuGuohui, line 46: We've already saved the `wizNetifConfigData_` if the data is invalid, which has synced the cached, so we don't need to recall the data, plus that we're not checking the returned value actually.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+        init();
+    } else {
+        validateWizNetifConfigData();
+    }
+
+    if (configData) {
+        recallWizNetifConfigData();
+        memcpy(configData, &wizNetifConfigData_, sizeof(WizNetifConfigData));
+        // logWizNetifConfigData(wizNetifConfigData_);
+    }
```
XuGuohui, line 126: else return an error.

hal/network/lwip/wiznet/wiznetif_config.h
```diff
@@ -0,0 +1,98 @@
...
+#include "static_recursive_mutex.h"
+
+#ifdef __cplusplus
+
+namespace particle {
+
+namespace net {
+
+struct __attribute__((packed)) WizNetifConfigData {
+    uint16_t size;
```
XuGuohui, line 29: Prefer to add `version` field as we normally do.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    if (!initialized_) {
+        init();
+    } else {
+        validateWizNetifConfigData();
+    }
+
+    if (configData) {
+        recallWizNetifConfigData();
+        memcpy(configData, &wizNetifConfigData_, sizeof(WizNetifConfigData));
```
XuGuohui, line 124: We'd better initialize the `size` field of `configData` before we call this method and use `memcpy(configData, &wizNetifConfigData_, std::min(configData->size, wizNetifConfigData_->size));` for backwards compatibility.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+        validateWizNetifConfigData();
+    }
+
+    if (configData) {
+        memcpy(&wizNetifConfigData_, configData, sizeof(WizNetifConfigData));
+        // logWizNetifConfigData(wizNetifConfigData_);
+        saveWizNetifConfigData();
+    } else {
+        return SYSTEM_ERROR_BAD_DATA;
```
XuGuohui, line 107: Prefer `SYSTEM_ERROR_INVALID_ARGUMENT`

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+    WizNetifConfigLock lk;
+
+    if (!initialized_) {
+        init();
+    } else {
+        validateWizNetifConfigData();
+    }
+
+    if (configData) {
+        memcpy(&wizNetifConfigData_, configData, sizeof(WizNetifConfigData));
```
XuGuohui, line 103: We might copy and save invalid `configData`, so need to check it before doing so.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+int WizNetifConfig::validateWizNetifConfigData() {
+    if (wizNetifConfigData_.size != sizeof(WizNetifConfigData)) {
+        memset(&wizNetifConfigData_, 0, sizeof(wizNetifConfigData_));
+        wizNetifConfigData_.size = sizeof(WizNetifConfigData);
+        wizNetifConfigData_.cs_pin = PIN_INVALID;
+        wizNetifConfigData_.reset_pin = PIN_INVALID;
+        wizNetifConfigData_.int_pin = PIN_INVALID;
+
+        saveWizNetifConfigData();
```
XuGuohui, line 139: I would like to seperate this method to one that accepts the config data to be checked and validates it, and the other to set default config data and sync it to cache.

hal/src/tron/network/network.cpp
```diff
@@ -111,7 +112,13 @@ int if_init_platform(void*) {
     CHECK(hal_get_mac_address(HAL_DEVICE_MAC_ETHERNET, mac, HAL_DEVICE_MAC_ADDR_SIZE, nullptr));
 
     if (HAL_Feature_Get(FEATURE_ETHERNET_DETECTION)) {
-        en2 = new WizNetif(HAL_SPI_INTERFACE1, D5, D3, D4, mac);
+        WizNetifConfigData wizNetifConfigData;
```
XuGuohui, line 115: Initialize the `size` field.

hal/src/tron/network/network.cpp
```diff
@@ -111,7 +112,13 @@ int if_init_platform(void*) {
...
+        WizNetifConfig::instance()->getConfigData(&wizNetifConfigData);
+        hal_pin_t cs_pin = (wizNetifConfigData.cs_pin == PIN_INVALID) ? D5 : wizNetifConfigData.cs_pin;
```
XuGuohui, line 117: Might be good to innitialize the config data with default pins, so that it's easy for us to change the default pins for all platforms, intead of traversing diffeerent folders to modify platform-specific `network.cpp`.

hal/network/lwip/ifapi.cpp
```diff
@@ -1153,11 +1155,38 @@ int if_event_handler_del(if_event_handler_cookie_t cookie) {
...
+            // ---
+            // auto dsreq = (if_req_driver_specific*)req;
+            // auto bnetif = getBaseNetif(iface);
+            // CHECK_TRUE(bnetif, -1);
+            // bnetif->request(dsreq, sizeof(dsreq));
+            if (reqsize != sizeof(if_wiznet_pin_remap)) {
+                return -1;
+            }
+
+            if_wiznet_pin_remap* wzpr = (if_wiznet_pin_remap*)req;
```
avtolstoy, line 1177: This is throwaway-ish stuff we all agree on that, but honestly not ideal. How about we just keep `WizNetif` instance if `FEATURE_ETHERNET_DETECTION` is enabled no matter if `WizNetif::initInterface()` detects presence of it or not (so getting rid of https://github.com/particle-iot/device-os/blob/develop/hal/src/tron/network/network.cpp#L120) ?

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: IIRC, for whatever reason it wasn't detecting changes in the structure.  This was on the FOTA PR though.  I left it here, but didn't try it again.  Maybe it would work on this simpler struct.

avtolstoy: There is a typo in there `, sizeof(tempData) != 0))` vs `, sizeof(tempData)) != 0)

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: That's a good point, do you think we should do a isPinValid() type check on the pins?  PIN_INVALID is also a valid option.  The size will be set correctly in `if_request()`

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: If we add a check for pin value, can we still bundle these two errors into BAD DATA or is INVALID ARGUMENT better, or separate errors for each error type?

hal/network/lwip/wiznet/wiznetif_config.cpp
XuGuohui: `isPinValid()` should be better.

hal/network/lwip/wiznet/wiznetif_config.cpp
XuGuohui: I'd prefer returning `SYSTEM_ERROR_NONE` and restore the default pin values.

hal/src/tron/network/network.cpp
technobly: This will only be for P2 for the time being.

technobly: As per our team discussion this morning we will release for all platforms that support Ethernet.

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: Hah, nice!  I'll convert to this method as it's simpler.

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: We should return SYSTEM_ERROR_INVALID_ARGUMENT if the user tries to input invalid data, but we can also restore defaults as well in that case.

hal/network/lwip/ifapi.cpp
technobly: This ended up not working out, but found a compromise for the user app that works pretty nice because it doesn't unintentionally toggle the Ethernet GPIO pins on first boot.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,206 @@
...
+// int WizNetifConfig::deleteWizNetifConfigData() {
+//     WZCFG_LOG_DEBUG(INFO, "Deleting cached wizNetifConfigData");
+//     int result = SystemCache::instance().del(SystemCacheKey::WIZNET_CONFIG_DATA);
+//     return (result < 0) ? result : SYSTEM_ERROR_NONE;
+// }
+
+int WizNetifConfig::setConfigData(const WizNetifConfigData* configData) {
+    WizNetifConfigLock lk;
+
+    if (!initialized_) {
```
avtolstoy, line 120: Nit: might want to move the check inside `init()` itself perhaps and just call it?

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -0,0 +1,206 @@
...
+        init();
+    }
+
+    CHECK_TRUE(configData, SYSTEM_ERROR_INVALID_ARGUMENT);
+
+    recallWizNetifConfigData();
+
+    if (configData->size == 0) {
```
avtolstoy, line 145: It's the caller responsibility to set `size` correctly, so ideally this should be an error condition.

hal/network/lwip/wiznet/wiznetif_config.h
```diff
@@ -0,0 +1,131 @@
...
+#include "platforms.h"
+
+#ifdef __cplusplus
+
+namespace particle {
+
+namespace net {
+
+#if PLATFORM_ID == PLATFORM_ARGON || PLATFORM_ID == PLATFORM_BORON || PLATFORM_ID == PLATFORM_P2 || PLATFORM_ID == PLATFORM_TRACKERM
+#define WIZNETIF_CS_PIN_DEFAULT    (D5)
```
avtolstoy, line 35: Might be a good idea to perhaps move these into `hal_platform_config.h`?

avtolstoy (approved):

hal/network/lwip/wiznet/wiznetif_config.h
technobly: I ended up moving them after our discussion this morning.

PR #2617, opened 2023-02-02
[test]: add more watchdog test cases.

XuGuohui:
As the title describes.

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/watchdog/watchdog.cpp
+++ b/user/tests/wiring/watchdog/watchdog.cpp
@@ -18,6 +18,20 @@
 #include "application.h"
 #include "unit-test/unit-test.h"
 
+static void startWatchdog(const WatchdogConfiguration& config) {
+    WatchdogInfo info;
+    assertEqual(0, Watchdog.init(config));
+    assertEqual(0, Watchdog.start());
+    assertEqual(0, Watchdog.getInfo(info));
+    assertTrue(info.state() == WatchdogState::STARTED);
+}
+
+static void checkState(WatchdogState state) {
+    WatchdogInfo info;
+    assertEqual(0, Watchdog.getInfo(info));
+    assertTrue(info.state() == state);
+}
+
 test(WATCHDOG_01_capabilities) {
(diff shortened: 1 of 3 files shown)


user/tests/wiring/watchdog/watchdog.cpp
```diff
@@ -55,8 +68,117 @@ test(WATCHDOG_02_default) {
...
+    assertEqual(0, pushMailbox(MailboxEntry().type(MailboxEntry::Type::RESET_PENDING), 5000));
+    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::STOP).duration(10s));
+}
+
+test(WATCHDOG_06_default_running_in_none_hibernate_mode_2) {
+    assertTrue(System.resetReason() == RESET_REASON_WATCHDOG);
+    checkState(WatchdogState::DISABLED);
+}
+
+static retained uint32_t magick = 0;
```
avtolstoy, line 127: Are retained variables kept both on Gen 3 and P2/Tracker M after WDG reset?

avtolstoy (approved):

user/tests/wiring/watchdog/watchdog.cpp
XuGuohui: Not on P2 based platforms

avtolstoy: There is `hal_backup_ram_sync()`, so in theory you can perform that before WDG reset and have backup ram restored on boot.

XuGuohui: We stilll need to figure out how we can promise the `retained` key word works as expected implicitly.

avtolstoy: debug peripheral watchpoint for that memory region on writes?

XuGuohui: I'll create a SC ticket for tracking this problem.

PR #2619, opened 2023-02-06
[wiring][gen3] Allow gen3 to select internal ADC reference source

scott-brust:
### Problem
Customer wishes to change ADC reference source from the default external VCC source to the internal voltage reference because it has greater accuracy on their particular design. 

### Solution

Add Wiring API to change the source at runtime. Setting is not persistent

### Steps to Test

I used a boron to test ADC reading on pin A5 with 2v from a bench supply. 

### Example App

```c
void setup() {
   */
}

void loop() {
    static uint8_t i = 0;
    int analogPin = A5;
    int value = 0;
    String source;

    // Set ADC reference source
    if (i++ % 2) {
        analogSetReference(AdcReference::INTERNAL);
    } else {
        analogSetReference(AdcReference::VCC);
    }

    source = (analogGetReference() == AdcReference::INTERNAL ? "INTERNAL" : "VCC     ");

    // Read ADC
    value = analogRead(analogPin);
    Log("ADC Source: %s ADC Value: %d", source.c_str(), value);
    
    delay(2000);
}
```
Sample output comparing the two sources:
```
0000444524 [app] INFO: ADC Source: VCC      ADC Value: 2574
0000446525 [app] INFO: ADC Source: internal ADC Value: 2594
0000448526 [app] INFO: ADC Source: VCC      ADC Value: 2569
0000450527 [app] INFO: ADC Source: internal ADC Value: 2592
0000452528 [app] INFO: ADC Source: VCC      ADC Value: 2582
0000454529 [app] INFO: ADC Source: internal ADC Value: 2586
0000456530 [app] INFO: ADC Source: VCC      ADC Value: 2583
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/adc_hal.cpp
+++ b/hal/src/rtl872x/adc_hal.cpp
@@ -298,6 +298,6 @@ int hal_adc_set_reference(uint32_t reference, void* reserved) {
     return SYSTEM_ERROR_NOT_SUPPORTED;
 }
 
-int hal_adc_get_reference(uint32_t reference, void* reserved) {
+int hal_adc_get_reference(void* reserved) {
     return (int)HAL_ADC_REFERENCE_DEFAULT;
 }
--- a/hal/src/rtl872x/adc_hal.cpp
+++ b/hal/src/rtl872x/adc_hal.cpp
@@ -293,3 +293,7 @@ int hal_adc_sleep(bool sleep, void* reserved) {
 int hal_adc_calibrate(uint32_t reserved, void* reserved1) {
     return Adc::instance().calibration();
 }
+
+int hal_adc_set_reference(uint32_t reference, void* reserved) {
+    return SYSTEM_ERROR_NOT_SUPPORTED;
+}
--- a/hal/src/rtl872x/adc_hal.cpp
+++ b/hal/src/rtl872x/adc_hal.cpp
@@ -297,3 +297,7 @@ int hal_adc_calibrate(uint32_t reserved, void* reserved1) {
 int hal_adc_set_reference(uint32_t reference, void* reserved) {
     return SYSTEM_ERROR_NOT_SUPPORTED;
 }
+
+int hal_adc_get_reference(uint32_t reference, void* reserved) {
+    return (int)HAL_ADC_REFERENCE_DEFAULT;
+}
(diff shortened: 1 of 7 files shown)


scott-brust:
Some questions that come to mind:
1. Should we provide a `get()` method to return the current ADC reference source? 
2. Should we even have a `AdcReference::DEFAULT` option? 
3. Or should we just have `AdcReference::INTERNAL` and/or rename `AdcReference::VCC` to `AdcReference::EXTERNAL` to be consistent?
4. We are allowing the user to override the OTP / HW info bits to default to internal ADC reference. I assume this is ok?
5. Should I add tests in `wiring/no_fixture/gpio`?

avtolstoy:
1. Given the fact that the default reference can be affected by hardware revision etc... Perhaps, yes.
2. I'd say yes, because again, it's not a simple answer of what is the default (internal or external or whatever else)
5. wiring/api + perhaps some manual test app?

XuGuohui (approved):

wiring_globals/src/spark_wiring_gpio.cpp
```diff
@@ -374,3 +374,7 @@ uint32_t pulseIn(hal_pin_t pin, uint16_t value) {
 void setDACBufferred(hal_pin_t pin, uint8_t state) {
   HAL_DAC_Enable_Buffer(pin, state);
 }
+
+int analogSetReference(AdcReference reference) {
+  return hal_adc_set_reference((uint32_t)reference, nullptr);
```
avtolstoy, line 379: Better use `to_underlying` from `enumclass.h`

scott-brust:
> 1. Given the fact that the default reference can be affected by hardware revision etc... Perhaps, yes.
> 2. I'd say yes, because again, it's not a simple answer of what is the default (internal or external or whatever else)
> 3. x
> 4. x
> 5. wiring/api + perhaps some manual test app?

Should have addressed these comments in latest commit, let me know what you think

hal/src/rtl872x/adc_hal.cpp
```diff
@@ -293,3 +293,11 @@ int hal_adc_sleep(bool sleep, void* reserved) {
 int hal_adc_calibrate(uint32_t reserved, void* reserved1) {
     return Adc::instance().calibration();
 }
+
+int hal_adc_set_reference(uint32_t reference, void* reserved) {
+    return SYSTEM_ERROR_NOT_SUPPORTED;
+}
+
+int hal_adc_get_reference(void* reserved) {
+    return (int)HAL_ADC_REFERENCE_DEFAULT;
```
avtolstoy, line 302: Probably better to specify which one is the default. I think for rtl872x that is probably internal.

avtolstoy (approved):

PR #2622, opened 2023-02-13
Ensure thread07 test executes as intended

scott-brust:
### Problem

Test `THREAD_07_particle_process_behavior_when_threading_enabled` fails in most cases with current `develop`. The problem is that when an Async task is inserted into the application thread queue, we only call `Particle.process()` once to execute the task. This works fine unless there are other work items in the queue ahead of that task. 

### Solution

For now, we will simply exhaust the queue by calling `Particle.process()` repeatedly. See discussion [here](https://s.slack.com/archives/CS0T32ZKK/p1676072611430139?thread_ts=1676047385.672259&cid=CS0T32ZKK)
Longer term we need to investigate if all queue items should be consumed when explicitly calling `Particle.process()` outside of the normal system idle loop.  

### Steps to Test

Run test runner test `THREAD_07_particle_process_behavior_when_threading_enabled`

### Example App

`THREAD_07_particle_process_behavior_when_threading_enabled`

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/wiring/no_fixture/thread.cpp
+++ b/user/tests/wiring/no_fixture/thread.cpp
@@ -205,10 +205,7 @@ test(THREAD_07_particle_process_behavior_when_threading_enabled)
     assertEqual((int)test_val_fn1, 0);
     // The increment task that was added to the Application Queue via invoke_async is not guaranteed to be the first and only item in the Application Thread Queue
     // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
-    bool processed = false;
-    do {
-        processed = Particle.process();
-    } while (processed);
+    while (Particle.process());
 
     assertEqual((int)test_val_fn1, 1);
     // Unblock system thread
--- a/user/tests/wiring/no_fixture/thread.cpp
+++ b/user/tests/wiring/no_fixture/thread.cpp
@@ -203,7 +203,11 @@ test(THREAD_07_particle_process_behavior_when_threading_enabled)
     assertTrue(Particle.connected());
 
     assertEqual((int)test_val_fn1, 0);
-    Particle.process();
+    for (int i = 0; i < 20; i++) {
+        // The increment task that was added to the Application Queue via invoke_async is not guarenteed to be the first and only item in the Application Thread Queue
+        // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
+        Particle.process();    
+    }
     assertEqual((int)test_val_fn1, 1);
     // Unblock system thread
     test_val = 0;
--- a/user/tests/wiring/no_fixture/thread.cpp
+++ b/user/tests/wiring/no_fixture/thread.cpp
@@ -203,11 +203,13 @@ test(THREAD_07_particle_process_behavior_when_threading_enabled)
     assertTrue(Particle.connected());
 
     assertEqual((int)test_val_fn1, 0);
-    for (int i = 0; i < 20; i++) {
-        // The increment task that was added to the Application Queue via invoke_async is not guarenteed to be the first and only item in the Application Thread Queue
-        // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
-        Particle.process();    
-    }
+    // The increment task that was added to the Application Queue via invoke_async is not guaranteed to be the first and only item in the Application Thread Queue
+    // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
+    bool processed = false;
+    do {
+        processed = Particle.process();
+    } while (processed);
+
     assertEqual((int)test_val_fn1, 1);
     // Unblock system thread
     test_val = 0;
--- a/system/inc/active_object.h
+++ b/system/inc/active_object.h
@@ -420,9 +420,9 @@ public:
         run();
     }
 
-    void process()
+    bool process()
     {
-        ActiveObjectQueue::process();
+        return ActiveObjectQueue::process();
     }
 };
 
--- a/system/inc/system_cloud.h
+++ b/system/inc/system_cloud.h
@@ -289,7 +289,7 @@ bool spark_sync_time_pending(void* reserved);
 system_tick_t spark_sync_time_last(time32_t* tm32, time_t* tm);
 
 
-void spark_process(void);
+bool spark_process(void);
 bool spark_cloud_flag_connected(void);
 
 /**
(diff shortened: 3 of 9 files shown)


technobly (approved):

user/tests/wiring/no_fixture/thread.cpp
```diff
@@ -203,7 +203,11 @@ test(THREAD_07_particle_process_behavior_when_threading_enabled)
     assertTrue(Particle.connected());
 
     assertEqual((int)test_val_fn1, 0);
-    Particle.process();
+    for (int i = 0; i < 20; i++) {
+        // The increment task that was added to the Application Queue via invoke_async is not guarenteed to be the first and only item in the Application Thread Queue
+        // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
+        Particle.process();
```
avtolstoy, line 209: We probably need to look at the other tests that use `Particle.process()` as well and add a similar thing in there.

How about we make this a bit more deterministic? We can extend `spark_process()` to return a bool/int to indicate whether there are any more events to process (void -> some type is fine as far as ABI goes) in the queue at least in the case of threaded mode and at least in the tests we can make a decision based on that.

scott-brust: I agree there are likely other tests that rely on this behavior that might fail in this scenario. I dont know if we want to find and change all of them right now.

If we want to improve this, we will have to change `Particle.process()` itself right? 
This means changing the `CloudClass::process` method to something like
```
    static bool process(void) {
            application_checkin();
            return spark_process();
    }
```
Or if we want it to process everything in the queue:
```
    static void process(void) {
            application_checkin();
            bool process = false;
            do {
                process = spark_process();
            } while (process);
    }
```
Or do this inside `spark_process()` anyways without having to change dynalib:
```
void spark_process(void)
{
    // application thread will pump application messages
#if PLATFORM_THREADING
    if (APPLICATION_THREAD_CURRENT()) {
        if (system_thread_get_state(NULL)) {
            bool process = false;
            do {
                process = ApplicationThread.process();
            } while (process);
        } else {
            Spark_Idle_Events(true);
        }
    }
#else
    // run the background processing loop, and specifically also pump cloud events
    Spark_Idle_Events(true);
#endif // PLATFORM_THREADING
}
```

I think any fix to this problem shouldnt involve having to loop over `Particle.process()` everywhere it is used. But this will change the system task behavior as well....

avtolstoy: What I'm proposing 'for now' is:
```diff
diff --git a/system/inc/system_dynalib_cloud.h b/system/inc/system_dynalib_cloud.h
index 107d824e8..cb3464975 100644
--- a/system/inc/system_dynalib_cloud.h
+++ b/system/inc/system_dynalib_cloud.h
@@ -37,7 +37,7 @@ DYNALIB_BEGIN(system_cloud)
 
 DYNALIB_FN(0, system_cloud, spark_variable, bool(const char*, const void*, Spark_Data_TypeDef, spark_variable_t*))
 DYNALIB_FN(1, system_cloud, spark_function, bool(const char*, p_user_function_int_str_t, void*))
-DYNALIB_FN(2, system_cloud, spark_process, void(void))
+DYNALIB_FN(2, system_cloud, spark_process, bool(void))
 DYNALIB_FN(3, system_cloud, spark_cloud_flag_connect, void(void))
 DYNALIB_FN(4, system_cloud, spark_cloud_flag_disconnect, void(void))
 DYNALIB_FN(5, system_cloud, spark_cloud_flag_connected, bool(void))
diff --git a/system/src/system_cloud.cpp b/system/src/system_cloud.cpp
index c7153339d..0e348dc8a 100644
--- a/system/src/system_cloud.cpp
+++ b/system/src/system_cloud.cpp
@@ -257,13 +257,13 @@ int spark_cloud_disconnect(const spark_cloud_disconnect_options* options, void*
     return 0;
 }
 
-void spark_process(void)
+bool spark_process(void)
 {
        // application thread will pump application messages
 #if PLATFORM_THREADING
     if (APPLICATION_THREAD_CURRENT()) {
         if (system_thread_get_state(NULL)) {
-            ApplicationThread.process();
+            return ApplicationThread.process();
         } else {
             Spark_Idle_Events(true);
         }
@@ -272,6 +272,7 @@ void spark_process(void)
     // run the background processing loop, and specifically also pump cloud events
     Spark_Idle_Events(true);
 #endif // PLATFORM_THREADING
+    return false;
 }
 
 String spark_deviceID(void)
diff --git a/wiring/inc/spark_wiring_cloud.h b/wiring/inc/spark_wiring_cloud.h
index f4587e8c2..12ef74437 100644
--- a/wiring/inc/spark_wiring_cloud.h
+++ b/wiring/inc/spark_wiring_cloud.h
@@ -369,9 +369,9 @@ public:
         spark_cloud_flag_connect();
     }
     static void disconnect(const CloudDisconnectOptions& options = CloudDisconnectOptions());
-    static void process(void) {
-               application_checkin();
-               spark_process();
+    static bool process(void) {
+        application_checkin();
+        return spark_process();
     }
     static String deviceID(void) { return SystemClass::deviceID(); }
 
```

That way we can for now 'fix' the tests to loop around Particle.process() until it stops saying `true` which means that the queue has been finally flushed, and in the future we can also potentially modify the behavior of `Particle.process()` specifically, or whatever else again based on the queue state.

scott-brust: ok, thats what I was thinking originally. ill add that and we can loop on `Particle.process()` in this test, which is better than just polling an arbitrary amount.

user/tests/wiring/no_fixture/thread.cpp
```diff
@@ -203,7 +203,13 @@ test(THREAD_07_particle_process_behavior_when_threading_enabled)
     assertTrue(Particle.connected());
 
     assertEqual((int)test_val_fn1, 0);
-    Particle.process();
+    // The increment task that was added to the Application Queue via invoke_async is not guaranteed to be the first and only item in the Application Thread Queue
+    // Call process() repeatedly in order to ensure that all potential messages in the queue are consumed, including the increment function
+    bool processed = false;
+    do {
+        processed = Particle.process();
```
avtolstoy, line 210: Nit: `while (Particle.process());` looks simpler? :)

And again, I'd probably add this to all the other tests where `Particle.process()` is used just in case, as I think the ones we outlined as having `Particle.process()` have been known to fail occasionally.

avtolstoy (approved):

user/tests/wiring/no_fixture/thread.cpp
scott-brust: Originally I was going to only execute a max of 20 times to avoid the possibility of an infinite loop, but I think the only way that can happen is if the async task being processed adds more item(s) back to the application queue. Ill make the change. 

Ill glance over the other tests that use `Particle.process()` and see if they need to be iterated on as well

technobly (approved):

PR #2623, opened 2023-02-15
rtl872x: Dcache fixes for exflash HAL

XuGuohui:
### Problem
The read data from external flash might be incorrect. This issue cannot be reliably reproduced, but once it encountered, the issue persists on every wakup from hibernate mode, and the RGB will give several blinks while it is restoring the server key and address.

### Solution
Clean the Dcache after reading data from external flash.

### Test App
```
#include "application.h"

SYSTEM_MODE(MANUAL);

void setup() {
    delay(5s);
    System.sleep(SystemSleepConfiguration().mode(SystemSleepMode::HIBERNATE).duration(3s));
}

void loop() {
}
```

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/exflash_hal.cpp
+++ b/hal/src/rtl872x/exflash_hal.cpp
@@ -59,11 +59,11 @@ typedef enum {
 #error "Unsupported external flash"
 #endif
 
-void dcacheInvalidateAligned(uintptr_t ptr, size_t size) {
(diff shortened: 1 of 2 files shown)


avtolstoy (approved):

avtolstoy:
@XuGuohui Should we perhaps on wake-up invalidate the whole cache just in case as well?

XuGuohui:
We have re-enabled Cache in `Set_System()` on every boot.

PR #2614, opened 2023-01-25
[trackerm] Fix ioexpander interrupt config to pass version check

scott-brust:
### Problem

Before merging the shared interrupts PR, I added a check to validate the interrupt config version. This had the unintended consequence of rejecting the MCP23 IO expander interrupt configuration and breaking its operation. 

### Solution

Add the right structure version

### Steps to Test

Run trackerM, if the modem powers on the IO expander is working

### Example App
tinker

### References

Links to the Community, Docs, Other Issues, etc..

---

### Completeness

- [x] User is totes amazing for contributing!
- [ ] Contributor has signed CLA ([Info here](https://github.com/spark/firmware/blob/develop/CONTRIBUTING.md))
- [ ] Problem and Solution clearly stated
- [ ] Run unit/integration/application tests on device
- [ ] Added documentation
- [ ] Added to CHANGELOG.md after merging (add links to docs and issues)

PR #2624, opened 2023-02-16
[rtl872x] hal: fix heap allocation issue in interrupt hal and postpone the mode button initialization.

avtolstoy:
### Problem

PR #2608 introduced a regression in MODE button handling for RTL872x-based platforms (P2, Tracker M).

### Solution

1. hal_button_init() is postponed until global constructors in system-part1 have been called and malloc has been enabled
2. interrupts_hal refactored to not remove entries unnecessarily (as our Vector implementation anyway does not free the memory) and instead safely invalidate them
3. Thread/interrupt-safety improvements for interrupts_hal
4. `wiring/shared_interrupt` test fix for configuration version to make it work after #2614

### Steps to Test

- `wiring/shared_interupt`
- MODE button presses should allow listening mode entry/exit as well as signal strength bars query
- Button should also still work correctly in bootloader

### Example App

N/A

### References

- [SC115613]
- #2608 
- #2614

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/interrupts_hal.cpp
+++ b/hal/src/rtl872x/interrupts_hal.cpp
@@ -116,7 +116,7 @@ public:
         decltype(callbacks_) temp;
         bool done = false;
         while (!done) {
-            temp.reserve(callbacks_.capacity() + 1);
+            CHECK_TRUE(temp.reserve(callbacks_.capacity() + 1), SYSTEM_ERROR_NO_MEMORY);
             ATOMIC_BLOCK() {
                 if (temp.capacity() == (callbacks_.capacity() + 1)) {
                     // No modifications occured, we can safely copy here
--- a/hal/src/rtl872x/interrupts_hal.cpp
+++ b/hal/src/rtl872x/interrupts_hal.cpp
@@ -43,6 +43,8 @@ using spark::Vector;
 using namespace particle;
 #endif
 
+#include <algorithm>
+
 extern uintptr_t link_ram_interrupt_vectors_location[];
 static uint32_t hal_interrupts_handler_backup[MAX_VECTOR_TABLE_NUM] = {};
 
@@ -67,9 +69,9 @@ public:
             return handler != callback.handler;
         }
 
-        hal_interrupt_handler_t handler;
-        void* data;
-        uint8_t chainPriority; // The lower the value, the higher the priority
+        hal_interrupt_handler_t handler = nullptr;
+        void* data = nullptr;
+        uint8_t chainPriority = 0; // The lower the value, the higher the priority
     };
 
     InterruptConfig()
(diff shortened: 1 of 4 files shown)


hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -78,48 +80,90 @@ class InterruptConfig {
...
-        CHECK_TRUE(temp.insert(idx, cb), SYSTEM_ERROR_NO_MEMORY);
-        ATOMIC_BLOCK() { 
-            spark::swap(temp, callbacks_);
+        if (!needAppend) {
+            return 0;
+        }
+
+        // Need to increase the size of the vector, which we cannot do from an ISR context
+        CHECK_FALSE(hal_interrupt_is_isr(), SYSTEM_ERROR_INVALID_STATE);
+        decltype(callbacks_) temp;
```
scott-brust, line 116: This is equivalent to `Vector<InterruptCallback> temp;` right?

hal/src/rtl872x/interrupts_hal.cpp
```diff
@@ -78,48 +80,90 @@ class InterruptConfig {
...
+        bool done = false;
+        while (!done) {
```
scott-brust, line 118: Is the intent of the loop here to retry `Vector::reserve` / `realloc()` if it fails to allocate a new larger vector? 
Can this get stuck if we can never allocate a larger vector? I guess that is an out of memory situation and we expect to fail in other ways?

scott-brust (approved):
Just a few questions

hal/src/rtl872x/interrupts_hal.cpp
avtolstoy: Aaand added

scott-brust: Ok so the check
```
if (temp.capacity() == (callbacks_.capacity() + 1)) {
```
is our best guess at detecting if `callbacks_` was modified after we reserved a new vector

avtolstoy: Correct. The modifications are either in-place in the existing buffer, or require a realloc, so capacitiy change is an indication that realloc happened, which is what we care about.

avtolstoy: Just to clarify realloc is always to a larger size in our vector implementation, to reduce capacity one has to manually call `trimToSize()`, this does not happen automatically. In addition to that this particular usage in interrupts_hal is also either in-place modification with InterruptCallback invalidation/replacement or append/realloc to a larger size.

scott-brust: I see, when a callback is removed, its handler is set to null, but it is left in the vector. 
When a callback is added, if we find a null handler, we know it is empty and can use that item in the vector.
If we dont find a null handler, then all callbacks are valid and we need to grow `callbacks_` vector. 

For 'removed' handlers in the vector, they will all have priority `0`, so sorting by priority wont really matter for them. 
When all handlers are invoked, they will be skipped as well since they have a null callback. So no issues there.

PR #2634, opened 2023-03-08
[rtl872x] hal: I2C may fail to read/write from/to slave device.

XuGuohui:
### Problem

1. RTL872x-based platforms acting as I2C master may failed to communicate with certain type of slave device.
2. RTL872x-based platforms acting as I2C slave may failed to communicate with I2C master.

### Solution

1. Re-enabled I2C before performing the I2C transmission.
2. Fix I2C slave race conditions.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL)

uint8_t i2cAddr = 0x3C; // SSD1306

void setup() {

    Wire.begin();

    uint8_t test[2] = {0x55, 0xaa};

    bool stop = true;
    for (uint8_t i = 0; i < 4; i++) {
        Wire.beginTransmission(i2cAddr);
        Wire.write(test, sizeof(test)); 
        int ret = Wire.endTransmission(stop);
        if (ret != 0) {
            Log.info("TX FAILED");
            while (1);
        }
        if (i < 2) {
            Wire.beginTransmission(i2cAddr);
            Wire.write(test, sizeof(test));
            int ret = Wire.endTransmission(true);
            if (ret != 0) {
                Log.info("TX -> TX FAILED, stop: %d", stop);
                while (1);
            }
        } else {
            int cnt = Wire.requestFrom(i2cAddr, sizeof(test), true);
            if (cnt != sizeof(test)) {
                Log.info("TX -> RX FAILED, stop: %d", stop);
                while (1);
            }
        }
        stop = !stop;
    }

    stop = true;
    for (uint8_t i = 0; i < 4; i++) {
        int cnt = Wire.requestFrom(i2cAddr, sizeof(test), stop);
        if (cnt != sizeof(test)) {
            Log.info("RX FAILED");
            while (1);
        }
        if (i < 2) {
            Wire.beginTransmission(i2cAddr);
            Wire.write(test, sizeof(test));
            int ret = Wire.endTransmission(true);
            if (ret != 0) {
                Log.info("RX -> TX FAILED");
                while (1);
            }
        } else {
            int cnt = Wire.requestFrom(i2cAddr, sizeof(test), true);
            if (cnt != sizeof(test)) {
                Log.info("RX -> RX FAILED");
                while (1);
            }
        }
        stop = !stop;
    }

    Log.info("Test done");

    delay(1s);
    System.reset();
}
 
void loop() {

}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/i2c_hal.cpp
+++ b/hal/src/rtl872x/i2c_hal.cpp
@@ -119,6 +119,7 @@ public:
             SPARK_ASSERT(txBuffer_.buffer_ && rxBuffer_.buffer_);
             heapBuffer_ = true;
         }
+        slaveRxCacheDepth_ = rxBuffer_.size() * 2;
         I2C_StructInit(&i2cInitStruct_);
         configured_ = true;
         unlock();
@@ -152,9 +153,9 @@ public:
             i2cInitStruct_.I2CAckAddr = 0; // This is the peer device's slave address in master mode, not needed herein.
         } else {
             if (hal_interrupt_is_isr()) {
-                return SYSTEM_ERROR_INVALID_STATE;
+                SPARK_ASSERT(false);
             }
-            slaveRxCache_ = std::make_unique<uint8_t[]>(SLAVE_RX_CACHE_DEPTH);
+            slaveRxCache_ = std::make_unique<uint8_t[]>(slaveRxCacheDepth_);
             if (!slaveRxCache_) {
                 SPARK_ASSERT(false);
             }
@@ -314,17 +315,9 @@ public:
         uint32_t quantity = std::min((size_t)config->quantity, rxBuffer_.size());
 
         // Dirty-hack: It may not generate the start signal when communicating with certain type of slave device.
-        if (!masterRestarted()) {
-            I2C_Cmd(i2cDev_, DISABLE);
-            if (!WAIT_TIMED(transConfig_.timeout_ms, I2C_CheckFlagState(i2cDev_, BIT_IC_STATUS_ACTIVITY) == 1)) {
-                reset();
-                LOG(TRACE, "SYSTEM_ERROR_I2C_BUS_BUSY");
-                return SYSTEM_ERROR_I2C_BUS_BUSY;
-            }
-            I2C_Cmd(i2cDev_, ENABLE);
+        if (reEnableIfNeeded() != SYSTEM_ERROR_NONE) {
+            return 0;
         }
-
-        clearIntStatus();
 
         setAddress(config->address);
 
@@ -345,7 +338,7 @@ public:
             // Wait for I2C_FLAG_RFNE flag
             if (WAIT_TIMED_ROUTINE(config->timeout_ms, I2C_CheckFlagState(i2cDev_, BIT_IC_STATUS_RFNE) == 0, checkAbrt) < 0) {
                 reset();
-                LOG(TRACE, "Wait BIT_IC_STATUS_RFNE timeout");
+                LOG_DEBUG(TRACE, "Wait BIT_IC_STATUS_RFNE timeout");
                 goto ret;
             }
             if(checkAbrt()) {
@@ -411,17 +404,7 @@ public:
         }
 
         // Dirty-hack: It may not generate the start signal when communicating with certain type of slave device.
-        if (!masterRestarted()) {
-            I2C_Cmd(i2cDev_, DISABLE);
-            if (!WAIT_TIMED(transConfig_.timeout_ms, I2C_CheckFlagState(i2cDev_, BIT_IC_STATUS_ACTIVITY) == 1)) {
-                reset();
-                LOG(TRACE, "SYSTEM_ERROR_I2C_BUS_BUSY");
-                return SYSTEM_ERROR_I2C_BUS_BUSY;
-            }
-            I2C_Cmd(i2cDev_, ENABLE);
-        }
-
-        clearIntStatus();
+        CHECK(reEnableIfNeeded());
 
         setAddress(transConfig_.address);
 
@@ -524,7 +507,7 @@ public:
     }
 
 private:
-    enum i2cSlaveStatus {
+    enum I2cSlaveStatus {
         I2C_SLAVE_STOPPED,
         I2C_SLAVE_STARTED,
         I2C_SLAVE_RESTARTED,
@@ -544,7 +527,8 @@ private:
               onReceived_(nullptr),
               mutex_(nullptr),
               slaveRxCache_(nullptr),
-              slaveRxCacheLen_(0) {
+              slaveRxCacheLen_(0),
+              slaveRxCacheDepth_(HAL_PLATFORM_I2C_BUFFER_SIZE(HAL_I2C_INTERFACE1) * 2) {
     }
     ~I2cClass() = default;
 
@@ -552,6 +536,20 @@ private:
         return (i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_START_DET) &&
                 !(i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_STOP_DET);
                 (i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_ACTIVITY);
+    }
+
+    int reEnableIfNeeded() {
+        if (!masterRestarted()) {
+            I2C_Cmd(i2cDev_, DISABLE);
+            if (!WAIT_TIMED(transConfig_.timeout_ms, I2C_CheckFlagState(i2cDev_, BIT_IC_STATUS_ACTIVITY) == 1)) {
+                reset();
+                LOG_DEBUG(TRACE, "SYSTEM_ERROR_I2C_BUS_BUSY");
+                return SYSTEM_ERROR_I2C_BUS_BUSY;
+            }
+            I2C_Cmd(i2cDev_, ENABLE);
+        }
+        clearIntStatus();
+        return SYSTEM_ERROR_NONE;
     }
 
     bool stopDetected() {
@@ -592,8 +590,8 @@ private:
                 slaveStatus_ = I2C_SLAVE_RESTARTED;
             } else if (intStatus & BIT_IC_INTR_STAT_R_START_DET) {
                 slaveStatus_ = I2C_SLAVE_STARTED;
-            } else {
                 // FIXME: It might be RESTARTED -> STOPPED
+            } else {
                 slaveStatus_ = I2C_SLAVE_STOPPED;
             }
             return true;
@@ -608,7 +606,7 @@ private:
         }
         rxBuffer_.reset();
         if (reset) {
-            memset(slaveRxCache_.get(), 0xFF, SLAVE_RX_CACHE_DEPTH);
+            memset(slaveRxCache_.get(), 0xFF, slaveRxCacheDepth_);
             slaveRxCacheLen_ = 0;
         } else {
             slaveRxCacheLen_ = slaveRxCacheLen_ - len;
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -379,6 +383,10 @@ class I2cClass {
             return SYSTEM_ERROR_INVALID_STATE;
         }
 
+        // Dirty-hack: It may not generate the start signal when communicating with certain type of slave device.
+        I2C_Cmd(i2cDev_, DISABLE);
```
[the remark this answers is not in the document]
XuGuohui: Good point! I'll test it out.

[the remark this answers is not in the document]
XuGuohui: We must not re-enable I2C when the next operation is initiated by the "re-start" signal. I have workrounded this in the latest commit.

XuGuohui:
@avtolstoy @scott-brust The PR is updated and ready for review again.

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -502,58 +578,140 @@ class I2cClass {
...
-            instance->rxBuffer_.reset();
-            instance->txBuffer_.reset();
+    // Return:
+    // 1: some of the read data belongs to next session
+    // 0: all of the read data belongs the current session
+    bool slaveRead() {
+        uint16_t reportLen = 0;
+        uint32_t intStatus;
+        while (I2C_CheckFlagState(i2cDev_, (BIT_IC_STATUS_RFNE | BIT_IC_STATUS_RFF))) {
+            slaveRxCache_[slaveRxCacheLen_++] = (uint8_t)I2C_ReceiveData(i2cDev_);
```
avtolstoy, line 626: @XuGuohui What if `SLAVE_RX_CACHE_DEPTH` is less than RX buffer and the transaction is also over `SLAVE_RX_CACHE_DEPTH` ? We can overflow here looks like.

Why not use the RX buffer itself and extend it over the configured size if required for peripheral behavior weirdness? E.g. x2?

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -150,14 +152,21 @@ class I2cClass {
...
+            }
             i2cInitStruct_.I2CMaster  = I2C_SLAVE_MODE;
             i2cInitStruct_.I2CAckAddr = address; // This is the local device's slave address in slave mode
         }
         I2C_Init(i2cDev_, &i2cInitStruct_);
 
         // Enable restart function
         i2cDev_->IC_CON |= BIT_CTRL_IC_CON_IC_RESTART_EN;
-        i2cDev_->IC_RX_TL = std::min(rxBuffer_.size(), (size_t)255); // Maximum RX FIFO depth: 255 + 1
+        i2cDev_->IC_RX_TL = std::min(rxBuffer_.size(), (size_t)0); // FIFO depth 1
```
scott-brust, line 169: Isnt this always going to be `0`? 

I think its ok to have a low (or even zero) RX threshold, but we should make sure that is the intent here. 
Note: the documentation for this register says the values can range from 0-255, "with the additional restriction that it may not be set to a value larger than the depth of the buffer". The "buffer" they are referring to is the 16 byte buffer implemented in hardware.

If we do change this to not be zero, it should not be more than 16.

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -474,19 +507,60 @@ class I2cClass {
...
+              slaveRxCache_(nullptr),
+              slaveRxCacheLen_(0),
+              slaveRxCacheDepth_(HAL_PLATFORM_I2C_BUFFER_SIZE(HAL_I2C_INTERFACE1) * 2) {
     }
     ~I2cClass() = default;
 
+    bool masterRestarted() {
+        return (i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_START_DET) &&
+                !(i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_STOP_DET);
+                (i2cDev_->IC_RAW_INTR_STAT & BIT_IC_INTR_STAT_R_ACTIVITY);
```
scott-brust, line 538: Does this line do anything? 
Is the `;` at the end of line 537 intentional? Should it be an extra condition ie `&&`?

scott-brust (commented):
I still need to finish reviewing the `i2cSlaveIntHandler` state machine, and run the example app with a real SSD1306 device (I have one on order), but here are some comments in the meantime. 

Also, for what its worth, I ran the `wiring/i2c_master_slave` tests with two photon 2s, and the master tests passed:
```
This is Master
Master message: I26HMFSESJNFYEMHQW3EFU26BNG
Slave message: PLYRPOKDWXF3FG7PHHTYUGHE2CRIVLN
Test I2C_01_Master_Slave_Master_Variable_Length_Transfer passed.
Test I2C_02_Master_Slave_Master_Variable_Length_Transfer_Slave_Tx_Buffer_Underflow passed.
Test I2C_03_Master_Slave_Master_WireTransmission_And_Short_Timeout passed.
Test I2C_04_Master_Slave_Master_Variable_Length_Transfer_With_WireTransmission_And_Standard_Timeout passed.
Test I2C_05_Master_Slave_Master_Variable_Length_Restarted_Transfer passed.
Test summary: 5 passed, 0 failed, and 0 skipped, out of 5 test(s).
```

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -150,14 +152,21 @@ class I2cClass {
             i2cInitStruct_.I2CMaster  = I2C_MASTER_MODE;
             i2cInitStruct_.I2CAckAddr = 0; // This is the peer device's slave address in master mode, not needed herein.
         } else {
+            if (hal_interrupt_is_isr()) {
+                SPARK_ASSERT(false);
```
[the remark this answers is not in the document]
XuGuohui: It just ensures we don't allocate memory in an ISR.

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: I know the FIFO can range from 0-255, but it won't work reliably if we set the depth to a certain value. As what I observed, if for example we set it to 16, the ISR won't trigger even if we have received 16 bytes and sometimes the SCL will be stretched and the master finally timeout before ISR is triggered.

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: Ah, good catch! It should be `&&`.

hal/src/rtl872x/i2c_hal.cpp
scott-brust: Does setting `i2cDev_->IC_RX_TL=0` ensure that the ISR will always be triggered on each byte reception? 
In other words, have you seen the ISR not trigger when only receiving one byte?

XuGuohui: Not really. When thee first byte is received and ISR is triggeered, we need to continuously read available data as long as the RX FIFO is not empty in the ISR, i.e. the slave might have received several more bytes when we execute the ISR. If we only read one byte in ISR, we might have lost data. It seams that the slave hardware will continue accepting data and set flags accordingly, regardless of whether software has handled this flags in time. In the section 19.2.8.1.3 of the UM0400 (user manual), it emphasize something specifically.

scott-brust:
I am trying to understand the purpose of the `slaveRxCache_`. I see that we read data from the hardware fifo into the cache here. 
If we detect a start or stop bit while we are reading data from the FIFO, we know that the data in the cache (potentially) belongs to two separate i2c transactions. (Correct me if I am wrong)

However, it doesnt appear that we know *which* bytes belong to *which* transaction? 

For example: if a master was sending 6 bytes of data to a P2 slave
```
A B C STOP START D E F
``` 
We would read bytes `A B C` into the cache, detect the stop/start condition, set `reportLen = 3`. 
However, if bytes `D E F` were still in the FIFO, we would continue to read them and update the cache length each time, since the interrupt status bits would still be set, right? Or does the peripheral not work like this?

So in the end, we would have 6 bytes in the cache, but we wouldnt know `A B C` were in one transaction, while `D E F` were in another?

Do we work around this by having a 1 byte FIFO right now? (Side note: I tried raising `IC_RX_TL = 8` just as a test, and all of the master/slave tests failed...)

Is the intent of `slaveRxCache_` to buffer data across stop/start conditions while only putting data from a single transaction into the actual HAL `rxBuffer_`?

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -502,58 +576,140 @@ class I2cClass {
...
-            instance->slaveReadRxFifo();
         }
-        if (intStatus & BIT_IC_INTR_STAT_R_STOP_DET) {
-            I2C_ClearINT(instance->i2cDev_, BIT_IC_INTR_STAT_R_STOP_DET);
-            instance->slaveReadRxFifo();
+        uint8_t data = 0xFF;
+        if (txBuffer_.get(&data) == 1) {
+            I2C_SlaveSend(i2cDev_, data);
+        } else {
+            I2C_SlaveSend(i2cDev_, 0xFF);
```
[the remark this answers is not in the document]
XuGuohui: yes, it simulates that the slave has released the bus.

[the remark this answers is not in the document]
XuGuohui: yes, it simulates that the slave has released the bus.

XuGuohui:
>we would continue to read them and update the cache length each time, since the interrupt status bits would still be set, right?

Correct.

>So in the end, we would have 6 bytes in the cache, but we wouldnt know A B C were in one transaction, while D E F were in another

Yes.

>Do we work around this by having a 1 byte FIFO right now?

Exactly.

>Is the intent of slaveRxCache_ to buffer data across stop/start conditions while only putting data from a single transaction into the actual HAL rxBuffer_

Bingo!

hal/src/rtl872x/i2c_hal.cpp
avtolstoy: @scott-brust No need to change this for master as well, we are not using interrupt mode and not preconfiguring the number of bytes to be recvd, see `requestFrom()` implementation where we just manually trigger receiving each byte with `IC_DATA_CMD` in a loop, so this setting does not affect master behavior at all. All good.

XuGuohui:
>Which I assume is ok?

You're right.

>Do you have a device that reproduces the behavior we are trying to handle here?

Just run the I2C unit test. You can add logs in I2C hal to verify this scenario.

scott-brust:
> > Which I assume is ok?
> 
> You're right.
> 
> > Do you have a device that reproduces the behavior we are trying to handle here?
> 
> Just run the I2C unit test. You can add logs in I2C hal to verify this scenario.

You are correct, a restarted message will be split across two invocations of `onReceived_()`
Master Side:
```
    const char * MASTER_TEST_START = "ABC";
    const char * MASTER_TEST_END = "DEF";
    
    USE_WIRE.beginTransmission(I2C_ADDRESS);
    USE_WIRE.write((uint8_t *)MASTER_TEST_START, strlen(MASTER_TEST_START));
    USE_WIRE.endTransmission(false);

    USE_WIRE.beginTransmission(I2C_ADDRESS);
    USE_WIRE.write((uint8_t *)MASTER_TEST_END, strlen(MASTER_TEST_END));
    USE_WIRE.endTransmission(true);
```

Slave side log showing two invocations and 6 total ISR triggers for the 6 bytes:
```
< ABC
On Rec Slave RX count: 5
< DEF
On Rec Slave RX count: 6
```
On the analyzer it looks like this
Screen Shot 2023-04-04 at 10 08 47 AM

scott-brust (approved):

PR #2635, opened 2023-03-09
nRF52: watchdog timeout is not accurate

XuGuohui:
### Problem

As the title describes. The method of calculating the watchdog reload value may cause the intermediate value overflow, which results in an incorrect reload value being loaded to register.

### Solution

Recent nRF5 SDK has fix this. The PR simply extracts the watchdog related files from the recent SDK and replace the existing files.

### Example App

```c
#include "application.h"

SYSTEM_MODE(MANUAL);

void setup() {
    Watchdog.init(WatchdogConfiguration().timeout(240s));
    Watchdog.start();
}

void loop() {
    delay(1s);
    Log.info("Running...");
}
```

### References

https://community.particle.io/t/hardware-watchdog-in-device-os-5-3-0/64031/16

---

### Completeness

- [ ] Added documentation


avtolstoy (approved):

user/tests/wiring/watchdog/watchdog.cpp
```diff
@@ -178,6 +178,34 @@ test(WATCHDOG_07_notify_2) {
     assertEqual(magick, 0xdeadbeef);
 }
 
+#include "nrf_wdt.h"
+test(WATCHDDOG_08_reload_value_is_calculated_correctly) {
```
avtolstoy, line 182: Nit: I'd generalize this test probably and just move the platform-dependent functionality to get current reload value etc into a separate function.

avtolstoy (approved):

PR #2637, opened 2023-03-24
[rtl872x] hal: fix BLE notifying data issue.

XuGuohui:
### Problem

Invoking `BleCharacteristic::setValue()` in characteristic's on written callback will send data multiple times due to retries.

### Solution

Do not wait for data to be sent if sending data from the BLE event dispatching thread

---

### Completeness

- [ ] Added documentation


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2980,6 +2980,10 @@ ssize_t BleGatt::notifyValue(hal_ble_attr_handle_t attrHandle, const uint8_t* bu
                 }
                 uint8_t retry = 3;
                 bool success;
+                if (BleEventDispatcher::getInstance().isThreadCurrent()) {
```
avtolstoy, line 2983: Nit: might look better if we just get out of the do/while loop below based on `isThreadCurrent()` ?

avtolstoy (approved):

PR #2639, opened 2023-03-29
[r510] enable PS (packet switched) Only mode for R510 modems (Boron/BSoM/ESoMX)

technobly:
### Problem

Currently R510 based devices (Boron/BSoM/ESoMX) power up with CS (Circuit Switched) and PS (Packet Switched) operation modes both enabled.  CS mode is not used, therefor it may be disabled which simplifies the data centric nature of the device.

### Solution

Enable PS Only operation mode on first boot.  This feature is non-volatile and only needs to be checked and set once. However, if the cellular registration process should linger, we will also double check that its set correctly during the registration intervening process.

### Steps to Test

- Test with (Boron/BSoM/ESoMX) devices that have never been setup before.  Flash the base firmware without letting the device boot.  Flash new system firmware with tinker-serial-debugging for logging, and cold or warm boot the device.  First boot should show AT+CEMODE? query == 2 and AT+CEMODE=0 followed by AT+CFUN=16.  Successive boots will not show these commands.  Other than this, it's quite difficult to tell that CS mode has been disabled (AT+CREG=2 URCs would need to be enabled, or AT+CREG? querying would need to be performed in a user app.
- Test with a deactivated SIM for approx. 10 minutes that an additional AT+CEMODE? querying is performed between AT+CFUN=0 and AT+CFUN=1.

---

### Completeness


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1758,6 +1774,68 @@ int SaraNcpClient::checkSimReadiness(bool checkForRfReset) {
...
+            // Already in required state
+            return SYSTEM_ERROR_NONE;
+        }
+    }
+
+    int r = SYSTEM_ERROR_UNKNOWN;
+
+    r = parser_.execCommand(1000, "AT+CEMODE=%d",(int)cemode);
+
+    CHECK_PARSER(r);
```
avtolstoy, line 1831: `CHECK_PARSER_OK` ?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,6 +1097,17 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
         CHECK(checkNetConfForImsi());
     }
 
+    // AT+CEMODE? query can take up to 24 seconds to poll without CellularFunctionality::MINIMUM first, and we don't
```
avtolstoy, line 1100: Have we changed the decision on only applying this change during intervention?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,6 +1097,17 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
...
+    // want to be dropping to CellularFunctionality::MINIMUM every boot / connection.
+    // Check cached value to see if we need to change CEMODE=0 (PS_ONLY).
+    if (getOperationModeCached() != particle::to_underlying(CellularOperationMode::PS_ONLY)) {
```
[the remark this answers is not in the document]
technobly: Yeah, I was also thinking that and had it half coded... then removed it to keep the same design patterns already established with CFUN.  It's not very nice to have to convert to underlying all over in higher level code.  I'll switch to that way.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: As discussed this morning, yes we were intending on applying this change at power up for the fastest application of CEMODE... however it would also be nice not to have to apply this forcibly after an OTA update to new firmware, so we'll look at moving the code to a "cold boot" location, instead of the current "warm boot" location.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2243,6 +2321,7 @@ int SaraNcpClient::interveneRegistration() {
                 psd_.reset();
                 registrationInterventions_++;
                 CHECK_PARSER_OK(setModuleFunctionality(CellularFunctionality::MINIMUM));
+                CHECK_PARSER_OK(setOperationMode(CellularOperationMode::PS_ONLY, true /* check */));
```
scott-brust, line 2330: This is in the block checking for 
```
    if (ncpId() != PLATFORM_NCP_SARA_R410 && ncpId() != PLATFORM_NCP_SARA_R510) {
```
So presumably this is for SARA U260? Do we want to send `AT+CEMODE` to non 410/510 modems?

hal/network/ncp_client/sara/sara_ncp_client.h
```diff
@@ -171,6 +171,10 @@ class SaraNcpClient: public CellularNcpClient {
     int waitAtResponseFromPowerOn(ModemState& modemState);
     int disablePsmEdrx();
     int checkSimReadiness(bool checkForRfReset = false);
+    int getOperationModeCached();
+    int setOperationModeCached(CellularOperationMode cemode);
+    int getOperationMode();
+    int setOperationMode(CellularOperationMode cemode, bool check = false, bool save = false);
```
scott-brust, line 177: Currently It looks like we always save the operation mode to the cache. Can we simplify this function by omitting `save` parameter? 
Also, is there ever a situation where we dont want to `check` first? 
Can the logic in this function be simplified to 
1) Check operation mode
2) if it doesnt match, send the CEMODE command
3) Cache the new value
4) return

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2275,6 +2354,7 @@ int SaraNcpClient::interveneRegistration() {
                 eps_.reset();
                 registrationInterventions_++;
                 CHECK_PARSER_OK(setModuleFunctionality(CellularFunctionality::MINIMUM));
+                CHECK_PARSER_OK(setOperationMode(CellularOperationMode::PS_ONLY, true /* check */));
```
scott-brust, line 2363: If I understand `interveneRegistration()` correctly, we should only ever configure the module to Packet Switch only if we have been running for `REGISTRATION_TWILIO_HOLDOFF_TIMEOUT` amount of time (ie 5 minutes) and we have been denied network registration ie `+CEREG:3`?

Is this is the failure state that is fixed with the `PS_ONLY` changes?

hal/network/ncp_client/sara/sara_ncp_client.cpp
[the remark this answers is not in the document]
technobly: FYI, I didn't add the CHECK with the changes because first read will be SYSTEM_ERROR_NOT_FOUND and throw it into a boot loop.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -146,6 +148,9 @@ const int UBLOX_WAIT_AT_RESPONSE_WHILE_UUFWINSTALL_PERIOD = 10000;
 const unsigned CHECK_SIM_CARD_INTERVAL = 1000;
 const unsigned CHECK_SIM_CARD_ATTEMPTS = 10;
 
+const uint32_t SYSTEM_CACHE_OPERATION_MODE_VALID = 0x97c10000;
```
scott-brust, line 151: I'm assuming this is just a random magic pattern to ensure the cached values are valid when read back? 
Are we worried about data in the system cache being unreliable?

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: As it turns out, this is not a reliable way of applying PS Only mode, so we must make the one time change and deal with the minor disconnect/reconnect.

hal/network/ncp/cellular/cellular_ncp_client.h
```diff
@@ -81,6 +81,14 @@ enum class CellularAccessTechnology {
     LTE_NB_IOT = 9
 };
 
+enum class CellularOperationMode {
+    NONE = -1,
+    PS_ONLY = 0,
+    CS_PS_MODE = 2,
```
scott-brust, line 87: Once a device switches to `PS_ONLY` mode, is there ever a reason to switch it back to `CS_PS_MODE`? 
Is there a reason we arent unconditionally switching everything to `PS_ONLY`? (Presumably because we arent fully sure of the behavior change yet?)

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: At this time we are only targeting R510 based devices, so yeah that should be removed.

hal/network/ncp_client/sara/sara_ncp_client.h
technobly: We don't force the cache save in the registration intervening code, but need to force it on first boot.  I was going to optimize the check part as well, but decided against it because it makes the code less modular.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Not really, this is only run here because it's convenient with the CFUN toggle and we are still not connecting.  It's possible that it could be useful here though as a check.  Worth leaving in I think.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: This is mainly to be sure that a value has been set by CEMODE code only and not some other way, "PTCL" :D  I guess the fact that the KEY is found is a similar thing, but it makes me feel better as a double check, and it allowed me to invalidate the data easily while testing.

hal/network/ncp/cellular/cellular_ncp_client.h
technobly: No reason to switch back to CS_PS_MODE other than for testing purposes.  Other devices have been found to already be in PS_ONLY mode by default, so R510 only is to keep the change focused.

avtolstoy (approved):

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,6 +1097,19 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
         CHECK(checkNetConfForImsi());
     }
 
+    // AT+CEMODE? query can take up to 24 seconds to poll without CellularFunctionality::MINIMUM first, and we don't
+    // want to be dropping to CellularFunctionality::MINIMUM every boot / connection.
+    // Check cached value to see if we need to change CEMODE=0 (PS_ONLY).
+    CellularOperationMode cemode;
```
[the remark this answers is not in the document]
technobly: I made sure it was set in getOperationModeCached(), but will initialize it here as well.

PR #2640, opened 2023-03-29
[r510] enable PS (packet switched) Only mode for R510 modems (Boron/BSoM/ESoMX)

technobly:
### Problem

Currently R510 based devices (Boron/BSoM/ESoMX) power up with CS (Circuit Switched) and PS (Packet Switched) operation modes both enabled.  CS mode is not used, therefor it may be disabled which simplifies the data centric nature of the device.

### Solution

Enable PS Only operation mode on first boot.  This feature is non-volatile and only needs to be checked and set once. However, if the cellular registration process should linger, we will also double check that its set correctly during the registration intervening process.

### Steps to Test

- Test with (Boron/BSoM/ESoMX) devices that have never been setup before.  Flash the base firmware without letting the device boot.  Flash new system firmware with tinker-serial-debugging for logging, and cold or warm boot the device.  First boot should show AT+CEMODE? query == 2 and AT+CEMODE=0 followed by AT+CFUN=16.  Successive boots will not show these commands.  Other than this, it's quite difficult to tell that CS mode has been disabled (AT+CREG=2 URCs would need to be enabled, or AT+CREG? querying would need to be performed in a user app.
- Test with a deactivated SIM for approx. 10 minutes that an additional AT+CEMODE? querying is performed between AT+CFUN=0 and AT+CFUN=1.

---

### Completeness

### DIFF
--- a/hal/network/ncp_client/sara/sara_ncp_client.cpp
+++ b/hal/network/ncp_client/sara/sara_ncp_client.cpp
@@ -46,6 +46,8 @@ LOG_SOURCE_CATEGORY("ncp.client");
 #include <lwip/memp.h>
 #include "enumclass.h"
 
+#include "system_cache.h"
+
 #undef LOG_COMPILE_TIME_LEVEL
 #define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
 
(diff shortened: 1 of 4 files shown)


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1092,6 +1097,21 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
         CHECK(checkNetConfForImsi());
     }
 
+    if (ncpId() == PLATFORM_NCP_SARA_R510) {
```
scott-brust, line 1100: This check is because we support other non-R510 SARA modems in the 4.x release line, right? Specifically this is the U260?

scott-brust (approved):

PR #2641, opened 2023-04-04
[p2] fixes SPI speed settings

technobly:
### Problem

P2/Photon2 SPI speeds were half of what was set, while SPI1 was twice that set.

### Solution

- Clocks were flipped between the two, and unfortunately SPI is 25Mhz max now, and SPI1 is 50MHz max now.

- Tests improved.

- Docs need to be updated.

### Steps to Test

`TEST=wiring/no_fixture_spi`

### References

sc-115656

---

### Completeness

- [ ] Added documentation


avtolstoy (approved):

rickkas7:
Firmware reference, P2 and Photon 2 datasheets, and migration guides updated to include a note in the SPI section. 
> If you are using SPI, Device OS 5.3.1 or later is recommended. Prior to that version, SPI ran at half of the set speed, and SPI1 ran at double the set speed. 
> Timing has also been improved for large DMA transfers; prior to 5.3.1, there could be 1 µs gaps for every 16 bytes of data transferred.

PR #2642, opened 2023-04-05
[rtl872x] hal: do not restore backup ram out of flash in case of soft or wdg reset

avtolstoy:
### Problem

A regression was introduced in #2633 where stale backup ram backup (sic!) was restored out of flash in case of soft or watchdog reset, which actually do preserve RAM contents, so this action is not required and has the potential of restoring stale data instead.

### Solution

Add `retained_system` marker field to check whether backup ram looks valid and only restore data out of flash if it's not.

### Steps to Test

`slo/connect_time` should pass and have correct statistics about connection times.

### Example App

N/A

### References

- Fixes regression in #2633

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/backup_ram_hal.cpp
+++ b/hal/src/rtl872x/backup_ram_hal.cpp
@@ -28,6 +28,7 @@
 #include "rtl8721d.h"
 #include "timer_hal.h"
 #include "static_recursive_mutex.h"
+#include "scope_guard.h"
 
 // NOTE: we are using a dedicated flash page for this as timings for writing
 // into fs vs raw page are about 10x
@@ -47,6 +48,9 @@ constexpr system_tick_t syncIntervalMs = 10000;
 StaticRecursiveMutex backupMutex;
 
 }
+
+constexpr uint32_t HAL_BACKUP_RAM_VALID_VALUE = 0x4a171bc4;
+retained_system uint32_t g_backupRamValidMarker;
 
 class BackupRamLock {
 public:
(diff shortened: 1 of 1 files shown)


scott-brust (approved):

hal/src/rtl872x/backup_ram_hal.cpp
```diff
@@ -87,7 +91,15 @@ int hal_backup_ram_init(void) {
     // NOTE: using SDK API here as last reset info in core_hal is initialized later
     platform_system_flags_t dctFlags = {};
     dct_read_app_data_copy(DCT_SYSTEM_FLAGS_OFFSET, &dctFlags, DCT_SYSTEM_FLAGS_SIZE);
-    if ((BOOT_Reason() & BIT_BOOT_DSLP_RESET_HAPPEN) || SYSTEM_FLAG(restore_backup_ram) == 1 || dctFlags.restore_backup_ram == 1) {
+    // In case of software reset or watchdog reset (software reset also uses watchdog reset),
+    // RAM is preserved and we may end up restoring stale data out of flash instead.
+    // Check g_backupRamValidMarker value which resides in system retained area, if it's valid
+    // our backup RAM is fine as-is and we shouldn't restore anything.
+    SCOPE_GUARD({
+        g_backupRamValidMarker = HAL_BACKUP_RAM_VALID_VALUE;
```
scott-brust, line 99: This block ensures that when `hal_backup_ram_init()` exits, `g_backupRamValidMarker` will always be set to `HAL_BACKUP_RAM_VALID_VALUE`. 
So `g_backupRamValidMarker` will only be reset if system RAM is lost (like when we sleep), which is the indication that we *DO* need to restore backup ram. We check for this below. 
Do I understand correctly?

avtolstoy: Yeah. Doing this on exit with `SCOPE_GUARD()` simply because of `CHECK()` potentially interrupting the flow of execution.

technobly (approved):

PR #2647, opened 2023-04-11
[rtl872x] fix sleep PSRAM

avtolstoy:
### Description

1. Application thread stack is statically allocated from SRAM
2. Heap allocations prefer SRAM first now, and then PSRAM, previously it was the opposite
3. ^^ Makes System thread also have stack on SRAM, not PSRAM
4. Minor: system thread has a distinct thread name vs default active_object
5. PSRAM suspend/resume procedure issues resolved
6. Stub out direct flash read from BLE SDK code (we do not allow direct flash access without our HAL functions)

### Steps to Test

- wiring/watchdog
- `runner/mailbox` with System.reset() replaced with System.sleep()

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -428,21 +457,61 @@ class SleepClass {
             PSRAM_datawr = ((temp << 8) | (temp << 24)) | (temps | (temps << 16));
             PSRAM_CTRL_CA_Gen(psram_ca, (BIT11 | BIT0), PSRAM_LINEAR_TYPE, PSRAM_REG_SPACE, PSRAM_WRITE_TRANSACTION);
             PSRAM_CTRL_DPin_Reg(psram_ca, &PSRAM_datawr, PSRAM_WRITE_TRANSACTION);
+
+            // Disable memory access from CPU
```
scott-brust, line 461: Was this supposed to be commented our or not? 
We shouldnt need to disable CPU access to PSRAM since we are going to sleep here shortly anyways (scheduler is disabled, IRQs disabled, nothing else should execute that could access PSRAM)?

hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -428,21 +457,61 @@ class SleepClass {
...
-        /* wait psram exit half sleep mode */
-        DelayUs(100);
-        DCache_Invalidate((uint32_t)PSRAM_BASE, 32);
+        // Disable memory access from CPU
+        PSRAM_DEV->CSR |= BIT_PSRAM_MEM_IDLE;
+        while(!(PSRAM_DEV->CSR & BIT_PSRAM_MEM_IDLE));
+        uint8_t PSRAM_CA[6];
+        // FIXME: No idea how to align addresses properly here to correctly read things out, reading beginning of PSRAM seems to work though
+        PSRAM_CTRL_CA_Gen(PSRAM_CA, 0, PSRAM_LINEAR_TYPE, PSRAM_MEM_SPACE, PSRAM_READ_TRANSACTION);
+        for (int i = 0; i < 10; i++) {
```
scott-brust, line 484: We will loop 10x attempting to wake PSRAM? I guess thats the best we can do, we dont really have any other recourse if we cannot access PSRAM at this point.

hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -428,21 +457,61 @@ class SleepClass {
...
+            PSRAM_DEV->CSR &= (~BIT_PSRAM_DPIN_MODE);
+            PSRAM_DEV->CSR |= PSRAM_DPIN_READ_MODE;
+
+            PSRAM_DEV->CCR = BIT_PSRAM_DPIN;
+            // 150us should be enough, but we'll anyway retry a few times just in case
+            for (int j = 0; j < 150; j++) {
+                if (PSRAM_DEV->CCR & BIT_PSRAM_DPIN) {
+                    break;
+                }
+                DelayUs(1);
```
scott-brust, line 499: `DelayUs()` Is a ROM function right? 
Do we know if it is safe to call when attempting to re-enable PSRAM?

scott-brust (approved):

hal/src/rtl872x/sleep_hal.cpp
avtolstoy: Ideally I'd love to have it here, but this causes problems on boot-up if reset while in sleep, so better to leave commented out.

hal/src/rtl872x/sleep_hal.cpp
avtolstoy: I haven't seen it fail once with this approach honestly. We can potentially fail out and reset/panic.

PR #2649, opened 2023-04-17
[rtl872x] hal: wifi stack stopping to work issue

avtolstoy:
This was kind of a known issue which we removed a workaround for in https://github.com/particle-iot/device-os/commit/1772dfa86d1ccfb2cc763cdc2509cee35bea0fd0 because of BLE/WiFi/coexistence reinitialization issues. This sequence somewhat seems to work, but we need to probably account for advertisement state etc.

@XuGuohui Any suggestions on how to restore the BLE state correctly? Same as in sleep code?

## TODO
1. Look into the sequence for restoring BLE stack state. Same as in sleep code?
2. Check whether WiFi/BLE coexistence still works.


XuGuohui:
Yeah, I think we just need to restore the BLE advertisment state as what we do in sleep.

scott-brust (approved):

hal/network/ncp_client/realtek/rtl_ncp_client.cpp
```diff
@@ -431,6 +431,10 @@ int RealtekNcpClient::scan(WifiScanCallback callback, void* data) {
         }
         return RTW_SUCCESS;
     }, (void*)&ctx);
+    if (rtlError) {
+        LOG(WARN,"wifi_scan_networks err %d", rtlError);
```
avtolstoy, line 435: Nit/style: space after `,`

XuGuohui:
Going to run the test again.

PR #2650, opened 2023-04-18
[rtl872x] BLE scanning panic

XuGuohui:
### Problem

BLE scanning may finally panic after running for a long time, bacause of corrupted BLE stack.

### Solution

Restart BLE stack, instead of panic.

### Steps to Test

`user/tests/app/ble/scanner` or `user/tests/app/ble/uart_central`

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -1629,9 +1629,12 @@ int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
                 HAL_Delay_Milliseconds(10);
             }
             if (isScanning_) {
-                stop();
-                init();
-                start();
+                int ret = stop();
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1626,7 +1628,12 @@ int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
                 }
                 HAL_Delay_Milliseconds(10);
             }
-            SPARK_ASSERT(!isScanning_);
+            if (isScanning_) {
+                stop();
```
avtolstoy, line 1632: What if it's deadlocked etc? Should we check return values here and SPARK_ASSERT if it fails to reinitialize perhaps?

XuGuohui: sounds good.

PR #2653, opened 2023-05-02
[rtl872x] Fix burnin GPIO and SPI Flash tests for photon 2

scott-brust:
### Problem

Both the GPIO and SPI Flash burn in tests fail on Photon 2 test jig
Additionally, GPIO tests were failing due to logical efuse data reads being incorrect and not returning an error code.

### Solution

1. GPIO Tests erroneously included SWD+SWC pins, remove them from the test matrix
2. SPI Flash tests are directly accessing flash. Flash should only be accessed through the HAL to manage the XIP state correctly
3. Require all logical efuse data to be consistent across two consecutive reads

### Steps to Test

Run the burn in tests with the tinker app

### Example App
Force burn to start in with this change to tinkers setup
```c
void setup()
{
    //Setup the Tinker application here

    //Register all the Tinker functions
    Particle.function("digitalread", tinkerDigitalRead);
    Particle.function("digitalwrite", tinkerDigitalWrite);

    Particle.function("analogread", tinkerAnalogRead);
    Particle.function("analogwrite", tinkerAnalogWrite);

#if HAL_PLATFORM_RTL872X && defined(ENABLE_FQC_FUNCTIONALITY)
    BurninTest::instance()->setup(true);
#endif
}
```
Jumper together the GPIOS as defined in the photon 2 pinmap:
```
A0, D10, A1, A2, A5, D5, S4, D4, S3, D3, SCK, D2, MOSI, SCL, MISO, SDA
```
Connect A0-D10, A1-A2, etc.

### References

---

### Completeness

- [ ] Added documentation


user/applications/tinker/src/burnin_test.cpp
```diff
@@ -36,6 +40,67 @@ extern uintptr_t platform_system_part1_flash_start;
 
 namespace particle {
 
+namespace detail {
+
+static const uint32_t softCrc32Tab[] = {
```
scott-brust, line 45: Not sure if its better to copy/paste this here, or include in separate file?

user/applications/tinker/src/burnin_test.cpp
```diff
@@ -353,6 +418,9 @@ bool BurninTest::testSram() {
 	return test_passed;
 }
 
+const int SPI_FLASH_BUFFER_SIZE = 4096;
+static uint8_t spi_flash_buffer[SPI_FLASH_BUFFER_SIZE] = {};
```
scott-brust, line 422: Should this be dynamically allocated instead?

technobly (approved):

hal/src/rtl872x/deviceid_hal.cpp
```diff
@@ -56,9 +57,20 @@ int readLogicalEfuse(uint32_t offset, uint8_t* buf, size_t size) {
 
     memset(efuseBuf, 0xff, LOGICAL_EFUSE_SIZE);
 
-    auto ret = EFUSE_LMAP_READ(efuseBuf);
-    CHECK_TRUE(ret == EFUSE_SUCCESS, SYSTEM_ERROR_INTERNAL);
+    bool dataConsistent = false;
+    for (int i = 0; i < 5 && !dataConsistent; i++) {
+        EFUSE_LMAP_READ(efuseBuf);
+        uint32_t crc1 = HAL_Core_Compute_CRC32(efuseBuf, LOGICAL_EFUSE_SIZE);
```
scott-brust, line 63: How do we want to handle this call for bootloader?
1. Add `HAL_Core_Compute_CRC32` to bootloader / expose `Compute_CRC32`?
2. Exclude this check from bootloader and accept the 4.5% failure rate?
3. include another implementation of `Compute_CRC32` in bootloader like we do for prebootloader

user/applications/tinker/src/burnin_test.cpp
avtolstoy: You can just include the one from tests probably or move it to a common location (e.g. `services`)

user/applications/tinker/src/burnin_test.cpp
avtolstoy: Nit/style: `camelCase`.

Honestly everywhere in this source as well. Let's make sure to keep things consistent: `snake_case` only in old code if rest is using `snake_case` already or in C API interfaces (implementation `camelCase`).

hal/src/rtl872x/deviceid_hal.cpp
avtolstoy: 1 is the right choice.

avtolstoy (approved):

user/applications/tinker/src/burnin_test.cpp
scott-brust: Ok, i'm all over the place with style, ill change things to camelCase in a single commit to make it clear what are only formatting changes

PR #2657, opened 2023-06-01
Fix/nanopb string max size

avtolstoy:
Includes a bugfix for one more issue I found in `spark_wiring_string`.


technobly (approved):
LGTM!  Tested with Argon on 5.3.2 vs. this PR on Particle iOS app with Wi-Fi setup.

PR #2658, opened 2023-06-06
Combined MFG firmware

XuGuohui:
This PR supports detecting and copying an alternate system part firmware from an offset of the flash to the target address. 

### Steps to Test

Collaborate with https://github.com/particle-iot/realtek_ambd_sdk/pull/3 

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/src/main.c
+++ b/bootloader/src/main.c
@@ -411,12 +411,6 @@ int main(void)
     }
     else if (USB_DFU_MODE == 0)
     {
-#if HAL_PLATFORM_RTL872X
-        if (!is_application_valid(ApplicationAddress, NULL)) {
-            FLASH_AddMfgSystemModuleSlot();
-        }
-#endif
-
 #ifdef FLASH_UPDATE_MODULES
         /*
          * Update Internal/Serial Flash based on application_dct=>flash_modules settings
@@ -451,6 +445,11 @@ int main(void)
             uint32_t stack = *(volatile uint32_t*)entry;
             jump_to_system(addr, stack);
         }
+#if HAL_PLATFORM_RTL872X
+        else if (FLASH_AddMfgSystemModuleSlot() == 0) {
+            HAL_Core_System_Reset_Ex(RESET_REASON_UPDATE, 0, NULL);
+        }
+#endif
 #if !HAL_PLATFORM_NRF52840 && !HAL_PLATFORM_RTL872X
         else
         {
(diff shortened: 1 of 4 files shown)


platform/MCU/rtl872x/src/flash_mal.c
```diff
@@ -553,6 +553,24 @@ static int invalidateUpdateSlot(platform_flash_modules_t* module, size_t offset)
     return dct_write_app_data(&module->magicNumber, offset + offsetof(platform_flash_modules_t, magicNumber), sizeof(module->magicNumber));
 }
 
+int FLASH_AddMfgSystemModuleSlot(void) {
+    int ret = SYSTEM_ERROR_INTERNAL;
+    uint32_t systemPartOffset = MFG_COMBINED_FW_START_ADDRESS + FLASH_ModuleLength(FLASH_INTERNAL, MFG_COMBINED_FW_START_ADDRESS) + 4/*CRC32*/;
```
XuGuohui, line 558: Ideally we should parse the binary to figure out the actual offset of system part1, instead of a fixed relative offset.

bootloader/src/main.c
```diff
@@ -411,6 +411,12 @@ int main(void)
     }
     else if (USB_DFU_MODE == 0)
     {
+#if HAL_PLATFORM_RTL872X
+        if (!is_application_valid(ApplicationAddress, NULL)) {
```
avtolstoy, line 415: We should not call this all the time. Let's set DCT flag or something like that if we detect valid system-part at least once, and not call this anymore.

avtolstoy (changes requested):

bootloader/src/main.c
XuGuohui: What about?
```diff
--- a/bootloader/src/main.c
+++ b/bootloader/src/main.c
@@ -445,6 +445,11 @@ int main(void)
             uint32_t stack = *(volatile uint32_t*)entry;
             jump_to_system(addr, stack);
         }
+#if HAL_PLATFORM_RTL872X
+        else if (FLASH_AddMfgSystemModuleSlot() == 0) {
+            HAL_Core_System_Reset_Ex(RESET_REASON_UPDATE, 0, NULL);
+        }
+#endif
 #if !HAL_PLATFORM_NRF52840 && !HAL_PLATFORM_RTL872X
         else
         {
```
This way we don't have to complicate the DCT layout and just at the cost of reseting device one more time.

avtolstoy (approved):

PR #2659, opened 2023-06-06
[rtl872x] usb: HID

avtolstoy:
### Description

USB HID (Mouse / Keyboard) support for P2 / Photon 2.

Same APIs as previously on Photon / P1 / Electron:
https://docs.particle.io/reference/device-os/api/keyboard/keyboard/
https://docs.particle.io/reference/device-os/api/mouse/mouse/

### Caveats

1. `Mouse.moveTo()` might not be moving the cursor (at least on Linux). Might need to come up with a different digitizer descriptor. Wayland uses multiple pointers, so standard pointer doesn't move. Might be the same nowadays on Windows/OSX.
2. We are out of USB endpoints to run CDC + HID at the same time. When both are enabled we are again using a hack from Electron times where one of the CDC endpoints is in fact invalid. This works fine on Linux, double check Windows / OSX behavior, but should probably behave fine as well, as again we've done the same thing with Electron.

### DIFF
--- a/hal/src/rtl872x/usbd_hid.cpp
+++ b/hal/src/rtl872x/usbd_hid.cpp
@@ -1,0 +1,379 @@
+/*
+ * Copyright (c) 2023 Particle Industries, Inc.  All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation, either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "logging.h"
+#include "usbd_hid.h"
+#include <cstring>
+#include <algorithm>
+#include "usbd_wcid.h"
+#include "appender.h"
+#include "check.h"
+#include "scope_guard.h"
+#include "endian_util.h"
+#include <algorithm>
+#include <mutex>
+#include "service_debug.h"
+
+using namespace particle::usbd;
+
+namespace {
+
+const char DEFAULT_NAME[] = HAL_PLATFORM_USB_PRODUCT_STRING " " "HID Mouse / Keyboard";
+
+uint8_t __attribute__((aligned(4))) DEFAULT_HID_DESCRIPTOR[] = {
+    0x05, 0x01,                    // USAGE_PAGE (Generic Desktop)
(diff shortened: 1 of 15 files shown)


hal/src/rtl872x/usbd_hid.h
```diff
@@ -0,0 +1,95 @@
...
+private:
+    int handleInSetupRequest(SetupRequest *req);
+    int handleOutSetupRequest(SetupRequest *req);
+    size_t reportDescriptorLen() const;
+
+protected:
+    HidClassDriver();
+
+private:
+    char* name_ = nullptr;
```
XuGuohui, line 81: Just curiosity: when we should use member initialization list and member initializer like this?

hal/src/rtl872x/usbd_hid.cpp
```diff
@@ -0,0 +1,379 @@
...
+    0x65, 0x00,                    //       UNIT (None)
+    0x75, 0x10,                    //       REPORT_SIZE (16)
+    0x95, 0x02,                    //       REPORT_COUNT (2)
+    0x81, 0x02,                    //       INPUT (Data,Var,Abs)
+    0xc0,                          //     END_COLLECTION
+    0xc0,                          //   END_COLLECTION
+    0xc0                           // END_COLLECTION
+};
+
+const size_t DEFAULT_HID_DESCRIPTOR_DIGITIZER_LEN = 65;
```
XuGuohui, line 143: nit: `constexpr`

hal/src/rtl872x/usbd_hid.h
avtolstoy: General rule of thumb: either everything is initialized in constructor initializer list, or constructor is default and member variables are initialized inline. USB-stuff is second case, so I just left it similar to other USB things.

hal/src/rtl872x/usbd_hid.cpp
```diff
@@ -0,0 +1,379 @@
...
+        return SYSTEM_ERROR_BUSY;
+    }
+
+    if (size > hid::MAX_DATA_PACKET_SIZE) {
+        return SYSTEM_ERROR_TOO_LARGE;
+    }
+
+    txState_ = true;
+
+    memcpy(buffer_, report, size);
```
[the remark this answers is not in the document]
avtolstoy: DMA is not used USB stack as of right now (we had issues with that) so it's ok.

hal/src/rtl872x/usbd_hid.cpp
```diff
@@ -0,0 +1,379 @@
...
+    return SYSTEM_ERROR_INVALID_ARGUMENT;
+}
+
+int HidClassDriver::dataIn(unsigned ep, particle::usbd::EndpointEvent ev, size_t len) {
+    if (ep != epIn_) {
+        return SYSTEM_ERROR_UNKNOWN;
+    }
+
+    if (!txState_) {
+        return 0;
```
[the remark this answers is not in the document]
avtolstoy: This one should be no-error, as otherwise the rest of the class drivers will be asked to handle this endpoint event, even though we know for a fact that it should have been us.

XuGuohui (approved):

eberseth (approved):

PR #2660, opened 2023-06-08
Ledger API

sergeuz:
# Do not merge, superseded by https://github.com/particle-iot/device-os/pull/2715.

This is the first part of the Ledger functionality covering its storage logic and system/application APIs. Below are some links to simplify navigation in the new code:

**Ledger**

- [Application API](https://github.com/particle-iot/device-os/blob/feature/ledger/wiring/inc/spark_wiring_ledger.h)
- [System API](https://github.com/particle-iot/device-os/blob/feature/ledger/system/inc/system_ledger.h)
- [Storage logic](https://github.com/particle-iot/device-os/blob/feature/ledger/system/src/system_ledger_internal.cpp)

**Auxillary classes**

- [Variant](https://github.com/particle-iot/device-os/blob/feature/ledger/wiring/inc/spark_wiring_variant.h)
- [Map](https://github.com/particle-iot/device-os/blob/feature/ledger/wiring/inc/spark_wiring_map.h)

**Outstanding issues**

- The Ledger is disabled on Tracker as we currently don't have enough flash space available on that platform. I still had to disable system error messages (`HAL_PLATFORM_ERROR_MESSAGES`) on Tracker, which freed about 2K of flash space, to accommodate for other changes in this PR.
- The `rmrf()` function added in this PR sometimes fails in load tests. The [issue](https://github.com/particle-iot/device-os/blob/4d1f9569565b45363a9b05787b0ce8c1ccc5f647/services/src/file_util.cpp#L102-L103) only affects the `ledger_purge*` functions and may be specific to LittleFS.

**Testing**

This PR adds some basic unit tests for the auxillary classes but the unit tests for the Ledger itself will be added in another PR with the synchronization logic. For now, the `wiring/ledger` test can be used to validate the basic logic of the Ledger.

**References**

- sc-118798


eberseth:
Could you please add doxygen comments in the new wiring headers?  The comments are automatically parsed by Workbench (VS Code) and appear as tips when implementing those classes/APIs in user apps.

monkbroc (commented):
Looks great! I left some initial comments.

hal/src/nRF52840/littlefs/filesystem.h
```diff
@@ -86,6 +97,10 @@ struct FsLock {
         filesystem_unlock(fs_);
```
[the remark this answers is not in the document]
sergeuz: Good catch, I'll fix it :+1:

system/src/system_ledger_internal.cpp
```diff
@@ -0,0 +1,880 @@
...
+    open_ = true;
+    closeFileGuard.dismiss();
+    return 0;
+}
+
+int LedgerWriter::write(const char* data, size_t size) {
+    if (!open_) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    if (src_ == LedgerWriteSource::USER && dataSize_ + size > LEDGER_MAX_DATA_SIZE) {
```
monkbroc, line 793: If this happens, the ledger write will terminate with an error but `setLedgerData` doesn't set the discard flag on `ledger_close`. Will temporary ledger files accumulate or are they cleaned up some other way?

sergeuz: The destructor of `LedgerWriter` discards the data.

sergeuz: Hm, actually the app API doesn't close the stream with the discard flag set if writing fails. Good catch!

sergeuz: Specifically, the scope guard that gets executed when a write error occurs should be closing the stream with `LEDGER_STREAM_CLOSE_DISCARD`: https://github.com/particle-iot/device-os/blob/feature/ledger/wiring/src/spark_wiring_ledger.cpp#L147-L149

hal/src/nRF52840/littlefs/filesystem.h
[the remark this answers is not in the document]
sergeuz: Fixed.

system/src/system_ledger_internal.cpp
sergeuz: Fixed.

monkbroc (approved):
Very nice work! I'm approving this PR so it is ready to merge when it's time.

wiring/src/spark_wiring_json.cpp
```diff
@@ -477,16 +489,30 @@ spark::JSONWriter& spark::JSONWriter::value(unsigned long val) {
...
+spark::JSONWriter& spark::JSONWriter::value(long long val) {
+    writeSeparator();
+    printf("%lld", val);
+    state_ = NEXT;
+    return *this;
+}
+
+spark::JSONWriter& spark::JSONWriter::value(unsigned long long val) {
+    writeSeparator();
+    printf("%llu", val);
```
eberseth, line 501: I don't think that our libc support this format type, `%llu`

wiring/inc/spark_wiring_platform.h
```diff
@@ -166,5 +170,9 @@
 #define Wiring_Watchdog 0
 #endif // Wiring_Watchdog
 
+#ifndef Wiring_Ledger
+#define Wiring_Ledger 0
+#endif
+
```
eberseth, line 176: Please add Variant and Map defines here as well in case we want to reuse them for external libraries.
```cpp
#define Wiring_Variant 1
#define Wiring_Map 1
```

PR #2662, opened 2023-06-12
[nrf] Add Factory Reset Tests

scott-brust:
### Problem

Factory Reset is a feature where an application module can be stored in flash at a specified `EXTERNAL_FLASH_FAC_ADDRESS` address and copied over the running application either by the firmware itself via `System.factoryReset()` or via a USB control request. 

This feature is used internally and by some customers but is not tested. We should expand the automated testing to cover these use cases.

### Solution

Add tests to OTA update suite. They are mostly the same as the multiple ota tests.

### Steps to Test

Use test runner on a gen 3 platform
```
device-os-test --test-dir ~/develop/device-os/user/tests/integration --device-os-dir ~/develop/device-os run ota/factory_reset argon -v -- -fixture
```

### Example App
See test app 

### References

Side note: Realtek platforms do not support Factory Reset (yet). There are a few things that need to be added to support it:
- There is no fixed Factory Reset module address / `factory_module` defined in `ota_module_bounds.c`. I dont know if we want to / can support allocated a fixed portion of flash for this purpose.
- The user application load address is dynamic, but we can calculate where it should be by working backwards from the user module bounds end location
- We do have an entry in the `flash_modules` DCT field. 
- We do support backup registers / retained memory for system flags already
---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/tests/integration/ota/factory_reset/factory_reset.spec.js
+++ b/user/tests/integration/ota/factory_reset/factory_reset.spec.js
@@ -1,0 +1,127 @@
+suite('Factory Reset');
+
+platform('gen3');
+systemThread('enabled');
+
+const { HalModuleParser, ModuleInfo, updateModulePrefix, updateModuleSuffix, updateModuleCrc32 } = require('binary-version-reader');
+const tempy = require('tempy');
+
+const { readFile } = require('fs').promises;
+const path = require('path');
+
+let api = null;
+let auth = null;
+let device = null;
+let deviceId = null;
+
+async function flash(ctx, binFile, { timeout = 120000, mayFail = false } = {}) {
+	const appName = path.basename(binFile, '.bin');
+	await api.flashDevice({ deviceId, files: { [appName]: binFile }, auth });
+	const ok = await waitFlashStatusEvent(ctx, timeout);
+	if (!ok && !mayFail) {
+		throw new Error('Update failed');
+	}
+	return ok;
+}
+
+async function waitFlashStatusEvent(ctx, timeout) {
+	let timeoutAt = Date.now() + timeout;
+	let data = null;
+	for (;;) {
+		const t = timeoutAt - Date.now();
+		if (t <= 0) {
+			throw new Error("Event timeout");
+		}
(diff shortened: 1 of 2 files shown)


user/tests/integration/ota/factory_reset/factory_reset.spec.js
```diff
@@ -0,0 +1,127 @@
...
+	auth = this.particle.apiClient.token;
+	device = this.particle.devices[0];
+	deviceId = device.id;
+});
+
+test('01_disable_resets_and_connect', async function () {
+	// See the test app
+});
+
+test('02_flash_binary', async function () {
```
scott-brust, line 61: There should be an easy way to refactor this to a single function, I had a hard time debugging it so left it as is for now

technobly (approved):

avtolstoy:
This is blocked by https://github.com/particle-iot/device-os-test-runner/pull/25

user/tests/integration/ota/factory_reset/factory_reset.spec.js
```diff
@@ -0,0 +1,117 @@
...
+});
+
+test('08_move_ota_binary_to_factory_slot', async function () {
+	// See the test app: Copy OTA binary to Factory Reset slot, update DCT to denote valid Factory Reset module present
+});
+
+test('09_usb_command_factory_reset', async function () {
+	let usbDevice = await device.getUsbDevice();
+	device.setWillDetach(true);
+	await usbDevice.factoryReset({ timeout: 10000 });
```
scott-brust, line 112: Should I have some sort of try/catch around this? The test runner implementation of device reset handles a number of errors and updates its connection state

avtolstoy: Under normal conditions, this should probably work out just fine. All we care about is that factory reset request is sent to the device/response is received and as far as test runner is concerned we've set `willDetach`, so it should expected a loss of USB connection to device.

avtolstoy (approved):

PR #2664, opened 2023-06-21
[rtl872x] dynamically enable Wi-Fi stack on demand and fix BLE race c…

XuGuohui:
### Problem

RTL872x-specific:

As of right now, the Wi-Fi stack is enabled on startup to support BLE functionality and it is always enabled during the application's life time. Thus, it consumes extremely high power that is around 60mA.

There is also deadlock and race condition in BLE HAL. The race condition happens when BLE is connected, followed by  entering sleep mode.

### Solution

Dynamically enable Wi-Fi stack on demand, e.g. requesting to enter BLE provisioning mode or calling `BLE.on()`/`WiFi.on()`.

Regarding the deadlock in BLE HAL, the workaround is to not acquire the `connectionsMutex_` in disconnect APIs, because they don't actually operate the connections data. The race condition is fixed by always clearing the advertising state explicitly in the stop function.

---

### Completeness

- [ ] Added documentation


system/src/ble_control_request_channel.cpp
```diff
@@ -1044,6 +1044,11 @@ int BleControlRequestChannel::initProfile() {
         LOG_DEBUG(TRACE, "BLE profile already initialized");
         return SYSTEM_ERROR_INVALID_STATE;
     }
+#if HAL_PLATFORM_RTL872X
+    SCOPE_GUARD ({
+        hal_ble_stack_deinit(nullptr);
```
avtolstoy, line 1049: What if `BLE.on()` was called previously by user app? We should probably check the state before this was called?

system/src/ble_provisioning_mode_handler.cpp
```diff
@@ -351,6 +351,9 @@ int BleProvisioningModeHandler::exit() {
         preSrData_.clear();
         ctrlReqAdvData_.clear();
         ctrlReqSrData_.clear();
+#if HAL_PLATFORM_RTL872X
+        hal_ble_stack_deinit(nullptr);
```
avtolstoy, line 355: Same here probably?

system/src/ble_control_request_channel.cpp
XuGuohui: Fixed

XuGuohui: Fixed

hal/inc/ble_hal.h
```diff
@@ -495,6 +495,15 @@ int hal_ble_stack_init(void* reserved);
...
 int hal_ble_stack_deinit(void* reserved);
 
+/**
+ * Check if BLE stack is initialized.
+ *
+ * @param[in]   reserved    Reserved for future use.
+ *
+ * @returns     true if it is initialized, otherwise, false.
+ */
+bool hal_ble_is_initialized(void* reserved);
```
avtolstoy, line 505: Just a nit: for things like this perhaps might be a better idea from the get-go introduce a method to return an enum `state`, instead of one purpose `initialized` / `not initialized`.

avtolstoy (approved):

PR #2668, opened 2023-07-11
Assets OTA

avtolstoy:
### Description

Assets OTA

### TODO

1. Automated test for assets OTA
2. system describe communication unit tests are broken due to migration from JSON to protobuf

#### P2 / Photon 2

1. Update device to latest available Device OS (5.4.1)
1.1. The easiest is `device-os-flash --all-devices 5.4.1`
2. Build/flash bootloader
2.2. `(bootloader) $ make PLATFORM=p2 COMPILE_LTO=n clean all`
2.3. Flash bootloader `device-os-flash --all-devices $DEVICE_OS_DIR/build/target/bootloader/platform-32-m/bootloader.bin`
3. Build/flash Device OS and application in whichever convenient way (e.g. Workbench with `deviceOS@source`)

#### Gen 3 (nRF52840 platforms)

1. Update device to latest available Device OS (5.4.1)
1.1. The easiest is `device-os-flash --all-devices 5.4.1`
2. Build/flash bootloader
2.2. `(bootloader) $ make PLATFORM=boron COMPILE_LTO=n clean all`
2.3. Flash bootloader `device-os-flash --all-devices $DEVICE_OS_DIR/build/target/bootloader/platform-13-m/bootloader.bin`
3. Build/flash Device OS and application in whichever convenient way (e.g. Workbench with `deviceOS@source`)

### Example App

See `app/assets`.

### References

N/A

---

### Completeness

- [ ] Added documentation


modules/shared/nRF52840/part1_build.mk
```diff
@@ -49,6 +49,11 @@ ASFLAGS += -I$(COMMON_BUILD)/arm/startup
 ASFLAGS +=  -Wa,--defsym -Wa,SPARK_INIT_STARTUP=0
 ASFLAGS += -D__STACKSIZE__=$(MAIN_STACK_SIZE) -D__STACK_SIZE=$(MAIN_STACK_SIZE)
 
+# Tracker
+ifeq ($(PLATFORM_ID),26)
+USE_PRINTF_FLOAT ?= n
```
monkbroc, line 54: Is this a breaking change for customer applications using printf?

avtolstoy: Yes, but I have a workaround for this.

wiring/inc/spark_wiring_asset.h
```diff
@@ -0,0 +1,95 @@
...
+    size_t size() const;
+
+    bool isValid() const;
+    bool isReadable();
+
+    int available() override;
+    int read() override;
+    int peek() override;
+    // Not present in Stream
+    virtual int read(char* buffer, size_t size);
```
eberseth, line 46: For convenience, can we add an `uint8_t`  overload `virtual int read(uint8_t* buffer, size_t size);`?

mrlambchop (approved):

XuGuohui (changes requested):
Great job 👍 ! This is a huge PR and I just had a brief review. The main problem I found is we use memory allocation in many places without checking the result. This is a potential risk. I'll go though the asset OTA process as well.

wiring/src/spark_wiring_system.cpp
```diff
@@ -99,6 +100,66 @@ bool SystemClass::enableFeature(LoggingFeature) {
...
+    info.size = sizeof(info);
+    int r = asset_manager_get_info(&info, nullptr);
+    if (r) {
+        return assets;
+    }
+    SCOPE_GUARD({
+        asset_manager_free_info(&info, nullptr);
+    });
+    for (size_t i = 0; i < info.required_count; i++) {
+        assets.append(ApplicationAsset(&info.required[i]));
```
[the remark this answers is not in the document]
avtolstoy: @XuGuohui Sure, it might fail, the problem is that there is no easy way to deal with that in a sane manner here with the current API and even if we change it it doesn't bring any joy (e.g. if we return `SYSTEM_ERROR_NO_MEMORY`). What we do now is we simply not report the asset as available or required. If you under low memory conditions there is no way to read the assets anyway (memory required for decompression etc).

eberseth (approved):

wiring/inc/spark_wiring_asset.h
```diff
@@ -0,0 +1,95 @@
...
+#include "asset_manager_api.h"
+#include "spark_wiring_stream.h"
+#include <memory>
+#include "spark_wiring_vector.h"
+#include <functional>
+
+namespace particle {
+
+class ApplicationAsset: public Stream {
+public:
```
eberseth, line 30: doxygen comments please for  public methods here

wiring/inc/spark_wiring_system.h
```diff
@@ -1037,6 +1038,31 @@ class SystemClass {
 #endif
     }
 
+#if HAL_PLATFORM_ASSETS
+    static spark::Vector<ApplicationAsset> assetsRequired();
+    static spark::Vector<ApplicationAsset> assetsAvailable();
+    static int assetsHandled(bool state = true);
+
+    static int onAssetsOta(OnAssetsOtaCallback cb, void* context = nullptr);
```
eberseth, line 1046: doxygen comments for public methods please

hal/src/rtl872x/ota_module_bounds.c
```diff
@@ -48,9 +48,9 @@ const module_bounds_t module_km0_part1 = {
 
 // OTA region, to be updated.
 module_bounds_t module_ota = {
-    .maximum_size = 0x200000, //2M
+    .maximum_size = 0x180000, // 1.5MB
```
eberseth, line 51: Can we get these values from the linker script?

system/inc/asset_manager_api.h
```diff
@@ -0,0 +1,279 @@
...
+    uint16_t required_count;
+
+    asset_manager_asset* available;
+    uint16_t available_count;
+} asset_manager_info;
+
+struct asset_manager_stream;
+typedef struct asset_manager_stream asset_manager_stream;
+
+typedef enum asset_manager_consumer_state {
```
eberseth, line 150: Please comment these cases about what the do or represent

system/src/asset_manager.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+int parseAssetInfo(InputStream* stream, size_t size, Asset& asset) {
+    std::unique_ptr<module_info_name_ext_t> nameExt;
+    AssetHash hash;
+    while (size > 0) {
+        module_info_extension_t ext = {};
+        CHECK(stream->peek((char*)&ext, sizeof(ext)));
+        CHECK_TRUE(ext.length <= size, SYSTEM_ERROR_BAD_DATA);
+        if (ext.type == MODULE_INFO_EXTENSION_NAME) {
+            CHECK_TRUE(ext.length > sizeof(module_info_name_ext_t), SYSTEM_ERROR_BAD_DATA);
```
eberseth, line 73: Just verifying `>` here and `>=` (line 78) are intentionally different.

system/src/asset_manager.cpp
```diff
@@ -0,0 +1,479 @@
...
+    CHECK(stream_->read((char*)&prefix, sizeof(prefix)));
+    CHECK_TRUE(prefix.module_function == MODULE_FUNCTION_ASSET, SYSTEM_ERROR_BAD_DATA);
+    CHECK_TRUE(prefix.flags & MODULE_INFO_FLAG_DROP_MODULE_INFO, SYSTEM_ERROR_BAD_DATA);
+    CHECK_FALSE(prefix.flags & MODULE_INFO_FLAG_PREFIX_EXTENSIONS, SYSTEM_ERROR_BAD_DATA);
+    CHECK_TRUE(moduleSize == ((uintptr_t)prefix.module_end_address - (uintptr_t)prefix.module_start_address + sizeof(uint32_t)), SYSTEM_ERROR_BAD_DATA);
+    bool compressed = prefix.flags & MODULE_INFO_FLAG_COMPRESSED;
+
+    LOG(INFO, "prefix func=%d flags=%x start=%x end=%x compressed=%d", prefix.module_function, prefix.flags, prefix.module_start_address, prefix.module_end_address, compressed);
+
+    // Comperssion header
```
eberseth, line 375: "Compression"

system/src/asset_manager.cpp
```diff
@@ -0,0 +1,479 @@
...
+          compressed_(false),
+          size_(0),
+          originalSize_(0),
+          dataOffset_(0),
+          dataSize_(0) {
+}
+
+AssetReader::AssetReader(const char* filename)
+        : AssetReader((InputStream*)nullptr) {
+    fileStream_ = std::make_unique<FileInputStream>(filename, FILESYSTEM_INSTANCE_ASSET_STORAGE);
```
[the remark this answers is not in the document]
avtolstoy: This is checked in other methods.

technobly (approved):
Works great! :shipit:

modules/shared/nRF52840/part1_build.mk
monkbroc: Can you elaborate on the workaround? I tried `Serial.printlnf("pi=%f", M_PI);` on a Tracker with 5.4.1 and I got the expected `pi=3.141593` but with this branch, I get `pi=`

eberseth: What is the workaround because Tracker Edge/Monitor Edge uses a lot of floats and doubles

eberseth: Mostly in the jsmn JSON writer but also logging of various sensor data and GPS

services/inc/storage_streams.h
```diff
@@ -0,0 +1,462 @@
...
+public:
+    FileInputStream(const char* filename, filesystem_instance_t instance, size_t offset = 0)
+            : fs_(nullptr),
+              file_{},
+              remaining_(0),
+              offset_(offset),
+              isOpen_(false) {
+        fs_ = filesystem_get_instance(instance, nullptr);
+        open(filename);
+    }
```
sergeuz, line 139: Consider only providing a default constructor and exposing the `open()` method so that the possible error when opening the file can be properly handled by the calling code.

services/inc/storage_streams.h
```diff
@@ -0,0 +1,462 @@
...
+
+    int peek(char* data, size_t size) override {
+        if (!remaining_) {
+            return SYSTEM_ERROR_END_OF_STREAM;
+        }
+        CHECK_TRUE(data, SYSTEM_ERROR_INVALID_ARGUMENT);
+        size = std::min(size, remaining_);
+        fs::FsLock lock(fs_);
+        CHECK(lfs_file_seek(&fs_->instance, &file_, offset_, LFS_SEEK_SET));
```
sergeuz, line 161: Make sure `CHECK_FS` is used with LittleFS functions here and in other places in this PR.

sergeuz (commented):
Posting the comments I have so far. Will continue looking into this tomorrow.

system/src/asset_manager.cpp
```diff
@@ -0,0 +1,487 @@
...
+    }
+#if HAL_PLATFORM_ASSETS_OTA_OVERLAP
+    if (size > EXTERNAL_FLASH_OTA_SAFE_LENGTH) {
+        LOG(WARN, "Receiving module of size %u (function = %d) will invalidate asset storage", size, moduleFunction);
+        // We cannot receive an asset that will also cause the asset filesystem to be overwritten
+        if (moduleFunction == MODULE_FUNCTION_ASSET) {
+            return SYSTEM_ERROR_TOO_LARGE;
+        }
+        // Otherwise unmount/invalidate filesystem
+        CHECK(formatStorage(false /* remount */));
```
sergeuz, line 331: Note that `prepareForOta` is called after the update manager's `startUpdate` which erases the OTA section. It likely wasn't intended to work that way.

Also, what would happen if the app tries to read an asset after the asset filesystem was invalidated? Would the app get an error or cause a UB? Is that a supported use case at all?

system/src/asset_manager.cpp
```diff
@@ -0,0 +1,487 @@
...
+            SYSTEM_ERROR_NOT_ENOUGH_DATA);
+    size_t moduleSize = CHECK(stream_->availForRead());
+
+    // Prefix
+    module_info_t prefix = {};
+    CHECK(stream_->read((char*)&prefix, sizeof(prefix)));
+    CHECK_TRUE(prefix.module_function == MODULE_FUNCTION_ASSET, SYSTEM_ERROR_BAD_DATA);
+    CHECK_TRUE(prefix.flags & MODULE_INFO_FLAG_DROP_MODULE_INFO, SYSTEM_ERROR_BAD_DATA);
+    CHECK_FALSE(prefix.flags & MODULE_INFO_FLAG_PREFIX_EXTENSIONS, SYSTEM_ERROR_BAD_DATA);
+    CHECK_TRUE(moduleSize == ((uintptr_t)prefix.module_end_address - (uintptr_t)prefix.module_start_address + sizeof(uint32_t)), SYSTEM_ERROR_BAD_DATA);
```
sergeuz, line 378: Let's add a comment that `sizeof(uint32_t)` is the CRC.

system/src/asset_manager_api.cpp
```diff
@@ -0,0 +1,172 @@
...
+    CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
+    auto internal = new Info();
+    CHECK_TRUE(internal, SYSTEM_ERROR_NO_MEMORY);
+    internal->required = AssetManager::instance().requiredAssets();
+    internal->available = AssetManager::instance().availableAssets();
+
+    info->internal = (void*)internal;
+
+    bool ok = false;
+    SCOPE_GUARD({
```
sergeuz, line 61: We also have `NAMED_SCOPE_GUARD` that can be dismissed.

system/src/asset_manager_api.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+int asset_manager_free_info(asset_manager_info* info, void* reserved) {
+    CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
+    if (info->internal) {
+        auto internal = static_cast<Info*>(info->internal);
+        delete internal;
+    }
+    if (info->required) {
+        free(info->required);
```
sergeuz, line 99: `free()` is used to free the memory allocated with `new`.

Same [here](https://github.com/particle-iot/device-os/pull/2668/files#diff-4c430d849bb9061cdddf693086b5b007baaafbf62d9dd4ef3b87aefcb7619d03R102).

system/src/asset_manager_api.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+int asset_manager_get_info(asset_manager_info* info, void* reserved) {
+    CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
+    auto internal = new Info();
+    CHECK_TRUE(internal, SYSTEM_ERROR_NO_MEMORY);
+    internal->required = AssetManager::instance().requiredAssets();
+    internal->available = AssetManager::instance().availableAssets();
+
+    info->internal = (void*)internal;
```
sergeuz, line 58: This internal info doesn't seem to be used by any of the API calls. If so, let's remove it and save on memory.

system/inc/asset_manager_api.h
```diff
@@ -0,0 +1,292 @@
...
+            memset(buf.data(), 0, buf.size());
+            bytes2hexbuf_lower_case((const uint8_t*)hash_.data(), hash_.size(), buf.data());
+            return String(buf.data());
+        }
+    }
+    return String();
+}
+
+inline bool AssetHash::operator==(const AssetHash& other) const {
+    return (type_ == other.type_ && hash_ == other.hash_);
```
sergeuz, line 112: Is it intentional that the asset name is not taken into account here? It's worth adding a comment if so.

sergeuz (approved):
Good job. I don't think I found anything major, although I must admit I don't fully follow the logic where a file is used to store the decompression buffer. Hopefully that is well covered by tests.

modules/shared/nRF52840/part1_build.mk
avtolstoy: Just to re-iterate: this is a temporary change and will be resolved (restored and some other stuff thrown away) before this PR is merged. This is failing a lot of our automated tests, so won't be unnoticed.

system/src/asset_manager.cpp
eberseth: Using `>` would allow for a minimum of one additional character, which is probably null.

system/src/asset_manager_api.cpp
avtolstoy: That's the storage for at the very least hash and name.

sergeuz: Ah, didn't notice that :+1:

services/inc/storage_streams.h
avtolstoy: I will leave this one as-is just for simplicity of not having to carry around `filename` (see `AssetReader` usage as well).

system/inc/asset_manager.h
```diff
@@ -0,0 +1,158 @@
+/*
```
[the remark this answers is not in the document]
avtolstoy: Also required by hal unfortunately (ota_flash_hal.cpp)

PR #2673, opened 2023-07-19
[rtl872x] WPA3 and hidden SSID support

avtolstoy:
### Description

Support for WPA3 and hidden SSIDs (note that correct channel plan e.g. US needs to be set for 5GHz hidden SSID support to work, otherwise 5GHz scans are passive).

### Steps to Test

1. Configure WPA3 credentials over Serial in listening mode and attempt to connect (need to select '6' for WPA3, won't be prompted as an option)
2. Configure credentials with hidden SSID with a test app and attempt to connect
```c++
    WiFiCredentials cred;
    cred.setSsid("Particle");
    cred.setPassword("Particle");
    cred.setCipher(WLAN_CIPHER_AES);
    cred.setSecurity(spark::SecurityType::WPA2);
    cred.setHidden(true);
    WiFi.setCredentials(cred);
```

### Problems

CLI uses hardcoded serial prompts as triggers (https://github.com/particle-iot/particle-cli/blob/master/src/cmd/serial.js#L1003), adding WPA3 option there breaks older CLI.

Should we migrate to using USB requests instead?

---

### Completeness

- [ ] Added documentation


technobly (approved):

avtolstoy:
Note for future reference: there is still a problem with Realtek SDK and WPA version, we'll look into that before 5.5.0 final.

PR #2674, opened 2023-07-21
Enables UPSV=1 low power mode for R510 when idle for >=9.2s

technobly:
### Problem

- R510 modem consumes more power than R410 modem by design.  This is unexpected behavior for some customers migrating from R410 to R510.

### Solution

- Reduce power consumption for R510 based devices by enabling UPSV=1 mode which will automatically reduce power consumption after 9.2s of being idle.  Note: Some extreme cases of the modem requiring over 2 minutes to drop into a low power state have been observed and are being investigated.  Typically though, if the AT interface is idle, the modem will automatically reduce power consumption.

### Steps to Test

- Compile `APP=tinker-serial1-debugging` and observe after device is idle, power consumption drops from 30-38mA avg. down to 8-12mA avg.
- Integration test - `device-os-test run bsom wiring/no_fixture_long_running -v -- -fixture`
- Integration test - `device-os-test run boron wiring/no_fixture_long_running -v -- -fixture`
- Integration test - `device-os-test run esomx wiring/no_fixture_long_running -v -- -fixture`

### References

- https://github.com/particle-iot/gsm0710muxer/pull/11

---

### Completeness

- [ ] Added documentation


hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -527,7 +529,10 @@ int SaraNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
         }
     }
 
-    int err = muxer_.writeChannel(UBLOX_NCP_PPP_CHANNEL, data, size);
+    int err = -1;
```
avtolstoy, line 532: This should be `0` by default, right? Otherwise we'll fall into `disable()` below on line 544.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2182,14 +2228,22 @@ int SaraNcpClient::getMtu() {
...
         }
     } else {
+        sleepNoPPPWrite_ = true;
+        // R510 may be in low power mode, wake it up so we can get our muxer data channel suspend
+        // echo, before we get into sleep and potentially prematurely wake by "network activity" (RX data)
+        if (ncpId() == PLATFORM_NCP_SARA_R510) {
+            CHECK(waitAtResponse(5000, 2000));
+        }
         CHECK_TRUE(muxer_.suspendChannel(UBLOX_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
+        HAL_Delay_Milliseconds(100); // allow a bit of time for muxer echo response
```
avtolstoy, line 2246: Should also set keepalive to 0 here perhaps?

hal/src/nRF52840/sleep_hal.cpp
```diff
@@ -816,6 +833,7 @@ static void fpu_sleep_prepare(void) {
 
 static int enterStopBasedSleep(const hal_sleep_config_t* config, hal_wakeup_source_base_t** wakeupReason) {
     int ret = SYSTEM_ERROR_NONE;
+    uint32_t ncpId_ = platform_primary_ncp_identifier(); // save before external flash is put to sleep
```
avtolstoy, line 836: Nit: remove `_`

user/tests/wiring/no_fixture_long_running/network.cpp
```diff
@@ -75,7 +75,7 @@ bool udpEchoTest(UDP* udp, const IPAddress& ip, uint16_t port, const uint8_t* se
 #define __stringify(x) #x
 
 #ifndef UDP_ECHO_SERVER_HOSTNAME
-#define UDP_ECHO_SERVER_HOSTNAME not_defined
+#define UDP_ECHO_SERVER_HOSTNAME publish-receiver-udp.particle.io
```
avtolstoy, line 78: This should be reverted, right?

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Yeah, I'll change to `int err = gsm0710::GSM0710_ERROR_NONE;`

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: I'm guessing you mean the PPP echo timer, since the muxer AT channel is already suspended?  PPP data writes should also be blocked with `sleepNoPPPWrite_` already?

avtolstoy: No, I was mainly thinking about muxer keepalives, to avoid sending a keepalive right before sleep, going into it and then waking up wrongly.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+    const uint32_t WAIT_FOR_LOW_POWER_ACTIVE_MS = 20000;
+    const uint32_t WAIT_FOR_LOW_POWER_MEAS_MS = 10000;
+    const uint32_t LOW_POWER_ATTEMPTS_MAX = 10; // XXX: extreme cases take up to 2.5 minutes to drop into low power mode
+    do {
+        Particle.connect();
+        assertTrue(waitFor(Particle.connected, CLOUD_CONNECT_TIMEOUT));
+        delay(WAIT_FOR_LOW_POWER_ACTIVE_MS); // wait for UPSV=1 default delay of ~9.2s before modem drops into low power mode idle mode.
+
+        period = measureAvgPeriodMs(CTS1, WAIT_FOR_LOW_POWER_MEAS_MS);
+        if (period <= 1280 + 128 && period >= 1280 - 128) { // 1.28s +/- 10% if in low power mode
```
scott-brust, line 94: The idea here is to check for ~1.28s pulses on modem CTS1 as an indication that we are in low power mode? 
I assume this line is normally high (or low) during normal operation and if we see the pulses then we can safely assume we have entered UPSV mode?

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+        return;
+    }
+
+    bool publishResult = false;
+    bool lowPowerResult = false;
+    int attempts = 0;
+    uint32_t period = 0;
+    const uint32_t WAIT_FOR_LOW_POWER_ACTIVE_MS = 20000;
+    const uint32_t WAIT_FOR_LOW_POWER_MEAS_MS = 10000;
+    const uint32_t LOW_POWER_ATTEMPTS_MAX = 10; // XXX: extreme cases take up to 2.5 minutes to drop into low power mode
```
scott-brust, line 87: nit: It looks like you can move these three `const` variables to the namespace scope to avoid having to redeclare them in multiple tests?

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+    do {
+        Particle.connect();
+        assertTrue(waitFor(Particle.connected, CLOUD_CONNECT_TIMEOUT));
+        delay(WAIT_FOR_LOW_POWER_ACTIVE_MS); // wait for UPSV=1 default delay of ~9.2s before modem drops into low power mode idle mode.
```
scott-brust, line 91: Does the connect() and delay() need to be inside the do/while loop? 
Or can we just do it once and then only check the state of the CTS1 pin in the loop?

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+        period = measureAvgPeriodMs(CTS1, WAIT_FOR_LOW_POWER_MEAS_MS);
+        if (period <= 1280 + 128 && period >= 1280 - 128) { // 1.28s +/- 10% if in low power mode
+            lowPowerResult = true;
+        } else {
+            lowPowerResult = false;
+        }
+
+        publishResult = Particle.publish("my_event_low_power", "event data low power", PRIVATE | WITH_ACK);
```
scott-brust, line 100: Does it matter if this always publishes even if `lowPowerResult` is false?
The test runner side will pass, but if the modem doesnt enter low power the device will still fail the test right?

user/tests/wiring/no_fixture_power_saving/no_fixture_power_saving.spec.js
```diff
@@ -0,0 +1,128 @@
...
+test('POWER_SAVING_01_particle_publish_publishes_an_event_after_low_power_active', async function() {
+    if (skipTest) {
+        return;
+    }
+
+    const data = await this.particle.receiveEvent('my_event_low_power');
+    expect(data).to.equal('event data low power');
+});
+
+test('POWER_SAVING_02_register_function_and_connect_to_cloud', async function() {
```
scott-brust, line 82: What is this test testing? It looks like the device connects and stays in low power mode, then the test runner side queries it for its device functions? 
This is to ensure that the cloud side can communicate with the device even when the modem enters low power mode?

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+    assertLessOrEqual(period, 1280 + 128);
+}
+
+test(POWER_SAVING_03_call_function_and_check_return_value_after_low_power_active) {
+    if (modem_type != DEV_SARA_R510) {
+        skip();
+        return;
+    }
+
+    // See no_fixture_long_running.spec.js
```
scott-brust, line 144: The implication is that we confirmed the device is in power save mode in test 2, so test 3 can continue on the test runner side (which tests that the cloud side can call a function when the device is in low power mode)?

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+        Particle.connect();
+        assertTrue(waitFor(Particle.connected, CLOUD_CONNECT_TIMEOUT));
+        delay(WAIT_FOR_LOW_POWER_ACTIVE_MS); // wait for UPSV=1 default delay of ~9.2s before modem drops into low power mode idle mode.
+
+        period = measureAvgPeriodMs(CTS1, WAIT_FOR_LOW_POWER_MEAS_MS);
+        if (period <= 1280 + 128 && period >= 1280 - 128) { // 1.28s +/- 10% if in low power mode
+            lowPowerResult = true;
+        } else {
+            lowPowerResult = false;
+        }
```
scott-brust, line 130: nit: it seems like you could refactor the low power check into a helper function since the same logic is done in a couple different places. Something like
```
bool enteredSleepMode() {
    <lines 125-130>
    return lowPowerResult;
}
```
or return an int if you want the exact pulse period, etc.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    assertEqual(0, pushMailbox(MailboxEntry().type(MailboxEntry::Type::RESET_PENDING), 30000));
+    system_tick_t start = millis();
+    SystemSleepConfiguration config;
+    config.mode(SystemSleepMode::ULTRA_LOW_POWER)
+        .duration(60s)
+        .network(NETWORK_INTERFACE_CELLULAR);
+    SystemSleepResult result = System.sleep(config);
+
+    // in sleep for 60s, should wake up after 20s due to function call
```
scott-brust, line 204: The function call should come from the test-runners `before()` function right? 
It looks like 30s after every test start the cloud side will call the test function on the device, which should wake it in this isntance

scott-brust (commented):
Tests passed for me when I ran them locally on a boron:
```
device-os-test --test-dir ~/develop/device-os/user/tests/integration --device-os-dir ~/develop/device-os run wiring/no_fixture_power_saving boron -v -- -fixture
Test directory: /Users/scottbrust/develop/device-os/user/tests/integration
Device OS directory: /Users/scottbrust/develop/device-os
Firmware binaries directory: /Users/scottbrust/develop/device-os/build/integration
Generating test matrix
Initializing API client
URL: https://api.particle.io
Authenticating with access token
Signed in as device-os-ci@particle.io
Subscribing to device events
Initializing device manager
Enumerating USB devices
Retrieving devices from the cloud
Device pool:
e00fce6801bbf404f387daf9 (scott-us-brn404x-dvt3-mxic-1)
No fixtures configured
Running tests


  No fixture power saving
    boron
    Target device: e00fce6801bbf404f387daf9 (scott-us-brn404x-dvt3-mxic-1)
    Flashing application: no_fixture_power_saving.bin
    Getting device tests
      systemThread=disabled
        Initializing device test suite
        Running device test: POWER_SAVING_00_setup
        Device test passed
        ✓ POWER_SAVING_00_setup (7.1s)
        Running device test: POWER_SAVING_01_particle_publish_publishes_an_event_after_low_power_active
        Device test passed
        Waiting cloud event: my_event_low_power
        ✓ POWER_SAVING_01_particle_publish_publishes_an_event_after_low_power_active (44.2s)
        Running device test: POWER_SAVING_02_register_function_and_connect_to_cloud
        Device test passed
        ✓ POWER_SAVING_02_register_function_and_connect_to_cloud (37.6s)
        Running device test: POWER_SAVING_03_call_function_and_check_return_value_after_low_power_active
        Device test passed
        ✓ POWER_SAVING_03_call_function_and_check_return_value_after_low_power_active (1.5s)
        Running device test: POWER_SAVING_04_check_function_argument_value
        Device test passed
        ✓ POWER_SAVING_04_check_function_argument_value (5.9s)
        Running device test: POWER_SAVING_05_system_sleep_with_configuration_object_ultra_low_power_mode_wake_by_network
        Device test notified about expected reset
        Device test passed
        ✓ POWER_SAVING_05_system_sleep_with_configuration_object_ultra_low_power_mode_wake_by_network (85.4s)
        Running device test: POWER_SAVING_06_check_function_argument_value
        Device test passed
        ✓ POWER_SAVING_06_check_function_argument_value (5.9s)
        Running device test: POWER_SAVING_99_cleanup
        Device test passed
        ✓ POWER_SAVING_99_cleanup
      systemThread=enabled
        Initializing device test suite
        Running device test: POWER_SAVING_00_setup
        Device test passed
        ✓ POWER_SAVING_00_setup (7.2s)
        Running device test: POWER_SAVING_01_particle_publish_publishes_an_event_after_low_power_active
        Device test passed
        Waiting cloud event: my_event_low_power
        ✓ POWER_SAVING_01_particle_publish_publishes_an_event_after_low_power_active (75s)
        Running device test: POWER_SAVING_02_register_function_and_connect_to_cloud
        Device test passed
        ✓ POWER_SAVING_02_register_function_and_connect_to_cloud (36.6s)
        Running device test: POWER_SAVING_03_call_function_and_check_return_value_after_low_power_active
        Device test passed
        ✓ POWER_SAVING_03_call_function_and_check_return_value_after_low_power_active (2.4s)
        Running device test: POWER_SAVING_04_check_function_argument_value
        Device test passed
        ✓ POWER_SAVING_04_check_function_argument_value (5.9s)
        Running device test: POWER_SAVING_05_system_sleep_with_configuration_object_ultra_low_power_mode_wake_by_network
        Device test notified about expected reset
        Device test passed
        ✓ POWER_SAVING_05_system_sleep_with_configuration_object_ultra_low_power_mode_wake_by_network (85.5s)
        Running device test: POWER_SAVING_06_check_function_argument_value
        Device test passed
        ✓ POWER_SAVING_06_check_function_argument_value (5.9s)
        Running device test: POWER_SAVING_99_cleanup
        Device test passed
        ✓ POWER_SAVING_99_cleanup
    Resetting device


  16 passing (7m)
```

I also saw the device go into power save mode with the Otii power measurements
Screen Shot 2023-07-28 at 1 24 40 PM

How did you observe the `CTS1` pin when in power save mode? Did you do testing on an EVK? I think the only board of ours that I might be able to probe is an esomx...

scott-brust:
I probed an ESOMX and can see the CTS pulses in power save mode 👍 
Screen Shot 2023-07-28 at 4 25 22 PM

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1702,7 +1715,11 @@ int SaraNcpClient::initMuxer() {
     // Initialize muxer
     muxer_.setStream(serial_.get());
     muxer_.setMaxFrameSize(UBLOX_NCP_MAX_MUXER_FRAME_SIZE);
-    muxer_.setKeepAlivePeriod(UBLOX_NCP_KEEPALIVE_PERIOD);
+    if (ncpId() == PLATFORM_NCP_SARA_R510 && connState_ == NcpConnectionState::DISCONNECTED) {
+        muxer_.setKeepAlivePeriod(UBLOX_NCP_KEEPALIVE_PERIOD_R510);
```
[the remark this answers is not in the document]
technobly: This ensures the keep alive is disabled on init in general.  If you just turn on cellular, but do not connect, it will be DISCONNECTED.  In that case we still want it to drop into low power mode.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Correct, normally low when modem UART is active.  High normally in low power mode, with a short pulse every 1.28s.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: The idea behind adding it to the loop, was if we become disconnected while waiting for the low power state, it's less deterministic.  If it stays connected the whole time, there won't be an issue calling these again.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Doesn't matter, all of these get checked when the loop times out or exits early due to the correct state.

user/tests/wiring/no_fixture_power_saving/no_fixture_power_saving.spec.js
technobly: This sets up the device to have a Particle function, and ensures the device is in low power mode.  Then the JS side runs and checks that the functions are registered with the cloud (describe sent).  The real test of low power mode with functions is then the next test case, when the JS side sends a Function call.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Correct.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: That's what measureAvgPeriodMs() is for.  I thought about refactoring more, but there is an assertTrue in there and wanted to keep that visible in the test.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Right, it was 20s and I increased it to 30s.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1780,6 +1797,35 @@ int SaraNcpClient::checkSimReadiness(bool checkForRfReset) {
...
+    if (check) {
+        if ((int)upsv == CHECK(getPowerSavingValue())) {
+            // Already in required state
+            return SYSTEM_ERROR_NONE;
+        }
+    }
+
+    int r = SYSTEM_ERROR_UNKNOWN;
+
+    r = parser_.execCommand(1000, "AT+UPSV=%d",(int)upsv);
```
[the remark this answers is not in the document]
technobly: > How did you observe the CTS1 pin when in power save mode? Did you do testing on an EVK? I think the only board of ours that I might be able to probe is an esomx...

I soldered some tiny tiny wires to the bottom of a Boron :D

> This would be muxer_.setKeepAlivePeriod(0) right? I think it is probably worth doing. Its not clear to me that muxer_.suspendChannel() will stop processTimeouts() from being called and sending a keep alive message.

suspendChannel does stop all keepAlives, and sleepNoPPPWrite_ also ensures they are blocked.  I've tried to keep the changing of keepalives in to places, init and on state change.  That way they are not called repeatedly, and ensures they are called when needed and only when needed.

> We are relying on the default Timeout value of 9.2s for the module to sleep right?

Correct.  We talked about decreasing it, but it seems to work well enough at the default value.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Updated the comment.

user/tests/wiring/no_fixture_power_saving/power_saving_mode.cpp
technobly: Refactored it a bit.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: I tested `muxer_.suspendChannel()` and it in fact does not stop muxer keepAlives on the AT channel.  So even for R410, this would prevent it from staying asleep while trying to wake on network activity.  I added some code to force this to be disabled when we disable urcs (presumed for sleep usage).

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -19,7 +19,6 @@
 
 #include "logging.h"
 #define WIZNETIF_CONFIG_LOG_CATEGORY "net.en.cfg"
-LOG_SOURCE_CATEGORY(WIZNETIF_CONFIG_LOG_CATEGORY)
```
avtolstoy, line 22: Stray change?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2069,8 +2123,18 @@ int SaraNcpClient::enterDataMode() {
...
     // There is some kind of a bug in 02.19 R410 modem firmware in where it does not accept
     // any AT commands over the second muxed channel, unless we send something over channel 1 first.
-    if (ncpId() == PLATFORM_NCP_SARA_R410) {
-        CHECK(waitAtResponse(parser_, 5000));
+    //
+    // If R510 and CONNECTED already, in the case where we are in low power mode and the keepalives
+    // may be diabled/longer, just in case we are trying to resume a broken connection, let's timeout
+    // faster with an AT/OK check prior to the following AT+CGATT? call that takes 90s to timeout.
+    // This will also skip a follow up AT+CPWROFF call because ready_ will be set to false.
+    if ((ncpId() == PLATFORM_NCP_SARA_R410) ||
```
avtolstoy, line 2131: @technobly Can we leave `waitAtResponse()` for R410 as it was before just to reduce the set of affected platforms? There is a reason why it is that way (see comment above).

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2182,14 +2246,32 @@ int SaraNcpClient::getMtu() {
...
+        // R510 may be in low power mode, wake it up so we can get our muxer data channel suspend
+        // echo, before we get into sleep and potentially prematurely wake by "network activity" (RX data)
+        if (ncpId() == PLATFORM_NCP_SARA_R510) {
+            CHECK(waitAtResponse(5000, 2000));
+        }
         CHECK_TRUE(muxer_.suspendChannel(UBLOX_NCP_AT_CHANNEL) == 0, SYSTEM_ERROR_INTERNAL);
+
+        // muxer_.suspendChannel(UBLOX_NCP_AT_CHANNEL) does not block muxer AT channel keepalives. All modems will
+        // require this to be disabled to wake on network activity, and not prematurely wake on the echo of the keepalive.
+        muxer_.setKeepAlivePeriod(UBLOX_NCP_KEEPALIVE_PERIOD_R510);
```
avtolstoy, line 2273: Nit: this is a bit confusing, as doesn't apply to just R510, might be worth just using `0` here`.

avtolstoy (changes requested):
Just the R410 `waitResponse()` thing is a bit worrying.

hal/network/lwip/wiznet/wiznetif_config.cpp
technobly: This is required for `DEBUG_BUILD=y`

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: If we leave that, and the AT interface dies... it will take a lot longer for the system to recover.  As far as I can tell from the previous code, it's just sending an AT/OK check on channel 1.  So this does the same thing, but also sets parserError_ in case of a timeout.

hal/network/ncp_client/sara/sara_ncp_client.cpp
technobly: Sure, I'll add a `UBLOX_NCP_KEEPALIVE_PERIOD_DISABLED` constant.

hal/network/ncp_client/sara/sara_ncp_client.cpp
avtolstoy: I'm more concerned with R410. There is a separate reason we issue multiple AT/AT/AT requests here, so we can leave that as-is and just have execCommand for R510, right?

technobly: Ok, from the comment, I didn't get that it requires multiple AT/AT/AT before OK on channel 1.  I believe that would still happen here, but I'll leave it the way it was for now just to not disturb it.

avtolstoy (approved):

PR #2677, opened 2023-08-01
rtl872x: set RXD sample delay for SPI0@50MHz

YutingYou:
### Problem

For the external flash read/write test, SPI0 and SPI1 passed test at 25MHz. SPI0 failed to communicate with the flash at the highest speed 50MHz.

Writting the regular data `0x55 0x55 0x55 ...` to the flash, which is 01010101 waveform on the oscilloscope, we read back the data `0xaa 0xaa ... 0xaa 0x1a 0xaa...`, this anomaly seems to have a pattern, the data appears to have shifted by one bit.

### Solution

The data corruption issue can be fixed by setting the sample delay to one clock cycle.

Please refer to < UM0400 Ameba-D User Manual > 25.2.3.1.1 RXD Sample Delay

Screenshot 2023-08-01 at 14 11 36

Screenshot 2023-08-01 at 14 11 41

### References

[CH119867]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/spi_hal.cpp
+++ b/hal/src/rtl872x/spi_hal.cpp
@@ -322,6 +322,13 @@ public:
             u32 rtlClockDivider = 256;
             clockDivToRtlClockDiv(config_.clockDiv, &rtlClockDivider);
             SSI_SetBaudDiv(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, rtlClockDivider);
+
+            // Set sample delay for SPI0@50MHz
+            if (rtlClockDivider == 2 && SPI_DEV_TABLE[rtlSpiIndex_].SPIx == SPI0_DEV) {
+                SSI_SetSampleDelay(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, 0x1);
+            } else {
+                SSI_SetSampleDelay(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, 0x0);
+            }
         }
 
         // Set bit order


hal/src/rtl872x/spi_hal.cpp
```diff
@@ -322,6 +322,13 @@ class Spi {
             u32 rtlClockDivider = 256;
             clockDivToRtlClockDiv(config_.clockDiv, &rtlClockDivider);
             SSI_SetBaudDiv(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, rtlClockDivider);
+
+            // Set sample delay for SPI0@50MHz
+            if (rtlClockDivider == 2 && SPI_DEV_TABLE[rtlSpiIndex_].SPIx == SPI0_DEV) {
+                SSI_SetSampleDelay(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, 0x1);
+            } else {
+                SSI_SetSampleDelay(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, 0x0);
```
XuGuohui, line 330: Is it working fine with other `rtlClockDivider`?

XuGuohui (approved):

hal/src/rtl872x/spi_hal.cpp
YutingYou: This issue only happens when using the high speed SPI, in case of affecting low speed SPI, we only set it for SPI0@50MHz (`rtlClockDivider ==2`).

XuGuohui: How do you know the sample delay being `0` is working with other low speed SPI? This is also the new introduced change to the low speed SPI.

YutingYou: Because `0x0` is the default value
Screenshot 2023-08-01 at 15 18 39

scott-brust:
These changes do not affect the onboard SPI flash, correct? These are only for the external SPI peripherals?

YutingYou:
@scott-brust It doesn't affect the onboard SPI flash, because the onboard SPI flash is using the QSPI bus while this test is for the SPI bus.

PR #2678, opened 2023-08-02
Combine PA[20] and VBAT_MEAS pins. Update names with orson v0.3 pinmap

scott-brust:
### Problem

MSoM EVT `A7`/`VBAT_MEAS` is only usable as ADC, not as digital IO

### Solution

MSoM DVT should connect both the `VBAT_MEAS` and `PA20` pins to form a more functional `A7`/`D28` pin that is capable of both ADC and digital IO

### Steps to Test

On EVT MSOM with the EVB, jumper together A7 and SOM11 to form the combined A7/VBAT_MEAS pin

### Example App

```c
#include "Particle.h"

Serial1LogHandler logHandler1(115200, LOG_LEVEL_ALL, 
{
    {
        "app", LOG_LEVEL_TRACE }, // Default logging level for all application messages
    }
);
SYSTEM_MODE(AUTOMATIC);
SYSTEM_THREAD(ENABLED);

void setup() {
}

void loop() {
    if (Serial1.available() > 0) {
        auto commandByte = Serial1.read();

        // All names for the same pin, just to make sure they all work. D26 for comparison with "normal" io
        const int pinAliases[] = {A7, D28, VBAT_MEAS, D26};
        const char * pinNames[] = {"A7", "D28", "VBAT_MEAS", "D26"};

        static int pinIndex = 0;
        static int pin = pinAliases[pinIndex];
        static const char * pinName = pinNames[0];

        switch ((char)commandByte) {
            case '1':
            {
                if (++pinIndex > 3) {
                    pinIndex = 0;
                }
                pin = pinAliases[pinIndex];
                pinName = pinNames[pinIndex];
                Log.info("Using pin number %d name %s", pin, pinName);
                break;
            }
            case '2':
            {
                int val = analogRead(pin);
                Log.info("ADC read on pin %s result %d", pinName, val);
                break;
            }
            case '3':
            {
                static bool outputHigh = true;
                Log.info("Digital write on pin %s state %d", pinName, outputHigh);

                pinMode(pin, OUTPUT);
                digitalWrite(pin, outputHigh ? HIGH : LOW);
                outputHigh = !outputHigh;
                break;
            }
            case '4':
            {
                int val = digitalRead(pin);
                Log.info("Digital read on pin %s result: %d", pinName, val);
                break;
            }
            default:
                break;
        }
    }
}
```

### References

See discussion [here](https://s.slack.com/archives/C046BM6DBLL/p1690364003547589)

---

### Completeness

- [ ] Added documentation


hal/src/msom/pinmap_defines.h
```diff
@@ -61,7 +61,7 @@
 #define A4                  D15
 #define A5                  D14
 #define A6                  D29
-#define A7                  VBAT_MEAS
```
scott-brust, line 64: Right now `A7` `D28` `WKP` and `VBAT_MEAS` are all the same logical pin. Should we get rid of the `VBAT_MEAS` define altogether? It should not be used by name?

eberseth: Probably only used to refer to manufacturer's pinout.  We can refer to it as A7 Internally.

XuGuohui (commented):
@YutingYou Should we make some changes to the gpio hal to make sure the PA20 is configured properly when A7 is used as analog input?

eberseth (approved):

eberseth:
> 

Yes

YutingYou:
@XuGuohui Yes, remember to configure `PA20` to `PIN_MODE_NONE` to ensure it is in HIGH-Z mode when we call `analogRead()`.

hal/src/msom/pinmap_defines.cpp
```diff
@@ -69,7 +69,7 @@ static hal_pin_info_t pinmap[TOTAL_PINS] = {
 /* BGVINT        - 40 */ { RTL_PORT_B,     31, PIN_MODE_NONE, PF_NONE, ADC_CHANNEL_NONE, PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 0},
 /* BGDTR         - 41 */ { RTL_PORT_B,     30, PIN_MODE_NONE, PF_NONE, ADC_CHANNEL_NONE, PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 0},
 /* GNSS_ANT_PWR  - 42 */ { RTL_PORT_A,      2, PIN_MODE_NONE, PF_NONE, ADC_CHANNEL_NONE, PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 0},
-/* VBAT_MEAS/A7  - 43 */ { RTL_PORT_NONE,   0, PIN_MODE_NONE, PF_NONE, 7,                PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 0},
+/* VBAT_MEAS/A7  - 43 */ { RTL_PORT_A,     20, PIN_MODE_NONE, PF_NONE, 7,                PWM_INSTANCE_NONE, PWM_CHANNEL_NONE, 0},
```
scott-brust, line 72: Should we go as far as to remove it from the array of pins here? (and adjust all the dependend constants etc)

PR #2680, opened 2023-08-04
rtl872x: overshoot issue

YutingYou:
### Problem

During the EVT test, we observed significant overshoot. This issue occurs on every high-speed bus, including SPI, I2C, PWM, I2S, and DMIC.

### Solution

There are four drive strength can be configured for RTL872x, all the GPIO were configured to `PAD_DRV_STRENGTH_3` at boot.
- PAD_DRV_STRENGTH_0: 4mA
- PAD_DRV_STRENGTH_1: 8mA   (Severe overshoot, not recommended to use it)
- PAD_DRV_STRENGTH_2: 12mA
- PAD_DRV_STRENGTH_3: 16mA  (Severe overshoot, not recommended to use it)

We compared the overshoot issue with each drive strength configuration and found `PAD_DRV_STRENGTH_0` and `PAD_DRV_STRENGTH_2` were better than `PAD_DRV_STRENGTH_1` and `PAD_DRV_STRENGTH_3`

Below is the clock pin of SPI0@50MHz
Screenshot 2023-08-02 at 15 19 33

Selecting PAD_DRV_STRENGTH_2 could be a solution to avoid the overshoot issue while maintaining a high drive strength.

Below, you will find a more detailed comparison between PAD_DRV_STRENGTH_2 and PAD_DRV_STRENGTH_3 across different buses on the Orson platform.

Screenshot 2023-08-04 at 16 01 42

### Steps to Test

- Compare the waveforms of each high-speed bus on P2 or Orson before and after applying the fix.

### References

[CH119867]
[CH119866]
[CH120087]
[CH119868]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/gpio_hal.h
+++ b/hal/inc/gpio_hal.h
@@ -73,6 +73,8 @@ PinMode hal_gpio_get_mode(hal_pin_t pin);
 void hal_gpio_write(hal_pin_t pin, uint8_t value);
 int32_t hal_gpio_read(hal_pin_t pin);
 uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value);
+hal_gpio_drive_t hal_gpio_get_drive_strength(hal_pin_t pin);
+int hal_gpio_configure_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive);
 
 #include "gpio_hal_compat.h"
 
--- a/hal/inc/gpio_hal.h
+++ b/hal/inc/gpio_hal.h
@@ -73,7 +73,7 @@ PinMode hal_gpio_get_mode(hal_pin_t pin);
 void hal_gpio_write(hal_pin_t pin, uint8_t value);
 int32_t hal_gpio_read(hal_pin_t pin);
 uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value);
-hal_gpio_drive_t hal_gpio_get_drive_strength(hal_pin_t pin);
+int hal_gpio_get_drive_strength(hal_pin_t pin, hal_gpio_drive_t* drive);
 int hal_gpio_configure_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive);
 
 #include "gpio_hal_compat.h"
--- a/hal/inc/gpio_hal.h
+++ b/hal/inc/gpio_hal.h
@@ -74,7 +74,7 @@ void hal_gpio_write(hal_pin_t pin, uint8_t value);
 int32_t hal_gpio_read(hal_pin_t pin);
 uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value);
 int hal_gpio_get_drive_strength(hal_pin_t pin, hal_gpio_drive_t* drive);
-int hal_gpio_configure_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive);
+int hal_gpio_set_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive);
 
 #include "gpio_hal_compat.h"
 
--- a/hal/src/rtl872x/gpio_hal.cpp
+++ b/hal/src/rtl872x/gpio_hal.cpp
@@ -159,6 +159,7 @@ int hal_gpio_configure(hal_pin_t pin, const hal_gpio_config_t* conf, void* reser
         }
 
         Pinmux_Config(hal_pin_to_rtl_pin(pin), PINMUX_FUNCTION_GPIO);
+        hal_gpio_configure_drive_strength(pin, HAL_GPIO_DRIVE_DEFAULT);
 
         // Pre-set the output value if requested to avoid a glitch
         if (conf->set_value && (mode == OUTPUT || mode == OUTPUT_OPEN_DRAIN || mode == OUTPUT_OPEN_DRAIN_PULLUP)) {
@@ -434,7 +435,7 @@ uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value) {
     return (hal_timer_micros(nullptr) - pulse_start);
 }
 
-hal_gpio_drive_t hal_gpio_get_drive_strength(hal_pin_t pin) {
+int hal_gpio_get_drive_strength(hal_pin_t pin, hal_gpio_drive_t* drive) {
     uint32_t rtlPin = hal_pin_to_rtl_pin(pin);
 
     /* get PADCTR */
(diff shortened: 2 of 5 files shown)


hal/inc/gpio_hal.h
```diff
@@ -73,6 +73,8 @@ PinMode hal_gpio_get_mode(hal_pin_t pin);
 void hal_gpio_write(hal_pin_t pin, uint8_t value);
 int32_t hal_gpio_read(hal_pin_t pin);
 uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value);
+hal_gpio_drive_t hal_gpio_get_drive_strength(hal_pin_t pin);
```
avtolstoy, line 76: Nit: If we are later to expose this in dynalib might perhaps be better to do `int hal_gpio_get_drive_strength(hal_pin_t hal_gpio_drive_t* drive)` to catch errors (e.g. wrong pin, input only pin etc).

hal/src/rtl872x/gpio_hal.cpp
```diff
@@ -433,3 +433,65 @@ uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value) {
...
+        // DVOS won't configure PAD_DRV_STRENGTH_1 and PAD_DRV_STRENGTH_3 due to the
+        // severe overshoot, so it is the default setting
+        default:
+            drive = HAL_GPIO_DRIVE_DEFAULT; break;
+    }
+
+    return drive;
+}
+
+int hal_gpio_configure_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive) {
```
avtolstoy, line 465: `hal_gpio_config_t::drive_strength` should also be handled in `hal_gpio_configure`, right?

hal/inc/gpio_hal.h
YutingYou: Fixed.

hal/src/rtl872x/gpio_hal.cpp
YutingYou: Yes, set to `HAL_GPIO_DRIVE_DEFAULT ` in the `hal_gpio_configure `

avtolstoy: Hm. What I meant is that it should be called in `hal_gpio_configure` with that field as argument, so ```hal_gpio_configure(pin, config->drive_strength)```

hal/src/rtl872x/gpio_hal.cpp
```diff
@@ -433,3 +434,64 @@ uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value) {
...
+
+    // Pad driving strength
+    //   - PAD_DRV_STRENGTH_0: 4mA
+    //   - PAD_DRV_STRENGTH_1: 8mA   (Severe overshoot, not recommended to use it)
+    //   - PAD_DRV_STRENGTH_2: 12mA
+    //   - PAD_DRV_STRENGTH_3: 16mA  (Severe overshoot, not recommended to use it)
+    switch(drive) {
+        case HAL_GPIO_DRIVE_STANDARD: drvStrength = PAD_DRV_STRENGTH_0; break;
+        case HAL_GPIO_DRIVE_HIGH:     drvStrength = PAD_DRV_STRENGTH_2; break;
+        case HAL_GPIO_DRIVE_DEFAULT:  drvStrength = PAD_DRV_STRENGTH_3; break;
```
scott-brust, line 477: Using `HAL_GPIO_DRIVE_DEFAULT` will set `PAD_DRV_STRENGTH_3`? I thought we did not want to use this mode? 
In your opinion, what should the default drive strength be? `PAD_DRV_STRENGTH_0`?

hal/src/rtl872x/gpio_hal.cpp
```diff
@@ -433,3 +434,64 @@ uint32_t hal_gpio_pulse_in(hal_pin_t pin, uint16_t value) {
...
+        // DVOS won't configure PAD_DRV_STRENGTH_1 and PAD_DRV_STRENGTH_3 due to the
+        // severe overshoot, so it is the default setting
+        default:
+            *drive = HAL_GPIO_DRIVE_DEFAULT; break;
+    }
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int hal_gpio_configure_drive_strength(hal_pin_t pin, hal_gpio_drive_t drive) {
```
scott-brust, line 465: Would it be clearer if this function was renamed `hal_gpio_set_drive_strength()` to match `hal_gpio_get_drive_strength()`?

hal/src/rtl872x/gpio_hal.cpp
```diff
@@ -159,6 +159,7 @@ int hal_gpio_configure(hal_pin_t pin, const hal_gpio_config_t* conf, void* reser
         }
 
         Pinmux_Config(hal_pin_to_rtl_pin(pin), PINMUX_FUNCTION_GPIO);
+        hal_gpio_configure_drive_strength(pin, HAL_GPIO_DRIVE_DEFAULT);
```
scott-brust, line 162: The `hal_gpio_config_t* conf` parameter here has a field for `drive_strength`. Should we use that field instead of the default? 
(I believe the NRF MCU had GPIO settings for drive strength as well, so perhaps we can re-use this field for the RTK platform now)

See `hal_gpio_mode()` for the config struct, and perhaps we initialize `conf.drive_strength = HAL_GPIO_DRIVE_DEFAULT` there?

This should also implement `pinSetDriveStrength()` wiring API function as well.

hal/src/rtl872x/gpio_hal.cpp
YutingYou: I misunderstood that, I'll rework this API

hal/src/rtl872x/gpio_hal.cpp
YutingYou: @scott-brust  Realtek defaults to using PAD_DRV_STRENGTH_3 as the default drive strength, as indicated in the log. The overshoot issue occurs on high-speed buses, and it has been resolved in each peripheral driver. However, considering the possibility of introducing other side effects (which might be an unnecessary concern), I have decided to set the default drive strength to `PAD_DRV_STRENGTH_3`.
```
hal_gpio_get_drive_strength, pin: D0, drvStrength: 3
hal_gpio_get_drive_strength, pin: D1, drvStrength: 3
hal_gpio_get_drive_strength, pin: D2, drvStrength: 3
...
hal_gpio_get_drive_strength, pin: D18, drvStrength: 3
hal_gpio_get_drive_strength, pin: D19, drvStrength: 3
hal_gpio_get_drive_strength, pin: D20, drvStrength: 3
```

hal/src/rtl872x/gpio_hal.cpp
YutingYou: To be honest, I'd also prefer `hal_gpio_set_drive_strength()`, I intended to keep it consistent with `hal_gpio_configure`, well, let's rename it.

hal/src/rtl872x/gpio_hal.cpp
scott-brust: I see, its probably the right decision to change drive strength only for peripherals you validate. Makes sense to leave the default in other cases to the current setting.

hal/src/rtl872x/gpio_hal.cpp
YutingYou: 👍🏻 Fixed, the `pinSetDriveStrength()` function is already available for the Realtek platform.

scott-brust (approved):

PR #2685, opened 2023-08-15
Gen3: fix uart rx data corruption

XuGuohui:
### Problem

The current mechanism for UART data receiving is depending on the `NRF_UARTE_EVENT_ENDRX` event and a timer counter that is connected with the `NRF_UARTE_EVENT_RXDRDY` event through PPI. When the `NRF_UARTE_EVENT_ENDRX` is asserted, the receiver is stopped automatically by hardware. Everytime we service the `NRF_UARTE_EVENT_ENDRX` we restart the UART receiver. Only if the receiver is started, the data we received can be moved into the RAM buffer. If we query the data when UART is receiving data, the timer can tell us how many bytes are received. But note that when `NRF_UARTE_EVENT_RXDRDY` is generated, it doesn't mean that the received byte has been moved to the RAM buffer.

Screenshot 2023-08-15 at 15 40 53

The problem of the current mechanism is if the `NRF_UARTE_EVENT_ENDRX` event is postponed to deal with by, for example, other interrupts with higher priority, while UART is receiving data, then the `NRF_UARTE_EVENT_RXDRDY` is asserted, but the receiver has stopped and the received byte cannot be moved to the RAM buffer, and we lost the byte. The longer the `NRF_UARTE_EVENT_ENDRX` being postponed, the more data will lose. This happens at the middle of a data receiving frame when the `NRF_UARTE_EVENT_ENDRX` is asserted, so using a larger buffer doesn't fix the issue.

-> incoming data stream:  00 01 02 03 04 05 06 07 08 09 ...
-> 1. rx buffer  [ (free spaces) | (acquired length = 5) 00 01 02 03 04 ] <- NRF_UARTE_EVENT_ENDRX, 05 06 07 ... may be lost
-> 2. rx buffer  [ (free spaces) | (acquired length = 10000,  commited = 9995) 00 01 02 03 04 ] <- NRF_UARTE_EVENT_ENDRX, 05 06 07 ... may be lost

The above is the typical circumstance how data is lost. Sometimes we don't lose data, but we may break a data frame. For example, Normally the `NRF_UARTE_EVENT_ENDRX` is generated before the `NRF_UARTE_EVENT_RXDRDY` of the byte `N` as shown in the attached screenshot. If the `NRF_UARTE_EVENT_ENDRX` is not services in time, the `NRF_UARTE_EVENT_RXDRDY` will increase the timer counter by 1, followed by servicing the `NRF_UARTE_EVENT_ENDRX` event, in which case the timer counter is cleared. In this case, we did received the byte N and the byte `N` can be moved to the RAM buffer once we restart the receiver in ISR, but since the timer counter is cleared, we will start counting the received data with byte `N+1`.

1. Assume we are going to receive contiguous 43 bytes
2. When the 42th byte is received, the `NRF_UARTE_EVENT_ENDRX` is serviced and timer counter is cleared, but the 43th byte is ready to be moved to the RAM buffer. Since user keep querying and read the received data, `available()` is now `0`.
3. Assume the 2nd transaction will acquire `10` bytes from the RX ring buffer.
4. Receiver restarted and the 43th byte is moved to the RAM buffer, but `available()` is still `0`. The first 43-bytes transaction timeout
5. 2nd 43-bytes is receiving. When the next `NRF_UARTE_EVENT_ENDRX` is asserted, there is `10` bytes in the ring buffer, but the timer counter is `9`, since the timer start counting when the first byte of the 2nd transaction is received. Now the `available()` is `10`, since we `rxBuffer_.acquireCommit(rxBuffer_.acquirePending());`(pending is 1) when we handle the `NRF_UARTE_EVENT_ENDRX` event.
6. Assume we will acquire `40` bytes from the RX ring buffer. Now there is still 34 bytes to be received for the 2nd transaction.
7. The timer counter increases by 34. The total `available()` since step5 is 44.
8. So the 1st transaction we received 42 bytes and the 2nd transaction we receved 44 bytes. 

### Solution

1. Use double buffering and enable `NRF_UARTE_SHORT_ENDRX_STARTRX` shortcut and the `NRF_UARTE_INT_RXSTARTED_MASK`, by doing this the receiver will automatically started when it detected the `NRF_UARTE_EVENT_ENDRX` and we prepare the next buffer for the next transaction when the current receiver started.

2. Do not clear the timer counter. The timer is always counting and there is a variable for tracking how many bytes are commited. It doesn't matter even if the timer overflows. We can easily handle this situation by comparing the counter value and the variable.

### Step To Test

1. Flash the below examples to two gen3 devices.
2. Connect these two devices over the Serial1 interface.
3. Connect the USB serial to start monitoring the test result.

### Example App

Argon (receiver):
```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

// set to serial port configuration for test
#define SERIAL (Serial1)
#define SERIAL_BAUD (115200)
#define SERIAL_TX_INTERVAL_MS (20)
#define SERIAL_RX_TIMEOUT_MS (50)

// comment in/out to test with or without a thread scanning for BLE adverts
#define FEATURE_BLE

// for Jacuzzi this is a valid a valid modbus request for 19 registers starting
// at address 3000, can be modified for alternate test cases
uint8_t TX_BUF[] = "12345678";
constexpr size_t TX_LEN = sizeof(TX_BUF) - 1; // -1 to skip null terminator

constexpr size_t RX_EXPECT_LEN = 43; // expected length of modbus response
constexpr uint8_t testBuff[RX_EXPECT_LEN] = {
    0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x10,
    0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x20,
    0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x28, 0x29, 0x30,
    0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40,
    0x41, 0x42, 0x43
};

#ifdef FEATURE_BLE

static Thread *thread = nullptr;

void _ble_scan_result_cb(const BleScanResult *result, void *context) {
    return;
}

void _ble_scan_thread_f() {
    // short repeated scans
    BLE.setScanTimeout(10);

    // disable active scan
    // only get advert, don't request scan response for less overhead and
    // hopefully reduced power consumption on sensors
    BleScanParams params = {};
    params.version = BLE_API_VERSION;
    params.size = sizeof(BleScanParams);
    BLE.getScanParameters(params);
    params.active = false;
    BLE.setScanParameters(params);

    while(true) {
        static system_tick_t last = millis();
        if ((millis() - last) > 5000) {
            Log.info("BLE Scan thread.");
            last = millis();
        }
        BLE.scan(_ble_scan_result_cb, nullptr);
    }
}

#endif

void hex_dump(LogLevel level, uint8_t *data, int len, const char *prefix = nullptr) {
    if(prefix) {
        Log.print(level, prefix);
    }
	// Log.print(level, "\t");
	Log.dump(level, data, len);
	Log.print(level, "\n\r");
}

void setup() {
    delay(1s);

    SERIAL.begin(SERIAL_BAUD);

#ifdef FEATURE_BLE
    BLE.on();
    BLE.setPPCP(12, 12, 0, 200); // increase BLE speed, may break third party

    thread = new Thread("sensor_ble_scan",
        _ble_scan_thread_f,
        OS_THREAD_PRIORITY_DEFAULT);
#endif

    Particle.connect();
}

void loop() {
    static int rx_timeout = 0, buffer_overrun = 0, crc_error = 0, rx_len_bad = 0, total_tx = 0, total_rx = 0;
    static bool timeout = false;
    static size_t rx_len = 0;

    if (timeout) {
        rx_timeout++;
        timeout = false;
    }

    // check for system blocking by monitoring for long loop delays
    // if we don't get a spam of these then should be servicing the serial port
    // on a reasonable timeframe
    static system_tick_t last_loop_ms = millis();
    auto loop_delta_ms = millis() - last_loop_ms;
    if (loop_delta_ms > 10) {
        Log.info("loop delay: %lu", loop_delta_ms);
    }
    last_loop_ms = millis();

    // handle sending TX buffer when requested
    // not running test long enough to care about rollover on millis()
    static system_tick_t next_request_tx = millis();
    static auto last_tx_ms = millis();
    if (next_request_tx && millis() > next_request_tx) {
        uint8_t count = SERIAL.available();
        if (count) {
            Log.info("!!!!available before TX: %d, rx_len: %d, millis: %ld", count, rx_len, millis());
        }
        SINGLE_THREADED_BLOCK() {
            SERIAL.write(TX_BUF, TX_LEN);
            SERIAL.flush();
        }
        last_tx_ms = millis();
        next_request_tx = 0;
        total_tx++;
    }

    // it not currently sending and it has been too long then the rx must have
    // failed
    // call a timeout and request the next send
    if (!next_request_tx && (millis() - last_tx_ms > 1000)) {
        next_request_tx = millis();
        timeout = true;
        Log.error("RX-TIMEOUT millis: %ld", millis());
    }

    // receive bytes into accumulating frame
    static uint8_t rx_buf[256];
    static auto last_rx_ms = millis();
    while (SERIAL.available()) {
        if (timeout) {
            Log.error("Timeout, but received first data now. millis: %ld", millis());
            timeout = false;
        }
        int byte = SERIAL.read();
        if (byte >= 0) {
            if (rx_len == 0) {
                total_rx++;
            }
            if (rx_len < sizeof(rx_buf)) {
                rx_buf[rx_len++] = (uint8_t)byte;
            } else {
                Log.error("BUFFER-OVERRUN");
                buffer_overrun++;
                hex_dump(LOG_LEVEL_ERROR, rx_buf, rx_len, "RX: ");
                rx_len = 0;
            }
        }
        last_rx_ms = millis();
    }
    // terminate frame after timeout from last received byte
    // validate expected length and crc
    auto time_since_last_rx = millis() - last_rx_ms;
    if (rx_len && time_since_last_rx > SERIAL_RX_TIMEOUT_MS) {
        bool error = false;
        if (rx_len != RX_EXPECT_LEN) {
            Log.info("RX-LEN-ERROR=%u (%lu ms)", rx_len, time_since_last_rx);
            error = true;
            rx_len_bad++;
        } else {
            if (memcmp(testBuff, rx_buf, rx_len)) {
                crc_error++;
                error = true;
            } 
        }
        if (error) {
            Log.info("total=%d, rx_total=%d, timeout=%d, overrun=%d, rx_len=%d, crc=%d", total_tx, total_rx, rx_timeout, buffer_overrun, rx_len_bad, crc_error);
            hex_dump(LOG_LEVEL_INFO, rx_buf, rx_len, "RX: ");
            if (rx_len_bad || crc_error) {
                delay(1000);
                SPARK_ASSERT(false);
            }
        }

        rx_len = 0;
        // schedule the next request after delay
        next_request_tx = millis() + SERIAL_TX_INTERVAL_MS;
    }

    auto now_sec = System.uptime();
    static auto last_sec = now_sec;
    if ((now_sec - last_sec) >= 3) {
        Log.info("total=%d, timeout=%d, overrun=%d, rx_len=%d, crc=%d, total_rx=%d", total_tx, rx_timeout, buffer_overrun, rx_len_bad, crc_error, total_rx);
        last_sec = now_sec;
    }
}
```

Boron:

```c
#include "Particle.h"

SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

#define SERIAL (Serial1)
#define SERIAL_BAUD (115200)

const char TX_BUF[] = "12345678";

constexpr uint8_t testBuff[43] = {
    0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x10,
    0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19, 0x20,
    0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x28, 0x29, 0x30,
    0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x40,
    0x41, 0x42, 0x43
};

void setup() {
    SERIAL.begin(SERIAL_BAUD);
}

int rxCount = 0;
int lastCount = 0;

uint16_t availableForWrite = 0;

void loop() {
    static uint8_t rxBuf[16];
    static uint8_t rxLen = 0;
    
    if (SERIAL.available()) {
        while (SERIAL.available()) {
            uint8_t temp = SERIAL.read();
            if (rxLen < strlen(TX_BUF)) {
                rxBuf[rxLen++] = temp;
            }
        }
    }
    if (rxLen == strlen(TX_BUF)) {
        if (!memcmp(rxBuf, TX_BUF, rxLen)) {
            // Log.info("Received TX request. Send test buffer.");
            availableForWrite = SERIAL.availableForWrite();
            SINGLE_THREADED_BLOCK() {
                SERIAL.write(testBuff, sizeof(testBuff));
                SERIAL.flush();
            }
            rxCount++;
        }
        memset(rxBuf, 0x00, sizeof(rxBuf));
        rxLen = 0;
    }

    auto now_sec = System.uptime();
    static auto last_sec = now_sec;
    if ((now_sec - last_sec) >= 3) {
        last_sec = now_sec;
        if (rxCount != lastCount) {
            lastCount = rxCount;
        } else {
            Log.info("rxCount=%d, availableForWrite: %d", rxCount, availableForWrite);
        }
    }
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/usart_hal.cpp
+++ b/hal/src/nRF52840/usart_hal.cpp
@@ -240,7 +240,7 @@ public:
 
         state_ = HAL_USART_STATE_DISABLED;
         transmitting_ = false;
-        receiving_ = 0;
+        usingAltRxBuffer_ = false;
         config_ = {};
         rxBuffer_.reset();
         rxBufferAlt_.reset();
@@ -262,8 +262,8 @@ public:
 
         state_ = HAL_USART_STATE_SUSPENDED;
         transmitting_ = false;
-        receiving_ = 0;
         rxBuffer_.prune();
+        rxBufferAlt_.prune();
         txBuffer_.reset();
 
         return SYSTEM_ERROR_NONE;
@@ -278,16 +278,10 @@ public:
     ssize_t data() {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         RxLock lk(uarte_);
-        particle::services::RingBuffer<uint8_t>* bufferCurrent = nullptr;
-        particle::services::RingBuffer<uint8_t>* bufferPre = nullptr;
-        if (usingAltRxBuffer_) {
-            bufferCurrent = &rxBufferAlt_;
-            bufferPre = &rxBuffer_;
-        } else {
-            bufferCurrent = &rxBuffer_;
-            bufferPre = &rxBufferAlt_;
-        }
-        ssize_t d = bufferPre->data() + bufferCurrent->data();
+        particle::services::RingBuffer<uint8_t>* currRxBuff = nullptr;
+        particle::services::RingBuffer<uint8_t>* preRxBuff = nullptr;
+        acquireRxBuffer(&currRxBuff, &preRxBuff);
+        ssize_t d = preRxBuff->data() + currRxBuff->data();
         if (receiving_) {
             // IMPORTANT: It seems that the timer value (which is a counter for every time UARTE generates RXDRDY event)
             // may potentially get larger than the configured RX DMA transfer:
@@ -297,16 +291,16 @@ public:
             // </quote>
             // We'll be extra careful here and make sure not to consume more than we can, otherwise
             // we may put the ring buffer into an invalid state as there are not safety checks.
-            size_t timerCount = timerValue();
-            size_t toConsumeCount;
+            uint16_t timerCount = timerValue();
+            uint16_t toConsumeCount;
             if (timerCount >= consumedTimerCount_) {
                 toConsumeCount = timerCount - consumedTimerCount_;
             } else {
                 toConsumeCount = 65536 - consumedTimerCount_ + timerCount;
             }
-            const ssize_t toConsume = std::min(bufferCurrent->acquirePending(), toConsumeCount);
+            const ssize_t toConsume = std::min(currRxBuff->acquirePending(), (size_t)toConsumeCount);
             if (toConsume > 0) {
-                bufferCurrent->acquireCommit(toConsume);
+                currRxBuff->acquireCommit(toConsume);
                 d += toConsume;
                 consumedTimerCount_ = (consumedTimerCount_ + toConsume) % 65536;
             }
@@ -326,25 +320,19 @@ public:
         const size_t readSize = std::min((size_t)maxRead, size);
         CHECK_TRUE(readSize > 0, SYSTEM_ERROR_NO_MEMORY);
         RxLock lk(uarte_);
-        particle::services::RingBuffer<uint8_t>* bufferCurrent = nullptr;
-        particle::services::RingBuffer<uint8_t>* bufferPre = nullptr;
-        if (usingAltRxBuffer_) {
-            bufferCurrent = &rxBufferAlt_;
-            bufferPre = &rxBuffer_;
-        } else {
-            bufferCurrent = &rxBuffer_;
-            bufferPre = &rxBufferAlt_;
-        }
-        ssize_t remainLen = bufferPre->data();
+        particle::services::RingBuffer<uint8_t>* currRxBuff = nullptr;
+        particle::services::RingBuffer<uint8_t>* preRxBuff = nullptr;
+        acquireRxBuffer(&currRxBuff, &preRxBuff);
+        ssize_t remainLen = preRxBuff->data();
         SCOPE_GUARD ({
             if (remainLen == 0 && !(uarte_->SHORTS & NRF_UARTE_SHORT_ENDRX_STARTRX)) {
                 prepareNextRingBuffer();
                 nrf_uarte_shorts_enable(uarte_, NRF_UARTE_SHORT_ENDRX_STARTRX);
             }
         });
-        ssize_t r = CHECK(bufferPre->get(buffer, std::min((size_t)remainLen, readSize)));
+        ssize_t r = CHECK(preRxBuff->get(buffer, std::min((size_t)remainLen, readSize)));
         if (r >= 0 && r < (ssize_t)readSize) {
-            r += CHECK(bufferCurrent->get(buffer + r, readSize - r));
+            r += CHECK(currRxBuff->get(buffer + r, readSize - r));
         }
         return r;
     }
@@ -355,7 +343,15 @@ public:
         const size_t peekSize = std::min((size_t)maxRead, size);
         CHECK_TRUE(peekSize > 0, SYSTEM_ERROR_NO_MEMORY);
         RxLock lk(uarte_);
-        return rxBuffer_.peek(buffer, peekSize);
+        particle::services::RingBuffer<uint8_t>* currRxBuff = nullptr;
+        particle::services::RingBuffer<uint8_t>* preRxBuff = nullptr;
+        acquireRxBuffer(&currRxBuff, &preRxBuff);
+        ssize_t remainLen = preRxBuff->data();
+        if (remainLen) {
+            return preRxBuff->peek(buffer, peekSize);
+        } else {
+            return currRxBuff->peek(buffer, peekSize);
+        }
     }
 
     ssize_t write(const uint8_t* buffer, size_t size) {
@@ -434,16 +430,12 @@ public:
                 nrf_uarte_event_clear(uarte_, NRF_UARTE_EVENT_ENDRX);
                 nrf_uarte_event_clear(uarte_, NRF_UARTE_EVENT_RXDRDY);
 
-                particle::services::RingBuffer<uint8_t>* buffer = nullptr;
-                if (usingAltRxBuffer_) {
-                    buffer = &rxBufferAlt_;
-                } else {
-                    buffer = &rxBuffer_;
-                }
+                particle::services::RingBuffer<uint8_t>* currRxBuff = nullptr;
+                acquireRxBuffer(&currRxBuff, nullptr);
 
-                if (buffer->acquirePending() > 0) {
-                    size_t remain = buffer->acquirePending();
-                    buffer->acquireCommit(remain);
+                if (currRxBuff->acquirePending() > 0) {
+                    size_t remain = currRxBuff->acquirePending();
+                    currRxBuff->acquireCommit(remain);
                     consumedTimerCount_ = (consumedTimerCount_ + remain) % 65536;
                 }
 
@@ -545,8 +537,6 @@ private:
         nrf_uarte_hwfc_pins_disconnect(uarte_);
         nrfx_prs_release(uarte_);
 
-        disableTimer();
-
         // Configuring the input pins' mode to PIN_MODE_NONE will disconnect input buffer to enable power savings
         // Configuring the output pins' mode to INPUT_PULLUP to not polluting the state on other side.
         hal_gpio_mode(txPin_, end ? PIN_MODE_NONE : INPUT_PULLUP);
@@ -587,24 +577,29 @@ private:
         }
     }
 
+    void acquireRxBuffer(particle::services::RingBuffer<uint8_t>** current, particle::services::RingBuffer<uint8_t>** another) {
+        if (current) {
+            *current = usingAltRxBuffer_ ? &rxBufferAlt_ : &rxBuffer_;
+        }
+        if (another) {
+            *another = usingAltRxBuffer_ ? &rxBuffer_ : &rxBufferAlt_;
+        }
+    }
+
     size_t prepareNextRingBuffer(bool flush = false) {
         // Updates current size
-        particle::services::RingBuffer<uint8_t>* buffer = nullptr;
-        if (usingAltRxBuffer_) {
-            buffer = &rxBuffer_;
-        } else {
-            buffer = &rxBufferAlt_;
-        }
+        particle::services::RingBuffer<uint8_t>* nextRxBuff = nullptr;
+        acquireRxBuffer(nullptr, &nextRxBuff);
 
         // To keep the received data contiguous
-        if (buffer->data()) {
+        if (nextRxBuff->data()) {
             return 0;
         }
 
-        buffer->acquireBegin();
+        nextRxBuff->acquireBegin();
 
-        const size_t acquirable = buffer->acquirable();
-        const size_t acquirableWrapped = buffer->acquirableWrapped();
+        const size_t acquirable = nextRxBuff->acquirable();
+        const size_t acquirableWrapped = nextRxBuff->acquirableWrapped();
         size_t rxSize = std::max(acquirable, acquirableWrapped);
 
         if (RESERVED_RX_SIZE) {
@@ -634,7 +629,7 @@ private:
             return 0;
         }
 
-        auto ptr = buffer->acquire(rxSize);
+        auto ptr = nextRxBuff->acquire(rxSize);
 #ifdef DEBUG_BUILD
             SPARK_ASSERT(ptr);
 #endif // DEBUG_BUILD
@@ -647,11 +642,14 @@ private:
             return;
         }
 
-        rxBuffer_.acquireBegin();
-        const size_t acquirable = rxBuffer_.acquirable();
-        const size_t acquirableWrapped = rxBuffer_.acquirableWrapped();
+        particle::services::RingBuffer<uint8_t>* currRxBuff = nullptr;
+        acquireRxBuffer(&currRxBuff, nullptr);
+
+        currRxBuff->acquireBegin();
+        const size_t acquirable = currRxBuff->acquirable();
+        const size_t acquirableWrapped = currRxBuff->acquirableWrapped();
         size_t rxSize = std::max(acquirable, acquirableWrapped);
-        auto ptr = rxBuffer_.acquire(rxSize);
+        auto ptr = currRxBuff->acquire(rxSize);
         nrf_uarte_rx_buffer_set(uarte_, ptr, rxSize);
 
         ++receiving_;
@@ -685,7 +683,10 @@ private:
                     rxto = true;
                 }
             }
+            data(); // Commit last received data
+            receiving_ = 0;
         }
+        disableTimer();
     }
 
     void stopTransmission() {
@@ -729,7 +730,7 @@ private:
         consumedTimerCount_ = 0;
     }
 
-    size_t timerValue() {
+    uint16_t timerValue() {
         nrf_timer_task_trigger(timer_, NRF_TIMER_TASK_CAPTURE0);
         return nrf_timer_cc_read(timer_, NRF_TIMER_CC_CHANNEL0);
     }
@@ -817,7 +818,7 @@ private:
 
     volatile bool transmitting_;
     volatile uint8_t receiving_;
-    volatile size_t consumedTimerCount_;
+    volatile uint16_t consumedTimerCount_;
     volatile bool usingAltRxBuffer_;
 
     Config config_ = {};
--- a/hal/src/nRF52840/usart_hal.cpp
+++ b/hal/src/nRF52840/usart_hal.cpp
@@ -32,6 +32,7 @@
 #include "interrupts_hal.h"
 #include "usart_hal_private.h"
 #include "timer_hal.h"
+#include "scope_guard.h"
 
 #if PLATFORM_ID == PLATFORM_TRACKER
 #include "i2c_hal.h"
@@ -57,10 +58,10 @@ class RxLock {
 public:
     RxLock(NRF_UARTE_Type* uarte)
             : uarte_(uarte) {
-        nrf_uarte_int_disable(uarte, NRF_UARTE_INT_ENDRX_MASK);
+        nrf_uarte_int_disable(uarte, NRF_UARTE_INT_ENDRX_MASK | NRF_UARTE_INT_RXSTARTED_MASK);
     }
     ~RxLock() {
-        nrf_uarte_int_enable(uarte_, NRF_UARTE_INT_ENDRX_MASK);
+        nrf_uarte_int_enable(uarte_, NRF_UARTE_INT_ENDRX_MASK | NRF_UARTE_INT_RXSTARTED_MASK);
     }
 
 private:
@@ -112,7 +113,8 @@ public:
               rtsPin_(rts),
               transmitting_(false),
               receiving_(0),
-              rxConsumed_(0),
+              consumedTimerCount_(0),
+              usingAltRxBuffer_(false),
               evGroup_(nullptr) {
         evGroup_ = xEventGroupCreate();
         SPARK_ASSERT(evGroup_);
@@ -138,6 +140,7 @@ public:
         CHECK_TRUE(conf.tx_buffer_size, SYSTEM_ERROR_INVALID_ARGUMENT);
 
         rxBuffer_.init((uint8_t*)conf.rx_buffer, conf.rx_buffer_size);
+        rxBufferAlt_.init(secondaryRxBuffer_, sizeof(secondaryRxBuffer_));
         txBuffer_.init((uint8_t*)conf.tx_buffer, conf.tx_buffer_size);
 
         configured_ = true;
@@ -213,7 +216,7 @@ public:
 
         disableInterrupts();
 
-        nrf_uarte_int_enable(uarte_, NRF_UARTE_INT_ENDRX_MASK | NRF_UARTE_INT_ENDTX_MASK);
+        nrf_uarte_int_enable(uarte_, NRF_UARTE_INT_RXSTARTED_MASK | NRF_UARTE_INT_ENDRX_MASK | NRF_UARTE_INT_ENDTX_MASK);
 
         NRFX_IRQ_PRIORITY_SET(nrfx_get_irq_number((void *)uarte_), prio_);
         NRFX_IRQ_ENABLE(nrfx_get_irq_number((void *)uarte_));
@@ -240,6 +243,7 @@ public:
         receiving_ = 0;
         config_ = {};
         rxBuffer_.reset();
+        rxBufferAlt_.reset();
         txBuffer_.reset();
 
         return SYSTEM_ERROR_NONE;
@@ -274,7 +278,16 @@ public:
     ssize_t data() {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         RxLock lk(uarte_);
-        ssize_t d = rxBuffer_.data();
+        particle::services::RingBuffer<uint8_t>* bufferCurrent = nullptr;
+        particle::services::RingBuffer<uint8_t>* bufferPre = nullptr;
+        if (usingAltRxBuffer_) {
+            bufferCurrent = &rxBufferAlt_;
+            bufferPre = &rxBuffer_;
+        } else {
+            bufferCurrent = &rxBuffer_;
+            bufferPre = &rxBufferAlt_;
+        }
+        ssize_t d = bufferPre->data() + bufferCurrent->data();
         if (receiving_) {
             // IMPORTANT: It seems that the timer value (which is a counter for every time UARTE generates RXDRDY event)
             // may potentially get larger than the configured RX DMA transfer:
@@ -284,11 +297,18 @@ public:
             // </quote>
             // We'll be extra careful here and make sure not to consume more than we can, otherwise
             // we may put the ring buffer into an invalid state as there are not safety checks.
-            const ssize_t toConsume = std::min(rxBuffer_.acquirePending(), timerValue() - rxConsumed_);
+            size_t timerCount = timerValue();
+            size_t toConsumeCount;
+            if (timerCount >= consumedTimerCount_) {
+                toConsumeCount = timerCount - consumedTimerCount_;
+            } else {
+                toConsumeCount = 65536 - consumedTimerCount_ + timerCount;
+            }
+            const ssize_t toConsume = std::min(bufferCurrent->acquirePending(), toConsumeCount);
             if (toConsume > 0) {
-                rxBuffer_.acquireCommit(toConsume);
-                rxConsumed_ += toConsume;
+                bufferCurrent->acquireCommit(toConsume);
                 d += toConsume;
+                consumedTimerCount_ = (consumedTimerCount_ + toConsume) % 65536;
             }
         }
         return d;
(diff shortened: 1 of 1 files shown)


avtolstoy:
> The problem of the current mechanism is if the NRF_UARTE_EVENT_ENDRX event is postponed to deal with by, for example, other interrupts with higher priority, while UART is receiving data, then the NRF_UARTE_EVENT_RXDRDY is asserted, but the receiver has stopped and the received byte cannot be moved to the RAM buffer, and we lost the byte. The longer the NRF_UARTE_EVENT_ENDRX being postponed, the more data will lose. This happens at the middle of a data receiving frame when the NRF_UARTE_EVENT_ENDRX is asserted, so using a larger buffer doesn't fix the issue.

UARTE peripheral has its own FIFO, so those bytes should not be lost, but should be in the RX FIFO. We might have to force-flush them, but they are not supposed to be lost, unless of course we go over RX FIFO size.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -284,11 +297,18 @@ class Usart {
...
             // We'll be extra careful here and make sure not to consume more than we can, otherwise
             // we may put the ring buffer into an invalid state as there are not safety checks.
-            const ssize_t toConsume = std::min(rxBuffer_.acquirePending(), timerValue() - rxConsumed_);
+            size_t timerCount = timerValue();
+            size_t toConsumeCount;
+            if (timerCount >= consumedTimerCount_) {
+                toConsumeCount = timerCount - consumedTimerCount_;
+            } else {
+                toConsumeCount = 65536 - consumedTimerCount_ + timerCount;
+            }
```
jlovinger-particle, line 306: Just make `timerValue()` return a `uint16_t` here and use `uint16_t` instead of size_t. There are plenty of places that already assume that. Such as setting the width of the timer itself to 16-bit and magic numbers of 65536.

The math breaks if `timerCount` is actually the full width of the underlying `size_t` so I had to read through more code to figure out it was really constrained to 16-bit. Making the types match that constraint makes it more clear and ensures a simple `timerCount - consumedTimerCount_` always presents the correct value without manually adding in 65536.

hal/src/nRF52840/usart_hal.cpp
```diff
@@ -542,21 +587,25 @@ class Usart {
...
+        } else {
+            buffer = &rxBufferAlt_;
         }
 
-        // Updates current size
-        rxBuffer_.acquireBegin();
+        // To keep the received data contiguous
+        if (buffer->data()) {
+            return 0;
+        }
```
jlovinger-particle, line 602: Somewhere decided we needed to prepare the next buffer but it has unconsumed bytes so rejected. Is this a surface for the original problem to present again? Or something else bad?

XuGuohui: If we don't read out received data in time, then yes. But this is a common issue even if we use a super large buffer if user don't read out the data in time. We will configure the another buffer in the `read()` method as long as the another buffer is empty.

XuGuohui:
Not that double buffering is used in the official SDK's UART driver as well and there is a chipset level register to support this feature, i.e. the `SHORTS` register. The main purpose of the H/W feature is to support receiving continuous data stream without losing any byte, such that we can get rid of software latency.

avtolstoy:
@XuGuohui I understand how double-buffering works and why it's implemented for DMA peripherals. I'm trying to understand why slapping another 64-byte buffer on top of existing one solves the problem. There is no difference at which point we run out of buffer space: in the first buffer or the second one. The underlying problem at least to me still looks the same: at the end of the DMA transfer further received data into the RX FIFO is for some reason lost and not transferred over into RX buffer? Is it lost because of the failure in our UART HAL to correctly flush it or schedule another DMA transfer or is this just a failure of the consumer of this data to read it out and free up space in RX buffer and we just overflow? If it's the second one, double buffering does not help and the proper solution is flow control (software or hardware) or adjustment of buffer sizes for a particular application which we have an API for.

XuGuohui:
>There is no difference at which point we run out of buffer space

This is not an issue about running out of buffer space. As you said the issue is "at the end of the DMA transfer further received data into the RX FIFO is for some reason lost and not transferred over into RX buffer". The root causes I would sumerize are:

avtolstoy:
Can we try dealing with these issues then? For #1 avoid locking or change the method (e.g. primask or baseirq instead) or avoiding locking all together at least in `available()`? I doubt UARTs would be interfering with each other as far as interrupts are concerned, so most likely this is `RxLock` and caller doing things in busy manner?

For #2 the same, whatever changes e.g. like in this PR changing to a wrapping counter to avoid resets etc.

XuGuohui:
For #1 it is either because the `Serial1`'s interrupt priority is lower than others, or the `RxLock`. In my opinion, as long as there are other ISRs with higher priority blocking the `Serial1`'s ISR, not only the AT UART ISR, but also SoftDevice ISRs, changing the primask or basepri doesn't seams to help, but may introduce more problems to other critical peripherals. Removing the `RxLock` in `available()` seams to be fine. The defect is user might not get the exact number of bytes in the buffer. Anyway, so long as there is one way blocking the execution of  the `Serial1`'s handler, fixing the another way doesn't actually fix the issue.

hal/src/nRF52840/usart_hal.cpp
jlovinger-particle: So it will continue to safely fill the current buffer without issue so long as it has room? Why is there a need for two buffers then?

XuGuohui: >it will continue to safely fill the current buffer without issue so long as it has room?

Yes. The secondary buffer aims to make sure when the ENDRX event is assert hardware can be immediately ready to receive the incoming data, even if the ENDRX event is not handled by ISR. We currently have to wait until the ENDRX event is handled by software and then start to receive the incoming data. There is a gap between when the ENDRX event is asserted and when the receiver is ready to receive data. This gap causes data lost.

avtolstoy:
> For #1 it is either because the Serial1's interrupt priority is lower than others, or the RxLock. In my opinion, as long as there are other ISRs with higher priority blocking the Serial1's ISR, not only the AT UART ISR, but also SoftDevice ISRs, changing the primask or basepri doesn't seams to help, but may introduce more problems to other critical peripherals.

Let's try to figure out what exactly causes the delay in processing of ENDRX then, right now it sounds we still have no idea and guess it's SoftDevice or other interrupts or our locking mechanism.

With introduction of an additional buffer and scheduling sequential RX DMA transactions we are just hiding the underlying problem here.

XuGuohui:
I made a diff based on the develop branch to debug the root cause:
```diff
diff --git a/hal/src/nRF52840/usart_hal.cpp b/hal/src/nRF52840/usart_hal.cpp
index b143fec05..c0ed90222 100644
--- a/hal/src/nRF52840/usart_hal.cpp
+++ b/hal/src/nRF52840/usart_hal.cpp
@@ -32,6 +32,7 @@
 #include "interrupts_hal.h"
 #include "usart_hal_private.h"
 #include "timer_hal.h"
+#include "concurrent_hal.h"
 
 #if PLATFORM_ID == PLATFORM_TRACKER
 #include "i2c_hal.h"
@@ -58,13 +59,18 @@ public:
     RxLock(NRF_UARTE_Type* uarte)
             : uarte_(uarte) {
         nrf_uarte_int_disable(uarte, NRF_UARTE_INT_ENDRX_MASK);
+        st_ = HAL_disable_irq(); // IRQ >= 2
+        // os_thread_scheduling(false, nullptr);
     }
     ~RxLock() {
         nrf_uarte_int_enable(uarte_, NRF_UARTE_INT_ENDRX_MASK);
+        HAL_enable_irq(st_);
+        // os_thread_scheduling(true, nullptr);
     }
 
 private:
     NRF_UARTE_Type* uarte_;
+    int st_;
 };
 
 class TxLock {
@@ -112,7 +118,7 @@ public:
               rtsPin_(rts),
               transmitting_(false),
               receiving_(0),
-              rxConsumed_(0),
+              consumedTimerCount_(0),
               evGroup_(nullptr) {
         evGroup_ = xEventGroupCreate();
         SPARK_ASSERT(evGroup_);
@@ -274,6 +280,8 @@ public:
     ssize_t data() {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         RxLock lk(uarte_);
+        unsigned long start = HAL_Timer_Get_Micro_Seconds();
+        unsigned long end = start;
         ssize_t d = rxBuffer_.data();
         if (receiving_) {
             // IMPORTANT: It seems that the timer value (which is a counter for every time UARTE generates RXDRDY event)
@@ -284,11 +292,25 @@ public:
             // </quote>
             // We'll be extra careful here and make sure not to consume more than we can, otherwise
             // we may put the ring buffer into an invalid state as there are not safety checks.
-            const ssize_t toConsume = std::min(rxBuffer_.acquirePending(), timerValue() - rxConsumed_);
+            uint16_t timerCount = timerValue();
+            uint16_t toConsumeCount;
+            if (timerCount >= consumedTimerCount_) {
+                toConsumeCount = timerCount - consumedTimerCount_;
+            } else {
+                toConsumeCount = 65536 - consumedTimerCount_ + timerCount;
+            }
+            const ssize_t toConsume = std::min(rxBuffer_.acquirePending(), (size_t)toConsumeCount);
             if (toConsume > 0) {
                 rxBuffer_.acquireCommit(toConsume);
-                rxConsumed_ += toConsume;
                 d += toConsume;
+                consumedTimerCount_ = (consumedTimerCount_ + toConsume) % 65536;
+            }
+            end = HAL_Timer_Get_Micro_Seconds();
+            unsigned long elapsed = end - start;
+            if (uarte_ == NRF_UARTE0 && elapsed > 20) {
+                if (nrf_uarte_event_check(uarte_, NRF_UARTE_EVENT_ENDRX)) {
+                    LOG(TRACE, "Lock: %ld, txCount_: %ld", elapsed, txCount_);
+                }
             }
         }
         return d;
@@ -306,14 +328,25 @@ public:
         const size_t readSize = std::min((size_t)maxRead, size);
         CHECK_TRUE(readSize > 0, SYSTEM_ERROR_NO_MEMORY);
         RxLock lk(uarte_);
+        unsigned long start = HAL_Timer_Get_Micro_Seconds();
+        unsigned long end = start;
         ssize_t r = CHECK(rxBuffer_.get(buffer, readSize));
         if (!receiving_) {
             startReceiver();
+            LOG(TRACE, "Restart");
+        }
+        end = HAL_Timer_Get_Micro_Seconds();
+        unsigned long elapsed = end - start;
+        if (uarte_ == NRF_UARTE0 && elapsed > 20) {
+            if (nrf_uarte_event_check(uarte_, NRF_UARTE_EVENT_ENDRX)) {
+                LOG(TRACE, "Lock2: %ld, txCount_: %ld, ", elapsed, txCount_);
+            }
         }
         return r;
     }
 
     ssize_t peek(uint8_t* buffer, size_t size) {
+        LOG(TRACE, "lastRxSize_: %ld, rxSize_: %d, pending: %d, missed: %d, endrxGap_: %ldus", lastRxSize_, rxSize_, rxBuffer_.acquirePending(), missed_, endrxGap_);
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         const ssize_t maxRead = CHECK(data());
         const size_t peekSize = std::min((size_t)maxRead, size);
@@ -338,6 +371,8 @@ public:
     }
 
     ssize_t flush() {
+        missed_ = 0;
+        txCount_++;
         while (true) {
             while (transmitting_) {
                 // FIXME: busy loop
@@ -362,6 +397,9 @@ public:
 
     void pump() {
         if (!willPreempt() && isEnabled()) {
+            if (uarte_ == NRF_UARTE0) {
+                LOG(ERROR, "pump");
+            }
             interruptHandler();
         }
     }
@@ -370,6 +408,35 @@ public:
         BaseType_t yield = pdFALSE;
         bool eventGenerated = false;
 
+        if (hal_interrupt_is_isr()) {
+            if (uarte_ != NRF_UARTE0 && nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ENDRX)) {
+                missed_++;
+                LOG(ERROR, "Missed1:%d", missed_);
+            }
+            if (uarte_ == NRF_UARTE0 && !nrf_uarte_int_enable_check(NRF_UARTE0, NRF_UARTE_INT_ENDRX_MASK) && nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ENDRX)) {
+                missed_++;
+                // LOG(ERROR, "Missed2: %d, INT: %08X", missed_, NRF_UARTE0->INTENSET);
+                uint16_t event = 0x0000;
+                event |= nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_CTS);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_NCTS)) << 1);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_RXDRDY)) << 2);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ENDRX)) << 3);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_TXDRDY)) << 4);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ENDTX)) << 5);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ERROR)) << 6);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_RXTO)) << 7);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_RXSTARTED)) << 8);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_TXSTARTED)) << 9);
+                event |= ((uint16_t)(nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_TXSTOPPED)) << 10);
+                LOG(ERROR, "Missed2:%d,%04X,%08X", missed_, event, NRF_UARTE0->INTENSET);
+                for (uint8_t i = 0; i < 4; i++) {
+                    LOG(ERROR, "%08X", NVIC->ISPR[i]);
+                }
+            }
+        }
+
         if (nrf_uarte_int_enable_check(uarte_, NRF_UARTE_INT_RXDRDY_MASK)) {
             if (nrf_uarte_event_check(uarte_, NRF_UARTE_EVENT_RXDRDY)) {
                 nrf_uarte_int_disable(uarte_, NRF_UARTE_INT_RXDRDY_MASK);
@@ -398,11 +465,13 @@ public:
                 nrf_uarte_event_clear(uarte_, NRF_UARTE_EVENT_ENDRX);
                 nrf_uarte_event_clear(uarte_, NRF_UARTE_EVENT_RXDRDY);
 
-                nrf_timer_task_trigger(timer_, NRF_TIMER_TASK_CLEAR);
-                rxConsumed_ = 0;
+                nrf_timer_task_trigger(NRF_TIMER4, NRF_TIMER_TASK_CAPTURE0);
+                endrxGap_ = nrf_timer_cc_read(NRF_TIMER4, NRF_TIMER_CC_CHANNEL0);
 
                 if (rxBuffer_.acquirePending() > 0) {
-                    rxBuffer_.acquireCommit(rxBuffer_.acquirePending());
+                    size_t remain = rxBuffer_.acquirePending();
+                    rxBuffer_.acquireCommit(remain);
+                    consumedTimerCount_ = (consumedTimerCount_ + remain) % 65536;
                 }
 
                 --receiving_;
@@ -481,6 +550,12 @@ public:
     }
 
 private:
+    volatile size_t rxSize_;
+    volatile size_t lastRxSize_ = 0;
+    volatile size_t missed_;
+    volatile unsigned long endrxGap_ = 0;
+    size_t txCount_ = 0;
+
     int disable(bool end = true) {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
 
@@ -596,6 +671,8 @@ private:
             } else {
                 nrf_uarte_task_trigger(uarte_, NRF_UARTE_TASK_FLUSHRX);
             }
+            lastRxSize_ = rxSize_;
+            rxSize_ = rxSize;
         }
     }
 
@@ -651,13 +728,23 @@ private:
         nrf_ppi_channel_endpoint_setup(ppi_, (uint32_t)&uarte_->EVENTS_RXDRDY,
                 (uint32_t)&timer_->TASKS_COUNT);
         nrf_ppi_channel_enable(ppi_);
+
+        if (uarte_ == NRF_UARTE0) {
+            NRFX_IRQ_DISABLE(nrfx_get_irq_number((void*)NRF_TIMER4));
+            nrf_timer_mode_set(NRF_TIMER4, NRF_TIMER_MODE_TIMER);
+            nrf_timer_bit_width_set(NRF_TIMER4, NRF_TIMER_BIT_WIDTH_32);
+            nrf_timer_frequency_set(NRF_TIMER4, NRF_TIMER_FREQ_1MHz);
+            nrf_timer_task_trigger(NRF_TIMER4, NRF_TIMER_TASK_START);
+            nrf_ppi_channel_endpoint_setup(NRF_PPI_CHANNEL6, (uint32_t)&NRF_UARTE0->EVENTS_ENDRX, (uint32_t)&NRF_TIMER4->TASKS_CLEAR);
+            nrf_ppi_channel_enable(NRF_PPI_CHANNEL6);
+        }
     }
 
     void disableTimer() {
         nrf_timer_task_trigger(timer_, NRF_TIMER_TASK_CLEAR);
         nrf_timer_task_trigger(timer_, NRF_TIMER_TASK_SHUTDOWN);
         nrf_ppi_channel_disable(ppi_);
-        rxConsumed_ = 0;
+        consumedTimerCount_ = 0;
     }
 
     size_t timerValue() {
@@ -748,7 +835,7 @@ private:
 
     volatile bool transmitting_;
     volatile uint8_t receiving_;
-    volatile size_t rxConsumed_;
+    volatile uint16_t consumedTimerCount_;
 
     Config config_ = {};
 
@@ -777,6 +864,9 @@ Usart* getInstance(hal_usart_interface_t serial) {
 }
 
 void uarte0InterruptHandler(void) {
+    // if (!nrf_uarte_int_enable_check(NRF_UARTE0, NRF_UARTE_INT_ENDRX_MASK) && nrf_uarte_event_check(NRF_UARTE0, NRF_UARTE_EVENT_ENDRX)) {
+    //     LOG(ERROR, "uarte0Int: %08X", NRF_UARTE0->INTENSET);
+    // }
     getInstance(HAL_USART_SERIAL1)->interruptHandler();
 }
```

In the test, I introduced a timer to capture the time it elapsed (`endrxGap_`) after the ENDRX is asserted and before it is handled. The test is also capable of detecting whether the handling of the ENDRX is posponed by AT UART or the TX path of UARTE0.

|Test cases|Cloud connected|BLE enabled|RxLock|Result|Comment|
|---|---|---|---|---|---|
|1|YES|NO|disable scheduler|failure|other IRQs' preemption|
|2|YES|YES|Disable IRQs (BASEPRI=2)|failure|SD's IRQs|
|3|YES|NO|Disable IRQs (BASEPRI=2)|failure|unknown|

#1 Logs:
>0000126000 [app] INFO: total=1552, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=1552
0000128070 [hal] TRACE: Lock: 27, txCount_: 1579
0000128123 [app] INFO: RX-LEN-ERROR=42 (51 ms)
0000128123 [hal] TRACE: lastRxSize_: 45, rxSize_: 44, pending: 20, missed: 1, endrxGap_: 585us
0000128123 [app] INFO: total=1579, rx_total=1579, timeout=0, overrun=0, rx_len=1, crc=0
RX: 010203040506070809101112131415161718192021222324262728293031323334353637383940414243

#2 Logs:
>0003092000 [app] INFO: total=40151, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=40151
0003095000 [app] INFO: total=40190, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=40190
0003096688 [app] INFO: BLE Scan thread.
0003096841 [app] INFO: RX-LEN-ERROR=41 (51 ms)
0003096841 [hal] TRACE: lastRxSize_: 41, rxSize_: 39, pending: 25, missed: 1, endrxGap_: 651us
0003096841 [app] INFO: total=40214, rx_total=40214, timeout=0, overrun=0, rx_len=1, crc=0
RX: 0102030405060708091011121314151617181920212223242526272829303132333637383940414243

#3 Logs:
>0015821000 [app] INFO: total=205871, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=205871
0015824000 [app] INFO: total=205910, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=205910
0015827000 [app] INFO: total=205949, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=205949
0015829252 [hal] TRACE: Restart
0015829252 [hal] TRACE: Restart
0015829252 [hal] TRACE: Restart
0015829256 [app] INFO: loop delay: 20
0015829306 [app] INFO: RX-LEN-ERROR=42 (51 ms)
0015829306 [hal] TRACE: lastRxSize_: 33, rxSize_: 31, pending: 25, missed: 0, endrxGap_: 17us
0015829306 [app] INFO: total=205979, rx_total=205979, timeout=0, overrun=0, rx_len=1, crc=0
RX: 010203040506070809101112131415161718192021222324252627282930313233343536373839404142

The root cause is eventually the ENDTX event is handled in time while the peer device is keeping sending data. This is caused by `RxLock`. During `RxLock` is acquired, the RX INT is disabled, and the following events may postpone the ENDRX handling:
1. OS scheduling
2. Other ISRs with higher priority
3. TX path of itsellf

The above logs show us how these circumstance happened. I also tried bumpping the RX INT's priority to `2` and disablinng BLE in user application and disbalinng IRQ through `BASEPRI`, but it eventually failed:
>0015344079 [comm.coap] TRACE: ACK 0.00  size=4 token= id=13937
0015345000 [app] INFO: total=199771, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=199771
0015348000 [app] INFO: total=199810, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=199810
0015350579 [app] INFO: RX-LEN-ERROR=36 (51 ms)
0015350579 [hal] TRACE: rxSize_: 32, pending: 26, missed: 0, endrxGap_: 2170us
0015350580 [app] INFO: total=199843, rx_total=199843, timeout=0, overrun=0, rx_len=1, crc=0
RX: 010203040506070809101112131415161718192021222324252627282930313233343536

There is a weird thing if `LOG_FROM_ISR` is enabled, which is the UARTE0 ISR is executed even if there is no related pending IRQ, neither there is UART INT enabled. I tried using polling mechanism for the TX path so that the ENDTX INT can be disabled and only the ENDRX INT is enabled, but the UARTE ISR is serviced while `RxLock`, Logs:
>0000057000 [app] INFO: total=698, timeout=0, overrun=0, rx_len=0, crc=0, total_rx=698
0000059116 [hal] ERROR: Missed2:1,030D,00000100
0000059169 [app] INFO: RX-LEN-ERROR=42 (51 ms)
0000059169 [hal] TRACE: lastRxSize_: 59, rxSize_: 58, pending: 34, missed: 0, endrxGap_: 524us
0000059169 [app] INFO: total=726, rx_total=726, timeout=0, overrun=0, rx_len=1, crc=0
RX: 010203040506070809101112131415161718192021222324262728293031323334353637383940414243

The data loss usually happened at the middle of a data frame, so it's nothing about buffer underrun.

avtolstoy:
@XuGuohui Shouldn't we capture the time difference between ENDRX -> Timer PPI and `rf_uarte_int_enable_check(uarte_, NRF_UARTE_INT_ENDRX_MASK)` and not `NRF_UARTE_INT_RXDRDY_MASK`?

XuGuohui:
>Shouldn't we capture the time difference between ENDRX -> Timer PPI and nrf_uarte_int_enable_check(uarte_, NRF_UARTE_INT_ENDRX_MASK) and not NRF_UARTE_INT_RXDRDY_MASK?

I did. No? Which lines did you point at? On ENDRX asserted I clear the timer counter and when it handles the ENDRX event I capture the time it elasped since the event is asserted.

avtolstoy:
Ah, the diff looks weird and makes it look like it's part of `RXDRDY` handling. #2 and #3 do look somewhat weird though still. Even 17us latency is quite a lot without supposedly anything interfering and 0.5ms is abysmal (I can understand #1 where thread scheduling is not disabled so things might be affected by that). Do we perhaps have some frequently firing interrupt perhaps?

XuGuohui:
>Do we perhaps have some frequently firing interrupt perhaps?

USB logging interrupt probably. But during `RxLock` with `BASEPRI = 2` the issue persisted though, such as the test result #2 and #3.

PR #2689, opened 2023-08-24
Network Interface Tester First Draft

scott-brust:
### Problem

We need to determine the quality of the various network connections available on the msom platform. We should exercise them and gather relevant metrics to inform that decision

### Solution

Send UDP data to an echo server across each interface and collect diagnostics. 

⚠️ This PR is not intended to be merged but rather serve as a point of discussion on how this feature should be implemented. Its very much a work in progress

### Steps to Test

Run example app, bring up any number of network interfaces, use the `1` command to test the interface, then `2` to print some stats about the test diagnostics.

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(AUTOMATIC);
SYSTEM_THREAD(ENABLED);

Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL, 
{
    { "app", LOG_LEVEL_ALL }, 
    //{ "sys.power", LOG_LEVEL_TRACE },
    { "comm.protocol", LOG_LEVEL_TRACE },
    { "system", LOG_LEVEL_TRACE },
    { "network", LOG_LEVEL_TRACE },
    { "ncp.at", LOG_LEVEL_TRACE },
    { "system.nm", LOG_LEVEL_TRACE },
    { "ncp.client", LOG_LEVEL_TRACE },
    { "net.rltkncp", LOG_LEVEL_TRACE },
    { "net.ifapi", LOG_LEVEL_TRACE },
    { "net.ppp.client", LOG_LEVEL_TRACE },
    { "service.ntp", LOG_LEVEL_TRACE },
    { "net.en", LOG_LEVEL_TRACE },
    { "mux", LOG_LEVEL_TRACE }
}   
);

struct NetIfDiagnostics {
    network_interface_t interface;
    uint32_t dnsResolutionAttempts;
    uint32_t dnsResolutionFailures;
    uint32_t socketConnAttempts;
    uint32_t socketConnFailures;
    uint32_t txBytes;
    uint32_t rxBytes;
    uint32_t rxTimeouts;
    uint32_t avgPacketRoundTripTime;
};

void logNetworkStates() {
    Log.info("Ethernet Ready: %d WiFi ready: %d Cellular ready: %d", Ethernet.ready(), WiFi.ready(), Cellular.ready());
}

void setup() {

#if HAL_PLATFORM_WIFI
    WiFi.on();
#endif
#if HAL_PLATFORM_CELLULAR
    Cellular.on();
#endif
    Serial1.begin(115200);
}
 
void loop() {
    if (!Serial1.available()) {
        return;
    }

    while (Serial1.available()) {
        char c = Serial1.read();
        logNetworkStates();

        if (c == 'p') {
            Particle.publishVitals(particle::NOW);
        }
        else if(c == '1') {
            // Call interface test
            system_internal(4, nullptr);
        }
        else if(c == '2') {
            // lwip stats, including link and udp stats
            // Link *doesnt* include wifi? 
            // Tracks packets across all interfaces, doesnt reset when interfaces are brought up/down
            stats_* lwipStats = (stats_*)system_internal(5, nullptr); 
            Log.info("LWIP link  xmit: %u recv %u drop %u err %u", 
                lwipStats->link.xmit, lwipStats->link.recv, lwipStats->link.drop, lwipStats->link.err);
            Log.info("LWIP udp   xmit: %u recv %u drop %u err %u", 
                lwipStats->udp.xmit, lwipStats->udp.recv, lwipStats->udp.drop, lwipStats->udp.err);

            // our stats
            const Vector<NetIfDiagnostics>* netIfTesterDiagnostics = (const Vector<NetIfDiagnostics>*)(system_internal(6, nullptr));
            for (auto& i: *netIfTesterDiagnostics) {
                Log.info("interface %lu dns_att:%lu dns_fail: %lu tx %lu rx %lu", 
                    i.interface, i.dnsResolutionAttempts, i.dnsResolutionFailures, i.txBytes, i.rxBytes);
            }
        }
    }
}
```
Example output:

```
0000470837 [app] INFO: Ethernet Ready: 0 WiFi ready: 1 Cellular ready: 1
0000470839 [app] INFO: LWIP link  xmit: 107 recv 105 drop 0 err 0.             // LWIP stats before the test
0000470844 [app] INFO: LWIP udp   xmit: 45 recv 42 drop 2 err 0
0000470850 [app] INFO: interface 3 dns_att:3 dns_fail: 3 tx 0 rx 0
0000470857 [app] INFO: interface 4 dns_att:3 dns_fail: 1 tx 10 rx 10          // Our stats before the test
0000470862 [app] INFO: interface 5 dns_att:3 dns_fail: 0 tx 15 rx 15
0000471837 [app] INFO: Ethernet Ready: 0 WiFi ready: 1 Cellular ready: 1     // WIFI and cellular interfaces are up + configured. The cloud connection is established over wifi, but that is not explicit here
0000471864 [system.nd] INFO: IF #3 failed to resolve DNS for publish-receiver-udp.particle.io : 0.0.0.0       // The ethernet interface is not enabled so DNS resolution fails
0000471893 [system.nd] INFO: Testing IF #4 with publish-receiver-udp.particle.io : 54.234.140.231        // Wifi succeeds
0000473719 [system.nd] INFO: UDP begin 1 send: 5 receive: 5 message: Hello
0000473740 [system.nd] INFO: Testing IF #5 with publish-receiver-udp.particle.io : 54.234.140.231      // Cellular succeeds
0000474825 [system.nd] INFO: UDP begin 1 send: 5 receive: 5 message: Hello
0000476326 [app] INFO: Ethernet Ready: 0 WiFi ready: 1 Cellular ready: 1
0000476328 [app] INFO: LWIP link  xmit: 110 recv 108 drop 0 err 0                 // LWIP stats show 3 new link packets
0000476333 [app] INFO: LWIP udp   xmit: 50 recv 47 drop 2 err 0                 // UDP shows 5
0000476339 [app] INFO: interface 3 dns_att:4 dns_fail: 4 tx 0 rx 0               // Ethernet DNS failure increment
0000476346 [app] INFO: interface 4 dns_att:4 dns_fail: 1 tx 15 rx 15            // Wifi bytes sent/received incremented
0000476351 [app] INFO: interface 5 dns_att:4 dns_fail: 0 tx 20 rx 20
```


system/src/system_network_diagnostics.h
```diff
@@ -85,4 +86,36 @@ class NetworkDiagnostics {
...
+    static NetIfTester* instance();
+    void testInterfaces();
+
+    const Vector<NetIfDiagnostics>* getDiagnostics();
+
+private:
+    int testInterface(NetIfDiagnostics* diagnostics);
+
+    const uint16_t UDP_ECHO_PORT = 40000;
+    const char * UDP_ECHO_SERVER_HOSTNAME = "publish-receiver-udp.particle.io";
```
scott-brust, line 115: Hard coding a UDP echo server endpoint, but this should probably be a configured or stored somewhere else

system/src/system_network_manager.cpp
```diff
@@ -1064,7 +1066,7 @@ int NetworkManager::powerInterface(if_t iface, bool enable) {
         if (ifState->pwrState != IF_POWER_STATE_UP && ifState->pwrState != IF_POWER_STATE_POWERING_UP) {
             ifState->pwrState = IF_POWER_STATE_POWERING_UP;
         }
-        LOG(TRACE, "Request to power on the interface");
+        LOG(TRACE, "Request to power on interface %d", index);
```
scott-brust, line 1069: This is just to help understand which network interface is being powered on now that its ambiguous

system/src/system_task.cpp
```diff
@@ -785,6 +786,16 @@ void* system_internal(int item, void* reserved)
...
+    case 4: {
+        particle::NetIfTester::instance()->testInterfaces();
+        return nullptr;
+    }
+    case 5: {
+        return reinterpret_cast<void*>(&lwip_stats);
+    }
+    case 6: {
+        return (void*)particle::NetIfTester::instance()->getDiagnostics();
+    }
```
scott-brust, line 798: Helper functions to allow user space app to call into dvos to test the new stuff

system/src/system_network_diagnostics.h
eberseth: Isn't this stored in DCT?

system/src/system_network_diagnostics.cpp
```diff
@@ -72,17 +161,24 @@ class NetworkCache
         system_tick_t m = millis();
         if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL)
         {
+            ts_ = millis();
 #if Wiring_Cellular
```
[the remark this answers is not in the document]
eberseth: My RAT doesn't seem to be displayed right.  I think it should be "LTE"
```json
{
  "device": {
    "network": {
      "cellular": {
        "radio_access_technology": "unknown",
        "operator": "AT&T Wireless Inc.",
        "cell_global_identity": {
          "mobile_country_code": 310,
          "mobile_network_code": "410",
```

eberseth:
How are the interfaces encoded for this output?  IF#1 is cellular and IF#2 WiFi?
```
0000756758 [app] INFO: Ethernet Ready: 0 WiFi ready: 1 Cellular ready: 1
0000756831 [system.nd] INFO: IF #3 failed to resolve DNS for publish-receiver-udp.particle.io : 0.0.0.0
0000756855 [system.nd] INFO: Testing IF #4 with publish-receiver-udp.particle.io : 54.234.140.231
0000758420 [system.nd] INFO: UDP begin 1 send: 5 receive: 5 message: Hello
0000758442 [system.nd] INFO: Testing IF #5 with publish-receiver-udp.particle.io : 54.234.140.231
0000759523 [system.nd] INFO: UDP begin 1 send: 5 receive: 5 message: Hello
0000804680 [app] INFO: Ethernet Ready: 0 WiFi ready: 1 Cellular ready: 1
0000804682 [app] INFO: LWIP link  xmit: 199 recv 193 drop 1 err 1
0000804688 [app] INFO: LWIP udp   xmit: 56 recv 87 drop 10 err 0
0000804694 [app] INFO: interface 3 dns_att:2 dns_fail: 2 tx 0 rx 0
0000804699 [app] INFO: interface 4 dns_att:2 dns_fail: 0 tx 10 rx 10
0000804706 [app] INFO: interface 5 dns_att:2 dns_fail: 1 tx 5 rx 5
```

system/src/system_network_diagnostics.cpp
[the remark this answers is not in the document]
scott-brust: Device service is responsible for that lookup from RAT to a technology string. Its possible that the field might not have gotten populated as expected due to some AT command failure. Ill look what command is supposed to populate that field that the server looks up

system/src/system_network_diagnostics.h
scott-brust: The device service address is stored in DCT, not this echo server that im using for testing here. 
I think long term we want to have a more official hosted/monitored service that we point to rather than this test instance.

scott-brust:
Binding the cloud connection to a specific network interface is still a work in progress. There are issues with closing/recreating the socket with it bound to a different interface. Will keep working through the details but the last commit should be roughly the direction we want to go in.

scott-brust:
> Binding the cloud connection to a specific network interface is still a work in progress. There are issues with closing/recreating the socket with it bound to a different interface. Will keep working through the details but the last commit should be roughly the direction we want to go in.

Cellular + Wifi connection binding works as expected. I can use the test app to bind to either and I can see the cloud traffic goes through the bound interface. Swapping between the two works fine with the cloud connection

Ethernet binding is much less reliable, but I think that is due to the current issues with ethernet + cloud connection in general with p2/msom platform

wiring/src/spark_wiring_network.cpp
```diff
@@ -53,6 +53,25 @@ NetworkClass& NetworkClass::from(network_interface_t nif) {
     }
 }
 
+network_interface_t NetworkClass::from(NetworkClass& network) {
```
scott-brust, line 56: Should we just implement a way to get `iface_`?

scott-brust:
```    
operator network_interface_t() const {
        return iface_;
    }
```
Is this supposed to be a function call operator or something else?

avtolstoy: https://en.cppreference.com/w/cpp/language/cast_operator

scott-brust: Ah ok, so we already have this: `static_cast<network_interface_t>(network)`, ill remove this function then as its redundant

system/src/system_network_manager.cpp
```diff
@@ -618,7 +618,7 @@ void NetworkManager::handleIfLink(if_t iface, const struct if_event* ev) {
         /* Interface link state changed to DOWN */
         if (state_ == State::IP_CONFIGURED || state_ == State::IFACE_LINK_UP) {
             if (countIfacesWithFlags(IFF_UP | IFF_LOWER_UP) == 0) {
-                transition(State::IFACE_UP);
+                transition(State::IFACE_DOWN);
```
scott-brust: I *think* this should be `State::IFACE_DOWN` and not `State::IFACE_UP`?

hal/network/ncp/cellular/cellular_hal.cpp
```diff
@@ -41,6 +41,8 @@ namespace {
 
 using namespace particle;
 
+static CellularDevice cachedInfo = {};
```
scott-brust: ~Not sure if there is a better place to cache this. But this simple implementation works for now~
TODO: Add these fields to `SystemCache`

scott-brust:
Refactored to use `sock_poll()` instead of UDPClient
Running the reachability test looks like this now (Ethernet is disconnected):
```
0001170027 [system.cm] TRACE: Resolving publish-receiver-udp.particle.io#40000
0001170107 [system.cm] INFO: test socket # 1, name en2 bound to Ethernet connected to 54.234.140.231#40000
0001170112 [system.cm] INFO: test socket # 2, name wl4 bound to WiFi     connected to 54.234.140.231#40000
0001170122 [system.cm] INFO: test socket # 3, name pp3 bound to Cellular connected to 54.234.140.231#40000
0001170131 [system.cm] TRACE: sock 1 packet # 1 tx > 262
0001170137 [system.cm] TRACE: sock 2 packet # 1 tx > 262
0001170142 [system.cm] TRACE: sock 3 packet # 1 tx > 262
0001171219 [system.cm] TRACE: sock 2 packet # 1 rx < 262
0001171222 [system.cm] TRACE: sock 2 packet # 2 tx > 262
0001171642 [system.cm] TRACE: sock 3 packet # 1 rx < 262
0001171645 [system.cm] TRACE: sock 3 packet # 2 tx > 262
0001172305 [system.cm] TRACE: sock 2 packet # 2 rx < 262
0001172307 [system.cm] TRACE: sock 2 packet # 3 tx > 262
0001172923 [system.cm] TRACE: sock 3 packet # 2 rx < 262
0001172925 [system.cm] TRACE: sock 3 packet # 3 tx > 262
0001173391 [system.cm] TRACE: sock 2 packet # 3 rx < 262
0001173393 [system.cm] TRACE: sock 2 packet # 4 tx > 262
0001174204 [system.cm] TRACE: sock 3 packet # 3 rx < 262
0001174206 [system.cm] TRACE: sock 3 packet # 4 tx > 262
0001174476 [system.cm] TRACE: sock 2 packet # 4 rx < 262
0001174478 [system.cm] TRACE: sock 2 packet # 5 tx > 262
0001175483 [system.cm] TRACE: sock 3 packet # 4 rx < 262
0001175484 [system.cm] INFO: Ethernet sent 1 packets, got 0 packets, avg rtt: 0
0001175491 [system.cm] INFO: WiFi     sent 5 packets, got 4 packets, avg rtt: 1083
0001175498 [system.cm] INFO: Cellular sent 4 packets, got 4 packets, avg rtt: 1333
```

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,464 @@
...
+        
+        LOG(TRACE, "sock %d packet # %d tx > %d", diagnostics->socketHandle, diagnostics->txPacketCount, r);
+    }
+    return r;
+};
+
+int NetIfTester::receiveTestPacket(NetIfDiagnostics* diagnostics) {
+    int r = sock_recv(diagnostics->socketHandle, diagnostics->rxBuffer, diagnostics->testPacketSize, MSG_DONTWAIT);
+    if (r > 0) {
+        // TODO: validate packet contents
```
scott-brust: Validate sequence number

system/src/system_cloud_connection.cpp
```diff
@@ -62,6 +62,11 @@ void spark_cloud_udp_port_set(uint16_t port)
     cloud_udp_port = port;
 }
 
+uint16_t spark_cloud_udp_port_get()
```
scott-brust, line 65: Is there a more authoritative source for the cloud connection port to use?

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,500 @@
...
+    if (preferred) {
+        preferredNetwork_ = network;
+    } else if (network == preferredNetwork_) {
+        preferredNetwork_ = NETWORK_INTERFACE_ALL;
+    }
+}
+
+network_handle_t ConnectionManager::getPreferredNetwork() {
+    //LOG(INFO, "getPreferredNetwork %lu", preferredNetwork_);
+    return preferredNetwork_;
```
scott-brust, line 105: TODO: Switch this to the combined protobuf configuration setting

system/src/system_network_diagnostics.cpp
```diff
@@ -66,23 +67,58 @@ class NetworkCache
...
-#elif Wiring_WiFi
-            sig_ = WiFi.RSSI();
-#endif
+        if (ts_ == 0 || (m - ts_) >= NETWORK_INFO_CACHE_INTERVAL) {
             ts_ = millis();
+            refreshCachedValues = true;
+        }
+
+// TODO: Better differentiator for "ACM"
+#if PLATFORM_ID == PLATFORM_MSOM
```
scott-brust: It feels like I should create a platform `#define` for ACM support. ideas for a name?

system/src/system_network_diagnostics.h
```diff
@@ -22,8 +22,22 @@
 
 namespace particle {
 
+enum {
+  DIAG_TYPE_APP_INT = 1 << 11,
+  DIAG_TYPE_APP_UINT = 2 << 11,
+  DIAG_TYPE_APP_FIXED_S16_16 = 3 << 11,
+};
```
scott-brust, line 33: This will go away when server side supports the alternate signal diagnostics

system/src/system_cloud.cpp
```diff
@@ -324,6 +331,20 @@ int spark_get_connection_property(unsigned property, void* data, size_t* size, v
...
         }
         return getConnectionProperty(protocol::Connection::MAX_FUNCTION_ARGUMENT_SIZE, data, size);
+    case SPARK_CLOUD_BIND_NETWORK_INTERFACE: {
+        if (*size >= sizeof(network_interface_t)) {
+            *((network_interface_t*)data) = CloudConnectionSettings::instance()->boundInterface();
+            return 0;    
+        }
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    case SPARK_CLOUD_GET_NETWORK_INTERFACE: {
```
scott-brust, line 341: Should this be excluded with an #ifdef since this relies on the NETIF api / LWIP getsockopt function to determine what interface is bound to the cloud socket?

system/src/system_cloud_internal.cpp
```diff
@@ -522,11 +522,13 @@ int Spark_Save(const void* buffer, size_t length, uint8_t type, void* reserved)
 
 int Spark_Restore(void* buffer, size_t max_length, uint8_t type, void* reserved)
 {
-	size_t length = 0;
-	int error = HAL_System_Backup_Restore(0, buffer, max_length, &length, nullptr);
-	if (error)
-		length = 0;
-	return length;
+    // Force new session handshake for now
+    return 0;
```
scott-brust, line 527: Do we want to merge this as is? or leave the session restoration problem for later

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,500 @@
...
+}
+
+ConnectionTester* ConnectionTester::instance() {
+    static ConnectionTester* tester = new ConnectionTester();
+    return tester;
+}
+
+static int getCloudHostnameAndPort(uint16_t * port, char * hostname, int hostnameLength) {
+    // Use this test server until real echo server is ready
+    strcpy(hostname, "publish-receiver-udp.particle.io");
```
scott-brust: Should we revert this and use the actual server address even if the DTLS message isnt implemented yet? This means internet testing will fail until its deployed, but at least we wont be pointing at the publish receiver endpoint

system/src/system_connection_manager.h
```diff
@@ -0,0 +1,102 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SYSTEM_CONNECTION_MANAGER_H
+#define SYSTEM_CONNECTION_MANAGER_H
+
+#include "hal_platform.h"
+
+#if HAL_PLATFORM_IFAPI
```
scott-brust, line 23: I think `ConnectionTester` class is heavily reliant on the socket api and should be included with `#if HAL_PLATFORM_IFAPI`, but should `ConnectionManager` be available to all platforms?

wiring/inc/spark_wiring_cloud.h
```diff
@@ -365,7 +366,8 @@ class CloudClass {
 
     static bool connected(void) { return spark_cloud_flag_connected(); }
     static bool disconnected(void) { return !connected(); }
-    static void connect(void) {
+    static void connect(const spark::NetworkClass& network = spark::Network) {
+        spark_set_connection_property(SPARK_CLOUD_BIND_NETWORK_INTERFACE, static_cast<network_interface_t>(network), nullptr, nullptr);
```
scott-brust, line 370: What should the behavior be when a user calls `Particle.connect(Cellular)` when cellular is off? 
Do we want to turn the hardware on? or keep the current behavior where Particle.connect()/disconnect() is decoupled from the network interface state?

system/src/system_connection_manager.h
```diff
@@ -0,0 +1,102 @@
...
+    int pollSockets(struct pollfd * pfds, int socketCount, int packetDataLength);
+    int sendTestPacket(ConnectionMetrics* metrics, int length);
+    int receiveTestPacket(ConnectionMetrics* metrics);
+    void cleanupSockets(bool recalculateMetrics = true);
+    ConnectionMetrics* metricsFromSocketDescriptor(int socketDescriptor);
+    bool testPacketsOutstanding();
+
+    const uint8_t REACHABILITY_TEST_MSG = 252;
+    const unsigned REACHABILITY_MAX_PAYLOAD_SIZE = 256;
+    const unsigned REACHABILITY_TEST_DURATION_MS = 5000;
```
scott-brust: This is arbitrary, what should this be?

technobly: I think we said maybe this should be changed so it doesn't blindly run for 5s, but maybe try to run 5 times and timeout after 5 seconds?

system/src/system_connection_manager.h
technobly: I think we should add a PLATFORM_ define for this and enable it only on MSOM platform for now.

system/src/system_connection_manager.cpp
technobly: I think we have a new endpoint for this now?

system/src/system_cloud.cpp
technobly: Yes this probably should be included with the PLATFORM_ define instead of PLATFORM_MSOM.

hal/network/ncp/cellular/cellular_hal.cpp
```diff
@@ -188,9 +203,25 @@ int cellular_device_info(CellularDevice* info, void* reserved) {
...
+    int r = SystemCache::instance().get(SystemCacheKey::CELLULAR_DEVICE_INFO, (uint8_t*)&cacheRead, sizeof(cacheRead));
+
     const NcpClientLock lock(client);
-    // Ensure the modem is powered on
-    CHECK(client->on());
+
+    // If modem is off, return the cached values if present
+    if (client->ncpPowerState() != NcpPowerState::ON) {
+        if (r != sizeof(CellularDeviceCached) || cacheRead.version != CELLULAR_DEVICE_VERSION) {
+            SystemCache::instance().del(SystemCacheKey::CELLULAR_DEVICE_INFO);
```
technobly, line 214: Instead deleting this on bad data (I don't think we use .del anywhere for cached objects), should we instead initialize it with sane defaults?  It mostly likely doesn't exist if this errors here, so deleting it won't do anything.  This is how you could also handle it if and when the version changes as well, so might be good to add this placeholder now.

system/src/system_connection_manager.cpp
scott-brust: Yes this was removed

wiring/inc/spark_wiring_cloud.h
scott-brust: DIscussed this [here](https://s.slack.com/archives/CS0T32ZKK/p1700246458097789?thread_ts=1700241989.915009&cid=CS0T32ZKK), leaving this behavior as currently implemented

system/src/system_cloud_connection.cpp
technobly: `cloud_udp_port` is it, so I think you are good.  Do you want to update `port = cloud_udp_port;` to `port = spark_cloud_udp_port_get()` as well?  If we are not planning on using `spark_cloud_udp_port_set()`, we could remove it.

system/src/system_connection_manager.h
scott-brust: Currently its a tradeoff of how much traffic do we want to generate vs how long do we want to block the system task, as the connection tester is done on the system thread currently. 
We should probably lower the test time, or run the test on its own thread.

system/src/system_cloud_connection_posix.cpp
```diff
@@ -298,6 +314,7 @@ int system_cloud_disconnect(int flags)
 int system_cloud_send(const uint8_t* buf, size_t buflen, int flags)
 {
     (void)flags;
+    LOG(TRACE, "TX s_state.socket %d buflen %d", s_state.socket, buflen);
```
technobly: LOG_DEBUG ?

system/src/system_cloud_connection_posix.cpp
```diff
@@ -328,13 +345,16 @@ int system_cloud_recv(uint8_t* buf, size_t buflen, int flags)
         }
     }
 
+    if (recvd) {
+        LOG(TRACE, "RX s_state.socket %d buflen %d recvd %d", s_state.socket, buflen, recvd);
```
technobly: LOG_DEBUG ?

user/applications/blank/application.cpp
```diff
@@ -0,0 +1,342 @@
...
+        Log.info("Ethernet is on");
+        uint8_t macAddrBuf[8] = {};
+        uint8_t* macAddr = Ethernet.macAddress(macAddrBuf);
+        if (macAddr != nullptr) {
+            Log.info("Ethernet MAC: %02x %02x %02x %02x %02x %02x",
+                    macAddr[0], macAddr[1], macAddr[2], macAddr[3], macAddr[4], macAddr[5]);
+        }
+        Ethernet.connect();
+        waitFor(Ethernet.ready, 30000);
+        Log.info("Ethernet.ready: %d", Ethernet.ready());
```
technobly: Comment out the above 3 lines by default?  Perhaps you can move this `blank/application.cpp` to `user/tests/apps/connection-manager.cpp` ?

user/applications/blank/application.cpp
```diff
@@ -0,0 +1,342 @@
...
+SYSTEM_MODE(SEMI_AUTOMATIC);
+SYSTEM_THREAD(ENABLED);
+STARTUP(System.enable(SYSTEM_FLAG_PM_DETECTION));
+
+retained uint8_t resetRetry = 0;
+#define ETHERNET_RETRY_MAX (10)
+
+#define CMD_SERIAL Serial1
+
+SerialLogHandler logHandler(LOG_LEVEL_ALL,
```
technobly: Change to `Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL,`

user/applications/blank/application.cpp
```diff
@@ -0,0 +1,342 @@
+#include "Particle.h"
+
+SYSTEM_MODE(SEMI_AUTOMATIC);
+SYSTEM_THREAD(ENABLED);
+STARTUP(System.enable(SYSTEM_FLAG_PM_DETECTION));
+
+retained uint8_t resetRetry = 0;
+#define ETHERNET_RETRY_MAX (10)
+
+#define CMD_SERIAL Serial1
```
technobly: Change to `Serial`

technobly (approved):

hal/network/ncp/cellular/cellular_hal.cpp
scott-brust: I dont think there are any defaults we could pick. This data is read from the modem, so we cannot guess what it may be. 
The intent here is 
1. If the modem is powered off, check if we have cached modem information to avoid turning on the modem
2. If there is no cached data, or the data is of an older format. Delete the data. The next time the modem is on and `cellular_device_info()` is called, we will update the cache then

system/src/system_cloud_connection_posix.cpp
```diff
@@ -14,14 +14,19 @@
  * You should have received a copy of the GNU Lesser General Public
  * License along with this library; if not, see <http://www.gnu.org/licenses/>.
  */
-
+#undef LOG_COMPILE_TIME_LEVEL
+#define LOG_COMPILE_TIME_LEVEL LOG_LEVEL_ALL
+#include "logging.h"
+LOG_SOURCE_CATEGORY("comm.cloud.posix")
```
scott-brust: TODO: remove before merging

system/src/system_connection_manager.h
technobly: So that there is a more reliable amount of data used for this tester, a predetermined count of packets seems reasonable?  If you do it based on time alone, you will have more or less packets depending on how fast the networks are at that given time, but more packets than a certain amount will likely not contribute to a better test value.  Still though you might have some really slow networks and a timeout will be required.

system/src/system_network_diagnostics.cpp
scott-brust: There is `HAL_PLATFORM_NCP_COUNT` but that will include Tracker platforms. Perhaps a brand new define like `HAL_PLATFORM_AUTOMATIC_CONNECTION_MANAGEMENT`?

hal/network/ncp/cellular/cellular_hal.cpp
```diff
@@ -199,6 +230,20 @@ int cellular_device_info(CellularDevice* info, void* reserved) {
     if (info->size >= offsetof(CellularDevice, radiofw) + sizeof(CellularDevice::radiofw)) {
         CHECK(client->getFirmwareVersionString(info->radiofw, sizeof(info->radiofw)));
     }
+
+    // Update the cached values if they dont match the last queried data. 
+    CellularDeviceCached cacheWrite = {};
+    cacheWrite.size = sizeof(CellularDeviceCached);
+    cacheWrite.version = CELLULAR_DEVICE_VERSION_V1;
```
avtolstoy: `CELLULAR_DEVICE_VERSION`

system/src/system_cloud_internal.h
```diff
@@ -194,6 +195,10 @@ class CloudConnectionSettings {
 
     static CloudConnectionSettings* instance();
 
+    network_interface_t &boundInterface() {
```
avtolstoy: @scott-brust Let's use getter/setter pattern and not expose references/pointers to internals if possible.

system/src/system_cloud.cpp
```diff
@@ -324,6 +331,25 @@ int spark_get_connection_property(unsigned property, void* data, size_t* size, v
...
+#if HAL_PLATFORM_AUTOMATIC_CONNECTION_MANAGEMENT
+        if (*size >= sizeof(network_interface_t)) {
+            *((network_interface_t*)data) = ConnectionManager::instance()->getCloudConnectionNetwork();
+            return 0;    
+        }
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+#else
+        return SYSTEM_ERROR_NOT_SUPPORTED;
```
avtolstoy: This could instead default to `NETWORK_INTERFACE_ALL`

wiring/inc/spark_wiring_cloud.h
avtolstoy, line 370: This should also possibly check current cloud network interface and if they don't match perform a disconnect?

system/src/system_task.cpp
```diff
@@ -794,6 +802,52 @@ void* system_internal(int item, void* reserved)
     case 3: {
         return reinterpret_cast<void*>(system_cloud_get_socket_handle());
     }
+#if PLATFORM_ID != PLATFORM_GCC && PLATFORM_ID != PLATFORM_NEWHAL
```
avtolstoy: This should be removed from normal release builds. If we need some of these for testing, let's wrap them under some normally-disabled compile-time macro, otherwise add a set of unstable APIs

system/src/system_network_diagnostics.h
```diff
@@ -22,8 +22,23 @@
...
+
 class NetworkDiagnostics {
 public:
+    // FIXME: Share this enum on device service side?
+    enum NetworkInterface {
```
avtolstoy: As discussed, let's use `network_interface_index`-compatible identifiers with the 'fixed' numbers, so
```
    NETWORK_INTERFACE_ALL = 0,
    NETWORK_INTERFACE_LOOPBACK = 1,
    // NETWORK_INTERFACE_MESH = 2, // Deprecated
    NETWORK_INTERFACE_ETHERNET = 3,
    NETWORK_INTERFACE_CELLULAR = 4,
    NETWORK_INTERFACE_WIFI_STA = 5,
    NETWORK_INTERFACE_WIFI_AP = 6
}
```

system/src/system_connection_manager.h
avtolstoy: We are using this now as a replacement for the NTP-based internet test, so this should be enabled on all platforms, not only for ACM purposes.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+        pollSockets(pfds, socketCount, packetDataLength);    
+    }
+
+    // Step 5: Cleanup test metrics, close sockets, calculate updated diagnostics
+    testSuccessful = true;
+
+    return 0;
+}
+
+const Vector<ConnectionMetrics>* ConnectionTester::getConnectionMetrics(){
```
avtolstoy: This is for debug purposes, but again let's avoid returning references/pointers to internal members, especially ones that will cause thread safety issues.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+        case NETWORK_INTERFACE_WIFI_STA:
+            return "WiFi    ";
+#endif
+        default:
+            return "";
+    }
+}
+
+ConnectionManager::ConnectionManager() {
+    preferredNetwork_ = NETWORK_INTERFACE_ALL;
```
avtolstoy: Nit: move to constructor initializer list.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+
+    if (system_cloud_is_connected(nullptr) == 0) {
+        socklen_t len = sizeof(socketNetIfIndex);
+        sock_handle_t cloudSocket = system_cloud_get_socket_handle();
+        //int r = 
+        sock_getsockopt(cloudSocket, SOL_SOCKET, SO_BINDTODEVICE, &socketNetIfIndex, &len);
+        //LOG(TRACE, "getCloudConnectionNetwork %d : %lu", r, socketNetIfIndex);
+    }
+
+	return socketNetIfIndex;
```
avtolstoy: Indentation.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    if (preferredNetwork_ != NETWORK_INTERFACE_ALL && spark::Network.from(preferredNetwork_).ready()) {
+        LOG(TRACE, "Using preferred network: %lu", preferredNetwork_);
+        return preferredNetwork_;
+    }
+
+    // 3: If no preferred network, use the 'best' network based on criteria
+    // 3.1: Network is ready: ie configured + connected (see ipv4 routable hook)
+    // 3.2: Network has best criteria based on network tester stats (vector should be sorted in "best" order)
+    for (auto& i: *ConnectionTester::instance()->getConnectionMetrics()) {
+        if (spark::Network.from(i.interface).ready()) {
```
avtolstoy: Do not use wiring network classes in system. This should be `network_ready()`.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+
+    // TODO: Determine a specific interface to bind to, even in the default case.
+    // ie: How should we handle selecting a cloud connection when no interfaces are up/ready?
+    // We should have some historical stats to rely on and then bring that network up? 
+    return bestNetwork;
+}
+
+ConnectionTester::ConnectionTester() {
+    network_interface_t interfaceList[] = { 
+        NETWORK_INTERFACE_ETHERNET,
```
avtolstoy: Nit: should probably also guard with `NETWORK_INTERFACE_ETHERNET`

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    // TODO: Determine a specific interface to bind to, even in the default case.
+    // ie: How should we handle selecting a cloud connection when no interfaces are up/ready?
+    // We should have some historical stats to rely on and then bring that network up? 
+    return bestNetwork;
+}
+
+ConnectionTester::ConnectionTester() {
+    network_interface_t interfaceList[] = {
```
avtolstoy: Nit: const

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    
+    for (const auto& i: interfaceList) {
+        struct ConnectionMetrics interfaceDiagnostics = {};
+        interfaceDiagnostics.interface = i;
+        metrics_.append(interfaceDiagnostics);
+    }
+}
+
+ConnectionTester* ConnectionTester::instance() {
+    static ConnectionTester* tester = new ConnectionTester();
```
avtolstoy:
```
static ConnectionTester tester;
return &tester;
```

system/src/system_connection_manager.h
```diff
@@ -0,0 +1,102 @@
...
+namespace particle { namespace system {
+
+struct ConnectionMetrics {
+    network_interface_t interface;
+    int socketDescriptor;
+    uint8_t *txBuffer;
+    uint8_t *rxBuffer;
+    int testPacketSize;
+    uint32_t testPacketSequenceNumber;
+    int txPacketCount;
```
avtolstoy: Nit: use unsigned types where it makes sense.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+    metrics->txBuffer = txBuffer;
+    metrics->rxBuffer = rxBuffer;
+    return 0;
+};
+
+int ConnectionTester::sendTestPacket(ConnectionMetrics* metrics, int length) {
+    int r = 0;
+    // Only send a new packet if we have received the previous one
```
avtolstoy: We should still send another packet after some timeout probably.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+};
+
+int ConnectionTester::receiveTestPacket(ConnectionMetrics* metrics) {
+    int r = sock_recv(metrics->socketDescriptor, metrics->rxBuffer, metrics->testPacketSize, MSG_DONTWAIT);
+    if (r > 0) {
+        CHECK_TRUE(r == metrics->testPacketSize, SYSTEM_ERROR_BAD_DATA);
+
+        if (memcmp(metrics->rxBuffer, metrics->txBuffer, r)) {
+            // Did not receive the exact same message sent
+            LOG(ERROR, "test socket on interface %d did not receive the same echo data");
```
avtolstoy: Nit: turn some of these non-critical errors into warnings.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+        0x8000,                // Differentiate interfaces by epoch field
+        {},                    // Sequence number
+        0                      // Payload length
+    };
+
+    int headerLength = sizeof(msg);
+    int totalMessageLength = packetDataLength + headerLength;
+    //LOG(TRACE, "packetDataLength %u headerLength %d totalMessageLength %d", packetDataLength, headerLength, totalMessageLength);
+
+    msg.epoch |= metrics->interface;
```
avtolstoy, line 295: Nit/TODO: use the same 'fixed' `network_interface_index` universally on all platforms, same as in vitals/diagnostics.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    msg.length = packetDataLength;
+    uint32_t sequenceNumber = nativeToBigEndian(metrics->testPacketSequenceNumber);
+
+    msg.epoch = nativeToBigEndian(msg.epoch);
+    msg.length = nativeToBigEndian(msg.length);
+    memcpy(&msg.sequence_number, &sequenceNumber, sizeof(sequenceNumber));
+
+    // TODO: Make data random instead
```
avtolstoy:
```
#include "random.h"
Random rand;
rand.gen(metrics->txBuffer + sizeof(msg), packetDataLength);
```

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+    memset(metrics->txBuffer, 0xFF, totalMessageLength);    
+    memcpy(metrics->txBuffer, &msg, headerLength);
+    metrics->testPacketSize = totalMessageLength;
+    // LOG_DUMP(TRACE, metrics->txBuffer, totalMessageLength);
+    return 0;
+};
+
+int ConnectionTester::pollSockets(struct pollfd* pfds, int socketCount, int packetDataLength) {
+    int pollCount = sock_poll(pfds, socketCount, 1000);
```
avtolstoy: Should probably avoid using a timeout here to avoid blocking system loop.

system/src/system_connection_manager.h
```diff
@@ -0,0 +1,102 @@
...
+#include "hal_platform.h"
+
+#if HAL_PLATFORM_IFAPI
+
+#include "system_network.h"
+#include "spark_wiring_vector.h"
+
+namespace particle { namespace system {
+
+struct ConnectionMetrics {
```
avtolstoy, line 30: Nit/TODO: this should probably be a proper C++ class with constructor, reset method etc.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+
+            LOG(INFO,"%s sent %d packets, got %d packets, avg rtt: %d", 
+                netifToName(i.interface), 
+                i.txPacketCount, 
+                i.rxPacketCount,
+                i.avgPacketRoundTripTime);
+        }
+
+        sock_close(i.socketDescriptor);
+        i.socketDescriptor = -1;
```
avtolstoy: Nit/TODO: proper C++ class with constructor/copy construction/reset method etc.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+        return SYSTEM_ERROR_NETWORK;
+    }
+
+    SCOPE_GUARD({
+        cleanupSockets(testSuccessful);
+        netdb_freeaddrinfo(info);
+    });
+
+    int socketCount = 0;
+    struct pollfd* pfds = (pollfd*)malloc(sizeof(pollfd) * metrics_.size());
```
avtolstoy: Could also better use unique_ptr: `auto pfds = std::make_unique<pollfd[]>(metrics_.size());`

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+        free(pfds);
+    });
+    
+    // Step 2: Create, bind, and connect sockets for each network interface to test
+    for (struct addrinfo* a = info; a != nullptr; a = a->ai_next) {
+
+        // For each network interface to test, create + open a socket with the retrieved server address
+        // If any of the sockets fail to be created + opened with this server address, return an error
+        for (auto& connectionMetrics: metrics_) {
+            if (!spark::Network.from(connectionMetrics.interface).ready()) {
```
avtolstoy: Do not use wiring network classes here.

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+            r = sock_setsockopt(s, SOL_SOCKET, SO_BINDTODEVICE, &ifr, sizeof(ifr));
+            if (r) {
+                LOG(ERROR, "test socket=%d, failed to sock_setsockopt to IF %s, errno=%d", s, ifr.ifr_name, errno);
+                sock_close(s);
+                return SYSTEM_ERROR_NETWORK;
+            }
+
+            r = sock_connect(s, a->ai_addr, a->ai_addrlen);
+            if (r) {
+                LOG(ERROR, "test socket=%d, failed to connect to %s#%u, errno=%d", s, serverHost, serverPort, errno);
```
avtolstoy, line 421: socket fd leak. Use `NAMED_SCOPE_GUARD`

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,489 @@
...
+            pfds[socketCount].fd = connectionMetrics.socketDescriptor;
+            pfds[socketCount].events = (POLLIN | POLLOUT);
+            socketCount++;
+
+            CHECK(allocateTestPacketBuffers(&connectionMetrics));
+        }
+    }
+    
+    // Step 4: Send/Receive data on the sockets for the duration of the test time
+    unsigned packetDataLength = random(1, REACHABILITY_MAX_PAYLOAD_SIZE);
```
avtolstoy: TODO: ideally each packet should be random-sized.

system/src/system_task.cpp
scott-brust: Ill delete these additions before merging

system/src/system_network_diagnostics.h
scott-brust: Should this enum definition be in `system_defs.h` like the existing `network_interface_index` enum? 
And then use `NetworkInterface::NETWORK_INTERFACE_CELLULAR` etc instead of `NETWORK_INTERFACE_CELLULAR` ?

system/src/system_connection_manager.cpp
scott-brust: Right now the Internet test / Connection test effectively runs in a busy loop for `REACHABILITY_TEST_DURATION_MS` seconds (currently 2.5s). 
Should the test be run in its own thread to not block the system?

system/src/system_connection_manager.cpp
```diff
@@ -0,0 +1,491 @@
...
+    uint16_t tmpport = 0;
+
+    int r = SYSTEM_ERROR_NETWORK;
+    bool testSuccessful = false;
+
+    getCloudHostnameAndPort(&tmpport, tmphost, sizeof(tmphost));
+    snprintf(tmpserv, sizeof(tmpserv), "%u", tmpport);
+    LOG(TRACE, "Resolving %s#%s", tmphost, tmpserv);
+    // TODO: get addrinfo/server IP from DNS lookup using the specific interface? 
+    r = netdb_getaddrinfo(tmphost, tmpserv, &hints, &info);
```
scott-brust, line 359: This is becoming a problem when different interfaces DNS servers resolve the device service hostname to different IPs and we try to use the same IP to try to connect with all the interfaces.
IE Some situations im seeing are wifi+ethernet sockets will connect but cellular doesnt and vice versa.

scott-brust: For example
Cellular + WIFI DNS servers:
```
[2023-11-28 11:45:14.380] 0000204617 [app] INFO: DNS IP 8.8.4.4
[2023-11-28 11:45:14.380] 0000204621 [app] INFO: DNS IP 8.8.8.8
[2023-11-28 11:45:14.385] 0000204624 [app] INFO: DNS IP 192.168.1.1
```
Server IP Resolves to `3.222.143.118#5684`

Ethernet + Cellular + WIFI DNS servers:
```
[2023-11-28 11:45:51.616] 0000241854 [app] INFO: DNS IP 192.168.0.1
[2023-11-28 11:45:51.622] 0000241858 [app] INFO: DNS IP 8.8.4.4
[2023-11-28 11:45:51.622] 0000241861 [app] INFO: DNS IP 8.8.8.8
[2023-11-28 11:45:51.627] 0000241865 [app] INFO: DNS IP 192.168.1.1
```

Server IP resolves to a different address `3.215.122.234#5684`
For whatever reason, cellular sometimes cannot reach this IP, or the traffic is much slower than with its own resolved IP

system/src/system_connection_manager.cpp
scott-brust: I picked an arbitrary 500ms time for sending the next packet. We still block the system, but since this test is only ever run when we lose the cloud connection I think its acceptable for now

wiring/inc/spark_wiring_cloud.h
scott-brust: I dont see any easy way to disconnect and reconnect if the current cloud connection is on a different network here in the wiring layer. 
`spark_cloud_flag_connect()/spark_cloud_flag_disconnect()` appear to set a flag that is handled in in the system loop. Is there a way to get the system to disconnect immediately from the app/wiring layer?

avtolstoy: This should probably instead be handled internally in system. E.g. checking if we are connected and bind interface != currently used one and performing disconnect/reconnect.

We could also opt to leave this as-is for now for 5.6.0.

system/src/system_connection_manager.cpp
scott-brust: Should be resolved now

wiring/inc/spark_wiring_cloud.h
scott-brust: Ok, lets handle internally in next release after 5.6.0

PR #2693, opened 2023-09-12
[wiring] Add 64-bit string conversions

eberseth:
### Description

Add 64-bit integer types (`long long` and `unsigned long long`) support to the `particle::String` class.

### Steps to Test

* `tests/unit_tests/wiring/string`

### Example App

```cpp
void setup() {
  Serial.begin();
  constexpr uint64_t number64 {12345678901234ULL}; // 0x00000b3a_73ce2ff2
  auto str64 = String(number64, HEX);
  Serial.printlnf("Big number in hex: %s", str64.c_str());
}

void loop() {

}
```
Result
```
Big number in hex: b3a73ce2ff2
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/src/spark_wiring_string.cpp
+++ b/wiring/src/spark_wiring_string.cpp
@@ -31,6 +31,7 @@
 #include <limits.h>
 #include <ctype.h>
 #include <stdlib.h>
+#include <charconv>
 #include "string_convert.h"
 
 //These are very crude implementations - will refine later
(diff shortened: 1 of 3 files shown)


wiring/src/string_convert.cpp
```diff
@@ -109,6 +109,55 @@ __attribute__((weak)) char* utoa(unsigned a, char* buffer, int radix) {
     return ultoa(a, buffer, radix);
 }
 
+__attribute__((weak)) char* ulltoa(uint64_t value, char* buffer, int radix, char pad) {
```
avtolstoy, line 112: `std::to_chars` should work for both all (including 64-bit) types, we should probably get rid of the current Arduino implementations and just streamline `std::to_chars` usage.

avtolstoy:
@eberseth This doesn't solve `printf` et al problem (including things like `Log.info` etc). How about we patch that one as well? I think I have a branch somewhere that guts out `_printf_i` and `_vsfprintf_r` with `%ll*` and `%L*` support, the problem previously was with bugs in gcc linker, unable to override newlib libc. Should be working fine nowadays, I think.

avtolstoy:
@eberseth https://github.com/particle-iot/device-os/pull/2694

eberseth:
> @eberseth This doesn't solve `printf` et al problem (including things like `Log.info` etc). How about we patch that one as well? I think I have a branch somewhere that guts out `_printf_i` and `_vsfprintf_r` and augments with `%ll*` and `%L*` support, the problem previously was with bugs in gcc linker, unable to override newlib libc. Should be working fine nowadays, I think.

@avtolstoy You are right, it doesn't solve the `printf` and friends.  I think your suggestion about patching it sound great.

wiring/src/string_convert.cpp
eberseth: I can go along with that

eberseth: done

wiring/src/spark_wiring_string.cpp
```diff
@@ -153,6 +154,22 @@ String::String(unsigned long value, unsigned char base)
 	*this = buf;
 }
 
+String::String(long long value, unsigned char base)
+{
+	init();
+	char buf[66] {};  // Assuming a maximum of 64-bit binary representation + '-' + null terminator.
```
avtolstoy, line 160: Let's make this a constant. This should also be 24 for worst case scenario for 64-bit ints (including hex/octal and other cases).

Style: Use `buf[24] = {}` initialization. Same in other places.

wiring/src/spark_wiring_string.cpp
```diff
@@ -442,6 +473,20 @@ StringSumHelper & operator + (const StringSumHelper &lhs, unsigned long num)
 	return a;
 }
 
+StringSumHelper & operator + (const StringSumHelper &lhs, long long num)
+{
+	StringSumHelper &a = const_cast<StringSumHelper&>(lhs);
+	if (!a.concat(num)) a.invalidate();
```
avtolstoy, line 479: Style:
```c++
if (!a.concat(num)) {
    a.invalidate();
}
```
Same in other places.

wiring/src/spark_wiring_string.cpp
```diff
@@ -774,6 +819,17 @@ long String::toInt(void) const
 	return 0;
 }
 
+long long String::toLongLongInt(unsigned char base) const
+{
+	if (buffer) return strtoll(buffer, NULL, base);
```
avtolstoy, line 824: Nit/style: `nullptr`

wiring/src/spark_wiring_string.cpp
eberseth: done

wiring/src/spark_wiring_string.cpp
eberseth: done

wiring/src/spark_wiring_string.cpp
eberseth: done

eberseth:
@avtolstoy Implemented a bunch of whitespace corrections as well.  There were a mix of tabs and spaces for indent.

wiring/src/spark_wiring_string.cpp
```diff
@@ -153,6 +154,22 @@ String::String(unsigned long value, unsigned char base)
 	*this = buf;
 }
 
+String::String(long long value, unsigned char base)
+{
+	init();
+	char buf[66] {};  // Assuming a maximum of 64-bit binary representation + '-' + null terminator.
+	(void)std::to_chars(buf, buf + sizeof(buf) - 1, value, (int)base);
```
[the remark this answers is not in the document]
avtolstoy: Still in?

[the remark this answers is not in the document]
eberseth: just kidding, removed :)

avtolstoy (approved):

PR #2698, opened 2023-09-20
[gen3] hal: fixes the issue that UARTE RX may loss data.

XuGuohui:
Please check out: https://github.com/particle-iot/device-os/pull/2685

The main idea to fix this issue is to make sure the ENDRX event is handled ASAP so that the UARTE receiver is started in time to accept incoming data.

### Completeness

- [ ] Added documentation


avtolstoy (approved):
@XuGuohui I'm approving this and we are going to merge and release this as part of 4.2.0 even if we may still be seeing some potential problems. We can continue looking into them, but this should provide enough of an improvement to previously much more broken behavior.

PR #2701, opened 2023-09-28
[msom] support 16MB GD flash.

XuGuohui:
As the title describes.

This PR doesn't include the updating of memory map.

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/exflash_hal.cpp
+++ b/hal/src/rtl872x/exflash_hal.cpp
@@ -59,6 +59,7 @@ typedef enum {
     GENERIC_FLASH_CMD_DP            = 0xB9,
     GENERIC_FLASH_CMD_WREN          = 0x06,
     GENERIC_FLASH_CMD_WRDI          = 0x04,
+    GENERIC_FLASH_CMD_RDSR          = 0x05,
 } hal_exflash_cmd_t;
 
 #define HAL_QSPI_FLASH_OTP_SECTOR_COUNT_GD25        3
@@ -375,7 +376,7 @@ static bool isSecureOtpMode(uint32_t normalContent) {
 // GD25 flash specific definitions
 
 #define RTK_SPIC_FIFO_SIZE                  64
-#define	RTK_SPIC_FIFO_HALF_SIZE	            (RTK_SPIC_FIFO_SIZE / 2)
+#define RTK_SPIC_FIFO_HALF_SIZE             (RTK_SPIC_FIFO_SIZE / 2)
 
 #define RTK_SPIC_BIT_OFFSET_AUTO_LEN_DUM    ((uint32_t)0)
 #define RTK_SPIC_BIT_WIDTH_AUTO_LEN_DUM     ((uint32_t)16)
@@ -408,6 +409,18 @@ enum RtkSpicFlashDataWidth {
     RTK_SPIC_DATA_WORD = 2
 };
 
+static void rtkSpiFlashSetDataReg(uint32_t index, uint32_t data, enum RtkSpicFlashDataWidth byteWidth) {
+    if (index > 31) {
+        return;
+    } else if (byteWidth == RTK_SPIC_DATA_BYTE) {
+        SPIC->dr[index].byte = (data & 0x000000ff);
+    } else if (byteWidth == RTK_SPIC_DATA_HALF) {
+        SPIC->dr[index].half = (data & 0x0000ffff);
+    } else if (byteWidth == RTK_SPIC_DATA_WORD) {
+        SPIC->dr[index].word = data;
+    }
+}
+
 static uint32_t rtkSpiFlashGetDataReg(uint32_t index, enum RtkSpicFlashDataWidth byteWidth) {
     uint32_t data;
     if (index > 31) {
@@ -424,6 +437,16 @@ static uint32_t rtkSpiFlashGetDataReg(uint32_t index, enum RtkSpicFlashDataWidth
     return data;
 }
 
+static void rtkSpiSetRxMode() {
+    SPIC->ctrlr0 = (SPIC->ctrlr0 | 0x0300);
+    SPIC->ctrlr0 = (SPIC->ctrlr0 & 0xfff0ffff);
+}
+
+static void rtkSpiSetTxMode() {
+    SPIC->ctrlr0 = SPIC->ctrlr0 & 0xfffffcff;
+    SPIC->ctrlr0 = (SPIC->ctrlr0 & 0xfff0ffff);
+}
+
 static void rtkSpiFlashWaitBusy() {
     while (1) {
         if (RTK_SPIC_BIT_GET_UNSHIFTED(SPIC->sr, RTK_SPIC_BIT_OFFSET_SR_TXE)) {
@@ -433,6 +456,23 @@ static void rtkSpiFlashWaitBusy() {
             break;
         }
     }
+}
+
+static uint8_t gd25GetStatus() {
+    /* Disable SPI_FLASH */
+    SPIC->ssienr = 0;
+    /* Set Ctrlr1; 1 byte data frames */
+    SPIC->ctrlr1 = 1;
+    /* Set tuning dummy cycles */
+    RTK_SPIC_BITS_SET_VAL(SPIC->auto_length, RTK_SPIC_BIT_OFFSET_AUTO_LEN_DUM, RTK_SPIC_READ_TUNING_DUMMY_CYCLE, RTK_SPIC_BIT_WIDTH_AUTO_LEN_DUM);
+    /* set ctrlr0: RX_mode */
+    rtkSpiSetRxMode();
+    /* set flash_cmd: write cmd to fifo */
+    rtkSpiFlashSetDataReg(0, GENERIC_FLASH_CMD_RDSR, RTK_SPIC_DATA_BYTE);
+    /* Enable SPI_FLASH */
+    SPIC->ssienr = 1;
+    rtkSpiFlashWaitBusy();
+    return rtkSpiFlashGetDataReg(0, RTK_SPIC_DATA_BYTE);
 }
 
 static void rtkSpiFlashEnableCsRead(uint32_t len) {
@@ -445,7 +485,25 @@ static void rtkSpiFlashEnableCsRead(uint32_t len) {
     rtkSpiFlashWaitBusy();
 }
 
-static void rtkSpiFlashSendCmdAddr(uint8_t cmd, uint32_t addr, uint32_t dummyCycle) {
+static void rtkSpiFlashEnableCsWriteErase() {
+    /* Enable SPI_FLASH */
+    SPIC->ssienr = 1;
+    rtkSpiFlashWaitBusy();
+    /* Check flash is in write progress or not */
+    while ((gd25GetStatus() & 0x1)) {}
+}
+
+static void rtkSpiFlashSendCmd(uint8_t cmd) {
+    /* Disble SPI_FLASH */
+    SPIC->ssienr = 0;
+    /* set ctrlr0: TX mode */
+    rtkSpiSetTxMode();
+    /* set flash_cmd: wren to fifo */
+    rtkSpiFlashSetDataReg(0, cmd, RTK_SPIC_DATA_BYTE);
+    rtkSpiFlashEnableCsWriteErase();
+}
+
+static void rtkSpiFlashRxCmdAddr(uint8_t cmd, uint32_t addr, uint32_t dummyCycle) {
     /* Disable SPI_FLASH */
     SPIC->ssienr = 0;
     if (dummyCycle != 0) { // spi_flash_set_dummy_cycle(dev, dummy);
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +370,298 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+#define DW_BIT_GET_UNSHIFTED(__datum, __bfws)                               \
+    ((uint32_t) ((__datum) & DW_BIT_MASK(__bfws)))
+
+enum spi_flash_byte_num {
+    DATA_BYTE = 0,
+    DATA_HALF = 1,
+    DATA_WORD = 2
+};
+
+__attribute__((section(".ram.text"), noinline))
```
avtolstoy: Any difference if we run out of RAM or PSRAM?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +370,298 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+    CHECK_TRUE(security_registers_addr + size <= HAL_QSPI_FLASH_OTP_SIZE_GD25, SYSTEM_ERROR_INVALID_ARGUMENT);
+
+    uint32_t address = ((security_registers_index + 1) << 12) + security_registers_addr;
+    uint8_t addrBuf[3] = {};
+    addrBuf[0] = (uint8_t)((address >> 16) & 0xFF);
+    addrBuf[1] = (uint8_t)((address >> 8) & 0xFF);
+    addrBuf[2] = (uint8_t)((address >> 0) & 0xFF);
+
+    // We may need to improve this
+    auto ptr = (uint8_t*)malloc(size + 3);
```
avtolstoy: Memory leak?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -400,32 +673,62 @@ int hal_exflash_write_special(hal_exflash_special_sector_t sp, uintptr_t addr, c
...
+            return SYSTEM_ERROR_INVALID_ARGUMENT;
+        }
+        uint32_t address = ((otpPage + 1) << 12);
+
+        // Erase security registers
+        uint8_t buf[3] = {};
+        buf[0] = (uint8_t)((address >> 16) & 0xFF);
+        buf[1] = (uint8_t)((address >> 8) & 0xFF);
+        buf[2] = (uint8_t)((address >> 0) & 0xFF);
+        FLASH_TxCmd(0x44, sizeof(buf), buf);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -400,32 +673,62 @@ int hal_exflash_write_special(hal_exflash_special_sector_t sp, uintptr_t addr, c
...
+        FLASH_WaitBusy(WAIT_WRITE_DONE);
+        
+        // Write/Erase disable
+        FLASH_TxCmd(0x04, 0, nullptr);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -400,32 +673,62 @@ int hal_exflash_write_special(hal_exflash_special_sector_t sp, uintptr_t addr, c
...
 }
 
 #endif // MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
 
 int hal_exflash_erase_special(hal_exflash_special_sector_t sp, uintptr_t addr, size_t size) {
-    /* We only support OTP sector and since it's One Time Programmable, we can't erase it */
+    if (flash_init_para.FLASH_Id == FLASH_ID_GD) {
+        ExFlashLock lk(false); // Stop thread scheduler
+        // Write/Erase enable
+        FLASH_TxCmd(0x06, 0, nullptr);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +370,298 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+    // We may need to improve this
+    auto ptr = (uint8_t*)malloc(size + 3);
+    if (!ptr) {
+        return SYSTEM_ERROR_NO_MEMORY;
+    }
+    memcpy(ptr, addrBuf, 3);
+    memcpy(ptr + 3, buf, size);
+
+    // Write/Erase enable
+    FLASH_TxCmd(0x06, 0, nullptr);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +370,298 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+    FLASH_WaitBusy(WAIT_WRITE_EN);
+
+    FLASH_TxCmd(0x42, size + 3, ptr);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +370,298 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+    FLASH_WaitBusy(WAIT_WRITE_DONE);
+    
+    // Write/Erase disable
+    FLASH_TxCmd(0x04, 0, nullptr);
```
avtolstoy: magic number

hal/src/rtl872x/exflash_hal.cpp
XuGuohui: Nope. These new functions should always be called after boot up. Would you like to remove the restriction to save more RAM sapce?

hal/src/rtl872x/exflash_hal.cpp
XuGuohui: Good catch!

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Just to avoid unnecessary attributes for readability. If they work fine out of regular PSRAM-copied `.text`, I'd just leave them there.

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -400,32 +665,62 @@ int hal_exflash_write_special(hal_exflash_special_sector_t sp, uintptr_t addr, c
...
+            return SYSTEM_ERROR_INVALID_ARGUMENT;
+        }
+        uint32_t address = ((otpPage + 1) << 12);
+
+        // Erase security registers
+        uint8_t buf[3] = {};
+        buf[0] = (uint8_t)((address >> 16) & 0xFF);
+        buf[1] = (uint8_t)((address >> 8) & 0xFF);
+        buf[2] = (uint8_t)((address >> 0) & 0xFF);
+        FLASH_TxCmd(GD_FLASH_CMD_ERS_SEC, sizeof(buf), buf);
```
scott-brust: Do we even want to implement the ability to erase GD OTP data?

XuGuohui: I would leave it there. Because it's not our first time that the OTP info is programmed with wrong data in factory. With MXIC flash we have to replace the flash with a brand new one. But with GD flash it gives us a chance to re-program the OTP info without reworking the hardware.

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -400,32 +665,62 @@ int hal_exflash_write_special(hal_exflash_special_sector_t sp, uintptr_t addr, c
...
+        FLASH_WaitBusy(WAIT_WRITE_DONE);
+        
+        // Write/Erase disable
+        FLASH_TxCmd(GENERIC_FLASH_CMD_WRDI, 0, nullptr);
```
scott-brust: Should we return `SYSTEM_ERROR_NONE` after this command? 
It looks like execution will continue past the `if (flash_init_para.FLASH_Id == FLASH_ID_GD) {}` block and return ERROR NOT SUPPORTED?

hal/src/rtl872x/exflash_hal.cpp
scott-brust: Ok, makes sense. 
As currently implemented, this will erase the whole 1024 byte register of the given address, right? 
If we have to use this, I guess its ok to erase everything and start over?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +372,288 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+
+    FLASH_TxCmd(GD_FLASH_CMD_PGM_SEC, size + 3, ptr);
+    FLASH_WaitBusy(WAIT_WRITE_DONE);
+    
+    // Write/Erase disable
+    FLASH_TxCmd(GENERIC_FLASH_CMD_WRDI, 0, nullptr);
+    return SYSTEM_ERROR_NONE;
+}
+
+static int gd25PerformSecurityRegistersOperation(Gd25SecurityRegistersOperation operation, uintptr_t addr, uint8_t* data_buf, size_t data_size) {
```
scott-brust, line 654: Does this function, and functions called within it, (ie `gd25ReadSecurityRegisters`/`gd25WriteSecurityRegisters`) also need to also be declared in the `.ram.text` section?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +372,288 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+    if (!ptr) {
+        return SYSTEM_ERROR_NO_MEMORY;
+    }
+    SCOPE_GUARD({
+        if (ptr) {
+            free(ptr);
+        }
+    });
+    memcpy(ptr, addrBuf, 3);
+    memcpy(ptr + 3, buf, size);
```
scott-brust: We need to allocate an intermediate buffer because the address bytes + data need to be contiguous when sent with `FLASH_TxCmd()`?
Does the data need to be in heap / PSRAM or not?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +372,288 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+
+    /* setting auto mode CS_H_WR_LEN, address length(3 byte) */
+    SPIC->auto_length = ((SPIC->auto_length & 0xfffcffff) | 0x00030000);
+    SPIC->addr_length = 3;
+
+    rtkSpiFlashSendCmdAddr(cmd, address, dummyCycle);
+
+    uint32_t cnt;
+    uint32_t data;
+    if (size <= RTK_SPIC_FIFO_HALF_SIZE) { // using RTK_SPIC_FIFO_SIZE it won't work if size >32
```
scott-brust: Do we want to limit reads to 32 bytes and get rid of the else block? Then do repeated 32 byte reads as needed?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +372,288 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+        }
+    });
+    memcpy(ptr, addrBuf, 3);
+    memcpy(ptr + 3, buf, size);
+
+    // Write/Erase enable
+    FLASH_TxCmd(GENERIC_FLASH_CMD_WREN, 0, nullptr);
+    FLASH_WaitBusy(WAIT_WRITE_EN);
+
+    FLASH_TxCmd(GD_FLASH_CMD_PGM_SEC, size + 3, ptr);
```
scott-brust: ⚠️ I seem to fault when writing >=10 bytes of data. 9 works fine, but anything more fails

scott-brust (commented):
The main issue is that I cant seem to get writes >9 bytes to succeed. Can you reproduce the same behavior?

hal/src/rtl872x/exflash_hal.cpp
XuGuohui: We don't have to, as all these functions are executed in PSRAM.

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,31 +372,288 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+            memcpy(buf, &data, 4);
+            buf += 4;
+        }
+        /* t is transfer data times,
+         * t = (size - first xfer data) / fifo size
+         */
+        uint32_t t = (size - RTK_SPIC_FIFO_HALF_SIZE) / RTK_SPIC_FIFO_HALF_SIZE;
+        address = address + RTK_SPIC_FIFO_HALF_SIZE;
+        while (t > 0) {
+            rtkSpiFlashSendCmdAddr(cmd, address, dummyCycle);
```
scott-brust: This looks like when we want to read more than 32 bytes of data in total, we need to do it with repeated 32 byte read commands (ie we have to re-send the read command `0x48` and the incrementing address)?
Seems strange this is how we would have to do this since the data sheet implies that we can continue reading and the address should just increment

XuGuohui: I think this is the limitation of how the realtek's SPIC works.

scott-brust (approved):

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -364,71 +373,409 @@ static bool isSecureOtpMode(uint32_t normalContent) {
...
+#define RTK_SPIC_BIT_GET_UNSHIFTED(reg, offset)                                     \
+    ((uint32_t)((reg) & RTK_SPIC_BIT_MASK(offset)))
+
+enum RtkSpicFlashDataWidth {
+    RTK_SPIC_DATA_BYTE = 0,
+    RTK_SPIC_DATA_HALF = 1,
+    RTK_SPIC_DATA_WORD = 2
+};
+
+static void rtkSpiFlashSetDataReg(uint32_t index, uint32_t data, enum RtkSpicFlashDataWidth byteWidth) {
```
avtolstoy, line 412: Nit: instead of static might be better to just move to an anonymous namespace.

avtolstoy (approved):

PR #2704, opened 2023-10-20
[msom] Support burnin tests

scott-brust:
### Problem

Factory burn in needs to support MSoM hardware, specifically the GPIO pinout

### Solution

Adds MSoM DVT pinmap to GPIO tests. 
Basic cellular test: turn on modem, query IMEI/ICCID/Modem firmware
GNSS test: turn on modem, configure for GNSS, poll for a fix for up to 90s. If there is no lock after 90s, the test fails

### Steps to Test

Jumper the following pins together on the M2 EVB (based off the silkscreen names):
```
CTS   <--> RTS
CS    <--> MISO
MOSI  <--> SCK
GPIO0 <--> GPIO1
PWM0  <--> PWM1
PWM2  <--> PWM3
SOM6  <--> SOM7
A0    <--> A1
A2    <--> A3


Jumper wire for these two sets:
EVT HARDWARE:
SOM5 <--> SOM12
A4     <--> SOM11

DVT HARDWARE:
SOM5 <--> SOM12
A4   <--> A7
```
Remove all of the Ethernet and SD card jumpers

### Example App

Run Tinker. To force burnin mode change this line in tinker setup
```
BurninTest::instance()->setup();  -->  BurninTest::instance()->setup(true);
```

### References
Ask scott for more info on gpio pins

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/applications/tinker/src/burnin_test.cpp
+++ b/user/applications/tinker/src/burnin_test.cpp
@@ -57,6 +57,7 @@ BurninTest::BurninTest() {
 		&particle::BurninTest::testCpuLoad,
 #if PLATFORM_ID == PLATFORM_MSOM
 		&particle::BurninTest::testCellularModem,
+		&particle::BurninTest::testGnss,
 #endif
 	};
 
@@ -68,7 +69,8 @@ BurninTest::BurninTest() {
 		"SPI_FLASH", 
 		"CPU_LOAD",
 #if PLATFORM_ID == PLATFORM_MSOM
-		"CELL_MODEM"
+		"CELL_MODEM",
+		"GNSS"
 #endif
 	};
 }
@@ -101,8 +103,12 @@ void BurninTest::setup(bool forceEnable) {
 	}
 
 	Particle.disconnect();
-
-    logger_ = std::make_unique<Serial1LogHandler>(115200, LOG_LEVEL_INFO);
+	Cellular.off();
+	LogCategoryFilters burninFilters = {
+        	{ "ncp.at", LOG_LEVEL_TRACE },
+        	{ "app", LOG_LEVEL_INFO }
+	};
+    logger_ = std::make_unique<Serial1LogHandler>(115200, LOG_LEVEL_INFO, burninFilters);
     
 	// Detect if backup SRAM has a failed test in it (IE state is "TEST FAILED")
 	Log.info("BURN IN START: ResetReason: %d State: %d ErrorMessage: %s", System.resetReason(), (int)BurninState, BurninErrorMessage);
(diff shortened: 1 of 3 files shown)


user/applications/tinker/src/burnin_test.cpp
```diff
@@ -459,5 +473,147 @@ bool BurninTest::testCpuLoad() {
...
+	// EXAMPLE:
+	// $<TalkerID>GGA,<UTC>,<Lat>,<N/S>,<Lon>,<E/W>,<Quality>,<NumSatUsed>,<HDOP>,<Alt>,M,<Sep>,M,<DiffAge>,<DiffStation>*<Checksum><CR><LF>
+	// +QGPSGNMEA: $GPGGA,213918.00,3804.405282,N,12209.922544,W,1,05,3.0,145.9,M,-26.4,M,,*5E
+
+    //Log.info("%s", buf);
+
+	*gnssLocked = false;
+
+	const int MAX_GPGGA_STR_LEN = 256;
+    char gpggaSentence[MAX_GPGGA_STR_LEN] = {};
```
scott-brust, line 518: Id like to fix the mix of tabs/space (change all tabs to 4 spaces) in all the fqc/burnin files. 
Should I just do a whitespace/formatting only commit or will that make this too difficult to review?

eberseth: I think we can hide whitespace changes.  Go ahead and even squash commits if you want.

eberseth (approved):

user/applications/tinker/src/burnin_test.cpp
```diff
@@ -459,5 +473,147 @@ bool BurninTest::testCpuLoad() {
...
+    device.size = sizeof(device);
+    if (cellular_device_info(&device, NULL)) {
+		strcpy(BurninErrorMessage, "Failed to get cellular info");
+    	return false;
+    }
+    
+    Log.info("Cell modem ICCID: %s IMEI: %s FW: %s", device.iccid, device.imei, device.radiofw);
+
+	Cellular.off();
+	waitFor(Cellular.isOff, 30000);
```
technobly: Do we want to check if it turns off?  Maybe add `turnModemOff()`?  It could also not explicitly fail the test with `false` if it doesn't turn off.

user/applications/tinker/src/burnin_test.cpp
technobly: We can hide whitespace changes so please do make a commit for whitespace changes, and also no need to squash it together.

user/applications/tinker/src/burnin_test.cpp
```diff
@@ -459,5 +473,147 @@ bool BurninTest::testCpuLoad() {
...
+
+	// If no lock in X minutes, fail
+	if (!gnssLocked) {
+		strcpy(BurninErrorMessage, "No GNSS lock after 90s");
+	}
+
+	// Turn off Cell modem + GNSS antenna
+	digitalWrite(GNSS_ANT_PWR, LOW);
+	Cellular.off();
+	waitFor(Cellular.isOff, 30000);
```
technobly: `turnModemOff()` ?  It could also not explicitly fail the test with `false` if it doesn't turn off.

technobly (approved):

user/applications/tinker/src/burnin_test.cpp
scott-brust: Its not required that the modem be off for the test to be considered successful. This is just an attempt to maintain state after the test is complete. If the modem fails to power on for subsequent cell/gnss tests, that is a legitimate failure though.

PR #2710, opened 2023-11-15
[rtl872x] fix multiple BLE issues

XuGuohui:
### Problem

1. BLE scanning may stall. BLE scanning API acquires the BLE lock. When error occured in BLE stack during scanning, it tries to reset the stack, which then sends a command to a queue for an internal thread to process. If the thread is trying to acquire the BLE lock to process other command in queue before handling the command to reset the stack, then deadlock occurs.
2. Due to reset of the BLE stack, for whatever reason, all of the on-going BLE activities are stopped. This typically causes the BLE provisioning mode not functioning well.
3. BLE states's event may be notified in messed order.

### Solution

1. Unlock the BLE lock before resetting the BLE stack if required.
2. Cache BLE current states and restore them after the BLE stack is reset successfully.
3. Do not completely rely on the notified state event. Use SDK API to check the state before waiting the target state is accomplished. 

### Steps to Test

Build and run the attached user application.

### Example App

```c
#include "Particle.h"

// delay after which to enable provisioning mode
// set to a very high value to effectively disable
#define PROVISION_MODE_DELAY (10)
// #define PROVISION_MODE_DELAY (100000)

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);

void scan_result_cb(const BleScanResult *result, void *context) {
    Log.info("\tgot advert");
}

void scan_thread_f() {
    BLE.on();

    // short repeated scans for improved chance to catch all sensor adverts
    // (each scan will only return the first advert from a device so we can't
    // use long-running scan)
    BLE.setScanTimeout(50);

    // disable active scan
    // only get advert, don't request scan response for less overhead and
    // hopefully reduced power consumption on sensors
    BleScanParams params = {};
    params.version = BLE_API_VERSION;
    params.size = sizeof(BleScanParams);
    BLE.getScanParameters(params);
    params.active = false;
    BLE.setScanParameters(params);

    while (true) {
        Log.info("start scan");
        unsigned int t0 = millis();
        BLE.scan(scan_result_cb, nullptr);
        unsigned int t1 = millis();
        if (t1 - t0 > 500) {
            Log.info("\tscan took too long: %dms", t1 - t0);
        }
        Log.info("\tend scan");
        delay(1);
    }
}

void setup() {
    // while (!Serial.isConnected()) {
        // delay(100);
    // }
    static Thread *thread = new Thread("sensor_ble_scan", scan_thread_f, OS_THREAD_PRIORITY_DEFAULT);
}

void loop() {
    static bool provision_once = true;
    static uint32_t last_sec = System.uptime();
    uint32_t now_sec = System.uptime();

    if (provision_once && now_sec > PROVISION_MODE_DELAY) {
        provision_once = false;

        // ---------Setup device name---------
        char buf[16] = {0};
        char name[BLE_MAX_DEV_NAME_LEN+1];
        auto len = hal_get_device_serial_number(buf, sizeof(buf), nullptr);
        // name gets last 6 of the serial number to be (mostly) unique
        snprintf(name, sizeof(name), "SmartTub-%s", &(buf[len-6]));
        Log.info("before BLE.setDeviceName");
        BLE.setDeviceName(name);
        Log.info("after BLE.setDeviceName");

        // ---------Set company ID---------
        Log.info("before BLE.setProvisioningCompanyId");
        BLE.setProvisioningCompanyId(0x1234);
        Log.info("after BLE.setProvisioningCompanyId");

        // ---------BLE provisioning mode---------
        Log.info("before BLE.provisioningMode");
        BLE.provisioningMode(true);
        Log.info("after BLE.provisioningMode");
    }

    if (now_sec - last_sec >= 10) {
        last_sec = now_sec;
        Log.info("loop: getProvisioningStatus: %d", BLE.getProvisioningStatus());
    }
}
```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -497,7 +497,7 @@ private:
     struct BleGapCache {
         bool isAdv;
         bool isScan;
-        bool isconn;
+        bool isConn;
     };
 
     struct BleConnection {
@@ -1163,7 +1163,7 @@ int BleGap::start() {
     LOCAL_DEBUG("wait GAP_INIT_STATE_STACK_READY");
     CHECK(waitState(BleGapDevState().init(GAP_INIT_STATE_STACK_READY), BLE_STATE_DEFAULT_TIMEOUT, true /* force poll */));
     btStackStarted_ = true;
-    if (cache_.isAdv || cache_.isconn) {
+    if (cache_.isAdv || cache_.isConn) {
         LOG(TRACE, "Restore advertising state");
         startAdvertising(); // Despite of the result.
     }
@@ -1180,7 +1180,7 @@ int BleGap::stop(bool restore) {
     if (restore) {
         cache_.isAdv = isAdvertising_;
         cache_.isScan = isScanning_;
-        cache_.isconn = connectedAsBlePeripheral();
+        cache_.isConn = connectedAsBlePeripheral();
     } else {
         cache_ = {};
     }
@@ -1646,7 +1646,7 @@ int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
 
     SCOPE_GUARD ({
         if (isScanning_) {
-            const int LE_SCAN_STOP_RETRIES = 1;
+            const int LE_SCAN_STOP_RETRIES = 10;
             for (int i = 0; i < LE_SCAN_STOP_RETRIES; i++) {
                 // This has seen failing a number of times at least
                 // with btgap logging enabled. Retry a few times just in case,
@@ -2216,9 +2216,15 @@ void BleGap::handleDevStateChanged(T_GAP_DEV_STATE newState, uint16_t cause) {
                         newState.gap_init_state, newState.gap_adv_state, newState.gap_adv_sub_state, newState.gap_scan_state, newState.gap_conn_state, cause);
     RtlGapDevState nState;
     nState.state = newState;
+    // In case of race condition in stack
+    RtlGapDevState s;
+    if (le_get_gap_param(GAP_PARAM_DEV_STATE, &s.state) == GAP_CAUSE_SUCCESS && nState.raw != s.raw) {
+        LOCAL_DEBUG("!!!!!!!!! Race condition. Notified: %02X, Get: %02X", nState.raw, s.raw);
+        nState.state = s.state;
+    }
     // NOTE: this event is generated before the connection is established
-    if (newState.gap_adv_state != state_.state.gap_adv_state && newState.gap_adv_state == GAP_ADV_STATE_IDLE) {
-        if (newState.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_CONN) {
+    if (nState.state.gap_adv_state != state_.state.gap_adv_state && nState.state.gap_adv_state == GAP_ADV_STATE_IDLE) {
+        if (nState.state.gap_adv_sub_state == GAP_ADV_TO_IDLE_CAUSE_CONN) {
             isAdvertising_ = false; // adv timer has higher priority, just in case that the BLE_CMD_STOP_ADV_NOTIFY
                                     // has been enqueued, followed by nontifying the ADV stopped event. Under
                                     // certain circumstance notifying the ADV stopped event will cause BLE re-adv,
@@ -2229,12 +2235,6 @@ void BleGap::handleDevStateChanged(T_GAP_DEV_STATE newState, uint16_t cause) {
         }
     }
     state_.raw = nState.raw;
-    // In case of race condition in stack
-    RtlGapDevState s;
-    if (le_get_gap_param(GAP_PARAM_DEV_STATE, &s.state) == GAP_CAUSE_SUCCESS && nState.raw != s.raw) {
-        LOCAL_DEBUG("!!!!!!!!! Race condition. Notified: %02X, Get: %02X", nState.raw, s.raw);
-        state_.raw = s.raw;
-    }
     os_semaphore_give(stateSemaphore_, false);
 }
 
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -151,6 +151,15 @@ namespace {
     res;                                                                        \
 })
 
+#if 0
+#define LOCAL_DEBUG(_fmt, ...) \
+        do { \
+            LOG(TRACE, _fmt "\r\n", ##__VA_ARGS__); \
+        } while (false)
+#else
+#define LOCAL_DEBUG(_fmt, ...)
+#endif
+
 constexpr uint32_t BLE_OPERATION_TIMEOUT_MS = 60000;
 constexpr system_tick_t BLE_ENQUEUE_TIMEOUT_MS = 5000;
 
@@ -234,7 +243,7 @@ private:
     void* allEvtQueue_;
     void* ioEvtQueue_;
     bool started_;
-    static constexpr uint8_t BLE_EVENT_THREAD_PRIORITY = 3; // Higher than application thread priority, which is 2
+    static constexpr uint8_t BLE_EVENT_THREAD_PRIORITY = OS_THREAD_PRIORITY_NETWORK;
     static constexpr uint8_t MAX_NUMBER_OF_GAP_MESSAGE = 0x20;
     static constexpr uint8_t MAX_NUMBER_OF_IO_MESSAGE = 0x20;
     static constexpr uint8_t MAX_NUMBER_OF_EVENT_MESSAGE = (MAX_NUMBER_OF_GAP_MESSAGE + MAX_NUMBER_OF_IO_MESSAGE);
@@ -352,7 +361,7 @@ private:
 class BleGap {
 public:
     int start();
-    int stop();
+    int stop(bool restore = true);
     int init();
 
     bool initialized() const {
@@ -443,6 +452,14 @@ public:
         return SYSTEM_ERROR_NONE;
     }
 
+    bool isQueueAvailable() const {
+        uint8_t command;
+        if (!os_queue_peek(cmdQueue_, &command, 0, nullptr)) {
+            return true;
+        }
+        return false;
+    }
+
     void lockMode(bool lock) {
         bleInLockedMode_ = lock;
     }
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -477,6 +494,12 @@ class BleGap {
         BLE_CMD_START_ADV
     };
 
+    struct BleGapCache {
+        bool isAdv;
+        bool isScan;
+        bool isconn;
```
avtolstoy: Nit: `isConn`

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1585,13 +1646,14 @@ int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
 
     SCOPE_GUARD ({
         if (isScanning_) {
-            const int LE_SCAN_STOP_RETRIES = 10;
+            const int LE_SCAN_STOP_RETRIES = 1;
```
avtolstoy: Remove loop entirely if no longer required? I think when we ran into this previously a few retries did help, but it was always very difficult to reliably reproduced without logging enabled.

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1616,8 +1681,13 @@ int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
     CHECK_RTL(le_scan_start());
     isScanning_ = true;
     // GAP_SCAN_STATE_SCANNING may be propagated immediately following the GAP_SCAN_STATE_START
-    if (waitState(BleGapDevState().scan(GAP_SCAN_STATE_SCANNING))) {
-        return SYSTEM_ERROR_TIMEOUT;
+    if (waitState(BleGapDevState().scan(GAP_SCAN_STATE_START), 10, true) && waitState(BleGapDevState().scan(GAP_SCAN_STATE_SCANNING), 10, true)) {
```
avtolstoy, line 1684: 10ms sounds like a very tight condition?

hal/src/rtl872x/ble_hal.cpp
XuGuohui: I can't reproduce this. But if you did observe the retries help, let's keep it as-is at cost of 100ms postponed.

hal/src/rtl872x/ble_hal.cpp
XuGuohui: That's designated. I just utilize the `waitState()` to do a quick check before a long wait.

scott-brust: Is it critical to call `BleGapDevState().scan(GAP_SCAN_STATE_START)` before `BleGapDevState().scan(GAP_SCAN_STATE_SCANNING)`? 
That seems to be the only real change here? 

Like you mention, we still call `waitState(BleGapDevState().scan(GAP_SCAN_STATE_SCANNING))` for the full wait duration after this

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1005,19 +1030,28 @@ void BleGap::bleCommandThread(void *context) {
     BleGap* gap = (BleGap*)context;
     while (true) {
         uint8_t command;
-        if (!os_queue_take(gap->cmdQueue_, &command, CONCURRENT_WAIT_FOREVER, nullptr)) {
+        if (!os_queue_peek(gap->cmdQueue_, &command, CONCURRENT_WAIT_FOREVER, nullptr)) {
```
scott-brust, line 1033: What is the effective difference between `peek` on the queue rather than `take`?

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1953,17 +2023,13 @@ int BleGap::setPairingConfig(const hal_ble_pairing_config_t* config) {
...
-        if (ok && adv) {
-            startAdvertising();
-        }
-    });
     if (btStackStarted_) {
         CHECK(stop());
         CHECK(init());
     }
+    SCOPE_GUARD({
+        start();
```
scott-brust, line 2031: This is no longer `CHECK()`'d. Could `start()` fail and the error be not returned to the caller of `setPairingConfig()`? Or is this not a concern?

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2104,22 +2166,42 @@ bool BleGap::isPaired(hal_ble_conn_handle_t connHandle) {
...
             if (BleGapDevState(s).copySetParamsFrom(state) == state) {
                 return SYSTEM_ERROR_NONE;
             }
             HAL_Delay_Milliseconds(BLE_WAIT_STATE_POLL_PERIOD_MS);
+            now = hal_timer_millis(nullptr);
         }
     } else {
-        auto end = hal_timer_millis(nullptr) + timeout;
-        for (auto now = hal_timer_millis(nullptr); now < end; now = hal_timer_millis(nullptr)) {
+        for (now = now; now < end; now = hal_timer_millis(nullptr)) {
```
scott-brust, line 2202: What is the intent behind `now = now`?

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2147,6 +2229,12 @@ void BleGap::handleDevStateChanged(T_GAP_DEV_STATE newState, uint16_t cause) {
         }
     }
     state_.raw = nState.raw;
+    // In case of race condition in stack
+    RtlGapDevState s;
+    if (le_get_gap_param(GAP_PARAM_DEV_STATE, &s.state) == GAP_CAUSE_SUCCESS && nState.raw != s.raw) {
+        LOCAL_DEBUG("!!!!!!!!! Race condition. Notified: %02X, Get: %02X", nState.raw, s.raw);
+        state_.raw = s.raw;
```
scott-brust: My understanding of this function is that we get a state change notification from the BLE Event Dispatcher, which includes the new BLE state. Is this correct?

This new behavior here is to also query the BLE stack for the actual state. If the queried state differs from the notification state, something is out of sync. Its possible the BLE stack changed states again between the original notification and now. Is this correct?

If we are also updating `state_` with the newly queried value, should we also act on the potential changes like we do above? IE check `newState.gap_adv_state != state_.state.gap_adv_state`  etc.

scott-brust (commented):
I have some questions about general behavior. 

I do see these errors in the app somewhat frequently:
```
0016113700 [app] INFO: start scan
0016113721 [app] INFO:  got advert
0016113747 [app] INFO:  got advert
0016113825 [app] INFO:  got advert
0016113849 [app] INFO:  got advert
0016114006 [app] INFO:  got advert
0016114134 [app] INFO:  got advert
0016114213 [hal.ble] TRACE: Failed to stop scanning, resetting stack
0016114216 [hal.ble] TRACE: Going to stop the stack...
0016114227 [app] INFO:  got advert
0016114243 [app] INFO:  got advert
0016114248 [app] INFO:  got advert
0016114587 [net.lwip_rltk] INFO: promisc_deinit TODO
0016114589 [hal] INFO: WiFi off
0016114607 [hal] INFO: rltk_wlan_set_netif_info: 0, 94:94:4a:04:29:f8
0016114944 [hal] INFO: WiFi on
0016115277 [hal.ble] TRACE: Restore advertising state
0016115285 [app] INFO:  scan took too long: 1584ms
0016115287 [app] INFO:  end scan
```
The app continues to scan / advertise so I assume this is expected?

I also see this pattern frequently:
```
0016229913 [app] INFO: start scan
0016230041 [app] INFO:  got advert
0016230141 [app] INFO:  got advert
0016230352 [app] INFO:  got advert
0016230353 [app] INFO:  got advert
0016230437 [app] INFO:  scan took too long: 523ms
0016230439 [app] INFO:  end scan
```
Is the 50ms BLE scan timeout per beacon? 
I see the time difference between some of the individual scan results is >50ms, is this expected?

hal/src/rtl872x/ble_hal.cpp
XuGuohui: >Is it critical to call BleGapDevState().scan(GAP_SCAN_STATE_START) before BleGapDevState().scan(GAP_SCAN_STATE_SCANNING)?

Nope. Just treat it as a check (`force` poll is set to `true`) without waiting any time (although there is 10ms). The `GAP_SCAN_STATE_START` and `GAP_SCAN_STATE_SCANNING` sometimes are notified in wrong sequence.

XuGuohui:
>The app continues to scan / advertise so I assume this is expected?

Yes.

> 0016230437 [app] INFO:  scan took too long: 523ms

The scan timeout can be changed in user app through `BLE.setScanTimeout(50);`, where the actual timeout is `50 * 10`ms. So that's expected result.

hal/src/rtl872x/ble_hal.cpp
XuGuohui: In `int BleGap::stop(bool restore)` you'll see this code snippet:
```
            if (isQueueAvailable()) {
                LOCAL_DEBUG(" >>>>>>>>>>>> s_bleMutex.unlock() <<<<<<<<<<<<<<");
                s_bleMutex.unlock();
            }
```
This is going to fix a deadlock in BLE driver. Using `peek` will set `isQueueAvailable()` to let us know if there is command that might be acquiring the BLE lock and need to be processed and unlock the lock accordingly.

hal/src/rtl872x/ble_hal.cpp
XuGuohui: If `start()` failed, then the stack is resided in an invalid state and the stack will try restarting again when next BLE API that depends on the stack is called. So it's fine to not check the resut. Also I didn't ever notice the failure of the `start()` so far.

hal/src/rtl872x/ble_hal.cpp
XuGuohui: This is just to make the code more readable. I did observe that due to thread scheduling if `now` is set in `for()`, the time elapsed between it and when `end` is set will make `now < end` false. Note that in BLE scan method we have `if (waitState(BleGapDevState().scan(GAP_SCAN_STATE_START), 10, true) && waitState(BleGapDevState().scan(GAP_SCAN_STATE_SCANNING), 10, true)) {`, which sets the timeout to 10ms. It's easily to make `now < end` false so that we may lose a chance to check the state.

hal/src/rtl872x/ble_hal.cpp
XuGuohui: #1  Correct 
#2 I don't think the BLE stack state is changed in between. It's most likely that, for example, normally the advertising state changes in the sequence of `GAP_ADV_STATE_ADVERTISING -> GAP_ADV_STATE_STOP -> GAP_ADV_STATE_IDLE`, but sometimes the events are notified in wrong sequence `GAP_ADV_STATE_ADVERTISING -> GAP_ADV_STATE_IDLE -> GAP_ADV_STATE_STOP`. The double check is just to correct the internal cached `state_`.

And yes, we should first check the polled state before the check routine above.

scott-brust (approved):

PR #2711, opened 2023-11-17
Asset OTA fixes

avtolstoy:
### Description

This PR introduces a number of Asset OTA related fixes/enhancements:

1. `CTRL_REQUEST_GET_ASSET_INFO` control request to fetch information about required/available assets on device over BLE/USB
2. `CTRL_REQUEST_GET_MODULE_INFO` refactored to align with new protobuf-based System Describe and encoding for both the control request and system describe coalesced into a single implementation to save on flash space
3. `System.assetsAvailable()` hides extra assets that might still be in storage but are not currently required, essentially this is now `available ∩ required`
4. `seek()` forward and backwards support in wiring Asset OTA APIs
5. [Factory application with asset dependencies is now correctly parsed](https://app.shortcut.com/particle/story/123329)
6. Overwriting an asset in use by the application (due to OTA or local flashing) is handled safely

### Testing

- `ota/assets`

NOTE: Requires latest `@particle/device-os-test-runner`

### Related

- https://github.com/particle-iot/particle-usb/pull/96
- https://github.com/particle-iot/device-os-protobuf/pull/22


system/src/system_info_encoding.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    cb->arg = this;
+    cb->funcs.encode = [](pb_ostream_t* strm, const pb_field_iter_t* field, void* const* arg) {
+        const auto self = (const EncodeFirmwareModules*)*arg;
+        for (unsigned i = 0; i < self->sysInfo_.module_count; ++i) {
+            const auto& module = self->sysInfo_.modules[i];
+            if (!is_module_function_valid((module_function_t)module.info.module_function)) {
+                continue;
+            }
+            // If sending a describe, skip factory modules entirely just in case
```
avtolstoy, line 118: This will be reverted once this is safe.

system/src/system_info.cpp
```diff
@@ -558,70 +490,8 @@ bool system_module_info_pb(appender_fn appender, void* append_data, void* reserv
...
-        }
-        return true;
-    };
-#if HAL_PLATFORM_ASSETS
-    auto availableAssets = AssetManager::instance().availableAssets();
-    if (availableAssets.size() > 0) {
-        pbDesc.assets.arg = (void*)&availableAssets;
-        pbDesc.assets.funcs.encode = encodeAssetDependencies;
-    }
+    EncodeAssets assets(&pbDesc.assets, AssetManager::instance().availableAssets());
```
monkbroc, line 494: I think this should use `availableAndRequiredAssets()` to avoid showing assets still on the device but not required by the application on the Console.

system/src/control/storage.cpp
```diff
@@ -191,74 +173,24 @@ int firmwareUpdateDataRequest(ctrl_request* req) {
...
+    EncodeFirmwareModules modules(&pbRep.modules, EncodeFirmwareModules::Flag::FORCE_SHA);
     CHECK(encodeReplyMessage(req, PB(GetModuleInfoReply_fields), &pbRep));
     return 0;
 }
 
+#if HAL_PLATFORM_ASSETS
+int getAssetInfo(ctrl_request* req) {
+    PB(GetAssetInfoReply) pbRep = {};
+    using namespace particle::system;
+    EncodeAssets assetsAvailable(&pbRep.available, AssetManager::instance().availableAssets());
```
monkbroc, line 187: Similarly I think this should use `availableAndRequiredAssets()` for consistency with the describe and Wiring API. The CLI flashes the new application before assets so returning available and required assets in the control request will work.

monkbroc (approved):
Looks great! I tested the control requests. I left a comment about making the list of available assets consistent between describe, control request and System.assetsAvailable

keeramis (approved):
I tested `available` and `required` works as expected. 
Tested `assetsAvailable()` return the assets which are available and required 👍 
Checked `getFirmwareModuleInfo()` and `getAssetInfo()` work as expected 👍 

I did not test the app crashing issue.

system/src/system_info.cpp
avtolstoy: This is for cloud: we want the DS to be aware of what is entirely on the device to avoid OTAing things over unnecessarily.

system/src/control/storage.cpp
avtolstoy: I'd disagree. I'd like for system describe and control requests to represent the full picture and for wiring APIs to be representative of what's required by the application. Again, same thing: to avoid re-flashing things unnecessarily that are already on the device.

system/src/system_info.cpp
avtolstoy: As discussed before the "use-case" for this is:
1. product fw version 1 assets (1,2,3)
2. product fw version 2 assets (2,3)
3. product fw version 3 assets (1,2,3)

Currently apart from the product application OTA (and depending on whether Device OS upgrade happened of course) the assets in this particular case will not get OTAd at all post #1.

monkbroc: Assets wouldn't get flashed unnecessarily if this used `availableAndRequiredAssets()` either since after 1, it would return assets 1, 2, 3 available, after 2, it would return 2, 3 available and after 3, it would return 1, 2, 3. The OTA planner only triggers after the device sends the new describe message. But since it works both ways it's fine like this.

PR #2715, opened 2023-12-11
Ledger API and sync

sergeuz:
### Problem

This PR supersedes https://github.com/particle-iot/device-os/pull/2660 and https://github.com/particle-iot/device-os/pull/2675.

Below are some links to simplify navigation in the new code:

- [Ledger application API](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/wiring/inc/spark_wiring_ledger.h)
- [Ledger system API](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/system/inc/system_ledger.h)
- [Ledger storage layer](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/system/src/ledger/ledger.cpp)
- [Ledger state machine](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/system/src/ledger/ledger_manager.cpp)
- [Synchronization protocol](https://github.com/particle-iot/device-os-protobuf/blob/main/cloud/ledger.proto)
- [New CoAP API](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/communication/inc/coap_api.h)
- [Variant class](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/wiring/inc/spark_wiring_variant.h)
- [Map class](https://github.com/particle-iot/device-os/blob/ledger-sync-merged/sc-120056/wiring/inc/spark_wiring_map.h)

### Steps to Test

See the [validation plan](https://www.notion.so/particle/Ledger-Validation-7b64402038734c28af40b2531da8a323).

The integration tests in this PR require the device's product to have the following ledger definitions:

- `test-device-to-cloud` – type: Device to Cloud.
- `test-cloud-to-device` – type: Cloud to Device; scope: Device.

### References

- sc-118798
- sc-119316
- sc-120056
- sc-122990
- sc-124092


sergeuz:
@avtolstoy @monkbroc As discussed, marking this as ready for review even though I'm still debugging the deadlock issue and need to add a few integration tests.

avtolstoy:
I've freed up some space on Tracker and re-enabled Ledger feature flag for it.

communication/inc/coap_channel_new.h
```diff
@@ -0,0 +1,154 @@
...
+#include "ref_count.h"
+
+namespace particle::protocol {
+
+class CoapMessageDecoder;
+class Protocol;
+
+namespace experimental {
+
+// This class implements the new experimental protocol API that allows the system to interact with
```
monkbroc, line 36: Can we remove the experimental namespace? I'd rather have an explanation like "this supercedes the CoAP API xyz and should be used for new message types" than one that says "new and experimental". It won't always be new and if we're using it for real messages, it's not experimental anymore.

services/inc/ref_count.h
```diff
@@ -0,0 +1,148 @@
...
+ * Smart pointer for reference counted objects.
+ */
+template<typename T>
+class RefCountPtr {
+public:
+    RefCountPtr() :
+            RefCountPtr(nullptr) {
+    }
+
+    RefCountPtr(T* ptr) :
```
monkbroc, line 67: I get the semantics of this constructor and the copy constructor incrementing the reference count, and the destructor decreasing it. I don't understand the semantics of `wrap` and `unwrap` and why this class needs both interfaces.

monkbroc: Looking at more of the code, `wrap` and `unwrap` are used to convert between the the smart pointer and the raw pointer type, sometimes when passing the pointer to a container that deals with opaque types. That makes sense, but the usage is hard to follow because many uses of `unwrap()` are paired with a call to `release()` on the reference counted object. Ideally, every use of `wrap()` would be paired with a user of `unwrap()` and there would be no need to call `release()` directly the same way there is no need to call `addRef()` directly.

^^ The above paragraph may not be 100% accurate, but it's a note that it's hard to reason about the ownership and safety of the smart pointers above. Usually I'd look for matching pairs calls to `new/delete` or `malloc/free` in add/remove functions. Here's it's harder to do this because `wrap/unwrap` and `addRef/release` don't match.

communication/src/coap_channel_new.cpp
```diff
@@ -0,0 +1,1457 @@
...
+
+template<typename T, typename E>
+inline void addRefToList(T*& head, RefCountPtr<E> elem) {
+    addToList(head, elem.unwrap());
+}
+
+template<typename T, typename E>
+inline void removeRefFromList(T*& head, const RefCountPtr<E>& elem) {
+    removeFromList(head, elem.get());
+    elem->release();
```
monkbroc, line 161: It looks odd that you have to manually release the `RefCount` here. Is there a way to make it symmeric with `addRefToList` by using `RefCountPtr::wrap` and letting the RAII class `RefCountPtr` call `release` on the inner pointer when it goes out of scope?

communication/src/coap_channel_new.cpp
```diff
@@ -0,0 +1,1457 @@
...
+    }
+    auto h = findInList(connHandlers_, [=](auto h) {
+        return h->callback == callback;
+    });
+    if (!h) {
+        std::unique_ptr<ConnectionHandler> h(new(std::nothrow) ConnectionHandler(callback, callbackArg));
+        if (!h) {
+            return SYSTEM_ERROR_NO_MEMORY;
+        }
+        addToList(connHandlers_, h.release());
```
monkbroc, line 677: What's the advantage of using a unique_ptr here since you're immediately passing ownership to the list?

communication/src/coap_channel_new.cpp
```diff
@@ -0,0 +1,1457 @@
...
+    }
+    return size;
+}
+
+void CoapChannel::destroyMessage(coap_message* apiMsg) {
+    if (!apiMsg) {
+        return;
+    }
+    auto msg = RefCountPtr<CoapMessage>::wrap(reinterpret_cast<CoapMessage*>(apiMsg)); // Take ownership
+    clearMessage(msg);
```
monkbroc, line 593: This is the pattern I would expect in `removeRefFromList`. First take ownership of the reference counted object, remove it from the list, then let the smart pointer go out of scope to decrement the reference count.

communication/inc/coap_channel_new.h
sergeuz: Happy to use a different name if you have a better suggestion. `experimental` was chosen for two reasons:
- It immediately catches an eye when used in the old CoAP code where the new implementation is sometimes called in a way that breaks the separation between the existing CoAP layers. `experimental` helps to indicate that some of those changes are temporary.
- There are just too many _channel_ classes in the comms library already and isolating this one in its own namespace seemed a good way to reduce ambiguity.

services/inc/ref_count.h
sergeuz: Julien and I discussed this at length. My main takeaway from the feedback is that `release()` and `addRef()` are considered a low-level API that should not ideally be used in C++ code for which there's a smart pointer class `RefCountPtr`. I agree with that, however, I think there's at least one exception where direct usage of `release()` can be justified: if there's a reference to an object that is _not managed_ by a smart pointer in the first place, it's should be okay to call `release()` on it as it would, arguably, express the intent in a more direct and readable way as opposed to wrapping the pointer via `RefCountPtr::wrap()` and letting the resulting smart pointer immediately go out of scope.

In the code in question, `release()` is only called on objects that either came through the C-based dynalib interface or are stored in an intrusive list container where it's not possible to use smart pointers. For those objects there's always a "dangling" reference and `release()` destroys it in an explicit way.

With that said, this is quite a subjective topic. I'd be okay replacing all occurrences of `release()` with calls to `wrap()` if we find the latter more readable.

system/src/ledger/ledger_manager.cpp
```diff
@@ -0,0 +1,1599 @@
...
+            ctx->syncPending = info.syncPending();
+            if (ctx->syncDir == LEDGER_SYNC_DIRECTION_CLOUD_TO_DEVICE) {
+                ctx->lastUpdated = info.lastUpdated();
+            }
+            // Context objects are sorted by ledger name
+            bool found = false;
+            auto it = findContext(contexts, ctx->name, found);
+            assert(!found);
+            it = contexts.insert(it, std::move(ctx));
+            if (it == contexts_.end()) {
```
monkbroc: Is this supposed to be `contexts.end()`? This whole loop is working with a local collection of sync contexts.

communication/src/coap_channel_new.cpp
sergeuz: I'm just being defensive here: if someone else adds another error condition before the ownership is passed to the list, they won't have a chance to leak the memory because it's already managed by a smart pointer.

system/src/ledger/ledger_manager.cpp
sergeuz: Good catch! This code used to populate the global `contexts_` previously.

system/src/ledger/ledger_manager.cpp
```diff
@@ -0,0 +1,1599 @@
...
+    return 0;
+}
+
+int LedgerManager::removeLedgerData(const char* name) {
+    if (!*name) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    std::lock_guard lock(mutex_);
+    // TODO: Allow removing ledgers regardless of the device state or if the given ledger is in use
+    if (state_ != State::OFFLINE) {
```
monkbroc, line 441: This will be important to document:
To remove the Ledger data from a device (for example, during development or to simulate the first cloud connection), go offline `Particle.disconnect()` then purge the Ledgers `Ledger::removeAll()`

Correct this if it's wrong.

sergeuz: Just disconnecting from the cloud won't be enough if the application has valid references to the ledger instances being deleted. The most reliable way would probably be to delete ledgers early on startup, before a cloud connection is initiated and any ledger instances are created.

This is quite a limitation so I think removing it should be one of the first improvements we do the ledger API once this PR is merged.

system/src/ledger/ledger_manager.cpp
```diff
@@ -0,0 +1,1599 @@
...
+    retryDelay_ = 0;
+    state_ = State::OFFLINE;
+}
+
+int LedgerManager::receiveRequest(CoapMessagePtr& msg, int reqId) {
+    if (state_ < State::READY) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    // Get the request type. XXX: It's assumed that the message fields are encoded in order of their
+    // field numbers, which is not guaranteed by the Protobuf spec in general
```
monkbroc, line 573: Can you add a test in https://github.com/particle-iot/device-os-protobuf/blob/main/src/index.test.js to make sure this remains true?

sergeuz: That's a good idea. There are integration tests in this PR that validate this indirectly but we only run them in production environment.

system/src/ledger/ledger_manager.cpp
```diff
@@ -0,0 +1,1599 @@
...
+    }
+    writer->updateInfo(LedgerInfo().lastUpdated(lastUpdated));
+    stream_.reset(writer.release());
+    msg_ = std::move(msg);
+    // Read the first chunk of the ledger data
+    CHECK(receiveLedgerData());
+    return 0;
+}
+
+int LedgerManager::receiveSubscribeResponse(CoapMessagePtr& msg, int result) {
```
monkbroc, line 822: What would happen if a response went missing? For example, you send a subscribe request and the server sends the subscribe response but it gets lost along the way. Is the timeout mechanism at the CoAP layer? I noticed the calls to `coap_begin_request` have timeout set to 0. I'm not sure if it means "use default timeout" or "no timeout".

sergeuz: Yeah, the old CoAP layer would take care of that and close the connection with a timeout error. 0 means the [default timeout](https://github.com/particle-iot/device-os/blob/07308ada47f7344250e7a06a281ca3d006a5f7f4/communication/inc/coap_api.h#L261) but at present it's ignored since the server only uses piggybacked CoAP responses for which only the ACK timeouts apply.

system/src/ledger/ledger_manager.cpp
```diff
@@ -0,0 +1,1599 @@
...
+    return 0;
+}
+
+int LedgerManager::receiveSetDataResponse(CoapMessagePtr& /* msg */, int result) {
+    assert(state_ == State::SYNC_TO_CLOUD && curCtx_ && curCtx_->syncDir == LEDGER_SYNC_DIRECTION_DEVICE_TO_CLOUD &&
+            curCtx_->taskRunning);
+    if (result == 0) {
+        LOG(TRACE, "Sent ledger data: %s", curCtx_->name);
+        LedgerInfo newInfo;
+        auto now = CHECK(getMillisSinceEpoch());
```
monkbroc, line 727: If the RTC clock was invalid here (or in other methods in this file), would the operation get retried in a subsequent run of the state machine, or would the state machine get stuck in State::SYNC_TO_CLOUD?

sergeuz: If you take a look at the CoAP callbacks that drive most of this state machine's logic, they all transition the manager to the `FAILED` state on errors and then call into the `run()` method asynchronously that will retry the sync eventually. In this case specifically, the error from `getMillisSinceEpoch()` will be propagated to `responseCallback()` which has the above logic implemented.

wiring/src/spark_wiring_ledger.cpp
```diff
@@ -0,0 +1,434 @@
...
+        callback(std::move(ledger), arg);
+    });
+}
+
+int Ledger::onSync(OnSyncFunction callback) {
+    if (!isValid()) {
+        return Error::INVALID_STATE;
+    }
+    return setSyncCallback(instance_, std::move(callback));
+}
```
monkbroc, line 384: On the topic of making things consistent, would we want to make all the methods to provide callbacks in the user application consistent using the pattern above? We have several patterns in the Wiring layer for callbacks, like Particle.subscribe and System.on. Could we mark some of the patterns like passing a pointer to a method and a pointer to an instance deprecated and tell developers to use lambdas? We can update the public docs as well.

sergeuz: I have a minor concern related to the fact that `std::function` is not exception-safe so, strictly speaking, it should be avoided in embedded code, but otherwise I'm not opposed to the suggested change in our approach to how callbacks are managed. It would also be nice to have a discussion about the default OOM handling behavior in general as part of making that decision.

wiring/inc/spark_wiring_variant.h
```diff
@@ -0,0 +1,1058 @@
...
+            return areEqual(first, second);
+        }
+    };
+
+    template<typename T>
+    struct IsAlternativeType {
+        static constexpr bool value = false;
+    };
+
+    VariantType v_;
```
monkbroc, line 800: Tiny nit: can you move this to just below the definition of `VariantType` for readability?

wiring/src/spark_wiring_ledger.cpp
sergeuz: Btw, I simplified the code that registers the `onSync` callback following your feedback in one the previous PRs. Hope it's clearer now.

wiring/inc/spark_wiring_map.h
```diff
@@ -0,0 +1,501 @@
...
+
+    friend void swap(Map& map1, Map& map2) {
+        using std::swap; // For ADL
+        swap(map1.entries_, map2.entries_);
+        swap(map1.cmp_, map2.cmp_);
+    }
+
+private:
+    Vector<Entry> entries_;
+    CompareT cmp_;
```
monkbroc, line 498: Does `cmp_` have to be an instance variable or could it be a compile time template parameter? It seems like the map should only have to store its entries. This would bring the storage down for `VariantMap` and potentially make `Variant` more space efficient.

wiring/inc/spark_wiring_variant.h
```diff
@@ -0,0 +1,1058 @@
...
+     */
+    static Variant fromJSON(const JSONValue& val);
+
+    friend void swap(Variant& var1, Variant& var2) {
+        using std::swap; // For ADL
+        swap(var1.v_, var2.v_);
+    }
+
+private:
+    typedef std::variant<std::monostate, bool, int, unsigned, int64_t, uint64_t, double, String, VariantArray, VariantMap> VariantType;
```
monkbroc: `sizeof(Variant) = 24` for now so it's worth documenting that an array of 32 bit integers will take 6x the memory when stored as a variant array.

monkbroc (approved):
I'm happy to mark my review as approved on this large amount of work! :clap:

communication/inc/coap_channel_new.h
monkbroc: OK. You can make the final decision here on wheter to change and to what.

services/inc/ref_count.h
monkbroc: I'm OK with keeping the usage of `release()` in those cases. I'd prefer additional comments around the ownership changes because of the bad consequences of getting it wrong for future maintainers.

system/src/ledger/ledger_manager.cpp
monkbroc: It doesn't sound too bad in practice as removing a Ledger from a device isn't a normal thing that a program has to do.
Did you end up adding a Wiring API to declare which Ledgers are used by an application, so the cleanup of unused Ledgers is automatic?

system/src/ledger/ledger_manager.cpp
monkbroc: Great!

wiring/src/spark_wiring_ledger.cpp
monkbroc: Yes, I like the simplication. Thank you.

wiring/inc/spark_wiring_map.h
monkbroc: It's not worth doing now since the `VariantType` would stay at 24 bytes due to `String`

system/src/ledger/ledger_manager.cpp
sergeuz: Yep, there's `Particle.useLedgers()` method that does that: https://github.com/particle-iot/device-os/blob/07308ada47f7344250e7a06a281ca3d006a5f7f4/wiring/inc/spark_wiring_cloud.h#L460-L472

communication/inc/coap_channel_new.h
sergeuz: I'll keep `experimental` until we have proper unit tests for this API.

services/inc/ref_count.h
sergeuz: Added more comments for now.

communication/src/coap_channel_new.cpp
sergeuz: Discussed [here](https://github.com/particle-iot/device-os/pull/2715#discussion_r1468230265).

communication/src/coap_channel_new.cpp
sergeuz: Discussed [here](https://github.com/particle-iot/device-os/pull/2715#discussion_r1468230265).

system/src/ledger/ledger_manager.cpp
sergeuz: Fixed.

system/src/ledger/ledger_manager.cpp
sergeuz: Added a test in device-os-protobuf: https://github.com/particle-iot/device-os-protobuf/pull/29.

wiring/inc/spark_wiring_variant.h
sergeuz: Fixed.

wiring/inc/spark_wiring_map.h
sergeuz: > Does cmp_ have to be an instance variable or could it be a compile time template parameter?

I don't think it does as long as we don't want to support comparator objects that can be configured at runtime or are expensive to construct. I can't think of any other reason why it's stored as a member variable in STL which I used as a reference.

wiring/inc/spark_wiring_variant.h
sergeuz: cc @rickkas7

wiring/inc/spark_wiring_map.h
sergeuz: Let's keep it like that for now given that we can't optimize memory usage for variants by removing it.

system/src/ledger/ledger.cpp
```diff
@@ -0,0 +1,891 @@
...
+}
+
+int Ledger::updateInfo(const LedgerInfo& info) {
+    std::lock_guard lock(*this);
+    // Open the file with the current ledger data
+    char path[MAX_PATH_LEN + 1];
+    CHECK(getCurrentFilePath(path, sizeof(path), name_));
+    FsLock fs;
+    lfs_file_t file = {};
+    // TODO: Rewriting parts of a file is inefficient with LittleFS. Consider using an append-only
```
avtolstoy, line 380: This particular use-case of only modifying from the end should be fine for the most part though.

avtolstoy (approved):
Great job! There are a few future changes to Device OS that we might make to make things a bit better/simpler as discussed, hope we get to that at some point.

I did review in detail lock usage in `LedgerManager` and `Ledger`, which is probably the most likely source of problems, so far looks good, but might consider refactoring that at some point a bit just to avoid introducing problems in the future when making an accidental change.

sergeuz:
> I did review in detail lock usage in LedgerManager and Ledger, which is probably the most likely source of problems, so far looks good, but might consider refactoring that at some point a bit just to avoid introducing problems in the future

@avtolstoy Yeah, I had to debug quite a few race conditions around that code and regretted not making `LedgerReader` and `LedgerWriter` trivial wrappers over the file handle that would simply forward all API calls to the parent ledger instance, so that we could at least simplify locking between ledger streams and instances. I'll consider that refactoring when I work on this code again.

hal/src/rtl872x/concurrent_hal.cpp
```diff
@@ -434,6 +434,7 @@ int os_semaphore_give(os_semaphore_t semaphore, bool reserved)
  */
 int os_timer_create(os_timer_t* timer, unsigned period, void (*callback)(os_timer_t timer), void* const timer_id, bool one_shot, void* reserved)
 {
+    // TODO: Return an error if the period is 0 (see https://www.freertos.org/FreeRTOS-timers-xTimerCreate.html)
```
eberseth, line 437: @avtolstoy  Should we submit a bug to fix in device OS on separate PR?

proto_defs/internal/ledger.proto
```diff
@@ -0,0 +1,30 @@
...
+
+package particle.firmware;
+
+import "cloud/ledger.proto";
+import "nanopb.proto";
+
+/**
+ * Ledger info.
+ */
+message LedgerInfo {
```
eberseth, line 11: Is there a need for a sequence number in the unlikely event of out-of-order Ledger sync because of slow transfer networks and modem quirks?

wiring/inc/spark_wiring_variant.h
```diff
@@ -0,0 +1,1058 @@
...
+     * Supported alternative types.
+     */
+    enum Type {
+        NULL_, ///< Null type (`std::monostate`).
+        BOOL, ///< `bool`.
+        INT, ///< `int`.
+        UINT, ///< `unsigned`.
+        INT64, ///< `int64_t`.
+        UINT64, ///< `uint64_t`.
+        DOUBLE, ///< `double`.
```
eberseth, line 84: All `float` types will have to be promoted?

test/unit_tests/wiring/variant.cpp
```diff
@@ -0,0 +1,474 @@
...
+        {
+            Variant v(123ll);
+            checkVariant<int64_t>(v, 123);
+        }
+        {
+            Variant v(123ull);
+            checkVariant<uint64_t>(v, 123);
+        }
+        {
+            Variant v(123.0f);
```
eberseth, line 207: Check also for NaN.  I know people who use that value regularly in code to signify uninitiated values.

proto_defs/internal/ledger.proto
sergeuz: There's `update_count` that gets incremented every time the ledger is modified but it's only used internally by Device OS so that it knows which snapshot of ledger data exactly is being synchronized at a given time. It was introduced to work around a subtle race condition around [this code](https://github.com/particle-iot/device-os/blob/59325889b628f591756c40c9cc5fa33dc73c05a9/system/src/ledger/ledger.cpp#L878-L882) possible when a device-to-cloud ledger is modified by the user while it's already being synchronized with the cloud. Without the counter, the ledger manager may prematurely clear the persistent `syncPending` flag and if the device happens to reset around the same time, the last change to the ledger wouldn't be synchronized when the device restarts until the ledger is modified again.

Out-of-order syncs are not possible because only one sync can be in progress at a time. In the worst case, the system may synchronize the same ledger data twice after recovering from a disconnect that happened at the end of the previous sync attempt.

wiring/inc/spark_wiring_variant.h
sergeuz: Yes. This class isn't very efficient for storing values of primitive types in the first place so it doesn't make much sense to support all possible variations of those types. 32- and 64-bit integers are handled separately because 32-bit arithmetic is done in hardware which is generally not the case for `float` and `double`.

test/unit_tests/wiring/variant.cpp
sergeuz: Having a test for NaN won't hurt but I don't see how it's different from any other floating point value in this context. We're not testing serialization or anything like that here.

sergeuz: Special floating point values are tested when serializing/deserializing a variant as CBOR, for example: https://github.com/particle-iot/device-os/blob/59325889b628f591756c40c9cc5fa33dc73c05a9/test/unit_tests/wiring/variant.cpp#L433-L441.

wiring/src/spark_wiring_ledger.cpp
sergeuz: @avtolstoy This can be a topic to discuss with a broader group of Device OS engineers.

avtolstoy: :+1: But just to say in advance this is the same pattern that we use in recently introduced APIs. GuoHui and I have gone through this during BLE introduction in detail and have come to the conclusion that function pointer with context argument + std::function wrapper most of the times is sufficient. One extra step is:
```c++
template<typename T>
void onBlah(void(T::*callback)(void arg1, void arg2), T* instance);
```

with no real dedicated implementation and just `std::bind` to produce `std::function` and delegate to its `onBlah` implementation.

avtolstoy (approved):
:+1:

sergeuz:
Alright, I'm going to merge this at last. Thanks for all the feedback, everybody!

PR #2718, opened 2023-12-13
Ensure consistent BLE scan behavior

XuGuohui:
### Problem

`BLE.scan()` and `BLE.scanWithFilter()` filter duplicate results on Gen3, but it's not true on Gen4.

### Solution

Make the BLE HAL report all of the scanned results and implement a filtering mechanism in wiring layer instead.

### Steps to Test
Build and  run the `tests/app/ble/scanner` app.

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/src/spark_wiring_ble.cpp
+++ b/wiring/src/spark_wiring_ble.cpp
@@ -2378,6 +2378,12 @@ private:
      */
     static void onScanResultCallback(const hal_ble_scan_result_evt_t* event, void* context) {
         BleScanDelegator* delegator = static_cast<BleScanDelegator*>(context);
+
+        if (!delegator->filter_.allowDuplicates() && delegator->isCachedDevice(event->peer_addr)) {
+            return;
+        }
+        delegator->cachedDevices_.append(event->peer_addr);
+
         BleScanResult result = {};
         result.address(event->peer_addr).rssi(event->rssi)
               .scanResponse(event->sr_data, event->sr_data_len)
@@ -2554,6 +2560,15 @@ private:
         return true;
     }
 
+    bool isCachedDevice(const BleAddress& address) const {
+        for (const auto& addr : cachedDevices_) {
+            if (address == addr) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     Vector<BleScanResult> resultsVector_;
     BleScanResult* resultsPtr_;
     size_t targetCount_;
@@ -2561,6 +2576,7 @@ private:
     std::function<void(const BleScanResult*)> scanResultCallback_;
     BleOnScanResultStdFunction scanResultCallbackRef_;
     BleScanFilter filter_;
+    Vector<BleAddress> cachedDevices_;
 };
 
 int BleLocalDevice::setScanTimeout(uint16_t timeout) const {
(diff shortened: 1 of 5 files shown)


wiring/src/spark_wiring_ble.cpp
```diff
@@ -2378,6 +2378,12 @@ class BleScanDelegator {
      */
     static void onScanResultCallback(const hal_ble_scan_result_evt_t* event, void* context) {
         BleScanDelegator* delegator = static_cast<BleScanDelegator*>(context);
+
+        if (!delegator->filter_.allowDuplicates() && delegator->isCachedDevice(event->peer_addr)) {
+            return;
+        }
+        delegator->cachedDevices_.append(event->peer_addr);
```
avtolstoy: This should not be populated in case of `allowDuplicates()` to avoid uncontrollable growth of the vector, as otherwise it will continue to fill-up with duplicates.

Same can potentially happen normally depending on scan duration, although highly unlikely, but should potentially be taken care of somehow as well.

avtolstoy (changes requested):

wiring/src/spark_wiring_ble.cpp
```diff
@@ -2378,6 +2378,14 @@ class BleScanDelegator {
      */
     static void onScanResultCallback(const hal_ble_scan_result_evt_t* event, void* context) {
         BleScanDelegator* delegator = static_cast<BleScanDelegator*>(context);
+
+        if (!delegator->filter_.allowDuplicates()) {
+            if (delegator->isCachedDevice(event->peer_addr)) {
+                return;
+            }
+            delegator->cachedDevices_.append(event->peer_addr);
```
avtolstoy, line 2386: @XuGuohui Do you think it would make sense to put a hard limit here on a maximum number of results?

XuGuohui: I think it wouldn't make sense, cos when the vector reaches the limit the device will report a large amount of duplicates, which more looks like a bug and might make the user panic. In whichever way we go we should document the limitation.

avtolstoy: I'm talking specifically about the case of non-duplicates. E.g. a moving device with a long scan duration and capturing a large number of devices around.

XuGuohui: What about introducing an option `BleScanFilter.cachedDeviceSize()` so that user can explicitly notice the limitation, otherwise, the cache can be uncontrollable to grow?

avtolstoy: Doubt anyone would use it :) Some relatively large default (e.g. 512-1024) is probably better, but honestly up to you to leave as-is potentially as well.

avtolstoy (approved):

wiring/src/spark_wiring_ble.cpp
XuGuohui: Cool. I'll leave it as-is, as you said it's highly unlikely happen and we can react on it when there is customers encounter a shortage of RAM.

XuGuohui:
Cc. @rickkas7 Introduced an option `allowDuplicates()` for `BleScanFilter`.

PR #2721, opened 2023-12-28
[msom] Add GNSS test FQC command

scott-brust:
### Problem

Msom hardware supports a built in GNSS engine along with the cellular modem. The mfg-cli needs to be updated to test this GNSS functionality as part of FQC testing at the factory. A new tinker:gnss-test command has been added to enable testing of both M404 (Quectel BG95-M5 modem/GNSS) and M524(Quectel EG91 modem/GNSS) hardware.

### Solution

Add support for the new `particlemfg tinker:gnss-test` command

### Steps to Test
Use `mfg-cli` to test [per instructions here](https://github.com/particle-iot/cli/pull/199)

### Example App

Use tinker

### References

See [CLI PR here](https://github.com/particle-iot/cli/pull/199)
See [slack here](https://s.slack.com/archives/C046BM6DBLL/p1703577091536279)

### DIFF
--- a/user/applications/tinker/src/burnin_test.cpp
+++ b/user/applications/tinker/src/burnin_test.cpp
@@ -557,7 +557,18 @@ static int callbackGPSGGA(int type, const char* buf, int len, bool* gnssLocked) 
     return 1;
 }
 
-bool BurninTest::testGnss() {
+static int callbackQGPS(int type, const char* buf, int len, bool* cmdSuccess) {
+    //Log.trace("%d : %s", strlen(buf), buf);
+
+    // If string is `OK` or `+CME ERROR: 504` (ie GNSS already started) then GPS engine is enabled
+    if (!strcmp(buf, "\r\n+CME ERROR: 504\r\n") || !strcmp(buf, "\r\nOK\r\n")) {
+        *cmdSuccess = true;
+    }
+    return 0;
+}
+
+
+bool BurninTest::initGnss() {
     // Turn on GNSS + Modem
     pinMode(GNSS_ANT_PWR, OUTPUT);
     digitalWrite(GNSS_ANT_PWR, HIGH);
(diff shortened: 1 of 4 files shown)


YutingYou (approved):
LGTM, also verified on M524(EG91-EX) and M404(BG95-M5)

```
$ npm run-script mfg-cli:start tinker:gnss-test start

> cli-monorepo@1.0.0 mfg-cli:start /Users/yuting/projects/js_prj/particle-cli
> ./packages/mfg-cli/bin/run "tinker:gnss-test" "start"

{
    "pass": true
}

$ ./packages/mfg-cli/bin/run tinker:gnss-test status
{
    "pass": true,
    "message": "Found 1 satellites"
}
```

user/applications/tinker/src/burnin_test.cpp
```diff
@@ -568,25 +579,40 @@ bool BurninTest::testGnss() {
 
     // Enable GNSS. It can take some time after the modem AT interface comes up for the GNSS engine to start
     const int RETRIES = 10;
-    int r = 0;
-    for (int i = 0; i < RETRIES && r != RESP_OK; i++) {
-        r = Cellular.command("AT+QGPS=1\r\n");
+    
+    bool success = false; 
+    for (int i = 0; i < RETRIES && !success; i++) {
```
eberseth, line 584: This loop is used elsewhere as well.  It would be handy to have a macro/template that wraps retry loops with delay.  Don't worry about it today.

user/applications/tinker/src/burnin_test.cpp
```diff
@@ -568,25 +579,40 @@ bool BurninTest::testGnss() {
...
-    for (int i = 0; i < RETRIES && r != RESP_OK; i++) {
-        r = Cellular.command("AT+QGPSCFG=\"priority\",0");
-        delay(1000);
+    hal_device_hw_info deviceInfo = {};
+    hal_get_device_hw_info(&deviceInfo, nullptr);
+    if (deviceInfo.ncp[0] == PLATFORM_NCP_QUECTEL_BG95_M5) {
+        int r = 0;
+        // Configure antenna for GNSS priority
+        for (int i = 0; i < RETRIES && r != RESP_OK; i++) {
+            r = Cellular.command("AT+QGPSCFG=\"priority\",0");
```
eberseth, line 600: The `AT+QGPSCFG="priority"` command should take a maximum of 300ms to execute.  I see you don't have a timeout on the command call.

user/applications/tinker/src/fqc_test.cpp
```diff
@@ -486,27 +495,191 @@ bool FqcTest::wifiScanNetworks(JSONValue req) {
...
+    int i = 1;
+    while (token) {
+        //Log.trace("%d %s", i, token);
+        token = strtok(NULL, delimiters);
+        i++;
+
+        switch (i) {
+            case 5: // TotalNumSat 
+            {   
+                int numberSattelites = atoi(token);
```
eberseth: Spelling

user/applications/tinker/src/fqc_test.cpp
```diff
@@ -486,27 +495,191 @@ bool FqcTest::wifiScanNetworks(JSONValue req) {
...
+    const char * delimiters = ", ";
+    char * token = strtok(gsvSentence, delimiters);
+    int i = 1;
+    while (token) {
+        //Log.trace("%d %s", i, token);
+        token = strtok(NULL, delimiters);
+        i++;
+
+        switch (i) {
+            case 5: // TotalNumSat
```
eberseth: There are two constellations reported in the GSV sentences.  This method may take the last set only, which is GLONASS.

eberseth (approved):

user/applications/tinker/src/burnin_test.cpp
scott-brust: Right, I just use the default 10 second timeout. I dont care about timing just that it succeeded.

user/applications/tinker/src/fqc_test.cpp
scott-brust: Yes, this just lazily finds the last GSV sentence satellite count. I didnt do any more parsing since its not clear what data (if any) we care about returning to the host. 

If we can see any number of satellites, then I consider gnss working and return success. 
If we want this test to do more, then we need to specify that and I can improve it.

eberseth: You may want to target the constellation that would be provided by the GNSS simulator during manufacturing

scott-brust: Eugene posted an example simulator that reported GPS, but we dont know for sure what the factory will end up using. This will work with any constellation. 

If we care about detecting/parsing each individual constellation I can do that, but we should make sure this will be used somewhere.

PR #2722, opened 2024-01-09
MSoM: enable session resumption

XuGuohui:
As the title describes.

### Problem

1. Session resumption is currentlly disabled on MSoM
2. When session resumption is enabled, it may use stale session info that is restored from backup flash, which results in long connection time.

### Solution

1. Use the session info in KM4 RAM if it is valid.
2. Otherwise, use the session info in backup flash if it is valid.
3. Invalidate the session info in backup flash once it is restored.
4. If there is no valid session info, force a new handshake.

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/backup_ram_hal.cpp
+++ b/hal/src/rtl872x/backup_ram_hal.cpp
@@ -29,6 +29,7 @@
 #include "timer_hal.h"
 #include "static_recursive_mutex.h"
 #include "scope_guard.h"
+#include "dtls_session_persist.h"
 
 // NOTE: we are using a dedicated flash page for this as timings for writing
 // into fs vs raw page are about 10x
@@ -39,6 +40,8 @@ extern uintptr_t platform_backup_ram_all_end;
 extern uintptr_t platform_backup_ram_persisted_flash_start;
 extern uintptr_t platform_backup_ram_persisted_flash_end;
 extern uintptr_t platform_backup_ram_persisted_flash_size;
+
+extern SessionPersistDataOpaque session;
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


hal/src/rtl872x/backup_ram_hal.cpp
```diff
@@ -116,14 +139,17 @@ int hal_backup_ram_sync(void* reserved) {
     BackupRamLock lk;
     if (memcmp((void*)&platform_backup_ram_all_start, backupRamShadow, sizeof(backupRamShadow))) {
         memcpy(backupRamShadow, (void*)&platform_backup_ram_all_start, sizeof(backupRamShadow));
-        if (SYSTEM_FLAG(restore_backup_ram) != 1) {
-            SYSTEM_FLAG(restore_backup_ram) = 1;
+        if (!(SYSTEM_FLAG(restore_backup_ram) & SYSTEM_FLAG_RESTOR_BACKUP_RAM_MASK)) {
+            SYSTEM_FLAG(restore_backup_ram) |= SYSTEM_FLAG_RESTOR_BACKUP_RAM_MASK;
             dct_write_app_data(&system_flags, DCT_SYSTEM_FLAGS_OFFSET, DCT_SYSTEM_FLAGS_SIZE);
```
scott-brust, line 144: We store `system_flags` to DCT in addition to keeping it in retained SRAM in case we lose power and want to restore some retained values from flash?

hal/src/rtl872x/backup_ram_hal.cpp
```diff
@@ -102,12 +105,32 @@ int hal_backup_ram_init(void) {
...
         // Woke up from deep sleep
         CHECK(hal_flash_read((uintptr_t)&platform_backup_ram_persisted_flash_start, (uint8_t*)&platform_backup_ram_all_start,
                 (size_t)&platform_backup_ram_persisted_flash_size));
+
+        if ((BOOT_Reason() & BIT_BOOT_DSLP_RESET_HAPPEN) && !(SYSTEM_FLAG(restore_backup_ram) & SYSTEM_FLAG_SESSION_DATA_STALE_MASK)) {
+            // The session info restored out of flash is valid
+            // Invalidate the session info in flash, cos it's probably stale from now on
+            auto s = new SessionPersistDataOpaque();
+            if (s) {
+                std::swap(*s, session);
```
scott-brust, line 122: `session` should be restored from the backup ram flash page above (line 114), correct? 
Why do we want to erase it from flash here? Wont we immediately re-write it when `hal_backup_ram_sync` runs again? 
Isnt the original issue that the values in `session` (which is in backup RAM) will almost always be out of sync with the data written to flash during `hal_backup_ram_sync()` due to how frequently it changes? 
It doesnt seem likely that we will successfully resume a session with the current approach?

scott-brust (commented):
This did work for me when entering/exiting `ULTRA_LOW_POWER` sleep on msom:
```
0000102293 [system] INFO: Cloud: connecting
0000102301 [system] INFO: Read Server Address = type:1,domain:$id.udp.particle.io
0000102307 [system] INFO: Loaded cloud server address and port from session data
0000102315 [system] TRACE: Address type: 1
0000102319 [system] TRACE: Cloud socket=0, family=2, type=2, protocol=17
0000102329 [system] INFO: Cloud socket=0, connecting to 18.214.12.187#5684
0000102336 [system] TRACE: 0 Bound cloud socket to lwip interface wl4
0000102344 [system] TRACE: Cloud socket=0, connected to 18.214.12.187#5684
0000102352 [system] TRACE: Updating cloud keepalive for AF_INET: 25000 -> 25000
0000102360 [system] TRACE: Applying new keepalive interval now
0000102366 [system] INFO: Cloud socket connected
0000102370 [system] INFO: Starting handshake: presense_announce=0
0000102376 [comm.protocol.handshake] INFO: Establish secure connection
0000102386 [comm.dtls] INFO: session has 0 uses
0000102392 [comm.dtls] INFO: (CMPL,RENEG,NO_SESS,ERR) restoreStatus=0
0000102396 [comm.dtls] INFO: out_ctr 0,1,0,0,0,0,0,16, next_coap_id=d
0000102402 [comm.dtls] INFO: restored session from persisted session data. next_msg_id=13
0000102412 [comm.dtls] INFO: session cmd (CLS,DIS,MOV,LOD,SAV): 2
```

Im not sure if this was by coincidence though that the session coap counters were still in sync? 
Is this session restore supposed to work in other use cases beside `STOP` and `ULTRA_LOW_POWER` modes? 
edit: It worked for me when calling `System.reset()`

hal/src/rtl872x/backup_ram_hal.cpp
XuGuohui: Yes.

hal/src/rtl872x/backup_ram_hal.cpp
XuGuohui: >Why do we want to erase it from flash here?

In case that the device is reset before periodic `hal_backup_ram_sync()` is called. A session can be successfully restored for all of the sleep modes, as long as the cloud connection is disconnected gracefully. Yes, the original issue is due to the stale session data in flash, that's why in the PR we mark the session data in flash as invalid once we update the session data in RAM and mark it as valid when the `hal_backup_ram_sync()` is called.

avtolstoy (approved):

PR #2734, opened 2024-01-30
Protected feature

avtolstoy:
TODO


bootloader/src/nRF52840/usbd_dfu_mal.h
```diff
@@ -24,19 +24,23 @@
 
 #define INTERNAL_FLASH_START_ADD        0x00000000
 #define INTERNAL_FLASH_END_ADDR         0x00100000
-#define INTERNAL_FLASH_IF_STRING        "@Internal Flash   /0x00000000/1*004Ka,47*004Kg,192*004Kg,4*004Kg,4*004Kg,8*004Ka";
+#define INTERNAL_FLASH_IF_STRING        "@Internal Flash   /0x00000000/1*004Ka,47*004Kg,192*004Kg,4*004Kg,4*004Kg,8*004Ka"
+#define INTERNAL_FLASH_IF_STRING_PROT   "@Internal Flash   /0x00000000/1*004Kb,47*004Kb,192*004Kb,4*004Kb,4*004Kb,8*004Kb"
```
monkbroc, line 28: A comment explaining the reason for the different letters would be helpful so that someone looking at this in the future doesn't need to reverse engineer what the letters mean and why it's important to change them. This comment can wait for the next PR.

services/inc/services_dynalib.h
```diff
@@ -96,6 +97,8 @@ DYNALIB_FN(48, services, devicetree_tree_lock, int(void*))
 DYNALIB_FN(49, services, devicetree_tree_get, int(void*, uint32_t, void*))
 DYNALIB_FN(50, services, devicetree_string_dictionary_lookup, const char*(uint32_t, void*))
 DYNALIB_FN(51, services, devicetree_hash_string, uint32_t(const char*, size_t))
+DYNALIB_FN(52, services, security_mode_set, int(module_info_security_mode, void*))
```
monkbroc: Why do we need security_mode_set in dynalib? Shouldn't the bootloader be the only caller of this method?

avtolstoy: Needed for testing, it also doesn't allow protected->non-protected, so should be fine, but I'll remove this.

monkbroc (approved):

scott-brust (approved):
See testing discussion [here](https://s.slack.com/archives/CS0T32ZKK/p1706639045664839)

PR #2736, opened 2024-02-09
[system] Fix bad int to bool conversion for clearCredentials

eberseth:
### Problem

It was reported that `WiFi.clearCredentials()` was returning a bad value for success in this Community post [WiFi.clearCredentials() fails? on p2 @ 5.4.1](https://community.particle.io/t/wifi-clearcredentials-fails-on-p2-5-4-1).

Rick dug into the issue and saw that there was an integer to bool conversion gone bad.

### Solution

Added a comparison to zero with the `int clearConfiguration()` call so that `int() == 0` results in a boolean `true` in `network_clear_credentials()`.

Added comments for `network_clear_credentials()` usage.

### Steps to Test

Run this while credentials are set on the device.

```cpp
if(!WiFi.clearCredentials()) {
    Log.warn("clear credentials failed!");
}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/system/inc/system_network.h
+++ b/system/inc/system_network.h
@@ -102,6 +102,15 @@ typedef WLanCredentials NetworkCredentials;
  * @return 0 on success.
  */
 int network_set_credentials(network_handle_t network, uint32_t flags, NetworkCredentials* creds, void* reserved);
+/**
+ *
+ * @param network   The network to configure the credentials.
+ * @param flags     Flags. set to 0.
+ * @param creds     The credentials to set. Set to NULL.
+ * @param reserved  For future expansion. Set to NULL.
+ * @return true     Successfully cleared the given network credentials
+ * @return false    Failed to clear the given network credentials
+ */
 bool network_clear_credentials(network_handle_t network, uint32_t flags, NetworkCredentials* creds, void* reserved);
 
 void network_setup(network_handle_t network, uint32_t flags, void* reserved);
--- a/system/src/system_network_manager_api.cpp
+++ b/system/src/system_network_manager_api.cpp
@@ -438,7 +438,7 @@ bool network_clear_credentials(network_handle_t network, uint32_t, NetworkCreden
     } else {
         if_t iface;
         if (!if_get_by_index(network, &iface)) {
-            return NetworkManager::instance()->clearConfiguration(iface);
+            return NetworkManager::instance()->clearConfiguration(iface) == 0;
         }
     }
     return false;


system/inc/system_network.h
```diff
@@ -102,6 +102,15 @@ typedef WLanCredentials NetworkCredentials;
  * @return 0 on success.
  */
 int network_set_credentials(network_handle_t network, uint32_t flags, NetworkCredentials* creds, void* reserved);
+/**
+ *
+ * @param network   The network to configure the credentials.
+ * @param flags     Flags. set to 0.
+ * @param creds     The credentials to set. Should not be NULL.
```
technobly: `creds` is not used in the function, and is normally set to NULL.  When clearing creds, this argument is N/A and should be set to NULL.

system/src/system_network_manager_api.cpp
```diff
@@ -438,7 +438,7 @@ bool network_clear_credentials(network_handle_t network, uint32_t, NetworkCreden
     } else {
         if_t iface;
         if (!if_get_by_index(network, &iface)) {
-            return NetworkManager::instance()->clearConfiguration(iface);
+            return NetworkManager::instance()->clearConfiguration(iface) == 0;
```
technobly, line 441: Looks good, similar to line 437.  Curious what the error was from, possibly SYSTEM_ERROR_INVALID_ARGUMENT due to the WiFi interface not being turned on before calling WiFi.clearCredentials() ?  Docs don't say it's required, but likely is.

technobly (changes requested):

system/inc/system_network.h
eberseth: ok, will change description

system/src/system_network_manager_api.cpp
eberseth: With a return type of bool there is not much other than success or fail

technobly: I agree, I'm just wondering why this failed to return true in the first place.  Normally `clearConfiguration()` would return SYSTEM_ERROR_NONE (0), but the other possibility is SYSTEM_ERROR_INVALID_ARGUMENT which is a negative error code that would still cause this to return `false` with the new `== 0` if that is the case.  So what is causing that... possibly WiFi radio not powered on?

eberseth:
@technobly yea/nay?

technobly (approved):

PR #2737, opened 2024-02-13
Enable Ledger and filesystem APIs on GCC platform

sergeuz:
This PR adds minimal implementations of `exflash_hal` and `concurrent_hal` for the GCC platform, sufficient to support the LittleFS and Ledger APIs on that platform.

By default, the contents of the external flash is stored in RAM and is not persistent, meaning that the virtual device will always start with an empty filesystem. The firmware executable can be provided with a `--flash_file` option to enable persistent storage:

```
./my_firmware --flash_file=flash.bin
```

In RAM and on disk, the contents of the external flash is stored in a sparse array to ensure that RAM and disk space are only allocated for the data actually in use by the firmware and we don't potentially spike the IO and RAM usage in load tests.


system/src/ledger/ledger_manager.cpp
```diff
@@ -259,9 +260,12 @@ int LedgerManager::init() {
     if (!buf) {
         return SYSTEM_ERROR_NO_MEMORY;
     }
+    auto fsInstance = filesystem_get_instance(FILESYSTEM_INSTANCE_DEFAULT, nullptr);
+    assert(fsInstance);
+    FsLock fs(fsInstance);
```
monkbroc, line 265: Why does this one need to change but not the one in `LedgerManager::getLedgerNames`?

monkbroc (approved):

system/src/ledger/ledger_manager.cpp
sergeuz: The manager is expected to be initialized in `getLedgerNames()` and `instance()` will take care of that.

sergeuz: Note that the main change in `init()` is that it now calls `filesystem_mount()` and the latter expects a `filesystem_t` instance, not `lfs_t`, hence the difference.

hal/src/gcc/exflash_hal.cpp
```diff
@@ -0,0 +1,210 @@
...
+int hal_exflash_init(void) {
+    return 0;
+}
+
+int hal_exflash_uninit(void) {
+    return 0;
+}
+
+int hal_exflash_read(uintptr_t addr, uint8_t* data_buf, size_t data_size) {
+    try {
```
avtolstoy, line 173: While this pattern is not used in normal Device OS sources for real devices, would be nice to introduce some CHECK-style macro or something along those lines to forward `SYSTEM_ERROR_XXX` error codes on exceptions to improve readability.

system/src/ledger/ledger_manager.cpp
```diff
@@ -1586,14 +1590,13 @@ LedgerManager* LedgerManager::instance() {
     // XXX: Lazy initialization is used so that ledger instances can be requested in the global
     // scope by the application. It's dangerous because the application's global constructors are
     // called before the system is fully initialized but seems to work in this case
-    static volatile bool initCalled = false;
-    if (!initCalled) {
+    static std::once_flag onceFlag;
+    std::call_once(onceFlag, []() {
```
avtolstoy, line 1594: :+1:

avtolstoy (approved):

PR #2741, opened 2024-02-19
Prevent OTA progress events from blocking system thread

sergeuz:
### Problem

Emitting a firmware update progress event blocks the system thread if the event queue of the application thread is full.

### Solution

Introduce a flag for `system_notify_event()` that allows the system to drop the event if the event queue is full.

This PR also removes the unused event queue implementation based on the `cpp::channel` library. Interestingly, that seems to have caused a problem when linking the wiring library with a user app that uses `Print::print(double)` in my local environment even though it's not reproducible on the CI. As a workaround, I inlined methods that get conditionally compiled based on the `PARTICLE_WIRING_PRINT_NO_FLOAT` macro since we only build the wiring library once and can't have different compile time options for it depending on where it's used.

### Steps to Test

1. Flash the test application to a device OTA.
2. When the device restarts, flash some large enough firmware to the device OTA.
3. Validate that flashing finishes in a reasonable time.

### Test Application

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC)
SYSTEM_THREAD(ENABLED)

namespace {

void onFirmwareUpdateEvent(system_event_t event, int param, void* data) {
    if (param == firmware_update_progress) {
        HAL_Delay_Milliseconds(3000);
    }
}

} // namespace

void setup() {
    System.on(firmware_update, onFirmwareUpdateEvent);
    Particle.connect();
}

void loop() {
}
```

### References

- [sc-125421]


mrlambchop:
Folks - whilst I appreciate the work gone into fix this OTA issue, I recognize this an old troubled friend in the world of event queues and message passing and would like to discuss this at the system design level before it lands. Will pop into a standup to discuss!

avtolstoy (approved):

PR #2742, opened 2024-02-20
[rtl872x] ota: fix bounds check, add ota/min_max_app_size test

avtolstoy:
### Problem

See [SC125387].

The gist of the underlying problem is that due to how user module bounds are handled and because user modules "grow" left (i.e. end address is static/fixed) when receiving a compressed application the end address does not match the intended static/fixed end address. E.g. running a small app:

```
      moduleStartAddy: '85cc000',
      moduleEndAddy: '85ffffc',
```
user module bounds are:
```
maximum_size = 0x180000, start_address = 0x85cc000, end_address = 0x8600000
```

With compression some larger application becomes:
```
module_start_address = 0x85cc000, module_end_address = 0x85ce213
```
End address won't pass validation in `fetch_module`

### Solution

Fix bounds check in `fetch_module`. Add `ota/min_max_app_size` test suite to validate that:
1. OTA min -> max update works (Default cloud-side compression settings)
2. OTA max -> min update works (Default cloud-side compression settings)
3. USB min -> max update works
4. USB max -> min update works
5. USB min -> max with compression update works
6. USB max -> min with compression update works

where 'min' is the test application and 'max' is the test application binary padded to maximum supported application binary size.

### Alternative solution

We could have fixed the compression method for RTL platforms where apps grow left (static/fixed end address), but such a change would require a significant diff to `flash_mal` and potentially other places. I simply don't want to introduce any more complexity to the fix.

### Steps to Test

Run `ota/min_max_app_size`.

### Example App

N/A

### References

- [SC125387]

---

### Completeness

- [ ] Added documentation


XuGuohui:
I don't get how `module_end_address` in module info can be less than the end address of the user bound. The `module_end_address` is assigned with `link_module_end`, which is always `0x85ffffc`?

*discussed in slack*

XuGuohui (approved):

PR #2743, opened 2024-02-21
Fixes std::call_once reentrancy problem

avtolstoy:
### Problem

`std::call_once` implementation in libstdc++ is not re-entrant and uses a common shared mutex:

```c++
// XXX: This is not defined in our toolchain, no thread local storage (TLS)
#undef _GLIBCXX_HAVE_TLS
// XXX

  template<typename _Callable, typename... _Args>
    void
    call_once(once_flag& __once, _Callable&& __f, _Args&&... __args)
    {
      // _GLIBCXX_RESOLVE_LIB_DEFECTS
      // 2442. call_once() shouldn't DECAY_COPY()
      auto __callable = [&] {
	  std::__invoke(std::forward<_Callable>(__f),
			std::forward<_Args>(__args)...);
      };
#ifdef _GLIBCXX_HAVE_TLS
      __once_callable = std::__addressof(__callable);
      __once_call = []{ (*(decltype(__callable)*)__once_callable)(); };
#else
      unique_lock<mutex> __functor_lock(__get_once_mutex());
      __once_functor = __callable;
      __set_once_functor_lock_ptr(&__functor_lock);
#endif

      int __e = __gthread_once(&__once._M_once, &__once_proxy);

#ifndef _GLIBCXX_HAVE_TLS
      if (__functor_lock)
        __set_once_functor_lock_ptr(0);
#endif

#ifdef __clang_analyzer__
      // PR libstdc++/82481
      __once_callable = nullptr;
      __once_call = nullptr;
#endif

      if (__e)
	__throw_system_error(__e);
    }
```

### Solution

A workaround for that. Going forward we should ideally get rid of `std::call_once` usage altogether and replace with our own implementation. Supporting `__gthread_xxx` implementations for `std` internals which constantly change between GCC releases is not trivial.

### Steps to Test

Not adding tests as `std::call_once` platform-specific implementations are not exposed to the application anyway.

M SoM build out of this branch should correctly boot in to the application.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/portable/FreeRTOS/bits/gthr-default.h
```diff
@@ -14,8 +14,14 @@
...
 typedef int __gthread_key_t;
-typedef std::atomic_flag __gthread_once_t;
 
+typedef enum __gthread_once_state_t {
+    GTHREAD_ONCE_STATE_NOT_INITIALIZED = 0,
+    GTHREAD_ONCE_STATE_RUNNING = 1,
+    GTHREAD_ONCE_STATE_INITIALIZED = 2
+} __gthread_once_state_t;
+
+typedef std::atomic<__gthread_once_state_t> __gthread_once_t;
```
scott-brust, line 24: The `__gthread_once_t` type is an atomic type with the above enum. 
We use the standard atomic operations to load / compare the once state value and use that to decide if we need to call the once function or not. 

Is this doing the same [thing](https://en.cppreference.com/w/cpp/thread/once_flag) as `once_flag`?
Is that correct?

hal/src/portable/FreeRTOS/bits/gthr.cpp
```diff
@@ -0,0 +1,68 @@
...
+extern std::unique_lock<std::mutex>*& __get_once_functor_lock_ptr();
+
+} // std
+
+// XXX/FIXME: std::call_once platform-specific implementations are not exported to the application
+// so we should probably just get rid of std::call_once altogether and replace with a separate implememntation.
+extern "C" int __gthread_once (__gthread_once_t* once, void (*func) (void))
+{
+    if (once->load(std::memory_order_acquire) == GTHREAD_ONCE_STATE_INITIALIZED) {
+        // Short path, already initialized
```
scott-brust, line 39: This is the case where `func` has already been called once and nothing needs to be done?

hal/src/portable/FreeRTOS/bits/gthr.cpp
```diff
@@ -0,0 +1,68 @@
...
+        return 0;
+    }
+
+    // Make a local copy, this is a workaround for a non-entrant implementation in libstdc++
+    auto callable = std::__once_functor;
+    auto lock = std::__get_once_functor_lock_ptr();
+    if (lock) {
```
scott-brust, line 46: What case would `__get_once_functor_lock_ptr` return a non zero value? 
I see below we set it back to null with `__set_once_functor_lock_ptr(0)`?

hal/src/portable/FreeRTOS/bits/gthr.cpp
```diff
@@ -0,0 +1,68 @@
...
+        // This won't be relocked in <mutex> std::call_once implementation
+        lock->unlock();
+    }
+    // std::__once_functor() is not usable from this point on, local callable should be used
+
+    auto state = GTHREAD_ONCE_STATE_NOT_INITIALIZED;
+    if (once->compare_exchange_strong(state, GTHREAD_ONCE_STATE_RUNNING,
+            std::memory_order_acq_rel)) {
+        // Switched to initializing/running state
+        callable();
```
scott-brust, line 58: This is where the actual one time function is called right?

hal/src/portable/FreeRTOS/bits/gthr.cpp
```diff
@@ -0,0 +1,68 @@
...
+        once->store(GTHREAD_ONCE_STATE_INITIALIZED, std::memory_order_release);
+    } else {
+        // Wait for some other thread to complete initialization
+        while (once->load(std::memory_order_acquire) == GTHREAD_ONCE_STATE_RUNNING) {
```
scott-brust, line 62: The `else` condition here is when the `once` state is something other than `GTHREAD_ONCE_STATE_NOT_INITIALIZED` ?
This can only happen if another thread is calling `__gthread_once` with the same `once` state object?
In this scenario we just yield until the other thread completes and the state variable is equal, then return?
We dont need to do any kind of atomic compare since we are polling it?

scott-brust (approved):
Asked some questions to make sure I understand the behavior

hal/src/portable/FreeRTOS/bits/gthr-default.h
avtolstoy: This is an actual underlying implementation of `std::once_flag`. It's platform-specific, so we have to provide it. This is not new one, I just moved some things around. See deleted `gtrh-default.h`

hal/src/portable/FreeRTOS/bits/gthr.cpp
avtolstoy: Correct.

hal/src/portable/FreeRTOS/bits/gthr.cpp
avtolstoy: Just a precaution, but this should always be not null. This is a pointer to ` unique_lock<mutex> __functor_lock(__get_once_mutex())` created right before call into this function.

hal/src/portable/FreeRTOS/bits/gthr.cpp
avtolstoy: Yes

hal/src/portable/FreeRTOS/bits/gthr.cpp
avtolstoy: This case is for `RUNNING` state: both threads fell through the first `UNINITIALIZED` check, the other thread managed to 'acquire' the atomic and set it to `RUNNING` + run `callable()`, our thread failed to acquire the atomic at that time and just waits for the second thread to complete running `callable()` and set state to INITIALIZED

XuGuohui (approved):

PR #2745, opened 2024-02-27
Process system tasks while cloud handshake is in progress

sergeuz:
### Problem

System tasks in `SystemISRTaskQueue` are not processed while the DTLS handshake or Hello request is in progress, which, among other things, prevents control requests from working. Story details: [sc-125370](https://app.shortcut.com/particle/story/125370/usb-control-requests-stuck-while-connecting-to-the-cloud).

### Solution

Process the task queue while the handshake with the Cloud is in progress.


avtolstoy (approved):
Would be nice if we could process system event queue as well i.e. make `spark_protocol_handshake()` non-blocking, but for now this works as a workaround for USB requests and other things.

monkbroc (approved):

PR #2747, opened 2024-02-29
[Gen4] fixes BLE scanning may break.

XuGuohui:
### Problem
BLE scans for BLE advertisements will quickly hit a failure and become unable to scan. The BLE.scan API will start to return immediately without performing the scan.

### Solution
Reset the BT stack when it failed to start scanning.

### Steps to Test
Run the following test app.

### Example App

```cpp
#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);

int advert_count;
void scan_result_cb(const BleScanResult *result, void *context) {
    advert_count++;
    // Log.info("\tgot advert");
}

void scan_thread_f() {
    unsigned int short_scan_count = 0;
    while (true) {
        unsigned int t0, t1;
        advert_count = 0;
        t0 = millis();
        BLE.scan(scan_result_cb, nullptr);
        t1 = millis();
        if (t1 - t0 < 50) {
            short_scan_count++;
        } else {
            short_scan_count = 0;
            if (t1 - t0 > 500) {
                Log.info("\tscan too long: %dms", t1 - t0);
            }
        }
        if (advert_count) {
            Log.info("BLE scan got %d adverts (duration %ums)", advert_count, t1 - t0);
        }
        if (short_scan_count && !(short_scan_count % 1000)) {
            Log.info("BLE scan too short last %d scans (ex. %dms)", short_scan_count, t1 - t0);
        }
        delay(1);
    }
}

void setup() {
    while (!Serial.isConnected()) {
        delay(100);
    }

    BLE.on();
    // quick scans should see the issue more quickly
    BLE.setScanTimeout(10);

    static Thread *thread = new Thread("sensor_ble_scan", scan_thread_f, OS_THREAD_PRIORITY_DEFAULT);
}
```

### References
SC125720

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -1629,8 +1629,10 @@ int BleGap::getScanParams(hal_ble_scan_params_t* params) const {
 int BleGap::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
     CHECK(start());
     CHECK_FALSE(isScanning_, SYSTEM_ERROR_INVALID_STATE);
+    bool resetStack = false;
 
     SCOPE_GUARD ({
+        clearPendingResult();
         if (isScanning_) {
             const int LE_SCAN_STOP_RETRIES = 10;
             for (int i = 0; i < LE_SCAN_STOP_RETRIES; i++) {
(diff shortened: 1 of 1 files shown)


avtolstoy (approved):
@XuGuohui This works as a workaround. Do we have any understanding on why this happens especially if we are not doing any WiFi comms?

XuGuohui:
I think it's that the BT stack state machine has race condition, causing that certain even is not progagated to high layer application.

eberseth:
As Andrey mentioned this is a workaround.  We'll need to continue looking for a root cause after this.

jlovinger-particle:
In testing the fix, it does appear to resolve the immediate issue of scans getting into a persistent bad state. Hopefully we can get to a better root cause and fix. One item I did notice, however, is the logging might point somewhat towards the possible issue.

The following snippet shows one successful scan (2 adverts on first line) then the bad sequence on next scan. The logging from Device-OS suggests the scan failed to start `failed to start scanning`. But when the BLE.scan() finally returned (total duration 2342ms) it actually reported a a larger number of adverts by fair margin than the normal 100ms scans. So seems the scan _did_ start but we failed to catch that for whatever reason.

```
0001239357 [app] INFO: BLE scan got 2 adverts (duration 113ms) (5675/5743)
0001240452 [hal.ble] TRACE: failed to start scanning.
0001240507 [hal.ble] TRACE: Going to stop the stack...
0001240893 [net.lwip_rltk] INFO: promisc_deinit TODO
0001240951 [hal] INFO: WiFi off
0001241000 [hal] INFO: rltk_wlan_set_netif_info: 0, 94:94:4a:04:53:d8
0001241411 [hal] INFO: WiFi on
0001241781 [app] INFO: BLE scan got 11 adverts (duration 2342ms) (5676/5744)
0001241863 [app] INFO: 	scan too long: 2342ms
```

PR #2753, opened 2024-03-11
Fix msom usb + ble setup names

mrlambchop:
### Problem

Msom setup shows the wrong name: Mseries. It should be M-SoM 
Same for P2 BLE setup and B-SoM.

Original by ScottBrust

### Solution

Fix the hardcoded USB + BLE setup strings

### Steps to Test

Build + flash branch. Setup msom on setup.particle.io, name should show as M SoM

### Example App

tinker

### References

---

### Completeness

- [ ] Added documentation


eberseth (approved):

avtolstoy:
There is a few other places where this needs to be changed e.g. https://github.com/particle-iot/device-os/blob/develop/hal/src/b5som/app_usbd_string_config.h#L127. I'll take care of it before merging.

PR #2755, opened 2024-03-12
[rtl872x] some more fixes

avtolstoy:
1. Fixes connectivity tester crash
2. `WRAP_BLE_EVT_LOCK` missing in a couple of places
3. Moved coex deinitialization into `rtwCoexStop` under a lock
4. Removed stop scan retries for now, looks like this is no longer necessary
5. Increase coex thread stack size, occasionally seen stack overflow assertion

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -72,7 +72,6 @@ extern "C" {
 #include "scope_guard.h"
 #include "rtl_system_error.h"
 #include "radio_common.h"
-#include "freertos/wrapper.h"
 
 #include "timer_hal.h"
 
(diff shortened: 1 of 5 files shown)


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1233,19 +1221,7 @@ int BleGap::stop(bool restore) {
...
-            // will call into rtw_coex_run_enable(123, false) which will finally cleanup the mutex
-            rtwCoexRunDisable(0);
-            rtw_coex_bt_enable(*(void**)rltk_wlan_info[0].dev->priv, 0);
-            HAL_Delay_Milliseconds(100);
-            rtwCoexRunEnable(0);
-            rtw_coex_wifi_enable(*(void**)rltk_wlan_info[0].dev->priv, 0);
-            rtwCoexRunDisable(0);
-            rtw_coex_wifi_enable(*(void**)rltk_wlan_info[0].dev->priv, 1);
-            rtwCoexCleanupMutex(0);
+            rtwCoexStop();
```
eberseth, line 1224: Wrapped lines above into `rtxCoexStop()` and apply a mutex within.

eberseth (approved):

PR #2758, opened 2024-03-15
[system] network: force IFF_UP if requested and no credentials exist

avtolstoy:
### Description

WiFi should be connected after `joinNewNetwork` when there were no wifi credentials on the device at all as well as there were some present before.


eberseth (approved):

Kategrode:
https://app.shortcut.com/particle/story/125990/wifi-network-does-not-remain-connected-after-calling-join-known-network

PR #2760, opened 2024-04-04
[rtl872x] wifi: reset the stack when there are no scan results and scan is requested to connect to AP

avtolstoy:
### Description

This was reverted in 5.8.0, still seeing this happen in some cases so bringing it back in with some minor changes (`forConnect`)


scott-brust (approved):

Kategrode:
https://app.shortcut.com/particle/story/125908/wi-fi-join-persistent-join-failures-for-jacuzzi

PR #2761, opened 2024-04-05
[msom] M404/BG95M5 disable 2g, only scan for LTEM

scott-brust:
### Problem

Seeing unexplained connectivity / DNS errors on 2g connections sometimes with MSOM

### Solution

Disable 2g networks until we can investigate further

### Steps to Test
Run any app

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -1439,8 +1439,12 @@ int QuectelNcpClient::registerNet() {
             r = CHECK_PARSER(respNwMode.readResult());
             CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);
         #if PLATFORM_ID == PLATFORM_MSOM
-            if (nwScanMode != 0) {
-                CHECK_PARSER(parser_.execCommand("AT+QCFG=\"nwscanmode\",0,1")); // AUTO
+            // M404/BG95M5 should be LTEM only, ie scan mode 3
+            // M524/EG91EX should be AUTO (both LTEM and 2G), ie scan mode 0
+            int desiredNwScanMode = (ncpId() == PLATFORM_NCP_QUECTEL_BG95_M5) ? 3 : 0;
+
+            if (nwScanMode != desiredNwScanMode) {
+                CHECK_PARSER(parser_.execCommand("AT+QCFG=\"nwscanmode\",%d,1", desiredNwScanMode));
             }
 
             if (ncpId() == PLATFORM_NCP_QUECTEL_BG95_M5) {


avtolstoy (approved):

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1439,8 +1439,12 @@ int QuectelNcpClient::registerNet() {
             r = CHECK_PARSER(respNwMode.readResult());
             CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);
         #if PLATFORM_ID == PLATFORM_MSOM
-            if (nwScanMode != 0) {
-                CHECK_PARSER(parser_.execCommand("AT+QCFG=\"nwscanmode\",0,1")); // AUTO
+            // M404/BG95M5 should be LTEM only, ie scan mode 3
+            // M524/EG91EX should be AUTO (both LTEM and 2G), ie scan mode 0
+            int desiredNwScanMode = (ncpId() == PLATFORM_NCP_QUECTEL_BG95_M5) ? 3 : 0;
```
technobly, line 1444: This check will never test for M524 because it is not PLATFORM_NCP_QUECTEL_BG96 || PLATFORM_NCP_QUECTEL_BG95_M5, but rather PLATFORM_NCP_QUECTEL_EG91_EX.  I think we might need to add a  `if (isQuecCat1Device()) { }` block if we want to ensure `nwscanmode` is on the default of 0 for those devices.  For this block you can simple check `if (nwScanMode != 3) {` and set to 3 if not.

technobly (commented):
I think we also want to remove checks for CREG? for M-SoM if we are disabling 2G.
https://github.com/particle-iot/device-os/pull/2761/files#diff-186ce5679947e07e5ce6c3d4fce9cdce5ebd61e4c315bcf0658db6106ea54299R1487

And potentially make sure the +CREG URCs are not causing any issues.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
scott-brust: Good point, I didnt realize this wasnt checking for EG91. So M524 NWSCAN is not being set, presumably being left as default (ie enable 2g)?

This does disable 2g for specifically M404/BG95M5, and leaves it enabled for BG95M6 devices (which I dont think there is any aside from a few samples?)

scott-brust:
> I think we also want to remove checks for CREG? for M-SoM if we are disabling 2G. https://github.com/particle-iot/device-os/pull/2761/files#diff-186ce5679947e07e5ce6c3d4fce9cdce5ebd61e4c315bcf0658db6106ea54299R1487
> 
> And potentially make sure the +CREG URCs are not causing any issues.

I think there is a larger refactor around CREG if/when we figure out 2g stuff. For now I dont think it hurts anything

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: Yeah I think EG91_EX will be left at defaults, 0.

technobly:
If the intention is to disable 2G connections for the time being for M-SoM, I think you'll want to disable AT+CREG=2 and AT+CREG? commands so that the system doesn't take action on any registration/de-registration events due to those +CREG: responses.

Kategrode:
https://app.shortcut.com/particle/story/127103/ensure-correct-modem-configuration-when-enabling-2g-support

PR #2763, opened 2024-04-10
Add set credentials usb request

scott-brust:
### Problem

The existing `JoinNewNetworkRequest` protobuf message has some behavior that might be undesireable in some circumstances.
- The wifi network specified in the request must be present for the credentials to be stored
- The device will power on wifi and attempt to connect to wifi (not necessarily using the new credentials if there are existing credentials/networks known)

### Solution

For use cases like a factory environment where the end wifi network is not present, it is desired to set credentials without attempting to immediately power on wifi and connect. The `SetNetworkCredentialsRequest` will simply store the credentials to the devices internal list of wifi credentials.

Some items to note:
- Currently, devices in the 5x line of Device OS can store 10 wifi credentials.
- ⚠️  If a device has 10 credentials and is asked to store more, it will discard the last credential in the list and replace it with the specified one. Not sure if this is the clearest behavior

### Steps to Test

I used `particle-usb` with the protobuf module updated to the branch [here](https://github.com/particle-iot/device-os-protobuf/tree/sc-126826/set-credentials-usb-request)
Install / build the definitions as in the example.

I added the following snippet to `particle-usb.js` to test the device OS parsing/handling of the new request
```
async function testDevices(){
    const devices = await getDevices();

    for (let device of devices) {
    console.log("Device Type: ", device.type);
    }

	let device = devices[0];
	await device.open();

	let message = {
		type_id : 507, // CTRL_REQUEST_WIFI_SET_NETWORK_CREDENTIALS
		ssid: "TP-Link_F6E0",
		bssid: null,
		security: 4, //   WPA_WPA2_PSK = 4
		hidden: false,
		credentials: {
			type: 1, // CredentialsType.PASSWORD
			password: "78019348"
		}
	}

	for (let iter = 0; iter < 20; iter++) {
		message.ssid = "network".concat(iter);
		let rep = await device._sendAndHandleProtobufRequest('wifi.SetNetworkCredentialsRequest', message, { timeout: 30000 });
		console.log( rep);
	}
}

testDevices();
```

Test with `node particle-usb.js`

### Example App
The `connection-manager` app has some helpful commands to list/clear wifi APs to validate testing. See `a` and `l` commands. 

### Relevant PRs
https://github.com/particle-iot/device-os-protobuf/pull/31


keeramis (approved):
I have tested this command via the Particle USB and I am not seeing any issues with the new control request as of now. It simply puts the new SSID to the config file in the filesystem and does not try to do a rejoin. I believe the workflow is as expected with the existing control requests including the new one.

avtolstoy (approved):

PR #2764, opened 2024-04-22
[Gen 4] change coexistence behavior in wifi connecting or wifi connected states with background BLE low prio activity

avtolstoy:
TODO


hal/src/rtl872x/rtl_sdk_support.cpp
```diff
@@ -65,6 +69,13 @@ namespace {
 #if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
 uint8_t radioStatus = RTW_RADIO_NONE;
 RecursiveMutex radioMutex;
+
+volatile bool s_wifiConnectionState = false;
+volatile bool s_tdmaSkip = false;
+
+uint32_t s_coexTable[3] = {0x55555555, 0xaaaa5555, 0xf0ffffff};
```
avtolstoy: 0x55555555 == BT > WLAN, 0xaaaa5555 == WLAN high-Priority > BT > WLAN low-Priority

bootloader/prebootloader/src/rtl872x/part1/include.mk
```diff
@@ -1,8 +1,20 @@
 MAIN_STACK_SIZE = 4096
 
+# # FIXME
```
scott-brust, line 3: Are these fixed in the prebootloader-part1 makefile below?

bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
```diff
@@ -324,10 +326,15 @@ void sleepProcess(void) {
              */
             auto config = sleepConfigShadow.config();
             if (config->mode == HAL_SLEEP_MODE_HIBERNATE) {
+                SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
+                taskENTER_CRITICAL();
                 configureDeepSleepWakeupSource(config);
                 enterDeepSleep();
                 // It should not reach here
             } else {
+                SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
```
scott-brust, line 335: This disables systick, which should stop the freertos scheduler from swapping in other tasks? So effectively we stop FreeRTOS at this point? Hibernate will restart, so we dont care about starting the OS again?

hal/inc/ble_hal.h
```diff
@@ -1153,6 +1153,15 @@ ssize_t hal_ble_gatt_client_write_without_response(hal_ble_conn_handle_t conn_ha
...
 
+/**
+ * Internal use.
+ * 
+ * @param data 
+ * @param size 
+ * @param reserved 
+ * @return int 
+ */
+int hal_ble_internal(int type, void* data, size_t size, void* reserved);
```
scott-brust, line 1164: Just double checking we still want to include this hook. I assume its still useful for testing stuff?

hal/src/rtl872x/concurrent_hal_impl.h
```diff
@@ -35,8 +35,8 @@ typedef uint8_t os_thread_prio_t;
 /* Default priority is the same as the application thread */
 #define OS_THREAD_PRIORITY_DEFAULT       (2)
 #define OS_THREAD_PRIORITY_CRITICAL      (9)
-#define OS_THREAD_PRIORITY_NETWORK       (7)
-#define OS_THREAD_PRIORITY_NETWORK_HIGH  (8)
+#define OS_THREAD_PRIORITY_NETWORK       (6)
+#define OS_THREAD_PRIORITY_NETWORK_HIGH  (7)
```
scott-brust, line 39: Do these changes have any effect? Do we know what priority the SDK threads run at? Do we need to take into account their priorities perhaps?

hal/src/rtl872x/rtl_sdk_support.cpp
```diff
@@ -190,10 +364,30 @@ void rtwRadioAcquire(RtwRadio r) {
...
     if (radioStatus != RTW_RADIO_NONE) {
+        static std::once_flag once;
+        std::call_once(once, [](){
+            static rtl_wifi_fw_ram_alloc __attribute__((aligned(32))) info;
+            info.size = sizeof(info);
+            info.start = (uint32_t)&link_prebootloader_wifi_fw_ram_start;
+            info.end = (uint32_t)&link_prebootloader_wifi_fw_ram_end;
+            DCache_CleanInvalidate((uint32_t)&info, sizeof(info));
+            rtl_wifi_fw_resp resp = {};
+            int r = km0_km4_ipc_send_request(KM0_KM4_IPC_CHANNEL_GENERIC, KM0_KM4_IPC_MSG_WIFI_FW_INIT, &info, sizeof(info), [](km0_km4_ipc_msg_t* msg, void* context) -> void {
```
scott-brust, line 375: What does the "wifi firmware" actually do? Why does it need to run on KM0?

scott-brust (commented):
Some general questions. Still need to try to understand the actual coexistence changes and their implications

bootloader/prebootloader/src/rtl872x/part1/include.mk
avtolstoy: Yes, this should probably be removed.

bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
avtolstoy: Yup. Exactly right.

hal/inc/ble_hal.h
avtolstoy: Yes, let's leave it in, might want to tune things up in runtime.

hal/src/rtl872x/concurrent_hal_impl.h
avtolstoy: This is just fixing the general problem: MAC network threads should be below LwIP TCP/IP thread.

hal/src/rtl872x/rtl_sdk_support.cpp
avtolstoy: It manages low power wifi modes and mainly for our use-case prioritization between wifi and bt slots. Pretty much the main thing it does in simple terms is set up an interrupt at TBTT interval (100ms generally, depends on AP settings) and depending on the coexistence configuration may switch the RF frontend between WiFi and BLE peripheral. Of course it's a bit more complex than simply that.

scott-brust: ok thanks for the general overview. I'm guessing wifi fw is running on KM0 for performance and hardware reasons? wifi fw is loaded into KM0 ROM?

avtolstoy: :shrug: That's how it's implemented by Realtek. Wifi fw is in `ambd_sdk/project/realtek_amebaD_va0_example/GCC-RELEASE/project_lp/asdk/lib/application/lib_wifi_fw.a`, it does use some of the ROM functions, but in general it just runs on KM0 through a set of threads / interrupt handlers. It can run XIP, but XIP is seriously broken as usual, so we run it out of KM4 SRAM (as we are out of KM0 SRAM).

scott-brust (approved):

Kategrode:
https://app.shortcut.com/particle/story/126574/wi-fi-stability-jacuzzi-wifi-connection-fail-on-5-8-0-while-on-2-4ghz

PR #2768, opened 2024-04-27
[RTL872x] multiple SPI fixes

XuGuohui:
### Problem
1. `SPI.transfer(byte)` may lose data
2. SPI settings before `SPI.begin()` won't apply
3. `SPI.transfer(txBuf, nullptr, size, callback)` will lose data if CS pin is deasserted after the callback is invoked
4. SPI slave DMA state machine is messed up if master just toggles the CS pin

### Steps to Test
`user/tests/wiring/spi_master_slave`

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/spi_hal.cpp
+++ b/hal/src/rtl872x/spi_hal.cpp
@@ -193,6 +193,13 @@ public:
         }
         os_thread_scheduling(true, nullptr);
 
+        // Enable SPI Clock
+        if (rtlSpiIndex_ == 0) {
+            RCC_PeriphClockCmd(APBPeriph_SPI0, APBPeriph_SPI0_CLOCK, ENABLE);
+        } else {
+            RCC_PeriphClockCmd(APBPeriph_SPI1, APBPeriph_SPI1_CLOCK, ENABLE);
+        }
+
         std::lock_guard<Spi> lk(*this);
         if (isEnabled()) {
             CHECK(end());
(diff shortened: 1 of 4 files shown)


XuGuohui:
Tests to be done to make sure we don't run into regressions:
1. FRAM read/write
2. SD card
3. Neopixel
4. Gen3 <-> Gen4 mixed SPI master/slave tests

eberseth:
So far the master passes everything but the slave fails.  Commit 6334474e
```
Test 23_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_Synchronous_MODE2_LSB passed.
Test 24_SPI_Master_Slave_Master_Const_String_Transfer_DMA passed.
Test 25_SPI_Master_Slave_Master_Reception passed.
Test summary: 26 passed, 0 failed, and 0 skipped, out of 26 test(s).
```
```
Test 00_SPI_Master_Slave_Slave_Transfer passed.
Test 01_SPI_Master_Slave_Slave_Receiption passed.
Assertion failed: (SPI1.available()=0) == (strlen(txString)=1000), file tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp, line 366.
Test 02_SPI_Master_Slave_Slave_Const_String_Transfer_DMA failed.
Test summary: 2 passed, 1 failed, and 0 skipped, out of 3 test(s).
```

XuGuohui:
I have repeatedly run the `00_SPI_Master_Slave_Slave_Transfer` for whole night and so far so good. This is the most tricky test I've experienced, cos it just failed occassionally. I'll jump to tun the whole tests to figureout the failure of `02_SPI_Master_Slave_Slave_Const_String_Transfer_DMA`.

XuGuohui:
@eberseth Have you updated the system-part and test application for both SPI master and slave device?  I just ran the whole test several times and all passed.

XuGuohui:
The SPI master/slave fixture test always passed for me. I was using two Photon2(s) to run the test.

eberseth:
I am running two Borons.  I have updated the system parts on both.

eberseth (approved):

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -193,6 +194,13 @@ class Spi {
         }
         os_thread_scheduling(true, nullptr);
 
+        // Enable SPI Clock
```
scott-brust, line 197: Any reason we enable the clocks in `init()` now instead of `begin()`?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -216,47 +224,64 @@ class Spi {
...
             }
         }
+        return SYSTEM_ERROR_NONE;
+    }
 
-        // Set clock divider
-        if (config_.spiMode == SPI_MODE_MASTER) {
+    int setDataMode(uint8_t dataMode, bool force = false) {
+        config_.dataMode = dataMode;
+        if (SPI_DEV_TABLE[rtlSpiIndex_].SPIx->SSIENR & BIT_SSIENR_SSI_EN) {
```
scott-brust, line 252: This is a new check to see if the SPIx peripheral is enabled before configuring it? Was there a problem before when configuring data mode/clock/bit order without checking this?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -216,47 +224,64 @@ class Spi {
...
-                PAD_PullCtrl((uint8_t)hal_pin_to_rtl_pin(sclkPin_), GPIO_PuPd_UP);
+                SSI_SetDataSwap(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, BIT_CTRLR0_TXBITSWP, ENABLE);
+                SSI_SetDataSwap(SPI_DEV_TABLE[rtlSpiIndex_].SPIx, BIT_CTRLR0_RXBITSWP, ENABLE);
             }
         }
+        return SYSTEM_ERROR_NONE;
+    }
 
-        // Set clock divider
-        if (config_.spiMode == SPI_MODE_MASTER) {
```
scott-brust, line 237: The clock divider is now being set when operating in slave mode? I'm guessing this does not really do anything when in slave mode?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -445,7 +470,6 @@ class Spi {
 
     uint8_t transfer(uint8_t data) {
         CHECK_TRUE(isEnabled(), 0);
-        CHECK_TRUE(isBusy() == false, 0);
```
scott-brust, line 448: Removing this check allows the busy loop to spin below while a previous master transmission finishes, right? This kind of forces the calling thread to wait and prevent repeated `hal_spi_transfer()` calls from failing?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -445,7 +470,6 @@ class Spi {
...
         CHECK_TRUE(config_.spiMode == SPI_MODE_MASTER, 0);
```
scott-brust, line 474: I know this isnt a change you did in this PR, but shouldnt we return a non-zero error code if we are *not* in master mode? IE something like this:
```
CHECK_TRUE(config_.spiMode == SPI_MODE_MASTER, SYSTEM_ERROR_INVALID_STATE);
```
Or am I misunderstanding `CHECK_TRUE()`?

scott-brust (commented):
I tested the `spi_master_slave` wiring tests

- Photon 2 master SPI <--> Photon 2 slave SPI1
  - All fail: I assume SPI1 slave mode on RTL platform is still broken?
- Photon 2 Master SPI <--> Boron slave SPI1 
  - All passed: ```Test summary: 26 passed, 0 failed, and 0 skipped, out of 26 test(s).```
- Argon master SPI <--> Photon 2 slave SPI1
  - Mostly passed:

```
Assertion failed: (strncmp((const char *)SPI_Master_Rx_Buffer, "SPI Slave is doing good", sizeof("SPI Slave is doing good")) == 0=0) == (true=1), file tests/wiring/spi_master_s
lave/spi_master/spi_master.cpp, line 309.
Test 17_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_Synchronous_MODE1_LSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 1, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 495.
Test 18_SPI_Master_Slave_Master_Variable_Length_Transfer_No_DMA_MODE2_MSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 1, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 503.
Test 19_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_MODE2_MSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 1, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 511.
Test 20_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_Synchronous_MODE2_MSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 0, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 521.
Test 21_SPI_Master_Slave_Master_Variable_Length_Transfer_No_DMA_MODE2_LSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 0, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 529.
Test 22_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_MODE2_LSB failed.
Assertion failed: (SPI_Master_Slave_Change_Mode(0x02, 0, transferFunc)=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.cpp, line 537.
Test 23_SPI_Master_Slave_Master_Variable_Length_Transfer_DMA_Synchronous_MODE2_LSB failed.
Test 24_SPI_Master_Slave_Master_Const_String_Transfer_DMA passed.
Assertion failed: (strncmp((const char *)SPI_Master_Rx_Buffer_Supper, txString, strlen(txString)) == 0=0) == (true=1), file tests/wiring/spi_master_slave/spi_master/spi_master.
cpp, line 598.
Test 25_SPI_Master_Slave_Master_Reception failed.
Test summary: 18 passed, 8 failed, and 0 skipped, out of 26 test(s).
```

~I will review the DMA configuration / operation changes next. This looks like the most relevant changes in this PR~
I read through the DMA operation and it seems to make sense to me. I think the biggest change here is the DCache invalidation steps with the tx/rx buffers. 
I would have to debug the tests and HAL in depth to understand the above test failures since nothing is obviously wrong to me...

XuGuohui:
@scott-brust Have you updated the system-part that is built against the branch in this PR for Photon2 slave role as well?

hal/src/rtl872x/spi_hal.cpp
XuGuohui: Just in case. There is operation to the SPI peripheral in the `init()`

hal/src/rtl872x/spi_hal.cpp
XuGuohui: The intention was to make sure the `setDataMode()` (clock/bit_order) just set internal variables if `begin()` is not called yet. Also the SDK APIs demonstrate we should disable the peripheral before setting these configurations.

hal/src/rtl872x/spi_hal.cpp
XuGuohui: Clock setting doesn't apply to slave mode, so it's fine to remove the check.

hal/src/rtl872x/spi_hal.cpp
XuGuohui: Yes, it's consistent with Gen3. And this is the root cause why `SPI.transfer(byte)` loss data in some circumstances.

hal/src/rtl872x/spi_hal.cpp
XuGuohui: It is consistent with Gen3. And the wiring API `SPI.transfer()` returns an unsigned `byte` value, so it's fine to return `0` here.

scott-brust: Ok, makes sense. Returning `0` to indicate no data was sent seems correct. I was misinterpreting it as `SYSTEM_ERROR_NONE`

scott-brust:
> @scott-brust Have you updated the system-part that is built against the branch in this PR for Photon2 slave role as well?

Yes, Ill double check it tomorrow. Should all the tests pass with P2 Master <-> P2 slave? Or just Gen3 Master <-> P2 slave?

XuGuohui:
The fixture tests all passed with P2 Master <-> P2 slave on my side.

scott-brust:
> The fixture tests all passed with P2 Master <-> P2 slave on my side.

I cant get them to consistently pass. I always get failures at some point in the test run. Things break after that.  
Here is a log and capture of the tests failing, around 79 seconds into the capture. I have two devices connected with jumper wire on a breadboard, not sure if this is introducing errors or what. 
[spi_master_slave_test_failures.zip](https://github.com/particle-iot/device-os/files/15324246/spi_master_slave_test_failures.zip)

user/tests/wiring/spi_master_slave/spi_slave/spi_slave.cpp
eberseth, line 1: Please check the P2 wiring diagrams around line 146.  The SPI1 to SPI1 connections seem wrong

scott-brust (approved):
With latest revision P2 Master/slave tests pass both with SPI and SPI1 peripherals

PR #2769, opened 2024-05-13
Resumable OTA fixes

sergeuz:
### Problem

- `DISCARD_DATA` flag has no effect when it's set in a `StartUpdate` OTA request as opposed to `FinishUpdate`.
- Cached module data isn't invalidated if the module validation failed.

### References

- sc-127555


avtolstoy (approved):
Tested by manually triggering an error.

PR #2770, opened 2024-05-20
Allow temporarily disabling device protection

sergeuz:
### Description

Implement a challenge-response protocol between the device and the cloud to temporarily disable device protection. An external tool like the CLI transfers the information from the device to the API and back.

### Changes

Behavior of protected mode
* All the USB USART functions return a neutral result when called when device protection is currently active.
* LogHandler changes for protected mode are reversed.
  - The consequence is that `SerialLogHandler` does not output anything in protected mode, but starts outputing when protection is temporarily disabled, and other LogHandlers (including custom ones) still output in protected mode.

Control requests
* GetProtectedStateReply
  - `state` returns true when the device is currently protected
  - `overriden` returns true when the device would have been protected but protection is temporarily disabled
* SetProtectedStateRequest. New request (see below for use)

Unprotect algorithm
* SetProtectedStateRequest has 3 steps: prepare, confirm and reset.
* Prepare generates a random device nonce, and responds with a signature of the device id, the server nonce, and the device nonce. The response also include the fingerprint of the device key used for the signature. It also saves the time when prepare was done.
* Confirm checks that prepare was started less than a minute ago. It first checks that the server that generated the signature is the one we have a key for (KEY_MISMATCH error). It then verifies the signature sent by the server (INVALID_SIGNATURE error).
  - If everything checks out, device protection is disabled for 20 reboots and 24 hours.
* Reset clears the temporary unlock.
* The counter of remaining resets and remaining seconds for the temporary unlock is stored in system flags.
* In the bootloader system tick, a 1 second counter is decremented. When that gets to 0, the remaining seconds for the temporary unlock is decremented and the system flags are saved.
* Outside the bootloader, a 1 second OS timer is used to decrement the remaining seconds for the temporary unlock and save the system flags.

### References

- sc-127946


proto_defs/src/control/config.pb.h
```diff
@@ -47,6 +47,13 @@ typedef enum _particle_ctrl_Feature {
     particle_ctrl_Feature_ETHERNET_DETECTION = 1 /* Ethernet FeatherWing detection on boot */
 } particle_ctrl_Feature;
 
+/* Note: Use the same values as defined by the security_key_type enum */
+typedef enum _particle_ctrl_SetProtectedStateRequest_Action { 
+    particle_ctrl_SetProtectedStateRequest_Action_RESET = 0, /* Invalid key */
+    particle_ctrl_SetProtectedStateRequest_Action_DISABLE_REQUEST = 1, /* TCP device private key */
+    particle_ctrl_SetProtectedStateRequest_Action_DISABLE_CONFIRM = 2 /* TCP device public key */
```
monkbroc: These comments are weird. Where do they come from?

sergeuz: Yeah, saw it in the diff too. It's most likely a bug in the nanopb generator :shrug:

system/src/control/config.cpp
```diff
@@ -55,8 +60,203 @@ namespace control {
 
 namespace config {
 
+namespace {
+
 using namespace particle::control::common;
 
+struct SecurityModeChangeContext {
+    char clientNonce[32];
```
monkbroc: This will eventually contain a start time for the 1 minute lifetime of the client nonce, right?

system/src/control/config.cpp
```diff
@@ -55,8 +60,203 @@ namespace control {
...
+    HAL_Core_Write_Backup_Register(BKP_DR_08, 0); // Use default security mode
+}
+
+inline bool isSecurityModeOverridden() {
+    return HAL_Core_Read_Backup_Register(BKP_DR_08);
+}
+
+// Completion handler for system_ctrl_set_result()
+void systemResetCompletionHandler(int result, void* data) {
+    system_reset(SYSTEM_RESET_MODE_NORMAL, RESET_REASON_CONFIG_UPDATE, 0 /* value */, 0 /* flags */, nullptr /* reserved */);
```
monkbroc: Will this reset the device after the unlock request succeeds? Can this be avoided? We want to support use cases like gathering logs from a running protected device without resetting it.

sergeuz: Yeah, what's implemented here so far is the simplest and most reliable way to disable the protection and it does require a reset. I wanted to focus on the challenge protocol first and then look into what it would take to support changing the security mode on the fly. There are some technical implications around that, mostly about sychronization.

Note though that the way things are implemented now, logging can't be re-enabled without a reset anyway. We'll need to change that as well if that's the use case we'd want to support.

system/src/control/config.cpp
sergeuz: Yes :+1:

services/src/security_mode.cpp
```diff
@@ -108,6 +187,14 @@ int security_mode_check_request(security_mode_transport transport, uint16_t id)
         }
     }
 
+    if (transport == SECURITY_MODE_TRANSPORT_USB) {
+        switch (id) {
+            case CTRL_REQUEST_SET_PROTECTED_STATE: {
```
monkbroc: Let's allow CTRL_REQUEST_SET_PROTECTED_STATE under BLE as well so a mobile app can be used to unprotect a device.

services/src/security_mode.cpp
```diff
@@ -16,21 +16,121 @@
...
+    }
+    if (ext.security_mode == MODULE_INFO_SECURITY_MODE_PROTECTED) {
+        sNormalSecurityMode = ext.security_mode;
+        Load_SystemFlags();
+        if (system_flags.security_mode_override_value != 0xff) {
+            sCurrentSecurityMode = system_flags.security_mode_override_value;
+        } else {
+            sCurrentSecurityMode = sNormalSecurityMode;
+        }
+    }
```
monkbroc, line 135: I think you need an else to clear the security mode override if the device is not protected, otherwise the following situation can happen:
* device is protected
* device gets unlocked (protected = false, overriden = true)
* open bootloader gets flashed (protected = false, overriden = true)
* protected bootloader gets flashed. At this point, protected should be true and overriden should be false, but because the `security_mode_override_value` wasn't cleared, we get protected = false, overriden = true

sergeuz: Good catch, @keeramis! Fixed in https://github.com/particle-iot/device-os/pull/2770/commits/c116ce7e55a9710612c8e5eb07dceea2f21b5496.

sergeuz: Ended up refactoring that function so that it would also clear the override if it for whatever reason failed to parse the bootloader extensions.

sergeuz:
I'm marking this PR as ready for review even though it currently overflows on some of the platforms – we'll investigate that separately. The test steps can be found in sc-127946.

monkbroc (approved):

services/src/security_mode.cpp
```diff
@@ -120,3 +221,122 @@ int security_mode_check_request(security_mode_transport transport, uint16_t id)
...
+
+void security_mode_override_to_none() {
+    if (sNormalSecurityMode == MODULE_INFO_SECURITY_MODE_NONE) {
+        return;
+    }
+    ATOMIC_BLOCK() {
+        Load_SystemFlags();
+        system_flags.security_mode_override_value = MODULE_INFO_SECURITY_MODE_NONE;
+        system_flags.security_mode_override_reset_count = 20;
+        system_flags.security_mode_override_timeout = 24 * 60 * 60; // Seconds
```
avtolstoy, line 233: Nit: define a constant for this?

avtolstoy (approved):

PR #2771, opened 2024-05-21
Publish last reset event on session resumption

monkbroc:
### Problem

When the device reboots and resumes its session (for example after `System.reset()` and after pin reset on Gen 3), the `spark/device/last_reset` event is not published. It's only published when a new session is established.

### Solution

Publish last reset reason whenever the first cloud connection is established, regardless if it's a new session or a resumed session.

Note: there's a lot of duplication in `if (!session_resumed)` part of `Spark_Handshake` and it's not clear if additional things in the `if` should also be done in the `else`. Please think about this as part of the review for this PR.

### Steps to Test

Ensure a `spark/device/last_reset` is published in the following conditions
- Power on reset
- Pin reset (check Gen 3 & Gen 4)
- `particle usb reset`

### Example App

When device resets from `System.reset()`, `spark/device/last_reset` with data `user` should be published.

```c
void setup()
{
   System.enable(SYSTEM_FLAG_PUBLISH_RESET_INFO);
   System.enableFeature(FEATURE_RESET_INFO);
}
void loop()
{
  if (Particle.connected()) {
        delay(10000);
        System.reset();
  }
}
```

### References

https://app.shortcut.com/particle/story/128022/publish-last-reset-event-on-session-resumption

---

### Completeness

- [ ] Added documentation


eberseth:
@monkbroc Could you rebase this onto develop?

PR #2773, opened 2024-05-22
[bootloader] Implement bootloader reset reasons

scott-brust:
### Problem
The RTL bootloader has not been updated to use the backup register HAL to communicate reset reason to the system part application. This leads to undefined reset reasons in the following scenarios

1. When completing OTA bootloader update
2. When completing DFU update via button reset/hold

### Solution
The backup register HAL is implemented on RTL platform via retained SRAM, the bootloader just needs to write the "registers"

### Steps to Test
Checkout branch, compile bootloader, flash
Flash app via reset button hold and DFU
Log reset reason ie (`Log.info("Last Reset Reason %d", System.resetReason());`)
Verify that it is correct (`70` = firmware update, `40` = pin/power reset):
```
0000008051 [app] INFO: Last Reset Reason 70
```

Without bootloader update, the values are undefined:
```
0000058978 [app] INFO: Last Reset Reason -49520187
```

### References

See comment [here](https://s.slack.com/archives/C06G6BGTNS0/p1713393224076049)

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
+++ b/bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
@@ -223,7 +223,7 @@ static uint32_t boot_reason_set(void) {
     /* backup boot reason to REG_LP_SYSTEM_CFG2 */
     temp_bootcfg = HAL_READ32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2);
     temp_bootcfg |= tmp_reason;
-    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason); // Should tmp_reason be `temp_bootcfg`? If we have deepsleep or brownout boots, we will clear some of the bits in REG_LP_SYSTEM_CFG2
+    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason);
     
     return 0;
 }
--- a/bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
+++ b/bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
@@ -223,7 +223,7 @@ static uint32_t boot_reason_set(void) {
     /* backup boot reason to REG_LP_SYSTEM_CFG2 */
     temp_bootcfg = HAL_READ32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2);
     temp_bootcfg |= tmp_reason;
-    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason);
+    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason); // Should tmp_reason be `temp_bootcfg`? If we have deepsleep or brownout boots, we will clear some of the bits in REG_LP_SYSTEM_CFG2
     
     return 0;
 }
--- a/system/src/system_cloud_internal.cpp
+++ b/system/src/system_cloud_internal.cpp
@@ -695,6 +695,23 @@ bool publishSafeModeEventIfNeeded() {
     return true; // ok
 }
 
+void publishResetReasonIfNeeded() {
+    uint8_t flag = 0;
+    if (system_get_flag(SYSTEM_FLAG_PUBLISH_RESET_INFO, &flag, nullptr) == 0 && flag)
+    {
+        system_set_flag(SYSTEM_FLAG_PUBLISH_RESET_INFO, 0, nullptr); // Publish the reset info only once
+        int reason = RESET_REASON_NONE;
+        uint32_t data = 0;
+        if (HAL_Core_Get_Last_Reset_Info(&reason, &data, nullptr) == 0 && reason != RESET_REASON_NONE)
+        {
+            char buf[64];
+            formatResetReasonEventData(reason, data, buf, sizeof(buf));
+            LOG(INFO,"Send spark/device/last_reset event");
+            publishEvent("spark/device/last_reset", buf);
+        }
+    }
+}
+
 void handleServerMovedRequest(const char* reqData, size_t reqSize, ServerMovedResponseCallback respCallback, void* ctx) {
     clear_system_error_message();
 #if PLATFORM_ID != PLATFORM_GCC && PLATFORM_ID != PLATFORM_NEWHAL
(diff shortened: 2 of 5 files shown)


bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
```diff
@@ -223,7 +223,7 @@ static uint32_t boot_reason_set(void) {
     /* backup boot reason to REG_LP_SYSTEM_CFG2 */
     temp_bootcfg = HAL_READ32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2);
     temp_bootcfg |= tmp_reason;
-    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason);
+    HAL_WRITE32(SYSTEM_CTRL_BASE_LP, REG_LP_SYSTEM_CFG2, tmp_reason); // Should tmp_reason be `temp_bootcfg`? If we have deepsleep or brownout boots, we will clear some of the bits in REG_LP_SYSTEM_CFG2
```
scott-brust: This is unrelated to the PR, but I think there is a subtle bug in the RTK SDK, where the `REG_LP_SYSTEM_CFG2` might be getting written incorrectly on boot in some circumstances??

user/tests/wiring/watchdog/watchdog.cpp
```diff
@@ -47,7 +47,8 @@ test(WATCHDOG_00_setup_disconnect_power_off_ncp) {
     Cellular.off();
     waitFor(Cellular.isOff, 120000);
     assertTrue(Cellular.isOff());
-#elif Wiring_WiFi
+#endif
+#if Wiring_WiFi
```
scott-brust, line 51: Not related to the bootloader, but just a subtle thing I noticed while testing wdog on msom. Just a convenient spot to include this change

eberseth:
Just a general question on whether we should be providing a reason for disconnection in the vitals publish as well?  I did a button reset and the disconnect reason was "none".  I then powered off my msom and got a reason of "unknown".

```json
{
  "connection": {
    "status": "connecting",
    "error": 0,
    "attempts": 1,
    "disconnects": 0,
    "disconnect_reason": "none",
    "interface": "Cellular"
  },
  "coap": {
    "transmit": 9,
    "retransmit": 0,
    "unack": 0,
    "round_trip": 1526
  },
  "publish": {
    "rate_limited": 0
  }
}
```

bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
eberseth: Good question.  @avtolstoy ?

eberseth (approved):

monkbroc:
Connection vitals are not very useful. The counters tend to be 0 or 1. I think it's out of scope to rework that to be something more useful right now, but worth considering perhaps as part of ACM work.

eberseth:
> Connection vitals are not very useful. The counters tend to be 0 or 1. I think it's out of scope to rework that to be something more useful right now, but worth considering perhaps as part of ACM work.

I agree changing it may be out-of-scope of the reason fixes.  Just thinking that since we already know that we have reset for OTA or other that it would be easy to determine and report.

scott-brust:
> > Connection vitals are not very useful. The counters tend to be 0 or 1. I think it's out of scope to rework that to be something more useful right now, but worth considering perhaps as part of ACM work.
> 
> I agree changing it may be out-of-scope of the reason fixes. Just thinking that since we already know that we have reset for OTA or other that it would be easy to determine and report.

The vitals mostly contain [system diagnostics](https://github.com/particle-iot/device-os/blob/develop/services/inc/diagnostics.h#L89) stuff. The "connection" structure you point out is mostly the "Network" diagnostics. There are similar "Cloud" diagnostics (ie connect attempts, disconnects, disconnect reason). I dont know how well the Network disconnect reason diagnostic works

These are distinct from the `System.resetReason()`, which is not stored as a system diagnostic. I dont know if this is included in the vitals publish or not. I believe it should be published as a separate event

scott-brust:
Pulling in changes from https://github.com/particle-iot/device-os/pull/2771

system/src/system_cloud_internal.cpp
```diff
@@ -1267,22 +1284,7 @@ int Spark_Handshake(bool presence_announce)
         }
```
scott-brust: No idea if these are still relevant events to publish, as that seems to be the difference between resuming session and not.

system/src/system_cloud_internal.cpp
```diff
@@ -1293,6 +1295,7 @@ int Spark_Handshake(bool presence_announce)
         LOG(INFO,"cloud connected from existing session.");
 
         publishSafeModeEventIfNeeded();
+        publishResetReasonIfNeeded();
         Send_Firmware_Update_Flags();
```
scott-brust: feels like we should be able to move these out of the if/else block since theyre in both. I dont think their order particularly matters

eberseth: I would think that it is safe to do so.

system/src/system_cloud_internal.cpp
monkbroc: Publishing spark/device/ident is 100% dead code. It's from the 2015 PR https://github.com/particle-iot/device-os/pull/621

The device service doesn't do anything with this event if it was ever published which it won't since `HAL_Get_Device_Identifier` is an empty function on all platforms.

monkbroc: spark/device/claim/code is still a live code path but we don't use claim codes in the latest dev tools. We shouldn't delete this code, but it would be ok to send the claim code whether the connection is resumed or new. The device service will respond with spark/device/claim/confirm when it receives the claim code. When the device receives this, it will clear the claim code using `HAL_Set_Claim_Code(NULL)` and mark the device as claimed.

system/src/system_cloud_internal.cpp
monkbroc: I agree. See my comments above.

monkbroc (approved):

bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
avtolstoy: I think the only stuff we'd lose is some ROM-specific params, the boot reason should be accurate irrespective of that as it's collected from AON/SYSON registers into `tmp_reason`.

avtolstoy (approved):

bootloader/prebootloader/src/rtl872x/part1/rtl_support.c
scott-brust: Yes, there is nothing in there that we would care about. Just odd to me that it seems realtek would have a bug like this. I will remove this comment since I dont think we want to "fix" this problem

PR #2774, opened 2024-05-23
Supports auxiliary power control.

XuGuohui:
**NOTE: this PR is targeting the `feature/muon-som-evb` branch**

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

#define M2_BREAKOUT_BOARD   1

#if M2_BREAKOUT_BOARD
uint8_t auxPwrPin = D23;
#else // Muon
uint8_t auxPwrPin = D7;
#endif

STARTUP(

    conf.powerSourceMinVoltage(3880)
        .powerSourceMaxCurrent(3000)
        .batteryChargeVoltage(4200)
        .batteryChargeCurrent(900)
        .auxPowerControlPin(auxPwrPin)
        .feature(SystemPowerFeature::PMIC_DETECTION);
        // .feature(SystemPowerFeature::DISABLE);
    System.setPowerConfiguration(conf);

#if M2_BREAKOUT_BOARD
    remap.cs_pin = HAL_PLATFORM_ETHERNET_WIZNETIF_CS_PIN_DEFAULT; // default
    remap.reset_pin = HAL_PLATFORM_ETHERNET_WIZNETIF_RESET_PIN_DEFAULT; // default
    remap.int_pin = PIN_INVALID;
#else // Muon
    remap.cs_pin = A3;
    remap.reset_pin = PIN_INVALID;
    remap.int_pin = A4;
#endif
);

void setup() {
}

void loop() {
    if (Serial1.available()) {
        char c = Serial1.read();
        switch (c) {
            case 'i': {
                if_list* ifs = nullptr;
                if_get_list(&ifs);
                    if (iface->iface) {
                        uint8_t idx;
                    }
                }
                if_free_list(ifs);
                break;
            }
            case 'p': {
                Particle.connect();
                break;
            }
            case 'e': {
                Ethernet.connect();
                break;
            }
            case 'd': {
                Particle.disconnect();
                break;
            }
            default: break;
        }
    }
}
```

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/power_hal.h
+++ b/hal/inc/power_hal.h
@@ -22,6 +22,11 @@
 #include <stddef.h>
 #include "static_assert.h"
 
+#define HAL_POWER_CONFIG_VERSION_0          0
+#define HAL_POWER_CONFIG_VERSION_1          1
+
+#define HAL_POWER_CONFIG_VERSION            HAL_POWER_CONFIG_VERSION_1
+
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus */
@@ -61,13 +66,9 @@ typedef struct hal_power_config {
     uint16_t charge_current; // charge current
     uint16_t termination_voltage; // termination voltage
     uint8_t soc_bits; // bits precision for SoC calculation (18 (default) or 19)
-    /* Since the old firmware resets all the reserved fields to 0, we use this field
-     * to determine if the aux_pwr_ctrl_pin being 0 is set by user or just inherits
-     * from the old DCT due to firmware upgrading. */
-    uint8_t aux_pwr_ctrl;
     uint8_t aux_pwr_ctrl_pin; // pin number for auxiliary power control
     uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
-    uint8_t reserved2[2];
+    uint8_t reserved2[3];
     uint32_t reserved3[3];
 } hal_power_config;
 static_assert(sizeof(hal_power_config) == 32, "hal_power_config size changed");
--- a/hal/inc/power_hal.h
+++ b/hal/inc/power_hal.h
@@ -70,7 +70,7 @@ typedef struct hal_power_config {
     uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
     uint8_t reserved2[3];
     uint32_t reserved3[3];
-} hal_power_config;
+} __attribute__((packed)) hal_power_config;
 static_assert(sizeof(hal_power_config) == 32, "hal_power_config size changed");
 
 int hal_power_load_config(hal_power_config* conf, void* reserved);
--- a/hal/inc/power_hal.h
+++ b/hal/inc/power_hal.h
@@ -61,8 +61,13 @@ typedef struct hal_power_config {
     uint16_t charge_current; // charge current
     uint16_t termination_voltage; // termination voltage
     uint8_t soc_bits; // bits precision for SoC calculation (18 (default) or 19)
+    /* Since the old firmware resets all the reserved fields to 0, we use this field
+     * to determine if the aux_pwr_ctrl_pin being 0 is set by user or just inherits
+     * from the old DCT due to firmware upgrading. */
+    uint8_t aux_pwr_ctrl;
     uint8_t aux_pwr_ctrl_pin; // pin number for auxiliary power control
     uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
+    uint8_t reserved2[2];
     uint32_t reserved3[3];
 } hal_power_config;
 static_assert(sizeof(hal_power_config) == 32, "hal_power_config size changed");
--- a/hal/inc/power_hal.h
+++ b/hal/inc/power_hal.h
@@ -61,8 +61,9 @@ typedef struct hal_power_config {
     uint16_t charge_current; // charge current
     uint16_t termination_voltage; // termination voltage
     uint8_t soc_bits; // bits precision for SoC calculation (18 (default) or 19)
-    uint8_t reserved2;
-    uint32_t reserved3[4];
+    uint8_t aux_pwr_ctrl_pin; // pin number for auxiliary power control
+    uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
+    uint32_t reserved3[3];
 } hal_power_config;
 static_assert(sizeof(hal_power_config) == 32, "hal_power_config size changed");
 
--- a/hal/src/b5som/network/network.cpp
+++ b/hal/src/b5som/network/network.cpp
@@ -152,6 +152,23 @@ int if_init_platform(void*) {
     return 0;
 }
 
+int if_init_platform_postpone(void*) {
+    if (en2) {
+        uint8_t dummy;
+        if (!if_get_index(en2->interface(), &dummy)) {
+            auto wizNetif = reinterpret_cast<WizNetif*>(en2);
+            if (wizNetif->isPresent()) {
+                return 0;
+            }
+            netifapi_netif_remove(en2->interface());
+        }
+        LOG(INFO, "Ethernet interface is not present");
+        delete en2;
+        en2 = nullptr;
+    }
+    return 0;
+}
+
 extern "C" {
 
 struct netif* lwip_hook_ip4_route_src(const ip4_addr_t* src, const ip4_addr_t* dst) {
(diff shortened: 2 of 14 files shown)


system/src/main.cpp
```diff
@@ -735,6 +735,11 @@ void app_setup_and_loop(void)
     }
 #endif // HAL_PLATFORM_ASSETS
 
+    // Reset all persistent settings to factory defaults if necessary
+    // This should be called before system_part2_post_init() so that the
+    // configurations set in STARTUP() is not reset by factory reset
+    resetSettingsToFactoryDefaultsIfNeeded();
```
XuGuohui, line 749: Just wondering whether this makes sense or not. If not, I'll revert the change.

hal/src/b5som/network/network.cpp
```diff
@@ -152,6 +152,23 @@ int if_init_platform(void*) {
...
 
+int if_init_platform_postpone(void*) {
+    if (en2) {
+        uint8_t dummy;
+        if (!if_get_index(en2->interface(), &dummy)) {
+            auto wizNetif = reinterpret_cast<WizNetif*>(en2);
+            if (wizNetif->isPresent()) {
+                return 0;
+            }
+            netifapi_netif_remove(en2->interface());
```
scott-brust, line 163: The logic here is
1. User has enabled ethernet detection
2. The system power management initialization is complete
3. We still cannot communicate with the wiznet chip, so we assume it is powered off
4. We remove it from the list of valid lwip netifs (this is new behavior compared to previously where it would be left in the list, just always in the powered off state)

system/src/main.cpp
scott-brust: This seems correct to do? But also ok to accept that user settings will be lost when doing a factory reset? 

I am thinking about another scenario where a user app uses DCT settings in `STARTUP()` but then the DCT settings are factory reset after STARTUP. (this is the current behavior?)

system/src/system_power_manager.cpp
```diff
@@ -353,72 +414,10 @@ void PowerManager::loop(void* arg) {
...
   self->thread_ = os_thread_current(nullptr);
 
-  {
-    // FIXME: perform these before creating the thread
-
-    // Load configuration
-    self->loadConfig();
-
-    if (self->config_.flags & HAL_POWER_MANAGEMENT_DISABLE) {
-      LOG(WARN, "Disabled by configuration");
```
scott-brust, line 363: I think we should still log this warning?

system/src/system_power_manager.cpp
```diff
@@ -137,13 +138,73 @@ PowerManager* PowerManager::instance() {
...
+    hal_gpio_write(config_.aux_pwr_ctrl_pin, config_.aux_pwr_ctrl_pin_level);
+  }
+
+  if (config_.flags & HAL_POWER_MANAGEMENT_DISABLE) {
+    return;
+  }
+
+#if HAL_PLATFORM_POWER_MANAGEMENT_OPTIONAL
+  if (!detect()) {
+    return;
```
scott-brust, line 167: Maybe Log a warning here to make it clear PMIC isnt detected?

scott-brust: Also, previously this would call `self->deinit();`
Do we need to call this still in this case or if we are exiting during init, there is nothing to de-init at this point? (maybe just unconfigure the aux power control pin?)

system/src/system_power_manager.cpp
```diff
@@ -353,72 +414,10 @@ void PowerManager::loop(void* arg) {
   PowerManager* self = PowerManager::instance();
   self->thread_ = os_thread_current(nullptr);
 
-  {
-    // FIXME: perform these before creating the thread
```
scott-brust, line 357: Thank you for taking the opportunity to fix this. This PR Makes it much easier to read the power manager loop with all the initialization code move to the init() function

hal/inc/power_hal.h
```diff
@@ -61,8 +61,9 @@ typedef struct hal_power_config {
     uint16_t charge_current; // charge current
     uint16_t termination_voltage; // termination voltage
     uint8_t soc_bits; // bits precision for SoC calculation (18 (default) or 19)
-    uint8_t reserved2;
-    uint32_t reserved3[4];
+    uint8_t aux_pwr_ctrl_pin; // pin number for auxiliary power control
+    uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
+    uint32_t reserved3[3];
 } hal_power_config;
 static_assert(sizeof(hal_power_config) == 32, "hal_power_config size changed");
```
scott-brust, line 74: To be clear, we are using `uint8_t reserved2` as well as one item of the `uint32_t reserved3[]` array to store the new pin and pin state?

I am assuming since we dont pack this structure that this static assert still passes? There should be any issues reading this out of DCT?

scott-brust (approved):
I tested MSOM on an old EVB, with the onboard wiznet chip (just to make things easier)

I confirmed that the D23 / aux power pin is asserted on boot before setup() completes, and that the en2 netif is populated (because the EVB already powers it on by default).

I confirmed that when pulling the ethernet jumpers, and breaking ethernet communication to the msom, that the en2 netif is removed from the list of lwip IFs after boot. 

I think the PR as currently written can be used for muon / m2 breakout to control the aux power at boot

hal/inc/power_hal.h
XuGuohui: Good catch. we should only consume one byte of the `reserved3` field.

hal/src/b5som/network/network.cpp
XuGuohui: It did not keep the interface in the list previously if it is invalid, since it was not even successfully add to the list. It just reserves the index if failed to add the interface.

system/src/main.cpp
XuGuohui: Yes, it is the current behavior herein. If user is going to do a factory reset, then it doesn't make perfect sense that user app wants to restore something old from the DCT, but it would make more sense that user may want to set some new configurations to a brand new DCT, i.e after factory reset？

system/src/system_power_manager.cpp
XuGuohui: I think the previous `goto exit`, in which it calls `self->deinit()` is mainly for calling `os_thread_exit()` to exit the thread. But here is the very begining of the `init()` function, if the PMIC is not detected we can simply return. As for the aux power control pin, it is required when the power module for DC power supply is installed, on which there is no PMIC, but we still need to enable the aux power control pin.

eberseth (approved):

eberseth (approved):

hal/inc/power_hal.h
```diff
@@ -61,8 +66,10 @@ typedef struct hal_power_config {
     uint16_t charge_current; // charge current
     uint16_t termination_voltage; // termination voltage
     uint8_t soc_bits; // bits precision for SoC calculation (18 (default) or 19)
-    uint8_t reserved2;
-    uint32_t reserved3[4];
+    uint8_t aux_pwr_ctrl_pin; // pin number for auxiliary power control
+    uint8_t aux_pwr_ctrl_pin_level; // active level for auxiliary power control
+    uint8_t reserved2[3];
+    uint32_t reserved3[3];
```
avtolstoy, line 72: Let's add `((packed))` just in case.

hal/src/b5som/network/network.cpp
```diff
@@ -127,7 +127,7 @@ int if_init_platform(void*) {
         wizNetifConfigData.size = sizeof(WizNetifConfigData);
         wizNetifConfigData.version = WIZNETIF_CONFIG_DATA_VERSION;
         WizNetifConfig::instance()->getConfigData(&wizNetifConfigData);
-        en2 = new WizNetif(HAL_SPI_INTERFACE1, wizNetifConfigData.cs_pin, wizNetifConfigData.reset_pin, wizNetifConfigData.int_pin, mac);
+        en2 = new WizNetif(HAL_SPI_INTERFACE1, wizNetifConfigData.cs_pin, wizNetifConfigData.reset_pin, wizNetifConfigData.int_pin, mac, true);
```
avtolstoy: Nit: add `true /* postpone initialization */` (same in other places for readability)

avtolstoy (approved):

wiring/inc/spark_wiring_system_power.h
```diff
@@ -105,6 +107,20 @@ class SystemPowerConfiguration {
         return conf_.soc_bits;
     }
 
+    SystemPowerConfiguration& auxPowerControlPin(uint8_t pin, bool activeLevel = 1) {
```
avtolstoy: Nit: I'd use more verbose names here e.g. auxiliaryPowerControlPin() just for readability.

PR #2775, opened 2024-05-24
Makes the Ethernet interrupt pin optional.

XuGuohui:
**NOTE: this PR is targeting the `feature/muon-som-evb` branch**

### Problem

Currently W5500 interrupt pin is essential for Ethernet to be functional. However, for the Ethernet feather wing from Adafruit doesn't expose the interrupt pin to side headers, indicating that the interrupt pin is optional to the Ethernet interface.

### Solution

Allow the `int_pin` in `WizNetifConfigData` to be `PIN_INVALID` and use query scheme in `WizNetif` when the interrupt pin is not present.

### Steps to Test

1. Use the M.2 SoM breakout board and install the Ethernet feather wing from Adafruit. Do not fly wire the IRQ pin on the wing to any of the Particle SoM's IOs.
2. Run the following test app.

### Example App

```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

STARTUP(

    remap.cs_pin = PIN_INVALID; // default
    remap.reset_pin = PIN_INVALID; // default
    remap.int_pin = PIN_INVALID;
);

void setup() {
    while (!Serial.isConnected()) {
        delay(100);
    }
}

void loop() {
    if (Serial.available()) {
        char c = Serial.read();
        switch (c) {
            case 'i': {
                if_list* ifs = nullptr;
                if_get_list(&ifs);
                    if (iface->iface) {
                        uint8_t idx;
                    }
                }
                if_free_list(ifs);
                break;
            }
            case 'p': {
                Particle.connect();
                break;
            }
            case 'e': {
                Ethernet.connect();
                break;
            }
            case 'd': {
                Particle.disconnect();
                break;
            }
            default: break;
        }
    }
}
```

### References

N/A

---

### Completeness

- [ ] (REQUIRED) Added documentation

### DIFF
--- a/hal/network/lwip/wiznet/wiznetif.cpp
+++ b/hal/network/lwip/wiznet/wiznetif.cpp
@@ -337,7 +337,7 @@ void WizNetif::loop(void* arg) {
         if (p) {
             self->output(p);
         }
-        if (self->inRecv_) {
+        if (self->inRecv_ || self->interrupt_ == PIN_INVALID) {
             int r = 1;
             if (!p) {
                 r = self->input();
--- a/hal/network/lwip/wiznet/wiznetif.cpp
+++ b/hal/network/lwip/wiznet/wiznetif.cpp
@@ -116,6 +116,7 @@ const hal_spi_info_t WIZNET_DEFAULT_CONFIG = {
 };
 
 const int WIZNET_DEFAULT_TIMEOUT = 500;
+const int WIZNET_DEFAULT_TIMEOUT_POLL = 250;
 /* FIXME */
 const unsigned int WIZNET_INRECV_NEXT_BACKOFF = 50;
 const unsigned int WIZNET_DEFAULT_RX_FRAMES_PER_ITERATION = PBUF_POOL_SIZE / 2;
@@ -333,7 +334,7 @@ void WizNetif::loop(void* arg) {
     while(!self->exit_) {
         pbuf* p = nullptr;
         os_queue_take(self->queue_, (void*)&p, timeout, nullptr);
-        timeout = WIZNET_DEFAULT_TIMEOUT;
+        timeout = (self->interrupt_ == PIN_INVALID) ? WIZNET_DEFAULT_TIMEOUT_POLL : WIZNET_DEFAULT_TIMEOUT;
         if (p) {
             self->output(p);
         }
--- a/hal/network/lwip/wiznet/wiznetif_config.cpp
+++ b/hal/network/lwip/wiznet/wiznetif_config.cpp
@@ -147,9 +147,6 @@ int WizNetifConfig::getConfigData(WizNetifConfigData* configData) {
     if (configData->reset_pin == PIN_INVALID) {
         configData->reset_pin = HAL_PLATFORM_ETHERNET_WIZNETIF_RESET_PIN_DEFAULT;
     }
-    if (configData->int_pin == PIN_INVALID) {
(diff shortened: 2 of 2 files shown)


mrlambchop:
How does this work in safe mode when the app can't reconfigure the ethernet interface?

XuGuohui:
> How does this work in safe mode when the app can't reconfigure the ethernet interface?

The Ethernet config data is stored in system cache, which is a file in littlefs.

eberseth:
Does this Ethernet remaping interface already address this PR?  https://docs.particle.io/reference/device-os/api/ethernet/ethernet/#pin-configuration-ethernet

XuGuohui:
> Does this Ethernet remaping interface already address this PR? https://docs.particle.io/reference/device-os/api/ethernet/ethernet/#pin-configuration-ethernet

No, that feature requires the INT pin to be a valid pin. This PR removes the constraint based on that feature and also makes minor changes to the WizNetif to make sure Ethernet can receive packets without relying on interrupt.

hal/network/lwip/wiznet/wiznetif_config.cpp
```diff
@@ -147,9 +147,6 @@ int WizNetifConfig::getConfigData(WizNetifConfigData* configData) {
     if (configData->reset_pin == PIN_INVALID) {
         configData->reset_pin = HAL_PLATFORM_ETHERNET_WIZNETIF_RESET_PIN_DEFAULT;
     }
-    if (configData->int_pin == PIN_INVALID) {
-        configData->int_pin = HAL_PLATFORM_ETHERNET_WIZNETIF_INT_PIN_DEFAULT;
-    }
```
technobly, line 152: NOTE: We will have to update the docs to spell out the fact that if you want to set the INT pin back to it's default value, you must know what it is and set it explicitly.

See here: "If you want to restore one or more pins as their default value, use PIN_INVALID, and the default will be used for that pin."
https://docs.particle.io/reference/device-os/api/ethernet/pin-configuration-ethernet/

technobly (approved):

technobly:
> How does this work in safe mode when the app can't reconfigure the ethernet interface?

It is important to note that if a custom configuration is required, the correct system parts need to be flashed to the device so that it does not rely on booting into Safe Mode for the first connection to the Cloud.

This NOTE should also be added to the Docs.
https://docs.particle.io/reference/device-os/api/ethernet/pin-configuration-ethernet/

hal/network/lwip/wiznet/wiznetif_config.cpp
XuGuohui: Cc. @rickkas7

hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -327,7 +327,7 @@ void WizNetif::loop(void* arg) {
         if (p) {
             self->output(p);
         }
-        if (self->inRecv_) {
+        if (self->inRecv_ || self->interrupt_ == PIN_INVALID) {
```
avtolstoy, line 341: Is this working ok? We might need to lower the default timeout from current 500ms for this case I think. 500ms sounds like a lot.

avtolstoy (approved):

hal/network/lwip/wiznet/wiznetif.cpp
XuGuohui: This was tested to be working fine. So if the INT pin is not present, let's lower it to 250ms as the compromise.

PR #2777, opened 2024-05-29
[system] power: add method to specify the PMIC/FuelGauge interrupt pin.

XuGuohui:
**NOTE: this PR is targeting the `feature/auxiliary-power-control` branch**

### Problem
1. Currently the pin for PMIC interrupt is pre-defined per platform in DVOS and customer has to designed their own board according to the pin number for the PMIC interrupt.

2. We have allocated the A6 pin for the PMIC interrupt on the MSoM. For Muon, we'd like to make PMIC be capable of waking up device from hibernate mode, however the A6 pin doesn't not have this ability. Thus, for Muon we connect the PMIC interrupt pin to A7, which is capable of doing what we want. But DVOS doesn't has a method to change the default pin for PMIC interrupt.

### Solution
Introduce a new method `SystemPowerConfiguration::intPin(pin)` to specify another pin for PMIC/FuelGauge interrupt.

### Steps to Test

Hardware: SoM Eval board + MSoM

1. Flash the attached test app
2. Remove the `PM_INT` jumper on the SoM EVAL board
3. Fly wire the `A7` with the `PM_INT` pin
4. See if the system power manager behaves as expected

### Example App

```cpp
#include "application.h"
 
STARTUP(  
    System.setPowerConfiguration(SystemPowerConfiguration().feature(SystemPowerFeature::PMIC_DETECTION).intPin(A7));
);

SYSTEM_MODE(MANUAL);

void setup() {
}

void loop() {
}
```

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/wiring/inc/spark_wiring_system_power.h
+++ b/wiring/inc/spark_wiring_system_power.h
@@ -118,16 +118,16 @@ public:
         return conf_.aux_pwr_ctrl_pin;
     }
 
-    uint8_t auxPowerControlActiveLevel() const {
+    uint8_t auxiliaryPowerControlActiveLevel() const {
         return conf_.aux_pwr_ctrl_pin_level;
     }
     
-    SystemPowerConfiguration& intPin(uint8_t pin) {
+    SystemPowerConfiguration& interruptPin(uint8_t pin) {
         conf_.int_pin = pin;
         return *this;
     }
 
-    uint8_t intPin() const {
+    uint8_t interruptPin() const {
         return conf_.int_pin;
     }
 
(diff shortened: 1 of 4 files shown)


wiring/inc/spark_wiring_system_power.h
```diff
@@ -120,6 +121,15 @@ class SystemPowerConfiguration {
     uint8_t auxPowerControlActiveLevel() const {
         return conf_.aux_pwr_ctrl_pin_level;
     }
+    
+    SystemPowerConfiguration& intPin(uint8_t pin) {
```
avtolstoy: Nit: I'd use more verbose names here e.g. interruptPin() just for readability.

avtolstoy (approved):

PR #2782, opened 2024-06-06
[eg91e][eg91ex] enable 921600 baud rate

technobly:
### Problem

- EG91-E modems require A08 or above modem firmware version to support 921600 baud speeds, otherwise the default 460800 should be used.
- EG91-EX modems all support 921600 baud, but still using 460800.

### Solution

- Add a check for modem firmware version for EG91-E modems only at this time, and adjust the baud rate appropriately.
- Set EG91-EX to 921600 baud.

### Steps to Test

- With an EG91-E/EX modem, verify the appropriate baud rate is selected when using `APP=tinker-serial1-debugging` based on it's current firmware version.  You may manually set the result of `fwVersion_` to force a specific baud rate. [ @avtolstoy @Cheong2K would you please help verify this? ]
- Ensure no change with non EG91-E/EX modems [checked with B504]

### References

[sc-128144]

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -92,6 +92,7 @@ const auto QUECTEL_NCP_DEFAULT_SERIAL_BAUDRATE = 115200;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 460800;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_BG95_M5 = 921600;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_EG91_NAX = 921600;
+const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_EG91_EX = 921600; // version A08 or above
 
 const auto QUECTEL_NCP_MAX_MUXER_FRAME_SIZE = 1509;
 const auto QUECTEL_NCP_KEEPALIVE_PERIOD = 5000; // milliseconds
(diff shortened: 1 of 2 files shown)


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1073,10 +1074,30 @@ int QuectelNcpClient::getRuntimeBaudrate() {
...
     return runtimeBaudrate;
 }
 
+int QuectelNcpClient::getAppFirmwareVersion() {
+    // example output: EG91NAXGAR07A03M1G_30.004.30.004
+    // Right now we only care about EG91EFBR06AxxM4G*
+    auto resp = parser_.sendCommand("AT+QGMR");
+    int ver = 0;
+    int major = 0;
+    int n = CHECK_PARSER(resp.scanf("EG91EFBR06A%dM4G", &major));
```
scott-brust, line 1091: It looks like this firmware version `EG91EFBR06AxxM4G` is specific to the EG91-E? 

My EG91EX has this firmware version:
(I see your example EG91-NAX also has a different prefix)
```
0000012547 [ncp.at] TRACE: > AT+QGMR
0000012552 [ncp.at] TRACE: < EG91EXGAR08A14M1G_20.200.20.200
0000012556 [ncp.at] TRACE: < OK
0000012558 [ncp.client] INFO: QUERIED FW VERSION 0
```

So we will query for both EG91-E and EG91-EX modems, but the only firmware string that should match will be the EG91-E version. Is there an analogous EG91-EX firmware version we want to support 921600 baud on?

technobly: Ahh, good catch.  I was pretty sure I had tried this with an EX modem.  What SKU is that with the EX modem?  I'll make sure it also has the same requirement.

scott-brust: M524

technobly: I flashed the oldest firmware available for `EG91EXGAR08A03M1G_01.003.01.003` and it supports 921k baud, so we can limit the version check for EG91-E only, and force EG91-EX to 921k.  I updated the PR to reflect this.

avtolstoy (approved):

PR #2794, opened 2024-06-28
Makes the Ethernet reset pin optional.

XuGuohui:
**NOTE: this PR is targeting the `feature/auxiliary-power-control` branch**

### Problem
The Ethernet reset pin on Muon is controlled by a reset IC, instead of GPIO. The reset IC will delay up to 200ms after the Ethernet interface is powered up to reset the Ethernet, which is too late to add the Ethernet interface.

### Solution
1. Makes the reset pin of Ethernet optional, so that user can set the reset pin to `PIN_INVALID`
2. Does retries to detect the Ethernet interface so that we can detect it after the reset IC resets it.

### Steps to Test
Build and run the test app on MSoM/B5SoM/BSoM that is installed on the Muon.

### Example App
```cpp
#include "application.h"

SYSTEM_MODE(SEMI_AUTOMATIC);

uint8_t auxPwrPin = D7;

STARTUP({
    System.setPowerConfiguration(SystemPowerConfiguration().auxPowerControlPin(auxPwrPin).feature(SystemPowerFeature::PMIC_DETECTION));

    remap.cs_pin = A3;
    remap.reset_pin = PIN_INVALID;
    remap.int_pin = A4;
});

void setup() {
}

void loop() {
    if (Serial.available()) {
        char c = Serial.read();
        switch (c) {
            case 'i': {
                if_list* ifs = nullptr;
                if_get_list(&ifs);
                    if (iface->iface) {
                        uint8_t idx;
                    }
                }
                if_free_list(ifs);
                break;
            }
            case 'p': {
                Particle.connect();
                break;
            }
            case 'e': {
                Ethernet.connect();
                break;
            }
            default: break;
        }
    }
}
```

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/b5som/network/network.cpp
+++ b/hal/src/b5som/network/network.cpp
@@ -157,12 +157,16 @@ int if_init_platform_postpone(void*) {
         uint8_t dummy;
         if (!if_get_index(en2->interface(), &dummy)) {
             auto wizNetif = reinterpret_cast<WizNetif*>(en2);
-            if (wizNetif->isPresent()) {
-                return 0;
+            for (uint8_t retries = 0; retries < 5; retries++) {
+                if (wizNetif->isPresent()) {
+                    LOG(INFO, "W5500 present");
+                    return 0;
+                }
+                HAL_Delay_Milliseconds(50);
             }
+            LOG(INFO, "Remove Ethernet interface");
             netifapi_netif_remove(en2->interface());
         }
-        LOG(INFO, "Ethernet interface is not present");
         delete en2;
         en2 = nullptr;
     }
--- a/hal/src/b5som/network/network.cpp
+++ b/hal/src/b5som/network/network.cpp
@@ -157,7 +157,7 @@ int if_init_platform_postpone(void*) {
         uint8_t dummy;
         if (!if_get_index(en2->interface(), &dummy)) {
             auto wizNetif = reinterpret_cast<WizNetif*>(en2);
-            for (uint8_t retries = 0; retries < 5; retries++) {
+            for (uint8_t retries = 0; retries < 10; retries++) {
                 if (wizNetif->isPresent()) {
                     LOG(INFO, "W5500 present");
                     return 0;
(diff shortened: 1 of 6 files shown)


hal/src/b5som/network/network.cpp
```diff
@@ -157,12 +157,16 @@ int if_init_platform_postpone(void*) {
         uint8_t dummy;
         if (!if_get_index(en2->interface(), &dummy)) {
             auto wizNetif = reinterpret_cast<WizNetif*>(en2);
-            if (wizNetif->isPresent()) {
-                return 0;
+            for (uint8_t retries = 0; retries < 5; retries++) {
```
XuGuohui: 5 retries does not work reliably, we probably need to increase the retries.

hal/src/b5som/network/network.cpp
```diff
@@ -157,12 +157,16 @@ int if_init_platform_postpone(void*) {
         uint8_t dummy;
         if (!if_get_index(en2->interface(), &dummy)) {
             auto wizNetif = reinterpret_cast<WizNetif*>(en2);
-            if (wizNetif->isPresent()) {
-                return 0;
+            for (uint8_t retries = 0; retries < 10; retries++) {
+                if (wizNetif->isPresent()) {
```
avtolstoy: Maybe let's move this retry logic into `isPresent()` itself?

avtolstoy (approved):

avtolstoy (changes requested):
@XuGuohui If there is no reset pin configured we should perform a software reset in `hwReset()` instead using `MR_RST` MODE register bit.

avtolstoy:
Looks like there is a ready-to-use `wizchip_sw_reset()`

hal/src/b5som/network/network.cpp
XuGuohui: That will increase the boot time, since `isPresent()` is also invoked when creating the `Wiznetif`. If the auxiliary power is not enabled, those retries make no sense.

avtolstoy: In that case let's make this an optional parameter for `isPresent(bool retry = false)`

avtolstoy (approved):

PR #2796, opened 2024-07-05
Fix/ble stop scanning

avtolstoy:
### Problem

1. `BLE.stopScanning()` on Gen 4 is broken since 5.8.2 from inside ble scan result callback.
2. In general behavior of Gen 3/4 `scan()` and `stopScanning()` is currently slightly different. Also different with indefinite/timed scans.
3. No-fixture tests are missing for this.

### Solution

1. Fix this by making sure that only the scan caller semaphore is unblocked if `hal_ble_gap_stop_scanning()` is called from inside the ble event thread (the one ble results are being notified to the application from)
2. Streamline the behavior to be:
2.1. `BLE.scan()` is blocking no matter the timeout, but once scanning is started - BLE lock is not held, allowing other BLE operations to be performed.
2.2. `BLE.stopScanning()` works correctly when called both from inside the scan result callback/other BLE callbacks and from some other thread (be that application/system/timer/custom etc).
2.3. If called from a normal thread `BLE.stopScanning()` is a blocking operation and scanning MUST be stopped once its execution completes.
3. Add a set of tests to validate scan/stopScanning.

### Steps to Test

- wiring/no_fixture_ble
- wiring/ble_central_peripheral
- wiring/ble_scanner_broadcaster

For cross-platform testing with test runner: https://github.com/particle-iot/device-os-test-runner/pull/28

### Example App

```c
void setup() {
   */
}

void loop() {

}
```

### References

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/ble_hal.cpp
+++ b/hal/src/nRF52840/ble_hal.cpp
@@ -1418,8 +1418,11 @@ void BleObject::Observer::onScanGuardTimerExpired(os_timer_t timer) {
 }
 
 int BleObject::Observer::startScanning(hal_ble_on_scan_result_cb_t callback, void* context) {
+    BleLock lk;
+
     CHECK_FALSE(isScanning_, SYSTEM_ERROR_INVALID_STATE);
     SCOPE_GUARD ({
+        stopScanning();
         clearPendingResult();
     });
     ble_gap_scan_params_t bleGapScanParams = toPlatformScanParams();
@@ -1439,6 +1442,7 @@ int BleObject::Observer::startScanning(hal_ble_on_scan_result_cb_t callback, voi
             // We don't return here, as scanning may still timeout by Softdevice as expected.
         }
     }
+    lk.unlock();
     if (os_semaphore_take(scanSemaphore_, CONCURRENT_WAIT_FOREVER, false)) {
         SPARK_ASSERT(false);
         return SYSTEM_ERROR_TIMEOUT;
@@ -1447,9 +1451,12 @@ int BleObject::Observer::startScanning(hal_ble_on_scan_result_cb_t callback, voi
 }
 
 int BleObject::Observer::stopScanning() {
+    BleLock lk;
+
     if (!isScanning_) {
         return SYSTEM_ERROR_NONE;
     }
+
     if (os_timer_is_active(scanGuardTimer_, nullptr)) {
         os_timer_change(scanGuardTimer_, OS_TIMER_CHANGE_STOP, hal_interrupt_is_isr() ? true : false, 0, 0, nullptr);
     }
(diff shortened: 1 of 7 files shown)


hal/src/nRF52840/ble_hal.cpp
```diff
@@ -4133,7 +4132,6 @@ int hal_ble_gap_get_scan_parameters(hal_ble_scan_params_t* scan_params, void* re
 }
 
 int hal_ble_gap_start_scan(hal_ble_on_scan_result_cb_t callback, void* context, void* reserved) {
-    BleLock lk;
```
XuGuohui, line 4136: Need to remove the lock here for Gen4 as well?

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1783,8 +1784,6 @@ int BleObject::ConnectionsManager::connect(const hal_ble_conn_cfg_t* config, hal
     CHECK_TRUE(connections_.size() < BLE_MAX_LINK_COUNT, SYSTEM_ERROR_LIMIT_EXCEEDED);
     CHECK_TRUE(config, SYSTEM_ERROR_INVALID_ARGUMENT);
     CHECK_TRUE(connHandle, SYSTEM_ERROR_INVALID_ARGUMENT);
-    // Stop scanning first to give the scanning semaphore if possible.
-    CHECK(BleObject::getInstance().observer()->stopScanning());
```
XuGuohui, line 1787: It's inconsistent that we will still stop scanning when trying to connect on Gen4.

hal/src/nRF52840/ble_hal.cpp
```diff
@@ -1458,16 +1465,10 @@ int BleObject::Observer::stopScanning() {
     if (sd_ble_gap_scan_stop() != NRF_SUCCESS) {
         // LOG(ERROR, "Device is not in scanning state.");
     }
-    bool give = false;
-    ATOMIC_BLOCK() {
```
XuGuohui, line 1462: We might give the semaphore in BLE event thread on `BLE_GAP_EVT_TIMEOUT` event, followed by giving the semaphore herein again, it may result in simple return on next call to `startScanning()` when it tries taking the semaphore.

hal/src/nRF52840/ble_hal.cpp
avtolstoy: There is only a short iniailization check that is done under a lock in there. Gen 4 should already be doing the same thing with the lock: acquiring while starting the scan but releasing during it.

hal/src/nRF52840/ble_hal.cpp
avtolstoy: I think this should be fine as long as we make sure to guard `isScanning_` and further steps in `stopScanning()` under a BLE lock (same as it is right now on Gen 4). Is it safe to acquire it from inside the event handler for `BLE_GAP_EVT_TIMEOUT` ? You are more familiar with this.

XuGuohui: Currently the `BLE_GAP_EVT_TIMEOUT` event is handled in ISR directly, unlike other events those are push to the event queue. We can do that for the `BLE_GAP_EVT_TIMEOUT` event and acquire the lock in the event thread, but we probably run into deadlock. For instance,
```
Call a BLE API -> acquire lock -> waiting for event X to give semaphore -> release lock
                                  |-> handle event X in event thread -> waiting for the lock to releae
```
In that case we need to release the lock after the SDK API returns as same as in `startScanning()` for most of the BLE HAL APIs. This seams to be a significant change and require a ton of test work to make sure it won't introduce any regression. 
So the existing logic is right guarding the `isScanning_`, instead of using the lock. I would leave it as-is for now.

avtolstoy: Yeah, I double-checked with how it is right now, I think it should work fine as-is, so let's not overcomplicate this.

XuGuohui (approved):
1. no_fixture_ble passed
2. ble_central_peripheral passed

PR #2800, opened 2024-07-16
[rtl872x] fixes DMA enabled USART flush

technobly:
### Problem

- `Serial2.flush()` on `P2/Photon2` does not wait for bytes to be emptied from USART buffer on D4 (Serial2 TX), but instead when DMA transfer is complete.

### Solution

- Enable the IRQ used to signal TX buffer FIFO emptied for DMA enabled USARTs (Serial2 on P2/Photon2), along with a new `busy_` flag, to assist in waiting long enough for all bytes to be transmitted when busy waiting with `Serial2.flush()`
 
### Steps to Test

- Run integration tests `TEST=wiring/no_fixture`
- Run fixture test `TEST=wiring/serial_loopback2` (with TX jumpered to RX)
- Run example test app while watching the TX pin for each Serial1/2/3 (TX/D4/MOSI) interface and also the signalPin (A0) on Photon2.
- Run example test app with Serial1 on M-SoM ensuring the Cellular connection works properly.

### Example App

```c

#include "Particle.h"

SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);

SerialLogHandler logHandler(LOG_LEVEL_INFO);

// Try them all, signalPin should not go LOW before all bytes are flush()'d
// #define MY_SERIAL Serial1
#define MY_SERIAL Serial2
// #define MY_SERIAL Serial3

// P2/Photon2
// TX = Serial1 TX
// D4 = Serial2 TX
// MOSI = Serial3 TX

const pin_t signalPin = A0;
const int baudRate = 600;

auto TEST_PERIOD_MS = 5000;
auto lastTest = 0;

void runTest();

void setup() {
    pinMode(signalPin, OUTPUT);
    digitalWrite(signalPin, LOW);

    Particle.connect();
    waitFor(Particle.connected, 60000);

    MY_SERIAL.begin(baudRate);
}

void loop() {
    if (millis() - lastTest >= TEST_PERIOD_MS) {
        lastTest = millis();
        runTest();
    }
}

void runTest() {
    digitalWrite(signalPin, HIGH);
    int bufferSize = MY_SERIAL.availableForWrite();
    auto start = millis();

    for (int ii = 0; ii < 64; ii++) {
        char c = 0x20 + ii;
        MY_SERIAL.write(c);
    }
    auto writeComplete = millis();
    MY_SERIAL.flush();
    auto empty = millis();
    digitalWrite(signalPin, LOW);
    Log.info("bufferSize=%d writeComplete=%lu empty=%lu", bufferSize, writeComplete - start, empty - start);
}
```

### References

- https://community.particle.io/t/issues-with-serial2-in-a-photon2-p2-and-modbusclient-library/67177/4

---

### Completeness

### DIFF
--- a/hal/src/rtl872x/usart_hal.cpp
+++ b/hal/src/rtl872x/usart_hal.cpp
@@ -341,7 +341,7 @@ public:
     ssize_t flush() {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
         while (true) {
-            while (transmitting_) {
+            while (transmitting_ || busy_) {
                 // FIXME: busy loop
             }
             {
@@ -472,6 +472,7 @@ public:
         if (event & HAL_USART_PVT_EVENT_WRITABLE) {
             if (space() <= 0) {
                 TxLock lk(this);
+                xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
                 UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
                 // Temporarily disable TX DMA, otherwise, the TX FIFO won't be empty until all data is transferred by DMA.
                 UART_TXDMACmd(uartInstance, DISABLE);
@@ -574,7 +575,9 @@ public:
                 } else {
                     UART_TXDMACmd(uartInstance, ENABLE);
                     BaseType_t yield = pdFALSE;
-                    uart->busy_ = false;
+                    if (!uart->transmitting_) {
+                        uart->busy_ = false; // All bytes sent if no new DMA transfers are in progress
+                    }
                     if (xEventGroupSetBitsFromISR(uart->evGroup_, HAL_USART_PVT_EVENT_WRITABLE, &yield) != pdFAIL) {
                         portYIELD_FROM_ISR(yield);
                     }
@@ -843,7 +846,6 @@ private:
                 auto ptr = txBuffer_.consume(consumable);
                 for (size_t i = 0; i < consumable; i++, ptr++) {
                     UART_CharPut(uartInstance, *ptr);
-                    UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
                 }
                 curTxCount_ = consumable;
                 UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
--- a/hal/src/rtl872x/usart_hal.cpp
+++ b/hal/src/rtl872x/usart_hal.cpp
@@ -288,6 +288,7 @@ public:
         txBuffer_.reset();
         curTxCount_ = 0;
         transmitting_ = false;
+        busy_ = false;
         receiving_ = false;
         state_ = HAL_USART_STATE_DISABLED;
         return SYSTEM_ERROR_NONE;
@@ -300,6 +301,7 @@ public:
         dataInFlight(true /* commit */);
         CHECK(disable(false));
         transmitting_ = false;
+        busy_ = false;
         receiving_ = 0;
         rxBuffer_.prune();
         txBuffer_.reset();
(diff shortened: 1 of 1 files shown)


hal/src/rtl872x/usart_hal.cpp
```diff
@@ -344,7 +346,7 @@ class Usart {
             }
             {
                 AtomicBlock atomic(this);
-                if (!isEnabled() || txBuffer_.empty()) {
+                if ((!isEnabled() || txBuffer_.empty()) && !busy_) {
```
avtolstoy, line 349: This should probably be checked above along with `transmitting_`? `while (transmitting_ || busy_)`

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -831,12 +835,15 @@ class Usart {
...
+                busy_ = true;
+                UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
             } else {
                 // LOG UART doesn't support DMA transmission
                 consumable = std::min(MAX_UART_FIFO_SIZE, consumable);
                 auto ptr = txBuffer_.consume(consumable);
                 for (size_t i = 0; i < consumable; i++, ptr++) {
                     UART_CharPut(uartInstance, *ptr);
+                    UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
                 }
```
avtolstoy: Worst case scenario this might trigger for each single byte written here. The below case of enabling this interrupt after this is written is probably correct.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -572,6 +574,7 @@ class Usart {
                 } else {
                     UART_TXDMACmd(uartInstance, ENABLE);
                     BaseType_t yield = pdFALSE;
+                    uart->busy_ = false;
```
avtolstoy: Should probably check `transmitting_` here just in case to avoid a weird race condition between starting a new DMA transfer and finishing a previous one based on TX FIFO state.

avtolstoy: `enableEvent()` should also be patched probably to clear the tx bit before enabling the interrupt.
```c++
        if (event & HAL_USART_PVT_EVENT_WRITABLE) {
            if (space() <= 0) {
                TxLock lk(this);
                xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
                UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
                // Temporarily disable TX DMA, otherwise, the TX FIFO won't be empty until all data is transferred by DMA.
                UART_TXDMACmd(uartInstance, DISABLE);
                if (!transmitting_) {
                    startTransmission();
                }
            } else {
                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
            }
        }
```

hal/src/rtl872x/usart_hal.cpp
XuGuohui: Might try using `UART_WaitBusy()`, instead of introducing a new flag and configuring interrupt. This API is used in SDK exmaples.

XuGuohui (approved):

hal/src/rtl872x/usart_hal.cpp
XuGuohui: Yes, the interrupt might significantly slow down the performance for single byte per transmission.

hal/src/rtl872x/usart_hal.cpp
technobly: I was thinking the `while(true)` would keep it in the loop, but I suppose adding `while(transmitting_ || busy_){}` would prevent the atomic block from being set too frequently while `busy_ == true`.  I'll add that 👍🏼

hal/src/rtl872x/usart_hal.cpp
technobly: Hmm, actually I don't remember how that got there... I'll remove it from line 846.

avtolstoy (approved):

XuGuohui (approved):

PR #2802, opened 2024-07-18
Device protection fixes

sergeuz:
### Problem

The custom DfuSe command for resetting the device in safe mode (see https://github.com/particle-iot/device-os/pull/2770) doesn't work if the device is protected.

### Solution

Allow certain DfuSe commands when the device is protected.

This PR also adds another DfuSe command that allows re-enabling device protection without leaving DFU mode.

### Steps to Test

See https://github.com/particle-iot/particle-usb/pull/114.


avtolstoy (approved):

sergeuz:
Works as expected on NRF and RTL platforms.

PR #2804, opened 2024-07-23
Add support for event content types

sergeuz:
### Description

This PR extends the `Particle.publish()` and `Particle.subscribe()` APIs to support events with non-plain text data. The actual format of the data is determined by its content type:

```cpp
enum class ContentType {
    TEXT, // text/plain; charset=utf-8
    JPEG, // image/jpeg
    PNG, // image/png
    BINARY // application/octet-stream
};
```

This PR also includes the changes from [#2815](https://github.com/particle-iot/device-os/pull/2815).

### Example

Publishing binary data:
```cpp
char buf[] = { 0x00, 0x01, 0x02, 0x04 };
Particle.publish("my_event", buf, sizeof(buf), ContentType::BINARY);
```

Subscribing to binary events:
```cpp
void eventHandler(const char* name, const char* data, size_t size, ContentType type) {
    Log.info("Received event: %s; content type: %d; size: %u", name, (int)type, (unsigned)size);
    Log.dump(data, size);
    Log.print("\r\n");
}

void setup() {
    Particle.subscribe("my_event", eventHandler);
    // ...
}
```

When publishing/subscribing to events with a content type via the HTTP API, the event data is expected to be encoded as a [data URI](https://en.wikipedia.org/wiki/Data_URI_scheme).

### References

- [SDD](https://www.notion.so/particle/Typed-Publish-SDD-eb22480a35364c018c0154a7ab6e9262)


monkbroc (approved):
Looks great from my point of view. No additional feedback at this point.

sergeuz:
Rebased on `develop-6.x`.

avtolstoy (approved):

PR #2805, opened 2024-07-24
[Gen 4] WiFi stuff

avtolstoy:
### Description

1. Adds logging of wifi disconnect reason for Gen 4
2. Fixes incorrect recursive mutex usage in BLE HAL on Gen 4 platforms

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -437,17 +437,12 @@ public:
         CHECK_TRUE(cmdQueue_, SYSTEM_ERROR_INVALID_STATE);
         if (os_queue_put(cmdQueue_, &cmd, BLE_ENQUEUE_TIMEOUT_MS, nullptr)) {
             LOG(ERROR, "os_queue_put() failed.");
-            SPARK_ASSERT(false);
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1020,23 +1018,49 @@ void BleEventDispatcher::bleEventDispatchThread(void *context) {
...
+                    if (gap->bleCommandThreadExit_) {
+                        LOCAL_DEBUG("Exit ble cmd thread");
+                        break;
+                    }
                     gap->stopAdvertising();
                     LOCAL_DEBUG("notifyAdvStop in ble cmd thread");
                     if (command == BLE_CMD_STOP_ADV_NOTIFY) {
                         gap->notifyAdvStop();
                     }
+                    hal_ble_unlock(nullptr);
```
XuGuohui: This is not necessary since we have `SCOPE_GUARD()` above to unlock.

avtolstoy: Yup, forgot to remove after refactoring a bit. Fixed.

XuGuohui (approved):

PR #2806, opened 2024-07-24
Update the Tinker firmware to support configuring Muon.

XuGuohui:
### Problem
The factory default user app, i.e. the Tinker firmware, running on Particle M.2 SoMs that is installed on Muon may not work as expected, since some of the pins for controlling the power module and Etehrnet are not correct.

### Solution
Introduce two USB requests to
1. identify whether the base board is Muon or not
2. set correct control pins for PMIC and Ethernet.

### Steps to Test
1. Build and flash the tinker app to a Particle M.2 SoM
2. Update the `DVOS/user/tests/app/usb_ctrl_request/test/test.js` with the following content
3. Follow the `DVOS/user/tests/app/usb_ctrl_request/README.md` to run the test
```js
import * as control from './support/control';
import * as proto from './support/proto';
import * as usb from './support/usb';
import './support/init';

import * as semver from 'semver';
import * as randomstring from 'randomstring';

import { expect } from 'chai';

describe('USB requests', function() {
  let dev = null;

  before('find a Particle USB device', function() {
    return usb.enumDevices().then((devs) => {
      if (devs.length == 0) {
        throw new Error('No Particle USB devices found');
      }
      dev = devs[0]; // Use the first device in the list
    });
  });

  before('open device', function() {
    return dev.open();
  });

  after('close device', function() {
    if (dev) {
      return dev.close();
    }
  });

  describe('application control requests', function() {
    this.slow(2000); // Test application replies with a random delay

    it('identify base board should succeed', function() {
      const reqData = '{"BASE_BOARD_DETECT":null}';
      console.log(reqData);
      return control.sendRequest(dev, control.RequestType.APP_CUSTOM, reqData).should.become({
        result: control.Result.OK,
        data: '{"board":"muon"}'
      });
    });

    it('configure base board should succeed', function() {
        const reqData = '{"CONFIGURE_MODULE_BOARD":"muon"}';
        console.log(reqData);
        return control.sendRequest(dev, control.RequestType.APP_CUSTOM, reqData).should.become({
          result: control.Result.OK,
          data: '{"status":0}'
        });
      });
  });
});
```

### Example App
`make PLATFORM=msom all APP=tinker`

### References
N/A

---

### Completeness

- [ ] Added documentation


user/applications/tinker/src/board_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+void BoardConfig::configureBaseBoard(JSONValue value) {
+    detectI2cSlaves(false);
+    int ret;
+    if (isMuon_ && value.toString() == "muon") {
+        ret = configure(true);
+    } else if (!isMuon_ && value.toString() == "none") {
+        ret = configure(false);
+    } else {
+        ret = SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
```
monkbroc: Having the `isMuon` and `!isMuon` check will prevent resetting a module back to default values and will make end-to-end validation harder.

Here's the scenario I'm thinking of:
1. Take a new M-SoM out of the box and plug it into a Muon. At that moment the M-SoM is configured for "none".
2. Run through setup and check that the module was properly configured for "muon".
3. Before doing another test of setup, reset the M-SoM to "none" using a script. However you'll get invalid argument, so it's not possible to repeat steps 1 and steps 2 as if your device was a new M-SoM fresh out of the box.

monkbroc (commented):
Looks good to me with a comment about making it possible to change board configuration arbitrarily to make it easier to test end-to-end. I'll let folks on the Device OS team approve the PR.

user/applications/tinker/src/board_config.cpp
XuGuohui: Makes sense. This flag was used to avoid wrongly configuring those pins, but if the setup page can ensure that the correct JSON command is sent, then we're fine to remove the check.

user/applications/tinker/src/board_config.cpp
```diff
@@ -0,0 +1,159 @@
...
+        replyWriter_.name("board").value("muon");
+    } else {
+        replyWriter_.name("board").value("none");
+    }
+    replyWriter_.endObject();
+}
+
+void BoardConfig::configureBaseBoard(JSONValue value) {
+    detectI2cSlaves(false);
+    int ret;
```
avtolstoy: Nit: `int ret = SYSTEM_ERROR_INVALID_ARGUMENT` as default.

user/applications/tinker/src/board_config.cpp
avtolstoy: Let's remove this check, it's better not to restrict this call to set configuration to non-muon.

avtolstoy (approved):

avtolstoy (approved):

PR #2808, opened 2024-07-28
Add the HAL_PLATFORM_IF_INIT_POSTPONE flag.

XuGuohui:
### Problem
Gen3 platforms with Muon support are running out of flash space.

### Solution
Add the `HAL_PLATFORM_IF_INIT_POSTPONE` flag to exclude postponed network interface initialization for Gen3.

### Steps to Test
N/A

### Example App
N/A

### References
N/A

---

### Completeness

- [ ] Added documentation


hal/inc/hal_platform.h
```diff
@@ -174,6 +174,10 @@
 # endif /* HAL_PLATFORM_ETHERNET_WIZNETIF_INT_PIN_DEFAULT */
 #endif /* HAL_PLATFORM_ETHERNET */
 
+#ifndef HAL_PLATFORM_IF_INIT_POSTPONE
+#define HAL_PLATFORM_IF_INIT_POSTPONE (0)
```
avtolstoy, line 178: @XuGuohui Let's make this `(1)` by default and explicitly set `(0)` in `hal_platform_nrf52840_config.h`

avtolstoy (approved):

hal/inc/hal_platform.h
XuGuohui: In the case we need to set this to `(0)` in platform-specific config file for P2 and trackerm (non-SoM).

avtolstoy: We have flash space for it, so might as well enable?

XuGuohui: That's not ideal, since for P2 and TrackerM if the Wiznet chip is not present the Ethernet interface won't be removed, leading the `loop()` in Wiznetif is always executed with no meaning. The postpone initialization is maily for cases where the Power Module is used and the Ethernet is powered by auxiliary power.

avtolstoy: Okay, let's leave this as-is then :+1:

PR #2815, opened 2024-08-26
Encode/decode event data as a Variant

sergeuz:
### Description

- Promote the internal `Buffer` class to a public API.
- Extend `Variant` to support `Buffer` as one of its alternative types.
- Extend `Particle.publish()` to support publishing event data as a variant.
- Extend `Particle.subscribe()` to support a new callback type that takes event data as a variant.

`Buffer` is a general purpose class for storing arbitrary data in a dynamically allocated buffer. Note that one implication of supporting it as one of the `Variant`'s alternative types is that it makes it possible to store binary data in a ledger so this PR enables that as well.

When an event is published as a variant, it's serialized as CBOR on the wire. The server then converts it to JSON and forwards it to the subscribers. In the opposite direction, from the server to device, a conversion to CBOR is performed if necessary (see the examples below).

### Examples

Sending event data as a variant:

```cpp
Variant data;
data["a"] = 123;
data["b"] = 456;
Particle.publish("devout", data);
```

Subscribing via the CLI:

```
particle subscribe devout

{"name":"devout","data":"{\"a\":123,\"b\":456}",...}
```

Receiving event data as a variant:

```cpp
void eventHandler(const char* name, Variant data) {
    Log.info("Received event: %s", name);
    Log.print(data.toJSON());
    Log.print("\r\n");
}

void setup() {
    Particle.subscribe("devin", eventHandler);
}
```

When sending an event to the device, the content type of the event data can be arbitrary. On the wire, the data is converted to a CBOR which is then parsed by Device OS as a variant:

```
particle publish devin abc
particle publish devin 'data:application/json,[1,2,3,"abc"]'
particle publish devin 'data:application/cbor;base64,oWNhYmMYew=='
```

Device output:
```
0000368773 [app] INFO: Received event: devin
"abc"
0000377012 [app] INFO: Received event: devin
[1,2,3,"abc"]
0000398583 [app] INFO: Received event: devin
{"abc":123}
```


wiring/inc/spark_wiring_buffer.h
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * Get a pointer to the buffer data.
+     *
+     * @return Buffer data.
+     */
+    char* data() {
+        return d_.data();
+    }
```
sergeuz, line 66: The original `Buffer` implementation that we used internally exposed its data as `char*` so I did the same here to reduce changes in unrelated parts of Device OS that use this class. The Wiring library tends to use `uint8_t*` for binary data though so I'm wondering if it would be better to use that type instead.

user/tests/integration/communication/events/events.cpp
```diff
@@ -4,28 +4,37 @@
 namespace {
 
 String eventName;
-String eventData;
+Variant eventData;
```
monkbroc: Does it make sense to use EventData in the test to make sure the API doesn't break in the future?

user/tests/integration/communication/events/events.cpp
```diff
@@ -46,9 +55,16 @@ test(verify_max_event_data_size) {
...
-    assertTrue((bool)Particle.publish("my_event", (const char*)data, sizeof(data), ContentType::BINARY, WITH_ACK));
+test(publish_device_to_cloud_event_with_content_type) {
+    auto data = Buffer::fromHex("92b2d100238c490bd6e7a89050dbc5bbe53ff329b24cefadadd810d9c3a4f1b0e7748e7e2ecf48be4dd3ae08368f76a8d550ec139d5bca624e3c6b3cbc758565356c00afee12f2bd3ff22700dc4cc6fa02168ee5a1e6e9404a71d17da5a6b78d7d475ff2");
+    assertTrue((bool)Particle.publish("my_event", data.data(), data.size(), ContentType::BINARY, WITH_ACK));
+}
+
+test(publish_device_to_cloud_event_as_variant) {
+    Variant v;
```
monkbroc: Same here with publishing an EventData.

user/tests/wiring/api/cloud.cpp
```diff
@@ -168,6 +168,9 @@ test(api_spark_publish) {
 
     API_COMPILE(Particle.publish("event", String("data"), ContentType::TEXT));
     API_COMPILE(Particle.publish("event", String("data"), ContentType::TEXT, NO_ACK));
+
+    API_COMPILE(Particle.publish("event", Variant("data")));
+    API_COMPILE(Particle.publish("event", Variant("data"), NO_ACK));
```
monkbroc: Same for this file to make sure publish with EventData compiles.

monkbroc (approved):
Looks great!

avtolstoy (approved):

PR #2820, opened 2024-08-29
[tests] fixes for device-os-test-runner 0.11.0 changes

technobly:
- HIL job `prepare-binaries-platforms` pulls in the latest `device-os-test-runner`, and a tag change not present in `*.spec.js` was preventing tests from building.


avtolstoy (approved):
Supersedes https://github.com/particle-iot/device-os/pull/2814

PR #2822, opened 2024-08-29
[gen4] i2c: use transactions for fuelgauge and pmic on platforms with single i2c bus

avtolstoy:
### Description

On platforms with no dedicated i2c bus for fuelgauge and pmic, we should be making sure not to cut into application i2c transmissions. In order to do that we now use a transaction API which does not use I2C HAL rx/tx buffers and keeps things intact.

### Steps to Test

- wiring/no_fixture

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/msom/hal_platform_config.h
```diff
@@ -47,3 +47,6 @@
 #define HAL_PLATFORM_IF_INIT_POSTPONE (1)
 
 #define HAL_PLATFORM_AUTOMATIC_CONNECTION_MANAGEMENT (1)
+
+#define HAL_PLATFORM_HW_FORM_FACTOR_SOM (1)
```
XuGuohui, line 51: Is it never used for now?

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -477,13 +483,42 @@ class I2cClass {
...
+        if (waitStop && !WAIT_TIMED(transConfig.timeout_ms, !stopDetected())) {
             reset();
             return SYSTEM_ERROR_I2C_STOP_TIMEOUT;
         }
         return SYSTEM_ERROR_NONE;
     }
 
+    int transaction(const hal_i2c_transmission_config_t* txConfig, const hal_i2c_transmission_config_t* rxConfig) {
+        // A quick hack for now with a temp copy
+        particle::services::RingBuffer<uint8_t> txBuffer(txConfig ? txConfig->buffer : nullptr, txConfig ? txConfig->quantity : 0);
```
XuGuohui, line 495: We probably need to double check the `txConfig->buffer` and `rxConfig->buffer` to be non-nullptr.

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -477,13 +483,42 @@ class I2cClass {
...
+        particle::services::RingBuffer<uint8_t> rxBuffer(rxConfig ? rxConfig->buffer : nullptr, rxConfig ? rxConfig->quantity : 0);
+        if (txConfig) {
+            txBuffer.acquire(txConfig->quantity);
+            txBuffer.acquireCommit(txConfig->quantity);
```
XuGuohui, line 502: Why do we need this? Isn't done when constructing the `txBuffer`?

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -477,13 +483,42 @@ class I2cClass {
...
+        SCOPE_GUARD({
+            if (txConfig) {
+                std::swap(txBuffer, txBuffer_);
+            }
+            if (rxConfig) {
+                std::swap(rxBuffer, rxBuffer_);
+            }
+        });
+        if (txConfig) {
+            CHECK(endTransmission(!!(txConfig->flags & HAL_I2C_TRANSMISSION_FLAG_STOP), txConfig));
```
XuGuohui, line 520: Is `!!` typo?

wiring/src/spark_wiring_fuel.cpp
```diff
@@ -292,19 +293,44 @@ int FuelGauge::readRegister(byte startAddress, byte &MSB, byte &LSB) {
     CHECK_TRUE(i2c_.requestFrom(config) == 2, SYSTEM_ERROR_TIMEOUT);
     MSB = i2c_.read();
     LSB = i2c_.read();
-
+#else
+    config.quantity(1);
+    config.buffer(&startAddress);
+    config.stop(false); // Use repeated start
+    WireTransmission rxConfig(MAX17043_ADDRESS);
+    rxConfig.timeout(MAX17043_ADDRESS);
```
XuGuohui: `FUELGAUGE_DEFAULT_TIMEOUT` ?

XuGuohui (changes requested):

hal/src/msom/hal_platform_config.h
avtolstoy: Used in tests at the very least.

hal/src/rtl872x/i2c_hal.cpp
avtolstoy: Constructor only sets the buffer and size. This will mark as the buffer containing `quantity` bytes available to be consumed.

hal/src/rtl872x/i2c_hal.cpp
avtolstoy: flag bit to bool conversion.

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: What about using the `put(buffer, size)` directly?

XuGuohui (approved):

hal/src/rtl872x/i2c_hal.cpp
avtolstoy: We don't need to copy anything into the buffer, it already contains the data, we just need to move the head pointer. Apart from manually manipulating things, the easiest is to acquire + acquireCommit. The buffer may also very well be a constant from flash, so `put()` is definitely not the right way.

avtolstoy: Well, actually `put(nullptr, quantity)` will do the same, let's do that for clarity.

avtolstoy: Scratch that, there might be a bug in that particular use-case, tests fail. Sticking with acquire/acquireCommit.

PR #2825, opened 2024-09-03
[gen4] prebootloader: fix STOP/ULP sleep with disabled rtos, get correct wake-up reason

avtolstoy:
### Problem

IPC communications are currently broken with KM0 after 1 STOP/ULP sleep cycle when WiFi interface is not initialized and thus KM0 RTOS and WiFi firmware are not initialized.

### Solution

`taskENTER_CRITICAL()` and `taskEXIT_CRITICAL()` should only be used if RTOS has been started.

This PR also improves RTC wake-up reason detection.

### Steps to Test

Run the test app below with both `WiFi.on()` in `setup()` and without it. Wake-up reason should be correct, sleep time should be correct too in both cases through multiple sleep cycles.

### Example App

```c++
#include "application.h"

SerialLogHandler dbg1(LOG_LEVEL_ALL);
SYSTEM_THREAD(ENABLED);
SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {
    // IMPORTANT: Run with WiFi.on() below and without it, should work in either case correctly.
    // WiFi.on();
    delay(1000);
}

SystemSleepResult result;
system_tick_t start = 0, end = 0;

void loop() {
    delay(1000);
    Log.info("Start sleep test slept=%u result=%d", end - start, (int)result.wakeupReason());

    config.mode(SystemSleepMode::ULTRA_LOW_POWER)
        .gpio(A7, FALLING)
        .duration(30s);

    Log.info("going to ULP");
    delay(10);

    start = millis();
    result = System.sleep(config);
    end = millis();
}

```

### References

N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
+++ b/bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
@@ -301,6 +301,10 @@ void sleepInit(void) {
     km0_km4_ipc_on_request_received(KM0_KM4_IPC_CHANNEL_GENERIC, KM0_KM4_IPC_MSG_SLEEP, onSleepRequestReceived, nullptr);
 }
 
+// From rtl8721dlp_pmc.c
+extern "C" u32 WakeEventAon;
+extern "C" u32 WakeEvent;
+
 void sleepProcess(void) {
     // Handle sleep
     sleepResult[0] = SYSTEM_ERROR_NONE;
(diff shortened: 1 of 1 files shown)


bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
```diff
@@ -327,13 +331,22 @@ void sleepProcess(void) {
             auto config = sleepConfigShadow.config();
             if (config->mode == HAL_SLEEP_MODE_HIBERNATE) {
                 SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
-                taskENTER_CRITICAL();
+                if (xTaskGetSchedulerState() != taskSCHEDULER_NOT_STARTED) {
```
XuGuohui, line 334: Just wondering why this matters. Is it the root cause of the sleep duration being incorrect?

XuGuohui (approved):

bootloader/prebootloader/src/rtl872x/part1/sleep_handler.cpp
avtolstoy: In non-RTOS mode (when WiFi firmware is not initialized) this will disable interrupts and we'll no longer wake up KM0 on IPC interrupts. So 5 second 'sleep' was actually just a 5 second timeout waiting to get an IPC response.

PR #2839, opened 2024-10-17
[Gen4] ble: clear connecting flag on connected for peripheral role

XuGuohui:
### Problem
When a device is connected as BLE peripheral, it cannot connect to a BLE periperhal device.

### Solution
Reset the `connecting_` flag on connected as BLE peripheral.

### References
https://community.particle.io/t/ble-fails-to-connect-to-peripheral-as-central-after-connecting-as-a-peripheral-to-iphone/68477?u=ielec

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -2436,6 +2436,7 @@ void BleGap::handleConnectionStateChanged(uint8_t connHandle, T_GAP_CONN_STATE n
                 LOG_DEBUG(TRACE, "Connected as Central");
             } else {
                 connection.info.role = BLE_ROLE_PERIPHERAL;
+                connecting_ = false;
                 LOG_DEBUG(TRACE, "Connected as Peripheral");
             }
             connection.info.conn_handle = connHandle;


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2381,6 +2381,7 @@ void BleGap::handleConnectionStateChanged(uint8_t connHandle, T_GAP_CONN_STATE n
                 LOG_DEBUG(TRACE, "Connected as Central");
             } else {
                 connection.info.role = BLE_ROLE_PERIPHERAL;
+                connecting_ = false;
```
avtolstoy, line 2439: Should the check at https://github.com/particle-iot/device-os/pull/2839/files#diff-f291a584aa5be86f324f9fa672da24a8055116bbe0194745178ab5eb74ce25f7R2403 just handle both cases once we generate appropriate events etc?

XuGuohui: When the device is acting as BLE peripheral, the device won't initiate the ATT MTU updating process, thus, that it won't hit that check, unless the BLE central device initiates the process. Under some circumstances, the BLE central device may omit this process. But for Gen4 platform that acts as BLE central role, from what we observed, it always initiates the process.

avtolstoy (approved):

PR #2840, opened 2024-10-24
New cloud event API

sergeuz:
### Description

This PR introduces a [new API](https://github.com/particle-iot/device-os/blob/new-event-api/sc-131168/wiring/inc/spark_wiring_cloud_event.h) for publishing and subscribing to cloud events. The new API supports sending and receiving up to 16K of data in an event.

### Examples

Publishing a text event:

```cpp
CloudEvent event;
event.name("my_event");
event.data("abc");
Particle.publish(event);
// or
Particle.publish(CloudEvent().name("my_event").data("abc"));
```

Publishing a binary-encoded event:

```cpp
CloudEvent event;
event.name("my_event");
event.data("\x01\x02\x03\x04", 4);
event.contentType(ContentType::BINARY);
Particle.publish(event);
```

Using the event instance as a regular stream:

```cpp
CloudEvent event;
event.name("my_event");
for (int i = 0; i < 10; ++i) {
    event.write("abc", 3);
}
Particle.publish(event);
```

Polling the status of an event:

```cpp
CloudEvent event;
bool done = false;

void loop() {
    if (event.isNew()) {
        event = CloudEvent().name("my_event").data("abc");
        Particle.publish(event);
    }
    if (!done) {
        if (!event.isOk()) {
            // This catches any error that might have occurred while creating or sending the event
            Log.error("Failed to send event: %d", event.error());
            done = true;
        } else if (event.isSent()) {
            Log.info("Successfully sent event");
            done = true;
        }
    }
}
```

Using a callback to receive status updates for an event:

```cpp
CloudEvent event;

void loop() {
    if (event.isNew()) {
        event = CloudEvent().name("my_event").data("abc");
        event.onStatusChange([](auto event) {
            if (!event.isOk()) {
                Log.error("Failed to send event: %d", event.error());
            } else if (event.isSent()) {
                Log.info("Successfully sent event");
            }
        });
        Particle.publish(event);
    }
}
```

Subscribing to events:

```cpp
void eventHandler(CloudEvent event) {
    Log.info("Received event: %s", event.name());
    Log.info("Event data: %s", event.dataString().c_str());
    // or
    while (event.available() > 0) {
        Serial.print((char)event.read());
    }
}

void setup() {
    Particle.subscribe("my_event", eventHandler);
}
```


communication/src/v2/coap_payload.cpp
```diff
@@ -0,0 +1,215 @@
...
+        data += n;
+    }
+    if (bytesToWrite > 0) {
+        FsLock fs;
+        if (!file_) {
+            CHECK(createTempFile(fs.instance()));
+        }
+        assert(file_);
+        CHECK(seekInFile(fs.instance(), file_.get(), p - maxHeapSize_));
+        size_t n = CHECK_FS(lfs_file_write(fs.instance(), file_.get(), data, bytesToWrite));
```
monkbroc, line 155: todo: make sure that lfs_file_write can handle writing large buffers (>1000 bytes) in one shot or split the writes in here to the limit of lfs_file_write

wiring/inc/spark_wiring_cloud_event.h
```diff
@@ -0,0 +1,477 @@
...
+     * @return Event data.
+     */
+    String dataAsString() const;
+
+    /**
+     * Get the event data.
+     *
+     * @return Event data.
+     */
+    Variant dataAsVariant() /* FIXME: const */; // TODO: Rename?
```
monkbroc: Is this still relevant since we only support receiving binary payloads from large events?

wiring/inc/spark_wiring_cloud_event.h
```diff
@@ -0,0 +1,477 @@
...
+    bool valid() const {
+        return status() != Status::INVALID;
+    }
+
+    /**
+     * Check if the event is being sent to the Cloud.
+     *
+     * @return `true` if the event is being sent to the Cloud, otherwise `false`.
+     */
+    bool sending() {
```
monkbroc: There's no shorthand to see if a CloudEvent is in NEW state. Can you add a `new()` method?

wiring/inc/spark_wiring_cloud_event.h
```diff
@@ -0,0 +1,477 @@
...
+     * @return Event status.
+     */
+    Status status() const;
+
+    /**
+     * Check if the event is valid.
+     *
+     * @return `true` if the event is valid, otherwise `false`.
+     */
+    bool valid() const {
```
monkbroc: This one is odd since it checks if the state is not INVALID. Can you make this method `invalid()` so it's symmetrical with all the other state helpers?

wiring/inc/spark_wiring_cloud_event.h
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    int error() const;
+
+    /**
+     * Cancel sending the event.
+     *
+     * This method has no effect if the event is not currently being sent to the Cloud.
+     *
+     * A cancelled event cannot be published again.
```
monkbroc: Is this still true? What state does a cancelled event go to?

wiring/src/spark_wiring_cloud_event.cpp
```diff
@@ -0,0 +1,778 @@
...
+        }
+        return Error::INVALID_STATE;
+    }
+    auto payload = getValidPayload();
+    if (!payload) {
+        return error();
+    }
+    int r = coap_write_payload(payload, data, size, d_->pos, nullptr /* reserved */);
+    if (r < 0) {
+        if (r == Error::COAP_TOO_LARGE_PAYLOAD) {
```
monkbroc, line 669: Is there a way for an application to know how much room is left on the event (or what's the maximum total size of a large event) before calling write and getting COAP_TOO_LARGE_PAYLOAD?

wiring/src/spark_wiring_cloud_event.cpp
```diff
@@ -0,0 +1,778 @@
...
+        LOG(ERROR, "Event is being sent already");
+        return Error::INVALID_STATE;
+    }
+    if (!d_->name) {
+        LOG(ERROR, "Event name is missing");
+        return setFailed(Error::INVALID_STATE);
+    }
+    size_t size = this->size();
+    if (!RateLimiter::instance().take(size)) {
+        return Error::LIMIT_EXCEEDED;
```
monkbroc: So exceeding the amount of data in flight will return LIMIT_EXCEEDED, but the event will stay in the NEW state so it can be sent again later. Is there a way to query how much data is in flight to know when a new event can be sent?

monkbroc (commented):
Looks good. I mostly looked at the application API for large publishes.

communication/src/v2/coap_payload.cpp
sergeuz: @avtolstoy Is this still something on your team's radar or was it perhaps fixed already? It's that problem with writing to external flash when the source buffer is in internal flash, possibly at an unaligned address.

wiring/inc/spark_wiring_cloud_event.h
sergeuz: I mostly added it for feature parity with the existing API. The user can still send CBOR data in a binary event and use this method to parse it. Let me know if we don't want to support encoding and decoding variants in `CloudEvent` directly and I'll remove the related methods from it.

wiring/inc/spark_wiring_cloud_event.h
sergeuz: `new` is a reserved keyword so it will be `isNew()`. For consistency, I'll need to rename all the other shorthand methods to have an `is` prefix which I originally omitted for brevity.

wiring/inc/spark_wiring_cloud_event.h
sergeuz: I'd prefer to keep it this way so that it's consistent with other APIs unless you feel strongly about it. We only use `valid()` or `isValid()` everywhere else in Device OS.

wiring/inc/spark_wiring_cloud_event.h
sergeuz: Yeah, a cancelled event gets invalidated. Making it transition to a failed state instead is possible and would be more user friendly but it will require some additional work: https://github.com/particle-iot/device-os/blob/7c898958ce7da201e0cc91c0b08180c35335aed8/wiring/src/spark_wiring_cloud_event.cpp#L415-L417.

wiring/src/spark_wiring_cloud_event.cpp
sergeuz: The maximum supported size is exposed as `CloudEvent::MAX_SIZE`. The app can check how much space is left like this:
```cpp
if (event.pos() + 100 <= CloudEvent::MAX_SIZE) {
    event.write(data, 100);
}
```

wiring/src/spark_wiring_cloud_event.cpp
sergeuz: > the event will stay in the NEW state

That was an oversight. The event will transition to a failed state on this error as well.

> Is there a way to query how much data is in flight to know when a new event can be sent?

There's `canPublish()`:
```
if (CloudEvent::canPublish(3000)) {
    Particle.publish(CloudEvent().name("my_event").data(buf, 3000));
}
```

I was hesitant to define it as `Particle.canPublish()` as not to create confusion since we have different limits for the new and old APIs.

communication/src/v2/coap_payload.cpp
avtolstoy: I have a fix for this, should get a PR out today.

wiring/inc/spark_wiring_cloud_event.h
monkbroc: Or rename the initial state PENDING

wiring/inc/spark_wiring_cloud_event.h
monkbroc: ok, sounds good

wiring/inc/spark_wiring_cloud_event.h
monkbroc: OK. No need for additional work here.

wiring/inc/spark_wiring_cloud_event.h
sergeuz: I added `isNew` and renamed the other methods to `isSending`, `isOk`, etc. `PENDING` is good but may be confusing for events in progress. Also, incoming events are received in that state too.

communication/src/v2/coap_payload.cpp
avtolstoy: https://github.com/particle-iot/device-os/pull/2843

sergeuz: Thanks!

wiring/inc/spark_wiring_cloud_event.h
sergeuz: Fixed in https://github.com/particle-iot/device-os/pull/2840/commits/6b4d48953f2d8d06923d6e80680db5057faa4562.

avtolstoy (approved):

PR #2841, opened 2024-11-12
[Gen 4] Optimize QSPI flash writes and erasures

avtolstoy:
### Problem

1. We are unable to write anything over 8 bytes at a time to the QSPI flash.
2. QSPI flash clock speed and IO mode (2IO/4IO) are not correctly calibrated on boot
3. Erasures do not use 32K and 64K operations where possible

### Solution

1. Manually perform write transactions (in appropriate IO mode depending on flash type and hardware) and write up to 256 bytes at a time which is what flashes normally support
2. Change drive mode on QSPI flash pins and 'guess' correct dummy cycles of the flash in flash initialization
3. For erasures spanning multiple 4k blocks use optimized 32K and 64K operations, which are quicker. This should help OTA times.

### To validate

1. Sleep needs to be double checked including power measurements (could be leakage due to changed pin modes of QSPI pins)
2. OTP operations

### Steps to Test

**Make sure to upgrade prebootloader-part1 before system.**

- `wiring/no_fixture_stress` validates that the flash is in correct mode.

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -235,21 +544,16 @@ signed char XipControl::mpuEntry_ = -1;
...
-        b += rem;
-    }
-
-    return SYSTEM_ERROR_NONE;
+    return writeInBestMode(addr, (uint8_t*)data, size);
 }
 
 int hal_exflash_init(void) {
+    // ExFlashLock lk;
+    // FLASH_ClockSwitch(BIT_SHIFT_FLASH_CLK_XTAL, _FALSE);
```
scott-brust, line 556: We are doing this at the SDK level in `flash_calibration_backup()` right? 
It looks like this is called by prebootloader when it turns on the KM4. So the KM4 will boot with the higher flash speed, correct?
(Im guessing these lines are leftover from attempting to switch speeds once KM4 has powered on)

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -124,6 +134,212 @@ class RsipIfRequired {
...
+    RtkSpicFlashConfigGuard() {
+        ctrl0_ = SPIC->ctrlr0;
+        autoLen_ = SPIC->auto_length;
+        addrLen_ = SPIC->addr_length;
+        /* setting auto mode CS_H_WR_LEN, address length(3 byte) */
+        SPIC->auto_length = ((SPIC->auto_length & 0xfffcffff) | 0x00030000);
+        SPIC->addr_length = 3;
+    }
+    ~RtkSpicFlashConfigGuard() {
+        SPIC->ssienr = 0;
```
scott-brust, line 332: This disables the SPI controller once were done with the read/write/erase, right?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -124,6 +134,212 @@ class RsipIfRequired {
...
+        SPIC->ctrlr0 = ctrl0_;
+        SPIC->auto_length = autoLen_;
+        SPIC->addr_length = addrLen_;
```
scott-brust, line 335: These restore the `ctrlr0`, `auto_length` and `addr_length` registers to what they were before the read/write/erase?
Is there a concern that those will change with the flash operations? (I assume the auto increment address register should change?)

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -124,6 +134,212 @@ class RsipIfRequired {
...
+static void rtkSpiFlashTxCmdAddr(uint8_t cmd, uint32_t addr, uint32_t mode = RTK_SPIC_DATA_WIDTH_MODE_SINGLE) {
+    /* Disable SPI_FLASH */
+    SPIC->ssienr = 0;
+    /* set ctrlr0: TX mode, data_ch, addr_ch */
+    rtkSpiSetTxMode(mode);
+    /* set flash cmd + addr and write to fifo */
+    SPIC->dr[0].word = RTK_SPIC_CMD_ADDR_FORMAT(cmd, addr);
+}
+
+class RtkSpicFlashConfigGuard {
```
scott-brust, line 321: This config guard wasnt being used before, correct? 
And it should be in effect for all read/write/erase (and OTP) commands now?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -176,6 +392,99 @@ class ExFlashLock {
...
 };
 
+static int writeInBestModeWithinPageBoundaries(uint32_t addr, uint8_t* buf, size_t size) {
+    uint32_t address = addr;
+    uint8_t cmd = GENERIC_FLASH_CMD_PP;
+    uint32_t mode = RTK_SPIC_DATA_WIDTH_MODE_SINGLE;
+
+    if (flash_init_para.FLASH_cur_bitmode == SpicQuadBitMode) {
+        // Quad mode enabled
+        if (flash_init_para.FLASH_Id == FLASH_ID_GD) {
```
scott-brust, line 402: Where is `flash_init_para` populated? in KM0 on boot? Then copied to KM4 in `Set_System()`?
```
    _memcpy((void *)&flash_init_para, &flash_init_para_km0, sizeof(FLASH_InitTypeDef));
```
How are `FLASH_cur_bitmode` and `FLASH_Id` populated? They must be valid by the time exflash_hal is run right?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -176,6 +392,99 @@ class ExFlashLock {
...
+            cmd = GD_FLASH_CMD_4PP;
+            mode = RTK_SPIC_DATA_WIDTH_MODE_QUAD_I;
+        } else {
+            cmd = MXIC_FLASH_CMD_4PP;
+            mode = RTK_SPIC_DATA_WIDTH_MODE_QUAD_II;
```
scott-brust, line 407: Whats the difference between `RTK_SPIC_DATA_WIDTH_MODE_QUAD_I` and `RTK_SPIC_DATA_WIDTH_MODE_QUAD_II` ? 
Are these just differences between GD and MXIC protocols? These are both just saying the quad write command should have 3 byte addresses?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -176,6 +392,99 @@ class ExFlashLock {
...
+    size_t written = size;
+
+    while (size > 0) {
+        ExFlashLock lk(false);
+        size_t writeLen = 0;
+        {
+            rtkSpiFlashSendCmd(GENERIC_FLASH_CMD_WREN);
+            rtkSpiFlashTxCmdAddr(cmd, address, mode);
+            if (size > (RTK_SPIC_FIFO_HALF_SIZE - 4)) {
+                writeLen = RTK_SPIC_FIFO_HALF_SIZE - 4;
```
scott-brust, line 436: Is this an attempt to limit the number of bytes we write to the SPIC Data Register fifo? 
It should have 64, but we only want to use up to 32 at a time just to be conservative? 
There should be 1 byte of data, 3 bytes of address, then up to 28 bytes of data?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -176,6 +392,99 @@ class ExFlashLock {
...
+            } else {
+                writeLen = size;
+            }
+            uint32_t data;
+            uint32_t cnt = writeLen / 4;
+            for (uint32_t i = 0; i < cnt; i++) {
```
scott-brust, line 442: The logic is to write as much data in 4 byte word chunks first, then 2 byte half words, and then finally 1 byte to the SPIC data buffer. (This always writes to `SPIC->dr[0]` though?)

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -291,36 +595,61 @@ static bool is_block_erased(uintptr_t addr, size_t size) {
...
+    }
+    dcacheCleanInvalidateAligned(SPI_FLASH_BASE + start_addr, num_sectors * 4096);
+    return 0;
 }
 
 __attribute__((section(".ram.text"), noinline))
 int hal_exflash_erase_block(uintptr_t start_addr, size_t num_blocks) {
-    return erase_common(start_addr, num_blocks, EraseBlock);
+    for (size_t i = 0; i < num_blocks; i++) {
+        CHECK(erase_common(start_addr, GENERIC_FLASH_CMD_BE64K));
```
scott-brust, line 649: Do we actually do 64k block erases at some point in normal operation?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -291,36 +595,61 @@ static bool is_block_erased(uintptr_t addr, size_t size) {
...
 }
 
 __attribute__((section(".ram.text"), noinline))
 int hal_exflash_erase_sector(uintptr_t start_addr, size_t num_sectors) {
-    return erase_common(start_addr, num_sectors, EraseSector);
+    size_t sizeInBytes = num_sectors * 4096;
+    uintptr_t addr = start_addr;
+    while (sizeInBytes > 0) {
+        hal_exflash_cmd_t cmd = GENERIC_FLASH_CMD_SE;
+        if (addr % (64*1024) == 0 && sizeInBytes >= 64*1024) {
```
scott-brust, line 633: This checks if we are being asked to erase a 64k or 32k aligned series of sectors, and if so, coalesces them into a more efficient Block Erase command, right? 
Is there a concern that `start_addr` might not be sector aligned in the single sector erase case?

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -291,36 +595,61 @@ static bool is_block_erased(uintptr_t addr, size_t size) {
...
-    // LOG_DEBUG(ERROR, "Erased %lu %lukB blocks starting from %" PRIxPTR,
-    //           num_blocks, block_length / 1024, start_addr);
-
-    dcacheCleanInvalidateAligned(SPI_FLASH_BASE + start_addr, block_length * num_blocks);
-
-    return SYSTEM_ERROR_NONE;
+    rtkSpiFlashSendCmd(GENERIC_FLASH_CMD_WREN);
+    /* Disable SPI_FLASH */
+    SPIC->ssienr = 0;
+    rtkSpiSetTxMode();
```
scott-brust, line 619: This means that there is no data on the other IO lines? IE this is just a single line command + address transmission?

scott-brust (commented):
Still working through the PR, but wanted to submit questions so far.
Some TODOs still

- [x] - Finish reviewing some of the exflash_hal code changes
- [x] - Actually test the quad IO read/writes on my GD25 MSOM board, with QSPI traces (might not be most accurate but Id like to see all 4 IOs going)
- [ ] - Review + run the test cases

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Yes, this is done by prebootloader-part1 and once KM4 boots we should be at the appropriate clock speed and IO mode detected by prebootloader-part1.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Correct.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: As long as we are guarding only our own calls and do not use ROM functions, we should be good.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Only for those that we do manually through `SPIC` peripheral without any ROM/SDK stuff.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: It's all done in `rtl8721dlp_flashclk.c`, see the changeset https://github.com/particle-iot/realtek_ambd_sdk_public/compare/ede161210878fb0ee996f6edd0c4f81892109317...a6b17b249f9134ae5fe001d884558c88bf78a872#diff-798e71d08c3dcef6271de7277133461c1e34a9317546ece0b6702c722afbc0e7

and then yes, copied from KM0 to KM4. `FLASH_Id` is read from the flash (manufacturer id), bit mode is tested in `flash_rx_mode_switch()`

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: `QUAD_II` is both 24-bit address and data sent over 4 IO lines, `QUAD_I` is 24-bit address over IO0 and data over 4 IO lines. Just slightly different modes. Just depends on what the flash supports.

image
image

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: I don't know where the requirement of half FIFO comes from, I'll admit. Not documented well. But yes, this fits command + address + data up to half FIFO.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Yup.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Not really, no. But now we will be using 32K and 64K block erasures during at least OTA as `hal_exflash_erase_sector` optimizes this.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Yes, exactly. No concern as you are supposed to give a sector address, if it happens to be an address within the sector not at the start, it should still work.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: Yup
image

hal/src/rtl872x/exflash_hal.cpp
scott-brust: ok, so those commands will be tested implicitly during normal OTA testing 👍

hal/src/rtl872x/exflash_hal.cpp
scott-brust: Yes, I saw in the data sheet the sector erase command will erase the entire sector of the given address, even if it is not exactly at the start. 
I can see `hal_storage_erase()` calls `hal_exflash_erase_sector()`. Is there a concern that a user would call `hal_storage_erase()` with a non-sector aligned address and arbitrary size, causing a sector erase to erase data before the address desired?

avtolstoy: > causing a sector erase to erase data before the address desired

That is the intended behavior even in this case. And more importantly storage HAL is an unstable/experimental API, not for general use.

scott-brust: Ok, just wanted to make sure 👍

scott-brust:
Tested this with a probe-able QSPI chip on a non form factor MSOM and can see the 2io -> 4io switch when prebootloader-part1 turns on the KM4 👍 
SPI-QSPI-transition
I have salea capture if interested as well

scott-brust (approved):

PR #2842, opened 2024-11-26
Allow Device OS to initialize BG95-S5 modems

scott-brust:
### Problem

Device OS fails to init BG95-S5 modem due to failing `AT+CPIN?` commands. See discussion [here](https://s.slack.com/archives/C06K0107RE1/p1731541647363639?thread_ts=1731098248.864349&cid=C06K0107RE1)

### Solution
Skip querying CPIN for BG95-S5 modems. 
Note: At time of manufacture, the device will boot with no NCPID. So Quectel NCP client will be configured with `PLATFORM_NCP_UNKNOWN`. 
After the NCPID has been programmed, and the client is turned on, we need to query OTP for the NCPID in order to know that the unit has a BG95S5 and should skip the `AT+CPIN?` commands. 

### Steps to Test
Run DVOS + msom provisioning firmware

### Example App
Use provisioning firmware

### References

---

### Completeness

- [ ] Added documentation


hal/shared/cellular_ncp_dev_mapping.h
```diff
@@ -65,6 +65,15 @@ inline Dev cellular_dev_from_ncp(PlatformNCPIdentifier identifer) {
     case PLATFORM_NCP_QUECTEL_BG95_M5:
         device = DEV_QUECTEL_BG95_M5;
         break;
+    case PLATFORM_NCP_QUECTEL_EG800Q_NA:
```
technobly: Can we keep the EG800* stuff in separate PR, or do you need it to be included here?  You can skip over 20/21 for Dev, ..11/12 for PLATFORM_NCP_IDENTIFIER for now if you don't need it.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1135,8 +1135,12 @@ int QuectelNcpClient::initReady(ModemState state) {
         skipAll(serial_.get(), 1000);
         CHECK(waitAtResponse(10000));
 
-        // Select either internal or external SIM card slot depending on the configuration
-        CHECK(selectSimCard());
+        // Cannot use ncpId() as this is not programmed on first boot during manufacturing
+        // Read NCP ID from OTP when initializing the modem after provisioning
```
technobly, line 1139: Pretty sure ncpId() is the same as this though, since it's configured as `conf.ncpIdentifier(platform_primary_ncp_identifier());` in `hal/src/msom/network/network.cpp`

technobly (approved):
LGTM, please see my comments.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
scott-brust: This is a unique case where a brand new device is powered on for the first time. There will be nothing in OTP so `platform_primary_ncp_identifier()` will return `PLATFORM_NCP_UNKNOWN`. 
Once OTP is programmed, we want to be able to power on the modem to collect EID. However, `conf.ncpIdentifier` will still be `PLATFORM_NCP_UNKNOWN`. We want to explicitly check OTP for BG95_S5 in order to avoid the CPIN errors and be able to turn the modem on

hal/shared/cellular_ncp_dev_mapping.h
scott-brust: Not really, I just wanted to include all the new models to be complete as it appears there are already devices being provisioned with these NCP IDs

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: I see.  So we will be updating the NCPID after `conf` is already created, so we will have to read the NCPID again directly.  I wonder if the MFG script should reset the device after programming the NCPID to make sure all other ncpId() code gets the most up to date value?

scott-brust: It is not a script, it is a full running app. And we cannot reboot at this point because flash encryption has been enabled, but the encrypted MBR has not been programmed. The device will not boot if its reset at this point. The device must complete provisioning and report the details to the provisioning station so that the station can flash the encrypted MBR.

hal/shared/cellular_ncp_dev_mapping.h
scott-brust: Will remove as discussed

PR #2843, opened 2024-12-04
[gen3] fix internal -> external flash writes

avtolstoy:
### Problem

External flash writes must be performed from RAM and can't be done directly from internal flash due to DMA requirements.

### Solution

If data address is not in RAM, force `hal_flash_common_write` to perform writes through an intermediate 256 byte on-stack aligned buffer.

### Steps to Test

- wiring/no_fixture_stress

### Example App

N/A

### References

#2840 

---

### Completeness

- [ ] Added documentation


hal/src/nRF52840/flash_hal.c
```diff
@@ -105,7 +105,7 @@ int hal_flash_write(uintptr_t addr, const uint8_t* data_buf, size_t data_size)
     __flash_acquire();
 
     int ret = hal_flash_common_write(addr, data_buf, data_size,
-                                     &fstorage_perform_write, &hal_flash_common_dummy_read);
+                                     &fstorage_perform_write, &hal_flash_common_dummy_read, 0);
```
scott-brust, line 108: Im assuming whatever module in gen3 that uses `flash_hal.c` vs `flash_hal.cpp` wont encounter the issue that requires the temp buffer copy? 
(The root issue is when we are trying to write data that is already in external flash, to another location in external flash?)

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -564,7 +564,7 @@ int hal_exflash_uninit(void) {
 __attribute__((section(".ram.text"), noinline))
 int hal_exflash_write(uintptr_t addr, const uint8_t* data_buf, size_t data_size) {
     ExFlashLock lk;
-    CHECK(hal_flash_common_write(addr, data_buf, data_size, &perform_write, &hal_flash_common_dummy_read));
+    CHECK(hal_flash_common_write(addr, data_buf, data_size, &perform_write, &hal_flash_common_dummy_read, 0));
```
scott-brust, line 567: For gen4, the only way the intermediate buffer will be used is if there is an unaligned `data_buf`?

user/tests/wiring/no_fixture_stress/exflash.cpp
```diff
@@ -172,3 +205,17 @@ test(EXFLASH_02_rtl872x_validate_mode) {
     assertEqual((uint32_t)SPIC->baudr, (uint32_t)1);
 }
 #endif // HAL_PLATFORM_RTL872X
+
+test(EXFLASH_03_aligned_and_unaligned_writes_from_internal_flash_work) {
```
scott-brust, line 209: What part of this test is `from_internal_flash`? Does gen3 execute this code as XIP? 
`hal_storage_write()` is the external OTA flash section for both platforms right?

scott-brust (approved):

user/tests/wiring/no_fixture_stress/exflash.cpp
avtolstoy: `static uint8_t dummy[8192 + 4] __attribute((aligned(4))) = {0xff, 0xff, 0xff, 0xff};` is internal on gen 3, RAM on Gen 4, but the test is still kind of valid for gen 4.

avtolstoy: Oops, that needs to be `static const` now that you mention it.

hal/src/rtl872x/exflash_hal.cpp
avtolstoy: yes, same as before.

hal/src/nRF52840/flash_hal.c
avtolstoy: the issue is with internal flash to external flash. XIP is disabled, so we can't access external flash just through an address space normally. internal can come as an address in `data_buf`.

this is the check that fails: https://github.com/particle-iot/nrf5_sdk/blob/c3dbc4fa6b142a98062704e3ef68df5499200ae5/modules/nrfx/drivers/src/nrfx_qspi.c#L409

PR #2847, opened 2024-12-11
[network] postpone ethernet interface initialization to allow STARTUP() call to override pin configuration without a reset

avtolstoy:
### Description

As title says. Postpones Ethernet interface initialization until after power management is initialized for all the platforms (even without aux power support). This happens after global app constructors are called, so should allow `STARTUP()` to override pin configuration for W5500.

### Steps to Test

N/A

### Example App

N/A

### References

---

### Completeness

- [ ] Added documentation


hal/network/lwip/wiznet/wiznetif.h
```diff
@@ -83,10 +86,9 @@ class WizNetif : public BaseNetif {
 
 private:
     hal_spi_interface_t spi_;
-    hal_pin_t cs_;
-    hal_pin_t reset_;
-    hal_pin_t interrupt_;
-    bool postpone_;
+    hal_pin_t cs_ = PIN_INVALID;
+    hal_pin_t reset_ = PIN_INVALID;
+    hal_pin_t interrupt_ = PIN_INVALID;
```
technobly, line 91: Use defaults?
HAL_PLATFORM_ETHERNET_WIZNETIF_CS_PIN_DEFAULT
HAL_PLATFORM_ETHERNET_WIZNETIF_RESET_PIN_DEFAULT
HAL_PLATFORM_ETHERNET_WIZNETIF_INT_PIN_DEFAULT

hal/src/argon/network/network.cpp
```diff
@@ -173,6 +169,28 @@ int if_init_platform(void*) {
...
+        if (!if_get_index(en2->interface(), &dummy)) {
+            auto wizNetif = reinterpret_cast<WizNetif*>(en2);
+            WizNetifConfigData wizNetifConfigData;
+            wizNetifConfigData.size = sizeof(WizNetifConfigData);
+            wizNetifConfigData.version = WIZNETIF_CONFIG_DATA_VERSION;
+            WizNetifConfig::instance()->getConfigData(&wizNetifConfigData);
+            if (wizNetif->init(wizNetifConfigData) == 0) {
+                return 0;
+            }
+            LOG(INFO, "Remove Ethernet interface");
```
technobly, line 185: Should this be more like `LOG(ERROR, "Error Initializing Ethernet");`

I won't duplicate my comment for all of the rest of them ;-)

hal/src/argon/network/network.cpp
```diff
@@ -173,6 +169,28 @@ int if_init_platform(void*) {
     return 0;
 }
 
+#if HAL_PLATFORM_IF_INIT_POSTPONE
```
technobly, line 172: Should the call to `if_init_platform_postpone(nullptr);` in `main.cpp` be wrapped in `#if HAL_PLATFORM_IF_INIT_POSTPONE ... #endif` as well?

technobly (approved):
LGTM, please also see comments 👍🏼

technobly:
I think this may also require a `documentation` change to our example on how to setup the Wiznet Pin Config, simplifying it.

hal/src/argon/network/network.cpp
avtolstoy: Probably, but there is always a default weak empty function anyway and it was like that before.

hal/src/argon/network/network.cpp
avtolstoy: Yeah, might be better to make it something like `not detected`

hal/network/lwip/wiznet/wiznetif.h
avtolstoy: The defaults are in the WizNetConfig. I've left things as much as possible to how it was before: settings coming from `network.cpp`. Could set to defaults here too.

hal/network/lwip/wiznet/wiznetif.cpp
```diff
@@ -318,7 +328,7 @@ bool WizNetif::isPresent(bool retry) {
         cv = getVERSIONR();
         /* VERSIONR always indicates the W5500 version as 0x04 */
         if (cv != 0x04 && retry) {
-            HAL_Delay_Milliseconds(50);
+            HAL_Delay_Milliseconds(10);
```
XuGuohui, line 331: @avtolstoy The delay is not enough to detect the W5500 on Muon. Cc. @artyLee

PR #2849, opened 2024-12-11
[gen4] wifi: use BSSID from scan result, always perform scans before connect

avtolstoy:
### Description

Default connect sequence in RTL872x SDK doesn't seem to work well with single SSID/multiple BSSIDs, apply the same workaround as for ESP32-based platforms to always perform a scan and use the best access point (BSSID) based on signal strength to connect to, instead of just blindly letting the SDK to connect by SSID name.

(I2C log change is unrelated, just fixing an annoying log line that can occur with flaky I2C buses or devices that like to NACK a lot)

### Steps to Test

### Example App

N/A

### References

N/A
---

### Completeness

- [ ] Added documentation


technobly (approved):
LGTM

PR #2851, opened 2024-12-30
[Gen 3] stack size fixes, __cxa_acquire/release/atexit refactoring

avtolstoy:
### Problem

Gen 3 platforms trigger a hardfault on boot with #2840 every other time.

### Solution

1. Increase main stack size for Gen 3 platforms from 2K to 4K
2. Add asserts to linker files to make sure that stack sizes match the expected values
3. atexit is not needed for our platforms, so it's now stubbed out freeing additional flash space of unneccassary destructor implementations
4. `__cxa_acquire`/`__cxa_release` do not account for thread-safety, the implementation has been refactored.
5. Panic handler should not log anything, as that may trigger a secondary fault making debugging more difficult and causing a lock-up instead of a reset

### Steps to Test

#2840 cloud test app

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


sergeuz (approved):
Looks good :+1:

PR #2853, opened 2025-01-17
[Gen 4] wifi: prevent skbuf overflow in rtw_opt_skbuf

avtolstoy:
### Problem

Incorrect `MAX_SKB_BUF_SIZE` is defined in SDK `rtw_opt_skbuf.c` (an override source which allows us to increase the default number of packet buffers). The defined value is lower than what is expected by the rest of the closed-source `lib_wlan.a` and due to this an out of bounds access is very easily triggered with large enough network packets (slightly over 1000 bytes).

This usually causes one of these issues:
1. WiFi connection drop
2. Deadlock

### Solution

Make sure that the numbers match, add assertions.

### Steps to Test

TODO: Andrey to add on-device test.

As a quick test one can just heavily ping the WiFi IP of the device with payload over 1100 bytes e.g. `ping -i 0.03 -s 1100 1.2.3.4`

### Example App

N/A

### References

N/A

---

### Completeness

- [ ] Added documentation


sergeuz (approved):
Great job! :+1:

PR #2864, opened 2025-02-18
[network] interface-aware DNS lookups

avtolstoy:
### Problem

Despite the fact that we store all interface-specific DNS servers (statically configured or over DHCP/PPP etc) the resolver only uses the first and then on failure subsequent servers in this table. There is no API to tell the resolver to use only interface X.

This may have undesired effect even on cloud connectivity. For example, in case of M SoM with eth, wifi and cellular available, but eth and wifi providing bogus DNS responses (app firewall or other reasons), while the connection can still be performed over cellular, it might fail because we are unable to correctly resolve cloud server address.

### Solution

1. Introduce an API to perform interface-aware lookups
2. Make sure that when initiating a cloud connection we perform lookups through the ACM-selected network interface
3. Make sure that ACM round-robins interfaces for DNS lookups in case of reachability failures. This is not ideal, but is a trade-off to not perform up to currently 3 blocking DNS lookups per each interface, as that can take a signficant amount of time.

### Steps to Test

TODO

### References

https://github.com/particle-iot/lwip/pull/17

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/network/lwip/netdb_hal.cpp
+++ b/hal/network/lwip/netdb_hal.cpp
@@ -26,6 +26,7 @@
 #include <lwip/sockets.h>
 #include <errno.h>
 #include <algorithm>
+#include "resolvapi.h"
 
 struct hostent* netdb_gethostbyname(const char *name) {
     return lwip_gethostbyname(name);
@@ -42,26 +43,39 @@ void netdb_freeaddrinfo(struct addrinfo* ai) {
 
 int netdb_getaddrinfo(const char* hostname, const char* servname,
                       const struct addrinfo* hints, struct addrinfo** res) {
-    /* Change the behavior when AF_UNSPEC is used */
-    if (hints && hints->ai_family == AF_UNSPEC) {
-        struct addrinfo h = *hints;
+    return netdb_getaddrinfo_ex(hostname, servname, hints, res, nullptr);
+}
 
-        /* First perform a lookup with AF_INET6 */
-        h.ai_family = AF_INET6;
-        int rinet6 = lwip_getaddrinfo(hostname, servname, &h, res);
-
-        /* Next perform a lookup with AF_INET */
-        h.ai_family = AF_INET;
-        /* FIXME: expects that there is either 1 or 0 results from the previous call */
-        int rinet = lwip_getaddrinfo(hostname, servname, &h, rinet6 == 0 && *res ? &((*res)->ai_next) : res);
-
-        if (rinet6 == 0 || rinet == 0) {
-            return 0;
-        }
-
-        return std::max(rinet, rinet6);
+int netdb_getaddrinfo_ex(const char* hostname, const char* servname,
+                         const struct addrinfo* hints, struct addrinfo** res, if_t iface) {
+    uint8_t ifaceIndex = 0;
+    if (iface) {
+        if_get_index(iface, &ifaceIndex);
     }
-    return lwip_getaddrinfo(hostname, servname, hints, res);
+    if (hints && hints->ai_family == AF_UNSPEC /* && hints->ai_flags & AI_ADDRCONFIG ?*/) {
+        // For now defaulting as if AI_ADDRCONFIG is always set in case of AF_UNSPEC, simplifies things
+        if_protocol_state state = {};
+        if_get_protocol_state(iface, &state, nullptr);
+        if (state.ipv6 != IF_PROTOCOL_STATE_UNCONFIGURED) {
+            struct addrinfo h = *hints;
+
+            /* First perform a lookup with AF_INET6 */
+            h.ai_family = AF_INET6;
+            int rinet6 = lwip_getaddrinfo_ex(hostname, servname, &h, res, ifaceIndex);
+
+            /* Next perform a lookup with AF_INET */
+            h.ai_family = AF_INET;
+            /* FIXME: expects that there is either 1 or 0 results from the previous call */
+            int rinet = lwip_getaddrinfo_ex(hostname, servname, &h, rinet6 == 0 && *res ? &((*res)->ai_next) : res, ifaceIndex);
+
+            if (rinet6 == 0 || rinet == 0) {
+                return 0;
+            }
+
+            return std::max(rinet, rinet6);
+        }
+    }
+    return lwip_getaddrinfo_ex(hostname, servname, hints, res, ifaceIndex);
 }
 
 int netdb_getnameinfo(const struct sockaddr* sa, socklen_t salen, char* host,
--- a/system/src/system_cloud_connection_posix.cpp
+++ b/system/src/system_cloud_connection_posix.cpp
@@ -47,7 +47,7 @@ const unsigned CLOUD_SOCKET_HALF_CLOSED_WAIT_TIMEOUT = 5000;
 
 } /* anonymous */
 
-int system_cloud_resolv_address(int protocol, const ServerAddress* address, sockaddr* saddrCache, addrinfo** info, CloudServerAddressType* type, bool useCachedAddrInfo) {
+int system_cloud_resolv_address(int protocol, const ServerAddress* address, sockaddr* saddrCache, addrinfo** info, CloudServerAddressType* type, bool useCachedAddrInfo, network_handle_t interface, bool flushDnsCache) {
     CHECK_TRUE(info, SYSTEM_ERROR_INVALID_ARGUMENT);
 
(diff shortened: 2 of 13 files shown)


hal/network/lwip/netdb_hal.cpp
```diff
@@ -42,26 +43,39 @@ void netdb_freeaddrinfo(struct addrinfo* ai) {
...
+                         const struct addrinfo* hints, struct addrinfo** res, if_t iface) {
+    uint8_t ifaceIndex = 0;
+    if (iface) {
+        if_get_index(iface, &ifaceIndex);
+    }
+    if (hints && hints->ai_family == AF_UNSPEC /* && hints->ai_flags & AI_ADDRCONFIG ?*/) {
+        // For now defaulting as if AI_ADDRCONFIG is always set in case of AF_UNSPEC, simplifies things
+        if_protocol_state state = {};
+        if_get_protocol_state(iface, &state, nullptr);
+        if (state.ipv6 != IF_PROTOCOL_STATE_UNCONFIGURED) {
```
scott-brust, line 59: Is this check to avoid calling `lwip_getaddrinfo_ex()` unless we expect it to succeed? IE only resolve the ipv6 address if the interface is configured for ipv6? 
Is it important to avoid these duplicate calls to `lwip_getaddrinfo_ex()` for time reasons?

user/applications/tinker/src/board_config.h
```diff
@@ -34,7 +34,7 @@ class BoardConfig {
     char* reply();
     size_t replySize();
     
-private:
+public:
```
scott-brust: Was this intended? Are we trying to expose these functions for use with muon or something?
```
    void detectBaseBoard();
    void configureBaseBoard(JSONValue value);
    bool detectI2cSlaves();
    int configure(bool muon);
```

third_party/lwip/lwip
scott-brust, line 1: [ill review lwip changes in its PR](https://github.com/particle-iot/lwip/pull/17#pullrequestreview-2625051383)

system/src/system_cloud_connection_posix.cpp
```diff
@@ -148,6 +157,21 @@ int system_cloud_connect(int protocol, const ServerAddress* address, sockaddr* s
...
+                }
+                break;
+            }
+            case AF_INET6: {
+                if (!network_ready(cloudInterface, NETWORK_READY_TYPE_IPV6, nullptr)) {
+                    continue;
+                }
+                break;
+            }
+        }
```
scott-brust, line 173: Is this just to be more efficient and not attempt to connect to an ipv6 address if the cloud interface is not configured for it?

system/src/system_connection_manager.cpp
```diff
@@ -559,6 +584,24 @@ int ConnectionTester::prepare(bool fullTest) {
...
+                break;
+            }
+            case AF_INET6: {
+                if (!network_ready(NETWORK_INTERFACE_ALL, NETWORK_READY_TYPE_IPV6, nullptr)) {
+                    LOG_DEBUG(WARN, "skipping resolved AF_INET6 address, as ipv6 networking is not ready");
+                    continue;
+                }
+                break;
+            }
+        }
```
scott-brust, line 603: Same comment as in `system_cloud_resolv_address()` above. This is just to be more correct / quicker if IPV4/6 is not configured for *any* interface? 
Should we check if the specific `iface` being tested is `network_ready()`? Or use `NETWORK_INTERFACE_ALL`?

wiring/src/spark_wiring_network.cpp
```diff
@@ -116,14 +116,21 @@ bool NetworkClass::isPreferred() {
...
 
-IPAddress NetworkClass::resolve(const char* name) {
+IPAddress NetworkClass::resolve(const char* name, bool flushCache) {
     IPAddress addr;
 #if HAL_USE_INET_HAL_POSIX
     struct addrinfo *ai = nullptr;
     struct addrinfo hints = {};
     hints.ai_flags = AI_ADDRCONFIG;
+    if (flushCache) {
+        hints.ai_flags |= AI_FLUSHCACHE;
```
scott-brust, line 126: `AI_FLUSHCACHE` is an existing flag defined in lwip already, correct? 

To be specific, the intent is to ignore any cached addresses and instead do a new DNS query on the hostname? Should any cached address for *just* that hostname also be removed?

system/src/system_connection_manager.cpp
```diff
@@ -539,12 +549,27 @@ int ConnectionTester::pollSockets(struct pollfd* pfds, int socketCount) {
...
-    CHECK(getCloudHostnameAndPort(&info, &type, !fullTest));
+    network_handle_t iface = NETWORK_INTERFACE_ALL;
+    if (lastTestFailed) {
+        for (int i = 0; i < metrics_.size(); i++) {
+            int idx = roundRobinInterfaceForDns_ % metrics_.size();
+            if (network_ready(metrics_[idx].interface, 0, nullptr)) {
+                iface = metrics_[idx].interface;
+                struct ifreq ifr = {};
+                if_index_to_name(iface, ifr.ifr_name);
+                LOG(TRACE, "Last rechability test failed, using interface %s for DNS queries explicitly", ifr.ifr_name);
```
scott-brust, line 565: nit: typo on `rechability`

system/src/system_connection_manager.cpp
```diff
@@ -539,12 +549,27 @@ int ConnectionTester::pollSockets(struct pollfd* pfds, int socketCount) {
...
+                roundRobinInterfaceForDns_++;
+                break;
+            }
+            roundRobinInterfaceForDns_++;
+        }
+    }
+    CHECK(getCloudHostnameAndPort(&info, &type, !fullTest, iface));
```
scott-brust, line 572: Will using the `iface` from the round robin selection above filter out all the other resolved addresses from the other interfaces `dns_table`s? 
Are we relying on there being multiple entries in `info` in order to implicitly associate a resolved address with a network interface? Do we only test one interface with this change?

system/src/system_cloud_connection_posix.cpp
avtolstoy: Yup.

user/applications/tinker/src/board_config.h
avtolstoy: Ah, just a test change. I'll remove it.

hal/network/lwip/netdb_hal.cpp
avtolstoy: Yup. If AF_UNSPEC is specified and we don't have IPv6 connectivity, no point in resolving both AF_INET and AF_INET6

system/src/system_connection_manager.cpp
avtolstoy: Cloud is not reachable over IPv6 anyway, so, what we have now should be fine. I just found from some logs that sometimes we do get AAAA results returned (in case the DNS server is providing bogus results due to app fw/captive portal etc) and this just helps avoid using them.

wiring/src/spark_wiring_network.cpp
avtolstoy: Not an existing one. This is a new feature. It's just for that hostname.

system/src/system_connection_manager.cpp
scott-brust: I guess in situations where `lastTestFailed` did not fail, we will still test all interfaces. IE `iface = NETWORK_INTERFACE_ALL`. So in effect, we will round robin test 1 interface at a time with its DNS until it succeeds. Then all subsequent tests will use all available interfaces

scott-brust (approved):

system/src/system_connection_manager.cpp
avtolstoy: Something like that, yes. It's just a trade-off to get out of this weird state:
> This may have undesired effect even on cloud connectivity. For example, in case of M SoM with eth, wifi and cellular available, but eth and wifi providing bogus DNS responses (app firewall or other reasons), while the connection can still be performed over cellular, it might fail because we are unable to correctly resolve cloud server address.

Again, ideally during the test we'd resolve X number of times per each iface with its set of servers, but it's a blocking call and might take a while, so as a trade off - normally we use NETWORK_INTERFACE_ALL - so the default prioritization in DNS server list, in case we see a complete reachability failure we'll round robin through all ifaces/their dns servers to try to correctly resolve, yes, this might take several iterations - but we'll recover. And cloud connection layer subsequently will use the same servers/interface to resolve before establishing connection.

PR #2867, opened 2025-02-20
[gen4] hal: get rid of unnecessary constant delay(1) in HAL_USB_USART_Send_Data()

avtolstoy:
### Problem

We are always unnecessarily delaying 1ms on each byte sent over USB serial.

Moreover this happens even if there is no host connected, as `_blocking = true` by default and `Serial::write()` is implemented as following and always goes down to `HAL_USB_USART_Send_Data()`:

```c++
size_t USBSerial::write(uint8_t byte)
{
  if (HAL_USB_USART_Available_Data_For_Write(_serial) > 0 || _blocking) {
    return std::max(0, (int)HAL_USB_USART_Send_Data(_serial, byte));
  }
  return 0;
}
```

### Solution

Get rid of it. In addition to that fixes the FIXME about thread priorities. Don't need to delay(1) if we are below USB ISR processing task.

### Steps to Test
N/A

### Example App

N/A

### References

- [SC133780]

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/usb_hal.cpp
+++ b/hal/src/rtl872x/usb_hal.cpp
@@ -24,6 +24,9 @@
 #include <mutex>
 #include "usb_settings.h"
 #include "usbd_hid.h"
+// FIXME: not ideal, directly using freertos task APIs
+#include "FreeRTOS.h"
+#include "task.h"
 
 using namespace particle::usbd;
 
@@ -220,16 +223,24 @@ int32_t HAL_USB_USART_Send_Data(HAL_USB_USART_Serial serial, uint8_t data) {
     if ((__get_PRIMASK() & 1) || (__get_BASEPRI() != 0)) {
         return -1;
     }
-    int32_t available = -1;
-    do {
-        available = HAL_USB_USART_Available_Data_For_Write(serial);
-        if (available > 0) {
-            break;
+    int32_t available = HAL_USB_USART_Available_Data_For_Write(serial);
+    if (available == 0) {
+        bool needToYield = false;
+        auto prio = uxTaskPriorityGet(nullptr);
+        if (prio >= rtl::RTL_USBD_ISR_PROCESSING_THREAD_PRIORITY) {
+            needToYield = true;
         }
-        // TODO: check thread priorities here and delay only conditionally?
-        HAL_Delay_Milliseconds(1);
-    } while (available < 1 && available != -1);
-    if (HAL_USB_USART_Is_Connected(serial) && available > 0) {
+        while (true) {
+            available = HAL_USB_USART_Available_Data_For_Write(serial);
+            if (available > 0 || available < 0) {
+                break;
+            }
+            if (needToYield) {
+                HAL_Delay_Milliseconds(1);
+            }
+        }
+    }
+    if (available > 0 && HAL_USB_USART_Is_Connected(serial)) {
         return getCdcClassDriver().write(&data, sizeof(data));
     }
     return -1;
--- a/hal/src/rtl872x/usbd_driver.h
+++ b/hal/src/rtl872x/usbd_driver.h
@@ -35,6 +35,11 @@ namespace particle { namespace usbd {
 namespace rtl {
 
 const size_t TEMP_BUFFER_SIZE = 512;
+#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER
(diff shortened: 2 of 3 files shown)


hal/src/rtl872x/usb_hal.cpp
```diff
@@ -220,16 +223,24 @@ int32_t HAL_USB_USART_Send_Data(HAL_USB_USART_Serial serial, uint8_t data) {
...
-    if (HAL_USB_USART_Is_Connected(serial) && available > 0) {
+        while (true) {
+            available = HAL_USB_USART_Available_Data_For_Write(serial);
+            if (available > 0 || available < 0) {
+                break;
+            }
+            if (needToYield) {
+                HAL_Delay_Milliseconds(1);
+            }
+        }
```
scott-brust, line 241: Just to be clear, the logic is as follows

- If the CDC driver has no room to write data, we wait for it to have room here
- If the driver has *any* room to write (`available >0`) or returns an error (`available < 0`) then break out of the loop and write any available data
- If the driver does not have room to write, and we need to yield (ie the thread calling this write function has higher priority than the USB ISR thread), then we yield so the ISR has a chance to send data and make room to write. 
- if the driver does not have room to write, and we are lower priority than the ISR thread, then we busy wait. The ISR task should preempt us and make room to write eventually.

hal/src/rtl872x/usb_hal.cpp
```diff
@@ -220,16 +223,24 @@ int32_t HAL_USB_USART_Send_Data(HAL_USB_USART_Serial serial, uint8_t data) {
...
+    }
+    if (available > 0 && HAL_USB_USART_Is_Connected(serial)) {
```
scott-brust, line 243: If the USB UART is never connected, we should always have data available, but we will never call `getCdcClassDriver().write()` so in effect the data is thrown away and the CDC write buffer is never used, correct?

hal/src/rtl872x/usbd_driver.h
```diff
@@ -116,7 +116,7 @@ class RtlUsbDriver : public DeviceDriver {
         .speed = USB_SPEED_HIGH_IN_FULL,
         .dma_enable = 0, // ?
         .self_powered = 1,
-        .isr_priority = RTL_USBD_ISR_PRIORITY,
+        .isr_priority = rtl::RTL_USBD_ISR_PROCESSING_THREAD_PRIORITY,
```
scott-brust, line 119: Did we just want to move `RTL_USBD_ISR_PROCESSING_THREAD_PRIORITY` into the `rtl` namespace to move it somewhere more appropriate? Are we worried about people using it to assign their own thread priorities?

scott-brust (approved):
Just some questions

hal/src/rtl872x/usb_hal.cpp
avtolstoy: > If the driver has any room to write (available >0) or returns an error (available < 0) then break out of the loop and write any available data
Write only if `available > 0`, otherwise we simply error out.

Other things are correct.

hal/src/rtl872x/usb_hal.cpp
avtolstoy: If not connected, yes, we are throwing away. That's the expected behavior.

hal/src/rtl872x/usbd_driver.h
avtolstoy: Just moved it out of the class, renamed to be more explicit of what it is (not ISR priority, but a task priority) and there was a somewhat appropriate namespace already with some other constant, so just put it in there.

hal/src/rtl872x/usb_hal.cpp
scott-brust: Is there any concern now that we are busy waiting? I guess to the user perspective its still blocking (busy loop vs delay previously). This only should affect the system if a user has a high priority thread doing this write (and/or they are constantly sending a lot of USB data)?

avtolstoy: Normally the TX buffer should clear up pretty quickly, so I wouldn't be too concerned and it's the exact same behavior as on other platforms since Gen 2 era. There are of course corner cases with host stopping to query for data ungracefully, in which case we should still get out of waiting state, but that might take a few seconds based on some internal timers we set up.

PR #2868, opened 2025-03-20
Fix MBR flash_copy due to incorrect status register WIP bit detection

scott-brust:
### Problem

After manual provisioning of MSOM devices using MBR from `develop`, I discovered that prebootloader-mbr failed to copy prebootloader-part1 images from pending update section into the bootable slot. This rendered the device unbootable.

The issue is that the read / write / verify step of the flash copy was failing. Data written was not read back correctly. 
Upon examination of the bus, I could see that the driver was polling the Status Register (05) between Page Program (02) commands and in some cases the Write In Progress(WIP) bit was set. However, the driver was not waiting to poll again until that bit was cleared. New Page Program commands were issued while WIP bit is set. 
This resulted in unreliable data being written to flash, and the subsequent validation failures.

I looked into the driver to determine why the status register was being incorrectly read. I logged all of the values that were recorded during a `prebootloader-part1` update, and could see the status register returning unexpected values, some of which dont make sense
```
(gdb) p /x g_StatusRegisterValuesMBR
$2 = {0xf9, 0x0, 0x1, 0x0, 0xa, 0x0, 0x9c}
      ^ sum where status = 0         ^ status = 6
```

WIP and WEL bits are 0 and 1, so values up to `4` are valid combinations.
Values over `4` dont make sense in the MBR driver. We are not using the block protect bits 2 and above.
I can see the most common value, is `6`, which appears to be `4` left shifted 1 bit. 

### Solution(s)
The desired solutions in order should be
1. Fix the KM0 SPIC driver - No luck so far on this
2. Workaround this behavior in KM0 side - Shifting the bit in KM0 side works
3. Revert KM0 to ROM flash functions - Not needed just yet

If I shift the status register values in the KM0 (ie mbr) code, then the status register WIP polling works and the MBR can read/write/validate and correctly boot part1 images updates.

However, the KM4 code does *not* have this erroneous behavior. It always reports the correct status register values.

I cannot determine what the difference is in the SPI Controller structure between KM0 and KM4 cores. They both use the same driver, and largely the same init structure.
I do not know if this is a hardware issue or not, but the workaround is to shift "impossible" status register values to presumed WIP+WEL bit values *only* on KM0 core. This should be a safe compromise

### Steps to Test
Testing this will require the flash encryption key for a MSOM/P2 device.
You will need to build the prebootloader-mbr from this branch, and encrypt it with the device key
Flash the encrypted MBR
Do 
```
device-os-flash --all-devices <path_to_built_binaries>
device-os-flash --all-devices --control <path to prebootloader-part1.bin>
```
It should succeed

### DIFF
--- a/hal/src/rtl872x/exflash_hal.cpp
+++ b/hal/src/rtl872x/exflash_hal.cpp
@@ -205,6 +205,7 @@ static void rtkSpiFlashWaitBusy() {
     }
 }
 
+#ifdef ARM_CPU_CORTEX_M33
 static void rtkSpiSetRxMode() {
     SPIC->ctrlr0 = (SPIC->ctrlr0 | 0x0300);
     SPIC->ctrlr0 = (SPIC->ctrlr0 & 0xfff0ffff);
@@ -242,6 +243,7 @@ static uint8_t rtkSpiFlashReadSR() {
     rtkSpiFlashWaitBusy();
     return rtkSpiFlashGetDataReg(0, RTK_SPIC_DATA_BYTE);
 }
+#endif // ARM_CPU_CORTEX_M33
 
 static void rtkSpiFlashEnableCsWriteErase() {
     /* Enable SPI_FLASH */
@@ -249,15 +251,13 @@ static void rtkSpiFlashEnableCsWriteErase() {
     rtkSpiFlashWaitBusy();
     /* Check flash is in write progress or not */
     for (;;) {
-        uint8_t status = rtkSpiFlashReadSR();
-
+        uint8_t status = 0;
 #ifdef ARM_CPU_CORTEX_M23
-        // HACK: SPIC on KM0 reports seemingly invalid status register bits. 
-        // 6 should not be a status in the context of MBR, assume it is a 1 bit left shifted 4, and correct it
-        if (status == 6) {
-            status = status >> 1;
-        }
-#endif
+        // WORKAROUND: SPIC on KM0 reports invalid status register bits. ROM function does not
+        FLASH_RxCmd(GENERIC_FLASH_CMD_RDSR, 1, &status);
+#else 
+        status = rtkSpiFlashReadSR();
+#endif // ARM_CPU_CORTEX_M23
 
         if (!(status & 0x1)) {
             break;
--- a/hal/src/rtl872x/exflash_hal.cpp
+++ b/hal/src/rtl872x/exflash_hal.cpp
@@ -244,20 +244,20 @@ static uint8_t rtkSpiFlashReadSR() {
 }
 
 static void rtkSpiFlashEnableCsWriteErase() {
-    bool isKM0 = ((SCB->CPUID & SCB_CPUID_PARTNO_Msk) >> SCB_CPUID_PARTNO_Pos) == 0xD20 ? true : false;
-
     /* Enable SPI_FLASH */
     SPIC->ssienr = 1;
     rtkSpiFlashWaitBusy();
     /* Check flash is in write progress or not */
     for (;;) {
         uint8_t status = rtkSpiFlashReadSR();
-    
+
+#ifdef ARM_CPU_CORTEX_M23
         // HACK: SPIC on KM0 reports seemingly invalid status register bits. 
         // 6 should not be a status in the context of MBR, assume it is a 1 bit left shifted 4, and correct it
-        if (isKM0 && status == 6) {
+        if (status == 6) {
             status = status >> 1;
         }
+#endif
 
         if (!(status & 0x1)) {
             break;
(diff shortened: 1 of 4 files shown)


hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -554,6 +558,10 @@ static int perform_write(uintptr_t addr, const uint8_t* data, size_t size) {
 int hal_exflash_init(void) {
     // ExFlashLock lk;
     // FLASH_ClockSwitch(BIT_SHIFT_FLASH_CLK_XTAL, _FALSE);
+    uint32_t CPUID = (SCB->CPUID & SCB_CPUID_PARTNO_Msk) >> SCB_CPUID_PARTNO_Pos;
+    if (CPUID == 0xD20 ) {
+        isKM0 = true;
+    }
```
scott-brust: I tried reading `SCB->CPUID` each time we considered shifting the Status register value, but things slowed down dramatically. Detecting the KM0 core or not should only need to be done once on boot

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -44,6 +44,8 @@ extern "C" {
 extern uintptr_t platform_system_part1_flash_start;
 extern uintptr_t platform_flash_end;
 
+static bool isKM0 = false;
```
scott-brust: If we keep this approach this needs to go somewhere more appropriate

scott-brust:
Examples of captures failing on 6.3.0:
Write data:
failure-last4-bytes-fail-write
Read it back, it fails to validate
failiure-last4-bytes-fail-read-validation

On this branch, the WIP bit is polled, for example after this sector erase command:
success-sector-erase-status-polled

mrlambchop (approved):

avtolstoy (approved):

hal/src/rtl872x/exflash_hal.cpp
```diff
@@ -251,7 +250,15 @@ static void rtkSpiFlashEnableCsWriteErase() {
     /* Check flash is in write progress or not */
     for (;;) {
         uint8_t status = rtkSpiFlashReadSR();
-        // FLASH_RxCmd(GENERIC_FLASH_CMD_RDSR, 1, &status);
+
+#ifdef ARM_CPU_CORTEX_M23
+        // HACK: SPIC on KM0 reports seemingly invalid status register bits. 
+        // 6 should not be a status in the context of MBR, assume it is a 1 bit left shifted 4, and correct it
+        if (status == 6) {
```
avtolstoy: @scott-brust Do you know if it's always shifted? I'm worried about other bits affecting this workaround.

avtolstoy: also does this happen both on MXIC and GD?

avtolstoy: I would also probably compare with `FLASH_RxCmd(GENERIC_FLASH_CMD_RDSR, 1, &status)`, maybe we have some issue with dummy cycles?

scott-brust: > @scott-brust Do you know if it's always shifted? I'm worried about other bits affecting this workaround.

It is not always shifted, I recorded all the status register values during a prebootloader part1 update and this is what I came up with
```
(gdb) p /x g_StatusRegisterValuesMBR
$1 = {0x6e, 0x0, 0x11, 0x0, 0x3f, 0x0, 0xd8}
```
IE Status register was 0 `6E` times, 4 `3F` times, and 6 `D8` times etc. So sometimes we do get the valid `4` response

> also does this happen both on MXIC and GD?

Great question, ill check my other board 

> I would also probably compare with FLASH_RxCmd(GENERIC_FLASH_CMD_RDSR, 1, &status), maybe we have some issue with dummy cycles?

I can try this. I was worried that this would potentially not work the same now that we are managing the SPIC directly ourselves now. I can see if it returns the same shifted values or not

scott-brust: The status register is unreliable on MXIC parts as well. The shifting workaround does not solve the issue, 6.3.0 MBR on MXIC parts is still bricked.

The ` FLASH_RxCmd(GENERIC_FLASH_CMD_RDSR, 1, &status)` ROM command *works reliably on both parts*. We should probably just stick to this in the MBR

scott-brust:
I tested both MXIC and GD devices with MBR built at `fix-mbr-flash-copy@4fc8d6073` (ie latest commit).
I did the following update paths using `device-os-flash`
```
fix-mbr-flash-copy-4fc8d6073 --> 6.2.0 (ie ROM based flash driver) --> fix-mbr-flash-copy-4fc8d6073 --> 6.3.0 (refactored flash driver)
```
and verified that nothing was bricked. 

I think this is good to merge

PR #2871, opened 2025-04-21
[gen4] core: fix PSRAM heap size while in safe mode

avtolstoy:
### Problem

While in safe mode Gen 4 devices only use 32KB of PSRAM for heap. This is not enough. 

### Solution

Set the minimum PSRAM heap size to 128KB and additionally while in safe mode expand the PSRAM heap to full PSRAM size.

This PR additionally fixes a couple of issues in DCT implementation which doesn't expect e.g. open() calls to fail which leads to RAM and/or flash corruption in case of unexpected issues like being low on heap.

---

### Completeness

- [ ] Added documentation


technobly (approved):
LGTM!

scott-brust (approved):

PR #2872, opened 2025-06-05
Panic vitals, Gen 4 free_worker bugfix

avtolstoy:
### Description

1. This PR adds new panic vitals and makes sure that vitals are uploaded after a panic even after a session resumption (which normally does not happen)
```
{
  "uptime": 10,
  "memory": {
    "used": 303216,
    "total": 2925776
  },
  "protected_mode": "Open",
  "version": 20072,
  "part": {
    "hash": 1277916473
  },
  "user": {
    "part": {
      "hash": 588532806
    }
  },
  "panic": {
    "code": 10,
    "pc": 33747552,
    "lr": 33835835,
    "assertion": 34595205
  }
}
```

2. Fixes a bug in Gen 4 `free_worker` when putting an address to be freed onto the queue
3. Updates Gen 4 Cortex-M33 FreeRTOS port to upstream version, as the current one is buggy due to missing memory barriers after BASEPRI modification. It's supposed to be a Cortex-M4 bug, but it's present on M33 too. This is causing crashes sometimes.

### References

https://github.com/particle-iot/realtek_ambd_sdk_public/pull/6

---

### Completeness

- [ ] Added documentation


build/module.mk
```diff
@@ -249,15 +249,15 @@ define build_C_file
...
+	$(VERBOSE)$(CCACHE) $(CC) $(CFLAGS) $(CONLYFLAGS) -D__FILE_NAME__=\"$(notdir $<)\" -c -o $@ $<
 	$(call echo,)
 endef
 
 define build_CPP_file
 	$(call echo,'Building cpp file: $<')
 	$(call echo,'Invoking: ARM GCC CPP Compiler')
 	$(VERBOSE)$(MKDIR) $(dir $@)
-	$(VERBOSE)$(CCACHE) $(CC) $(CFLAGS) $(CPPFLAGS) -c -o $@ $<
+	$(VERBOSE)$(CCACHE) $(CC) $(CFLAGS) $(CPPFLAGS) -D__FILE_NAME__=\"$(notdir $<)\" -c -o $@ $<
```
scott-brust, line 260: This is to strip the full path from the `__FILE_NAME__` string in order to save space in the compiled binary right?

hal/src/nRF52840/core_hal.c
```diff
@@ -68,7 +69,7 @@ static struct Last_Reset_Info {
     uint32_t data;
 } last_reset_info = { RESET_REASON_NONE, 0 };
 
-void HardFault_Handler( void ) __attribute__( ( naked ) );
+void CommonFault_Handler( void ) __attribute__( ( naked ) );
```
scott-brust: How is this referenced elsewhere? I cant find any references, am I missing how this is used somewhere?

hal/src/nRF52840/core_hal.c
```diff
@@ -135,26 +136,61 @@ __attribute__((externally_visible)) void prvGetRegistersFromStack( uint32_t *pul
...
+
+    /* Go to infinite loop when Hard Fault exception occurs */
+    while (1) {
+        ;
+    }
 }
 
 
-void HardFault_Handler(void) {
+void CommondFault_Handler(uint32_t code) {
```
scott-brust: nit: is this a typo? `CommondFault_Handler` vs `CommonFault_Handler`

hal/src/nRF52840/core_hal.c
```diff
@@ -163,52 +199,62 @@ void HardFault_Handler(void) {
 }
```
scott-brust, line 199: I see this matches the rtl872x implementation now. The change is putting `code` into `r1` so it can be passed to `prvGetRegistersFromStack`?

hal/src/nRF52840/core_hal.c
```diff
@@ -163,52 +199,62 @@ void HardFault_Handler(void) {
...
+    data.size = sizeof(data);
+    data.code = HardFault; // FIXME, this should have a dedicated fault code
+    data.text = __FUNCTION__;
+    data.pc = (uintptr_t)_pc;
+    data.extra_code = (uintptr_t)_id;
+    data.registers[0] = (uintptr_t)_info;
+    PANIC_EXT(&data);
     while(1) {
         ;
     }
```
scott-brust, line 212: I havent looked at the nrf sdk too much, but I assume these are hooks from the SDK to call our fault handler when their code hits exceptions? 

When that happens we want to log the same kind of `PanicData` as much as possible? 

`_info` doesnt appear to be the normal stacked registers, but we are using `data.registers[0]` to store the data? From what I can tell in the SDK, these are pointers to different error data structures? 
I guess we are not too interested in deciphering these, but might as well store the pointer since we have it

hal/src/nRF52840/core_hal.c
```diff
@@ -163,52 +199,62 @@ void HardFault_Handler(void) {
...
 }
 
 void app_error_handler(uint32_t error_code, uint32_t line_num, const uint8_t * p_file_name)
 {
-    PANIC(HardFault,"HardFault");
+    PanicData data = {};
+    data.size = sizeof(data);
+    data.code = HardFault; // FIXME, this should have a dedicated fault code
+    data.lr = (uintptr_t)__builtin_return_address(0);
+    data.pc = __get_PC();
```
scott-brust, line 235: Will `__get_PC()` get the PC of some instruction inside `app_error_handler()` rather than what caused the error? Is this just the best we can do to fill this data out?

services/inc/panic.h
```diff
@@ -39,15 +42,42 @@ typedef enum {
...
+
+int panic_get_last_panic_data(PanicData* panic, void* reserved);
+void panic_set_last_panic_data_handled(void* reserved);
+
+#define PANIC_COMPAT(_code, _text, ...) ({ \
+        PanicData _data = {}; \
+        _data.size = sizeof(_data); \
+        _data.code = _code; \
+        _data.text = _text; \
+        _data.lr = (uintptr_t)__builtin_return_address(0); /* XXX: __get_LR? */ \
```
scott-brust, line 77: Is the question here between which is more accurate between the built in `__builtin_return_address` and our own read of `lr` register? Does this come down to how the `PANIC_COMPAT` macro is expanded wherever it is used?

services/inc/panic.h
```diff
@@ -39,15 +42,42 @@ typedef enum {
...
+        _data.pc = __get_PC(); \
```
scott-brust, line 78: This works because when the `PANIC_COMPAT` macro is expanded in whatever context its used and will have the PC of that function and therefore we will know what function caused the panic?

services/inc/services_dynalib.h
```diff
@@ -55,7 +56,7 @@ DYNALIB_FN(10, services, LED_Fade, void(Led_TypeDef))
 DYNALIB_FN(11, services, Get_LED_Brightness, uint8_t(void))
 
 DYNALIB_FN(12, services, set_logger_output, void(debug_output_fn, LoggerOutputLevel)) // Deprecated
-DYNALIB_FN(13, services, panic_, void(ePanicCode, void*, void(*)(uint32_t)))
```
scott-brust, line 58: Does this break anything? only if someone is calling `panic_` directly, which they should not be doing?

services/src/panic.c
```diff
@@ -46,29 +55,48 @@ void panic_set_hook(const PanicHook panicHookFunction, void* reserved)
...
 {
     #if HAL_PLATFORM_CORE_ENTER_PANIC_MODE
         HAL_Core_Enter_Panic_Mode(NULL);
     #else
         HAL_disable_irq();
     #endif // HAL_PLATFORM_CORE_ENTER_PANIC_MODE
 
+    memset(&g_retainedPanicData, 0, sizeof(g_retainedPanicData));
+    g_retainedPanicData.marker = PANIC_DATA_MARKER;
+    memcpy(&g_retainedPanicData.data, data, sizeof(PanicData));
```
scott-brust: kind of a nit: but do we want to only set the marker after the data has been copied? Swap the order of these two lines?

services/src/panic.c
```diff
@@ -46,29 +55,48 @@ void panic_set_hook(const PanicHook panicHookFunction, void* reserved)
...
     #if defined(RELEASE_BUILD) || PANIC_BUT_KEEP_CALM == 1
-        HAL_Core_System_Reset_Ex(RESET_REASON_PANIC, code, NULL);
+        HAL_Core_System_Reset_Ex(RESET_REASON_PANIC, data->code, NULL);
     #endif
 }
 
+int panic_get_last_panic_data(PanicData* panic, void* reserved) {
+    CHECK_TRUE(panic, SYSTEM_ERROR_INVALID_ARGUMENT);
+    CHECK_TRUE(g_retainedPanicData.marker == PANIC_DATA_MARKER && g_retainedPanicData.data.size > 0, SYSTEM_ERROR_NOT_FOUND);
+    memcpy(panic, &g_retainedPanicData.data, panic->size <= g_retainedPanicData.data.size ? panic->size : g_retainedPanicData.data.size);
```
scott-brust, line 92: `panic->size <= g_retainedPanicData.data.size ? panic->size : g_retainedPanicData.data.size`
Is the size check here to account for a growing `PanicData` structure? (not sure how that would work with a panic between firmware updates that change the structure)
Or just a sanity check?

scott-brust (commented):
Reviewed the panic handler stuff. My general notes about the call chains:

```
Non-fault panic calls:
PANIC() --> PANIC_COMPAT() --> panic_ext() --> LED flash + reset
PANIC_EXT()  ------------------^

Fault panic calls:
NVIC HW Fault Handlers -> Fault_Handler(panic_code) --> prvGetRegistersFromStack() --> PANIC_EXT()
```
Still need to review
- FreeRTOS changes

hal/src/nRF52840/core_hal.c
avtolstoy: Yup, already fixed.

hal/src/nRF52840/core_hal.c
avtolstoy: :+1: And having things consolidated in a single location pretty much.

hal/src/nRF52840/core_hal.c
avtolstoy: This is a fault that comes from SoftDevice, consider it more like an assertion. This is just our best attempt to gather all the info provided from SoftDevice (to be potentially accessed by vitals and `panic_get_last_panic_data`)

hal/src/nRF52840/core_hal.c
avtolstoy: Yup. As we are setting a code as HardFault here, the way to distinguish that this is `app_error_handler()` is a distinct PC. Hence the comment as well.

services/inc/panic.h
avtolstoy: Yes. I don't have a good answer which one is better. `__builtin_return_address` seems to be a bit more informative.

services/inc/panic.h
avtolstoy: Yes, exactly.

services/inc/services_dynalib.h
avtolstoy: No functional changes here, just types (which are of the same size as before). We are not using the last argument since Gen 2 era and the second argument is still a extra info string.

services/src/panic.c
avtolstoy: :+1: Good idea.

services/src/panic.c
avtolstoy: Just to account for future changes, yes. The changes between releases we probably do not care too much about. Same thing with any other retained structure (e.g. session data).

build/module.mk
avtolstoy: Yes. This is normally available in GCC 12+, but here we have to do it manually.

hal/src/nRF52840/core_hal.c
avtolstoy: It's a typo, should have uint32_t argument

hal/src/nRF52840/core_hal.c
```diff
@@ -163,52 +199,62 @@ void HardFault_Handler(void) {
...
 
 void app_error_fault_handler(uint32_t _id, uint32_t _pc, uint32_t _info) {
-    volatile uint32_t id = _id;
-    volatile uint32_t pc = _pc;
-    volatile uint32_t info = _info;
-    (void)id; (void)pc; (void)info;
-    PANIC(HardFault,"HardFault");
+    PanicData data = {};
+    data.size = sizeof(data);
+    data.code = HardFault; // FIXME, this should have a dedicated fault code
```
scott-brust, line 204: Is there any reason we cant add new values to `ePanicCode` enum? Is this definition maintained elsewhere like in device service and its a hassle to update there as well?

system/src/main.cpp
```diff
@@ -649,6 +653,61 @@ class RunTimeInfoDiagnosticData: public AbstractIntegerDiagnosticData {
...
+    }
+};
+
+ModuleShortHashDiagnosticData g_systemPartHashDiag(DIAG_ID_SYSTEM_SYSTEM_PART_HASH, DIAG_NAME_SYSTEM_SYSTEM_PART_HASH, &module_system_part1);
+UserModuleShortHashDiagnosticData g_userPartHashDiag(DIAG_ID_SYSTEM_USER_PART_HASH, DIAG_NAME_SYSTEM_USER_PART_HASH);
+SimpleUnsignedIntegerDiagnosticData g_panicCodeDiag(DIAG_ID_SYSTEM_PANIC_CODE, DIAG_NAME_SYSTEM_PANIC_CODE);
+SimpleUnsignedIntegerDiagnosticData g_panicPcDiag(DIAG_ID_SYSTEM_PANIC_PC, DIAG_NAME_SYSTEM_PANIC_PC);
+SimpleUnsignedIntegerDiagnosticData g_panicLrDiag(DIAG_ID_SYSTEM_PANIC_LR, DIAG_NAME_SYSTEM_PANIC_LR);
+SimpleUnsignedIntegerDiagnosticData g_panicAssertionStringDiag(DIAG_ID_SYSTEM_PANIC_ASSERTION_STRING, DIAG_NAME_SYSTEM_PANIC_ASSERTION_STRING);
+
```
scott-brust, line 713: Should this be a `SimpleUnsignedIntegerDiagnosticData` or some other diagnostic type? How does that work for string data? Is there a diagnostic type for strings?

third_party/nrf5_sdk/build.mk
```diff
@@ -11,7 +11,7 @@ TARGET_NRF5_SDK_NFC_SRC_PATH = $(TARGET_NRF5_SDK_PATH)/nrf5_sdk/components/nfc
 TARGET_NRF5_SDK_EXTERNAL_SRC_PATH = $(TARGET_NRF5_SDK_PATH)/nrf5_sdk/external
 
 # C source files included in this build.
-CSRC += $(call target_files,$(TARGET_NRF5_SDK_NRFX_SRC_PATH)/drivers/src/,*.c)
+CSRC += $(filter-out $(TARGET_NRF5_SDK_NRFX_SRC_PATH)/drivers/src/nrfx_i2s.c,$(call target_files,$(TARGET_NRF5_SDK_NRFX_SRC_PATH)/drivers/src/,*.c))
```
scott-brust, line 14: This excludes `nrfx_i2s.c` from the NRF SDK build? Assuming this is to save space since its unneeded?

hal/src/rtl872x/free_worker.h
```diff
@@ -82,7 +82,7 @@ int FreeRunnable::enqueue(void* ptr) {
 
     // Using FromISR variant here as it's not that safe to call normal FreeRTOS APIs
     // under a critical section/disabled interrupts.
-    int r = xQueueSendFromISR((QueueHandle_t)queue_, ptr, nullptr);
+    int r = xQueueSendFromISR((QueueHandle_t)queue_, &ptr, nullptr);
```
scott-brust, line 85: This was broken because we needed to pass a ***pointer*** to the item to be put into the queue (which is the actual pointer that should be free'd)? Not the pointer itself? 
Im assuming the other usages of `xQueueSendFromISR()` are correct in concurrent hal?

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2458,7 +2458,7 @@ void BleGap::handleConnectionStateChanged(uint8_t connHandle, T_GAP_CONN_STATE n
             if (connecting_ && connection.info.role == BLE_ROLE_CENTRAL) {
                 connecting_ = false;
                 // See: handleMtuUpdated()
-                // os_semaphore_give(connectSemaphore_, false);
+                os_semaphore_give(connectSemaphore_, false);
```
scott-brust: The gist of the BLE changes are that we give the `connectSempahore_` after connecting in the central role, rather than when the MTU is updated? How was this causing problems? We were giving the semaphore on MTU negotiation before connection was complete?

avtolstoy: This should not be here, covered by a different PR. This was a temporary fix. I'll get rid of it.

hal/src/rtl872x/free_worker.h
avtolstoy: Correct.

third_party/nrf5_sdk/build.mk
avtolstoy: Yup. `I2S_IRQHandler` was not being eliminated during linking and it takes quite a lot of flash space. This solves the issue.

system/src/main.cpp
avtolstoy: Vitals do not support string data. This just posts the address in the flash of the assertion string. Provided we have the elf, we should easily access this string.

hal/src/nRF52840/core_hal.c
avtolstoy: We can, but this requires DS changes and others, not a great time to deal with that. So for now HardFault.

hal/src/rtl872x/free_worker.h
scott-brust: This is a subtle one, good catch

scott-brust (approved):

PR #2873, opened 2025-06-05
BLE issues

avtolstoy:
### Description

This PR fixes two issues:
1. [Gen 3] We've introduced a regression in 5.9.0 in https://github.com/particle-iot/device-os/pull/2796 where BLE mutex in some cases (out of memory pool) may be acquired from inside an ISR handler (additionally we are slightly increasing by 1K the pool size)
2. [Gen 4] Devices in central role are hitting a connect semaphore acquisition timeout assertion with some peripheral that do not respond to MTU exchanges.

### Tests

Need to run BLE fixture tests (Gen4 - Gen4, Gen 4 - Gen 3) in both roles.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -1274,7 +1274,7 @@ int BleGap::stop(bool restore) {
                 os_queue_destroy(cmdQueue_, nullptr);
                 cmdQueue_ = nullptr;
             } else {
-                uint8_t command;
+                Command command = {};
                 while (!os_queue_take(cmdQueue_, &command, 0, nullptr)) {}
             }
         }
--- a/hal/src/rtl872x/ble_hal.cpp
+++ b/hal/src/rtl872x/ble_hal.cpp
@@ -108,6 +108,9 @@ void __wrap_bt_coex_handle_specific_evt(uint8_t* p, uint8_t len) {
 
 namespace {
 
+// Give MTU exchange 1.5 seconds to complete
+constexpr system_tick_t BLE_ATT_MTU_EXCHANGE_TIMEOUT_MS = 1500;
+
 // FIXME: should probably move under some class here
 StaticRecursiveMutex sBleEventMutex;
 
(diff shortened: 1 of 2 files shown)


hal/src/rtl872x/ble_hal.cpp
```diff
@@ -1090,7 +1104,9 @@ void BleGap::bleCommandThread(void *context) {
                     break;
                 }
                 gap->startAdvertising();
-            } else if (command == BLE_CMD_EXIT_THREAD) {
+            } else if (command.cmd == BLE_CMD_MTU_EXCHANGE_TIMEOUT) {
+                gap->handleMtuUpdated((uint8_t)(uintptr_t)command.arg /* connHandle */, 0);
```
scott-brust, line 1108: Why does `command.arg` need to be cast twice?

hal/src/rtl872x/ble_hal.cpp
```diff
@@ -2456,8 +2492,16 @@ void BleGap::handleConnectionStateChanged(uint8_t connHandle, T_GAP_CONN_STATE n
             evt.params.connected.info = &connection.info;
             notifyLinkEvent(evt);
             if (connecting_ && connection.info.role == BLE_ROLE_CENTRAL) {
-                connecting_ = false;
+                // Postpone setting connecting_ = false until MTU exchange happens (or times out)
+                // as by the standard you are not allowed to perform any reads/writes etc during an MTU exchange
+                SPARK_ASSERT(!os_timer_is_active(attMtuExchangeTimer_, nullptr));
+                os_timer_set_id(attMtuExchangeTimer_, (void*)(uintptr_t)connHandle);
```
scott-brust, line 2498: Is this a FreeRTOS feature to store data with a timer? I have not noticed this before

scott-brust (approved):
nit: in RTL `BleGap::stop()`, the command queue type is still uint8_t instead of `Command` struct
```
        if (cmdQueue_) {
            if (!cmdThread_) {
                os_queue_destroy(cmdQueue_, nullptr);
                cmdQueue_ = nullptr;
            } else {
                uint8_t command;
                while (!os_queue_take(cmdQueue_, &command, 0, nullptr)) {}
            }
        }
```
The items are just thrown away so it doesnt really matter but might help to keep types consistent.

scott-brust:
I still need to test actual hardware but so far it looks ok

hal/src/rtl872x/ble_hal.cpp
avtolstoy: If not, GCC will complain (error) about losing precision on void* -> uint8_t cast. void* -> uintptr_t (uint32_t) -> uint8_t avoids this.

hal/src/rtl872x/ble_hal.cpp
avtolstoy: Yup.

avtolstoy:
> nit: in RTL `BleGap::stop()`, the command queue type is still uint8_t instead of `Command` struct
> 
> ```
>         if (cmdQueue_) {
>             if (!cmdThread_) {
>                 os_queue_destroy(cmdQueue_, nullptr);
>                 cmdQueue_ = nullptr;
>             } else {
>                 uint8_t command;
>                 while (!os_queue_take(cmdQueue_, &command, 0, nullptr)) {}
>             }
>         }
> ```
> 
> The items are just thrown away so it doesnt really matter but might help to keep types consistent.

I'll fix this, otherwise it'll be a stack corruption :)

scott-brust:
[Did some more testing with gen3/gen4 HW](https://docs.google.com/spreadsheets/d/1k3bLHLgcur8iWNRqGGBRtoiSKwCsGxLYK2lm2nkUUJc/edit?gid=0#gid=0)

PR #2874, opened 2025-07-21
Increase system power manager stack size for all platforms

scott-brust:
### Problem
Using setup.particle.io on a B5SOM with Muon carrier board results in hard fault boot loop after flashing with 6.3.3 and enabling MUON support via USB control requests. Debugger showed that the system power manager thread was overflowing the stack. Gen3 platforms have a smaller stack than gen4

### Solution

Increase the stack size from 1024 to 1536 for all platforms

### Steps to Test
1. Verify that B5SOM / BSOM work on MUON carrier board with the USB control requests sent manually. They should just be the JSON payloads:
```
    {"BASE_BOARD_DETECT" : ''},
    {"CONFIGURE_MODULE_BOARD" : 'muon'},
```
2. Run SLO tests to make sure free RAM threshold isnt broken on Gen3 SOMs

### Example App

```c
void setup() {
   */
}

void loop() {

}
```

### References

[Internal Discussion](https://s.slack.com/archives/C04EP77AF/p1752698043664559)
---

### Completeness

- [ ] Added documentation


hal/src/b5som/hal_platform_config.h
```diff
@@ -40,3 +40,5 @@
 //#define HAL_PLATFORM_PPP_SERVER_USART_BAUDRATE (460800)
 #define HAL_PLATFORM_PPP_SERVER_USART_BAUDRATE (921600)
 #define HAL_PLATFORM_PPP_SERVER_USART_FLAGS (SERIAL_8N1 | SERIAL_FLOW_CONTROL_RTS_CTS) // For hw flow control (SERIAL_8N1 | SERIAL_FLOW_CONTROL_RTS_CTS)
+
+#define HAL_PLATFORM_POWER_MANAGEMENT_STACK_SIZE (1536)
```
circuscircle, line 44: Let's increase the default in `hal_platform.h` in general.

avtolstoy: Let's increase the default in `hal_platform.h` in general.

avtolstoy (approved):
The only thing we need to watch out for is the slo tests which test for free amount of RAM after boot. As long as that passes, we should be ok. If not - needs to be adjusted we don't have any other option.

avtolstoy (approved):

scott-brust:
SLO test passed
```
  Device startup service level objectives (SLOs)
    b5som
      systemThread=enabled
startupStats JSON { free_mem: 82432, app_flash_size: 17758 }
actual_free_mem=82432 target_free_mem=60000 platform=b5som
actual_app_flash_size=17758 target_app_flash_size=18105 platform=b5som
        ✓ slo startup stats (18.7s)
```

PR #2877, opened 2025-08-21
[test] fixes for M635 BG95-S5 CELLULAR_07_urcs

technobly:
### Problem

- CELLULAR_07_urcs failing on M635

### Solution

- Muxed AT command interface suspension not working the same way currently as other Quectel devices.  Use discrete AT commands instead to disable URCs.

### Steps to Test

- run `TEST=wiring/no_fixture_cellular`
- tested locally on B5SoM and M635 M-SoM

### Reference

- Ticket 212016

### DIFF
--- a/user/tests/wiring/no_fixture_cellular/cellular.cpp
+++ b/user/tests/wiring/no_fixture_cellular/cellular.cpp
@@ -189,8 +189,8 @@ test(CELLULAR_06_on_off_validity_check) {
 
     Cellular.off();
     waitFor(Cellular.isOff, MAX_CELLULAR_OFF_TIME);
-    assertFalse(Cellular.isOn());
     assertTrue(Cellular.isOff());
+    assertFalse(Cellular.isOn());
 
     ret = Cellular.command("AT\r\n");
     assertNotEqual(ret, (int)RESP_OK);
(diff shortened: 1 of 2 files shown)


user/tests/wiring/no_fixture_cellular/cellular.cpp
```diff
@@ -202,28 +202,60 @@ test(CELLULAR_06_on_off_validity_check) {
...
-    connect_to_cloud(HAL_PLATFORM_MAX_CLOUD_CONNECT_TIME);
 
 #if HAL_PLATFORM_GEN == 3
     assertEqual(Cellular.command("AT\r\n"), (int)RESP_OK);
+    int modem_type = -1;
+#if HAL_PLATFORM_NCP
+    modem_type = cellular_modem_type();
+#endif // HAL_PLATFORM_NCP
+
+    // Ensure we are disconnected, bad things can happen if messages are tx/rx
```
avtolstoy, line 216: Any specific issues you are seeing? It may interfere with vitals or system describe, but normally does work out fine.

technobly: It was actually some other command I was running after the AT interface was suspended, but it made me think that other things can happen at just the wrong time causing much longer testing delays due to the modem trying to power cycle at that test (if it was trying to send AT commands while suspended).  So the previous test ends with essentially power cycling the modem, so we really don't need to reconnect right away.  Best to leave cellular just powered on, and do the URC test... then connect to the cloud on the way out of that test.

avtolstoy (approved):

PR #2882, opened 2025-09-18
system_event network_credentials_added passes NetworkCredentials object

technobly:
### Problem

- When setting Wi-Fi network credentials over USB/BLE, NetworkCredentials object is not passed to system event network_credentials_added as it is when calling WiFi.setCredentials() directly.

### Solution

- Pass NetworkCredentials as required

### Steps to Test

- @jlovinger-particle to test via BLE setup process

### DIFF
--- a/system/src/control/wifi_new.cpp
+++ b/system/src/control/wifi_new.cpp
@@ -64,6 +64,49 @@ void bssidFromPb(MacAddress* bssid, const T& pbBssid) {
     }
 }
 
+void network_credentials_cleanup(void* ptr) {
+    auto creds = static_cast<NetworkCredentials*>(ptr);
+    free(creds);
+}
+
+NetworkCredentials* alloc_network_credentials_deep_copy(WifiNetworkConfig conf)
+{
+    const size_t ssid_bytes = conf.ssid() ? (strlen(conf.ssid()) + 1) : 0;
+    const size_t pass_bytes = conf.credentials().password() ? (strlen(conf.credentials().password()) + 1) : 0;
+    const size_t total = sizeof(NetworkCredentials) + ssid_bytes + pass_bytes;
+
+    unsigned char* mem = static_cast<unsigned char*>(malloc(total));
+    if (!mem) {
+        return nullptr;
+    }
+
+    auto creds = reinterpret_cast<NetworkCredentials*>(mem);
+    memset(creds, 0, sizeof(*creds));
+
+    unsigned char* p = mem + sizeof(NetworkCredentials);
+
+    if (ssid_bytes) {
+        char* ssid_buf = reinterpret_cast<char*>(p);
+        memcpy(ssid_buf, conf.ssid(), ssid_bytes);
+        creds->ssid = ssid_buf;
+        creds->ssid_len = static_cast<unsigned>(ssid_bytes - 1);
+        p += ssid_bytes;
+    }
+
+    if (pass_bytes) {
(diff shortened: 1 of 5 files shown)


system/src/control/wifi_new.cpp
```diff
@@ -64,6 +64,57 @@ void bssidFromPb(MacAddress* bssid, const T& pbBssid) {
...
+    const size_t ssid_bytes = conf.ssid() ? (strlen(conf.ssid()) + 1) : 0;
+    const size_t pass_bytes = conf.credentials().password() ? (strlen(conf.credentials().password()) + 1) : 0;
+    const size_t total = sizeof(NetworkCredentials) + ssid_bytes + pass_bytes;
+
+    unsigned char* mem = static_cast<unsigned char*>(malloc(total));
+    if (!mem) {
+        return nullptr;
+    }
+
+    auto* creds = reinterpret_cast<NetworkCredentials*>(mem);
```
avtolstoy: just `auto`, it is already a pointer

system/src/control/wifi_new.cpp
```diff
@@ -64,6 +64,57 @@ void bssidFromPb(MacAddress* bssid, const T& pbBssid) {
     }
 }
 
+void network_credentials_cleanup(void* ptr) {
+    auto* nc = static_cast<NetworkCredentials*>(ptr);
+    if (nc) {
+        if (nc->ssid && nc->ssid_len) {
+            memset(const_cast<char*>(nc->ssid), 0, nc->ssid_len);
```
avtolstoy: should be unnecessary, right? just freeing the single chunk should suffice

system/src/control/wifi_new.cpp
```diff
@@ -136,8 +187,18 @@ int joinNewNetwork(ctrl_request* req) {
     CHECK(ncpClient->on());
     // Set new configuration
     CHECK(wifiMgr->setNetworkConfig(conf, WifiNetworkConfigFlag::VALIDATE));
-    // TODO: Not adding NetworkCredentials for now as this object needs to be allocated on heap and then cleaned up
-    system_notify_event(network_credentials, network_credentials_added);
+    // Deep copy credentials and pass object to system_notify_event with a cleanup function
+    auto* creds = alloc_network_credentials_deep_copy(conf);
```
avtolstoy: just `auto`

system/src/control/wifi_new.cpp
```diff
@@ -307,7 +368,18 @@ int setNetworkCredentials(ctrl_request* req) {
     const auto wifiMgr = wifiNetworkManager();
     CHECK_TRUE(wifiMgr, SYSTEM_ERROR_UNKNOWN);
     CHECK(wifiMgr->setNetworkConfig(conf, WifiNetworkConfigFlag::NONE));
-    system_notify_event(network_credentials, network_credentials_added);
+    // Deep copy credentials and pass object to system_notify_event with a cleanup function
+    auto* creds = alloc_network_credentials_deep_copy(conf);
```
avtolstoy: just auto

avtolstoy (approved):

system/src/control/wifi_new.cpp
technobly: If we are going to set a callback to free the memory, I thought we might as well zero out the actual sensitive part as well.  But if you want to save these few bytes of memory, we can surely remove it.

PR #2886, opened 2025-11-21
Fixes MCP23S17 lock issue

XuGuohui:
### Problem
`Mcp23s17Lock lock();` in mcp23s17.cpp doesn't actually acuire the lock.

Affected platforms: tracker, trackerm

### Solution
Use `Mcp23s17Lock lock;` instead.

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/shared/mcp23s17.cpp
+++ b/hal/shared/mcp23s17.cpp
@@ -27,6 +27,7 @@
 #include "delay_hal.h"
 #include "spi_lock.h"
 #include <mutex>
+#include "scope_guard.h"
 
 using namespace particle;
 
@@ -58,6 +59,7 @@ Mcp23s17::Mcp23s17()
           ioExpanderWorkerThreadExit_(false),
           ioExpanderWorkerSemaphore_(nullptr),
           spiLock_(HAL_PLATFORM_MCP23S17_SPI, sDefaultSettings) {
+    begin();
 }
 
 Mcp23s17::~Mcp23s17() {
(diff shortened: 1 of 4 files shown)


hal/src/nRF52840/gpio_hal.cpp
```diff
@@ -170,6 +170,8 @@ int hal_gpio_configure(hal_pin_t pin, const hal_gpio_config_t* conf, void* reser
             hal_pin_set_function(pin, PF_NONE);
         }
 
+        CHECK(Mcp23s17::getInstance().begin());
```
avtolstoy: I understand the other one with `lock()` vs `lock` as far as construction goes. What's the issue with `begin()` in the constructor?

hal/shared/mcp23s17.cpp
```diff
@@ -58,15 +58,14 @@ Mcp23s17::Mcp23s17()
...
-    begin();
 }
 
 Mcp23s17::~Mcp23s17() {
 
 }
 
 int Mcp23s17::begin() {
-    Mcp23s17Lock lock();
+    Mcp23s17Lock lock;
```
avtolstoy: Once/if we switch over to gcc 11, we should really enable `-Wvexing-parse`. I don't think there is anything before 11 that can be done for cases like this unfortunately.

hal/src/nRF52840/gpio_hal.cpp
XuGuohui: First time calling`getInstance()` -> `begin()` -> `Mcp23s17Lock lock;` -> `Mcp23s17::getInstance().lock()`  -> call into `getInstance()` again causing reentrant issue. It didn't encounter the issue previously since `Mcp23s17::getInstance().lock()` wasn't really invoked.

avtolstoy: Ok, you are right. A simpler fix could be to just use lock()/unlock() specifically in begin() I guess?

XuGuohui: Yeah. It would be promised, in case that reading the IO expander in previous to `hal_gpio_configure()`. 👍

avtolstoy (approved):

PR #2888, opened 2025-11-28
Feature/early startup

XuGuohui:
**Note: This PR is targeting to the enhancement/gen4_boot_time branch**

### Problem
`STARTUP()` is not timing guaranteed. For example,
```cpp
class Test {
public:
    Test(uint8_t val = 0) {
        DiagPrintf("User App Constructor: %d\n", val);
    }
    ~Test() = default;
};

Test testInstance1(1);

STARTUP(
    DiagPrintf("STARTUP() 1\n");
);

STARTUP(
    DiagPrintf("STARTUP() 2\n");
);

Test testInstance2(2);
```
The output:
```
[+   2][253802946] User App Constructor: 1
[+   3][253802949] STARTUP() 1
[+   1][253802950] STARTUP() 2
[+   1][253802951] User App Constructor: 2
```
### Solution
Introduce a new API `PRE_STARTUP()`, which is executed in previous to the `STARTUP()` and is timing guaranteed.

### Example App
```cpp
void PRE_STARTUP() {
    Serial2.begin(115200);
    Serial2.println("PRE_STARTUP called");
}

void setup() {
}

void loop() {
}
```

### References
N/A

---

### Completeness

- [ ] Added documentation


technobly (approved):

system/src/main.cpp
```diff
@@ -850,8 +854,6 @@ void app_setup_and_loop(void)
     system_subscribe_event(aux_power_state, if_init_postpone, nullptr);
 #endif /* HAL_PLATFORM_LWIP */
 
-    system_power_management_init();
```
XuGuohui, line 853: This seams to be a breaking change. Calling system power manager APIs in `STARTUP()` will no more work. Might leave it as-is and also call `system_power_management_init();` before `PRE_STARTUP()`?

avtolstoy: ? I'm confused.

https://github.com/particle-iot/device-os/pull/2888/changes#diff-5920c626810f612b3105d4945621b64cd93f959b03e71a058fb19cfd1c6fc71bR839

XuGuohui: For example, if user wanted to disable system power manager in `STARTUP()`, it is too late to stop the system power manager thread. Calling `system_power_management_init()` twice won't fix the issue. We probbaly need to do something in `Event::ReloadConfig` to re-configure the whole system power manager according to the new config.

PR #2890, opened 2025-12-09
[fuelgauge] fixes access when i2c unintentionally disabled [sc-137389]

technobly:
### Problem

- Calls to `FuelGauge` in certain cases can result in incorrect readings if `FuelGauge::begin()` not called first.
- These cases are typically when using `hal_i2c_config_t acquireWireBuffer()` and setting a buffer size greater than 32.  This ends up re-initialing the I2C bus, and leaving it disabled.  Typically `begin()` would need to be called afterwards, but we do not document it as necessary.
- In addition to the i2c bus being disabled, the `FuelGauge` calls need to occur 5-10s after boot.  The reason for this is the `FuelGauge` i2c bus is shared with the `PMIC`, and if the `PMIC` is still going through it's initial configuration, it will detect the bus has been disabled by the `FuelGauge` call, and call `begin()` again.

### Solution

- Call `hal_i2c_begin()` automatically after `hal_i2c_init()` for `FuelGauge/PMIC`.

### Steps to Test

- Run `device-os-test run bsom wiring/no_fixture -v -- -fixture`

### References

- [sc-137389]

### DIFF
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -317,11 +317,19 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     os_thread_scheduling(false, nullptr);
     if (i2cMap[i2c].mutex == nullptr) {
         os_mutex_recursive_create(&i2cMap[i2c].mutex);
-    } 
+    }
 
     // Re-enable threading and capture the mutex
     os_thread_scheduling(true, nullptr);
     I2cLock lk(i2c);
+
+// sc-137389
+#if HAL_PLATFORM_PMIC_BQ24195
+    bool wasEnabled = false;
+    if (i2cMap[i2c].state == HAL_I2C_STATE_ENABLED) {
+        wasEnabled = true;
+    }
+#endif // HAL_PLATFORM_PMIC_BQ24195
 
     if (i2cMap[i2c].configured) {
         // Configured, but new buffers are invalid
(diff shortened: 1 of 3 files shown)


wiring/src/spark_wiring_fuel.cpp
```diff
@@ -276,12 +276,18 @@ int FuelGauge::wakeup() {
 
 
 int FuelGauge::readConfigRegister(byte &MSB, byte &LSB) {
+    if (!i2c_.isEnabled()) {  // sc-137389
```
avtolstoy: @technobly Can we fix this in hal_i2c_init() and make this works sanely in all cases? We already check whether we are configured in both Gen 3 and Gen 4 HALs, so the simple thing would be to at the end of hal_i2c_init restore the state to configured essentially calling hal_i2c_begin.

technobly: @avtolstoy I just looked into this, and it doesn't look like that is a viable option for all i2c interfaces.  It would work for fuel gauge, but not for something that required sending an address or custom i2c mode to be set in the hal_i2c_begin() method.

Since we know what i2c interface we're working with, if it's the internal one for PMIC/FuelGauge we could in fact call begin at the end of init, maybe also only if the interface was configured && enabled already upon entering init, and buffers were changing.  Just to keep the scope of this fix small.

technobly: Unless it's always one of these two options
```
void TwoWire::begin(void)
{
	hal_i2c_begin(_i2c, I2C_MODE_MASTER, 0x00, NULL);
}

void TwoWire::begin(uint8_t address)
{
	hal_i2c_begin(_i2c, I2C_MODE_SLAVE, address, NULL);
}
```
Then we just always use `I2C_MODE_MASTER, 0x00` unless mode is `I2C_MODE_SLAVE`, then we use  `I2C_MODE_SLAVE, address`.

technobly: I ended up fixing only for PMIC/FuelGauge to keep it clean.  This is the only one that is expected not to have to call begin() as well.  There may be a customer that init's but does not want the i2c interface active until they explicitly call begin().

avtolstoy (approved):

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -375,6 +375,13 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     memset((void *)i2cMap[i2c].rx_buf, 0, i2cMap[i2c].rx_buf_size);
     memset((void *)i2cMap[i2c].tx_buf, 0, i2cMap[i2c].tx_buf_size);
 
+// sc-137389
+#if HAL_PLATFORM_PMIC_BQ24195
```
avtolstoy, line 387: As discussed: gen4 + check for enabled state.

PR #2891, opened 2025-12-10
[b5som][eg91nax][quectel] fixes policyman/svc_mode setting preventing IMSI switching [sc-137388]

technobly:
### Problem

- A small number of EG91NAX devices have been found to have the `policyman/svc_mode` setting in an incorrect state on first boot, resulting in SIM IMSI switching not occurring.  This can lead to the device not being able to register with the cellular network.
- At this time only EG91NAX is affected.

### Solution

- Quectel has released an updated modem firmware `EG91NAXGAR07A03M1G_A0.002.A0.002` for new devices that automatically checks and fixes this setting. However it may not be practical for all customers to update devices to this firmware, thus we will provide a similar solution in Device OS.
- Check the policyman/svc_mode setting on warm/cold boot and every 60s during the registration process for good measure.

### Steps to Test

- Use a b5som EG91NAX based device with a deactivated SIM and attached antenna
- Run the included test app, and triple click the MODE button to set the policyman/svc_mode to the incorrect setting of "00".  Device should recover automatically to "02" on warm/cold boot, or during registration.

### Example App

```c
#include "Particle.h"
SYSTEM_MODE(SEMI_AUTOMATIC);
SYSTEM_THREAD(ENABLED);
bool button_clicked_3 = false;

{
        if (clicks >= 3) {
            button_clicked_3 = true;
        }
    }
}

void setup() {
    delay(5000);
    Particle.connect();
}

void loop() {
    if (button_clicked_3) {
        button_clicked_3 = false;
        Cellular.command(60000, "AT+QNVFW=\"policyman/svc_mode\",00");
    }
}
```

### References

- [sc-137388]

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -113,6 +113,8 @@ const unsigned REGISTRATION_TIMEOUT = 10 * 60 * 1000;
 const unsigned REGISTRATION_INTERVENTION_TIMEOUT = 15 * 1000;
 const unsigned REGISTRATION_TWILIO_HOLDOFF_TIMEOUT = 5 * 60 * 1000;
 
+const unsigned QUECTEL_POLICYMAN_SRV_MODE_CHECK_INTERVAL = 60 * 1000;
+
 const system_tick_t QUECTEL_COPS_TIMEOUT = 3 * 60 * 1000;
 const system_tick_t QUECTEL_CFUN_TIMEOUT = 3 * 60 * 1000;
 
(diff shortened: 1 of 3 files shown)


avtolstoy (approved):

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1397,6 +1404,43 @@ int QuectelNcpClient::setModuleFunctionality(CellularFunctionality cfun, bool ch
...
+
+    int r = SYSTEM_ERROR_UNKNOWN;
+
+    if (mode == CellularPolicymanServiceMode::FULL_SERVICE) {
+        r = parser_.execCommand(10000, "AT+QNVFW=\"policyman/svc_mode\",02");
+    } else if (mode == CellularPolicymanServiceMode::NO_SERVICE) {
+        r = parser_.execCommand(10000, "AT+QNVFW=\"policyman/svc_mode\",00");
+    }
+
+    CHECK_PARSER(r);
```
avtolstoy, line 1438: Maybe add a warning log if we don't get an OK back? But I guess we will anyway see that in the AT log..

technobly: Yep, we'd see an ERROR, which technically sounds worse than a warning.  In that case, the command is likely not supported by the version of EG91NAX.  Currently Quectel could not tell us which versions of EG91NAX support or do not support this command.  Thus we run it and operate on it only if it returns a valid response.

PR #2892, opened 2025-12-19
Environment variables

sergeuz:
For testing, the module version of the system firmware needs to be set to 6400 or higher.

Includes flash space optimizations ([#2895](https://github.com/particle-iot/device-os/pull/2895)).


.workbench/manifest.json
```diff
@@ -6,7 +6,7 @@
             "firmware": "deviceOS@source",
             "compilers": "gcc-arm@10.2.1",
             "tools": "buildtools@1.1.1",
-            "scripts": "buildscripts@1.15.0",
+            "scripts": "buildscripts@1.16.0",
```
monkbroc: TODO: update to support `env=env.json` in `project.properties`

hal/shared/filesystem.h
```diff
@@ -107,6 +113,48 @@ struct FsLock {
     filesystem_t* fs_;
 };
 
-} } /* particle::fs */
+// TODO: Implement Input/OutputStream?
+class File {
```
avtolstoy, line 117: As discussed, let's add a TODO for now to add `FsLock` to the methods in this class, and a note that it expects external lock management for now, just for visibility.

avtolstoy (approved):

services/inc/dct_file.h
```diff
@@ -126,12 +126,9 @@ class DctFile {
 
         LOG_DEBUG(INFO, "Filesystem mounted");
 
-        int r = lfs_mkdir(lfs(), "/sys");
```
XuGuohui, line 129: Without making the dir first, Gen4 runs into assertion failure at `SPARK_ASSERT(open(flags) == 0);`

sergeuz: Hm, do you know why? `/sys` is created in `filesystem_mount()` which is the very first thing `DctFile::init()` does.

XuGuohui: The bootloader has been excluded by `#if MODULE_FUNCTION != MOD_FUNC_BOOTLOADER` in `filesystem_mount()`, so it won't create the `/sys` folder on first boot.

sergeuz: Ahh, I see. Nice catch.

PR #2893, opened 2025-12-19
Adds support for I2C::setSpeed() from 10kHz to 99kHz in 1kHz increments

technobly:
### Feature

- Adds support for `I2C::setSpeed()` from 10kHz to 99kHz in 1kHz increments for Gen 3 and Gen 4 devices

### Steps to Test

- Run `device-os-test run bsom wiring/no_fixture -v -- -fixture` on SoM eval board with FuelGauge
- Run `device-os-test run msom wiring/no_fixture -v -- -fixture` on SoM eval board with FuelGauge

### DIFF
--- a/hal/inc/i2c_hal.h
+++ b/hal/inc/i2c_hal.h
@@ -84,6 +84,7 @@ typedef enum hal_i2c_state_t {
 } hal_i2c_state_t;
 
 /* Exported macros -----------------------------------------------------------*/
+#define CLOCK_SPEED_10KHZ          (uint32_t)10000
 #define CLOCK_SPEED_100KHZ         (uint32_t)100000
 #define CLOCK_SPEED_400KHZ         (uint32_t)400000
 #define HAL_I2C_DEFAULT_TIMEOUT_MS (100)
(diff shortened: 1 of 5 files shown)


avtolstoy (approved):
My main comment would be why not support 100kHz to 400kHz range as well? Works fine on both Gen 3 and Gen 4 (I modified HAL and the test).

technobly:
> My main comment would be why not support 100kHz to 400kHz range as well? Works fine on both Gen 3 and Gen 4 (I modified HAL and the test).

@avtolstoy mainly because the scope was set for below 100kHz only, but also because the Gen 3 formula doesn't calculate exactly the same constant for 100kHz.  It could be still be kept the same as it is but the 10kHz to 99kHz range expanded to 10kHz to 400kHz I suppose.  Check 400kHz constant, then 100kHz constant, then 10kHz to 400kHz range, else defaulted to 100kHz.  Of course the simplest code is just to check if in range or default to 100kHz, with potential minor breaking change of frequency at the old constants.  Easy changes though once we decide.

PR #2896, opened 2026-02-13
Cellular Env Vars

technobly:
# Feature

- Cellular Env Vars
  - PARTICLE_CELLULAR_PREFERRED_BANDS
  - PARTICLE_CELLULAR_FORBIDDEN_BANDS
  - PARTICLE_CELLULAR_PREFERRED_PLMN
  - HIL tests
  - Unit tests

### References

- sc-137969
- sc-137970
- sc-137973
- sc-137975
- sc-137984


hal/network/util/ncp_env_var.h
```diff
@@ -0,0 +1,42 @@
...
+#if HAL_PLATFORM_ENV
+
+#include <cstddef>
+#include <cstdint>
+
+namespace particle {
+
+#define MAX_CELLULAR_PREFFERED_PLMN_ENTRIES (4)
+
+int getEnvVar(const char* name, char* val, size_t valSize);
```
avtolstoy: These are already implemented in `system_env.h/cpp`

hal/network/util/ncp_env_var.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+bool hasEnv(const char* name) {
+    int r = system_get_env(name, nullptr /* buf */, 0 /* buf_size */, nullptr /* reserved */);
+    return r >= 0;
+}
+
+#endif // PLATFORM_ID == PLATFORM_GCC
+
+void hexString128toUint64Array(const char* hexString128, uint64_t* bands) {
```
avtolstoy: There is `hexToBytes()` already, which is much simpler. Just convert to flat uint8_t* buffer first and then construct uint64_t bands[0] and bands[1] depending on endianness etc out of specific bytes.

IMO this should be unit tested with example "1232131232155737758" -> bands[0] == 0x213213213 && bands[1] == 0x557851. Very easy to screw things up.

avtolstoy: Use `uint64_t bands[2]` in signature or even better just create a
```c++
struct CellularBandMask {
    CellularBandMask(const char* str);
    CellularBandMask(uint64_t low, uint64_t high = 0);
    CString toString() const;
    uint64_t low() const;
    uint64_t high() const;
    // ... etc
    uint64_t bands[2];
};
```

or something along the lines and use that everywhere. It's very awkward to pass around `uint64_t*`

You can then also define all your constants as this type and add whatever operators/methods to compare etc and keep essentially everything contained in this class.

hal/network/util/ncp_env_var.cpp
```diff
@@ -0,0 +1,195 @@
...
+    char envBands[32+1] = {};
+    if (hasEnv(varName)) {
+        if (getEnv(varName, envBands, sizeof(envBands)) == SYSTEM_ERROR_NONE) {
+            hexString128toUint64Array(envBands, bands);
+            // LOG(INFO, "%s=0x%llX%016llX", varName, bands[1], bands[0]);
+        }
+    }
+}
+
+int getEnvPreferredPlmn(const char* varName, char plmns[][7]) {
```
avtolstoy: Same thing here. `char plmns[][7]` is very awkward to see and pass around.
```
struct CellularPlmnList {
   CString plmns[MAX_CELLULAR_PREFFERED_PLMN_ENTRIES];
};
```

or even use `Vector<CString>`.

hal/network/util/ncp_env_var.cpp
```diff
@@ -0,0 +1,195 @@
...
+    if (!plmns || !varName) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+
+    char envPlmns[32+1] = {};
+    int plmnCount = 0;
+    if (hasEnv(varName)) {
+        if (getEnv(varName, envPlmns, sizeof(envPlmns)) == SYSTEM_ERROR_NONE) {
+            char* token = strtok(envPlmns, ",");
```
avtolstoy: Better use `strtok_r` in system. `strtok` will allocate per-thread reent struct for each thread that uses it and keep it around.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+    // LOG(INFO, "ENV BANDS: [65-128]=0x%llx, [1-64]=0x%016llx", envPreferredBands[1], envPreferredBands[0]);
+
+    // Log bandmask, and change bandmask if necessary
+    auto respBand = parser_.sendCommand("AT+QCFG=\"band\"");
+    uint64_t uint64LTEbandsCurrent[2] = {};
+    char qbandGSM_Str[4+1] = {};
+    char qbandLTE_Str[32+1] = {};
+    char qbandCatNB_Str[32+1] = {};
+    // BG95_M5: +QCFG: "band",0xf,0x100002000000000f0e189f,0x10004200000000090e189f <GSM>,<CAT-M1-1-128>,<CAT-NB-1-128>
+    auto retBand = CHECK_PARSER(respBand.scanf("+QCFG: \"band\",0x%4[^,],0x%32[^,],0x%32[^,]", qbandGSM_Str, qbandLTE_Str, qbandCatNB_Str));
```
avtolstoy: If `CellularBandMask` were to be used as I outlined below this could be much more simplified with using `%llx` scanf and then just a few `CellularBandMask operator==` operations.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+    // LOG(INFO, "%s,%s,%s", qbandGSM_Str, qbandLTE_Str, qbandCatNB_Str);
+    CHECK_PARSER_OK(respBand.readResult());
+
+    if (retBand == 3) {
+        hexString128toUint64Array(qbandLTE_Str, uint64LTEbandsCurrent);
+
+        if (uint64LTEbandsCurrent[0] != envPreferredBands[0] ||
+                uint64LTEbandsCurrent[1] != envPreferredBands[1]) {
+            if (envPreferredBands[1]) {
+                sprintf(qbandLTE_Str, "%llx%016llx", envPreferredBands[1], envPreferredBands[0]);
```
avtolstoy: use `snprintf` universally, do not use `sprintf`.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int QuectelNcpClient::parseEfSize(unsigned int fid, unsigned int& efSize) {
+    char fcpHexStr[128] = {};
+    int sw1, sw2;
+    auto resp = parser_.sendCommand("AT+CRSM=192,%u,0,0,30", fid);
+    while (resp.hasNextLine()) {
+        // Other URCs can pop up here before +CRSM and cause this to fail
```
avtolstoy: This shouldn't happen. With echo enabled once the first command character is echoed back the URCs are suppressed.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+        if (r != AtResponse::OK) {
+            lastError = r;
+        }
+    }
+    CHECK_PARSER_OK(lastError);
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int QuectelNcpClient::parseEfSize(unsigned int fid, unsigned int& efSize) {
```
avtolstoy: Nit: could also just return the size simply.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+    return SYSTEM_ERROR_NONE;
+}
+
+int QuectelNcpClient::syncUserPlmn(char envPreferredPlmn[][7], int preferredPlmnCount) {
+    if (!envPreferredPlmn) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+
+    Vector<unsigned int> cpolIdxs;
+    Vector<const char*> cpolPlmns;
```
avtolstoy: Use `CString`.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+    }
+
+    if (needsUpdate) {
+        CellularSignalQuality qual;
+        queryAndParseAtCops(&qual);
+        bool needsReselection = false;
+        if (cgi_.mobile_country_code > 0 && cgi_.mobile_network_code > 0) {
+            char currentPlmn[10+1] = {};
+            if (CGI_FLAG_TWO_DIGIT_MNC & cgi_.cgi_flags) {
+                sprintf(currentPlmn, "%03u%02u", cgi_.mobile_country_code, cgi_.mobile_network_code);
```
avtolstoy: Use `snprintf` universally.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1046,256 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+            } else {
+                sprintf(currentPlmn, "%03u%03u", cgi_.mobile_country_code, cgi_.mobile_network_code);
```
avtolstoy: Use `snprintf` universally.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1466,27 +1754,81 @@ int QuectelNcpClient::configureApn(const CellularNetworkConfig& conf) {
...
+                    break;
+                }
+            }
+            // Set the APN if any are true:
+            // - We don't match anything from the scanf()
+            // - cgdcontApnVal does not match expected APN
+            // - No APN is set or explicitly blank "" APN (apnLength == 0)
+            // - cgdcontIpVal is not "IP".
+            if (cgdcontApnVal[0] != '\0' &&
+                    strncmp(cgdcontApnVal, netConf_.hasApn() ? netConf_.apn() : "", sizeof(cgdcontApnVal)) == 0 &&
```
avtolstoy, line 1739: `strcmp` ? they are all null terminated, right?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1164,18 +1192,28 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
...
+            if (useEnvBands) {
+                if (ncpId() == PLATFORM_NCP_SARA_R410) {
+                    newProf = static_cast<int>(UbloxSaraUmnoprof::STANDARD_EUROPE);
+                } else { // R510
+                    newProf = static_cast<int>(UbloxSaraUmnoprof::STANDARD_GLOBAL);
+                }
+            }
             // TWILIO Super SIM
-            if (netConf_.netProv() == CellularNetworkProvider::TWILIO) {
+            else if (netConf_.netProv() == CellularNetworkProvider::TWILIO) {
```
avtolstoy: Nit/style: don't break blocks with a comment. `else if` kind of easily missed.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1272,6 +1309,146 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
...
+}
+
+int SaraNcpClient::syncUserPlmn(char envPreferredPlmn[][7], int preferredPlmnCount) {
+    if (!envPreferredPlmn) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+
+    // Query all currently set PLMNs
+    Vector<unsigned int> cpolIdxs;
+    Vector<const char*> cpolPlmns;
```
avtolstoy: Same as in quectel.

user/tests/integration/system/env/env.cpp
```diff
@@ -75,6 +80,145 @@ bool waitListening(bool state, system_tick_t timeout = LISTENING_MODE_STATE_CHAN
...
+    }
+
+    // Force skip if not testing on cellular network interface
+    auto network = (int)TestSuite::instance()->network();
+    if (network != NETWORK_INTERFACE_CELLULAR && network != NETWORK_INTERFACE_ALL) {
+        modemType = ModemType::UNSUPPORTED;
+        return modemType;
+    }
+
+    switch (getNcpId()) {
```
avtolstoy: Nit: manufacturer is already encoded in NCP ID. See `PLATFORM_NCP_MANUFACTURER()`

user/tests/integration/system/env/env.cpp
```diff
@@ -75,6 +80,145 @@ bool waitListening(bool state, system_tick_t timeout = LISTENING_MODE_STATE_CHAN
...
+
+        char hexBuf[17] = {};
+        snprintf(hexBuf, sizeof(hexBuf), "%llx", val);
+
+        return String(hexBuf);
+    }
+    return "";
+}
+
+// Simplifed checker for lowest 4 bands/bits only!!
```
avtolstoy: 😅  hexToBytes + uint8_t buffer + figure out byte and bit within byte? `CellularBandMask` if it were a struct could very well be reused too in tests.

user/tests/integration/system/env/env.cpp
```diff
@@ -430,6 +576,339 @@ test(16_particle_ethernet_enable_cleanup) {
 }
 #endif // HAL_PLATFORM_ETHERNET
 
+
+#if HAL_PLATFORM_CELLULAR
+
+test(17_particle_cellular_preferred_bands_init) {
```
avtolstoy: Let's move out these tests into a separate suite. `system/cellular_env` or something like that.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -99,6 +101,29 @@ const auto QUECTEL_NCP_MAX_MUXER_FRAME_SIZE = 1509;
 const auto QUECTEL_NCP_KEEPALIVE_PERIOD = 5000; // milliseconds
 const auto QUECTEL_NCP_KEEPALIVE_MAX_MISSED = 5;
 
+// Bandmask same for BG95-M1 ~ BG95-M6 & BG95-MF
+// Note: BG95-M1 does not support NB mode
+// BG95-M4 not supported (bands are different as well)
+const uint64_t QUECTEL_NCP_BANDMASK_CATM1_1_64_BG95 = 0xF0E189F;      // Bands 1,2,3,4,5,8,12,13,18,19,20,25,26,27,28 [all default enabled]
```
avtolstoy: `uint64_t getBandMaskByNcpIdForRAT(PlatformNCPIdentifier id, CellularAccessTechnology rat)` perhaps? Could also be in a separate header and reused by tests.

user/tests/integration/system/env/env.cpp
```diff
@@ -430,6 +576,339 @@ test(16_particle_ethernet_enable_cleanup) {
...
+
+        defaultMask = readCurrentLteMask();
+        Log.info("Default LTE band mask (%s): %s",
+                modemType == ModemType::QUECTEL ? "QUECTEL" : "UBLOX",
+                defaultMask.c_str());
+        assertTrue(defaultMask.length() > 0);
+
+
+        // band 1 or band 2 should be enabled
+        assertTrue(isBandEnabled(defaultMask, 1) || isBandEnabled(defaultMask, 2));
```
avtolstoy: Shouldn't we verify that the band mask is an exact match the default one? What if 1 or 2 are set and the rest are out of whack?

Upper 64-bit word should be tested too.

user/tests/integration/system/env/env.cpp
```diff
@@ -430,6 +576,339 @@ test(16_particle_ethernet_enable_cleanup) {
...
+
+test(19_particle_cellular_preferred_bands_set) {
+    if (getModemType() == ModemType::UNSUPPORTED) {
+        skip();
+        return;
+    }
+
+    assertTrue(System.hasEnv("PARTICLE_CELLULAR_PREFERRED_BANDS"));
+    assertEqual(System.getEnv("PARTICLE_CELLULAR_PREFERRED_BANDS"),
+            String("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFC")); // disables bands 1 & 2
```
avtolstoy: Upper 64-bit word should be covered by tests too I think.

user/tests/integration/system/env/env.cpp
```diff
@@ -430,6 +576,339 @@ test(16_particle_ethernet_enable_cleanup) {
...
+    });
+
+    newMask = readCurrentLteMask();
+    Log.info("LTE band mask after PREFERRED_BANDS env var (%s): %s",
+            modemType == ModemType::QUECTEL ? "QUECTEL" : "UBLOX",
+            newMask.c_str());
+    assertTrue(newMask.length() > 0);
+
+    // band 1 AND band 2 should be disabled
+    assertTrue(!isBandEnabled(newMask, 1) && !isBandEnabled(newMask, 2));
```
avtolstoy: Same as above.

user/tests/integration/system/env/env.spec.js
```diff
@@ -26,6 +26,9 @@ before(async function() {
 });
 
 test('01_particle_ble_enable_init', async function() {
+    // This is a bit of a hack, but should work fine
+    this.test.parent.particle.network = 'wifi';
```
avtolstoy: Leave this as-is. I have fixes for these tests for M SoM.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: we already use strncmp in the client so was trying to reuse things as much as possible.

avtolstoy: That's only used for `CONNECT 15000000000` responses, because we just want to compared `CONNECT == CONNECT` and ignore the rest.

hal/network/util/ncp_env_var.cpp
technobly: I swore we had that somewhere as well but couldn't find it... will rework this.

user/tests/integration/system/env/env.cpp
technobly: The lower two bands are common to all devices, upper is not.  So that's the thinking here... just testing that a set env var makes the appropriate change in ncp client.

user/tests/integration/system/env/env.spec.js
technobly: Ok, was mainly the issue for msom, but maybe this is why some other platforms are currently failing test 4... or could be they don't have wifi credentials set... still need to look at that.

user/tests/integration/system/env/env.cpp
avtolstoy: I get it, but we are essentially not testing any changes to the upper one right now for modems that do use this...

technobly: Ok, I think this is just u-blox anyways since quectel uses the bandmask as a 128bit hex string... so we shouldn't need to do it for quectel (even though we parse 64bits out at a time).  When I rework the hexToBytes it will change as well.

user/tests/integration/system/env/env.spec.js
technobly: @avtolstoy what's your fix for this so I can test locally with `msom --network=cellular|wifi` ?

avtolstoy: If it works for you, use this, the one I have is a bit more involved (now that I saw issues today with msom and --network). That's also part of the reason I'm suggesting we move cellular specific ones into a dedicated `system/cellular_env` test. Then you can just skip() if `TestSuite::instance()->network() != NETWORK_INTERFACE_CELLULAR && TestSuite::instance()->network() != NETWORK_INTERFACE_ALL`. They will not be tested with msom using wifi only and they will be tested with msom using cellular only.

avtolstoy: And would have `platform('cellular', 'msom');` in spec.js

technobly: Ok no problem, I moved my tests to `system/cellular_env`

user/tests/integration/system/env/env.cpp
technobly: There's no DYNALIB entry for this platform_primary_ncp_identifier()

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: This code is copied from sara_ncp_client.cpp which has been in use for a long time now, so I'd rather not change it.  The thought was that it was a bit safer to be explicit with the size of the compare.

avtolstoy: Ok, yup, I see now. thought you meant it was used here in quectel_ncp_client. :+1: leave as-is then.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: I just had a case where this CRSM command fired, got a +CGEV URC right after and nothing else for 90s then timed out.  Later I saw other cases of this scenario happening in logs and wasn't failing to finish the CRSM command... so must have been a temporary case.  I will revert this to the standard flow.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1022,232 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+            envPreferredBands = defaultBands;
+        }
+        envPreferredBands = (envPreferredBands & defaultBands) & ~envForbiddenBands;
+    } else {
+        envPreferredBands = defaultBands;
+    }
+
+    // Log bandmask, and change bandmask if necessary
+    auto respBand = parser_.sendCommand("AT+QCFG=\"band\"");
+    char qbandGSM_Str[4+1] = {};
```
avtolstoy: Nit/style: one case :)

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1022,232 @@ int QuectelNcpClient::checkNetConfForImsi() {
...
+    CHECK_TRUE(efSize <= 64, SYSTEM_ERROR_INTERNAL);
+    char ffBuf[128 + 1];
+    memset(ffBuf, 'F', hexLen);
+    ffBuf[hexLen] = '\0';
+    CHECK_PARSER(parser_.execCommand("AT+CRSM=214,%u,0,0,%u,\"%s\"", fid, efSize, ffBuf));
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int QuectelNcpClient::syncUserPlmn(Vector<CString>& envPreferredPlmn, int preferredPlmnCount) {
```
avtolstoy: `preferredPlmnCount` isn't necessary? You can always get the size of the vector. Also pass by `const &` if you are not going to be modifying `envPreferredPlmn`.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1020,6 +1022,232 @@ int QuectelNcpClient::checkNetConfForImsi() {
     return SYSTEM_ERROR_TIMEOUT;
 }
 
+int QuectelNcpClient::setupBands() {
+
+    CellularBandMask defaultBands(
+        getBandMaskByNcpIdForRAT(ncpId(), CellularAccessTechnology::LTE_CAT_M1, false /* upper */, false /* legacy */),
```
avtolstoy, line 1028: Do we care about the upper/lower split? Why not just return CellularBandMask here? For modems that do not support long band mask upper bits would be just 0 anyway?

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -1272,6 +1298,141 @@ int SaraNcpClient::selectNetworkProf(ModemState& state) {
...
+        if (r != AtResponse::OK) {
+            lastError = r;
+        }
+    }
+    CHECK_PARSER_OK(lastError);
+
+    return SYSTEM_ERROR_NONE;
+}
+
+int SaraNcpClient::syncUserPlmn(Vector<CString>& envPreferredPlmn, int preferredPlmnCount) {
```
avtolstoy: Same here.

hal/shared/ncp_band_mask.h
```diff
@@ -0,0 +1,208 @@
...
+
+    void setFromHexString(const char* hexStr) {
+        bands[0] = 0;
+        bands[1] = 0;
+        if (hexStr) {
+            hexString128toUint64Array(hexStr, bands);
+        }
+    }
+
+    void setFromDecimalStrings(const char* lowStr, const char* highStr = nullptr) {
```
avtolstoy, line 152: :+1: Common pattern is usually a static method that returns a constructed instance e.g. but this is also fine.
```
static CellularBandMask fromDecimalStrings(const char* low, const char* high = nullptr) {
    return CellularBandMask(low, high, 10); //
}
```

Nit: perhaps just make `const char* x, int base = 16` and `const char* low, const char* high, int base = 16` methods/constructors and can deal at the same time for both cases but it's fine as-is as well.

test/unit_tests/cellular/cellular.cpp
```diff
@@ -151,4 +152,320 @@ TEST_CASE("cellular_printable") {
         // printf("%s", output);
         REQUIRE(strncmp(output, "700,800,900,2100,2600", sizeof(output)) == 0);
     }
-}
\ No newline at end of file
+}
+
+TEST_CASE("hexString128toUint64Array") {
```
avtolstoy, line 158: Nit: ideally would test CellularBandMask here as well

test/unit_tests/cellular/cellular_stubs.h
```diff
@@ -0,0 +1,73 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+enum class CellularAccessTechnology {
```
avtolstoy: It should be possible to include both `platform_ncp.h` and `cellular_ncp_client.h` in tests. It's just a matter of providing the stubs for what they require, not having stub values for the contents themselves.

test/unit_tests/cellular/cellular.cpp
technobly: It's there at the bottom TEST_CASE("CellularBandMask") { https://github.com/particle-iot/device-os/pull/2896/changes/BASE..797491a2b6634509f6e05e53164281904c40ab53#diff-fd50e628151b28b0bbd894c39b39920d5062779a9acfbd2917e6446ed207228aR323

Along with tests for TEST_CASE("getBandMaskByNcpIdForRAT") {

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
technobly: Yes, I've kind of baked this in everywhere now.  It's kind of nice being able to deal with the 128bit bands in two 64bit chunks at times as well.

test/unit_tests/cellular/cellular_stubs.h
technobly: It was trying to pull in too many other headers from places deep within system and hit it with a hammer to make the tests run.  I'll let you work your magic on it.

avtolstoy: :+1: As discussed I'll clean this up before we merge.

avtolstoy (approved):

PR #2898, opened 2026-02-19
[gen3] USB Tethering

scott-brust:
USB CDC Tethering Implementation

  ## Overview

Extends the existing USART-based PPP tethering feature to support USB CDC as an alternative serial transport. The host connects to the device's USB CDC serial port, goes through the same AT command handshake and PPP negotiation flow, and gets a tethered network connection over USB instead of a hardware UART.

  ## Changes

  ### New files:
  - `hal/inc/usb_hal_private.h` — Private C API exposing FreeRTOS event group functions for USB CDC (`hal_usb_cdc_pvt_get_event_group_handle`, `hal_usb_cdc_pvt_wait_event`)
  - `hal/src/nRF52840/serial_usb_stream.h` / .cpp — SerialUSBStream class, a stream adapter that wraps the USB CDC HAL (HAL_USB_USART_* functions) behind the same EventGroupBasedStream interface used by the existing SerialStream (USART). Implements read, write, peek, skip, flush, waitEvent, and availForRead/availForWrite.

  ### Modified files:
  - `hal/src/nRF52840/usb_hal_cdc.c` — Added a FreeRTOS event group (ev_group) to the USB instance. READABLE is signaled from the RX_DONE ISR handler (via xEventGroupSetBitsFromISR) when data is successfully copied into the rx FIFO. WRITABLE is signaled from TX_DONE after schedule_tx frees FIFO space. An enableEvent function checks current FIFO state from thread context to handle the race where data is already available before waitEvent blocks.
  - `hal/network/lwip/pppservernetif.cpp` — PppServerNetif::start() now creates either a SerialStream or SerialUSBStream based on settings_.usbserial. AT command handler lambdas updated to pass this (instead of serial_.get()) to properly distinguish transport type for flow control (+IFC) and baud rate (+IPR) queries. The ATD connect handler similarly updated.
  - `wiring/inc/spark_wiring_tether.h` — Added TetherUSBConfig struct and TetherClass::bind(TetherUSBConfig) overload. Added TetherInterface enum and activeSerialInterface() accessor.
  - `wiring/src/spark_wiring_tether.cpp` — Implements the USB bind path: sets settings.serial = 0 and settings.usbserial to the USB serial instance, signaling PppServerNetif to use SerialUSBStream.

### Steps to Test

- Build + flash this branch + test app to B5SOM harware
- Connect B5SOM to SBC like Tachyon or RPI5
- Create + enable ppp connection over `ttyACM0` (see host setup below)

#### Host Setup
Tachyon and RPI need udev rules update to tell modem manager to not block the `ttyACM0` connection:
```
sudo tee /etc/udev/rules.d/99-particle-modem.rules << 'EOF'
SUBSYSTEM=="usb", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_DEVICE_PROCESS}="1"
SUBSYSTEM=="tty", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_CANDIDATE}="1", ENV{ID_MM_PORT_TYPE_AT_PRIMARY}="1"
EOF

sudo udevadm control --reload-rules
sudo udevadm trigger
sudo systemctl restart ModemManager
```

The rest of the setup is the ame as with USART tethering, just make sure the serial endpoint is the USB one instead. for example on tachyon
```
nmcli connection add type gsm con-name "particle" gsm.apn "particle" connection.interface ttyACM0 ppp.crtscts true autoconnect yes
```

Scan + list for modems
```
sudo mmcli -S
mmcli -L 
```

Activate PPP connection, test interface
```
nmcli c up particle
ping -I ppp0 google.com
```

Download test
```
curl --interface ppp0 -sS -o /dev/null -w "%{speed_download}\n" http://technobly.com/100/phrack.29.phk | awk '{ printf "Download speed: %.2f KiB/s\n", $1/1024 }'
```

Upload test
```
curl --interface ppp0 -T /dev/urandom -o /dev/null --max-time 10 \
  -w "Speed: %{speed_upload} bytes/sec\n" \
  http://speedtest.tele2.net/upload.php
```

### Example App
Use mhat firmware or following test snippet

```c
#include "application.h"

Serial1LogHandler logHandler(115200, LOG_LEVEL_ALL, 
{
    { "app", LOG_LEVEL_ALL }, 
}
);
SYSTEM_MODE(AUTOMATIC);

void setup() {


    // waitUntil(Serial.isConnected);
    // Enable Cellular
    Cellular.on();
    Cellular.connect();
    // Bind Tether interface to Serial with default settings (8n1 + RTS/CTS flow control)
    // Tether.bind(TetherSerialConfig().baudrate(921600).serial(Serial1));

    Tether.bind(TetherUSBConfig());
    Tether.on();
    Tether.connect();

    Particle.connect();
}

void loop() {
}
```

### References

Test download speeds around 36 kb/s
```
particle@tachyon-1c5472eb:~$ curl --interface ppp0 -sS -o /dev/null -w "%{speed_download}\n" http://technobly.com/100/phrack.29.phk | awk '{ printf "Download speed: %.2f KiB/s\n", $1/1024 }'
Download speed: 36.87 KiB/s
```

Test upload speeds around 65 kb/s
```
particle@tachyon-1c5472eb:~$ curl --interface ppp0 -T /dev/urandom -o /dev/null --max-time 10 \
  -w "Speed: %{speed_upload} bytes/sec\n" \
  http://speedtest.tele2.net/upload.php
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  639k    0     0    0  639k      0  65526 --:--:--  0:00:10 --:--:-- 64468
curl: (28) Operation timed out after 10000 milliseconds with 0 bytes received
Speed: 65526 bytes/sec
```

edit: Overnight soak test as of March 4
```
════════════════════════════════════════════════════════════
  RUNNING AVERAGES — ALL TESTS
════════════════════════════════════════════════════════════

  ── iperf3 ──

  iperf3 → LeaseWeb (LA)
    ↑ Sender   avg : 702.45 Kbits/sec  (n=99)
    ↓ Receiver avg : 489.41 Kbits/sec  (n=99)

  iperf3 → Clouvider (LA)
    ↑ Sender   avg : 684.22 Kbits/sec  (n=157)
    ↓ Receiver avg : 558.87 Kbits/sec  (n=157)

  ── curl ──

  curl upload → Tele2
    ↑ Speed avg : 48.74 KB/s  (n=327)

  curl download → technobly
    ↓ Speed avg : 26.49 KB/s  (n=327)
```


wiring/inc/spark_wiring_tether.h
```diff
@@ -24,12 +24,29 @@
...
 #include "system_network.h"
 #include "spark_wiring_usartserial.h"
+#include "spark_wiring_usbserial.h"
 #include "ifapi.h"
 #include "scope_guard.h"
 #include "check.h"
 
 namespace particle {
 
+enum class TetherInterface {
```
scott-brust, line 34: Not sure if we want to create this enum, its only used for determining if tethering is enabled for USB logging. I think its helpful to have USB logging when Tethering is on USART, so knowing which interface is active is helpful.

wiring/src/spark_wiring_tether.cpp
```diff
@@ -23,6 +23,19 @@ namespace particle {
 
 TetherClass Tether;
 
+TetherUSBConfig::TetherUSBConfig()
+        : usbSerial_(Serial) {
+}
+
+TetherUSBConfig& TetherUSBConfig::usbserial(USBSerial& s) {
+    usbSerial_ = s;
+    return *this;
```
scott-brust, line 32: There isnt really any configuration options for USB at the moment, so this class doesnt really do that much. We could probably combine to one `TetherConfig` class, but that would break existing code.

hal/network/lwip/pppservernetif.cpp
```diff
@@ -325,23 +342,39 @@ int PppServerNetif::start() {
...
         CHECK_TRUE(v > 0, SYSTEM_ERROR_INVALID_ARGUMENT);
-        CHECK(stream->setBaudRate(v));
+        CHECK_TRUE(self->serial_.get(), SYSTEM_ERROR_INVALID_STATE);
+
+        if (self->settings_.serial) {
+            auto stream = static_cast<SerialStream*>(self->serial_.get());
+            CHECK(stream->setBaudRate(v));
+        } else {
+            auto stream = static_cast<SerialUSBStream*>(self->serial_.get());
+            CHECK(stream->setBaudRate(v));
```
scott-brust: This does result in the USB detach/attach being called for not really any good reason. We can make this no op for USB if we want instead...

hal/network/api/ifapi_driver_specific.h
```diff
@@ -32,7 +32,7 @@ typedef enum if_req_driver_specific_type {
     IF_REQ_DRIVER_SPECIFIC_WIZNET_PIN_REMAP = 1,
     // FIXME: for compatibility
     IF_WIZNET_DRIVER_SPECIFIC_PIN_REMAP = IF_REQ_DRIVER_SPECIFIC_WIZNET_PIN_REMAP,
-    IF_REQ_DRIVER_SPECIFIC_PPP_SERVER_UART_SETTINGS = 2
```
avtolstoy, line 35: keep the `IF_REQ_DRIVER_SPECIFIC_PPP_SERVER_UART_SETTINGS` enum value, we've shared some apps before with this, so let's not break those.

hal/network/api/ifapi_driver_specific.h
```diff
@@ -42,14 +42,15 @@ typedef struct if_wiznet_pin_remap {
     uint16_t int_pin;
 } if_wiznet_pin_remap;
 
-typedef struct if_req_ppp_server_uart_settings {
```
avtolstoy, line 45: keep the `if_req_ppp_server_uart_settings` alias, we've shared some apps before with this, so let's not break those.

hal/network/api/ifapi_driver_specific.h
```diff
@@ -42,14 +42,15 @@ typedef struct if_wiznet_pin_remap {
...
+typedef struct if_req_ppp_server_serial_settings {
```
avtolstoy, line 46: add ` __attribute__ ((__packed__))` just in case

hal/network/lwip/pppservernetif.cpp
```diff
@@ -387,7 +420,7 @@ void PppServerNetif::loop(void* arg) {
                 char tmp[256] = {};
                 while (self->serial_->availForRead() > 0) {
                     int sz = self->serial_->read(tmp, sizeof(tmp));
-                    // LOG(INFO, "input %d", sz);
+                    LOG(INFO, "input %d", sz);
```
avtolstoy: Fix

hal/src/nRF52840/serial_usb_stream.cpp
```diff
@@ -0,0 +1,210 @@
...
+    phyOn_ = true;
+    return 0;
+}
+
+// CTS/RTS Configs are no ops
+int SerialUSBStream::setConfig(uint32_t config, unsigned int baudrate /* optional */) {
+    if (!phyOn_ || !enabled_) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    HAL_USB_USART_End(serial_);
```
avtolstoy: This shouldn't really do anything?

wiring/src/spark_wiring_logging.cpp
```diff
@@ -476,7 +477,8 @@ int spark::detail::LogFilter::nodeIndex(const Vector<Node> &nodes, const char *n
 // spark::StreamLogHandler
 void spark::StreamLogHandler::logMessage(const char *msg, LogLevel level, const char *category, const LogAttributes &attr) {
 #if PLATFORM_ID != PLATFORM_GCC && !defined(LOG_IN_LISTENING_MODE)
-    if (stream_ == &Serial && Network.listening()) {
+    if (stream_ == &Serial &&
+        (Network.listening() || (particle::Tether.isOn() && particle::Tether.activeSerialInterface() == particle::TetherInterface::USB))) {
```
avtolstoy: :+1:

wiring/src/spark_wiring_tether.cpp
avtolstoy: all good :+1:

wiring/inc/spark_wiring_tether.h
avtolstoy: This is definitely useful :+1:

hal/network/lwip/pppservernetif.cpp
avtolstoy: This shouldn't be necessary at all for USB serial. Baudrate value is virtual, we don't care about it at all.

scott-brust: ok, I will make it a no op for USB here and in the USB `setBaudRate()` and  `setConfig()` functions. 
I wasnt sure if we cared enough to propagate the change to the USB layer. I dont know if it really has any effect at the USB CDC level.

avtolstoy: No effect at all. It's just something that hosts sends us to use as a virtual baudrate. It would make sense if for example we were a USB CDC <-> UART (hardware) bridge and it asked to open CDC with 19200 and we would configure hardware UART with 19200 baudrate based on that. We have nothing like that, so we don't really care about the baudrate.

hal/src/nRF52840/serial_usb_stream.cpp
```diff
@@ -0,0 +1,210 @@
...
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    if (size == 0) {
+        return 0;
+    } else if (size > 1) {
+        // Only support peeking first byte
+        return SYSTEM_ERROR_NOT_SUPPORTED;
+    }
+
+    auto r = HAL_USB_USART_Receive_Data(serial_, 1);
```
scott-brust, line 95: There isnt a USB_USART hal function to peek more than one byte at a time, so I just left it at that. I dont think the PPP server uses this function at all though. Same with `skip()`, `flush()` and I think even `on()`.

avtolstoy: It's not necessary to peek more. We don't have support for that in HAL, so just do 1 byte at a time. Private UART HAL has sized buffer access, but it's just an optimization there. For simplicity I wouldn't worry about it for now.

hal/network/api/ifapi_driver_specific.h
scott-brust: I just renamed it, the enum value is the same. Its only used internally in `Tether.bind()` so I dont think its a breaking change but ill revert it just to be safe.

avtolstoy: We have some apps we shared before that used this lower layer API before we introduce wiring APIs. It's fine to rename - keep the renames, it's all good, just keep the old names too with the same enum value too just for compatibility reasons.

scott-brust:
One outstanding item is how the particle udev linux rules should be updated to allow modem manager to consider the B5SOM tty connection as a modem. Right now the particle cli disallows this by default, but when enabling USB tethering this needs to be disabled. IE this snippet needs to be created
```
sudo tee /etc/udev/rules.d/99-particle-modem.rules << 'EOF'
SUBSYSTEM=="usb", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_DEVICE_PROCESS}="1"
SUBSYSTEM=="tty", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_CANDIDATE}="1", ENV{ID_MM_PORT_TYPE_AT_PRIMARY}="1"
EOF
```

Should we incorporate this into the 'canonical' tethering setup script?

wiring/inc/spark_wiring_tether.h
```diff
@@ -94,6 +111,14 @@ class TetherClass : public spark::NetworkClass {
         return network_ready(*this, 0,  NULL);
     }
 
+    bool isOn(void) {
```
avtolstoy: Weird, this should be working as-is. It's in `NetworkClass`

scott-brust: oh, yes you are right. I think basically all of these functions in `TetherClass` are redundant

wiring/src/spark_wiring_logging.cpp
```diff
@@ -476,7 +477,11 @@ int spark::detail::LogFilter::nodeIndex(const Vector<Node> &nodes, const char *n
 // spark::StreamLogHandler
 void spark::StreamLogHandler::logMessage(const char *msg, LogLevel level, const char *category, const LogAttributes &attr) {
 #if PLATFORM_ID != PLATFORM_GCC && !defined(LOG_IN_LISTENING_MODE)
-    if (stream_ == &Serial && Network.listening()) {
+    bool usbTetherOn = false;
+#if HAL_PLATFORM_PPP_SERVER
+    usbTetherOn = ((particle::Tether.activeSerialInterface() == particle::TetherInterface::USB) && particle::Tether.isOn());
```
avtolstoy: This is still not short circuited. The first check should be `stream_ == &Serial`, so that we don't even do anything else in case of `Serial1LogHandler` etc.

wiring/src/spark_wiring_tether.cpp
```diff
@@ -23,6 +23,19 @@ namespace particle {
 
 TetherClass Tether;
 
+TetherUSBConfig::TetherUSBConfig()
+        : usbSerial_(Serial) {
+}
+
+TetherUSBConfig& TetherUSBConfig::usbserial(USBSerial& s) {
```
avtolstoy, line 30: How about we get rid of a separate config for usb serial and just add an overload to `TetherSerialConfig` for `serial(USBSerial& s)? There is nothing special between UART and CDC serial usage as far as tethering as of right now. Baudrate and config don't matter either. `TetherUSBConfig` sounds more like we are doing something special like RNDIS/NCM/ECM.

wiring/src/spark_wiring_logging.cpp
avtolstoy:
```suggestion
    if (stream_ == &Serial) {
        if (Network.listening()) {
            return;
        }
#if HAL_PLATFORM_PPP_SERVER
        if ((particle::Tether.activeSerialInterface() == particle::TetherInterface::USB) && particle::Tether.isOn()) {
            return;
        }
#endif // HAL_PLATFORM_PPP_SERVER
    }
```

hal/src/nRF52840/usb_hal_cdc.c
```diff
@@ -669,3 +689,40 @@ int usb_hal_set_state_change_callback(HAL_USB_State_Callback cb, void* context,
     }
     return SYSTEM_ERROR_NO_MEMORY;
 }
+
+EventGroupHandle_t eventGroup() {
+    return m_usb_instance.ev_group;
+}
+
+int enableEvent(HAL_USART_Pvt_Events event) {
```
avtolstoy: Nit: static or anonymous namespace, same below

hal/network/lwip/pppservernetif.cpp
```diff
@@ -289,29 +296,39 @@ int PppServerNetif::start() {
     }, nullptr));
     server_->addCommandHandler(AtServerCommandHandler(AtServerCommandType::TEST, "+IFC", "+IFC: (0-2)(0-2)"));
     server_->addCommandHandler(AtServerCommandHandler(AtServerCommandType::READ, "+IFC", [](AtServerRequest* request, AtServerCommandType type, const char* command, void* data) -> int {
-        auto stream = (SerialStream*)data;
-        CHECK_TRUE(stream, SYSTEM_ERROR_INVALID_STATE);
-        if (stream->config() & SERIAL_FLOW_CONTROL_RTS_CTS) {
-            request->sendResponse("+IFC: 2,2");
-        } else {
+        auto self = static_cast<PppServerNetif*>(data);
+        if (self->settings_.usbserial) {
```
avtolstoy: `settings_.usbserial` will be `0xff` if using hardware UART https://github.com/particle-iot/device-os/pull/2898/changes#diff-97131b9bf0a32b593dbb9536fa1af2b12f29457a58570fc08033ecd13ba5ea25R79

hal/network/lwip/pppservernetif.cpp
```diff
@@ -289,29 +296,39 @@ int PppServerNetif::start() {
...
-        if (v[0] == 2 || v[1] == 2) {
-            auto c = stream->config() | SERIAL_FLOW_CONTROL_RTS_CTS;
-            return stream->setConfig(c);
-        } else if (v[0] == 0 || v[1] == 0) {
-            auto c = stream->config() & (~SERIAL_FLOW_CONTROL_RTS_CTS);
-            return stream->setConfig(c);
-        }
+        auto self = static_cast<PppServerNetif*>(data);
+
+        if (self->settings_.serial) {
```
avtolstoy: Same thing here. `serial` will be `0xff` which will pass this check.

avtolstoy (commented):
Main thing: `if_req_ppp_server_serial_settings` `serial` and `usbserial` default value usage should be corrected.

wiring/src/spark_wiring_tether.cpp
scott-brust: Agree, I thought the same but figured we wanted to have an explicit `Tether.bind(TetherUSBConfig())` syntax. Ill consolidate things and just have `Tether.bind(TetherSerialConfig().serial(Serial));`

wiring/src/spark_wiring_tether.cpp
```diff
@@ -26,18 +26,30 @@ TetherClass Tether;
...
 }
 
 USARTSerial& TetherSerialConfig::serial() const {
     return serial_;
 }
 
+USBSerial& TetherSerialConfig::usbserial() const {
+    return usbSerial_;
+}
+
```
scott-brust: Not sure if this is an acceptable solution to having different possible underlying serial stream types

avtolstoy: Ugh, yeah, kinda weird. Let me try a few things if they look better..

scott-brust:
> One outstanding item is how the particle udev linux rules should be updated to allow modem manager to consider the B5SOM tty connection as a modem. Right now the particle cli disallows this by default, but when enabling USB tethering this needs to be disabled. IE this snippet needs to be created
> 
> ```
> sudo tee /etc/udev/rules.d/99-particle-modem.rules << 'EOF'
> SUBSYSTEM=="usb", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_DEVICE_PROCESS}="1"
> SUBSYSTEM=="tty", ATTRS{idVendor}=="2b04", ATTRS{idProduct}=="c019", ENV{ID_MM_DEVICE_IGNORE}="0", ENV{ID_MM_CANDIDATE}="1", ENV{ID_MM_PORT_TYPE_AT_PRIMARY}="1"
> EOF
> ```
> 
> Should we incorporate this into the 'canonical' tethering setup script?

[Updated tethering setup script to incorporate updated rules here](https://github.com/particle-iot-inc/m-hat-firmware/pull/1)

hal/network/lwip/pppservernetif.cpp
```diff
@@ -165,11 +166,17 @@ int PppServerNetif::start() {
...
 
-    auto serial = std::make_unique<SerialStream>((hal_usart_interface_t)settings_.serial, settings_.baud, settings_.config, DEFAULT_SERIAL_BUFFER_SIZE, DEFAULT_SERIAL_BUFFER_SIZE);
-    SPARK_ASSERT(serial);
-    serial_ = std::move(serial);
+    if (settings_.serial != IF_REQ_INVALID_SERIAL_INTERFACE) {
+        auto serial = std::make_unique<SerialStream>((hal_usart_interface_t)settings_.serial, (uint32_t)settings_.baud, (uint32_t)settings_.config, DEFAULT_SERIAL_BUFFER_SIZE, DEFAULT_SERIAL_BUFFER_SIZE);
+        SPARK_ASSERT(serial);
+        serial_ = std::move(serial);
+    } else {
+        auto serial = std::make_unique<SerialUSBStream>((HAL_USB_USART_Serial)settings_.usbserial, (uint32_t)settings_.baud, DEFAULT_SERIAL_BUFFER_SIZE, DEFAULT_SERIAL_BUFFER_SIZE);
```
avtolstoy, line 176: Nit: Might be a good idea to make this controllable with `HAL_PLATFORM_XXX` feature flag. E.g. we might want to enable just UART.

hal/src/nRF52840/usb_hal_cdc.c
```diff
@@ -506,23 +564,27 @@ int usb_uart_init(uint8_t *rx_buf, uint16_t rx_buf_size, uint8_t *tx_buf, uint16
     }
 
     m_usb_instance.mode = USB_MODE_CDC_UART;
+    m_usb_instance.ev_group = xEventGroupCreate();
```
avtolstoy, line 567: assert that it got created

avtolstoy (approved):

hal/network/lwip/pppservernetif.cpp
scott-brust: You're thinking of something analogous to `HAL_PLATFORM_PPP_SERVER_USART` like `HAL_PLATFORM_PPP_SERVER_USB`? 
The goal would be to ifdef out all of the usb serial code for platforms that wont support it? To save flash space, etc?

PR #2907, opened 2026-03-24
Adds compatible Tracker exrtc functions and fixes Gen 4 sleep duration sanity checks

avtolstoy:


hal/src/rtl872x/sleep_hal.cpp
```diff
@@ -463,11 +465,27 @@ class SleepClass {
...
+        if (rtc->ms > SLEEP_INTERNAL_RTC_MAX_SLEEP_TIME) {
+#if !HAL_PLATFORM_EXTERNAL_RTC
+            return SYSTEM_ERROR_NOT_SUPPORTED;
+#else
+            if (mode == HAL_SLEEP_MODE_POWER_OFF) {
+                hal_exrtc_status_t status = {};
+                status.size = sizeof(status);
+                status.version = HAL_EXRTC_API_VERSION;
+                if (hal_exrtc_get_status(HAL_EXRTC_INSTANCE_DEFAULT, &status, nullptr, nullptr) < 0 || !(status.status & HAL_EXRTC_STATUS_READY)) {
+                    return SYSTEM_ERROR_NOT_SUPPORTED;
```
scott-brust, line 483: So the conditions are
1. sleeping longer than internal max time
2. The platform has an external RTC
3. The sleep mode is power off
4. the exrtc is ok and status is ready
Then we can sleep for the long amount of time

scott-brust (approved):

PR #2909, opened 2026-03-30
Add lock around reading logical efuse

XuGuohui:
### Problem
Reading logical efuse sometimes failed even after retries, which results failed to get device ID, mobile secrect and serial number.

### Solution
Acquire `RT_DEV_LOCK_EFUSE` during reading logical efuse. 

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/rtl872x/efuse.cpp
+++ b/hal/src/rtl872x/efuse.cpp
@@ -24,6 +24,7 @@
 #include "check.h"
 
 extern "C" {
+#include "device_lock.h"
 #include "rtl8721d.h"
 #include "rtl8721d_efuse.h"
 }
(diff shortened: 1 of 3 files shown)


hal/src/rtl872x/efuse.cpp
avtolstoy, line 32: We also have readLogicalEfuse in deviceid_hal which is not used for whatever reason. Can we use a single implementation everywhere?

XuGuohui: Yeah. They are basically identical. Should we just delete the one in deviceid_hal?

avtolstoy: Yeap, but I like the idea that's in there: blocking only for the duration of EFUSE_LMAP_READ, so essentially just replace this function with contents of the other one?

XuGuohui: Sounds good.

avtolstoy (approved):

PR #2910, opened 2026-03-31
[msom] PPP Server Serial1 Tethering

scott-brust:
### Problem
Enabling the PPP server interface on MSoM caused a system deadlock. The PPP server's AT command loop uses waitEvent() to block while waiting for serial data, but the RTL872x USART HAL rejected waitEvent() on interrupt-mode UARTs (UART2/UART3) with SYSTEM_ERROR_NOT_SUPPORTED. Since Serial1 (LP_UART) uses interrupt mode, the pppserver thread spun in a tight loop at network priority, permanently starving the lower-priority system thread which was holding the lwIP core lock. NRF52840 is unaffected because its waitEvent() has no such restriction.

### Solution

  - Enable events for RTL872x usart in interrupt mode. 
  - Removed the useInterrupt() guards from waitEvent(), enableEvent(), and disableEvent() in the RTL872x USART HAL so event-based blocking works on all UARTs. The ISR already had the xEventGroupSetBitsFromISR infrastructure for both modes — the event bit signaling was just moved to execute unconditionally rather than only in mode-specific branches.
  - Guarded UART_RXDMACmd(ENABLE) / UART_TXDMACmd(ENABLE) in disableEvent() to avoid enabling DMA on UARTs without DMA channels.
  - Added MSoM platform support: enabled HAL_PLATFORM_PPP_SERVER, created the ps6 netif, added the NAT64 input hook, and bumped lwIP PPP/timeout pool sizes.

### Steps to Test

Run `user/tests/app/tethering/tethering.cpp` test app

### Example App

```
user/tests/app/tethering/tethering.cpp
```

### References

[Mhat docs if using that hardware](https://deploy-preview-119--particle-developer.netlify.app/m-hat/setup-with-raspberry-pi)

---


hal/src/rtl872x/usart_hal.cpp
```diff
@@ -499,29 +498,30 @@ class Usart {
 
     int disableEvent(HAL_USART_Pvt_Events event) {
         CHECK_TRUE(isEnabled(), SYSTEM_ERROR_INVALID_STATE);
-        CHECK_FALSE(useInterrupt(), SYSTEM_ERROR_NOT_SUPPORTED);
 
         auto uartInstance = uartTable_[index_].UARTx;
         if (event & HAL_USART_PVT_EVENT_READABLE) {
             RxLock lk(this);
             UART_INTConfig(uartInstance, (RUART_IER_ERBI | RUART_IER_ELSI | RUART_IER_ETOI), DISABLE);
```
avtolstoy: I think this is wrong in case of useInterrupt().
This should instead be:

```c++
        if (event & HAL_USART_PVT_EVENT_READABLE) {
            if (data() <= 0) {
                RxLock lk(this);
                if (!useInterrupt()) {
                    UART_INTConfig(uartInstance, (RUART_IER_ERBI | RUART_IER_ELSI | RUART_IER_ETOI), ENABLE);
                    UART_RXDMACmd(uartInstance, DISABLE);
                }
                if (!receiving_) {
                    startReceiver();
                }
            } else {
                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
            }
        }
```

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -499,29 +498,30 @@ class Usart {
...
+                UART_RXDMACmd(uartInstance, ENABLE);    
+            }
             xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
         }
 
         if (event & HAL_USART_PVT_EVENT_WRITABLE) {
             TxLock lk(this);
             UART_INTConfig(uartInstance, RUART_IER_ETBEI, DISABLE);
-            UART_TXDMACmd(uartInstance, ENABLE);
+            if (!useInterrupt()) {
```
avtolstoy: Same here:
```c++
            if (space() <= 0) {
                TxLock lk(this);
                xEventGroupClearBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
                if (!useInterrupt()) {
                    UART_INTConfig(uartInstance, RUART_IER_ETBEI, ENABLE);
                    // Temporarily disable TX DMA, otherwise, the TX FIFO won't be empty until all data is transferred by DMA.
                    UART_TXDMACmd(uartInstance, DISABLE);
                }
                if (!transmitting_) {
                    startTransmission();
                }
            } else {
                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_WRITABLE);
            }
```

hal/src/rtl872x/usart_hal.cpp
scott-brust: Ok, `disableEvent()` isnt used anywhere right now 😄 

If there is no readable data, do we want to clear the `HAL_USART_PVT_EVENT_READABLE` event? 
Also, if there *is* data available, do we still want to set the `HAL_USART_PVT_EVENT_READABLE` event when we are disabling these events? I guess im not sure what the use case of this function should be

avtolstoy: The key point: do not modify int config when in interrupt mode. The appropriate interrupts are already enabled/disabled when needed elsewhere. For DMA they are enabled/disable strictly here , the data tx/rx is managed separately.

> If there is no readable data, do we want to clear the HAL_USART_PVT_EVENT_READABLE event?

No, don't clear it, the reader will clear it when waiting on the event group.

> Also, if there is data available, do we still want to set the HAL_USART_PVT_EVENT_READABLE event when we are disabling these events? I guess im not sure what the use case of this function should be

Just set it, the performance impact would be comparable to checking whether the event is enabled.

scott-brust: Ok, to be clear,`UART_TXDMACmd()` and `UART_RXDMACmd()` are only disabled in `disableEvent()` and `enableEvent()` but im guessing thats ok since they will be re-enabled appropriately in the `uartTxRxIntHandler()`

avtolstoy: Looked at the diff, changing to DISABLE for DMACmd is wrong:
https://github.com/particle-iot/device-os/pull/2910/changes#diff-c751354fca72766e3e8cbdb322f83e8c868633507b69f80bdfc25527d8ba8fa2L508
https://github.com/particle-iot/device-os/pull/2910/changes#diff-c751354fca72766e3e8cbdb322f83e8c868633507b69f80bdfc25527d8ba8fa2L515

avtolstoy: The way this works in DMA mode: when we need to wake somebody up on readable/writable - we suspend DMA, use interrupt mode instead and generate readable/writable on first received/transmitted byte. And go back to dma mode.

For interrupt mode: just generate events when appropriate interrupts come in - don't manage interrupts.

hal/src/rtl872x/usart_hal.cpp
```diff
@@ -499,29 +501,43 @@ class Usart {
...
+                RxLock lk(this);
+                if (!useInterrupt()) {
+                    UART_INTConfig(uartInstance, (RUART_IER_ERBI | RUART_IER_ELSI | RUART_IER_ETOI), DISABLE);
+                    UART_RXDMACmd(uartInstance, ENABLE);
+                }
+                if (!receiving_) {
+                    startReceiver();
+                }
+            } else {
+                xEventGroupSetBits(evGroup_, HAL_USART_PVT_EVENT_READABLE);
```
avtolstoy: These changes look a bit weird, I would honestly not touch disableEvent(), and leave it as-is for DMA mode and for interrupt mode just make it no-op?

avtolstoy (approved):

PR #2911, opened 2026-03-31
[scripts] fix rtl872x.tcl flash script to be more reliable

avtolstoy:
Key changes to `rtl872x_flash_write_bin_ext`:
1. Replaced load_image with write_memory for JimTcl >= 0.80:
   - File is read once into memory (`set file_data [read $f]`)
   - Each 16KB chunk is extracted via string range
   - Bytes converted to list and written via `write_memory $data_buf_addr 8 $byte_list`
   - Eliminates temp file creation and the load_image address trick
2. Added per-chunk result checking:
   - After each `wait_response`, checks `mcu_msg_data_mem_addr + 0x00` for `RESULT_OK`
   - Fails fast with specific error message if MCU reports failure
3. Kept `load_image` fallback for older JimTcl (< 0.80):
   - Same logic as before, but with `puts -nonewline` (no newline hack needed)
   - Also includes per-chunk result checking
   - Fixed temp folder for windows
   - Added pid to the temp filenames so that multiple openocd flashers do not interfere with each other
   - Added temp file cleanup
4. Added `JIMTCL_VERSION` to global declarations in the function

The old load_image approach had a per-chunk failure probability of ~2-5% with DAPLink, compounding to very low success rates for large binaries (62 chunks = ~29% success at 2% per-chunk failure). The new write_memory approach:
- Eliminates the complex load_image address arithmetic that may interact poorly with CMSIS-DAP's transfer queuing
- Reads the file once (no repeated 512KB disk reads per 16KB chunk)
- Uses a simpler, more direct code path through OpenOCD

This should significantly improve reliability with DAPLink and also be much faster.

### DIFF
--- a/scripts/rtl872x.tcl
+++ b/scripts/rtl872x.tcl
@@ -123,7 +123,7 @@ proc openocd_print_memory {address length} {
     } else {
         mem2array value 8 $address $length
     }
-    
+
     echo ""
     echo "---------------------------------------------------------"
     echo "  Offset| 00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F"
@@ -286,6 +286,7 @@ proc rtl872x_flash_write_bin_ext {file address auto_erase auto_verify} {
     global MSG_FLASH_DATA_BUF_SIZE
     global RESULT_OK
     global RESULT_ERROR
+    global JIMTCL_VERSION
 
     rtl872x_init
     set size [file size $file]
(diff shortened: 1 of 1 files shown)


technobly (approved):

scripts/rtl872x.tcl
```diff
@@ -322,47 +323,89 @@ proc rtl872x_flash_write_bin_ext {file address auto_erase auto_verify} {
...
+
+                if {[expr {[openocd_read_register [expr {$mcu_msg_data_mem_addr + 0x00}]] != $RESULT_OK}]} {
+                    error "Flash write FAILED at offset $i (address 0x[format %X [expr {$op_file_address + $i}]])"
+            }
+        }
+        for {set i 0} {$i < $file_counts} {incr i} {
+            catch {file delete "$tmp_file_path/${tmp_prefix}_$i.bin"}
         }
     }
+    }
```
technobly: is this an extra closing curly brace?

technobly (approved):

scripts/rtl872x.tcl
avtolstoy: Indent issue. Fixed.

PR #2912, opened 2026-04-03
[msom] PPP Server USB Serial Tethering

scott-brust:
USB Serial tethering for Msom platform

### Steps to Test

Run tethering app

### Example App

```c
#include "application.h"

Serial1LogHandler dbg(115200, LOG_LEVEL_ALL);
// SerialLogHandler dbg(115200, LOG_LEVEL_ALL);

SYSTEM_MODE(SEMI_AUTOMATIC);

void setup() {
    // M-Hat requires this to enable USART communication to SBC
    pinMode(A0, OUTPUT);
    digitalWrite(A0, HIGH);

    // waitUntil(Serial.isConnected);
    // Enable Cellular
    Cellular.on();
    Cellular.connect();
    // Bind Tether interface to Serial1 @ 921600 baudrate with default settings (8n1 + RTS/CTS flow control)
    // Tether.bind(TetherSerialConfig().baudrate(921600).serial(Serial1));

    // Bind Tether to USB CDC interface
    Tether.bind(TetherUSBConfig());

    // Turn on Tether interface and bring it up
    Tether.on();
    Tether.connect();
    Particle.connect();
}

void loop() {

}
```

### References

---

### Completeness

- [ ] Added documentation


hal/src/msom/network/network.cpp
```diff
@@ -195,6 +207,14 @@ int if_init_platform(void*) {
     /* TODO: wl5 - Realtek NCP Access Point */
     (void)wl5;
 
+#if HAL_PLATFORM_PPP_SERVER
+    /* ps6 - PPP Server */
+    ps6 = new PppServerNetif();
+    if (ps6) {
+        ((PppServerNetif*)ps6)->init();
+    }
```
scott-brust, line 215: ps6 / Index 7 should be correct to match `NETWORK_INTERFACE_PPP_SERVER` for msom platform right? 
I think for b5som ps5 / index 6 is wrong?

hal/src/rtl872x/usb_hal.cpp
```diff
@@ -255,6 +262,37 @@ int32_t HAL_USB_USART_Send_Data_protected(HAL_USB_USART_Serial serial, uint8_t d
...
+            if (available > 0 || available < 0) {
+                break;
+            }
+            if (needToYield) {
+                HAL_Delay_Milliseconds(1);
+            }
+        }
+    }
+    if (available > 0 && HAL_USB_USART_Is_Connected(serial)) {
+        return getCdcClassDriver().write(data, size);
```
scott-brust: There is probably a better way to do multiple byte read/writes to the USB CDC driver. I just copied the single byte methods and changed this line

hal/src/rtl872x/usbd_cdc.cpp
```diff
@@ -61,6 +63,7 @@ CdcClassDriver::~CdcClassDriver() {
         txTimer_ = nullptr;
     }
 #endif // !HAL_PLATFORM_USB_SOF
+    vEventGroupDelete(eventGroup_);
```
scott-brust, line 66: Might want to check if eventGroup_ exists, it should here but just to be safe

hal/src/rtl872x/usbd_cdc.cpp
```diff
@@ -615,3 +626,29 @@ int CdcClassDriver::stopTxTimer() {
...
+    return SYSTEM_ERROR_NONE;
+}
+
+int CdcClassDriver::waitEvent(uint32_t events, system_tick_t timeout) {
+    CHECK_FALSE(enableEvent((HAL_USART_Pvt_Events)events), SYSTEM_ERROR_INVALID_STATE);
+    return xEventGroupWaitBits(eventGroup_, events, pdTRUE, pdFALSE, timeout / portTICK_RATE_MS);
+}
+
+int CdcClassDriver::eventGroup(EventGroupHandle_t* eventGroup) {
+    *eventGroup = eventGroup_;
```
scott-brust, line 645: might help to do error checking here too

wiring/src/spark_wiring_tether.cpp
```diff
@@ -95,7 +95,7 @@ int TetherClass::bind(const TetherUSBConfig config) {
         settings.base.type = IF_REQ_DRIVER_SPECIFIC_PPP_SERVER_SERIAL_SETTINGS;
         settings.serial = IF_REQ_INVALID_SERIAL_INTERFACE;
         settings.baud = HAL_PLATFORM_PPP_SERVER_USART_BAUDRATE;
-        settings.usbserial = config.usbserial();
+        settings.usbserial = config.usbserial() == Serial ? HAL_USB_USART_SERIAL : IF_REQ_INVALID_SERIAL_INTERFACE;
```
scott-brust: ill have to confirm if this is working as intended. I was getting `HAL_USB_USART_SERIAL1` due to a conversion from `config.usbserial()`. `Serial` doesnt have a method to return what `HAL_USB_USART_Serial` it is

Also may just want to hardcode this to `HAL_USB_USART_SERIAL`, or check if the platform even supports `HAL_USB_USART_SERIAL1`

hal/src/rtl872x/usbd_cdc.cpp
```diff
@@ -236,9 +239,12 @@ int CdcClassDriver::dataIn(unsigned ep, particle::usbd::EndpointEvent ev, size_t
     txBuffer_.consumeCommit(txBuffer_.consumePending());
     txState_ = false;
 
-
     startTx();
 
+    BaseType_t yield = pdFALSE;
+    if (xEventGroupSetBitsFromISR(eventGroup_, HAL_USART_PVT_EVENT_WRITABLE, &yield) != pdFAIL) {
+        portYIELD_FROM_ISR(yield);
+    }
```
scott-brust: Probably should check that there is still TX space available before signalling writeable?

hal/src/rtl872x/serial_usb_stream.cpp
```diff
@@ -0,0 +1,189 @@
+/*
+ * Copyright (c) 2026 Particle Industries, Inc.  All rights reserved.
```
avtolstoy, line 2: Move to hal/shared?

hal/src/rtl872x/usb_hal.cpp
avtolstoy: Just keep this one and delegate single byte one to it?

hal/src/rtl872x/usbd_cdc.cpp
avtolstoy: Would require acquiring tx lock etc. Leave as-is. consumePending() is definitely > 0 if we got here, so we did free up some space.

wiring/src/spark_wiring_tether.cpp
avtolstoy: We don't have any platforms with USBSerial1 as of right now. But this is not working correctly because you are comparing config.usbserial() with (bool)Serial (so, operator bool). I would add explicit interface() method to usb serial class similar to spark_wiring_usart.h and use that explicitly here.

scott-brust: Yeah figured this was not correct, ill add `interface()` to spark_wiring_usbserial.h

avtolstoy (approved):

PR #2913, opened 2026-04-08
Improve I2C driver reliability

XuGuohui:
### Problem
1. The system power manager references the wiring PMIC and FuelGauge driver, which will construct the `Wire` instance and initialize the TX/RX buffer in the I2C driver. The `Wire` instance in user app is separated instance and if user specify a larger buffer (>32 bytes) for the driver, the I2C will be deinitialized, which may results in corrupting the operation to PMIC and FuelGauge in the system power manager. This is fixed by https://github.com/particle-iot/device-os/pull/2890. The PR herein reverts some of those changes and fix the issue from another perspective.
2. (Gen4 only) `hal_i2c_transaction()` doesn't check if the I2C is enabled before going forward.
3. If the I2C interface that is used by system power manager is ended by `Wire.end()` in user app, the system power manager may return unexpected state.

### Solution
1. (Gen4 only) I2C is not disabled in `deInit()`, we should not change the enable state, but change the configured state only.
2. Checks I2C enable state in `hal_i2c_transaction()`
3.  Make sure the I2C interface is active before interacting with PMIC, FuelGauge and Am18x5 

### Example App
```cpp
#include "application.h"

SYSTEM_MODE(MANUAL);

size_t I2C_BUFFER_SIZE = HAL_PLATFORM_I2C_BUFFER_SIZE(HAL_PLATFORM_PMIC_BQ24195_I2C) + 1;
uint8_t _rx_buf[I2C_BUFFER_SIZE];
uint8_t _tx_buf[I2C_BUFFER_SIZE];
hal_i2c_config_t acquireWireBuffer()
{
    Log.info("acquireWireBuffer() in user app\n");
    hal_i2c_config_t config = {
        .rx_buffer = _rx_buf,
        .rx_buffer_size = I2C_BUFFER_SIZE,
        .tx_buffer = _tx_buf,
        .tx_buffer_size = I2C_BUFFER_SIZE
    };
    return config;
}

system_tick_t now = 0;

void setup() {
    delay(10000);
    now = millis();
}

void loop() {
    FuelGauge fuel(true);
    Log.info("Wire enabled: %d\r", Wire.isEnabled());
    float soc = fuel.getSoC();
    Log.info("%.2f", soc);
    delay(1000);
    if (millis() - now > 10000) {
        Log.info("Disable I2C");
        Wire.end();
        // The getSoC() herein will fail, but the power manager will re-enable the I2C interface
    }
}
```

Sample outputs:
>0000010320 [app] INFO: acquireWireBuffer() in user app
0000010321 [app] INFO: Wire enabled: 1
0000010323 [app] INFO: 87.21
0000011324 [app] INFO: Wire enabled: 1
0000011326 [app] INFO: 87.21
0000012328 [app] INFO: Wire enabled: 1
0000012330 [app] INFO: 87.21
0000013330 [app] INFO: Wire enabled: 1
0000013333 [app] INFO: 87.21
0000014333 [app] INFO: Wire enabled: 1
0000014335 [app] INFO: 87.21
0000015335 [app] INFO: Disable I2C
0000015336 [app] INFO: Wire enabled: 0
0000015336 [app] INFO: -1.00
0000016337 [app] INFO: Wire enabled: 0
0000016337 [app] INFO: -1.00
0000017338 [app] INFO: Wire enabled: 0
0000017338 [app] INFO: -1.00
0000018339 [app] INFO: Wire enabled: 0
0000018339 [app] INFO: -1.00
0000019340 [app] INFO: Wire enabled: 0
0000019340 [app] INFO: -1.00
0000020341 [app] INFO: Disable I2C
0000020341 [app] INFO: Wire enabled: 0
0000020342 [app] INFO: -1.00
0000021348 [app] INFO: Wire enabled: 1              <<< Battery detached
0000021350 [app] INFO: 87.21
0000022355 [app] INFO: Wire enabled: 1
0000022357 [app] INFO: 87.06
0000023359 [app] INFO: Wire enabled: 1
0000023361 [app] INFO: 86.92
0000024362 [app] INFO: Wire enabled: 1
0000024364 [app] INFO: 86.92
0000025364 [app] INFO: Disable I2C
0000025366 [app] INFO: Wire enabled: 1
0000025368 [app] INFO: 86.92
0000026369 [app] INFO: Wire enabled: 1

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -120,6 +120,7 @@ typedef struct nrf5x_i2c_info_t {
     os_mutex_recursive_t        mutex;
     bool                        configured;
     bool                        heap_buffer;
+    bool                        acquired;
 
     void (*callback_on_request)(void);
     void (*callback_on_receive)(int);
@@ -393,6 +394,7 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     i2cMap[i2c].tx_index_head = 0;
     i2cMap[i2c].tx_index_tail = 0;
     i2cMap[i2c].configured = true;
+    i2cMap[i2c].acquired = false;
     memset((void *)i2cMap[i2c].rx_buf, 0, i2cMap[i2c].rx_buf_size);
     memset((void *)i2cMap[i2c].tx_buf, 0, i2cMap[i2c].tx_buf_size);
 
@@ -478,7 +480,7 @@ void hal_i2c_end(hal_i2c_interface_t i2c, void* reserved) {
     }
 
     I2cLock lk(i2c);
-    if (hal_i2c_is_enabled(i2c, nullptr)) {
+    if (hal_i2c_is_enabled(i2c, nullptr) && !i2cMap[i2c].acquired) {
         if (twiUninit(i2c) == SYSTEM_ERROR_NONE) {
             i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;
         }
@@ -673,7 +675,7 @@ void hal_i2c_flush(hal_i2c_interface_t i2c, void* reserved) {
 }
 
 bool hal_i2c_is_enabled(hal_i2c_interface_t i2c,void* reserved) {
-    return i2cMap[i2c].state == HAL_I2C_STATE_ENABLED;
+    return i2cMap[i2c].state == HAL_I2C_STATE_ENABLED && i2cMap[i2c].configured;
 }
 
 void hal_i2c_set_callback_on_received(hal_i2c_interface_t i2c, void (*function)(int),void* reserved) {
@@ -808,4 +810,30 @@ int hal_i2c_sleep(hal_i2c_interface_t i2c, bool sleep, void* reserved) {
 int hal_i2c_transaction(hal_i2c_interface_t i2c, const hal_i2c_transmission_config_t* tx_config, const hal_i2c_transmission_config_t* rx_config, void* reserved) {
     // TODO
     return SYSTEM_ERROR_NOT_SUPPORTED;
+}
+
+int hal_i2c_acquire(hal_i2c_interface_t i2c, void* reserved) {
+    if (i2c >= HAL_PLATFORM_I2C_NUM) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    I2cLock lk(i2c);
+    i2cMap[i2c].acquired = true;
+    return SYSTEM_ERROR_NONE;
+}
+
+int hal_i2c_release(hal_i2c_interface_t i2c, void* reserved) {
+    if (i2c >= HAL_PLATFORM_I2C_NUM) {
+        return SYSTEM_ERROR_INVALID_ARGUMENT;
+    }
+    I2cLock lk(i2c);
+    i2cMap[i2c].acquired = false;
+    return SYSTEM_ERROR_NONE;
+}
+
+bool hal_i2c_is_acquired(hal_i2c_interface_t i2c, void* reserved) {
+    if (i2c >= HAL_PLATFORM_I2C_NUM) {
+        return false;
+    }
+    I2cLock lk(i2c);
+    return i2cMap[i2c].acquired;
 }
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -118,7 +118,6 @@ typedef struct nrf5x_i2c_info_t {
     size_t                      tx_index_tail;
 
     os_mutex_recursive_t        mutex;
-    bool                        configured;
     bool                        heap_buffer;
 
     void (*callback_on_request)(void);
@@ -254,6 +253,8 @@ static int twiUninit(hal_i2c_interface_t i2c, bool reset_pin_configuration = tru
         hal_pin_set_function(i2cMap[i2c].sda_pin, PF_NONE);
     }
 
+    i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;
+
     return SYSTEM_ERROR_NONE;
 }
 
@@ -344,7 +345,7 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     }
 #endif // HAL_PLATFORM_PMIC_BQ24195
 
-    if (i2cMap[i2c].configured) {
+    if (i2cMap[i2c].state != HAL_I2C_STATE_NOT_INITIALIZED) {
         // Configured, but new buffers are invalid
         if (!isConfigValid(config)){
             return SYSTEM_ERROR_INVALID_ARGUMENT;
@@ -358,7 +359,7 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
         if (i2cMap[i2c].heap_buffer) {
             free(i2cMap[i2c].rx_buf);
             free(i2cMap[i2c].tx_buf); 
-            i2cMap[i2c].configured = false;
+            // We've acquired lock, it's safe to not change the state here.
         }
     }
 
@@ -392,7 +393,6 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     i2cMap[i2c].rx_index_tail = 0;
     i2cMap[i2c].tx_index_head = 0;
     i2cMap[i2c].tx_index_tail = 0;
-    i2cMap[i2c].configured = true;
     memset((void *)i2cMap[i2c].rx_buf, 0, i2cMap[i2c].rx_buf_size);
     memset((void *)i2cMap[i2c].tx_buf, 0, i2cMap[i2c].tx_buf_size);
 
@@ -425,7 +425,7 @@ void hal_i2c_begin(hal_i2c_interface_t i2c, hal_i2c_mode_t mode, uint8_t address
     if (i2c >= HAL_PLATFORM_I2C_NUM) {
         return;
     }
-    if (!i2cMap[i2c].configured) {
+    if (i2cMap[i2c].state == HAL_I2C_STATE_NOT_INITIALIZED) {
         return;
     }
 
@@ -454,9 +454,7 @@ void hal_i2c_begin(hal_i2c_interface_t i2c, hal_i2c_mode_t mode, uint8_t address
     }
 #endif // PLATFORM_ID == PLATFORM_TRACKER
 
-    if (twiUninit(i2c) == SYSTEM_ERROR_NONE) {
-        i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;
-    }
+    twiUninit(i2c);
 
     if (i2cMap[i2c].state == HAL_I2C_STATE_DISABLED) {
         i2cMap[i2c].rx_index_head = 0;
@@ -482,9 +480,7 @@ void hal_i2c_end(hal_i2c_interface_t i2c, void* reserved) {
 
     I2cLock lk(i2c);
     if (hal_i2c_is_enabled(i2c, nullptr)) {
-        if (twiUninit(i2c) == SYSTEM_ERROR_NONE) {
-            i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;
-        }
+        twiUninit(i2c);
     }
 }
 
@@ -676,7 +672,7 @@ void hal_i2c_flush(hal_i2c_interface_t i2c, void* reserved) {
 }
 
 bool hal_i2c_is_enabled(hal_i2c_interface_t i2c,void* reserved) {
-    return i2cMap[i2c].state == HAL_I2C_STATE_ENABLED && i2cMap[i2c].configured;
+    return i2cMap[i2c].state == HAL_I2C_STATE_ENABLED;
 }
 
 void hal_i2c_set_callback_on_received(hal_i2c_interface_t i2c, void (*function)(int),void* reserved) {
@@ -712,9 +708,7 @@ int hal_i2c_reset(hal_i2c_interface_t i2c, uint32_t reserved, void* reserved1) {
     }
 
     // Important: we keep GPIO configuration intact
-    if (twiUninit(i2c, false) == SYSTEM_ERROR_NONE) {
-        i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;
-    }
+    twiUninit(i2c, false);
 
     // Just in case make sure that the pins are correctly configured (they should anyway be at this point)
     hal_gpio_config_t conf = {
--- a/hal/src/nRF52840/i2c_hal.cpp
+++ b/hal/src/nRF52840/i2c_hal.cpp
@@ -120,7 +120,6 @@ typedef struct nrf5x_i2c_info_t {
     os_mutex_recursive_t        mutex;
     bool                        configured;
     bool                        heap_buffer;
-    hal_i2c_acquire_flag_t      acquire_flag;
 
     void (*callback_on_request)(void);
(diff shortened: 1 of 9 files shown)


hal/src/nRF52840/i2c_hal.cpp
technobly, line 346: Is this still required? Maybe it is because `i2cMap[i2c].state = HAL_I2C_STATE_DISABLED;` is in hal_i2c_init() below?  Vs. in Gen 4 init() doesn't do this, and deInit() in gen 4 now has `state_ = HAL_I2C_STATE_DISABLED;` removed?

technobly, line 406: Is this still required? Maybe it is because i2cMap[i2c].state = HAL_I2C_STATE_DISABLED; is in hal_i2c_init() above?  Vs. in Gen 4 init() doesn't do this, and deInit() in gen 4 now has state_ = HAL_I2C_STATE_DISABLED; removed?  Furthermore, Gen 3 hal_i2c_init() appears to reset a bunch of the config to a default state, whereas Gen 4 does not do that.  I'm not sure if this can be changed without breaking existing implementation, but if so perhaps we should make it more like Gen 4?  Or if it's easier just to leave the existing fix, that's probably ok as well.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -393,6 +394,7 @@ int hal_i2c_init(hal_i2c_interface_t i2c, const hal_i2c_config_t* config) {
     i2cMap[i2c].tx_index_head = 0;
     i2cMap[i2c].tx_index_tail = 0;
     i2cMap[i2c].configured = true;
+    i2cMap[i2c].acquired = false;
```
technobly: This shouldn't have any affect on already implemented user code because `acquired` defaults to `false`, correct?

hal/src/nRF52840/i2c_hal.cpp
XuGuohui: It's required, since I2C peripheral is disabled in this process. But on Gen4, we just re-configure the buffer in this function, so we should keep the `state_` as-is.

XuGuohui: Ideally the init function should configure the buffer only, cos there is the `begin()` API for configuring the I2C peripheral and the init function is invoked in `Wire` constructor, configuring I2C peripheral before `begin()` seams like a race condition. I'd prefer leave the implementation of the init function for each generation as-is, since their I2C behavior is different. Changing the logic may introduce problem. We can revisit it if new issue is uncovered.

hal/src/nRF52840/i2c_hal.cpp
XuGuohui: Correct. Currently the I2C is only acquired by system power manager if enabled. User app can also acquire the I2C, in case that multiple kind of I2C devices are used and prevent one driver from disabling the other drivers inadvertently.

XuGuohui:
@avtolstoy Now that the Am1805 driver become part of the DVOS, we should also use `hal_i2c_transaction()` in the driver to prevent I2C transaction from being corrupted by other threads. What's your opinion?

XuGuohui:
> @avtolstoy Now that the Am1805 driver become part of the DVOS, we should also use `hal_i2c_transaction()` in the driver to prevent I2C transaction from being corrupted by other threads. What's your opinion?

False alert. The Am18x5 driver is using RAII to manipulate the HAL I2C lock, so it doesn't have threading safe issue. But `hal_i2c_acquire()/release()` are necessary so that the I2C in use won't be disabled advertently. A counter should be maintained in `hal_i2c_acquire()/release()`.

technobly:
@XuGuohui I will do some testing before I approve, please give me a little time.

technobly:
@XuGuohui I tested in a SoM Eval Board and `msom` and it works great.  The `Wire.end()` does not disable the PMIC/FuelGuage I2C driver.  When testing on `b5som` though, the previous fix still patches up the FuelGauge after initial 10s, but when explicitly attempting to terminate with `Wire.end()` it does appear to disable the interface and SoC returns with -1.0.  If I add `fuel.begin()` to the loop, the interface it re-enabled properly each time through.  So something is still not quite right for Gen 3, at least with the new addition of `.acquired` to prevent `Wire.end()` from disabling the interface.  Here's a slightly modified version of the app you shared that I used.

```cpp
#include "Particle.h"

SYSTEM_MODE(MANUAL);

STARTUP(System.enable(SYSTEM_FLAG_PM_DETECTION));

#define I2C_BUFFER_SIZE (32 + 1)
uint8_t _rx_buf[I2C_BUFFER_SIZE];
uint8_t _tx_buf[I2C_BUFFER_SIZE];
hal_i2c_config_t acquireWireBuffer()
{
    Log.info("acquireWireBuffer() in user app\n");
    hal_i2c_config_t config = {
        .rx_buffer = _rx_buf,
        .rx_buffer_size = I2C_BUFFER_SIZE,
        .tx_buffer = _tx_buf,
        .tx_buffer_size = I2C_BUFFER_SIZE
    };
    return config;
}

system_tick_t now = 0;

void setup() {
    delay(10000);
    now = millis();
}

void loop() {
    FuelGauge fuel(true);
    // fuel.begin(); // should work without this call
    Log.info("Wire enabled: %d\r", Wire.isEnabled());
    float soc = fuel.getSoC();
    Log.info("%.2f", soc);
    delay(1000);
    if (millis() - now > 10000) {
        Log.info("Disable I2C");
        Wire.end();
        now = millis();
    }
}
```

XuGuohui:
@technobly Fixed for Gen3 and added protection for Am18x5 driver as well.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -478,7 +481,7 @@ void hal_i2c_end(hal_i2c_interface_t i2c, void* reserved) {
     }
 
     I2cLock lk(i2c);
-    if (hal_i2c_is_enabled(i2c, nullptr)) {
+    if (hal_i2c_is_enabled(i2c, nullptr) && i2cMap[i2c].acquire_flag == HAL_I2C_ACQUIRE_NONE) {
```
avtolstoy: I'll be frank - I think gating `hal_i2c_end()` is not the answer. The fixes for `hal_i2c_is_enabled()` and lock race condition are great - that's a real issue, but this API for acquire/release based on some specific I2C peripherals is not ideal - it's good to have this (perhaps in a refcount form) for understanding whether peripheral is used or not and for example once released powering it down, but not for this specific use case/problem.

(Also as far as power manager goes - we do have external locking so the is_enabled + lock issues shouldn't apply there as well?)

You also have exactly the same problem with SPI (MCP23S17/Ethernet), GPIO (a bunch of things), UART (Tethering at the very least?), USB or any other peripheral shared by the system and application. An ad-hoc solution for just I2C here is not the answer.

From my perspective we either leave things as is, or if absolutely required in power manager use a pattern where we ensure that the underlying peripheral is configured (e.g. combined lock + enabled check + begin RAII object).

XuGuohui: The introduction of the acquire/release API intends to avoid the power manager and the external RTC being corrupted if `Wire.end()` is invoked in user application. If this is kind of the user's responsibility to watch out this, we can leave things as-is. The is_enabled issue caused by acquiring a larger buffer doesn't lean on this APIs.

avtolstoy: 1. Why is app calling Wire.end()?
2. If we want to make sure that underlying i2c interface is enabled in power manager - my comment on RAII object ensuring that still stands.

Can we please also add tests that will reproduce now on 6.4.0 and will not after this change? I guess part of my feedback is that I don't understand exactly what conditions we are trying to solve the problem against. Makes sense?

XuGuohui: 1. We don't know whether third-party I2C peripheral drivers will do it or not.
2. I was thinking conditionally end the I2C peripheral is a shorter path, instead of going through the `end()` -> `begin()` process if the case happened.
3. The issue now can be:
    1) `Wire.end()` will corrupted PMIC, FuelGauge and Am18x5 APIs
    2) `hal_i2c_transaction()` will accidently enable I2C peripheral if disabled, since `reEnableIfNeeded()` is called in `endTransmission()`, cause sending unexpected data on the bus.

avtolstoy: > Wire.end() will corrupted PMIC, FuelGauge and Am18x5 APIs

How? Power manager should be locking throughout the whole transaction if it's in several steps e.g. read/write

XuGuohui: `Wire.end()` will disable the I2C peripheral. All the mentioned I2C peripherals used by system are accessed by APIs without `begin` first.

XuGuohui: If ended, HAL I2C APIs will return error, cos `hal_i2c_is_enabled()` is `false`.

avtolstoy: It's not corruption though? If we are not checking errors on querying/writing to pmic/fuel gauge - that's a separate problem.
Re-enabling underlying i2c interface is a separate concern, we can make sure that power manager does that.

All I'm trying to say: let's not mix problems here.

XuGuohui: Sorry, I was not stating it clear. I meant the `Wire.end()` will make the I2C communications to those built-in peripherals corrupted, not the APIs themselves.

avtolstoy: And let's solve that problem at higher level - power manager. Either we make sure to bring up the i2c peripheral - or we just handle the errors properly.

XuGuohui: Same to the Am18x5 driver?

avtolstoy: Yes. If we care about power manager and am18x5 both working without caring about the state of the underlying i2c interface.

hal/src/nRF52840/i2c_hal.cpp
```diff
@@ -425,6 +425,9 @@ void hal_i2c_begin(hal_i2c_interface_t i2c, hal_i2c_mode_t mode, uint8_t address
     if (i2c >= HAL_PLATFORM_I2C_NUM) {
         return;
     }
+    if (!i2cMap[i2c].configured) {
```
avtolstoy: this is new

XuGuohui: yup. make sense?

wiring/src/spark_wiring_power.cpp
```diff
@@ -1154,6 +1154,11 @@ byte PMIC::getVersion() {
  *******************************************************************************/
 byte PMIC::readRegister(byte startAddress) {
     std::lock_guard<PMIC> l(*this);
+    if (!pmicWireInstance()->isEnabled()) {
```
avtolstoy: Let's do this outside of PMIC and keep these as-is. Same for fuelgauge. In power manager just make sure we are holding the i2c lock while interacting with pmic/fuelgauge.

hal/src/rtl872x/i2c_hal.cpp
```diff
@@ -137,11 +137,11 @@ class I2cClass {
             heapBuffer_ = false;
         }
         configured_ = false;
-        state_ = HAL_I2C_STATE_DISABLED;
```
avtolstoy, line 140: Can we maybe clean this up and either use enum state or bool state flags? It's very confusing right now even looking at this diff

wiring/src/spark_wiring_power.cpp
XuGuohui: Abort the changes that try enabling the I2C interface if disabled?

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: Sorry, I didn't get it. The `state_` is now a enum state already.

wiring/src/spark_wiring_power.cpp
avtolstoy: no, you would still do that, just outside of PMIC and FuelGauge classes, similar to what you do in AM18x5?

hal/src/rtl872x/i2c_hal.cpp
avtolstoy: you have state_ and separate configured_ sprinkled around as well, can we clean that up? you can't normally have state ENABLED and configured_ false?

wiring/src/spark_wiring_power.cpp
XuGuohui: There are a bunch of independent PMIC and Fuelgauge objects in the power manager, it's not ideal to add the check after where the objects are constructed.

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: Got it. Will see how to make sure the state is consistent without race condition.

wiring/src/spark_wiring_power.cpp
avtolstoy: You can use C++ composition or inheritance and just wrap PMIC and FuelGauge classes? That way you can have RAII locking + ensuring i2c is enabled? Instead of PMIC blah; you would simply use EnsurePMIC blah; or something like that>

hal/src/rtl872x/i2c_hal.cpp
XuGuohui: Updated.

XuGuohui:
Updated the PR's description.

system/src/system_power_manager.cpp
```diff
@@ -44,6 +44,26 @@ using namespace particle::power;
 
 namespace {
 
+class PmicGuard : public PMIC {
+public:
+  PmicGuard(bool lock = false) : PMIC(lock) {
+    if (!hal_i2c_is_enabled(HAL_PLATFORM_PMIC_BQ24195_I2C, nullptr)) {
+      hal_i2c_begin(HAL_PLATFORM_PMIC_BQ24195_I2C, I2C_MODE_MASTER, 0x00, nullptr);
+      SPARK_ASSERT(hal_i2c_is_enabled(HAL_PLATFORM_PMIC_BQ24195_I2C, nullptr));
```
avtolstoy, line 52: We are not locking in some places, so you have a race condition with begin and this check. If you are trying to be protective against app or something else disabling I2C - you need to always be locking. And SPARK_ASSERT will just hit and reset us, which is not great..

avtolstoy: I thnk we should revise power manager and make sure that we are safely locking while by default while performing PMIC/FG operations.

XuGuohui: Can we just make it `PmicGuard() : PMIC(true) {` and the same to FuelGauge? If we remove the assertion, there is no way to handle the error. Some operations may fail without notifying users.

avtolstoy: We can - yes. `SPARK_ASSERT` in that case is justified, not right now. The problem is that we need to be careful and after making this change review each and every place PmicGuard/FuelGaugeGuard are used to avoid deadlocks, unnecessarily blocking for significant about of time etc. This is highest priority thread, so this will affect things if there are any issues.

XuGuohui: Always acquiring the I2C lock is not ideal, as you pointed out that the power manager is running on critical priority and if user app has acquired the lock, that will definitely block the thread for unpredictable amount of time. Going back to the initial intention, we just want to make sure the I2C is not disabled before operating PMIC/FuelGauge. We can probably always acquire the lock ONLY if `hal_i2c_is_enabled()` is `false` in the `PmicGuard` and `FuelGaugeGuard`, which can significantly avoid race condition that is caused by user app.

avtolstoy (approved):

PR #2914, opened 2026-04-14
Support detecting/configuring M-HAT in Tinker

XuGuohui:
### Problem
The factory default user app, i.e. the Tinker firmware, running on Particle M.2 SoMs that is installed on M-HAT may not work as expected, since some of the pins for controlling the power module are not correct.

### Steps to Test
1. Build and flash the tinker app to a Particle M.2 SoM
2. Update the `DVOS/user/tests/app/usb_ctrl_request/test/test.js` with the following content
3. Follow the `DVOS/user/tests/app/usb_ctrl_request/README.md` to run the test
```js
import * as control from './support/control';
import * as proto from './support/proto';
import * as usb from './support/usb';
import './support/init';

import * as semver from 'semver';
import * as randomstring from 'randomstring';

import { expect } from 'chai';

describe('USB requests', function() {
  let dev = null;
  let detectedBoard = null;
  const supportedBoards = ['none', 'muon', 'm-hat'];

  before('find a Particle USB device', function() {
    return usb.enumDevices().then((devs) => {
      if (devs.length == 0) {
        throw new Error('No Particle USB devices found');
      }
      dev = devs[0]; // Use the first device in the list
    });
  });

  before('open device', function() {
    return dev.open();
  });

  after('close device', function() {
    if (dev) {
      return dev.close();
    }
  });

  describe('application control requests', function() {
    this.slow(2000); // Test application replies with a random delay

    it('identify base board should succeed', function() {
      const reqData = '{"BASE_BOARD_DETECT":null}';
      console.log(reqData);
      return control.sendRequest(dev, control.RequestType.APP_CUSTOM, reqData).then((rep) => {
        expect(rep.result).to.equal(control.Result.OK);
        const data = JSON.parse(rep.data);
        console.log(data);
        expect(data).to.have.property('board');
        expect(data.board).to.be.oneOf(supportedBoards);
        detectedBoard = data.board;
      });
    });

    it('configure base board should succeed', function() {
        expect(detectedBoard).to.be.oneOf(supportedBoards);
        const reqData = JSON.stringify({ CONFIGURE_MODULE_BOARD: detectedBoard });
        console.log(reqData);
        return control.sendRequest(dev, control.RequestType.APP_CUSTOM, reqData).should.become({
          result: control.Result.OK,
          data: '{"status":0}'
        });
      });
  });
});

```

### Example App
`make PLATFORM=msom all APP=tinker`

### References
N/A

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/user/applications/tinker/src/board_config.cpp
+++ b/user/applications/tinker/src/board_config.cpp
@@ -81,8 +81,9 @@ void BoardConfig::detectBaseBoard() {
         {ATSHA204A, 0x64},
         {LORA, 0x61},
     };
-    constexpr uint16_t muonI2cDevices = STUSB4500 | AM1805 | TMP112A | BQ24195 | MAX17043 | LORA;
-    constexpr uint16_t mhatI2cDevices = STUSB4500 | AM1805 | TMP112A | BQ24195 | MAX17043 | ATSHA204A;
+    constexpr uint16_t essentialI2cDevices = STUSB4500 | AM1805 | TMP112A | BQ24195 | MAX17043;
+    constexpr uint16_t muonI2cDevices = essentialI2cDevices | LORA;
+    constexpr uint16_t mhatI2cDevices = essentialI2cDevices | ATSHA204A;
 
     Wire.lock();
     Wire.begin();
@@ -91,6 +92,20 @@ void BoardConfig::detectBaseBoard() {
         Wire.beginTransmission(addrs[i].second);
         if (Wire.endTransmission() == 0) {
             i2cDeviceMask |= addrs[i].first;
+        }
+    }
+    // If swapping from "none" to "muon", auxiliary power is not enabled by default.
+    // Thus, the LoRa chip is not detected in the first scan.
+    if (i2cDeviceMask == essentialI2cDevices) {
+        PinMode mode = getPinMode(D7);
+        pinMode(D7, OUTPUT);
+        digitalWrite(D7, HIGH); // It's active HIGH on Muon
+        delay(350); // Wait for the LoRa chip to power on, 200ms is estimated to be not sufficient
+        Wire.beginTransmission(0x61); // Lora I2C address
+        if (Wire.endTransmission() == 0) {
+            i2cDeviceMask |= LORA;
+        } else {
+            pinMode(D7, mode);
         }
     }
     Wire.unlock();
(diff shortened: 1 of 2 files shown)


user/applications/tinker/src/board_config.h
```diff
@@ -35,13 +35,23 @@ class BoardConfig {
     size_t replySize();
     
 private:
+    enum __attribute__((packed)) I2cDevice {
```
avtolstoy: No need for packed, this is just a single enum, nothing to 1-byte align/pack.

avtolstoy (approved):

monkbroc (approved):
I updated setup web with this custom tinker. I ran setup on a B504 plugged into a M-HAT. It detected the M-HAT and performed the board configuration.

user/applications/tinker/src/board_config.cpp
```diff
@@ -72,10 +72,34 @@ size_t BoardConfig::replySize() {
...
+    constexpr std::pair<I2cDevice, uint8_t> addrs[] = {
+        {STUSB4500, 0x28},
+        {AM1805, 0x69},
+        {TMP112A, 0x48},
+        {BQ24195, 0x6B},
+        {MAX17043, 0x36},
+        {ATSHA204A, 0x64},
+        {LORA, 0x61},
+    };
+    constexpr uint16_t muonI2cDevices = STUSB4500 | AM1805 | TMP112A | BQ24195 | MAX17043 | LORA;
```
monkbroc: This version of tinker no longer detects my Muon v1.0
In the previous code, it looked for 5 I2C devices:
```
    constexpr uint8_t addrs[] = {
        0x28,   // STUSB4500 USB PD chip
        0x69,   // AM18x5 RTC
        0x48,   // TMP112A temperature sensor
        0x6B,   // BQ24195 PMIC
        0x36    // MAX17043 fuel gauge
    };
```
It now looks for 6 devices. The LORA at 0x61 is new and my Muon doesn't report this one.

Please verify this PR on a Muon v1.0 before merging this PR.

monkbroc (changes requested):

user/applications/tinker/src/board_config.cpp
XuGuohui: I was also using a Muon v1.0 to verify the changes and it worked well. I'm suspecting that you're using a EVT version that doesn't have the Lora firmware programmed?

user/applications/tinker/src/board_config.h
XuGuohui: Fixed.

user/applications/tinker/src/board_config.cpp
XuGuohui: It’s just that when the board type is `none` , the auxiliary power is not enabled by default. When mount it onto Muon the Lora is not powered, thus, the Lora is not detected. Fixed in the latest commit.

monkbroc (approved):
Tested on B504e on M-HAT 👍

PR #2915, opened 2026-04-16
API for sending APDUs

sergeuz:
This PR adds a system and control request APIs for sending command APDUs to the UICC.

#### Steps to test

Using [particle-usb/#126](https://github.com/particle-iot/particle-usb/pull/126):
```js
#!/usr/bin/env node

const { getDevices } = require('particle-usb');
const { setTimeout: delay } = require('node:timers/promises');

async function sendApdu(dev, cmd) {
	console.log('->', cmd.toString('hex'));
	const resp = await dev.sendApdu(cmd);
	console.log('<-', resp.toString('hex'));
	return resp;
}

async function run() {
	let dev;
	try {
		const devs = await getDevices();
		if (!devs.length) {
			throw new Error('No devices found');
		}
		await devs[0].open();
		dev = devs[0];

		console.log('Open channel');
		await sendApdu(dev, Buffer.from('0070000001', 'hex'));

		console.log('Select ISD-R');
		await sendApdu(dev, Buffer.from('01a4040010a0000005591010ffffffff890000010000', 'hex'));

		console.log('Close channel');
		await sendApdu(dev, Buffer.from('0070800100', 'hex'));

	} catch (err) {
		console.error(err);
	} finally {
		await dev.close();
	}
}

run();
```


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1439,6 +1449,12 @@ int QuectelNcpClient::initReady(ModemState state) {
         CHECK(setupBands());
     }
 
+    // Make sure all normal (non-extended) APDU channels are closed. Interestingly, the first
+    // channel seems to be always open after a power on
+    for (int i = 1; i <= 3; ++i) {
+        closeApduChannel(i);
+    }
```
sergeuz, line 1457: @technobly This is to ensure that we always have some APDU channels available for use when the device is reset and we lose the info about the modem/SIM state. However, on my M635 I noticed that the channel 1 is always open at this point, even after the modem has been power cycled, as if the modem itself opens it during the initialization. Do you think there going to be any problem if we close that channel here?

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -2113,6 +2129,149 @@ int QuectelNcpClient::startNcpFwUpdate(bool update) {
...
+    cmd.print(strBuf);
+    cmd.print("\"");
+
+    auto resp = cmd.send();
+    // Not using AtResponse::scanf() as it may allocate memory dynamically
+    CHECK_PARSER(resp.readLine(strBuf, sizeof(strBuf)));
+    CHECK_PARSER_OK(resp.readResult());
+
+    int respHexSize = 0;
+    int r = std::sscanf(strBuf, "+CSIM: %d", &respHexSize);
```
monkbroc, line 2184: Is it possible for responses from other commands to be sent back before +CSIM? Maybe some other modem URCs.

monkbroc (approved):
Looks good from the requirements side. I'll let Brett review from the Device OS cellular side.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
sergeuz: No, the AT parser takes care of separating interactive responses from URCs.

avtolstoy: This is modem-specific, but for ublox/quectel modules we use after 'A' (from AT) is echoed back we are not supposed to receive any URCs they are deferred until the command completes. We would otherwise have a lot of problems dealing with command responses vs URCs. This is not universal across vendors, but is expected in our use case -V250 standard also leaves it up to the vendor, but most vendors agree it's a bad idea to do otherwise (some ublox exceptions exist with cutoff being the \n from the command, not the first echoed attention character)

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
avtolstoy: I would propose we keep first channel open (reserved) and use only the next channels for our purposes.

hal/network/ncp_client/quectel/quectel_ncp_client.h
```diff
@@ -119,6 +121,8 @@ class QuectelNcpClient: public CellularNcpClient {
     volatile bool inFlowControl_ = false;
     bool checkImsi_ = false;
     unsigned int fwVersion_ = 0;
+    system::SystemTimer apduChannelTimer_;
```
avtolstoy, line 124: TIL. Didn't know we had this. freertos timer -> ISR task queue.

avtolstoy (approved):

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
sergeuz: Fixed.

sergeuz: Interestingly, the modem still opens the first channel, which is supposedly already open, when it receives an "open channel" command that doesn't specify a channel number explicitly :shrug:

technobly: I ran into the same thing on Tachyon, and only attempt to close channels 2-4.

PR #2916, opened 2026-04-21
Add netif specific cloud connection keepalives

scott-brust:
### Problem

Cloud connection timeouts are hard coded to 23 minutes for cellular and 25 seconds for wifi for ACM enabled platforms, (ie MSOM). This is overrideable with `Particle.keepAlive()` but will not distinguish between network interfaces. 

### Solution
Implement environment variables to specify what cloud timeouts should be used when connecting on each possible network interface. 
The order of precedence is
1. Per interface env var, ie `PARTICLE_CLOUD_KEEP_ALIVE_CELLULAR`, `PARTICLE_CLOUD_KEEP_ALIVE_WIFI` etc.
2. Global cloud interface keep alive `PARTICLE_CLOUD_KEEP_ALIVE`. This value is used regardless of network interface
3. If no env vars are specified, default to the current behavior where cellular is fixed 23 mins, all others are 25 seconds

If a user specifies a custom keep alive with `Particle.keepAlive()` it will override any of the current above settings. This will not be persistent across resets. 

All env var values are in seconds. 

### Steps to Test

1. Define following env vars for per-interface testing. Observe the values are respected when connecting
`PARTICLE_ETHERNET_CLOUD_KEEP_ALIVE`
`PARTICLE_CELLULAR_CLOUD_KEEP_ALIVE`
`PARTICLE_WIFI_CLOUD_KEEP_ALIVE` 

2. Delete the interface specific vars. Define the global default `PARTICLE_CLOUD_KEEP_ALIVE` value. Observe this value is used when opening a cloud connection

3. Delete all env vars and confirm defaults are used. Cellular = 23 mins, all others 25 seconds 

4. Use `Particle.keepAlive()` and confirm that the specified ping interval takes precedence.  

On device tests with Muon + MSOM
```
device-os-test --device-os-dir=/Users/scottbrust/develop/device-os run msom 00_before system/cellular_env system/env -- -fixture


  00_before: Preparation/Setup for HIL Testing
    msom
      systemThread=enabled
        ✓ 01_erase_factory_module (5.2s)
        ✓ 02_remove_static_ip
        ✓ 03_enable_listening_mode
mailbox msg { t: 1 }
Product ID: 22759
        ✓ 04_clear_env (3.9s)
spark/flash/status started
spark/flash/status success
        ✓ 05_restore_cloud_after_env_clear (6.9s)
mailbox msg { t: 1 }
        ✓ 06_finalize_env_clear (3.9s)
mailbox msg { t: 1 }
        ✓ 07_disable_external_rtc (3.8s)
        ✓ 08_verify_external_rtc_default_state
mailbox msg { t: 2, d: 'muon=true' }
Muon detected: yes
        ✓ 09_report_muon_presence
mailbox msg { t: 1 }
        ✓ 10_configure_muon_board_and_exrtc (3.8s)
        ✓ 11_verify_muon_exrtc_configuration

  Cellular env vars
    msom
      systemThread=disabled
PARTICLE_CELLULAR_PREFERRED_BANDS=100000000000000f0e189c
        ✓ 1_particle_cellular_preferred_bands_init (9.2s)
Default LTE band mask (QUECTEL): 100002000000000f0e189f
        ✓ 2_particle_cellular_preferred_bands_default (6.2s)
LTE band mask after PREFERRED_BANDS env var (QUECTEL): 100000000000000f0e189c
        ✓ 3_particle_cellular_preferred_bands_set (1.9s)
PARTICLE_CELLULAR_FORBIDDEN_BANDS=ffffffffffeffffffffffffff0f1e763
        ✓ 4_particle_cellular_forbidden_bands_init (4.5s)
Default LTE band mask (QUECTEL): 100002000000000f0e189f
        ✓ 5_particle_cellular_forbidden_bands_default (27.7s)
LTE band mask after FORBIDDEN_BANDS env var (QUECTEL): 100000000000000f0e189c
        ✓ 6_particle_cellular_forbidden_bands_set (2.4s)
        ✓ 7_particle_cellular_preferred_plmn_init (4s)
Default CPOL response: empty
        ✓ 8_particle_cellular_preferred_plmn_default (18s)
AT+CPOL response after PREFERRED_PLMN env var:
+CPOL: 1,2,"310410",0,0,0,1

+CPOL: 2,2,"310260",0,0,0,1

+CPOL: 3,2,"311480",0,0,0,1

        ✓ 9_particle_cellular_preferred_plmn_set (6.4s)
        ✓ 10_particle_cellular_keepalive
        ✓ 11_particle_cellular_preferred_plmn_cleanup
AT+CPOL response after PREFERRED_PLMN env var:
+CPOL: 1,2,"310410",0,0,0,1

+CPOL: 2,2,"310260",0,0,0,1

+CPOL: 3,2,"311480",0,0,0,1

        ✓ 12_particle_cellular_env_vars_cleared_verify_defaults
        ✓ 97_cleanup (8.2s)
spark/flash/status started
spark/flash/status success
        ✓ 98_cleanup (6s)
        ✓ 99_cleanup_1 (4.5s)
        ✓ 99_cleanup_2 (8.7s)
        ✓ 10_particle_cellular_preferred_plmn_cleanup (3.6s)
        ✓ 11_particle_cellular_env_vars_cleared_verify_defaults (2.4s)
      systemThread=enabled
PARTICLE_CELLULAR_PREFERRED_BANDS=100000000000000f0e189c
        ✓ 1_particle_cellular_preferred_bands_init (10s)
Default LTE band mask (QUECTEL): 100002000000000f0e189f
        ✓ 2_particle_cellular_preferred_bands_default (6.3s)
LTE band mask after PREFERRED_BANDS env var (QUECTEL): 100000000000000f0e189c
        ✓ 3_particle_cellular_preferred_bands_set (1.1s)
PARTICLE_CELLULAR_FORBIDDEN_BANDS=ffffffffffeffffffffffffff0f1e763
        ✓ 4_particle_cellular_forbidden_bands_init (4.7s)
Default LTE band mask (QUECTEL): 100002000000000f0e189f
        ✓ 5_particle_cellular_forbidden_bands_default (5.7s)
LTE band mask after FORBIDDEN_BANDS env var (QUECTEL): 100000000000000f0e189c
        ✓ 6_particle_cellular_forbidden_bands_set (1.7s)
        ✓ 7_particle_cellular_preferred_plmn_init (3.9s)
Default CPOL response: empty
        ✓ 8_particle_cellular_preferred_plmn_default (186s)
AT+CPOL response after PREFERRED_PLMN env var:
+CPOL: 1,2,"310410",0,0,0,1

+CPOL: 2,2,"310260",0,0,0,1

+CPOL: 3,2,"311480",0,0,0,1

        ✓ 9_particle_cellular_preferred_plmn_set (5.7s)
        ✓ 10_particle_cellular_keepalive
        ✓ 11_particle_cellular_preferred_plmn_cleanup
AT+CPOL response after PREFERRED_PLMN env var:
+CPOL: 1,2,"310410",0,0,0,1

+CPOL: 2,2,"310260",0,0,0,1

+CPOL: 3,2,"311480",0,0,0,1

        ✓ 12_particle_cellular_env_vars_cleared_verify_defaults
        ✓ 97_cleanup (8.2s)
spark/flash/status started
spark/flash/status success
        ✓ 98_cleanup (7s)
        ✓ 99_cleanup_1 (4s)
        ✓ 99_cleanup_2 (8.7s)
        ✓ 10_particle_cellular_preferred_plmn_cleanup (3.6s)
        ✓ 11_particle_cellular_env_vars_cleared_verify_defaults (2.4s)

  System level env vars
    msom
      systemThread=disabled
        ✓ 01_particle_ble_enable_init (7.9s)
        ✓ 02_particle_ble_enable_default (7.4s)
        ✓ 03_particle_ble_enable_true (7.4s)
        ✓ 04_particle_ble_enable_false (11.5s)
        ✓ 05_particle_default_cloud_keepalive (8.1s)
        ✓ 06_particle_wifi_enable_init (4s)
        ✓ 07_particle_wifi_enable_default (23.3s)
        ✓ 08_particle_wifi_enable_true (18.3s)
        ✓ 09_particle_wifi_enable_false (17.6s)
        ✓ 10_particle_wifi_enable_false_connect_through_other_ifaces (23s)
        ✓ 11_particle_wifi_enable_cleanup (3.9s)
        ✓ 12_particle_ethernet_enable_init (4.4s)
        ✓ 13_particle_ethernet_enable_default (7.5s)
        ✓ 14_particle_ethernet_enable_true (3.5s)
        ✓ 15_particle_ethernet_enable_false
        ✓ 16_particle_ethernet_enable_false_connect_through_other_ifaces (31.6s)
        ✓ 17_particle_ethernet_enable_cleanup (4.1s)
        ✓ 97_cleanup (3.8s)
spark/flash/status started
spark/flash/status success
        ✓ 98_cleanup (5s)
        ✓ 99_cleanup_1 (3.9s)
        ✓ 99_cleanup_2 (8.7s)
      systemThread=enabled
        ✓ 01_particle_ble_enable_init (7.9s)
        ✓ 02_particle_ble_enable_default (9s)
        ✓ 03_particle_ble_enable_true (9s)
        ✓ 04_particle_ble_enable_false (15s)
        ✓ 05_particle_default_cloud_keepalive (7.2s)
        ✓ 06_particle_wifi_enable_init (4.1s)
        ✓ 07_particle_wifi_enable_default (24s)
        ✓ 08_particle_wifi_enable_true (19.6s)
        ✓ 09_particle_wifi_enable_false (18.3s)
        ✓ 10_particle_wifi_enable_false_connect_through_other_ifaces (24.5s)
        ✓ 11_particle_wifi_enable_cleanup (4s)
        ✓ 12_particle_ethernet_enable_init (4.4s)
        ✓ 13_particle_ethernet_enable_default (9.1s)
        ✓ 14_particle_ethernet_enable_true (5.1s)
        ✓ 15_particle_ethernet_enable_false
        ✓ 16_particle_ethernet_enable_false_connect_through_other_ifaces (16.2s)
        ✓ 17_particle_ethernet_enable_cleanup (4.1s)
        ✓ 97_cleanup (3.8s)
spark/flash/status started
spark/flash/status success
        ✓ 98_cleanup (6.8s)
        ✓ 99_cleanup_1 (3.9s)
        ✓ 99_cleanup_2 (8.7s)


  89 passing (16m)
```

### Example App
any

### References
Example using global connection time of 45s

```
0000017305 [system.cm] TRACE: Using best network: WiFi
0000017330 [system] TRACE: Address type: 1
0000017353 [system] TRACE: Cloud socket=0, family=2, type=2, protocol=17
0000017380 [system] INFO: Cloud socket=0, connecting to 3.222.143.118#5684 using if 5
0000017420 [system] INFO: Bound cloud socket to lwip if 5 ("wl4")
0000017451 [system] TRACE: Cloud socket=0, connected to 3.222.143.118#5684
0000017484 [system] INFO: system_cloud_get_netif_keepalive 45000
0000017518 [system] TRACE: Updating cloud keepalive for AF_INET: 45000 -> 45000
0000017556 [system] TRACE: Applying new keepalive interval now
0000017587 [system] INFO: Cloud socket connected
```

keep alive timing matches (45s apart)
```
0000065271 [comm.coap] TRACE: Sending CoAP message
0000065299 [comm.coap] TRACE: CON 0.00  size=4 token= id=14
0000065480 [comm.coap] TRACE: Received CoAP message
0000065506 [comm.coap] TRACE: ACK 0.00  size=4 token= id=14
<45 seconds later>
0000110533 [comm.coap] TRACE: Sending CoAP message
0000110555 [comm.coap] TRACE: CON 0.00  size=4 token= id=15
0000110742 [comm.coap] TRACE: Received CoAP message
0000110768 [comm.coap] TRACE: ACK 0.00  size=4 token= id=15
```

---

### Completeness

- [ ] Added documentation


system/src/system_cloud_connection_posix.cpp
```diff
@@ -287,7 +287,7 @@ int system_cloud_connect(int protocol, const ServerAddress* address, sockaddr* s
 
         unsigned int keepalive = 0;
 #if HAL_PLATFORM_AUTOMATIC_CONNECTION_MANAGEMENT
-        keepalive = (cloudInterface == NETWORK_INTERFACE_CELLULAR ? HAL_PLATFORM_CELLULAR_CLOUD_KEEPALIVE_INTERVAL : HAL_PLATFORM_DEFAULT_CLOUD_KEEPALIVE_INTERVAL);
+        keepalive = system_cloud_get_netif_keepalive(cloudInterface);
 #else
         system_cloud_get_inet_family_keepalive(a->ai_family, &keepalive);
 #endif
```
scott-brust: Do we care about the ipv4/ipv6 keepalives or should we switch to cloud keep alive being determined just by network interface? IE Should we let these new env vars work for all platforms?

avtolstoy: At this point - we don't. This was essentially a way of differentiating mesh (ipv6) vs non-mesh at the time when we had that. So going forward this should be not per inet family, but per interface, correct.

communication/inc/protocol.h
```diff
@@ -375,6 +375,14 @@ class Protocol
 		pinger.set_interval(interval, source);
 	}
 
+	system_tick_t get_keepalive_interval() {
+		return pinger.get_interval();
+	}
+
+	keepalive_source_t get_keepalive_source() {
+		return pinger.get_keepalive_source();
```
scott-brust: Is there a reason to expose this to the user? Or get rid of this?

system/src/system_cloud_connection.cpp
```diff
@@ -340,6 +339,55 @@ int system_cloud_set_inet_family_keepalive(int af, unsigned int value, int flags
...
+            break;
+#endif
+#if HAL_PLATFORM_CELLULAR
+        case NETWORK_INTERFACE_CELLULAR:
+            keepAliveEnvVarName = "PARTICLE_CLOUD_KEEP_ALIVE_CELLULAR";
+            break;
+#endif
+#if HAL_PLATFORM_WIFI 
+        case NETWORK_INTERFACE_WIFI_STA:
+            keepAliveEnvVarName = "PARTICLE_CLOUD_KEEP_ALIVE_WIFI";
```
scott-brust: The existing particle env vars to enable network interfaces are called `PARTICLE_ETHERNET_ENABLE`, `PARTICLE_WIFI_ENABLE` etc. 
Should we have keep alives match this, ie `PARTICLE_ETHERNET_CLOUD_KEEPALIVE`? Also, is it `KEEP_ALIVE` or `KEEPALIVE`?

system/src/system_cloud_connection_posix.cpp
```diff
@@ -285,12 +285,7 @@ int system_cloud_connect(int protocol, const ServerAddress* address, sockaddr* s
...
         }
 
-        unsigned int keepalive = 0;
-#if HAL_PLATFORM_AUTOMATIC_CONNECTION_MANAGEMENT
-        keepalive = (cloudInterface == NETWORK_INTERFACE_CELLULAR ? HAL_PLATFORM_CELLULAR_CLOUD_KEEPALIVE_INTERVAL : HAL_PLATFORM_DEFAULT_CLOUD_KEEPALIVE_INTERVAL);
-#else
-        system_cloud_get_inet_family_keepalive(a->ai_family, &keepalive);
-#endif
+        unsigned int keepalive = system_cloud_get_netif_keepalive(cloudInterface);
         system_cloud_set_inet_family_keepalive(a->ai_family, keepalive, 1);
```
scott-brust: Should we completely move off of the `inet` approach to cloud keepalives? Or should this still be gated only for ACM platforms (ie just msom). These env vars need to work on all platforms right?

system/src/system_cloud_connection.cpp
scott-brust: Should we get rid of the ipv4 vs ipv6 keepalive concept and just use the logic in `system_cloud_get_netif_keepalive()` to determine what the cloud connection keep alive should be each time we connect the cloud socket?

wiring/inc/spark_wiring_cloud.h
```diff
@@ -475,6 +475,17 @@ class CloudClass {
     }
 
     inline static void keepAlive(std::chrono::seconds s) { keepAlive(s.count()); }
+
+    inline static std::chrono::seconds getKeepAlive() {
```
scott-brust: I dont know if its better to return `unsigned` or `std::chrono::seconds` for this. You can still get ms via
```
size_t size = 0;
size_t n = sizeof(size);
auto r = spark_get_connection_property(SPARK_CLOUD_PING_INTERVAL, &size, &n, nullptr /* reserved */);
```
But it shouldnt matter since you can only ever set keepalive in seconds.

user/tests/integration/system/cellular_env/cellular_env.spec.js
```diff
@@ -97,14 +97,22 @@ test('8_particle_cellular_preferred_plmn_default', async function () {
 test('9_particle_cellular_preferred_plmn_set', async function () {
     expect(device.mailBox).to.not.be.empty;
     console.log(device.mailBox[0].d);
+
+    await setEnvVarsAndFlash({
+        PARTICLE_CLOUD_KEEP_ALIVE_CELLULAR: '60',
+    });
```
scott-brust, line 104: Should these go in their own tests to make it clearer? Just got lazy and reused the end of this test to set the env var for the next one...

system/src/system_cloud_connection.cpp
mrlambchop: I agree on this - it needs updating to match existing

communication/inc/protocol.h
avtolstoy: I think knowing/querying current keepalive/configured value is useful - this one not so much - so something like `auto val = Particle.keepAlive()` makes sense.

system/src/system_cloud_connection.cpp
avtolstoy: Yes, it no longer makes sense and was a stop-gap solution for mesh. Now that we have ACM - obsolete.

system/src/system_cloud_connection_posix.cpp
avtolstoy: I'd get rid of inet stuff entirely.

user/tests/integration/system/cellular_env/cellular_env.spec.js
avtolstoy: eh, might as well leave here.

wiring/inc/spark_wiring_cloud.h
avtolstoy: I'd propose we instead:
```c++
    int keepAlive(unsigned sec);
    int keepAlive(std::chrono::seconds s) { keepAlive(s.count()); }
    int keepAlive();
```

If we really need milliseconds, we'll need to deal with this both on the setter side, so for now just leave everything at seconds here?

avtolstoy (approved):
There is still somewhat an unsolve side of this: you have env vars for setting keepalive on per interface basis, but you don't have a matching application API for it.

So, I think we are missing something like `Particle.keepAlive(unsigned seconds, network_interface_t iface)` to allow `Particle.keepAlive(10, WiFi)` and a matching getter e.g. `Particle.keepAlive(network_interface_t iface)`.

system/src/system_cloud_connection.cpp
avtolstoy: No strong preference here, but my 0.02: I usually prefer PREFIX_BIGGERCLASS_SMALLERCLASS_CONCRETE just because it makes things easier to search for and a bit more readable among a bunch of similar things.

This is just a matter of what 'class' we want to make this. Current - makes sense to me because it's prefix->cloud (main thing this applies too)->keepalive (small class of values with variation based on interface)->interface.

The other way around (network being the main thing) makes sense too somewhat, but perhaps a bit less?

wiring/inc/spark_wiring_cloud.h
scott-brust: Ok, so these are 'setters'
```C++
    int keepAlive(unsigned sec);
    int keepAlive(std::chrono::seconds s) { keepAlive(s.count()); }
```
And this is the 'getter', returning number of seconds
```C++
    int keepAlive();
```

system/src/system_cloud_connection.cpp
scott-brust: I think the current wording makes sense, but ill change it to match the existing system env vars just to be internally consistent, ie `PARTICLE_ETHERNET_CLOUD_KEEPALIVE`. 

Should it be `KEEPALIVE` or `KEEP_ALIVE`?

scott-brust:
> There is still somewhat an unsolve side of this: you have env vars for setting keepalive on per interface basis, but you don't have a matching application API for it.
> 
> So, I think we are missing something like `Particle.keepAlive(unsigned seconds, network_interface_t iface)` to allow `Particle.keepAlive(10, WiFi)` and a matching getter e.g. `Particle.keepAlive(network_interface_t iface)`.

This would be easy if the device could set its own env vars, but I understand why it doesnt work that way. 

Right now, if a user calls `Particle.keepAlive()` it will override any of the env vars or defaults (ie any `KeepAliveSource::SYSTEM` "sources"). I assume we want to keep this same behavior for `Particle.keepAlive(unsigned, network_interface_t)` ? 
Also, `Particle.keepAlive(unsigned)` is not persistent through reboots, assuming the same would be ok for per-interface user keepalives? 

Should these user per-interface keep alive overrides be stored in ram in `system_cloud_connection.cpp`? And add some method to set those to the wiring layer?

avtolstoy:
I'd keep things consistent with current `Particle.keepAlive()`, so: runtime only (no persistence), takes precedence and yes, we would need to keep them in system cloud connection layer and forward to protocol layer when needed.

system/inc/system_cloud.h
```diff
@@ -372,9 +373,23 @@ typedef enum spark_connection_property {
     SPARK_CLOUD_MAX_EVENT_DATA_SIZE = 3, ///< Maximum size of event data (get).
     SPARK_CLOUD_MAX_VARIABLE_VALUE_SIZE = 4, ///< Maximum size of a variable value (get).
     SPARK_CLOUD_MAX_FUNCTION_ARGUMENT_SIZE = 5, ///< Maximum size of a function call argument (get).
-    SPARK_CLOUD_GET_NETWORK_INTERFACE = 6 ///< Which interface is being used for the current cloud connection
+    SPARK_CLOUD_GET_NETWORK_INTERFACE = 6, ///< Which interface is being used for the current cloud connection
+    SPARK_CLOUD_NETIF_PING_INTERVAL = 7 ///< Per-interface keepalive override (set/get).
```
avtolstoy: Nit: `SPARK_CLOUD_GET_NETWORK_INTEFACE_PING_INTERVAL` ? just to be consistent with `SPARK_CLOUD_GET_NETWORK_INTERFACE` and `SPARK_CLOUD_PING_INTERVAL`

system/src/system_cloud_connection.cpp
```diff
@@ -31,27 +32,27 @@
...
 uint16_t cloud_udp_port = PORT_COAPS; // default Particle Cloud UDP port
 
+struct NetIfKeepAlive {
+    unsigned keepAlive;
+    particle::protocol::KeepAliveSource::Enum source;
+};
+
+// Indexed by network_interface_t. Only USER overrides are stored here; SYSTEM
+// values are resolved from env vars / HAL defaults in system_cloud_get_netif_keepalive().
+static NetIfKeepAlive netIfKeepAlives[NETWORK_INTERFACE_PPP_SERVER + 1] = {};
```
avtolstoy: Let's add an explicit `NETWORK_INTERFACE_MAX` or something like that.

wiring/inc/spark_wiring_cloud.h
```diff
@@ -475,6 +475,42 @@ class CloudClass {
     }
 
     inline static void keepAlive(std::chrono::seconds s) { keepAlive(s.count()); }
+
+    inline static int getKeepAlive() {
```
avtolstoy, line 479: `keepAlive()` vs `getKeepAlive()` as we discussed?

scott-brust: The issue is that there are two pairs of functions
1. The current active keep alive
```
inline static int keepAlive()  //get
inline static void keepAlive(unsigned sec) //set
```
2. The per-interface keep alive (with caveat that this may not be the active keep alive if the netif isnt being used!)
```
inline static int keepAlive(network_interface_t network) //get
inline static void keepAlive(unsigned sec, network_interface_t network) //set
```
The two 'get' functions signatures conflict since `network_interface_t` is typedef to `uint32_t`/ `unsigned`. I had to add 'get' to disambiguate.  

Should the function signature for the specific netif getter be something like this in order to not conflict? 
```
inline static int keepAlive(NetworkClass& network) 
```

avtolstoy: Ugh, that's right. Ok, let's do `getKeepAlive()` then, while keeping the old `keepAlive()` for compatibility.

scott-brust: There was no `keepAlive()` 'getter' only the setter functions. So I figured I would make them match since they're both new

avtolstoy: even better

system/inc/system_cloud.h
scott-brust: This is used to both get and set netif ping interval. Maybe just `SPARK_CLOUD_NETWORK_INTERFACE_PING_INTERVAL` ?

avtolstoy: :+1: even better.

avtolstoy (approved):

PR #2920, opened 2026-05-18
[test] adds tethering usb/serial1

technobly:
- [test] adds tethering usb/serial1
- Folds the standalone tethering HIL test (previously driven manually by tethering-tests-bw.sh + a docker container) into the automated 6x release pipeline.

### DIFF
--- a/user/tests/wiring/tethering_usb/tethering_usb.spec.js
+++ b/user/tests/wiring/tethering_usb/tethering_usb.spec.js
@@ -1,0 +1,138 @@
+'use strict';
+/* eslint no-undef: 'off' */
+
+// Tethering-specific glue. The helpers below (constants, rig loader,
+// runTetheringTest) are duplicated in ../tethering_serial1/tethering_serial1.spec.js
+// because the two scenarios share most of the host-side setup but pass a
+// different `mode` and live in sibling test directories. The only thing
+// hoisted out is generic docker plumbing (integration/test/docker.js).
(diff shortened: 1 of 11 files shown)


user/tests/wiring/tethering_usb/tethering_usb.spec.js
```diff
@@ -0,0 +1,195 @@
...
+const fs = require('fs');
+const path = require('path');
+
+const DEFAULT_URL = 'http://technobly.com/100/phrack.29.phk';
+const CONTAINER_NAME_PREFIX = 'tethering-hil';
+// Outer cap on the docker-run subprocess. The container's own
+// tethering-tests.sh enforces tighter per-step timeouts.
+const RUN_OVERHEAD_SEC = 300;
+
+function loadRigConfig() {
```
avtolstoy, line 30: Main comment is to probably decouple this common stuff into a requirable js file to use commonly between usb/serial1 spec.js?

see `user/tests/integration/test` folder for where such common code goes into.

avtolstoy (approved):

user/tests/wiring/tethering_usb/tethering_usb.spec.js
technobly: yeah, I didn't feel great about duplicating the JS code on both devices... but it wasn't able to find any shared JS code in the test folder itself.  A shared docker.js for the general stuff would be good, but there will still be some duplication in the spec.js files.

PR #2922, opened 2026-05-22
[rtl872x] usb: fix control interface class driver consuming dataIn/dataOut events for unrelated endpoints

avtolstoy:


technobly (approved):
LGTM, could even `& 0x0f` to make it even tighter, but 7f would allow for some kind of usb spec expansion we are unaware of at this time.

PR #2924, opened 2026-05-24
wiring/usartserial: add multi-byte USARTSerial write overload

TJett:
### Problem

`SerialN.write(buf, len)` on `USARTSerial` does not use a native buffer-write path. It falls back to `Print::write(const uint8_t*, size_t)`, which iterates through `write(uint8_t)` one byte at a time.

On P2 / Photon 2 this causes buffered UART writes to be fed into the HAL one byte at a time, which introduces large inter-byte gaps and makes `SerialN.write(buf, len)` take much longer than expected.

### Solution

Add a native `USARTSerial::write(const uint8_t* buffer, size_t size)` overload and route it to `hal_usart_write_buffer()`.

Additionally, add `hal_usart_write_buffer()` to the dynalib table. `hal_usart_read_buffer()` and `hal_usart_peek_buffer()` were exported at the same time for API consistency with the existing HAL buffer interface.

Files changed:
* [wiring/inc/spark_wiring_usartserial.h](https://github.com/particle-iot/device-os/compare/develop...TJett:device-os:develop#diff-de4d90f03c0f14e614f4163191172eb3f915442cb623b529eb21c1d5a48ee1dd)
* [wiring/src/spark_wiring_usartserial.cpp](https://github.com/particle-iot/device-os/compare/develop...TJett:device-os:develop#diff-fd84260164d43bd9bcef2ecd1041e71e641de1d1d0421621fccc844ff8aca5c6)
* [hal/inc/hal_dynalib_usart.h](https://github.com/particle-iot/device-os/compare/develop...TJett:device-os:develop#diff-bbef537fe0005217f0f9256be7dae4e110588c47d36b30141ed635d9186be05e)

### Steps to Test

Application-level testing on P2 / Photon 2:
* Build unmodified Device OS and run the example app below.
* Scope the UART TX pin while sending fixed-size buffers, for example 8-byte and 64-byte writes at 115200 Mbps.
* Measure the duration of SerialN.write(buf, len) using micros().
* Repeat with this PR applied.

Expected results:

* Before the fix:
  * SerialN.write(buf, len) takes significantly longer.
  * Visible idle gaps appear between transmitted bytes.
* After the fix:
  * SerialN.write(buf, len) returns much sooner.
  * Bytes transmit back-to-back without the added inter-byte gaps.
  
When running the example code below, I got the following results:

Before:
* Serial1.write avg=1326 us over 490 samples (64-byte packets at 115200 baud)

After:
* Serial1.write avg=129 us over 439 samples (64-byte packets at 115200 baud)

Existing serial tests were also run with a direct TX/RX loopback setup, but I do not currently have the recommended buffer fixture for the full loopback test coverage, so I am not including a new automated timing-based test in this PR (I also think direct timing measurements in the loopback test would be somewhat fragile).

### Example App

```c
#include "Particle.h"

SYSTEM_MODE(MANUAL);

SerialLogHandler logHandler(LOG_LEVEL_INFO);

constexpr uint32_t BAUD = 115200;
constexpr size_t WRITE_SIZE = 64;

uint8_t tx[WRITE_SIZE];
system_tick_t sampleStart = 0;
uint32_t samples = 0;
uint64_t totalWriteUs = 0;

void setup() {
    for (size_t i = 0; i < WRITE_SIZE; ++i) {
        tx[i] = 'A' + (i % 26);
    }

    Serial1.begin(BAUD);
    sampleStart = millis();
}

void loop() {
    system_tick_t startUs = micros();
    size_t written = Serial1.write(tx, sizeof(tx));
    system_tick_t elapsedUs = micros() - startUs;

    if (written == sizeof(tx)) {
        totalWriteUs += elapsedUs;
        ++samples;
    }

    if (millis() - sampleStart >= 5000) {
        if (samples > 0) {
            Log.info("Serial1.write avg=%lu us over %lu samples (%u-byte packets at %lu baud)",
                    (unsigned long)(totalWriteUs / samples),
                    (unsigned long)samples,
                    (unsigned)WRITE_SIZE,
                    (unsigned long)BAUD);
        }
        totalWriteUs = 0;
        samples = 0;
        sampleStart = millis();
    }

    delay(10);
}
```

### References

* [Particle community thread: Photon 2 Slow UART Serial TX Speeds](https://community.particle.io/t/photon-2-slow-uart-serial-tx-speeds/71260)

### AI Disclosure

I used Codex/GPT-5.4 to assist me in making this PR.

---

### Completeness

- [ ] Added documentation

### DIFF
--- a/hal/inc/hal_dynalib_usart.h
+++ b/hal/inc/hal_dynalib_usart.h
@@ -81,6 +81,9 @@ DYNALIB_FN(BASE_IDX2 + 3, hal_usart, hal_usart_break_detected, uint8_t(hal_usart
 DYNALIB_FN(BASE_IDX2 + 4, hal_usart, hal_usart_sleep, int(hal_usart_interface_t serial, bool, void*))
 DYNALIB_FN(BASE_IDX2 + 5, hal_usart, hal_usart_init_ex, int(hal_usart_interface_t, const hal_usart_buffer_config_t*, void*))
 DYNALIB_FN(BASE_IDX2 + 6, hal_usart, hal_usart_get_features, int(hal_usart_interface_t, uint32_t*, void*))
+DYNALIB_FN(BASE_IDX2 + 7, hal_usart, hal_usart_write_buffer, ssize_t(hal_usart_interface_t serial, const void* buffer, size_t size, size_t elementSize))
+DYNALIB_FN(BASE_IDX2 + 8, hal_usart, hal_usart_read_buffer, ssize_t(hal_usart_interface_t serial, void* buffer, size_t size, size_t elementSize))
+DYNALIB_FN(BASE_IDX2 + 9, hal_usart, hal_usart_peek_buffer, ssize_t(hal_usart_interface_t serial, void* buffer, size_t size, size_t elementSize))
 
 DYNALIB_END(hal_usart)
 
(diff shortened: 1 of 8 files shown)


TJett:
Superseded by https://github.com/particle-iot/device-os/pull/2935

PR #2928, opened 2026-06-06
Implements power manager env vars, [gen 3] reclaim >3KB flash w/`-flto-partition=one` [sc-139793]

technobly:
### Feature

- Implements power manager env vars
  - Implements two new env vars for Power Manager, `PARTICLE_POWER_INPUT_CURRENT` and `PARTICLE_POWER_CHARGE_CURRENT` which may be used to override default or user set `SystemPowerConfiguration ::powerSourceMaxCurrent` and `SystemPowerConfiguration:: batteryChargeCurrent` respectively.  Values will be checked and validated by System Power Manager when configuration is loaded.  Please use official documentation to set values appropriately.

### Internal

- #### [gen 3] reclaim >3KB flash w/`-flto-partition=one`

#### What

Add `-flto-partition=one` to the LTO_EXTRA_OPTIMIZATIONS link flags in `build/arm-tools.mk` (the `USE_LTO=1`
branch). Affects only LTO-enabled links `system-part1` on the Gen 3 platforms (minus argon)
that set `FORCE_LTO=1`. Non-LTO builds are untouched.

```diff
ifneq ($(LTO_EXTRA_OPTIMIZATIONS),)
-CFLAGS += -fmerge-all-constants
-LDFLAGS += -fmerge-all-constants
+CFLAGS += -fmerge-all-constants -flto-partition=one
+LDFLAGS += -fmerge-all-constants -flto-partition=one
endif
```

#### Why

`system-part1` is at the flash ceiling on boron, bsom, b5som, tracker, esomx.  Recent changes have overflowed
`APP_FLASH` on bsom.

By default GCC splits LTO code generation into ~32 partitions. Partition boundaries are
optimization barriers: cross-partition calls can't inline and identical constants/strings
can't merge.

`-flto-partition=one` generates code in a single partition: no boundaries.

#### Result (bsom, exact CI flags, gcc-arm 10.2.1)

| | default (~32 partitions) | `-flto-partition=one` | Δ |
|---|---|---|---|
| system-part1.bin | 573,072 | 569,960 | **−3,112 bytes** |
| headroom (573,440 region) | 368 | 3,480 | ~9.5× |

Recovers >3 KB on the most constrained platforms

#### Trade-offs / risk

- **Link time & memory**: single-partition codegen is serial and holds the whole program in
  one process, minutes-scale, hundreds of MB. Watch CI builder memory on first run.
  Fallback if it bites: `-flto-partition=balanced --param lto-partitions=8` (deterministic
  count, parallel, most of the size win).
- More aggressive inlining can surface latent bugs (same risk class as a toolchain bump), 
  will validate with a complete HIL test on all platforms.

### Steps to Test

- Run `system/env` tests
- Run all HIL tests

### References

- [sc-139793]

### DIFF
--- a/user/tests/integration/system/env/env.cpp
+++ b/user/tests/integration/system/env/env.cpp
@@ -655,9 +655,14 @@ bool waitAppliedPmicConfig(uint16_t inputLimit, uint16_t chargeCurrent, system_t
 
 test(18_particle_power_env_init) {
 #if POWER_ENV_TESTS
-    PMIC power(true);
-    power.begin();
-    skipPowerEnv = (power.getVersion() != BQ24195_VERSION);
+    {
+        // Scope the PMIC lock to the presence probe: holding it across
+        // System.setPowerConfiguration() can deadlock with the power manager
+        // thread, which takes the same lock while processing events
+        PMIC power(true);
+        power.begin();
+        skipPowerEnv = (power.getVersion() != BQ24195_VERSION);
+    }
     if (skipPowerEnv) {
         skip();
         return;
@@ -671,6 +676,7 @@ test(18_particle_power_env_init) {
     conf.feature(SystemPowerFeature::PMIC_DETECTION);
 #endif
     assertEqual(System.setPowerConfiguration(conf), 0);
+    System.enableFeature(FEATURE_RESET_INFO); // For the reset reason check in the next test
     System.clearEnv(false /* reset */);
     expectSystemReset();
     System.reset();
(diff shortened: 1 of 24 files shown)


user/tests/integration/system/env/env.cpp
```diff
@@ -613,6 +613,185 @@ test(17_particle_ethernet_enable_cleanup) {
...
+#else
+#define POWER_ENV_TESTS 0
+#endif
+
+#if POWER_ENV_TESTS
+
+namespace {
+
+// Default power configuration values as applied to the PMIC
+const uint16_t DEFAULT_PMIC_INPUT_CURRENT_LIMIT = 900; // mA
```
avtolstoy: these are already in `system_power.h` and exposed for tests, so let's use those to avoid having to change multiple places in case we modify.

avtolstoy (approved):

PR #2932, opened 2026-06-23
Adds iperf and improves PPP (cellular / tethering performance), related bugfixes

avtolstoy:
### Problem

The existing tethering HIL tests only validated basic setup/bind and a single HTTP download speed measurement. They needed to be improved with comprehensive bidirectional throughput testing (UDP/TCP, loss/jitter, device-direct cellular paths) to properly validate tethering performance and reliability across platforms.

Improving the tests uncovered several pre-existing platform issues:

1. **Gen 4 Serial1 (LP_UART) tether non-functional.** `enableEvent()` in the rtl872x USART interrupt-mode branch returned without re-asserting READABLE/WRITABLE from the current ring buffer state, unlike the DMA-mode branch. `waitEvent()` callers (PPP tether server) only progressed on the 1 s poll timeout, advancing one protocol step per second. ModemManager timed out identifying the device, PPP never came up.

2. **Gen 4 RTS/CTS flow control silently broken.** Two issues: (a) flow control enable used two back-to-back `|=` RMWs on the slow-clock-domain MCR register; the second read could return the pre-first-write value and clobber the AFE bit; (b) the RX ISR drained the FIFO into the ring even when the ring was full, discarding bytes and defeating auto-RTS since the FIFO never filled to deassert RTS. Auto flow control was silently disabled, host not throttled; under RX flood, frames were silently shredded (98% PPP frame loss observed) instead of backpressured.

3. **Gen 3 USART TX race condition.** `startTransmission()` was called after the TxLock released, and interrupt disable/enable around ENDTX wasn't in a critical section. A high-priority thread preempting between lock release and TX start could cause TX transmissions to never start.

4. **NAT TCP MSS clamping was a complete no-op.** Upstream code parsed options into a local buffer, rewrote MSS in the local copy, and never wrote it back to the pbuf (`pbuf_take_at` missing). Checksum was computed before the intended rewrite. Match condition only covered a narrow SYN/RST flag combo. Options buffer was 20 bytes (max is 40). Full-size DF segments (1460 MSS) were not clamped to the cellular path MTU (~1400 measured), so they were black-holed over the cellular link. Curl uploads and NAT-forwarded TCP timed out.

5. **PPP TX deadlock.** TX packets were sent inside the lwIP `output()` netif callback, which runs with the lwIP core lock held. The PPP TX path acquires the same lock to enqueue/forward. AB-BA deadlock between the tcpip thread and the PPP pump thread under traffic.

6. **USB starvation under cellular RX flood.** The gsm0710 muxer pump thread runs at NETWORK priority; the USB deferred-ISR processing thread runs one priority below. Under sustained cellular RX traffic the pump never yielded. USB ISR processing starved, CDC endpoint stalls, control requests time out, device appears wedged.

7. **Gen 4 USB CDC buffer reinitialization not possible.** `HAL_USB_USART_Init()` early-returned if the CDC driver was already enabled, so the CDC ring buffers could not be resized after first init - any caller wanting larger buffers was stuck with the 256-byte default. Additionally the default-buffer path `malloc`'d new buffers on every call (leak), and the reconfigure path did not detach/reinit/reattach the USB device.

8. **`os_thread_join()` use-after-free and idle priority bump.** FreeRTOS task enumeration could dereference a potentially already freed TCB. `vApplicationIdleHook()` priority bump only applied for 1 cycle via an atomic flag, so terminated thread data wasn't reclaimed until the idle thread got CPU at the lowest priority.

9. **SerialUSBStream and wiring issues.** `SerialUSBStream::write()` did not bound the write size to the available space in the CDC TX buffer; blocking write could attempt to send more data than the buffer could accept. The wiring `foreach_if_addr` macro dereferenced `i->if_addr->addr` without null-checking, crashing for interfaces without a hardware address.

10. **Container MM probes all host USB devices.** The container bind-mounts `/dev/bus/usb`, so ModemManager sees every host USB device. Without a catch-all ignore rule, MM probes every USB serial it finds - including devices belonging to other concurrent test containers, causing interference.

11. **Cloud connection check triggered on irrelevant interface state changes.** `forceCloudPingOrTest()` ran on every interface link/address change, including tether interfaces that can't carry the cloud connection. This caused unnecessary cloud connection checks when bringing up tethering.

Additionally, running iperf3 reliably on these constrained platforms required dedicated support work:

- Gen 3 (nRF52840, ~22 KB free heap, 5 TCP PCBs, 16 pbufs): iperf's per-interval JSON accumulation, 128 KB default TCP block size, and control-socket churn all needed bounding to fit the resource budget without collapsing mid-test
- Both platforms: iperf control sockets needed `SO_LINGER{1,0}` so the 5-deep TCP PCB pool isn't eaten by FIN_WAIT/TIME_WAIT across the many sequential test connections; the sync-close path needed bounding so the server thread can't wedge forever on a peer FIN that never arrives over the tether path

### Solution

**Extended tethering HIL suite** (20 tests per variant, USB + Serial1):
- iperf3 UDP sustain (down/up, normal + 64-byte) with adaptive bitrate search
- iperf3 TCP down/up
- HTTP download/upload speed tests
- Device-direct iperf3 full-path tests (device as client to public iperf3 servers, both UDP and TCP, both directions)
- Final stats report with per-platform link-budget assertions
- Common `IperfServer`/`IperfClient` test library; container-side `tethering.js` test helper with detached-container exec model (`startContainer`/`execContainer`/`stopContainer`)

**Fixes for the issues above:**

1. USART (rtl872x): fix interrupt-mode `enableEvent` to re-assert READABLE/WRITABLE from current buffer state (mirrors DMA branch). Fix FIFO overflow handling under RTS flow control (don't drain-and-discard on ring-full, let the FIFO fill so auto-RTS holds the peer). Fix slow-clock-domain MCR RMW losing AFE bits (single combined write). Add IER shadow register so interrupt config writes go through a critical-section RMW instead of a direct register RMW that loses bits on the slow clock domain. Line-status ISR now falls through to the RX drain path instead of starving RX under sustained error conditions.

2. USART (nRF52840): `startTransmission()` now called inside the TxLock so TX isn't started after the lock released. TxLock now uses `taskENTER_CRITICAL`/`taskEXIT_CRITICAL` around the ENDTX interrupt disable/enable to fix the race. Null-check the `consume()` return.

3. NAT: rewrote MSS clamping - write options back via `pbuf_take_at`, recompute checksum after, cover SYN + SYN-ACK, 20->40 byte options buffer. Clamps to 1360.

4. PPP client: output TX packets via a separate queue processed in the pump loop, outside the lwIP `output()` callback (fixes AB-BA deadlock). Non-blocking queue put for `EVENT_TX_DATA`. pbuf-allocated copies with drop-and-report-as-written on queue full (packet loss, not link error). pbuf pool drained on disconnect. Server LCP echo max fails 2->8, adaptive echo enabled. Report dropped frame length to avoid lwIP generating `ERR_IF` errors.

5. GSM0710 muxer: `GSM0710_COOP_YIELD_MS` (rtl872x) - pump yields 1 tick every 4 ms so USB ISR processing gets CPU under heavy cellular traffic.

6. Concurrent HAL: `os_thread_join()` uses `uxTaskGetSystemStateParticle()` callback to check task existence without dereferencing the handle (avoids stale/freed TCB). `vApplicationIdleHook()` checks `uxTaskGetNumberOfTasksWaitingTermination()` under a critical section and only restores idle priority when there's nothing to clean up.

7. USB CDC (rtl872x): `HAL_USB_USART_Init` now supports reconfiguration with larger buffers (detach, reinit, reattach) instead of early-returning when already enabled. Default buffer allocation reused across calls (static) to avoid leaking. `startRx`/`startTx` round transfer sizes down to a multiple of `MAX_DATA_PACKET_SIZE` (64) so the DWC2 endpoint can queue multi-packet IN/OUT transfers instead of being capped at one 64-byte packet per transfer. TX timeout timer distinguishes a real stall (remaining bytes not decreasing) from a slow host. `endpointTransferComplete()`/`endpointTransferRemaining()` accessors for DIEPINT/DOEPINT and DIEPTSIZ/DOEPTSIZ so the CDC class driver can detect transfer progress.

8. Serial stream: coalesce platform-specific `serial_stream.cpp`/`.h` (nRF52840, rtl872x) into a shared implementation under `hal/shared/`. `SerialUSBStream::write()` now bounds the write size to `HAL_USB_USART_Available_Data_For_Write()` and returns 0 when the buffer is full. Wiring: null-check `if_addr`/`addr` in the `foreach_if_addr` macro.

9. Container image: bake a catch-all udev rule (`77-mm-ignore-all-usb-tty.rules`) into the image that sets `ID_MM_DEVICE_IGNORE=1` on all USB tty devices by default. The per-test rule (`78-mm-tethering.rules`) overrides `ID_MM_DEVICE_IGNORE=0` for the target device/port only. This closes the startup race where MM probes other containers' devices before the per-test rule is written.

10. System network: `forceCloudPingOrTest(iface)` now checks `isCloudConnectionInterface()` and only schedules a cloud connection network check when the state change is on an interface that can actually carry the cloud connection, avoiding spurious checks on tether interface events.

**iperf3 support on constrained platforms:**
- lwIP: enable `SO_LINGER` (nRF52840 + rtl872x) for faster TCP PCB/pbuf cleanup via RST-on-close. rtl872x TCP performance improvements (`TCP_SND_BUF` 2->8xMSS, `TCP_WND` 4->16xMSS, `MEMP_NUM_TCP_SEG` 16->48). PPP_LCP_ADAPTIVE on both platforms. `PPP_INPROC_IRQ_SAFE` 1->0 (input now takes the tcpip core lock explicitly). rtl872x PPP/PRINTPKT debug now always on, synced with Gen 3 logging. `ppp_dbglog`/`ppp_error`/etc implementations in `lwip_logging.cpp`.
- PPP server netif: reduce output blocking wait to 100 ms. Add flow-control pause/resume mechanism - when `client_.input()` returns `NO_MEMORY` and flow control is configured, stop reading and wait for the pbuf pool to recover (memp event handler signals `RESUME`). Limit RX bytes per iteration to 2x PBUF_POOL_BUFSIZE. Serial1 flow-control state tracked from AT+IFC.
- Per-platform PPP TX queue size: nRF52840=4, rtl872x=16 (overridable via `HAL_PLATFORM_PPP_TX_QUEUE_SIZE`).
- nRF52840: `MODULE_FUNCTION` guard made `defined()`-safe for bootloader includes.

**New on-device test:**
- `UART_01_TxRingConsistentUnderHighPriorityThreadPreemption` (no_fixture_long_running): verifies TX ring buffer consistency under high-priority thread preemption for 60s - catches the Gen 3 USART HAL race condition between `startTransmission`, the TxLock, and the TX ISR (issue #3 above).

**Submodule updates** (separate PRs):
- https://github.com/particle-iot/lwip/pull/19 - `SO_LINGER{1,0}` unconditional abort on close; `tcp_pcb` double-free fix; `ERR_IF`->`EIO`; per-pcb adaptive LCP echo
- https://github.com/particle-iot/iperf/pull/1 - Device OS port of iperf 3.21 with RAM-constrained-platform hardening (JSON interval discard, blksize cap, bounded worker stack, socket-leak fixes, FD_CLR UB fix, SO_SNDTIMEO, bounded sync_close)
- https://github.com/particle-iot/gsm0710muxer/pull/12 - cooperative yield under heavy traffic
- https://github.com/particle-iot/freertos/pull/5 - safe task enumeration for `os_thread_join`

**CI/infra:**
- https://github.com/particle-iot/device-os-concourse-pipelines/pull/21 - container image -> Ubuntu 26.04 (iperf3 3.20), tcpdump, resolve-tty FTDI/CP210x support, baked-in MM ignore-all-usb-tty udev rule
- https://github.com/particle-iot/device-os-test-runner/pull/47 - report-builder: allow to attach additional data to the report (partice.data) - e.g. stats

### Steps to Test

**Required environment variables:**

- `TETHERING_DOCKER_IMAGE` - the HIL container image tag (e.g. `tethering-hil:3.2.0`)
- `TETHERING_ADAPTER_SERIAL` - USB-to-UART adapter serial for Serial1 variant (USB variant uses the device's own serial)

```
export TETHERING_DOCKER_IMAGE=tethering-hil:3.2.0
export TETHERING_ADAPTER_SERIAL=A5027JW4

device-os-test run <msom|b5som> <tethering_serial1|tethering_usb>
```

All 20 tests per variant pass on both b5som and msom.

### References

- https://github.com/particle-iot/lwip/pull/19
- https://github.com/particle-iot/iperf/pull/1
- https://github.com/particle-iot/gsm0710muxer/pull/12
- https://github.com/particle-iot/freertos/pull/5
- https://github.com/particle-iot/device-os-concourse-pipelines/pull/21
- https://github.com/particle-iot/device-os-test-runner/pull/47


technobly (approved):
😅

PR #2933, opened 2026-06-25
[m635] enable 2G fallback support

technobly:
### Problem

- 2G support was disabled for M-SoM temporarily, enabling only for M635 (not M404 at this time)
- `isQuecBG95xDevice()` did not include M5 and S5, and is used for modemPower/Reset timings

### Solution

- Enable and test 2G on M635 for correct operation
- Added `isQuec2g3gEnabled()` helper function to determine which devices use CREG/CGREG AT commands and need `nwscanmode` set to AUTO.  Also used to exclude BG96 from having 2G enabled.
- `nwscanseq` still isolated to M-SoM
- `interveneRegistration()` function now processed for Cat1 / 2g3g / Cat-M1 devices explicitly
- Note: EG91xxx devices were relying on factory defaults for `nwscanmode`, now set properly if needed
- Some BG95 devices and BG77 do not support `nwscanmode` and were excluded
- 3dB attenuation for 2G on M635 required by certification
- Fix `isQuecBG95xDevice()`

### Steps to Test

- Force M635 to connect on 2G by changing `int desiredNwScanMode = 1;`
- Auto mode (0) will be tested with DVOS release tests.
- Run on Dioptra on M635 and M404 for good measure (requires a few tweaks not included here)
- Double check timings for BG95-M5/S5 power on/off/reset

### References

- [sc-140313]
- [sc-140284]

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -96,6 +96,7 @@ inline system_tick_t millis() {
 const auto QUECTEL_NCP_DEFAULT_SERIAL_BAUDRATE = 115200;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE = 460800;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_BG95_M5 = 921600;
+const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_BG95_S5 = 921600;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_EG91_NAX = 921600;
 const auto QUECTEL_NCP_RUNTIME_SERIAL_BAUDRATE_EG91_EX = 921600; // version A08 or above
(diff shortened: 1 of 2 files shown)


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -1704,6 +1725,26 @@ int QuectelNcpClient::setPolicymanServiceMode(CellularPolicymanServiceMode mode,
...
+    auto resp = parser_.sendCommand(10000, "AT+QNVFR=\"/nv/item_files/rfnv/00025063\"");
+    char curVal[40 + 1] = {};
+    auto r = resp.scanf("+QNVFR: %40[^\n]", curVal);
+    // Soft response on ERROR
+    const int result = CHECK_PARSER(resp.readResult());
+    if (result != AtResponse::OK) {
+        return SYSTEM_ERROR_NONE;
+    }
+
+    if (r == 1 && strcmp(curVal, "5A0A2C0B780A140A0000F6099C09E80884080000") != 0) {
```
avtolstoy: Nit: make a constant with the value and use that universally in strcmp/`%s` execCommand?

avtolstoy (approved):

PR #2939, opened 2026-08-06
[Gen 3] Fix USB CDC write triggering an assertion failure; migrates wiring/usbserial to be run under test runner, more usb tests

avtolstoy:
### Problem

1. On Gen 3 a non-blocking `Serial.write()` racing a full TX buffer could SOS the device. `hal_usb_cdc_pvt_send_data()` `SPARK_ASSERT`ed that `app_fifo_write()` succeeds, but a full FIFO returns `NRF_ERROR_NO_MEM` - so writing into a full buffer was an assertion failure instead of a short write. `SerialUSBStream` had the same problem one layer up: errors from the HAL propagated out of `read()`/`peek()`/`write()` instead of being reported as "0 bytes done". **This is a 6.5.0 regression introduced in https://github.com/particle-iot/device-os/pull/2935**

2. `wiring/usbserial` was a manual test, so it never ran on HIL.

3. USB coverage beyond CDC data was basically nonexistent: control-request transport (chunking, concurrency, cancellation), descriptors/WCID, device modes and bus-reset robustness had no automated tests. `user/tests/app/usb_ctrl_request` predates the test runner and never runs.

### Solution

1. `hal_usb_cdc_pvt_send_data()` returns `SYSTEM_ERROR_NO_MEMORY` on a full FIFO instead of asserting; `SerialUSBStream` treats it as a zero-length transfer.

2. Migrated `wiring/usbserial` under the test runner (20 tests). Host side drives the CDC tty via a new `UsbCdcPort` helper (`user/tests/integration/test/usb_cdc.js`): resolves the tty by USB serial number through sysfs, toggles DTR via open/close, creates TX backpressure by pausing the reader. Covers blocking/non-blocking TX stress, RX stress, `isConnected()` transitions, RX buffer fill, `end()`/`begin()` while the port is closed.

3. New `wiring/usb` suite (18 tests, gen3 + gen4) - everything USB that isn't CDC data:
   - system control requests (`CTRL_REQUEST_ECHO`) across the allocation/chunking boundaries (1–10000 bytes: pooled vs async alloc at 64, host-side chunked SEND/RECV above 4096), 4 concurrent requests
   - app custom request path end-to-end, including deferred completion so requests stay `PENDING` across multiple host CHECK polls and complete out of order
   - raw vendor requests (`bRequest` 0x50: `DEVICE_ID`, `SYSTEM_VERSION`), raw INIT->RESET->CHECK cancellation, `DIAGNOSTIC_INFO`
   - descriptor assertions: VID/PID vs `@particle/device-constants`, iSerialNumber == device ID, per-platform interface layout, MS OS string + WCID compat ID / extended properties
   - listening mode enter/leave observed from both host and device; control requests keep working in listening mode
   - host-initiated bus reset x5 with recovery and `Serial.end()`/`begin()` re-enumeration stress

   Supersedes `user/tests/app/usb_ctrl_request` (`GET_MODULE_INFO` is already covered by the ota/* suites).

4. Test runner application: the request handler gained an `e` echo command with an optional delay - delayed requests are completed from `testAppLoop()`, same pattern as the old `usb_ctrl_request` app. Removed the dead `RequestHandler::instance()` singleton: the live handler is `TestSuite::reqHandler_`, and the stale singleton silently absorbed the loop hook (deferred requests queued on one object, drained on the other). Added `test/platform_util.js` - wiring specs can't `require()` node modules directly (module resolution uses the spec's realpath, which is outside `integration/`), so helpers with npm deps have to live in `integration/test/`.

### Steps to Test

```
device-os-test --device-os-dir=. run <platform> wiring/usbserial wiring/usb
```


hal/shared/serial_usb_stream.cpp
```diff
@@ -84,7 +84,11 @@ int SerialUSBStream::read(char* data, size_t size) {
         return 0;
     }
 
-    return hal_usb_cdc_pvt_recv_data(data, size);
+    auto r = hal_usb_cdc_pvt_recv_data(data, size);
+    if (r == SYSTEM_ERROR_NO_MEMORY) {
+        return 0;
+    }
+    return r;
```
scott-brust: On gen3, `hal_usb_cdc_pvt_recv_data` should only return either 
- `0` if the underlying fifo has no received data
- `1 <= x <= size` depending on how much data was read from the fifo. 
It shouldnt return `SYSTEM_ERROR_NO_MEMORY`
This doesnt have any functional effect though, since if nothing is read it will just return `r=0` as expected.

hal/shared/serial_usb_stream.cpp
scott-brust, line 109: I think there is an issue with this. 
On gen3 `read(nullptr, size)` -> `hal_usb_cdc_pvt_recv_data()` -> `app_fifo_read()` with a null buffer will be interpreted as querying the length of the fifo and return `NRF_SUCCESS` with the size skipped set to the entire length of the fifo:
```
uint32_t app_fifo_read(app_fifo_t * p_fifo, uint8_t * p_byte_array, uint32_t * p_size)
{
...
    const uint32_t byte_count    = fifo_length(p_fifo);
...
    (*p_size) = byte_count;

   // Check if application has requested only the size.
    if (p_byte_array == NULL)
    {
        return NRF_SUCCESS;
    }
...
```
I think we need to have a dummy buffer to actually consume the bytes

`skip()` doesnt appear to be used right now but if someone does use it, I dont think it will work

hal/shared/serial_usb_stream.cpp
```diff
@@ -94,7 +98,11 @@ int SerialUSBStream::peek(char* data, size_t size) {
     if (size == 0) {
         return 0;
     }
-    return HAL_USB_USART_Peek_Buffer(serial_, data, size);
+    auto r = HAL_USB_USART_Peek_Buffer(serial_, data, size);
+    if (r == SYSTEM_ERROR_NO_MEMORY) {
+        return 0;
```
scott-brust, line 103: For gen4, `SYSTEM_ERROR_NO_MEMORY` means there is no data to peek?

hal/src/nRF52840/usb_hal_cdc.c
```diff
@@ -788,7 +788,11 @@ int hal_usb_cdc_pvt_send_data(const char* data, size_t size) {
     }
 
     uint32_t bytesToWrite = size;
-    SPARK_ASSERT(app_fifo_write(&m_usb_instance.tx_fifo, data, &bytesToWrite) == NRF_SUCCESS);
+    uint32_t result = app_fifo_write(&m_usb_instance.tx_fifo, data, &bytesToWrite);
+    if (result == NRF_ERROR_NO_MEM) {
+        return SYSTEM_ERROR_NO_MEMORY;
+    }
+    SPARK_ASSERT(result == NRF_SUCCESS);
```
scott-brust, line 795: 👍

scott-brust (approved):
Works for me on `msom`
```
➜  device-os git:(fix/gen3-usb-cdc-migrate-usbserial-tests) ✗ device-os-test --device-os-dir=. run msom wiring/usbserial wiring/usb


  USB
    msom
      systemThread=enabled
        ✓ USB_00_SystemEchoShort (4.8s)
        ✓ USB_01_SystemEchoLarge (1s)
        ✓ USB_02_SystemEchoConcurrent
        ✓ USB_03_AppCustomEcho
        ✓ USB_04_AppCustomEchoDeferred (3.3s)
        ✓ USB_05_DiagnosticInfo
        ✓ USB_06_RawVendorRequests (1.1s)
        ✓ USB_07_ServiceRequestCancellation
        ✓ USB_08_DeviceDescriptor
        ✓ USB_09_InterfaceLayout
        ✓ USB_10_MsftOsStringDescriptor
        ✓ USB_11_WcidCompatIdDescriptor
        ✓ USB_12_EnterListeningMode (1.4s)
        ✓ USB_13_DeviceObservesListeningMode (1.3s)
        ✓ USB_14_DeviceObservesNormalMode
        ✓ USB_15_HostBusResetRecovery (16.9s)
          [2a] Hub-cycle disabled, device not found by portPath
          [2a] Hub-cycle disabled, device not found by portPath
          [2a] Hub-cycle disabled, device not found by portPath
          [2a] Hub-cycle disabled, device not found by portPath
Opened USB CDC port /dev/cu.usbmodem1102 for 0a10aced202194944a04b000
        ✓ USB_16_DeviceEndBeginStress (34.6s)
        ✓ USB_17_FinalCdcSanity

  USBSerial
    msom
      systemThread=enabled
Opened USB CDC port /dev/cu.usbmodem1102 for 0a10aced202194944a04b000
        ✓ USBSERIAL_00_RingBufferHelperIsSane (4.9s)
        ✓ USBSERIAL_01_SerialDoesNotDeadlockWhenInterruptsAreMasked
        ✓ USBSERIAL_02_ReadWrite
        ✓ USBSERIAL_03_ReadWriteVerifiesHostData
        ✓ USBSERIAL_04_isConnectedInitially
        ✓ USBSERIAL_05_ClosedPortWritesFailWithoutBlocking
        ✓ USBSERIAL_06_isConnectedDetectsReopenedPort
        ✓ USBSERIAL_07_EndBeginWhilePortIsClosed (4.3s)
        ✓ USBSERIAL_08_RxBufferSetup (1.1s)
        ✓ USBSERIAL_09_RxBufferFillsCompletelyFirst
        ✓ USBSERIAL_10_RxBufferFillsCompletelySecond
        ✓ USBSERIAL_11_RxBufferFillsCompletelyThird
        ✓ USBSERIAL_12_NonBlockingWriteStressSetup
        ✓ USBSERIAL_13_NonBlockingWriteStressHandlesBackpressure
        ✓ USBSERIAL_14_BlockingWriteStressWritesEveryBuffer (2.3s)
        ✓ USBSERIAL_15_NonBlockingCharacterWriteStressSetup
        ✓ USBSERIAL_16_NonBlockingCharacterWriteStressHandlesBackpressure (1.8s)
        ✓ USBSERIAL_17_BlockingCharacterWriteStressWritesEveryByte (16.3s)
        ✓ USBSERIAL_18_DeviceReceiveStressSetup
        ✓ USBSERIAL_19_DeviceReceiveStress (2.8s)


  38 passing (2m)
```

`user/tests/app/usb_ctrl_request predates the test runner and never runs`
Do we want to delete this app since its effectively been replaced?

hal/shared/serial_usb_stream.cpp
avtolstoy: Yup, looks like this is broken. Will fix and add tests, as it wasn't caught. Good catch.

hal/shared/serial_usb_stream.cpp
avtolstoy: Kinda weird, I admit. I'll take another look at peek/read in particular

hal/shared/serial_usb_stream.cpp
scott-brust: There aren't any tests for `SerialUSBStream` right now, right? Its only used for USB tethering, which have their own tests on top of it I suppose.

avtolstoy: yeah, but `HAL_USB_USART_Peek_Buffer()` and others are testable.

PR #2940, opened 2026-08-15
DVOS-35 [quectel] Fixes recovery of AT lockup when zero data while "CONNECTED" issue [sc-139209]

technobly:
# Problem

- A valid cellular connection with no-data due to IPX routing error at the serving cell provider results ultimately in triggering a bug in the Quectel EG91NAX modem firmware that kills the AT pre-processor, used over DLCI 1 muxer channel, while the muxer channel control is still responsive, PPP data channel and keep alive control channel are still responsive.  AT commands no longer are responded to by the modem.  The device is stuck in this condition with no way to recover other than a modem power cycle.

# Solution

- [quectel][ublox] detect and recover from an unresponsive AT interface

  - A Quectel EG91NAX firmware bug (present in both deployed and the latest available firmware, EG91NAXGAR07A03M1G_A0.005.A0.005) stops servicing the AT channel (DLCI 1) while continuing to service the muxer control channel (DLCI 0) and the PPP data channel (DLCI 2). Data keeps flowing, so `connectionState()` legitimately reports `CONNECTED`, and nothing in Device OS notices or recovers: the device stops responding to any `cellular_hal` call and never comes back without a power cycle.

  - Nothing detects it today. While `CONNECTED` the client issues no AT commands at all, everything ahead of the `connState_` != `CONNECTING` gate in `processEventsImpl()` is either one-shot (`checkRunningImsi()`, `configurePlmn()`) or a no-op. `parserError_` is therefore never set and `checkParser()`, which already implements the right escalation, is never reached. The muxer cannot see it either: both TEST and MSC keepalives are DLCI 0 control frames that the peer's multiplexer answers without involving its AT processor. Confirmed under gdb during a 49 minute stall with the MSC keepalive active, `keepAlivesMissed_` was still 0.

  - Adds `checkAtWhileConnected()`, a periodic AT probe that runs while `CONNECTED`. 3s timeout every 35s, declared on the 4th consecutive failure, so ~105s. The interval is set just above the 90s default command timeout in case a command uses a shorter one. The probe is kept out of the AT log, and deliberately does not use `CHECK_PARSER*`, which would set `parserError_` and route the next command into `checkParser()` -> `waitReady()` -> `modemHardReset()` instead of the deliberate sequence below.

  - Recovery is contained in the NCP client, in the same style as the existing network registration timeout. On declaration, `processEventsImpl()` calls a new `recoverModem()` and returns `SYSTEM_ERROR_TIMEOUT`. `PppNcpNetif::loop()` then brings the modem back through `on()` and `upImpl()` on its next pass, which is the existing chain, unchanged. No network interface changes are needed.

  - `recoverModem()` is shared with the network registration timeout, which previously inlined its own copy of the sequence. Three differences from the old inline version, all of which the registration timeout now also gets:

    1. `ncpState(OFF)` before `muxer_.stop()`, so the muxer channel state callback does not disable us unnecessarily. This also clears `ready_` and drives `connectionState(DISCONNECTED)`, taking PPP down before the muxer disappears underneath it. This part was already correct, kept as is.
    2. The muxer stop is bracketed by `serial_->enabled(false/true)`, the way `off()` does it, so it stops non-gracefully. Without this it closes every channel and waits out a CLD exchange with a modem that is not answering, measured at 1072ms of dead time. With it, 50ms.
    3. `apduChannelTimer_.stop()` and `apduChannel_` = 0, again as `off()` does. The modem is gone, so the inactivity timer must not fire and try to close the channel with AT commands. Quectel only, u-blox has no APDU channel.

  - It calls `modemPowerOff()` directly rather than `off()`, because `off()` starts with `modemSoftPowerOff()`, which asks an AT interface we have just proven dead to power itself down. On Quectel that is `AT+QPOWD` retried six times, on u-blox a single `AT+CPWROFF` at the 90s default timeout.

  - Also quietens `dataChannelWrite()`: once we are OFF the muxer channel is gone but PPP keeps writing until the interface is brought down, which produced 314 ERROR lines in 15.5s during recovery (under a heavy load soak test, but a real system will still see log spam).

  - The probe stands down while the device is asleep. `urcs(false)` suspends the AT muxer channel, so a probe issued there cannot be answered and would declare a false fault. Both clients now track this in `sleepUrcsDisabled_`, set in `urcs()`. On u-blox this is the existing `sleepNoPPPWrite_` renamed: it already marked exactly this state for the PPP write path, and now gates the probe as well, so there is one flag rather than two in lockstep. On Quectel the flag is new, and is set on both branches of `urcs()`, including the BG95 QINDCFG path where the AT channel stays up: a probe there would be answered, but it would wake the module, which is the thing sleep is trying to avoid.

  - The probe and recovery are enabled on every Quectel and u-blox part except R510 (see below), not just EG91NAX. Only EG91NAX has been seen to wedge, so everywhere else this is a safety net rather than a fix for a known fault, and the rig run below shows it does not false positive on the others.

  - The u-blox port is identical with one exception: R510 is excluded from the probe entirely. It runs `UPSV=1` and drops to low power after ~9s of UART idle, and while `CONNECTED` the muxer keepalive is disabled and the LCP echo stretched to 240s specifically to let it get there. The probe is a second, independent wake source with no phase relationship to the echo, so it costs roughly another 9s awake per period whatever interval it is given. Matching the intervals does not fix that, because nothing aligns the phase: the two timers free-run on different threads and drift, and at 180 degrees apart the longest sleep halves from ~231s to ~120s. The wedge has only ever been seen on EG91NAX, so that coverage is not worth a measurable power regression on a part that has never exhibited it.

  - The guard also stands down while `firmwareUpdateR510_` is set, because an R510 firmware install stops answering AT for up to 300s by design while `connState_` is still `CONNECTED`. That clause is redundant given the R510 exclusion above, since its only setter returns early on anything but R510. It is kept as documentation, and in case R510 probing is ever reinstated behind a phase-locked scheme.

- [hal] `quectel_ncp_client`: reset connection state on every `disconnect()` exit path

  - `resetRegistrationState()` and `connectionState(DISCONNECTED)` sat at the end of `disconnect()`, after `CHECK(checkParser())` and `CHECK_PARSER(execCommand("AT+CFUN=0,0"))`. Both early-return on failure, so if the modem stops answering AT commands, the teardown was requested but never recorded: `connState_` stayed `CONNECTED` with the modem in an unknown state.

  - That stale value disables the whole recovery chain in `PppNcpNetif::loop()`: all three branches are keyed on it, so `upImpl()` (needs `DISCONNECTED`), `downImpl()` and `off()` (needs `DISCONNECTED` before powering down) all become unreachable. The interface looks healthy and has no way out, which is one signature seen in the field: a device stopped passing data and did not recover without a power cycle.

  - Moved into a `SCOPE_GUARD` placed before `checkParser()`. `sara_ncp_client` already had this, no change needed there.

  - `disconnect()` also picks up the rest of `sara_ncp_client`'s behaviour, which it had drifted from. Two gaps, both traceable to SARA-only commits that never got a Quectel counterpart:

    1. No forced AT check. `checkParser()` short-circuits when `ready_` is true and `parserError_` is 0, which is exactly the state a wedge leaves behind, so `AT+CFUN=0,0` went out to a dead interface and blocked for the full 90s parser default. The forced 1s AT, added to SARA in `d981dcc0b` (an R510 UPSV commit), sets `parserError_` so `checkParser()` escalates instead: ~5s of probing, then `waitReady()` falls through to `modemHardReset()`, which is the recovery a wedged modem needs anyway. Note this means `disconnect()` can now hard reset the modem and leave ncpState at OFF; that is inherited from SARA rather than new behaviour, but it is heavier than the name suggests.
    2. Wrong CFUN timeout. The raw `execCommand("AT+CFUN=0,0")` used the 90s parser default, and was the only one of 14 CFUN call sites in the Quectel client not using `QUECTEL_CFUN_TIMEOUT`. The constant predates the line by a month; it was simply never picked up. Replaced with `setModuleFunctionality(CellularFunctionality::MINIMUM, false)`, matching SARA, which routes through the helper that already carries the 3 minute timeout. Identical on the wire, since MINIMUM is 0 and the helper formats `AT+CFUN`=%d,0. The Quectel declaration also gains the `check = false` default SARA already had, so the two signatures match. This drops the leftover `(void)r` and commented-out `CHECK_TRUE`, and means `disconnect()` now propagates parser errors it previously swallowed; the `SCOPE_GUARD` still tears down connection state on every path.

- [ublox] give R510 a longer AT wake budget after a parser error

  - R510 runs `UPSV=1`. The module sleeps its UART after ~9.2s of idle and needs a wake character, consuming the first one, so a command issued cold after an idle period goes unanswered. `waitReady()`'s parser-error branch allowed 2000ms, which is 2-3 AT attempts at the 1s retry period. Measured recoveries run to ~7.6s, so the modem was being hard reset while it was alive and about to answer.

  - Raised to 15s for R510 only, via `UBLOX_R510_UPSV_WAKE_TIMEOUT`, leaving every other part on the existing 2000ms. That is roughly 2x the worst recovery observed. Keyed on `ncpId()` rather than on UPSV actually being enabled, which is only equivalent because `initReady()` enables it unconditionally for R510.

  - `checkRuntimeState()` is deliberately left at 2000ms and carries a note saying so. Widening it was tried at 30s and made things strictly worse: across three affected boots the modem answered none of 30 one-per-second ATs and still needed the hard reset, so the only effect was pushing recovery from 15s out to 43s. Boot-time unresponsiveness is a different failure from the cooldown wake, where the modem always answers within 2-7s.

- [test] update and re-enable `wiring/no_fixture_power_saving`

  - Every test in this suite was disabled in `3387462de` with "TODO: Find a deterministic way to get these tests to pass". The skips are removed and the two sources of non-determinism are addressed.

  - The CTS measurement was wrong rather than merely noisy. In low power the line idles HIGH and blips LOW for only ~26ms every 1.28s (confirmed on a scope), and `pulseIn()` blocks the app thread while it waits for an edge, so any blip landing while the system thread has the CPU was simply lost. It also only accounted for the pulse widths it happened to catch, not the whole window. Replaced with a GPIOTE ISR that timestamps every transition, and a median rising-to-rising period rather than a mean, since a missed blip merges two periods into one of ~2x nominal and drags a mean out of band. Against the scope the ISR now reports median=1279ms with p25/p75 inside 1275..1284, and edge counts matching 2 per period exactly.

  - Two traps worth knowing, both commented at the point of use. `pinReadFast()` is not ISR-safe on its first call: `fastPinGetPinmap()`'s function-local static initialises via `__cxa_guard_acquire`, which asserts in interrupt context and panics SOS 10, so it is primed from thread context before the interrupt is attached. And the capture is attached once in setup rather than around each measurement, because `hal_interrupt_detach_ext()` ends with `hal_pin_set_function(pin, PF_NONE)`, clearing the pinmap's record that CTS belongs to the UART.

  - Whether the R510 enters low power is no longer asserted. It depends on paging and registration state the test does not control, so gating on it only made the suite flaky without saying anything about Device OS. It is reported instead, over the mailbox so it lands in the runner console rather than only on Serial1, alongside a diagnostic line carrying median, quartiles, edge count, span and wakes/min. What is asserted in its place is `AT+UPSV? == 1`, which is deterministic: it is Device OS's own configuration rather than modem behaviour, and it fails exactly when `initReady()` stops enabling UPSV.

  - `POWER_SAVING_06` (wake from ULP sleep by network) could not pass as written, for a structural reason rather than a flaky one. The runner runs each device test inside an awaited mocha `beforeEach` hook, and `waitTest()` only returns early when the device reports `STATUS_WAITING`, which requires an actual reset. `ULTRA_LOW_POWER` resumes in place, so the device stays `STATUS_RUNNING` for the whole sleep and the spec body that issues the waking function call cannot run until the device has already woken. `RESET_PENDING` does not change this; its job is telling the runner USB is about to detach, which is why the `HIBERNATE` tests in `sleep20` work (their wake is self-contained) and its ULP cellular test works (a BLE fixture peer supplies the stimulus).

  - Restructured so the wake is armed from test 05's spec body, not awaited, and gated on a mailbox message the device sends immediately before sleeping. Test 05's body runs before test 06_1's hook, so the call is already in flight while the device sleeps. The signal goes over the mailbox rather than a publish because a publish is UART traffic that would restart the modem's UPSV idle timer. The low power measurement was also dropped from 06 entirely: waking on inbound traffic does not depend on it, tests 01 and 02 already cover it, and it cost ~30s per run. Sleep duration raised to 120s to give the call room to land mid-sleep.

# Scope

- This only bites devices that do not otherwise talk to the modem. An application calling `Cellular.RSSI()` or `Cellular.command()` sets `parserError_` on the first timeout, and the second call escalates through `checkParser()` -> `waitReady()` -> `modemHardReset()` within roughly two command timeouts. That path exists today and is unchanged. The probe is what gives the same outcome to a PPP data-only device, which is the field signature.

# Validation

Earlier soak runs were on a slightly different shape of this change, where recovery was driven from PppNcpNetif rather than contained in the NCP client. The detector and its timing are the same, so the detection numbers carry over. Recovery times are re-measured below.

The review changes (sleep stand-down, R510 exclusion, `disconnect()` alignment) landed after the runs below and are compile-verified only. None of them touch the detector's timing or the recovery sequence on EG91NAX, so the detection and recovery numbers carry over unchanged. What is not re-measured is the `disconnect()` path against a wedged modem, which now escalates to a hard reset in ~5s instead of blocking on CFUN for 90s. R510 no longer probes at all, so the power question that motivated the exclusion needs no hardware validation.

The R510 wake budget and the power saving test are independent of the EG91NAX detector and recovery path, and are validated separately below.

- Validated the R510 wake budget and the updated suite together on the HIL rig, boron BRN404X (R510, modem firmware 03.30), both systemThread=disabled and enabled. All tests pass in both modes. Low power was detected on the first attempt in every measurement, at median 1279-1280ms and 46-47 wakes/min, with `AT+UPSV?` reading 1,2000,1 throughout.

- The wake budget was measured over a separate multi-cycle soak on esomx and boron before being set. Recoveries after a parser error clustered at 2.0-7.6s and always succeeded; with the old 2000ms budget every one of those would have been a hard reset on a live modem. `V_INT` stayed high throughout, and each of those hard resets then recovered on the first AT afterwards, which is what identified them as unnecessary.

- Builds clean with no new warnings on boron and esomx. Note that `device-os-test build` only compiles the user application, so a full modules build is needed to cover the HAL change.

- Validated on 6.4.1 over an 8.5 hour soak: 3 wedges, 3 recoveries, 94 data-mode transitions, no permanent failures. Recovery took 340/338/344s, in every case via the 5-minute cloud-failure network reset. The probe was deliberately stretched (60s x 10) for this run so the reset would win.

- Validated on 6.5.0 over a 2.4 hour soak with the probe at production values: 4 wedges, 4 recoveries, 29 data-mode transitions, no permanent failures. All four recovered via the probe rather than the cloud-failure reset (Hard resetting: 0), tripping on the 3rd consecutive failure every time, detect -> resume taking 28s in all four cases. This run used 30s/3/2s, the merged 35s/3/3s moves detection from ~90s to ~105s to sit just above the default AT command timeout.

- Validated on 6.5.0 with the probe stretched to 60s x 10 behaved identically to 6.4.1 over 6 hours: 6 wedges, 338-343s recoveries. 6.5.0's PPP TX-queue rework does not affect teardown.

- Validated on 6.5.0 with the soak removed entirely and the device pointed at an unreachable cloud IP, i.e. the shipping configuration with the probe as the sole detector: 13 data-mode transitions, 2 wedges, 2 recoveries, no MCU resets. Both recoveries were identical to within 100 ms, declared at 105.0s, a 15.5s modem power cycle, and data restored 29.6s after detection (134.6s from the first probe failure). 6 probe failures total = exactly 2 x 3, with no isolated failures, so no false positives.

- No false positives: across a separate 10.3 hour run, all 26 "AT unresponsive" declarations fell inside a genuine wedge, and all 78 probe failures were part of a streak that escalated (78 = 26 x 3), no isolated or spurious probe failures.

- Validated on the final shape on the HIL rig, b5som, both systemThread=disabled and enabled. Both reproduced and recovered, one of them twice because the fault recurred on the first transition after recovery. Declaration cadence was 105.000s from the first probe failure, unchanged. Detect -> data restored was 28.9s and 29.4s, against 30.4s before the non-graceful muxer stop was added. Breakdown of one recovery: 29ms to start the muxer stop, 50ms to stop it, 15.6s modem power off, 9.7s power on, then registration and PPP.

- Builds clean with no new warnings on every platform with a cellular NCP client: boron and esomx (u-blox), b5som, msom, tracker, trackerm and electron2 (Quectel).

- `off()` is unchanged. The refactor folded `off()` into `configModemPowerState(ModemOff)`, so the same test app was run on the branch and on released 6.5.0 and the logs compared from "Request to power off interface 4" onward. Identical line for line, the only difference being the modem's own power off duration (8584ms vs 8573ms). That includes the pre-existing second pass through `off()`: the muxer channel 0 close fires `disable()` during the first pass, so `ncpState(OFF)` is swallowed by its own `DISABLED` guard and the netif's next `enable()` re-runs `off()` 101ms later (102ms on the branch), which is where NCP state changed: 0 actually lands. Both builds do this.

- Run across the HIL rigs. Reproduced and recovered on msom "Muon" (B504e/EG91NAX) and B5SoM (B504e/EG91NAX). No reproduction on bsom, boron and esomx (R410/R510), tracker (BG96), msom (BG95-M5) or msom (BG95-S5), and no probe declarations on any of them either, so no false positives across five other modem types. The fault therefore looks specific to EG91NAX, and the probe is preventative on everything else.

# Test

- run integration test `wiring/no_fixture_cellular_long_running`
- run integration test `wiring/no_fixture_power_saving` (R510 only, skips itself on other NCPs)

### DIFF
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -378,44 +378,71 @@ int QuectelNcpClient::on() {
     return SYSTEM_ERROR_NONE;
 }
 
-int QuectelNcpClient::off() {
-    const NcpClientLock lock(this);
-    if (ncpState_ == NcpState::DISABLED) {
-        return SYSTEM_ERROR_INVALID_STATE;
+// Take the modem down. ModemOff is the ordinary off() sequence, the other reasons are faults where
+// the modem will not answer an AT command, so the sequence never issues one.
+// Caller must hold the client lock.
+int QuectelNcpClient::configModemPowerState(ModemPowerReason reason) {
+    const bool recovering = (reason != ModemPowerReason::ModemOff && reason != ModemPowerReason::Unknown);
+    int r = SYSTEM_ERROR_NONE;
+
+    if (recovering) {
+        LOG(WARN, "Resetting the modem due to %s", reason == ModemPowerReason::AtUnresponsive ?
+                "an unresponsive AT interface" : "the network registration timeout");
+        // We are going into an OFF state immediately before stopping the muxer
+        // otherwise the muxer channel state callback will disable us unnecessarily.
+        // off() deliberately leaves this until the end.
+        ncpState(NcpState::OFF);
+    } else {
+        // Try using AT command to turn off the modem first.
+        r = modemSoftPowerOff();
     }
-    // Try using AT command to turn off the modem first.
-    int r = modemSoftPowerOff();
 
     // Disable ourselves/channel, so that the muxer can potentially stop faster non-gracefully
     serial_->enabled(false);
     muxer_.stop();
     serial_->enabled(true);
 
-    // Disable voltage translator
-    modemSetUartState(false);
-
-    if (!r) {
-        LOG(TRACE, "Soft power off modem success");
-        // WARN: We assume that the modem can turn off itself reliably.
-    } else {
-        // Power down using hardware
+    if (recovering) {
+        // Not modemSoftPowerOff(), it would retry AT+QPOWD six times against a dead interface
         if (modemPowerOff() != SYSTEM_ERROR_NONE) {
-            LOG(ERROR, "Failed to turn off");
+            modemHardReset(true);
         }
-        // FIXME: There is power leakage still if powering off the modem failed.
+    } else {
+        // Disable voltage translator
+        modemSetUartState(false);
+
+        if (!r) {
+            LOG(TRACE, "Soft power off modem success");
+            // WARN: We assume that the modem can turn off itself reliably.
+        } else {
+            // Power down using hardware
+            if (modemPowerOff() != SYSTEM_ERROR_NONE) {
+                LOG(ERROR, "Failed to turn off");
+            }
+            // FIXME: There is power leakage still if powering off the modem failed.
+        }
+
+        // Disable the UART interface.
+        LOG(TRACE, "Deinit modem serial.");
+        serial_->on(false);
+
+        ready_ = false;
+        ncpState(NcpState::OFF);
     }
 
-    // Disable the UART interface.
-    LOG(TRACE, "Deinit modem serial.");
-    serial_->on(false);
-
-    ready_ = false;
-    ncpState(NcpState::OFF);
-
+    // The modem is gone, the inactivity timer must not fire and try to close the channel
     apduChannelTimer_.stop();
     apduChannel_ = 0;
 
     return SYSTEM_ERROR_NONE;
+}
+
+int QuectelNcpClient::off() {
+    const NcpClientLock lock(this);
+    if (ncpState_ == NcpState::DISABLED) {
+        return SYSTEM_ERROR_INVALID_STATE;
+    }
+    return configModemPowerState(ModemPowerReason::ModemOff);
 }
 
 int QuectelNcpClient::enable() {
@@ -2543,25 +2570,6 @@ int QuectelNcpClient::checkRunningImsi() {
     return 0;
 }
 
-// Power cycle the modem without issuing any AT command, for the faults where it will not answer one.
-// Caller must hold the client lock. Leaves the client OFF, PppNcpNetif::loop() brings it back up.
-void QuectelNcpClient::recoverModem() {
-    // We are going into an OFF state immediately before stopping the muxer
-    // otherwise the muxer channel state callback will disable us unnecessarily.
-    ncpState(NcpState::OFF);
-    // Disable ourselves/channel, so that the muxer can stop faster non-gracefully
-    serial_->enabled(false);
-    muxer_.stop();
-    serial_->enabled(true);
-    // Not off(), its modemSoftPowerOff() would retry AT+QPOWD six times first
-    if (modemPowerOff() != SYSTEM_ERROR_NONE) {
-        modemHardReset(true);
-    }
-    // The modem is gone, the inactivity timer must not fire and try to close the channel
-    apduChannelTimer_.stop();
-    apduChannel_ = 0;
-}
-
 // Periodic AT probe while CONNECTED. Returns false once the interface is declared unresponsive.
 //
 // Deliberately not CHECK_PARSER*, that would set parserError_ and route the next command into
@@ -2606,10 +2614,9 @@ int QuectelNcpClient::processEventsImpl() {
 
     // Must stay above the connState_ != CONNECTING early return below, the point is to run while CONNECTED
     if (!checkAtWhileConnected()) {
-        LOG(WARN, "Resetting the modem due to an unresponsive AT interface");
         // connectionState(CONNECTED) clears this too, but nothing does if we never reconnect
         atProbeFailStreak_ = 0;
-        recoverModem();
+        configModemPowerState(ModemPowerReason::AtUnresponsive);
         return SYSTEM_ERROR_TIMEOUT;
     }
 
@@ -2645,8 +2652,7 @@ int QuectelNcpClient::processEventsImpl() {
     CHECK_PARSER(parser_.execCommand("AT+QENG=\"servingcell\""));
 
     if (connState_ == NcpConnectionState::CONNECTING && millis() - regStartTime_ >= registrationTimeout_) {
-        LOG(WARN, "Resetting the modem due to the network registration timeout");
-        recoverModem();
+        configModemPowerState(ModemPowerReason::RegTimeout);
         return SYSTEM_ERROR_TIMEOUT;
     }
 
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -114,6 +114,12 @@ const auto QUECTEL_NCP_PPP_CHANNEL = 2;
 
 const auto QUECTEL_NCP_SIM_SELECT_PIN = 23;
 
+// AT probe: 3s timeout every 35s, declared on the 4th consecutive failure, so ~105s.
+// Just above the 90s default command timeout, in case a command uses a shorter one.
+const auto AT_PROBE_INTERVAL = 35000;
+const auto AT_PROBE_TIMEOUT = 3000;
+const auto AT_PROBE_MAX_INTERVALS = 3u;
+
 const unsigned REGISTRATION_CHECK_INTERVAL = 15 * 1000;
 const unsigned REGISTRATION_TIMEOUT = 10 * 60 * 1000;
 const unsigned REGISTRATION_INTERVENTION_TIMEOUT = 15 * 1000;
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -454,14 +454,18 @@ int QuectelNcpClient::disconnect() {
     if (connState_ == NcpConnectionState::DISCONNECTED) {
         return SYSTEM_ERROR_NONE;
     }
+
+    // Tear the connection state down unconditionally, on every exit path
+    SCOPE_GUARD({
+        resetRegistrationState();
+        connectionState(NcpConnectionState::DISCONNECTED);
+    });
+
     CHECK(checkParser());
     const int r = CHECK_PARSER(parser_.execCommand("AT+CFUN=0,0"));
     (void)r;
     // CHECK_TRUE(r == AtResponse::OK, SYSTEM_ERROR_UNKNOWN);
 
-    resetRegistrationState();
-
-    connectionState(NcpConnectionState::DISCONNECTED);
     return SYSTEM_ERROR_NONE;
 }
 
--- a/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
+++ b/hal/network/ncp_client/quectel/quectel_ncp_client.cpp
@@ -492,6 +492,9 @@ int QuectelNcpClient::updateFirmware(InputStream* file, size_t size) {
 }
 
 int QuectelNcpClient::dataChannelWrite(int id, const uint8_t* data, size_t size) {
+    // Once disable() has been called the muxer channel is gone, but PPP does not know that yet and keeps
+    // writing until the interface is brought down. This check prevents log spam.
+    CHECK_TRUE(ncpState_ == NcpState::ON, SYSTEM_ERROR_INVALID_STATE);
     // Just in case perform some state checks to ensure that LwIP PPP implementation
     // does not write into the data channel when it's not supposed do
     CHECK_TRUE(connState_ == NcpConnectionState::CONNECTED, SYSTEM_ERROR_INVALID_STATE);
@@ -2337,6 +2340,9 @@ void QuectelNcpClient::connectionState(NcpConnectionState state) {
     connState_ = state;
 
     if (connState_ == NcpConnectionState::CONNECTED) {
+        // A streak from a previous connection must not carry into this one.
+        // atProbeTime_ is deliberately left stale so the first probe runs immediately.
+        atProbeFailStreak_ = 0;
         inFlowControl_ = false;
         // Open data channel and resume it just in case
         int r = muxer_.openChannel(QUECTEL_NCP_PPP_CHANNEL);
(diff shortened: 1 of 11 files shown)


hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -2527,13 +2538,52 @@ int QuectelNcpClient::checkRunningImsi() {
     return 0;
 }
 
+// Periodic AT probe while CONNECTED
+//
+// Deliberately does not call parserError() via CHECK_PARSER*. That would route into checkParser() -> waitReady() ->
+// modemHardReset() from inside processEvents(), bypassing the network interface's orderly teardown. The
+// streak is published via atInterfaceUnresponsive() instead, and PppNcpNetif acts on it.
+void QuectelNcpClient::checkAtWhileConnected() {
```
avtolstoy: should probably be added similarly to sara ncp client too.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -454,14 +454,18 @@ int QuectelNcpClient::disconnect() {
     if (connState_ == NcpConnectionState::DISCONNECTED) {
         return SYSTEM_ERROR_NONE;
     }
+
+    // Tear the connection state down unconditionally, on every exit path
+    SCOPE_GUARD({
```
avtolstoy, line 492: :+1:

hal/network/lwip/cellular/pppncpnetif.cpp
```diff
@@ -169,6 +170,26 @@ void PppNcpNetif::loop(void* arg) {
...
+        // Nothing else will do this. connectionState() reports CONNECTED (PPP data really is flowing on
+        // its own muxer channel), which gates off upImpl(), downImpl() and off() alike, and no muxer
+        // keepalive can see the fault because TEST and MSC are both control-channel frames the peer's
+        // multiplexer answers itself. The 5-minute cloud-failure reset does recover it, but only when
+        // the cloud is failing, and at worst a full reset period later.
+        //
+        // disable() is lock-free and aborts the stuck command by disabling the muxer streams; the
+        // enable() above then runs off() on the next pass, and expectedNcpState_ == ON brings it back
+        // up. That is the existing recovery chain, simply made reachable.
+        if (self->celMan_->ncpClient()->ncpState() == NcpState::ON &&
```
avtolstoy: Could just error out of `processEvents()` perhaps? And honestly I think it would be cleaner to contain this logic inside the ncp client and not spill the specifics into the ncp netif - it's pretty simple and only cares about getting incoming events from ifapi and other layers and translating into ncp client calls/maintaining expected state.

We already have precedent with e.g. `registrationTimeout_`. Why not essentially do the same thing?

You anyway need `processEvents()` to run to detect unresponsive AT interface. If this thread is stuck in some AT command waiting a response which will not happen, this `disable()` call etc will not run anyway until that gets unstuck. If it's some other thread (e.g. app) - same thing - until the calling thread releases the mutex, `processEvents()` isn't going to run and isn't going to detect anything.

So, with that said, `processEvents()` is the single source of truth for running a periodic comms with the modem over the AT channel - so it makes sense to contain everything in there? Detect unresponsive modem - perform an immediate reset sequence etc.

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -2527,13 +2543,76 @@ int QuectelNcpClient::checkRunningImsi() {
     return 0;
 }
 
+// Power cycle the modem without issuing any AT command, for the faults where it will not answer one.
+// Caller must hold the client lock. Leaves the client OFF, PppNcpNetif::loop() brings it back up.
+void QuectelNcpClient::recoverModem() {
```
avtolstoy: I'd prefer if we could combine this with `off()`. e.g. either a common function called from here/processEvents and off(), or off() with an argument to perform power down without software reset or we add a conditional to whether we can perform soft power off or not. Makes sense?

Otherwise we have essentially the same logic duplicated in two places.

technobly: Heh, yeah... was considering something like this.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -2699,13 +2683,112 @@ int SaraNcpClient::checkRunningImsi() {
...
+    }
+
+    return SYSTEM_ERROR_NONE;
+}
+
+// Periodic AT probe while CONNECTED. Returns false once the interface is declared unresponsive.
+//
+// Deliberately not CHECK_PARSER*, that would set parserError_ and route the next command into
+// checkParser() -> waitReady() -> modemHardReset() instead of the reset sequence run by the caller.
+bool SaraNcpClient::checkAtWhileConnected() {
```
avtolstoy, line 2756: This should also not probe while URCs are disabled for sleep: `cellular_urcs(false)` suspends the AT channel (flow control) so the AT probes will not get any response. We already have `sleepNoPPPWrite_` for exactly this state, so perhaps reuse that or something along those lines?

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -2527,13 +2570,56 @@ int QuectelNcpClient::checkRunningImsi() {
     return 0;
 }
 
+// Periodic AT probe while CONNECTED. Returns false once the interface is declared unresponsive.
+//
+// Deliberately not CHECK_PARSER*, that would set parserError_ and route the next command into
+// checkParser() -> waitReady() -> modemHardReset() instead of the reset sequence run by the caller.
+bool QuectelNcpClient::checkAtWhileConnected() {
```
avtolstoy, line 2582: Same sleep issue here as in sara ncp client (except BG95 which uses QINDCFG.

hal/network/ncp_client/sara/sara_ncp_client.cpp
```diff
@@ -156,6 +156,12 @@ const int UUFWINSTALL_COMPLETE = 128;
 const int UBLOX_WAIT_AT_RESPONSE_WHILE_UUFWINSTALL_TIMEOUT = 300000;
 const int UBLOX_WAIT_AT_RESPONSE_WHILE_UUFWINSTALL_PERIOD = 10000;
 
+// AT probe: 3s timeout every 35s, declared on the 4th consecutive failure, so ~105s.
+// Just above the 90s default command timeout, in case a command uses a shorter one.
+const auto AT_PROBE_INTERVAL = 35000;
```
avtolstoy, line 170: Is this going to cause problems for R510? We specifically increase PPP LCP echo interval and muxer keepalive interval to allow it to go to powersave?

hal/network/ncp_client/quectel/quectel_ncp_client.cpp
```diff
@@ -454,14 +487,18 @@ int QuectelNcpClient::disconnect() {
...
         return SYSTEM_ERROR_NONE;
     }
+
+    // Tear the connection state down unconditionally, on every exit path
+    SCOPE_GUARD({
+        resetRegistrationState();
+        connectionState(NcpConnectionState::DISCONNECTED);
+    });
+
     CHECK(checkParser());
```
avtolstoy, line 503: I compared to sara ncp client which has this:
```
    // If we disconnect due to the AT interface being dead, a forced check
    // is required because ready_ is true and parserError_ is 0.
    const int r = parser_.execCommand(1000, "AT");
    if (r != AtResponse::OK) {
        parserError_ = r;
    }
    CHECK(checkParser());
```

I think it's worth sharing the behavior? Otherwise there is a chance to get stuck in CFUN for a long time?

avtolstoy (approved):

avtolstoy (approved):

Kategrode:
https://onedigi.atlassian.net/browse/DVOS-35

PR #2941, opened 2026-08-25
[build] working -jXX and ccache support; faster CI builds

avtolstoy:
### Problem

1. `make -jXX` has never worked. Recursive make relies on positional ordering of prerequisites in `all: prebuild $(MAKE_DEPENDENCIES) $(TARGET) postbuild`, which only holds for serial builds. Under `-j` the dependency libraries have no rule in the parent make instance (`No rule to make target 'libservices.a'`), submakes race each other in shared build directories and `make clean all -jXX` runs both goals in parallel, cleaning from under an ongoing build.
2. ccache has never worked either: `$(CCACHE)` is referenced in every `module.mk` recipe, but is not defined anywhere. Even when defined manually, bare `--specs=nano.specs` makes ccache bail out on every single compile (`Result: bad_compiler_arguments` - it tries to stat the specs file by its bare name), so nothing was ever cached.
3. CI builds everything single-threaded and rebuilds the same shared unit-test library sources for every one of ~37 test apps per platform. There is also no visibility into per-platform build times.
4. Bonus: CI builds have been running under GNU make 3.82 (2010) all this time - buildtools ships its own `make` into `/root/.particle/bin` which shadows the distro one.

### Solution

Build system:
1. Explicit order-only edges in `module.mk`: `$(ALLOBJ) $(LIB_DEPS) $(LINKER_DEPS): | $(MAKE_DEPENDENCIES) prebuild` and friends, so ordering no longer depends on serial make.
2. `.NOTPARALLEL` in coordinator makefiles that fan out submakes with overlapping dependency graphs (top-level, `modules/`, `main/`, prebootloader). The jobserver is still passed down, so each submake compiles at full `-j`.
3. A single make process cannot safely clean and rebuild the same files (make caches file timestamps across the clean), so coordinators split `clean` into a separate submake invocation and leaf makefiles fall back to serial goal processing when `clean` is combined with build goals.
4. `modules/<platform>/makefile` delegates matched every goal with `%:` and ran the full goal list once per goal (`clean all program-dfu` = three full builds, concurrent under `-j`). Now delegates once.
5. user-part is linked twice (`elf_fi` relinks with a generated RAM layout and replaces the ELF in place) - this flow is not parallel-safe by design, so these modules run their own recipes serially. rtl872x user-part prebuild was also deleting the intermediate ELF mid-run, which under `-j` meant make never relinked it - replaced with a forced relink via a phony prerequisite.
6. `rtl-flash` had no prerequisites at all and could flash a stale binary; `size`/`hex`/`lst` now run after `bin` (the bin recipe post-processes the ELF in place).
7. LTO link uses `-flto=auto` - parallel LTRANS where partitioning allows, no-op with `-flto-partition=one`.
8. `verbose.mk` considered any submake of a `-j` build silent, because `--jobserver-auth` in MAKEFLAGS contains an 's' :sigh:
9. ccache is auto-detected in `macros.mk` (opt out with `CCACHE=`), `--specs` is resolved to a full path so compiles are actually cacheable.

CI:
1. Release builds and device-os-test builds run with `-j$(nproc)`.
2. Buildpack builder bumped to 0.3.0: 26.04 base image used with ccache and removes the buildtools make 3.82 shim, so the distro make is used.
3. Per-platform build times in the Slack report.

Local numbers (16 cores, M SoM):

```
make -C modules PLATFORM=msom APP=tinker clean all       7m03s
make -C modules PLATFORM=msom APP=tinker clean all -j16  1m04s
```

Binaries are bit-identical between serial and `-jXX` builds (`system-part1.bin`/`tinker.bin` sha256-compared on both Gen3  and Gen 4 platforms).

CircleCI per-platform times (serial baseline -> this branch):

```
Argon     38m04s -> 20m45s
Boron     38m36s -> 20m45s
BSoM      46m18s -> 20m32s
B5SoM     42m59s -> 21m17s
Tracker   37m41s -> 18m57s
TrackerM  38m35s -> 18m48s
ESomX     39m08s -> 20m37s
P2        36m09s -> 19m14s
MSoM      34m16s -> 17m34s
Electron2 40m15s -> 19m49s
```

### Steps to Test

- `make PLATFORM=<platform> clean all -jXX` out of `modules/`, `main/`, `bootloader/`, `bootloader/prebootloader/`, also in combination with `program-dfu`
- Incremental `-jXX` builds, `TEST=`/`APP=` builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,7 +240,7 @@ jobs:
             TIMESTAMP_FILE=timestamp_$RANDOM
             mkdir -p $HOME/workspace/release
             date +%s > $HOME/workspace/release/$TIMESTAMP_FILE
-            apk -q update && apk -q add bash curl xz
+            apk -q update && apk -q add bash
       - run:
           name: "Fetch dependencies"
           shell: /bin/bash
@@ -268,19 +268,7 @@ jobs:
             export BUILD_PLATFORM="<< parameters.platform >>"
             export BUILDPACK_NORM=1
             export BUILD_PLATFORM_NORMALIZED=$(echo "$BUILD_PLATFORM" | sed 's/ /_/g')
-            # stage a modern ccache into the build container: the one shipped in ubuntu 20.04 (3.7.7)
-            # cannot share cache entries between objects with different output paths
-            CCACHE_ARG=""
-            if curl -fsSL --retry 3 -o /tmp/ccache.tar.xz https://github.com/ccache/ccache/releases/download/v4.11.3/ccache-4.11.3-linux-x86_64.tar.xz \
-                && echo "7766991b91b3a5a177ab33fa043fe09e72c68586d5a86d20a563a05b74f119c0  /tmp/ccache.tar.xz" | sha256sum -c -s -; then
-              tar xf /tmp/ccache.tar.xz -C /tmp
-              docker volume create ccache_ext
-              docker create -v ccache_ext:/ccache-ext --name ccseed alpine true
-              docker cp /tmp/ccache-4.11.3-linux-x86_64/ccache ccseed:/ccache-ext/ccache
-              docker rm ccseed
-              CCACHE_ARG="--volume ccache_ext:/ccache-ext --env CCACHE=/ccache-ext/ccache"
-            fi
-            export EXTRA_ARG="--env DEVICE_OS_CI_API_TOKEN=$DEVICE_OS_CI_API_TOKEN --env CI_BUILD_RELEASE=1 --name build --env VERSION=$TAG --env MAKE_JOBS=nproc ${CCACHE_ARG}"
+            export EXTRA_ARG="--env DEVICE_OS_CI_API_TOKEN=$DEVICE_OS_CI_API_TOKEN --env CI_BUILD_RELEASE=1 --name build --env VERSION=$TAG --env MAKE_JOBS=nproc"
             mkdir -p $HOME/workspace/release/$BUILD_PLATFORM_NORMALIZED
             export BUILD_START=$(date +%s)
             $HOME/firmware-buildpack-builder/scripts/ci | tee $HOME/workspace/release/$BUILD_PLATFORM_NORMALIZED/log | tee >(grep -E '^has_failures [0-9]+ ' > $HOME/failures/has_failures); export TMPRESULT=${PIPESTATUS[0]}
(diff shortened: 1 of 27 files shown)


.circleci/config.yml
```diff
@@ -268,9 +268,23 @@ jobs:
             export BUILD_PLATFORM="<< parameters.platform >>"
             export BUILDPACK_NORM=1
             export BUILD_PLATFORM_NORMALIZED=$(echo "$BUILD_PLATFORM" | sed 's/ /_/g')
-            export EXTRA_ARG="--env DEVICE_OS_CI_API_TOKEN=$DEVICE_OS_CI_API_TOKEN --env CI_BUILD_RELEASE=1 --name build --env VERSION=$TAG"
+            # stage a modern ccache into the build container: the one shipped in ubuntu 20.04 (3.7.7)
+            # cannot share cache entries between objects with different output paths
+            CCACHE_ARG=""
+            if curl -fsSL --retry 3 -o /tmp/ccache.tar.xz https://github.com/ccache/ccache/releases/download/v4.11.3/ccache-4.11.3-linux-x86_64.tar.xz \
+                && echo "7766991b91b3a5a177ab33fa043fe09e72c68586d5a86d20a563a05b74f119c0  /tmp/ccache.tar.xz" | sha256sum -c -s -; then
```
scott-brust: This is comparing `ccache-4.11.3-linux-x86_64.tar.xz` against the known/expected hash of `7766991b91b3a5a177ab33fa043fe09e72c68586d5a86d20a563a05b74f119c0`?

I trying it manually and `sha256sum -s` seems to complain on mac, claude seems to think its wrong on linux too.
```sha256sum: invalid option -- s```

Is this silently failing? Should this be `--status``?
You want to get rid of this eventually by moving to 24.04 in any case?

File: bootloader/prebootloader/makefile [hunk 2]
```diff
@@ -17,9 +17,20 @@ SUBDIR_GOALS := $(MAKECMDGOALS)
...
+clean: $(makefiles)
+
 $(makefiles):
 	$(call,echo,'Making module $@')
+ifneq (,$(filter clean,$(SUBDIR_GOALS)))
+	$(VERBOSE)$(MAKE) -C $(dir $@) clean $(MAKE_ARGS) $(MAKEOVERRIDES)
+ifneq (,$(strip $(filter-out clean,$(SUBDIR_GOALS))))
+	$(VERBOSE)$(MAKE) -C $(dir $@) $(filter-out clean,$(SUBDIR_GOALS)) $(MAKE_ARGS) $(MAKEOVERRIDES)
+endif
+else
```
scott-brust, line 29: These checks are making sure that if `clean` is passed that `clean` finishes before `all`?

scott-brust (approved):

.circleci/config.yml
avtolstoy: This does work on CircleCI. Yes, I will get rid of it before merging by upgrading buildpack base to 24.04/26.04.

bootloader/prebootloader/makefile
avtolstoy: Yup, exactly.

PR #2944, opened 2026-09-04
[gen3] [rtl872x] Fix hal malloc deadlocks

scott-brust:
### Problem

I observed a deadlock on boot with an MSOM. 

The backtrace showed contention on the malloc mutex between system thread and user application thread. Something was trying to malloc with scheduling disabled, which resulted in the deadlock.

### Solution

Investigation showed that its possible to deadlock during SPI and I2C hal initialization if the auxiliary power setting and ethernet detection is enabled on an msom (for example, on muon configuration). The Wiznet ethernet 'probe' is deferred to the user thread. If it happens to execute while the system thread holds the malloc mutex it can dead lock.

The solution is to allocate the HAL mutexes without disabling scheduling, and then swapping the created mutexes as needed under the section with scheduling disabled. 

This problem was present in a couple of places and was fixed similarly in all of them. 

Also, added an explicit assert in the malloc lock to PANIC if this condition is detected going forward. 

### Steps to Test
This condition is easy to replicate if you stress the malloc critical section see below

### Example App
This snippet when added to user app generates deadlocks reliably. With these PR changes it does not.

```c
static os_thread_t s_hammer;
static void mallocHammer(void*) {
    for (;;) { void* p = malloc(64); free(p); }
}

void setup()
{
    // Enable configuration at least once
    SystemPowerConfiguration powerConfig = System.getPowerConfiguration();
    powerConfig.auxiliaryPowerControlPin(D7).interruptPin(A7);
    System.setPowerConfiguration(powerConfig);

    os_thread_create(&s_hammer, "hammer", OS_THREAD_PRIORITY_DEFAULT, mallocHammer, nullptr, 1024);
    SPI1.begin();
    Wire.begin();
...
}

```

### References

[Internal discussion](https://s.slack.com/archives/C8QH35A73/p1788478919392809)

### DIFF
--- a/hal/src/nRF52840/core_hal.c
+++ b/hal/src/nRF52840/core_hal.c
@@ -747,6 +747,9 @@ static void init_malloc_mutex(void) {
 
 void __malloc_lock(struct _reent *ptr) {
     if (malloc_mutex) {
+#ifdef DEBUG_BUILD
+        SPARK_ASSERT(xTaskGetSchedulerState() != taskSCHEDULER_SUSPENDED);
+#endif
         if (!xSemaphoreTakeRecursive(malloc_mutex, MALLOC_LOCK_TIMEOUT_MS)) {
             PANIC(HeapError, "Semaphore Lock Timeout");
             while (1);
--- a/hal/src/nRF52840/core_hal.c
+++ b/hal/src/nRF52840/core_hal.c
@@ -747,9 +747,9 @@ static void init_malloc_mutex(void) {
 
 void __malloc_lock(struct _reent *ptr) {
     if (malloc_mutex) {
-#ifdef DEBUG_BUILD
-        SPARK_ASSERT(xTaskGetSchedulerState() != taskSCHEDULER_SUSPENDED);
-#endif
+        if (xTaskGetSchedulerState() == taskSCHEDULER_SUSPENDED) {
+            PANIC(HeapError, "malloc while suspended");
+        }
         if (!xSemaphoreTakeRecursive(malloc_mutex, MALLOC_LOCK_TIMEOUT_MS)) {
             PANIC(HeapError, "Semaphore Lock Timeout");
             while (1);
--- a/hal/src/rtl872x/spi_hal.cpp
+++ b/hal/src/rtl872x/spi_hal.cpp
@@ -188,11 +188,21 @@ public:
     }
 
     int init() {
-        os_thread_scheduling(false, nullptr);
         if (mutex_ == nullptr) {
-            os_mutex_recursive_create(&mutex_);
+            os_mutex_recursive_t mutex = nullptr;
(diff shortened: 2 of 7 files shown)


hal/src/nRF52840/core_hal.c
```diff
@@ -747,6 +747,9 @@ static void init_malloc_mutex(void) {
 
 void __malloc_lock(struct _reent *ptr) {
     if (malloc_mutex) {
+#ifdef DEBUG_BUILD
+        SPARK_ASSERT(xTaskGetSchedulerState() != taskSCHEDULER_SUSPENDED);
```
avtolstoy: This should ideally be enabled in release builds, but should be after take fails (again ideally). You saw that when scheduler is stopped the timeout essentially does not fire?

It should probably be safe to leave enabled here then, as you are checking for SUSPENDED only (we do have allocations before scheduler is started but that's a separate NOT_STARTED state). :shrug:

Also from the UX standpoint, would be better to generate `HeapError` perhaps?

hal/src/rtl872x/spi_hal.cpp
```diff
@@ -213,7 +223,15 @@ class Spi {
 
     int deinit() {
         end();
-        os_mutex_recursive_destroy(&mutex_);
```
avtolstoy, line 216: Nit: this is not used at all, i'd probably leave as-is or get rid of deinit altogether.

avtolstoy (approved):

hal/src/nRF52840/core_hal.c
scott-brust: > This should ideally be enabled in release builds, but should be after take fails (again ideally). You saw that when scheduler is stopped the timeout essentially does not fire?

The issue is that `xSemaphoreTakeRecursive()` never returns since the tick stops incrementing and the timeout comparison is never true. Thats why it never gets to `PANIC()`. 

>It should probably be safe to leave enabled here then, as you are checking for SUSPENDED only (we do have allocations before scheduler is started but that's a separate NOT_STARTED state). 🤷

Yeah, Freertos should return `taskSCHEDULER_NOT_STARTED` if the scheduler is not started. Otherwise it will be `taskSCHEDULER_RUNNING` or `taskSCHEDULER_SUSPENDED`. 
Things seem to run fine with these checks enabled in the normal build. It does add a little overhead but not that much. 

> Also from the UX standpoint, would be better to generate HeapError perhaps?

You mean do something like this instead of `SPARK_ASSERT`?
```
if (xTaskGetSchedulerState() == taskSCHEDULER_SUSPENDED) {
    PANIC(HeapError, "malloc while suspended");
}
```

avtolstoy: The overhead should be fine - compared to recursive mutex acquisition and free list traversal a single xTaskGetSchedulerState is nothing.

:+1: on `PANIC(HeapError, "malloc while suspended");`

PR #2946, opened 2026-09-16
[gen4] rng: switch to secure DRBG; backup RAM via filesystem; last panic info over USB; littlefs truncate bugfix backport

avtolstoy:
### Problem

A few things in one branch:

1. On Gen 4 there is no hardware RNG peripheral, so the random source has to be provided
in software: this adds a mbedtls CTR_DRBG over an ADC-based entropy source.

2. Backup RAM was persisted to a dedicated 4K flash page with a full 4K shadow copy on every
sync, and `.retained_system` had no defined cold boot state - on a cold boot the region kept
whatever was left in RAM, stale DTLS session included.

3. The last panic info was only visible through cloud vitals, published once per panic, with
the assertion text as a pointer value (an int). Nothing could read it over USB, and nothing
could clear it.

4. The linked ELF artifacts carry a placeholder hash and CRC, so an ELF alone cannot be matched
against the module a device reports over USB, and any tool that wants to symbolicate an address
has to find the sibling binary first.

### Solution

1. `HAL_RNG_GetRandomNumber()` is backed by a mbedtls CTR_DRBG (entropy_len 128, nonce_len 0).
The DRBG is seeded from `hal_entropy_read()`, which harvests the least significant bits of ADC
channel 8 (internal ground) samples: the raw bits are packed as they come and conditioning is
left to the DRBG itself, but the source runs continuous health tests while harvesting - a
repetition count test (catches a stuck or short-period signal) and an adaptive proportion test
over a 1024 bit window (catches a biased signal), per NIST SP 800-90B. A harvest that trips
either test fails instead of feeding bad bits to the DRBG. The DRBG seed lives in
`.retained_system` in a consume-and-replace slot (safe against torn writes): warm boot =
retained seed + a small ADC refresh, cold boot = a full 128 byte harvest plus a forced backup
RAM sync. DRBG init failures panic at boot.
2. ADC access is arbitrated by a new `hal_adc_lock()` / `particle::AdcLock`, shared by the
entropy harvest and the regular `hal_adc_read/sleep/calibrate` paths.
3. Wi-Fi and BLE are routed to the DRBG: `rtw_get_random_bytes` is linked with `--wrap`, and
`platform_random` is implemented in `ble_hal.cpp`.
4. Backup RAM is persisted to `/sys/backup_ram.bin` via `SimpleFileStorage` - same plumbing as
the OTA state file - with a CRC over the region to skip unchanged writes. `initRetainedSystem()`
copies the initial values from the flash LMA on a cold boot. The retained data validity marker
is bumped, so a device that takes this update re-initializes the retained system area once on
the first boot after the update; user-facing retained variables keep their values across
resets as before. The 4K flash page stays reserved; freeing it is a flash map change for
another day.
5. Two new control requests: `CTRL_REQUEST_GET_LAST_PANIC_INFO` (101) and
`CTRL_REQUEST_CLEAR_LAST_PANIC_INFO` (102), defined in `control/diagnostics.proto` (already
released as `@particle/device-os-protobuf` 4.0.2). Unlike the cloud vitals path, the reply
carries the actual assertion text, not the pointer value. The get returns NOT_FOUND when the
device is clean; the clear invalidates the retained marker and syncs the filesystem copy, so
it survives a power loss on gen4.
6. The `%.bin` rule in `build/module.mk` now derives the module suffix/CRC offsets from the
linker symbols instead of asserting on a 38 byte placeholder at the end of the file, and
back-writes the real SHA-256 and CRC-32 into the ELF's `.module_info_suffix` /
`.module_info_crc` (and the legacy CRC section on gen3). Binary output is unchanged, the step
can run again against an already patched ELF, and `objcopy -O binary` on the patched ELF
reproduces the shipped binary. Verified against all 33 released 6.5.0 modules (9 platforms,
both generations).
7. Backported the upstream littlefs fix for the `lfs_cache_proc()` assert
(`pcache->block == 0xffffffff`) on truncate -> seek(0) -> write of a multi-block file. Upstream
fixed this in May 2019 (12e464e, "writes following a truncate"); our tree pre-dates it, and
`/sys/backup_ram.bin` (4100 bytes vs a 4096 byte block) is the first multi-block
`SimpleFileStorage` file, so the assert became reachable with item 4. The panic this used to
produce (SOS 10, lfs.c:183, during wiring/backup_ram) is what got the panic info work started.
8. The test runner application forwards unrecognized USB control requests to a weak
`test_app_ctrl_request_handler()` hook, and its request parser no longer modifies the request
buffer in place (the in-place parse NUL-terminates tokens inside the buffer, which broke the
re-parse in the hook).

### Test suites

- `wiring/rng_quality` - statistical analysis of ~1.5 Mbit pulled over USB (monobit, chi2,
longest run, entropy, Berlekamp-Massey linear complexity), plus reset-reason checks around the
soft reset and initial generation tests.
- `wiring/backup_ram` - persistence across hibernate, cold boot init, file recreation; tests
also check the reset reason, so a test that expected a reset cannot pass when the device
actually panicked.
- `wiring/panic_info` - announces a real `SPARK_ASSERT` panic over a new `PANIC_PENDING`
mailbox message; the runner expects the reboot, verifies the record, symbolicates and clears it.
- `FS_POSIX_03_TruncateRewrite` in `wiring/filesystem` - regression test for item 7 (truncating
an open multi-block file and rewriting it from the start, the `SimpleFileStorage` sequence).

### Steps to Test

Use test runner from: https://github.com/particle-iot/device-os-test-runner/pull/49

```
device-os-test run msom wiring/rng_quality wiring/backup_ram wiring/panic_info wiring/filesystem -v
```

What a panic looks like in a run now:
```
Device test notified about expected panic
Device panicked as expected: panic: code=10 pc=0x23fcfdc lr=0x23fbb51 extra=0x0 text="panic_info.cpp:31 false" registers=0x0, 0x0, 0x0, 0x0, 0x0, 0x0
pc: once at tests/integration/wiring/panic_info/panic_info.cpp:31
lr: TestOnce::loop at tests/libraries/unit-test/unit-test.cpp:260
Device test passed```


avtolstoy:
Superseded by #2947 #2948 #2949 #2950

PR #2948, opened 2026-09-17
[gen4] backup ram: persist via filesystem; initialize retained system on cold boot

avtolstoy:
### Problem

Backup RAM was persisted to a dedicated 4K flash page with a full 4K shadow copy on every
sync, and `.retained_system` had no defined cold boot state - on a cold boot the region kept
whatever was left in RAM, stale DTLS session included.

### Solution

Backup RAM is persisted to `/sys/backup_ram.bin` via `SimpleFileStorage` - same plumbing as
the OTA state file - with a CRC over the region to skip unchanged writes. `initRetainedSystem()`
copies the initial values from the flash LMA on a cold boot. The retained data validity marker
is bumped, so a device that takes this update re-initializes the retained system area once on
the first boot after the update; user-facing retained variables keep their values across
resets as before. The 4K flash page stays reserved; freeing it is a flash map change for
another day.

New test suite `wiring/backup_ram`: persistence across hibernate, cold boot init, file
recreation; tests also check the reset reason, so a test that expected a reset cannot pass
when the device actually panicked.

### Steps to Test

```
device-os-test run msom wiring/backup_ram -v
```


technobly (commented):
This seems primarily for SYSTEM retained memory, and the session data... but does it also affect USER retained memory?

Might be nice to add a short table to the PR similar to what is found in Docs: https://docs.particle.io/reference/device-os/api/retained-memory/

Also Docs may need to be updated since it currently says a System.reset() will not retain USER backup ram... or maybe it needs an asterisk clarifying things.

avtolstoy:
This affects all backup RAM as far as where it's stored when going into hibernate or when it is explicitly synced with `System.backupRamSync()`. There are no user facing behavior changes compared to previous releases.

The only changes are where its stored now: instead of a single flash page write/erase cycles spread through filesystem pages and system backup RAM (retained_system) is properly initialized on cold boot in the same manner as user backup ram (retained) -- again, doesn't affect customer apps.

technobly (approved):

PR #2949, opened 2026-09-17
[system] control: last panic info over USB; [build] back-write module hash and crc into elf

avtolstoy:
### Problem

The last panic info was only visible through cloud vitals, published once per panic, with
the assertion text as a pointer value (an int). Nothing could read it over USB, and nothing
could clear it.

The linked ELF artifacts carry a placeholder hash and CRC, so an ELF alone cannot be matched
against the module a device reports over USB, and any tool that wants to symbolicate an
address has to find the sibling binary first.

### Solution

1. Two new control requests: `CTRL_REQUEST_GET_LAST_PANIC_INFO` (101) and
`CTRL_REQUEST_CLEAR_LAST_PANIC_INFO` (102), defined in `control/diagnostics.proto` (already
released as `@particle/device-os-protobuf` 4.0.2). Unlike the cloud vitals path, the reply
carries the actual assertion text, not the pointer value. The get returns NOT_FOUND when the
device is clean; the clear invalidates the retained marker and syncs the persisted backup RAM
copy, so it survives a power loss on gen4.
2. The `%.bin` rule in `build/module.mk` now derives the module suffix/CRC offsets from the
linker symbols instead of asserting on a 38 byte placeholder at the end of the file, and
back-writes the real SHA-256 and CRC-32 into the ELF's `.module_info_suffix` /
`.module_info_crc` (and the legacy CRC section on gen3). Binary output is unchanged, the step
can run again against an already patched ELF, and `objcopy -O binary` on the patched ELF
reproduces the shipped binary. Verified against all 33 released 6.5.0 modules (9 platforms,
both generations).
3. New `PANIC_PENDING` mailbox type for test apps, and the `wiring/panic_info` suite: a test
announces a real `SPARK_ASSERT` panic via `PANIC_PENDING`, then panics; the test runner
(companion PR) expects the reboot, verifies the record, symbolicates and clears it.

Companion test runner PR: https://github.com/particle-iot/device-os-test-runner/pull/49

### Steps to Test

```
device-os-test --elf-dir=<release dir> run msom wiring/panic_info -v
```

What a panic looks like in a run (with the companion runner PR):

```
Device test notified about expected panic
Device panicked as expected: panic: code=10 pc=0x23fcfdc lr=0x23fbb51 extra=0x0 text="panic_info.cpp:31 false" registers=0x0, 0x0, 0x0, 0x0, 0x0, 0x0
pc: once at tests/integration/wiring/panic_info/panic_info.cpp:31
lr: TestOnce::loop at tests/libraries/unit-test/unit-test.cpp:260
Device test passed
```


technobly (approved):
Tested locally on b5som and failed 1 in 3 tries.  Will share the test logs with you directly.  I'll run this in a loop unattended and see if it fails again.

PR #2953, opened 2026-09-18
adds eSIM-aware IDLE cellular state [sc-139159]

technobly:
## Problem

An eSIM-equipped device with no enabled profile cannot register on any network. That is the state of a factory-fresh eUICC, and it is also the state whenever tooling disables a profile in order to replace it. The device reports the Default ICCID (`89000123456789012358`) while it is there, which is a useful signal but only a symptom.

Device OS does not know registration is impossible, so it treats the failure as temporary and keeps trying. The network interface loop runs every 100 ms with no backoff at all. Each failed attempt powers the modem off, and the next tick powers it straight back on. Meanwhile the modem-layer recovery paths keep firing, because as far as they can tell the client is still trying to connect.

The wasted power and the flooded logs are annoying. The real problem is that **tooling cannot manage eSIM profiles on the device**, because the modem keeps getting reset in the middle of the operation.

## Solution

We want a cellular state where the modem stays powered on and reachable over AT, but deliberately does not try to register. The device enters that state on its own when it finds no usable eSIM profile, and leaves it when the connection is cycled after a profile has been enabled.

## Success criteria

- With no enabled profile, the modem reaches the IDLE state during power-on and stays there. No registration is ever attempted, no resets, no power cycling, verified well past the ten-minute registration timeout.
- While IDLE the device issues no AT traffic at all, and the status LED breathes blue, the same as a radio that is powered on but not connecting. It does not sit on the blinking green which means "connecting".
- With an enabled profile, nothing changes at all. Registration, PPP, cloud connect and the CWM500 certification testing all behave exactly as they do today.
- After enabling a profile, a disconnect followed by a connect leaves IDLE cleanly and registers on the new profile, with no reset in between.
- **An error reading the eSIM never puts the device into IDLE.** A device with no eSIM at all, or one that fails the query transiently, behaves exactly as it does today.
- Tooling can do its APDU work while the device is IDLE without anything resetting underneath it.
- After a profile change, the cloud sees the new ICCID on the next connection.  Previously it would only update if the application changed.

## Test

- RUN integration test `system/cellular_esim` on HIL against all platforms and devices


hal/network/api/ifapi.h
```diff
@@ -174,6 +175,13 @@ typedef enum if_phy_state_t {
     IF_PHY_STATE_ON = 0x02
 } if_phy_state_t;
 
+/* Interface is powered and reachable but deliberately not trying to connect */
+typedef enum if_idle_state_t {
+    IF_IDLE_STATE_NONE = 0x00,
+    IF_IDLE_STATE_OFF = 0x01,
```
avtolstoy, line 181: Perhaps we can just add a new flag or xflag and just use that if it's just an indicator on/off state? Makes things simpler without having to add new event etc.

user/tests/integration/system/cellular_esim/cellular_esim.cpp
```diff
@@ -0,0 +1,854 @@
+/*
```
avtolstoy, line 1: Any way we can just reuse the hal/network/ncp_client/quectel/esim_profiles.cpp?

avtolstoy (commented):
just a couple of small things to consider while i do a ful review

user/tests/integration/system/cellular_esim/cellular_esim.cpp
technobly: Might be able to reuse some bits but not a ton. This will all be recycled when device os esim manager is integrated as well.  Integration tests need to do more as well like enable/disable profiles which is not available in esim_profiles.cpp.

hal/network/api/ifapi.h
technobly: Any suggestions on how to pass and process a flag separately from other `if_event`'s?  An event seems logical, but I understand it's a lot more code than just a flag.

currently using `PppNcpNetif::notifyIdleState(bool idle)` to pass.

avtolstoy: I kinda dug into this a bit more -- let's keep things as-is. flags/xflags are poll only right now, I thought we had events for the changes too - apparently not.

avtolstoy:
@technobly You can rebase on `develop` now, all Gen 3 platforms should be ok as far as flash space.

system/src/system_network_manager.cpp
```diff
@@ -714,6 +718,21 @@ void NetworkManager::handleIfPhyState(if_t iface, const struct if_event* ev) {
...
 }
 
+void NetworkManager::handleIfIdleState(if_t iface, const struct if_event* ev) {
+    /* An interface that is powered but deliberately not connecting. The state machine parks in
+     * IFACE_UP because the interface is administratively up, which would otherwise leave us on the
+     * blinking NETWORK_CONNECTING signal. Show the radio as on instead. */
+    if (ev->ev_idle_state->state == IF_IDLE_STATE_ON) {
+        // Another interface may be carrying traffic, and this signal shares BACKGROUND with the rest
+        // of the network ones, so it would take the front of the queue and hide them
+        if (!isConnectivityAvailable()) {
```
avtolstoy: I think this should be `if (!isConnectivityAvailable() && !isEstablishingConnections())`, otherwise we are going to go into `NETWORK_ON` signal on platforms with multiple network interfaces incorrectly when e.g. trying to connect to both cellular and wifi.

avtolstoy (approved):

system/src/system_network_manager.cpp
technobly: I ended up with this check for `no other iface up` instead of the `if (!isConnectivityAvailable() && !isEstablishingConnections()) {`

https://github.com/particle-iot/device-os/commit/5eb57b497251da39758b90bdd87a432cec2ca7c4#diff-436d70bc8202b80b11d8da611b9c6d40307975ba781f54e630a52d11f7357348R726-R739

Main reason is `!isEstablishingConenctions()` sounds like it might be what you want, but it is true when `IFACE_UP` which IDLE mode IS equal to IFACE_UP when we are connecting.  It's just parked there, with no `IP_CONFIGURED`. So really what we want is to breathe blue if IDLE and no other iface up which IFF_UP gives us.

Tested a few cases with ethernet (connected cable / disconnected cable / and eth disabled) + cellular in IDLE and the LED is what we want in all cases.

